PR #58, opened 2012-01-20
implement support for singleton resources for DS.RESTAdapter

siebertm:
In order to support working with singleton resources, I modified the RESTAdapter to take the different URL scheme into account for that.
While fiddling with the URLs, I also made the calculations of URLs consistent (e.g. deleteRecord used the primaryKey attribute, while all others hard-coded 'id' as primary key)

Defining a singleton resource is pretty easy:

``` javascript
App.User = DS.Model.extend();
App.User.reopenClass({
  isSingleton: true
})
```

This pull request is by no means complete, since IMO we need to discuss about the API for that.
- how do i find such a singleton resource? DS.Store doesn't know about singletons, so these don't work:

``` javascript
App.store.find(User) # no ID, no query
App.store.find(User, 1) # but the found user most likely isn't ID 1, so the mapping clientId <=> ID breaks
App.store.findAll(User) # would then return a ModelArray with exactly 1 model. works but sucks
```

Maybe we should introduce a findFirst() API method to DS.Store and Adapter?
- how do we cope with the bulk methods? just throw an error?
- no docs yet (until the other stuff is settled)

Just give me some feedback on how you think it should work, i'll fix that then

### DIFF
--- a/packages/ember-data/lib/adapters/rest_adapter.js
+++ b/packages/ember-data/lib/adapters/rest_adapter.js
@@ -2,12 +2,13 @@ var get = Ember.get, set = Ember.set, getPath = Ember.getPath;
 
 DS.RESTAdapter = DS.Adapter.extend({
   createRecord: function(store, type, model) {
-    var root = this.rootForType(type);
+    var root = this.rootForType(type),
+        url = this.urlForModel(model, type),
+        data = {};
 
-    var data = {};
     data[root] = get(model, 'data');
 
-    this.ajax("/" + this.pluralize(root), "POST", {
+    this.ajax(url, "POST", {
       data: data,
       success: function(json) {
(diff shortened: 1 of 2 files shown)


lessless:
any progress on retrieving singular resource? Definitely a 'must have' feature

tomdale:
@cerdiogenes I'm definitely open to revisiting it. The biggest two areas of concern for me are:

1. How do you declare that a model is a singleton?
2. How do you find singleton models? `store.find('singleton-type')` will do a find all.

Regarding item 2, we've had some discussion about having more explicit finders that don't depend on the type of arguments passed to determine behavior. @igorT, where are we with that?

wecc:
@tomdale Related to item 2: #2728

tomdale:
@wecc Perfect, thank you.

PR #750, opened 2013-02-22
Added support for polymorphic associations.

cyril-sf:
``` js
App.User = DS.Model.extend({
 messages: DS.hasMany(App.Message, {polymorphic: true})
});

App.Message = DS.Model.extend({
  created_at: DS.attr('date'),
  user: DS.belongsTo(App.User)
});

App.Post = App.Message.extend({
  title: DS.attr('string')
});

App.Comment = App.Message.extend({
  body: DS.attr('string'),
  message: DS.belongsTo(App.Message, {polymorphic: true})
});
```

You need to configure the serializer to map to the correct type:

``` js
DS.RESTAdapter.configure('App.Post' {
  alias: 'post'
});
DS.RESTAdapter.configure('App.Comment' {
  alias: 'comment'
});
```

The expected payload for a polymorphic association with the REST adapter/serializer should contain the type:

```
{
    user: {
        id: 3,
        // For a polymorphic hasMany
        messages: [
            {id: 1, type:"post"},
            {id:1, type: "comment"}
        ]
    },

    comment: {
        id: 1,
        // For a polymorphic belongsTo
        message_id: 1,
        message_type: "post"
    }
}
```

Initializing an alias will automatically allow you to sideload this
type. The support for `sideloadAs` is only for backward compatibility.

Two hooks exist for the serialization and materialization of an embedded
polymorphic association:
- `addType` that adds the record's type to the serialized data
- `extractEmbeddedType` that retrieves the record's type from the
  payload.

Both hooks rely in the JSONSerializer on `keyForEmbeddedType`, which returns the string used to
represent the record's type in the payload. The default value is `'type'`

### DIFF
--- a/packages/ember-data/lib/serializers/json_serializer.js
+++ b/packages/ember-data/lib/serializers/json_serializer.js
@@ -13,8 +13,6 @@ DS.JSONSerializer = DS.Serializer.extend({
       this.set('transforms', DS.JSONTransforms);
     }
 
-    this.sideloadMapping = Ember.Map.create();
-
     this.configure({
       meta: 'meta',
       since: 'since'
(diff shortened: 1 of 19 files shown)


lolmaus:
Is this feature documented? I struggle to understand how it works, but i failed to find a single working JSBin demo.

lolmaus:
Okay, i got it working for a polymorphic hasMany: http://emberjs.jsbin.com/rivav/4/

But it doesn't seem to work with the [DS.EmbeddedRecordsMixin](http://emberjs.com/api/data/classes/DS.EmbeddedRecordsMixin.html): http://emberjs.jsbin.com/rivav/7/edit?html,js,output

DS.EmbeddedRecordsMixin is a native feature of Ember Data and is well-documented. I think that lack of support for it is a bug in this PR's implementation. Do you agree?

Issue #1111, opened 2013-08-07
Pluralize with same singular and plural causing assertion to break


dagda1:
If I have a class called App.UserSettings which is the same singular and plural and I define a plural like this:

```
App.RESTAdapter.configure('plurals',{
  user_settings: 'user_settings'
});
```

Then this assertion fails in _pluralizeAliases:

```
    aliases.forEach(function(key, type) {
      plural = self.pluralize(key);
      Ember.assert("The '" + key + "' alias has already been defined", !aliases.get(plural));
      aliases.set(plural, type);
    });
```

PR #1112, opened 2013-08-08
check key and plural match in _pluarlizeAliases

dagda1:
antidote to #1111


0x414f:
@wycats I ask because I'm not sure: why wasn't this relevant anymore at the time? (Some of us have been meaning to upgrade for a long time... :fearful:)

PR #1176, opened 2013-09-03
Support for manyToNone and manyToMany relationships

ssured:
Current implementation discards manyToNone and manyToMany relationships as it does not serialize any hasMany relations.

### DIFF
--- a/packages/ember-data/lib/serializers/json_serializer.js
+++ b/packages/ember-data/lib/serializers/json_serializer.js
@@ -120,7 +120,16 @@ DS.JSONSerializer = Ember.Object.extend({
     }
   },
 
-  serializeHasMany: Ember.K,
+  serializeHasMany: function(record, json, relationship) {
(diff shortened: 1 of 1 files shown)


tomichal:
@wycats @ssured Curious, why not also serialize the 'ManyToOne' by default?
(this code carried through all the way to https://github.com/emberjs/data/blob/7662ae859471a6529078674bbd9a9814d59490fe/packages/serializer/addon/json.js#L947)
Sorry if I'm missing something obvious!

runspired:
@tomichal I suspect the reasoning was around the 2 generals problem and playing the odds for what would be correct for most applications.

In a `ManyToNone` and `ManyToMany` situation you can bet pretty well on the API expecting to make patches/updates via the resource vs via a different API or the opposite side of the relationship.

In a `ManyToOne` situation it's more common for the single-side of the collection to be the "owner" and make the API request.

PR #1243, opened 2013-09-08
Failing test to demonstrate that changing a belongsTo relationship on a record doesn't dirty the record

robmonie:
It also doesn't update the underlying `data` hash. It does superficially update the relationship in so much as asking for the value of the property after setting the belongsTo relationship will return the record you set it to.

It appears that the belongsToWillChange and belongsToDidChange observers are not being fired correctly.

I noticed this in the pre-beta build as well and it appears to have been around for a while although I only noticed it recently so it's possible it's got something to do with a recent change to ember core around observers?

### DIFF
--- a/packages/ember-data/tests/unit/model/relationships_test.js
+++ b/packages/ember-data/tests/unit/model/relationships_test.js
@@ -659,7 +659,7 @@ test("changing a belongsTo relationship marks the record as dirty", function() {
 
   var Person = DS.Model.extend({
     name: DS.attr('string'),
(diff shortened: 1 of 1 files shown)


dgtman:
@wycats This was closed back in 2013 as a forthcoming refactor was supposed to address this.

However, as of ED 1.0.0-beta.14.1, this is still a problem today.  Is there a roadmap for when we might see this fix?

wecc:
@dgtman There's a discussion about this over at https://github.com/emberjs/rfcs/pull/21

Issue #1362, opened 2013-09-25
DS.hasMany with { async: true } errors out when there are duplicate IDs.


ohmslaw:
Let's say I have a pre-defined set of items, of which some are being added to a cart.  I am trying to keep the DB relatively denormalized so I would like to add the same item to the cart multiple times (instead of adding an intermediary object).  Works perfectly fine during the add/save (because the items are pre-loaded) but started running into a problem on refresh.  I mark the items as `DS.hasMany('item', { async: true })` so they could be loaded if needed.  During asynchronous loading the first duplicate will error out because it is already in a `loading` state.

Because of that if I side-load every item we have no problems, but if your user has 20 carts and each cart has 50 items that payload would larger then desired.  In my case the user will only ever select one cart to modify at a time so all the items would not need to be loaded up front.

PR #1382, opened 2013-09-28
Looking at the error caused by loading two items with the same id asynch...

ohmslaw:
I know it does not really fit the tests but I wanted to demonstrate the error I was seeing in https://github.com/emberjs/data/issues/1362.  I have not had time to familiarize myself with all the new paradigms in ember-data (new job & new twins) so I would not be shocked to learn I might just be implementing something wrong.

I believe this will run that test: [http://localhost:9292/?package=ember-data&testNumber=188](http://localhost:9292/?package=ember-data&testNumber=188)


wecc:
This does not seem to be an issue anymore. I've added a slightly modified test in #2774 and it passes.

PR #1436, opened 2013-10-08
Returning existing records in the store immediately when calling store.find()

diogomafra:
The method `find()` in ember-data 0.13 returns the existing data in the store and updates them when the request is completed. With ember-data 1.0 the `find` method is returning an empty array.

It's very useful to return the cached data because the app can display the existing records and update them in the background.


wecc:
Related: https://github.com/emberjs/rfcs/pull/27

PR #1491, opened 2013-11-06
Allow for polymorphic async relations

ssured:
Async relations are wrapped in a DS.PromiseObject, if so, unwrap before querying the typeKey


igorT:
Should be fixed by #2623

PR #1500, opened 2013-11-13
Added support for serializing polymorphic hasmany-to-none relations

ssured:


igorT:
We have this support now, closing

PR #1517, opened 2013-11-19
Pagination

aexmachina:
This PR adds the foundation for pagination support in Ember Data. Example usage using an 'infinite-scroll' `loadMore` approach can be [seen here](http://emberjs.jsbin.com/cofi/latest/edit). The goal of this PR is to provide the underlying requirements to support pagination, so that the community can iterate on rest of the implementation in userland until we've got something we're happy with.

The solution adds a `DS.Request` class to contain information about the parameters that were used to create a given `AdapterPopulatedArray`, thereby allowing for more pages of results to be fetched from the server.

I've folded the `sinceToken` into the `Request` class, but this is the only 'breaking' change. So `Adapter.findAll: function(store, type, sinceToken)` is now `Adapter.findAll: function(store, type, request)`.

Please feel free to offer any feedback - I've tried to stay close to the style of the rest of the project, but happy to take on board any feedback.

One possible concern is that I'm using `page` and `pageSize` parameters which is nice and simple but there's an edge case where the `pageSize` changes and leads to duplicate/missing records in the series of pages. I'd prefer to use `offset` and `limit`, but the common practice seems to be `page` and `pageSize` so I've used this approach instead.

# Todo

- Currently only `findQuery()` and `findAll()` support pagination, I'll add support for the other `find...()` methods now

### DIFF
--- a/packages/ember-data/lib/adapters/fixture_adapter.js
+++ b/packages/ember-data/lib/adapters/fixture_adapter.js
@@ -166,8 +166,9 @@ DS.FixtureAdapter = DS.Adapter.extend({
   },
 
   handlePagination: function(array, request) {
-    if( request.page && request.pageSize ) {
-      var start = (request.page - 1) * request.pageSize;
+    if( request.pageSize ) {
+      var page = request.page || 1,
+          start = (page - 1) * request.pageSize;
       return array.slice(start, start + request.pageSize);
     }
     else {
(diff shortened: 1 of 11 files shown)


igorT:
Just talked to @aexmachina, we will be using the new JSONApi meta for pagination, so closing this PR.
@wecc this PR would be good to review before starting work on pagination

oligriffiths:
@aexmachina What's the state of play with this?

aexmachina:
Long stale by now :)

PR #1656, opened 2014-01-13
Fixes hasMany records being duplicated after payload returns from server

kurko:
There is an edge case where a hasMany record gets duplicated (same ID).
Basically, if we generate a record locally using UUID or any custom ID
format, then push it to production which will return the same ID
previously defined, the store's array that holds records to be resolved
is going to duplicate both items, even if they have the same ID.

This commit fixes that by not allowing duplicated IDs in that array.

I need this badly for my app.

### DIFF
--- a/packages/ember-data/lib/system/store.js
+++ b/packages/ember-data/lib/system/store.js
@@ -1553,6 +1553,7 @@ function normalizeRelationships(store, type, data, record) {
     if (kind === 'belongsTo') {
       deserializeRecordId(store, data, key, relationship, value);
     } else if (kind === 'hasMany') {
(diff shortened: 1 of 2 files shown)


angusfretwell:
It seems like this issue has regressed and is present in 1.13.14

Znegl:
Getting the same bug with null IDs:
Saving new embedded hasMany records results in duplicate content in the frontend. One if the items in the frontend is marked as isNew.

Ember : 2.3.0
Ember Data : 2.3.3

kurko:
FYI @igorT fixed this in another branch 1,5h years ago. Given this is a PR, I suggest opening a new issue and linking back to here.

PR #1740, opened 2014-02-11
Add support for passing a query of type array to store.find.

arjand:
Since the query is passed directly to jQuery, this array follows
the format as specified by jQuery.param.

Use Case:
I need to send multiple filter requests to the server. Since the query is passed to jQuery, jQuery.param() is responsible for serializing the array. http://api.jquery.com/jQuery.param/
The example I added to the documentation follows jQuery's serializeArray() convention.


wecc:

igorT:
We have deprecated the overloaded find, and now you should be able to pass an array to store.query

PR #1826, opened 2014-03-21
Extract the filtering of fixtures to easily mimic server retrieval schemes

jwright:
Sometimes you want to find a record on the server with keys other than the ids. 

Here is an example:

```js
App.BlogRoute = Ember.Route.extend({
  model: function(params) {
    return this.store.find("blog", params.slug);
  }
});
```

Your API on the server should know to expect a `slug` as the parameter and retrieve the `Blog` object with that slug.

If you want to mimic this in your `FIXTURES`, you have to override the entire `find` function. This PR extracts the filtering for the fixture ids so you can override a simpler, more concise function that the `find` function calls.

This is what you would need for your `FixtureAdapter`.

```js
BlogAdapter = DS.FixtureAdapter.extend({
  filterFixturesByIds: function(fixtures, ids) {
    return fixtures.filter(function(blog) {
      return Ember.EnumerableUtils.indexOf(ids, blog.slug) !== -1;
    });
  }
});
```

### DIFF
--- a/packages/ember-data/lib/adapters/fixture_adapter.js
+++ b/packages/ember-data/lib/adapters/fixture_adapter.js
@@ -303,13 +303,7 @@ var FixtureAdapter = Adapter.extend({
     @param id
   */
   findFixtureById: function(fixtures, id) {
-    return Ember.A(fixtures).find(function(r) {
-      if(''+get(r, 'id') === ''+id) {
-        return true;
-      } else {
-        return false;
-      }
-    });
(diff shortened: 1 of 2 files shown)


bmac:
Hi @jwright. Thanks for this pr. We've deprecated the fixture adapter and moved it into its own Ember Addon found here: https://github.com/emberjs/ember-data-fixture-adapter.

I'm going to close this pr but feel free to reopen it against the new repo.

PR #1831, opened 2014-03-23
Zombie record when id query mismatch

bcardarella:
If I query for a record of id `1` and the server returns a record of id
`2` the store will have records of id `1` and id `2` as Ember Data
pre-emptively instantiates the record you are querying for.

A common use-case where this occurs is when querying for records with
a slug for an id. If I have a slug library server side for a Rails
application I do not want to just map my id column for the slug because
all of the relationships and foreign keys in my database are using the
auto-incrementing id column, not the slugs. The slugs are there for
presentation purposes only. However, on the client I will sometimes only
have a slug as the unique identifier to search for a given record on the
server.

This PR will do the following:

If you query the server for id `1` and it returns a record of id `2` it
will delete the previously instantiated record of id `1`.

If you already have a pre-existing record of id `1` that is committed
and query the server for id `1` and the server returns a record of id
`2` it will **not** delete the record of id `1`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -80,6 +80,8 @@
 * move some requires around
 * move grunt tasks into folders
 * ES6!
+* Ensure place-holder records do are removed after querying and an id
+  mismatch
 
 ### Ember Data 1.0.0-beta.6 _(January 25, 2014)_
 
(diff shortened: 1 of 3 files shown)


igorT:
We now have a solution for this, @bmac has a PR, https://github.com/emberjs/data/pull/3198 so closing this PR

PR #1832, opened 2014-03-24
Add force option to store find method

bcardarella:
Passing `true` as the last argument on find will always delegate the
record lookup to the adapter's find method rather than relying on the
local store's copy.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -80,6 +80,9 @@
 * move some requires around
 * move grunt tasks into folders
 * ES6!
+* Add `force` option to the store's `find` method which will bypass a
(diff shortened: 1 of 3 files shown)


wecc:
Related: https://github.com/emberjs/rfcs/pull/27

bmac:
Thanks for the pr @bcardarella. I'm going to close this since I believe [fetchById](http://emberjs.com/api/data/classes/DS.Store.html#method_fetchById) now solves this issue.

igorT:
Keeping this as a use case, as `findById` wouldn't take a force option right now afaik

igorT:
Nm, `store.fetchById` solves this problem

PR #1842, opened 2014-03-28
allow store to deserialize polymorphic relationships

bbanach:
With this change, no need to change the JSON structure to have all objects at the root, with referential ids in the value of the parent attributes.

I feel this is helpful in the fact that you don't need to create extra serializers to overwrite the extractSingle function in order to change traditional JSON to what Ember-data expects.  However, if you prefer to create specific serializers that will tranform your payload, this is a non-breaking change.


bmac:
Closing this pr as I believe the original issue has been solved by changes to the `EmbeddedRecordsMixin` because it no-longer tries to manipulate the data into a namespaced object.

PR #1854, opened 2014-04-06
Use transform to store attribute values in deserialized format.

rwjblue:
This forces the attribute values in `_attributes` to be stored in the deserialized format.

This essentially makes the internal values much more consistent. Currently if you bind a non-string value in a template (`boolean`, `date`, or `number`) the internal attribute is represented in string form (since `{{input}}` will always be a string). This is an issue if you use `Em.computed`'s on the attribute that expect a given attribute format.

For instance:

Storing a `count` field (`DS.attr('number')`), and using an `Em.computed.sum` accros multiple records starts simply appending strings after one is set with an `{{input}}`.

----

There may be a wonderful reason that this is not done at the moment, but it wasn't obvious to me.

### DIFF
--- a/packages/ember-data/lib/system/model/attributes.js
+++ b/packages/ember-data/lib/system/model/attributes.js
@@ -278,6 +278,11 @@ function attr(type, options) {
       Ember.assert("You may not set `id` as an attribute on your model. Please remove any lines that look like: `id: DS.attr('<type>')` from " + this.constructor.toString(), key !== 'id');
       var oldValue = getValue(this, key);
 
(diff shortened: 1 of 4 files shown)


xtian:
I ran into this issue recently as well. Anything I can do to help get this merged or investigate potential issues?

igorT:
I think we decided that this path is too hard to do properly and that transforming data coming into the record from users code is really user's responsibility. I am not sure that doing this automagically at the record layer even works properly, as it would probably create issue in different parts of your app, for example if you transform a select value without your select knowing about it, seems like your selection could easily get mismatched with the backing data.

PR #1869, opened 2014-04-11
Add a default filter function if none provided

alexspeller:
This adds a default filter function to `DS.Store#filter` which does the thing it will most commonly be used for anyway.


igorT:
I still think this is way to automagic and scary, and pretty easy to do if you need to yourself. Would be a great addon though

PR #1994, opened 2014-06-01
Honour the processData flag in RESTAdapter#ajaxOptions.

drwrf:
This change allows users to pass the `processData` flag to the `RESTAdapter#ajaxOptions`. This flag ultimately tells jQuery that it shouldn't perform any transformations to the data passed to the ajax request.

Previously, ember-data didn't honour this flag and would blindly `JSON.stringify` any data passed along for the request. This is normally the exact behaviour you want, but it prevents passing strings or instances of `FormData`, which is useful for file uploads and custom data you want to pass to the server.

Now, if `processData: false` is passed, then ember-data will not perform any transformations to the data object passed. It will also not assume that the content-type is `application/json`.

### DIFF
--- a/packages/ember-data/lib/adapters/rest_adapter.js
+++ b/packages/ember-data/lib/adapters/rest_adapter.js
@@ -632,14 +632,9 @@ var RESTAdapter = Adapter.extend({
     hash.dataType = 'json';
     hash.context = this;
 
-    if (type !== 'GET') {
-      if (FormData && hash.data instanceof FormData) {
-        hash.contentType = false;
-        hash.processData = false;
-      } else if (hash.data) {
-        hash.contentType = 'application/json; charset=utf-8';
-        hash.data = JSON.stringify(hash.data);
-      }
+    if (hash.data && hash.processData !== false && type !== 'GET') {
+      hash.contentType = 'application/json; charset=utf-8';
+      hash.data = JSON.stringify(hash.data);
     }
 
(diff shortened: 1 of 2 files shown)


davidrivera:
Any eta on this guys? Would really like to stop encoding my images in base64 just to accommodate this.

fivetanley:
How are you getting the form data into the adapter/serializer?

fivetanley:
@dwrf We think this is fine to merge, but would like to do the right thing automatically. (see @tomdale's comments). We'd like to special case FormData/others automatically, but we'd like to understand this use case better.

drwrf:
@tomdale Some of the original commits in the PR would go about detecting `FormData` and automatically do the right thing. I can resurrect that code but also continuing to honour `processData`. 

I get what you're saying about this being somewhat "insider baseball". It's really that this `ajax` method is kind of half leaking the fact that it's just calling `jQuery.ajax` under the hood, but also half hiding it away. I opted to continue with jQuery's conventions for consistency's sake but maybe that's not the goal. What do you think?

bmac:
@drwrf With this pr I'm still unsure of how a `FormData` object will end up in the adapter. Are using a custom serializer? Ideally when documenting this feature Ember Data would like to show the full flow of how a user can take advantage of this flag.

bmac:
I'm going to close this issue since there hasn't been much activity on it in a while. @drwrf feel free to re-open if you are still interested in this feature.

PR #2015, opened 2014-06-07
Add option for custom serializer to `Model#serialize`

tricknotes:
I want to switch serializer on special context.

For example:
* admin can create `Post` with original `Category`.
(This case `Post` will be serialized embedded `Category` **records**.)

* user can only create `Post` with chosen `Category`.
(This case `Post` will be serialized embedded `Category` **ids**.)


tricknotes:
Thanks @tomdale .
I agreed.

Certainly, this is a less commonly case.
I thinks it has no necessary to support flag for custom behavior.

PR #2038, opened 2014-06-18
[DOC] Update intro to FixtureAdapter

knownasilya:
Update some of the wording. Mentioned that it expects normalized data and a note about it not having a serializer.

Also added a blurb about it being async by default, and how to prevent that with an example.


bmac:
Thanks for the update @knownasilya. I'm going to close this for now since the Fixture adapter is being moved to an addon. If you'd like you can retarget against the https://github.com/emberjs/ember-data-fixture-adapter repo.

PR #2040, opened 2014-06-19
Store#findHasMany return value and documentation fix

watsonbox:
The documentation for `Store#findHasMany` was incorrectly changed in https://github.com/emberjs/data/commit/7ab338bea001437002b05a1965367e3e94c404c8.

I then realized that the intended behavior didn't work anyway so fixed that by returning the retrieved records from `_findHasMany`.


wecc:
`store.findHasMany()` has been changed (and correctly documented) in https://github.com/emberjs/data/commit/c9825d4a022dadf7f9cf17922dcfb3f228f8af0e

sly7-7:
Thanks @wecc for the heads up :+1:

Issue #107, opened 2012-02-21
read-only attributes


NickFranceschina:
would like short-cut to define a property as "read-only" without having to make computed property

```
email : DS.attr('string',{readOnly:true})
```

PR #2106, opened 2014-07-22
allow attributes to be excluded via the attrs hash

fivetanley:
(Reopening in hopes that travis will run it correctly this time).

refs #303

As per a discussion I had with @igort and @tomdale yesterday in #emberjs-dev:

This pull request allows you to declaratively (instead of overriding serializeAttribute and using `this._super`) not send attributes to the server when serializing.

So you can now say:

``` javascript
App.PersonSerializer = DS.JSONSerializer.extend({
  attrs: {
    admin: { exclude: true }
  }
});
```

and `admin` will be removed from the payload when serialized to the server.

I'm also thinking it would be handy to allow some sort of array syntax for declaring multiple properties if you have a lot:

``` javascript
App.PersonSerializer = DS.JSONSerializer.extend({
  excludedAttributes: [ 'admin' ]
});
```

Thoughts?

### DIFF
--- a/packages/ember-data/lib/serializers/json_serializer.js
+++ b/packages/ember-data/lib/serializers/json_serializer.js
@@ -67,6 +67,30 @@ export default Ember.Object.extend({
     });
     ```
 
+    You can also remove attributes by setting the `serialize` key to
+    false in your mapping object.
+
+    Example
+
+    ```javascript
+    App.PersonSerializer = DS.JSONSerializer.extend({
+      attrs: {
+        admin: {serialize: false},
+        occupation: {key: 'career'}
+      }
+    });
+    ```
+
+    When serialized:
+
+    ```javascript
+    {
+      "career": "magician"
+    }
+    ```
+
+    Note that the `admin` is now not included in the payload.
+
     @property attrs
     @type {Object}
   */
@@ -436,6 +460,11 @@ export default Ember.Object.extend({
       value = transform.serialize(value);
     }
 
+    // If attrs.key.serialize is false, do not include the value in the
+    // response to the server at all.
+    if (attrs && attrs[key] && attrs[key].serialize === false) {
+      return;
+    }
     // if provided, use the mapping provided by `attrs` in
     // the serializer
     if (attrs && attrs[key]) {
--- a/packages/ember-data/tests/integration/serializers/json_serializer_test.js
+++ b/packages/ember-data/tests/integration/serializers/json_serializer_test.js
@@ -179,6 +179,22 @@ test('Serializer should respect the attrs hash when serializing records', functi
   equal(payload.title_payload_key, "Rails is omakase");
 });
 
+test('Serializer respects `serialize: false` on the attrs hash', function(){
+  expect(2);
+  env.container.register("serializer:post", DS.JSONSerializer.extend({
+    attrs: {
+      title: {serialize: false}
+    }
+  }));
+
+  post = env.store.createRecord("post", { title: "Rails is omakase"});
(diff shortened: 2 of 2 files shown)


maxkwallace:
Does this still work in Ember Data 2? I'm using `"ember-data": "2.5.0-beta.1"`, and I've tried both `readOnly: false` and `serialize: false` without success.

sandstrom:
@maxkwallace it should. You'd use it sort of like this:

```js
import ApplicationSerializer from './application';

export default ApplicationSerializer.extend({
  attrs: {
    firstName: { serialize: false }
  }
})
```

(in e.g. `app/serializers/user.js`)

maxkwallace:
@sandstrom Thanks for the reply! I found that the follow does indeed work for my use case:

```
import DS from 'ember-data';
  attrs: {
    firstName: { serialize: false }
  },
});
```

(This was my mistake-- I was trying to disable serialization in the model itself using `firstName:   DS.attr('string', { serialize: false }),` rather than in a serializer.

pangratz:
> This was my mistake-- I was trying to disable serialization in the model itself using firstName: DS.attr('string', { serialize: false }), rather than in a serializer.

@maxkwallace is this something which has been wrongly documented somewhere? If so, this should definitely be fixed.

maxkwallace:
@pangratz not that I'm aware of. In terms of documentation for this issue, this right answer is out there, but it takes some work to find. I was in a hurry and didn't read through everything before posting.

When I search for "ember data read only attribute", here are the results that come up, in order ("correct answer" means sandstrom's recommendation):

1. https://github.com/emberjs/data/issues/107 (From 2012, first post proposes incorrect syntax, links to https://github.com/emberjs/data/pull/303 which shows the reference from this thread at the bottom, but doesn't make it explicit that that's where you should go for the resolution.)

2. [Tom's proposal from Dec. 2012](https://gist.github.com/tomdale/4263171) (Obviously very out of date, no link to the current solution.)

3. http://stackoverflow.com/questions/32829966/how-to-prevent-ember-data-from-saving-attribute-ie-read-only-attribute (Contains the correct answer!)

4. http://stackoverflow.com/questions/16588286/non-persistent-attributes-in-emberjs (Four different answers, one is sort-of correct, but it's not clear that that's the case.)

5. http://discuss.emberjs.com/t/read-only-attributes-in-ember-data-1-0-0-beta/2646/5 (Hasn't been updated since Nov. 2013.)

So there's a lot of noise and outdated info. I'm more than happy to go back and add clarifying links/info on the correct solution for 1, 2, 4 and 5. Would that be helpful? Is there a recommended best practice for this sort of thing?

coryondemand:
One thing about disabling in the serializer is that you will get a warning from models that do not use the attribute. For example, I don't want to serialize the attribute "history" on any of my models, but not all of my models have a history attribute. There are a few work arounds, but I just deal with the warnings.

WARNING: There is no attribute or relationship with the name `history` on `line-item`. Check your serializers attrs hash.

pixelhandler:
@coryondemand perhaps make an mixin and setup a few per-type serializers using that mixin to disable serializing the common attribute for those types.

pangratz:
@coryondemand you can also overwrite `serializeAttribute` in your application serializer and don't serialize when the attributes' name is `history`, see [this ember-twiddle](https://ember-twiddle.com/341469d996459f86b0213816b36e2f35?fileTreeShown=false&numColumns=3&openFiles=routes.application.js%2Ctemplates.application.hbs%2Cserializers.application.js) as an example.

Issue #1950, opened 2014-05-17
EmbeddedRecords with ActiveModelSerializer should append _attributes to keys


mehulkar:
When I embed records for a Rails backend, it's usually because I'm using the `accepts_nested_attributes` feature of ActiveRecord. Rails expects that embedded records will come in with the key `[plural_model_name]_attributes`. 

Example:

```json
{
  "author": {
      "name": "Mark Twain",
      "books_attributes": [{"id": "1", "title": "Tom Sawyer"}]
  }
}
```

The solution would be to do modify `serializeHasMany` in EmbeddedRecordsMixin to check if the ActiveModelSerializer and append the string.

Opening up this issue to see if this an appropriate addition. To implement, I'm not sure how to check what type of the ActiveModelSerializer (or an extension of it) is being used.

arnaldo2792:
May be I'm wrong, but I think that this issue should be solved by what you guys said in [#2138](https://github.com/emberjs/data/pull/2138)

amedr:
Yeah I actually opted for other approaches which actually look cleaner and stop needing this.

igorT:
This should now be much easier to customize, due to #2790 so you can customize the key in one direction. Maybe a change to AMS default(if we added keyForEmbeddedRecord?) or an addon would be great to get this out of the box.

mmahalwy:
Sorry guys, but any solution to this?

PR #2138, opened 2014-07-30
Follow rails conventions for serializing embedded objects

opsb:
In https://github.com/emberjs/data/issues/1950 there was a lot of talk about how people expected the embedded objects to be serialized with _attributes when they're embedded (as per the rails convention). This PR will add _attributes to any embedded objects when the EmbeddedRecordsMixin is used.

It does break the existing behaviour but I think it's fair to say the existing behaviour is broken for rails. If someone does need the existing behaviour they can revert to it by overriding the formatEmbeddedKey method e.g.

    App.ApplicationSerializer = DS.ActiveModelSerializer.extend(DS.EmbeddedRecordsMixin, {
      formatEmbeddedKey: function(key){
        return key;
      },

      attrs: {
        video: { embedded: 'always' }
      }
    });

### DIFF
--- a/packages/activemodel-adapter/lib/system/active_model_serializer.js
+++ b/packages/activemodel-adapter/lib/system/active_model_serializer.js
@@ -111,6 +111,10 @@ var ActiveModelSerializer = RESTSerializer.extend({
     return decamelize(attr);
   },
 
+  formatEmbeddedKey: function(key){
+    return key + "_attributes";
+  },
+
   /**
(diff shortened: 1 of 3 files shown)


tomdale:
Rather than making big breaking changes to fit the Rails case exclusively, let's leave this up to addons to experiment with. Ideally, people can use JSON API easily on both ends and won't need to worry about this case.

igorT:
This should be much easier to customize now due to #2790

PR #2145, opened 2014-07-31
Move key lookup for ajaxError into model and serializer

tpitale:
Moves the responsibility of looking up the key to
be used for error <=> attribute mapping into the
model's adapterDidInvalidate addError behavior and
uses the serializer to find errorKeyForAttribute/
Relationship to do the key conversion using similar
logic to serialization since the errors returned
from the active model adapter ajax call will have
keys that match the serialized form.

This has the added benefit of allowing a serializer
to override this behavior to perform key name
mapping much the way normalizeHash does inside
the serializer.


igorT:
I believe error normalization has been in the serializer and working for a while now, so closing this PR. Thanks!

Issue #2162, opened 2014-08-05
findHasMany should be called instead of findMany if a has-many relationship is undefined


tomdale:
Given a payload for a particular record, Ember Data currently resolves has-many relationships through the adapter using the following heuristic:
1. If the relationship is defined in the top-level `links` object, the store will call the adapter's `findHasMany` hook.
2. If the relationship is _not_ defined in the top-level `links` object, the store will call the adapter's `findMany` hook, passing it the array of IDs provided in the payload after removing the IDs of any records that are already loaded.

This behavior works well when the relationship is provided as a URL in the `links` object, or when the relationship is provided as an array of IDs.

However, there is a third, very common case, where the relationship should be populated by fetching a URL, but the URL is derived from information contained in the record, rather than being provided in the payload, hypermedia-style.

For example, imagine the following model:

``` js
// models/post.js
var attr = DS.attr,
    hasMany = DS.hasMany;
export default DS.Model.extend({
  title: attr(),
  body: attr(),
  comments: hasMany()
});
```

In this example, materializing the comments relationship by calling `post.get('comments')` should result in the client retrieving `/post/123/comments`. Thus, the server sends the following payload for the post with ID 123:

``` js
{
  "id": 123,
  "title": "Rails is Omakase",
  "body": "Trololol"
  // Note lack of comments relationship
}
```

This case is extraordinarily common (I ran into it when writing an app that uses the Portland public transit API, and Django's REST framework uses this approach), but currently getting this to work requires synthesizing a fake `links` entry in the serializer's `normalizePayload` hook and filling it with garbage data, which is a huge PITA.

@wycats and I think that it may be appropriate to add a third hook, which generates a URL for the relationship, and if that returns a value, the store will call `findHasMany` with the result. The current `links` behavior can be explained and implemented in terms of this new hook.

psteininger:
@tomdale @igorT has this been resolved in 2.0+ ? I have a use case where the nested route produces aggregate level data, from a large dataset.
```
/widgets/:id/scores_by_month 
```

amiel:
Having this would be super helpful for https://github.com/amiel/ember-data-url-templates, particularly to support the case @psteininger is talking about.

Is anyone else working on this? and is it still desired by others?

I'd be happy to give it a shot...

amiel:
[...]

PR #2168, opened 2014-08-07
Introduce buildURLForHasMany hook

tomdale:
**Not ready for merge**

This introduces a new hook, `buildURLForHasMany`, to address the issues discussed in #2162.

This commit slightly changes the behavior of relationship resolution, but I've tried to keep the changes as unobtrusive as possible.

Specifically:

1. If a relationship is provided as an array of IDs, or `null`, the adapter's `findMany` hook will be called (after filtering out already-loaded records).
2. If a relationship is provided via `{ links: { relationshipName: '/foo' } }`, the value (usually a URL) will be passed to the adapter's `findHasMany` hook.
3. If a relationship is not provided (i.e., it resolves to `undefined`), then the adapter's `buildURLForHasMany` hook will be called. The value returned from that hook will be passed to `findHasMany`, as though it were provided in the `links` object.

Implementation was relatively straightforward, but the feature does raise some questions.

Namely, what should the default serialization be for `hasMany` relationships? We have three common representations for inbound JSON: inline array, `links` URL, or implicit URL generated by the adapter. Applications may mix and match two or more of these forms. What should we send back to the server when the record is saved?

The easy answer is: use the payload provided by the server as a heuristic. If the relationship is specified as an inline array of IDs, send that back. Otherwise, if the relationship is a URL (implicit or explicit via `links`), send nothing.

However, this approach falls flat when we consider the case of client-created records, which may be created and saved before we ever see an exemplar payload from the server. Additionally, many APIs send JSON different than what they expect to receive during a save.

Finally, what happens when a has-many relationship is specified by a foreign key on the belongs-to side? If a `Post` has many `Comment`s, and the contents of the `comments` ManyArray is populated by loading `Comment` records that have a `post_id` field, the `Post` payload will not have a `comments` field, which will trigger the `buildURLForHasMany` hook. I remain unsure how to reconcile this hook with future SSOT plans.

We could disable the `buildURLForHasMany` hook in the case that we see a foreign key for the relationship, but that means that load order affects the way relationships are resolved. This feels both confusing and difficult to debug.

Ultimately, the more I try to untangle this problem, the more I believe that we want a DSL for concisely expressing serialization rules, without requiring users to write their own custom serializer for these common cases.

### DIFF
--- a/packages/ember-data/lib/adapters/rest_adapter.js
+++ b/packages/ember-data/lib/adapters/rest_adapter.js
@@ -508,6 +508,7 @@ export default Adapter.extend({
     //We might get passed in an array of ids from findMany
     //in which case we don't want to modify the url, as the
     //ids will be passed in through a query param
+
     if (id && !Ember.isArray(id)) { url.push(id); }
 
     if (prefix) { url.unshift(prefix); }
--- a/packages/ember-data/lib/system/relationships/has_many.js
+++ b/packages/ember-data/lib/system/relationships/has_many.js
@@ -52,14 +52,36 @@ function asyncHasMany(type, options, meta) {
       var rel;
       var promiseLabel = "DS: Async hasMany " + this + " : " + key;
       var resolver = Ember.RSVP.defer(promiseLabel);
+      var records = data[key];
+
+      // If a relationship is not specified by the adapter when it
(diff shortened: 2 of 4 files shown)


igorT:
Superseded by #2435

PR #2245, opened 2014-09-07
Consistentcy in iteration blocks and remove unnecessary code

chadhietala:
Just making some things consistent across files. Also removing some unnecessary code.


igorT:
This is super old, closing, i think we still prefer normal for loops if they work

PR #2321, opened 2014-09-27
Adds method to model that will return all its relationships on a RSVP.hash

eniolopes:
I'm still having trouble with one of the tests, it is failing in a non-deterministic way. Any ideas?


igorT:
Ping @eniolopes or reminder @igorT

bmac:
Hi @eniolopes. I'm going to close this issue as there hasn't been much activity in a while. Feel free to re-open if you would still like to work on this feature.

PR #2325, opened 2014-09-29
Adds support for pushing promise objects to hasMany relationships

jcollins1991:
Added the hasMany support further out as you suggested. Has a few tests for the various code-paths, not sure of the best way to test that all of them work as expected... Hard when theres like 10 ways to add a record to a `ManyArray` :(

@igorT

### DIFF
--- a/packages/ember-data/lib/system/record_arrays/many_array.js
+++ b/packages/ember-data/lib/system/record_arrays/many_array.js
@@ -108,7 +108,17 @@ export default RecordArray.extend({
       this.get('relationship').removeRecords(records);
     }
     if (objects){
-      this.get('relationship').addRecords(objects, idx);
+      var underlyingRecords = objects.map(function(record) {
+        if (record && record.then) {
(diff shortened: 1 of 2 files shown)


packages/ember-data/lib/system/record_arrays/many_array.js
```diff
@@ -108,7 +108,17 @@ export default RecordArray.extend({
       this.get('relationship').removeRecords(records);
     }
     if (objects){
-      this.get('relationship').addRecords(objects, idx);
+      var underlyingRecords = objects.map(function(record) {
+        if (record && record.then) {
+          var underlyingRecord = record.get('content');
```
bmac: This should probably be `var underlyingRecord = Ember.get(record, 'content'));`

fivetanley:
Closing due to inactivity and I think this can be implemented thanks to @pangratz on top of references: https://github.com/emberjs/data/pull/3303/files#diff-c9a4324bc502dfbcb034cc0cde828743R41

Issue #2212, opened 2014-08-21
Regression - Polymorphic relationship to unrelated classes no longer supported


mrloop:
Development builds of ember-data supported polymorphic relationship of classes which weren't subclasses of the same superclass. As of 1.0.0.beta.9 this is no longer the case.

[Working JSBin](http://emberjs.jsbin.com/cuxiva/3/edit) Using development build 1.0.0.beta.8
[Not working JSBin](http://emberjs.jsbin.com/visodu/1/edit) The same but using development build 1.0.0.beta.9

jrhe:
If you do polymorphic relationships without inheriting from a base class of the polymorphic type, it breaks creation of new record as BelongsToRelationship.prototype.addRecord asserts that the class must be a subtype of the polymorphic relationships class.

https://github.com/emberjs/data/blob/48c02654e8a524b390d1a28975416ee73f912d9e/packages/ember-data/lib/system/relationships/state/belongs_to.js#L63


Probably mixing up my words and can't really figure out how to distil this down into a JS bin but hopefully it makes sense. I'll try and work something up.

igorT:
@mrloop @lukegalea does https://github.com/emberjs/data/pull/2775 address your issues?

mrloop:
@igorT I'll be able next few of days, cheers.

mrloop:
Thanks @igorT this is working for me.

igorT:
Awesome. Merged #2775

PR #2345, opened 2014-10-05
allow polymorphic relationship types to not exist

fivetanley:
Previously, you had to have a model defined
for polymorphic relationships. However, the
polymorphic relationship belongsTo/hasMany
did not actually have to be an instanceof the
base type declared in DS.belongsTo(‘modelType’).

This commit removes that check for polymorphic
types only.

fixes GH-2212

### DIFF
--- a/packages/ember-data/lib/system/relationship-meta.js
+++ b/packages/ember-data/lib/system/relationship-meta.js
@@ -8,7 +8,7 @@ export function typeForRelationshipMeta(store, meta) {
     if (meta.kind === 'hasMany') {
       typeKey = singularize(typeKey);
     }
-    type = store.modelFor(typeKey);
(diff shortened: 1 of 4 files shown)


NullVoxPopuli:
So, how do I specify a relationship is being any type?

hjdivad:
> @hjdivad So, how do I specify a relationship is being any type?

The check is made using `instanceof` so the most straightforward way would be to have a base class for your app's models and use that as the type if your association could refer to literally any model.

@igorT was support ever added for polymorphic associations referring to mixins?  This might be closer to what @NullVoxPopuli is looking for.

aexmachina:
I would like to be able to have a polymorphic association that doesn't specify a superclass at all. I tried passing `DS.Model`, but it doesn't work. I needed to create a 'model' Model :(

PR #2358, opened 2014-10-07
Yield in pushMany: proof of concept

slindberg:
Pushing a large number of records into the store can be a very expensive operation. It gets even more expensive the more relations records have to each other. This can then result in the page being unresponsive for an extended period of time, most importantly blocking animations.

This is an experiment to explore solutions to extended blocking. Unfortunately it requires changing the behavior of `store#pushMany` to be asynchronous (something that has surprisingly little impact on ED internals). `pushMany` now looks for `DS.yieldFn` and calls it between each individual call to `store#push`:

```javascript
DS.yieldFn = function(next, index) {
  if (index % 3 === 0) {
    window.requestAnimationFrame(next);
  } else {
    next();
  }
};
```

This allows the browser to breathe every third call to `push`, which is enough for animations to appear smooth. The fixed batch size is a naive solution, but it serves to show what's possible.

### DIFF
--- a/packages/ember-data/lib/system/record_arrays/adapter_populated_record_array.js
+++ b/packages/ember-data/lib/system/record_arrays/adapter_populated_record_array.js
@@ -37,22 +37,25 @@ export default RecordArray.extend({
     @param {Array} data
   */
   load: function(data) {
+    var recordArray = this;
     var store = get(this, 'store');
     var type = get(this, 'type');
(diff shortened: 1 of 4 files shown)


igorT:
@slindberg performance has been an ongoing concern, I believe we should be a lot better now, but still have lots of room to improve. Let us know if this has been felt in your app, closing this PR as the solution is not one we are likely to go for

PR #1984, opened 2014-05-29
Allow errors on arbitrary properties, not just defined attributes or relationships.

alexspeller:
Addresses #1877

@tchak @igorT ping

Issue #1977, opened 2014-05-27
ActiveModel adapter 422 error handling


fsmanuel:
@igorT Is it possible to consider the `attrs` hash [here](https://github.com/emberjs/data/blob/master/packages/activemodel-adapter/lib/system/active_model_adapter.js#L113)?

Right now the errors from the server doesn't match custom mappings and are worthless.

Issue #1877, opened 2014-04-15
Showing server side validation errors for a base object


daveroberts:
Let's say I have an object Story with two properties: title and body.
I have validation rules for both the title and body, but I also have validation rules on the object as a whole, that don't apply to any one property directly.  For example, it's okay for a story to have a duplicate title, as long as the body is different.

In Rails, it's suggested that these base level validation errors are placed onto the "base" attribute (assuming your object doesn't have a base attribute already).  So the JavaScript errors that are returned look like this:

    {"errors":{"body":["can't be blank"],"title":["should begin with a capital letter"],"base":["A story with this same body and title already exist."]}}

On my Ember model for Story, if I do not add base as a property, Ember will not place the base validation errors onto my model object, and the base errors are not displayed on the page when the object is invalid.

JSBin without Base
http://jsbin.com/motuvaye/10/edit?html,js,output

1. Do not enter any information
2. Click create
3. Object is invalid.  Notice that the base error messages are not displayed on the form.
4. Fill out a new title and new body, the object changes from an invalid state to an uncommited state

JSBin with Base property:
http://jsbin.com/motuvaye/9/edit?html,js,output

If base is added as a property to the Ember model, the following behavior occurs:

1. Do not enter any information
2. Click create
3. Object is invalid.  The base error messages are displayed
4. Fill out a new title and new body.  The object does NOT change from invalid and thus an exception will be thrown by Ember if you try to save again.

The desired behavior would be to have a way to display this base error message, and let the object be saved after the user corrects both the title and body.

@alexspeller has made a pull request to allow saving objects in an invalid state.  This may help solve part of the problem at hand. https://github.com/emberjs/data/pull/1876

bmac:
Fixed by #1984

PR #2392, opened 2014-10-16
Refactored InvalidError handling into a serializer concern.

ahacking:
Refactored `InvalidError` handling into a serializer concern.

Addresses #1977 by ensuring error keys undergo the same mapping treatment that model attribute and relationship keys do.

Supports #1984 and #1877 by moving the model error serialization concern into the serializer. This PR should remove any objections for not applying all errors which have been sanctioned by the serializer to the record/model errors.

Paves the way for embedded record error handling, with the intent that errors can be mapped/normalized for the specific embedded records to which they apply.

### DIFF
--- a/packages/ember-data/lib/serializers/json_serializer.js
+++ b/packages/ember-data/lib/serializers/json_serializer.js
@@ -261,6 +261,16 @@ export default Ember.Object.extend({
   },
 
   /**
+    @method normalizeErrors
+    @private
+  */
+  normalizeErrors: function(type, hash) {
+    this.normalizeId(hash);
+    this.normalizeAttributes(type, hash);
+    this.normalizeRelationships(type, hash);
+  },
+
+  /**
     Looks up the property key that was set by the custom `attr` mapping
     passed to the serializer.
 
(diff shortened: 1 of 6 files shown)


joostdevries:
:heart_eyes: :heart_eyes: :heart_eyes:

packages/ember-data/lib/serializers/json_serializer.js
```diff
@@ -261,6 +261,16 @@ export default Ember.Object.extend({
   },
 
   /**
+    @method normalizeErrors
+    @private
+  */
+  normalizeErrors: function(type, hash) {
+    this.normalizeId(hash);
+    this.normalizeAttributes(type, hash);
+    this.normalizeRelationships(type, hash);
```
jbourassa: Hey @ahacking,

We're updating our app to latest Ember-Data and we've been having issue in our errors processing because of this. When using polymorphic relationships, `normalizeRelationships` will inspect the hash for a `type` property. Since this is an error payload, the `type` property won't be set.

Here's a [workaround implementation](https://gist.github.com/jbourassa/4fe88271d271524402a9) which works great for us. Do you see any issue with it? I'm thinking of opening a PR with it.

igorT: Can you open an issue for this please as well, just so we can track it

jbourassa: Done: https://github.com/emberjs/data/issues/3017

quaertym:
:fireworks: Thanks!

PR #2395, opened 2014-10-16
Defines raw fixtures getters/setters in FixtureAdapter

xcambar:
This PR is concerned about how one could easily extend the FixtureAdapter.
It is a refactor of FixtureAdapter with the final objective to add the following two new methods:

* `getRawFixturesForType(type)` which is a getter to the raw fixtures (as set in direct return from `type.FIXTURES`)
* `setRawFixturesForType(type, fixtures)` which is the corresponding setter: no check, nothing.

With the new methods, devs (_ie_, "I" :wink: ) will be able to extend more easily the FixtureAdapter.
The main issue with the current implementation of FixtureAdapter is the inconsistent usage of `type.FIXTURES` and `this.fixturesForType(type)`, which is (guts-guessing here) due to the fact that `type.FIXTURES` is a reference and `this.fixturesForType(type)` returns a new array.

My goal for a custom FixtureAdapter is to have a simpler/easier data-representation for fixtures that my non-tchnical colleagues can use to "play" with the apps and fixtures more naturally and not having to dig deep into the relations when adding fixtures (generating data on the fly is also discussed).

### DIFF
--- a/packages/ember-data/lib/adapters/fixture_adapter.js
+++ b/packages/ember-data/lib/adapters/fixture_adapter.js
@@ -10,6 +10,12 @@ var counter = 0;
 
 import Adapter from "ember-data/system/adapter";
 
+function generateRemoteCallback (data, type) {
+  return function () {
+    return this.wrapFixture(data, type);
+  };
+}
+
 /**
   `DS.FixtureAdapter` is an adapter that loads records from memory.
   It's primarily used for development and testing. You can also use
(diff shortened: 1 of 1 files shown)


bmac:
Hi @xcambar. The fixture adapter has been moved to its own repo here https://github.com/emberjs/ember-data-fixture-adapter. I'm going to close this issue here but feel free to reopen it on the new repo.

PR #2400, opened 2014-10-20
allowing to pass an has many relationship on create record. Fixes #1497

### LINKED ISSUE #1497: store.createRecord + record.save do not resolve hasMany-Relationships
I tried to use store.createRecord for persisting hasMany-relationships, but they are not resolved in store.createRecord (store.push works correctly). 

Here is a sample. Beware: There is no open REST api on jsbin, so requests are just "simulated". Open Firebug network protocol and look at "request payload":
http://jsbin.com/OJOJoFE/1/edit?html,js,output

Example code snippet:
```
App.Role = DS.Model.extend({
	name: DS.attr('string')
});
App.User = DS.Model.extend({
	username: DS.attr('string'),
	password: DS.attr('string'),
	roles: DS.hasMany('role')
});

    var model = {
      username: 'username',
      password: 'password',
      roles: [1,2]
    };

// create a record based on model data
record = this.get('store').createRecord('user', model);

// persist created record to REST 
// should send: username='username', password='password' and roles=[1,2]
// actually sends: username='username', password='password' and roles=[]
record.save();
```

sly7-7:
Fixes #1497
cc/ @igorT

### DIFF
--- a/packages/ember-data/lib/system/relationships/has-many.js
+++ b/packages/ember-data/lib/system/relationships/has-many.js
@@ -2,6 +2,7 @@
   @module ember-data
 */
 
+import computedPolyfill from "ember-data/utils/computed-polyfill";
 import Model from "ember-data/system/model";
 import normalizeModelName from "ember-data/system/normalize-model-name";
 
@@ -120,10 +121,18 @@ function hasMany(type, options) {
     key: null
   };
 
-  return Ember.computed(function(key) {
-    var relationship = this._relationships[key];
-    return relationship.getRecords();
-  }).meta(meta).readOnly();
+  return computedPolyfill({
+    get: function(key) {
+      var relationship = this._relationships[key];
+      return relationship.getRecords();
+    },
+    set: function(key, records) {
+      var relationship = this._relationships[key];
+      relationship.clear();
+      relationship.addRecords(records);
+      return relationship.getRecords();
+    }
+  }).meta(meta);
 }
 
 Model.reopen({
--- a/packages/ember-data/tests/integration/adapter/rest-adapter-test.js
+++ b/packages/ember-data/tests/integration/adapter/rest-adapter-test.js
@@ -369,8 +369,7 @@ test("create - a serializer's attribute mapping takes precedence over keyForRela
 
   run(function() {
     var comment = store.createRecord('comment', { id: "a-comment-id", name: "First!" });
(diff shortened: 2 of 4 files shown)


packages/ember-data/lib/system/relationships/has_many.js
```diff
@@ -104,10 +104,14 @@ function hasMany(type, options) {
     key: null
   };
 
-  return Ember.computed(function(key) {
+  return Ember.computed(function(key, records) {
     var relationship = this._relationships[key];
+    if (arguments.length === 2) {
+      relationship.removeRecords(relationship.getRecords());
```
wecc: When using `async: true` I think this line would kick off a server request, no? We probably don't want that when we do:

```
store.createRecord('parent', { children: [child1, child2] });
```

sly7-7:
@igorT One more ping for you ;)

sicasli:
+1

packages/ember-data/lib/system/relationships/has_many.js
```diff
@@ -104,10 +104,14 @@ function hasMany(type, options) {
...
+      relationship.addRecords(records);
+    }
     return relationship.getRecords();
-  }).meta(meta).readOnly();
+  }).meta(meta);
```
igorT: `model.set('posts',  [post1, post2[)` should work though?

packages/ember-data/lib/system/relationships/has_many.js
igorT: nice catch you don't want to do a `getRecords()` here, there is a relationship.clear() you can use instead

igorT:
@sly7-7 sorry for the late response, this completely fell of my radar, until I saw @sicasli comment. Left a comment, really happy to see this merged soon. Thanks!

sly7-7:
@igorT rebased, updated using relationship.clear() and computedPolyfill. Though I don't know what to do with https://github.com/emberjs/data/pull/2400#discussion_r21727341

igorT:
Your rebase seems off by 1, it seems to have pulled another commit with it.

igorT:
Otherwise looks good, ready to merge

sly7-7:
@igorT Thank for the review, do you want me to merge that after the right rebase ?

igorT:
as long as travis passes

sly7-7:
Ok, I don't know what was going on, but I had a commit about the beta.18 beeing released. This should be good now

igorT:
Thanks!

PR #2412, opened 2014-10-22
Add failing test on PromiseArray.createRecord when called before the hasMany is loaded

ZenCocoon:
2 bugs present in this PR:

1) Calling createRecord on a PromiseManyArray before the hasMany is loaded remove the newly created record once the promises resolves.
2) The test itself is not suppose to load comments before `post.get('comments')` gets called, but it appears to behave differently. Any idea?

/cc @igorT


bmac:
@ZenCocoon do you mind rebasing this pr?

bmac:
Committed as https://github.com/emberjs/data/commit/647e80a647d43ccc028c964f3f35fe203f202539. Thanks @ZenCocoon.

PR #2414, opened 2014-10-23
Don't refilter .all() and .find() if only properties changed

igorT:
Might need minor cleanup, maybe a unit test. Also would like to rename 'ready' 
@stefanpenner I think this solves sandstorm's problem and overall should lead to major perf


sandstrom:
Thanks Igor & Brendan!

PR #2428, opened 2014-10-29
[BUG?] embedded records mixin does not use the customized `keyForAttribute`

tarzan:
I would like to customize the `keyForAttribute` function in such a way that some keys are serialized with a different key(name). Defining this as an extension on the serializer works fine. However when this key is in a nested object it doesn't.

I wrote an integration spec reproducing the bug. Should I try to solve it, or is this expected behaviour?

### DIFF
--- a/packages/ember-data/tests/integration/serializers/embedded_records_mixin_test.js
+++ b/packages/ember-data/tests/integration/serializers/embedded_records_mixin_test.js
@@ -701,6 +701,45 @@ test("serialize with embedded object (belongsTo relationship)", function() {
   });
 });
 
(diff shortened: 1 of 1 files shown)


Soliah:
This happened to me on 1.0.0-beta.14.1 and it looks like it's because `DS.EmbeddedRecordsMixin`'s serialization methods don't use [`_getMappedKey`](https://github.com/emberjs/data/blob/c0fe627e09a57765df01601d8728ed2537f5035c/packages/ember-data/lib/serializers/json_serializer.js#L282), whereas `JSONSerializer` (which REST and ActiveModel serializers extend from) does.

PR #2435, opened 2014-10-29
update buildURLForHasMany hook

arenoir:
Implement #2168 buildURLForHasMany hook on current build. It uses an explicit buildURL option rather than depending on links and ids being nil.  I'm with @tomdale that a dsl is needed but I have needed/wanted this feature from day one. 

@igorT  let me know what you think #2162.

### DIFF
--- a/packages/ember-data/lib/system/store.js
+++ b/packages/ember-data/lib/system/store.js
@@ -2130,24 +2130,9 @@ function setupRelationships(store, record, data) {
   var typeClass = record.type;
 
   typeClass.eachRelationship(function(key, descriptor) {
-    var kind = descriptor.kind;
-    var value = data[key];
-    var relationship;
+    var relationship = record._relationships.get(key);
 
-    if (data.links && data.links[key]) {
-      relationship = record._relationships.get(key);
-      relationship.updateLink(data.links[key]);
-    }
-
-    if (value !== undefined) {
-      if (kind === 'belongsTo') {
-        relationship = record._relationships.get(key);
-        relationship.setCanonicalRecord(value);
-      } else if (kind === 'hasMany') {
-        relationship = record._relationships.get(key);
-        relationship.updateRecordsFromAdapter(value);
-      }
-    }
+    relationship.updateFromAdapter(data);
   });
 }
 
--- a/packages/ember-data/lib/adapters/build-url-mixin.js
+++ b/packages/ember-data/lib/adapters/build-url-mixin.js
@@ -238,6 +238,18 @@ export default Ember.Mixin.create({
     return url.join('/');
   },
 
+  /**
+    Builds a URL for hasMany associations given relationshipName, recordName and record.
+
+    Optionally used to generate a URL if payload does not include links or ids.
+
+    @method buildURLForHasMany
+    @param {String} relationshipName
+    @param {String} recordName
+    @param {DS.Model} record
+    @return {String} url
+  */
+  buildURLForHasMany: null,
 
   /**
     Determines the pathname for a given type.
--- a/packages/ember-data/lib/adapters/rest-adapter.js
+++ b/packages/ember-data/lib/adapters/rest-adapter.js
@@ -603,7 +603,6 @@ var RestAdapter = Adapter.extend(BuildURLMixin, {
 
   _stripIDFromURL: function(store, snapshot) {
     var url = this.buildURL(snapshot.modelName, snapshot.id, snapshot);
-
     var expandedURL = url.split('/');
     //Case when the url is of the format ...something/:id
     var lastSegment = expandedURL[expandedURL.length - 1];
--- a/packages/ember-data/lib/system/relationships/state/belongs-to.js
+++ b/packages/ember-data/lib/system/relationships/state/belongs-to.js
@@ -16,6 +16,22 @@ BelongsToRelationship.prototype = Ember.create(Relationship.prototype);
 BelongsToRelationship.prototype.constructor = BelongsToRelationship;
 BelongsToRelationship.prototype._super$constructor = Relationship;
 
+BelongsToRelationship.prototype.updateFromAdapter = function(data) {
+  var key   = this.key;
+  var value = data[key];
+
+  if (data.links && data.links[key]) {
+    this.updateLink(data.links[key]);
+  }
+
+  if (value === undefined) {
+    return;
+  }
+
+  this.setCanonicalRecord(value);
+  this.setHasData(true);
+};
+
 BelongsToRelationship.prototype.setRecord = function(newRecord) {
   if (newRecord) {
     this.addRecord(newRecord);
--- a/packages/ember-data/lib/system/relationships/state/has-many.js
+++ b/packages/ember-data/lib/system/relationships/state/has-many.js
@@ -24,6 +24,21 @@ ManyRelationship.prototype = Ember.create(Relationship.prototype);
 ManyRelationship.prototype.constructor = ManyRelationship;
 ManyRelationship.prototype._super$constructor = Relationship;
 
+ManyRelationship.prototype.updateFromAdapter = function(data) {
+  var key   = this.key;
+  var value = data[key];
+  var link  = data.links && data.links[key] || this.generateLink();
+
+  if (link) {
+    this.updateLink(link);
+  }
+
+  if (value !== undefined) {
+    this.updateRecordsFromAdapter(value);
+    this.setHasData(true);
+  }
+};
+
 ManyRelationship.prototype.destroy = function() {
   this.manyArray.destroy();
 };
@@ -190,6 +205,20 @@ ManyRelationship.prototype.getRecords = function() {
       this.manyArray.set('isLoaded', true);
     }
     return this.manyArray;
+  }
+};
+
+ManyRelationship.prototype.generateLink = function() {
+
+  if (!this.isAsync) {
+    return;
+  }
+
+  var adapter = this.store.adapterFor(this.type);
+  var buildUrl = adapter.buildURLForHasMany;
+
+  if (typeof buildUrl === "function") {
+    return buildUrl.call(adapter, this.key, this.inverseKey, this.record);
   }
 };
 
--- a/packages/ember-data/lib/system/relationships/state/relationship.js
+++ b/packages/ember-data/lib/system/relationships/state/relationship.js
@@ -7,6 +7,7 @@ function Relationship(store, record, inverseKey, relationshipMeta) {
   this.canonicalMembers = new OrderedSet();
   this.store = store;
   this.key = relationshipMeta.key;
+  this.type = relationshipMeta.type;
   this.inverseKey = inverseKey;
   this.record = record;
   this.isAsync = relationshipMeta.options.async;
@@ -20,7 +21,7 @@ function Relationship(store, record, inverseKey, relationshipMeta) {
 
 Relationship.prototype = {
   constructor: Relationship,
-
+  updateFromAdapter: Ember.K,
   destroy: Ember.K,
 
   clear: function() {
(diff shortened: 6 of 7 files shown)


arenoir:
@igorT any word on this? Is there anything I can do to help this along?

igorT:
@arenoir holidays and crap came in between, reviewing now. Feel bad about leaving this hanging, happy to pair with you/screenshare on this or any other ED/Ember issue as a penance.

packages/ember-data/lib/adapters/rest_adapter.js
```diff
@@ -549,6 +549,18 @@ export default Adapter.extend({
     return url;
   },
 
+
+  
+  buildURLForHasMany: function(relationshipName, type, record) {
```
igorT: I am not sure we want provide a default implementation, seems more like a fallback option?

packages/ember-data/lib/system/store.js
```diff
@@ -2005,22 +2013,8 @@ function setupRelationships(store, record, data) {
...
-
-    if (kind === 'belongsTo') {
-      if (value === undefined) {
-        return;
-      }
-      relationship.setCanonicalRecord(value);
-    } else if (kind === 'hasMany' && value) {
-     relationship.updateRecordsFromAdapter(value);
-    }
+    relationship.setup(data);
```
igorT: :+1:

igorT: Though i wouldn't call it setup, as it happens whenever there is a server update, maybe updateFromAdapter?

packages/ember-data/tests/integration/relationships/has_many_test.js
```diff
@@ -144,6 +144,52 @@ test("A serializer can materialize a hasMany as an opaque token that can be lazi
   });
 });
 
+test("If has-many options buildUrl is set, the adapter's buildURLForHasMany is called and the result is passed to findHasMany", function() {
+  Post.reopen({
+    comments: DS.hasMany('comment', { async: true, buildURL: true })
```
igorT: The model layer shouldn't have any knowledge of whether we want to buildURLs or not

packages/ember-data/lib/system/relationships/state/has_many.js
```diff
@@ -16,6 +16,25 @@ ManyRelationship.prototype = Ember.create(Relationship.prototype);
...
+ManyRelationship.prototype.setup = function(data) {
+  var key   = this.key,
+      value = data[key];
+
+  if (data.links && data.links[key]) {
+    this.updateLink(data.links[key]);
+  }
+
+  if (this.buildURL) {
+    var link = this.store.buildURLForHasMany(this.key, this.inverseKey, this.record);
```
igorT: If the URL is dynamic, we should probably not calculate it upfront, because it might depend on data that changes in the meantime.

igorT:
I feel like instead of passing buildURL as an option, we should fallback to it if t is defined and the the server did not provide us with ids or with a url.

packages/ember-data/lib/system/relationships/state/has_many.js
```diff
@@ -16,6 +16,25 @@ ManyRelationship.prototype = Ember.create(Relationship.prototype);
...
 
+ManyRelationship.prototype.setup = function(data) {
+  var key   = this.key,
+      value = data[key];
+
+  if (data.links && data.links[key]) {
+    this.updateLink(data.links[key]);
+  }
+
+  if (this.buildURL) {
```
igorT: Basically instead of the option, if no value and no link, we should set a flag that is 'generateLink' if there is a buildURLForHasMany in the adapter

igorT: We can then check the flag at https://github.com/arenoir/data/blob/build-hasmany-url/packages/ember-data/lib/system/relationships/state/has_many.js#L188 instead of relying on there being a link

igorT: And `findLink` can then generate a `link` if one is not provided

igorT:
Again, apologies for taking this long, PRs like these are a bit scary and I suck :(

arenoir:
@igorT you don't suck :) this is a overwhelming project. I am happy to be able to contribute.  

I will make the changes you noted and rebase.

arenoir:
@igorT take a look when you get some time.  The biggest change was moving the `buildUrlForHasMany` function to the `relationshipFromMeta` function.  I think it would be beneficial to add the same functionality to belongsTo relationships `buildUrlForBelongsTo`.  This would be useful when dealing with composite primary keys.

packages/ember-data/lib/system/relationship-meta.js
```diff
@@ -16,13 +16,24 @@ export function typeForRelationshipMeta(store, meta) {
   return type;
 }
 
+export function generateLink(store, kind, type) {
```
igorT: Why not put it on the relationship class?

packages/ember-data/lib/system/relationships/state/belongs_to.js
```diff
@@ -16,6 +16,21 @@ BelongsToRelationship.prototype = Ember.create(Relationship.prototype);
 BelongsToRelationship.prototype.constructor = BelongsToRelationship;
 BelongsToRelationship.prototype._super$constructor = Relationship;
 
+BelongsToRelationship.prototype.updateFromAdapter = function(data) {
+  var key   = this.key,
+      value = data[this.key];
```
igorT: Can you put another var here please? It makes debugging easier to have multiline var assignments

packages/ember-data/lib/system/relationships/state/has_many.js
```diff
@@ -10,12 +10,31 @@ var ManyRelationship = function(store, record, inverseKey, relationshipMeta) {
...
 ManyRelationship.prototype = Ember.create(Relationship.prototype);
 ManyRelationship.prototype.constructor = ManyRelationship;
 ManyRelationship.prototype._super$constructor = Relationship;
 
+ManyRelationship.prototype.updateFromAdapter = function(data) {
+  var key   = this.key,
+      value = data[key],
+      link  = data.links && data.links[key];
+
+  if (typeof this.generateLink === "function") {
```
igorT: Would this break for people who provide their own link. You just seem to override it here. Shouldn't the link be generated only if the user did not already provide one?

packages/ember-data/tests/integration/relationships/has_many_test.js
```diff
@@ -144,6 +144,52 @@ test("A serializer can materialize a hasMany as an opaque token that can be lazi
...
+    equal(id, "1", "find id was 1");
+
+    return Ember.RSVP.resolve({ id: 1 });
+  };
+
+  env.adapter.findMany = function() {
+    throw new Error("Adapter's findMany should not be called");
+  };
+
+  env.adapter.buildURLForHasMany = function(relationshipName, recordName, record) {
```
igorT: Can you add docs for this method in the basic Adapter/RESTAdapter?

packages/ember-data/tests/integration/relationships/has_many_test.js
```diff
@@ -144,6 +144,52 @@ test("A serializer can materialize a hasMany as an opaque token that can be lazi
   });
 });
 
+test("If has-many options buildUrl is set, the adapter's buildURLForHasMany is called and the result is passed to findHasMany", function() {
```
igorT: Can you add a test that assert that if buildURLForHasMany is provided, and user provided a URL, we always go to the user provided url and not generate a url

igorT:
`buildUrlForBelongsTo` seems like a right thing to do as well

packages/ember-data/lib/system/relationship-meta.js
arenoir: I may be misunderstanding the use but I was thinking If i put here the function would be looked up only once when the relationship is defined rater than each time the record is updated.

igorT: Not sure this makes a difference? You will call the function every time you want to generate a link no?

packages/ember-data/lib/system/relationships/state/has_many.js
arenoir: Yes I thought about that.  Should the adapter take precedence or the payload?

igorT: Well if we have a link from the payload no reason to generate our own. Also we only want to do this if we do not have `ids` provided.

packages/ember-data/lib/system/relationship-meta.js
arenoir: yes if the `generateLink` function is found it will call the returned function every time but if the function is not found it is only called once.

arenoir:
okay cool I will add the `buildUrlForBelongsTo` and take care of the other notes.

arenoir:
@igorT okay I took care of the issues you mentioned.  My tests pass, It looks like Travis ci is crapping out when running npm.

arenoir:
@igorT take a look when you get some time.  I think this is ready to go.

Myztiq:
+1 this is going to hopefully solve one of the biggest hurdles we have when using ember-data IMHO.

arenoir:
@bmac, @igorT, @fivetanley just brought this up to date with master.  Has this pr been discussed?  What is preventing it from being merged?  Is there anything I can do?

wecc:
Is my understanding correct that this PR and #2898 are solving the same problem, but in different ways?

arenoir:
Implementing [RFC #4](https://github.com/emberjs/rfcs/pull/4) looks like it would solve this problem.  Will it land in 1.0?  Until then I will keep this up to date.

arenoir:
@wecc I believe PR #2898 and RFC #4 make the buildUrl more customizable but still require  relationship payload from the server.  

This pr adds extra functionality to [generate](https://github.com/arenoir/data/blob/build-hasmany-url/packages/ember-data/lib/system/relationships/state/has-many.js#L28) the hasMany link automatically using the type and foreign_key.

bmac:
This pr looks good to me. :+1: It just needs a rebase and a review from @igorT.

arenoir:
I will rebase it today.

arenoir:
@bmac, I have rebased.  However there are two tests that fail because of the way I restructured the `setupRelationships` function in packages/ember-data/lib/system/store.js [line 2135](https://github.com/emberjs/data/pull/2435/files#diff-1636bea36bbe333a834ea11fbfbde48dR2135). 

The failing tests test the absence of new record relationships.  And I am guessing `record._relationships.get(key)` lazily creates a relationship.  

I like moving the logic of how to load the data into the relationship instance. 

So I guess the question is why shouldn't new records have empty relationships?

@igorT any thoughts?  Is this functionality that @tomdale originally [implemented](https://github.com/emberjs/data/pull/2168) still sought after?

arenoir:
Here is a work around.

https://gist.github.com/jamesarosen/79337a515f557d7e9cf6

greyhwndz:
@igorT: Any update on this?

bmac:
@arenoir any chance you have time to rebase this pr?

bmac:
@arenoir do you have time to rebase this pr and put it behind a feature flag?

arenoir:
@bmac I don't know if this is necessary anymore?  I am able build the link for relationships within the models serializer normalize function.  Let me know.  I could possibly spend some time on this next week.

```
#/serializers/job.js
function urlForAttachments(jobId) {
  return `/api/v2/private_attachments?job_id=${jobId}`;
}

export default ApplicationSerializer.extend({
  normalize(typeClass, hash, prop) {
    hash.links = {
      privateAttachments: urlForAttachments(hash.id)
    };
    return this._super(typeClass, hash, prop);
  }
});
```

bmac:
Thanks @arenoir. I'm willing to re-visit this in the future but for now I like the idea of building the link for relationships within the models serializer normalize function.

stefanpenner:
This seems to have gone stale. If this is something we want, @bmac suggested it go behind a feature flag, if not we can leave this closed.

PR #2467, opened 2014-11-12
Dematerialize rejected _find() if record isEmpty

### LINKED ISSUE #1747: Failed store.find with ID creates record cache (blocks ids later)
The following testcase:

On an yet empty adapter i do something like this:

1. store.find('somemodel', 4)
 a) this will try to lookup the record, but wont find it (ok)
 b) will trigger the "fail" promise in regards of "not found" (ok so far)

2. Now create a record with the ID 4
data = {'title':'TEST','id':4}
record = store.createRecord('test', data)

This will run into an exception which says, that the ID 4 is already taken. It is taken due to the "unsuccessful" find beforehand. The find method creates a record for an ID even if the ID has not been found.

Resuolution:
Do not register/create/cache ids, when the lookup in the adapter fails -> record does not yet exist. 
Eventhough the current case is rather an edgecase  (create records with a specific ID) it can happen and there is no benefit ( i can see ) of the current behavior.

### LINKED ISSUE #1767: requests for non existing dynamic segments create empty objects
If someone modifies the dynamic segment (object ID) in the URL of an Ember Application with Ember Data a new object is created. 

[Here is a JSbin](http://emberjs.jsbin.com/hurozaju/20#/colors) demonstrating this behaviour. If you click on one of the links and modify the dynamic segment (eg. type in an additional digit), you will be redirected to the "#/colors-page" with an additional color listed. The color-object is empty except of the ID (which equals the requested object).

This happens in every ember/ember-data version i tried...  (ranging from Ember-1.3.0/Ember-Data-1.0.0-beta.4 to  current canary builds).

Also i posted a [question on Stackoverflow](http://stackoverflow.com/questions/22019423/how-to-handle-requests-for-non-existing-dynamic-segments-in-ember) concerning this topic.

### LINKED ISSUE #1931: Request a nonexistent data by id still push to the store using RestfulAdapter
While request a nonexistent data by id, the server return 404.

for example, GET `localhost:5002/#/products/some_id_not_exists` return 404

but the store still got one product with the id `some_id_not_exists`, the other fields are undefined.
screenshot 2014-05-12 10 47 26
screenshot 2014-05-12 10 47 36

### LINKED ISSUE #2150: Unable to create record once find has been called
I am using FixtureAdapter for testing. I am trying to achieve the following:
- find in the store if a model with certain id 
- if such model exist, it is used for processing
- otherwise, a new model with that id is created for processing

        this.store.find('MyModel', { id: 'foo' }).then(function(model) {
            //do something with model
        }, function(reason) {
            var model = this.store.createRecord('MyModel', { id: 'foo' });
            //do something with model
        });

However, what I get is an error:
`Error: Assertion Failed: The id foo has already been used with another record of type MyModel.`

After tracing the Ember Data code, I found that:
* store.idToRecord is set after calling store.find:
    - [store.find with id=`foo`](https://github.com/emberjs/data/blob/master/packages/ember-data/lib/system/store.js#L409)
    - [store.findById](https://github.com/emberjs/data/blob/master/packages/ember-data/lib/system/store.js#L422)
    - [store.recordForId](https://github.com/emberjs/data/blob/master/packages/ember-data/lib/system/store.js#L438)
    - [store.idToRecord['foo'] being set](https://github.com/emberjs/data/blob/master/packages/ember-data/lib/system/store.js#L678)
* store.createRecord fails because store.idToRecord contains empty record of the id `foo`:
    - [store.createRecord with id=`foo`](https://github.com/emberjs/data/blob/master/packages/ember-data/lib/system/store.js#L240)
    - [store.buildRecord](https://github.com/emberjs/data/blob/master/packages/ember-data/lib/system/store.js#L240)
    - [assertion fails in store.buildRecord due to empty `foo` record in store.idToRecord created by store.find previously](https://github.com/emberjs/data/blob/master/packages/ember-data/lib/system/store.js#L1473)

There is a similar issue in [emberFire](https://github.com/firebase/emberFire/issues/42). My temporary workaround is to call `store.dematerializeRecord` before `store.createRecord` and it is obviously undesirable because `store.dematerializeRecord` is non-public.

wecc:
This dematerializes records that has been created when doing a `store.find('thing', 1)` and it fails/gets rejected. Not 100% sure if `isEmpty` is the correct condition here though.

Also, is this the behaviour we want?

Fixes #1747, fixes #1767, fixes #1931, fixes #2150

### DIFF
--- a/packages/ember-data/lib/system/store.js
+++ b/packages/ember-data/lib/system/store.js
@@ -1911,6 +1911,9 @@ function _find(adapter, store, type, id, record) {
     var record = store.getById(type, id);
     if (record) {
       record.notFound();
+      if (get(record, 'isEmpty')) {
+        store.dematerializeRecord(record);
+      }
     }
     throw error;
   }, "DS: Extract payload of '" + type + "'");
--- a/packages/ember-data/tests/integration/adapter/find_test.js
+++ b/packages/ember-data/tests/integration/adapter/find_test.js
@@ -120,3 +120,22 @@ test("When a single record is requested, and the promise is rejected, .find() is
     }));
   });
 });
+
+test("When a single record is requested, and the promise is rejected, the record should be dematerialized.", function() {
+  expect(2);
+
+  store = createStore({ adapter: DS.Adapter.extend({
+      find: function(store, type, id) {
+        return Ember.RSVP.reject();
+      }
+    })
+  });
+
+  run(function() {
+    store.find(Person, 1).then(null, async(function(reason) {
+      ok(true, "The rejection handler was called");
+    }));
+  });
+
+  ok(!store.hasRecordForId(Person, 1), "The record has been dematerialized");
+});


sly7-7:
@igorT ping ?

andrusieczko:
Hey, I feel that even if we dematerialize the record, there are some leftovers in cache...

When I did (just to check, could be moved to `dematerializeRecord`?):

    if (get(record, 'isEmpty')) {
      store.dematerializeRecord(record);
      store.typeMapFor(type).findAllCache.removeRecord(record); // remove cache
    }

it's better, but sometimes still doesn't work (I have to refresh my components to reflect the change).

With `dematerializeRecord` only, I get different results:
- `store.getById(id)` gives `null`
- `store.all(type).content.findBy('id', id) is not `null`

wecc:
@andrusieczko Nice catch! You are right, existing (cached) record arrays does not currently remove these records. I think this is the responsibility of `RecordArrayManager`, will investigate...

igorT:
@andrusieczko I think the reason it didn't work was because the proper method to call would be unloadRecord not dematerialize record. Can you test that one?

wecc:
#2781 addresses using `unloadRecord()` instead of `dematerializeRecord()`

andrusieczko:
@igorT, it's working perfectly now! Thanks! @wecc: thanks for the other PR

PR #2468, opened 2014-11-12
Support trailing slashes on urls for coalesceFindMany.

g-cassie:
Currently if you override your adapter as follows, coalesceFindRequests will break (silently).  
```javascript
export default DS.ActiveModelAdapter.extend({
    buildURL: function(type, id, record){
      return this._super(type, id, record) + '/';
    },
});
```
As _stripIDFromURL is marked as a private method I do not think it should be expected that users override it.  This provides a simple fix.  Happy to add a test if someone can suggest a simple approach for testing this.

I've noticed the build is failing. I'll wait to hear whether this would be a merge candidate before fixing. I will also need to make some adjustments to make sure that coalesce uses a trailing slash in the url if a trailing slash url is already being used.

### DIFF
--- a/packages/ember-data/lib/adapters/rest_adapter.js
+++ b/packages/ember-data/lib/adapters/rest_adapter.js
@@ -589,6 +589,10 @@ export default Adapter.extend({
     var url = this.buildURL(type.typeKey, record.get('id'), record);
 
     var expandedURL = url.split('/');
+    // Handle urls with trailing slash
+    if (expandedURL[expandedURL.length - 1] === '') {
+      delete expandedURL[expandedURL.length - 1];
+    }
     //Case when the url is of the format ...something/:id
     var lastSegment = expandedURL[ expandedURL.length - 1 ];
     var id = record.get('id');


dustinfarris:
What is keeping this from being merged?

bmac:
Looks like it needs a test case.

dustinfarris:
@g-cassie are you still working on tests for this?

g-cassie:
Sorry, I won't have any time to look at this again for at least a month.  I don't have much experience with Ember testing either.

dustinfarris:
@g-cassie ok, if you don't mind, I'm going to finish it then.  Need this for ember-django-adapter.

g-cassie:
@dustinfarris please do. Sorry I can't be of more help right now.

dustinfarris:
So I ended up overriding this in our custom adapter, which I believe is the right way forward.  This can be closed without merging in my opinion.

However, in doing so, I decided to greatly simplify the method by proxying to buildUrl.  These seems to have the same effect with much less effort.  Am I missing something?

https://github.com/dustinfarris/ember-django-adapter/blob/master/addon/adapters/drf.js#L132-L134

igorT:
@dustinfarris if buildUrl works, i think it's only on accident

benkonrath:
@igorT Can you be more specific about the problem with @dustinfarris's solution? Do you think it would be better to add tests for code in this PR and get it merged into ember-data?

PR #2501, opened 2014-11-21
[Bug] add assertion to prevent from rejecting records without a reason.

stefanpenner:
When rejecting because records were missing, we currently reject without a reason. This results in non-actionable failures.

I am actually pretty sure the find coalescing doesn't work even with these fixes

I'm going to implement out-side of ember data right now, I'll try to circle back this weekend and figure out whats wrong inside ember-data.

- [ ] the usage of without is still broken
- [ ] switch recordResolverPairs to be a map

### DIFF
--- a/packages/ember-data/lib/system/store.js
+++ b/packages/ember-data/lib/system/store.js
@@ -595,7 +595,7 @@ Store = Ember.Object.extend({
           _findMany(adapter, store, type, ids, requestedRecords).
             then(resolveFoundRecords).
             then(makeMissingRecordsRejector(requestedRecords)).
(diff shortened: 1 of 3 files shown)


stefanpenner:
I keep running into this, I can't believe I am the only one.....

I have no excuse but to fix this soon.

bmac:
@stefanpenner any plans to pick this up at some point? If not would you be able to write up a description of the steps that need to be done so someone else can take over this pr?

stefanpenner:
I suspect the work is to far diverged now, closing.

Issue #1893, opened 2014-04-25
Ember Data BelongTo Relationship :Duplicate record after save


maheshiv:
I am using ember data for my application . I defined two models with belongsTo relationship

```
App.Queue=DS.Model.extend({
time_added:DS.attr("string"),
description:DS.attr("string"),
video:DS.belongsTo("video")
});

App.Video=DS.Model.extend({
video_url:DS.attr("string"),
title:DS.attr('string'),
duration:DS.attr('number'),
queue:DS.belongsTo('queue')
});
```

For adding a video to queue , I do 

```
var queue=this.store.createRecord("queue",{time_added:"2014-04-26",description:"Good to have at this time"});
this.store.find('video',44).then(function(data){
queue.set('video',data);
queue.save();
});
```

response after save will be

```
{queue:{id:1,time_added:"2014-04-26",description:"Good to have at this time",video:44},
video:[{
id:44,
video_url:"beauty.png",
title:"Cool dance",
duration:"30",
queue:1
}]
}
```

Now it will create new record in the store . But the problem is after save when i check the queue computed property it shows "App.Queue:ember1724:null" [in belongsTo relationship] and when i click on this property it will fetch new record from the server with the same id and remaining values are undefined . I am having doubt that why it should fetch new queue record with same id if it is already there ?

I think you understand my problem . Thanks in advance .

PR #2537, opened 2014-11-30
[WIP] Fix for emmbeded records mixin to make it work with rails.

simcha:
Hi, I made some changes to EmbeddedRecordMixin to make it work with rails 
- New creation duplicate fix: Client id merged from https://github.com/tstirrat/ember-data/tree/embedded-client-id  #1893 (probobly)
- Add '_attributes' #2138
- Support for ids deserialization (default in ActiveModel::Serializers) 
- Support for deletion of objects by '_destroy'
- Properly marks dirty state of saved objects

It shall not brake non rails JSON and i have fixed the tests as well as added some new ones.

It is work in progress I do not like the client_id idea end especially the way it is stored in Ember. Same goes for the list of deleted records. But I think some subset of the solution or all of it after fixing could be useful for the others. I hope you can guide me in proper direction. Thanks in advance.


bmac:
Hi @simcha. Thanks for taking the time to work on this pr and sorry for not giving you any feedback until now. Unfortunately, I don't think we can introduce these changes into Ember Data's existing `EmbeddedRecordMixin` without breaking backwards compatibility. 

Instead, I think this would code make a great ember addon. If you have any trouble porting this to an addon please let me know and I would be happy to assist.

Issue #2553, opened 2014-12-03
Ember Data ^1.0.0-beta.11 should not recommend use of es5-sham


trabus:
Given [this issue](https://github.com/emberjs/data/issues/2488), can we assume it is safe to remove es5-sham from the IE8 recommendation on the front page?

PR #2554, opened 2014-12-03
Add es5-sham warning for IE8 support

trabus:
See [#2553]

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,8 +34,7 @@ after [setup](#setup). You'll find ember-data.js in the `dist` directory.
 
 #### Internet Explorer 8
 
-If you need to support Internet Explorer, you will need to use es5-shim.js and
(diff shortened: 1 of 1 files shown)


trabus:
Sorry no, I haven't had time to investigate further. We have it working with the ES5 shim, so it wasn't as high of a priority. I may have a little time next week, and we'll be upgrading our app to 1.9 as well.

trabus:
I finally tracked the issue down, and have determined that it was a utility that we were using that called Array.isArray(). Sorry for any confusion this has caused, IE8 is near impossible to debug, and I wasn't able to see what specifically was causing the error.

Running Ember 1.9.1 and Ember Data 1.0.0-beta.14 without the ES5-shim and all is well on IE8.

bmac:
Looks good to me. @fivetanley can you review and merge?

fivetanley:
thanks!

Issue #2557, opened 2014-12-04
canary - ember-data.js.map Not Found


mikepmunroe:
Upon installing Ember Data Canary, I have started to get the following exception,

GET http://localhost:4200/assets/ember-data.js.map 404 (Not Found)

gnapse:
I'm still getting this error, so what should I do to fix it? It doesn't appear to have any consequences, other than the annoying error message in the browser's console. I'm using the most recent starter kit with Ember 1.9.1, and Ember Data 1.0.0-beta.15-canary (as reported in the console).

bmac:
@gnapse With the starter kit you will need to manually include this file in the same directory as ember data. http://builds.emberjs.com/canary/ember-data.js.map

fivetanley:
We should include this in the starter kit.

gnapse:
Thanks @bmac for the solution. Regarding the inclusion of this in the starter kit @fivetanley, I don't think it should be done, as long as the starter kit continues shipping without ember data, as it does now. It should be noted though when you download ember data, that you also need this file.

BTW, what's this file needed for? As I mentioned above, my project is working perfectly without it. It's just the annoyance of having this error pop out in the console. But I assume there's more to it. What is it?

vasilakisfil:
Hello! I am not sure how relevant it is but when I start my app I get this error:
```bash
version: 0.1.6
Livereload server on port 35729
Serving on http://0.0.0.0:4200/
Path or pattern "ember-data.js.map" did not match any files
Error: Path or pattern "ember-data.js.map" did not match any files
    at Object.multiGlob (/home/vasilakisfil/Development/rails_tutorial_api/rails_tutorial_ember/node_modules/ember-cli/node_modules/broccoli-kitchen-sink-helpers/index.js:202:13)
    at /home/vasilakisfil/Development/rails_tutorial_api/rails_tutorial_ember/node_modules/ember-cli/node_modules/broccoli-static-compiler/index.js:28:27
    at $$$internal$$tryCatch (/home/vasilakisfil/Development/rails_tutorial_api/rails_tutorial_ember/node_modules/ember-cli/node_modules/rsvp/dist/rsvp.js:490:16)
    at $$$internal$$invokeCallback (/home/vasilakisfil/Development/rails_tutorial_api/rails_tutorial_ember/node_modules/ember-cli/node_modules/rsvp/dist/rsvp.js:502:17)
[...]

PR #2561, opened 2014-12-05
Include ember-data.js.map in S3 upload

wecc:
Added `ember.data.js.map` to be uploaded with `application/json` as content type, I figured it's JSON in the file and should probably be treated as such.

I also modified `fileObject()` to accept a `fileName` instead of `baseName` and `extension` since they weren't used by themselves anyways.

Somewhat related #2557

### DIFF
--- a/config/s3ProjectConfig.js
+++ b/config/s3ProjectConfig.js
@@ -1,30 +1,30 @@
 module.exports = function(revision, tag, date){
   return {
-    'ember-data.js': fileObject('ember-data', '.js', 'text/javascript', revision, tag, date),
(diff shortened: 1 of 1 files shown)


chavafarfan:
So what I did to have this source map served by Rails 4.2.0 was to follow this link 
https://raw.githubusercontent.com/components/ember-data/master/ember-data.js.map
Then copied the contents and pasted it into a new file on Sublime Text.
Saved the file as ember-data.js.map on vendor/assets/javascripts and everything good now

fivetanley:
this should be fixed in ember-rails https://github.com/emberjs/ember-rails/pull/424

PR #2551, opened 2014-12-03
Belongs To Polymorphic Async Bug Failing Test

jherdman:
This is a failing test for @sly7-7 regarding #1535.

PR #2571, opened 2014-12-10
Fix async polymorphic belongsTo serialization

### LINKED ISSUE #2385: Async Polymorphic relations don't serialize correctly
If you have a belongsTo that is async and polymorphic, when it tries to serialize the model, it throws an error because it didn't unwrap the promise.

This can be seen here in the console assert. http://emberjs.jsbin.com/zisoba/8/edit

The quick solution locally was for me to unwrap it (https://gist.github.com/aaronbhansen/5a4bb73cf4c16b22b924), but that might be a good longterm fix.

### LINKED ISSUE #2508: `RESTSerializer.serializePolymorphicType` broken for `async` relationships
When a relationship is defined as `polymorphic` *and* `async`, `serializePolymorphicType` raises `key is undefined` because the related object is wrapped in a promise, which breaks `belongsTo.constructor.typeKey`.

sly7-7:
Closes #2551, Fixes #2385, Fixes #2508, Closes #1491

Very naive implementation, also I didn't change anything in the embedded mixin because I don't know how/if it's supposed to be handled.

When reading #1535, I'm assuming the serialization process stays synchronous.

I think this should be considered as a breaking change because even if the signature does not change, the passed record to `serializePolymorphicType` is the inverse record instead of the record itself.

cc/ @jherdman @igorT

### DIFF
--- a/packages/activemodel-adapter/lib/system/active_model_serializer.js
+++ b/packages/activemodel-adapter/lib/system/active_model_serializer.js
@@ -4,8 +4,7 @@ import RESTSerializer from "ember-data/serializers/rest_serializer";
   @module ember-data
 */
 
-var get = Ember.get,
-    forEach = Ember.EnumerableUtils.forEach,
+var forEach = Ember.EnumerableUtils.forEach,
     camelize =   Ember.String.camelize,
     capitalize = Ember.String.capitalize,
     decamelize = Ember.String.decamelize,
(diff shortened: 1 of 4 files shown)


igorT:
@sly7-7 snapshots got done early, thanks to awesome work by @wecc so this is much easier to handle now, he'll extract the tests into #2623

sly7-7:
@igorT That is great news, thanks @wecc :+1:

Issue #1576, opened 2013-12-13
Ability to use query string params with store.find()


adamlc:
I've come across an issue where I need to append query string params to a find request.

So basically I need to do something like this:

```javascript
this.store.find('contact', 1, {embed: 'company'});
```

Which should produce a GET request like:

/contacts/1?embed=company

Going through the code I couldn't see a way of doing this. Has anyone got any thoughts or ideas on how I might achieve this?

seocahill:
Thanks for the patches hope the pull request is accepted

simo163:
Awesome, It worked perfect!!!!

igorT:
You will now be able to use `adapterOptions` to pass options to your adapter, using `findRecord`. We will soon have support for `include/embed` like things, but passing query params should also be easy.
Tracking this under https://github.com/emberjs/data/pull/3281

jpoiri:
When are you targeting this feature to be release?

abuiles:
`queryRecord` is now available in the store.

jpoiri:
Is this the right way to use it?

this.store.queryRecord('team', 77, {include_members: true}); Doesn't seem to work for me

wecc:
@jpoiri try `this.store.queryRecord('team', 77, { query: { include_members: true } });`

jpoiri:
@wecc I'm getting this now using the RestAdapter
:
Error: Assertion Failed: You tried to make a query but your adapter does not implement `queryRecord`

wecc:
@jpoiri https://github.com/emberjs/data/issues/3514

jpoiri:
thanks, I guess I'll have to wait

tschoartschi:
I also need this behaviour but I'm not sure how it works, because the following code snipped does not work for me.
```javascript
this.store.queryRecord('team', 77, { query: { include_members: true } });
```
I use ember-data 1.13.7 with the RESTAdapter and ember 1.13.6

What I did to get it working as expected (or as I would expect it):
```javascript
// SOMEWHERE IN A CONTROLLER OR ELSEWHERE 
    this.get('store').queryRecord(this.get('modelname'), {
        __id__: id,
        contentLanguage: newLang
    }).then(function (response) {
        // HANDLE SUCCESS 
    }, function (error) {
        // HANDLE ERROR
    });
````
And then I need to adapt buildURL in the RESTAdapter:

```javascript
import Ember from 'ember';
import DS from 'ember-data';

export default DS.RESTAdapter.extend({
[...]

Issue #1326, opened 2013-09-18
[Improvement?] find should take both id and query parameters


abobwhite:
I have a situation where I need to find a model via `/models/{id}` but to also provide query parameters: `/models/{id}?key=value`. The reason is that this particular model is under version control on the server. It can get "approved" and become publicly visible but another version can still be edited by the user in a "draft" state. When the user is logged in, they should be able to access the model in that "draft" state. There are two obvious solutions:

1) Enhance `find({id})` to optionally take query params: `find({id}, {/* draft : true */})`
2) Extend the model to be something like a "DraftModel" which would only serve as a wrapper to hit a different server endpoint: `/draftModels/{id}`

I prefer option 1 but I can see how opinions would be split between the options. Option 1 can be used to provide different information on the same model depending on a specific use-case and the state of the data.

Thoughts?

thomasvanlankveld:
+1 for `find('model', id, paramsObj);`. Feels only natural with `find('model', id);` and `find('model', paramsObj);` available. Is this really an edge case?

abobwhite:
@thomasvanlankveld I agree with you. There should be some standard way to do this. I do not believe it to be an edge case. It's fairly common practice even if it's not the most common scenario. It seems this and other related issues have been dismissed (closed) by core team members because they probably had bigger fish to fry. But currently, all solutions suggested making this work are a hack. I think the ultimate issue at hand here is that if we try ```this.store.find('myModel', {id : myId, otherParam: true})``` which used id as a param, the serializer expects the result to be an array because ```findQuery``` is invoked. If ```findQuery``` were to not assume array, then this might be a bit simpler and could conceivably be handled on the app-level rather than at an ember level.

That being said, I'm in favor of simply permitting query params on the ```find``` for a single model.

matixmatix:
+1 for `find('model', id, paramsObj)`. It's really frustrating using `find('myModel', {id : myId, otherParam: true})`, because the result is passed as an array. I don't believe this is really such an edge case, and would gladly see it getting implemented in the near future.

jeffjarchow:
+1 for find('model', id, paramsObj). This is something I find myself needing on many of my endpoints of my API. It just doesn't feel right having to fetch a single record as an array and extract the index I need every time.

[...]

PR #2584, opened 2014-12-12
Add queryOne method

thaume:
The findQueryOne method let developers query their server-side API for a single record and also pass a query parameter to this call.

For instance, you could do the following:

```javascript
this.store.findQueryOne('post', 1, { include: 'comments' });
```

### DIFF
--- a/packages/ember-data/lib/serializers/json-serializer.js
+++ b/packages/ember-data/lib/serializers/json-serializer.js
@@ -1254,6 +1254,24 @@ export default Serializer.extend({
     return this.extractArray(store, typeClass, payload, id, requestType);
   },
   /**
+    TODO: remove this in a couple of days
+
+    `extractQueryRecord` is a hook into the extract method used when a
+    call is made to `DS.Store#queryRecord`. By default this method is an
+    alias for [extractSingle](#method_extractSingle).
+
+    @method extractQueryRecord
+    @param {DS.Store} store
+    @param {DS.Model} typeClass
+    @param {Object} payload
+    @param {(String|Number)} id
+    @param {String} requestType
+    @return {Object} object A hash of deserialized object
+  */
+  extractQueryRecord: function(store, typeClass, payload, id, requestType) {
+    return this.extractSingle(store, typeClass, payload, id, requestType);
+  },
+  /**
     `extractFindMany` is a hook into the extract method used when a
     call is made to `DS.Store#findMany`. By default this method is
     alias for [extractArray](#method_extractArray).
--- a/packages/ember-data/lib/system/adapter.js
+++ b/packages/ember-data/lib/system/adapter.js
@@ -190,6 +190,44 @@ var Adapter = Ember.Object.extend({
   findQuery: null,
 
   /**
+    The `queryRecord()` method is invoked when the store is asked for a single
+    record through a query object.
+
+    In response to `queryRecord()` being called, you should always fetch fresh
+    data. Once found, you can asynchronously call the store's `push()` method
+    to push the record into the store.
+
+    Here is an example `queryRecord` implementation:
+
+    Example
+
+    ```javascript
+    App.ApplicationAdapter = DS.Adapter.extend({
+      queryRecord: function(store, typeClass, query) {
+        var url = [type.typeKey, id].join('/');
+
+        return new Ember.RSVP.Promise(function(resolve, reject) {
+          jQuery.getJSON(url, query).then(function(data) {
+            Ember.run(null, resolve, data);
+          }, function(jqXHR) {
+            jqXHR.then = null; // tame jQuery's ill mannered promises
+            Ember.run(null, reject, jqXHR);
+          });
+        });
+      }
+    });
+    ```
+
+    @method queryRecord
+    @param {DS.Store} store
+    @param {subclass of DS.Model} type
+    @param {Object} query
+    @param {String} id
+    @return {Promise} promise
+  */
+  queryRecord: null,
+
+  /**
     If the globally unique IDs for your records should be generated on the client,
     implement the `generateIdForRecord()` method. This method will be invoked
     each time you create a new record, and the value returned from it will be
--- a/packages/ember-data/lib/system/store.js
+++ b/packages/ember-data/lib/system/store.js
@@ -40,7 +40,8 @@ import {
   _findHasMany,
   _findBelongsTo,
   _findAll,
-  _query
+  _query,
+  _queryRecord
 } from "ember-data/system/store/finders";
 
 import coerceId from "ember-data/system/coerce-id";
@@ -1036,6 +1037,36 @@ Store = Service.extend({
     Ember.assert("You tried to load a query but your adapter does not implement `findQuery`", typeof adapter.findQuery === 'function');
 
     return promiseArray(_query(adapter, this, typeClass, query, array));
+  },
+
+  /**
+    This method delegates a query to the adapter. This is the one place where
+    adapter-level semantics are exposed to the application.
+
+    Exposing queries this way seems preferable to creating an abstract query
+    language for all server-side queries, and then require all adapters to
+    implement them.
+
+    This method returns a promise, which is resolved with a `RecordObject`
+    once the server returns.
+
+    @method queryRecord
+    @param {String or subclass of DS.Model} type
+    @param {any} query an opaque query to be used by the adapter
+    @return {Promise} promise
+  */
+  queryRecord: function(modelName, query) {
+    Ember.assert("You need to pass a type to the store's queryRecord method", modelName);
+    Ember.assert("You need to pass a query hash to the store's queryRecord method", query);
+    Ember.assert('Passing classes to store methods has been removed. Please pass a dasherized string instead of '+ Ember.inspect(modelName), typeof modelName === 'string');
+
+    var typeClass = this.modelFor(modelName);
+    var adapter = this.adapterFor(modelName);
+
+    Ember.assert("You tried to make a query but you have no adapter (for " + typeClass + ")", adapter);
+    Ember.assert("You tried to make a query but your adapter does not implement `queryRecord`", typeof adapter.queryRecord === 'function');
+
+    return promiseObject(_queryRecord(adapter, this, typeClass, query));
   },
 
   /**
--- a/packages/ember-data/tests/integration/store/query-record-test.js
+++ b/packages/ember-data/tests/integration/store/query-record-test.js
@@ -1,0 +1,63 @@
+var Person, store, env;
+var run = Ember.run;
+
+module("integration/store/query-record - Query one record with a query hash", {
+  setup: function() {
+    Person = DS.Model.extend({
+      updatedAt: DS.attr('string'),
+      name: DS.attr('string'),
+      firstName: DS.attr('string'),
+      lastName: DS.attr('string')
+    });
+
+    env = setupStore({
+      person: Person
+    });
+    store = env.store;
+  },
+
+  teardown: function() {
+    run(store, 'destroy');
+  }
+});
+
+test("It raises an assertion when no type is passed", function() {
+  expectAssertion(function() {
+    store.queryRecord();
+  }, "You need to pass a type to the store's queryRecord method");
+});
+
+test("It raises an assertion when no query hash is passed", function() {
+  expectAssertion(function() {
+    store.queryRecord('person');
+  }, "You need to pass a query hash to the store's queryRecord method");
+});
+
+test("When a record is requested, the adapter's queryRecord method should be called.", function() {
+  expect(1);
+
+  env.registry.register('adapter:person', DS.Adapter.extend({
+    queryRecord: function(store, type, query) {
+      equal(type, Person, "the query method is called with the correct type");
+      return Ember.RSVP.resolve({ id: 1, name: "Peter Wagenet" });
+    }
+  }));
+
+  run(function() {
+    store.queryRecord('person', { related: 'posts' });
+  });
+});
+
+test("When a record is requested, and the promise is rejected, .queryRecord() is rejected.", function() {
+  env.registry.register('adapter:person', DS.Adapter.extend({
+    queryRecord: function(store, type, query) {
+      return Ember.RSVP.reject();
+    }
+  }));
+
+  run(function() {
+    store.queryRecord('person', {}).then(null, async(function(reason) {
+      ok(true, "The rejection handler was called");
+    }));
+  });
+});
--- a/packages/ember-data/lib/system/store/finders.js
+++ b/packages/ember-data/lib/system/store/finders.js
@@ -165,3 +165,25 @@ export function _query(adapter, store, typeClass, query, recordArray) {
 
   }, null, "DS: Extract payload of findQuery " + typeClass);
 }
+
+export function _queryRecord(adapter, store, typeClass, query) {
+  var modelName = typeClass.modelName;
+  var promise = adapter.queryRecord(store, typeClass, query);
+  var serializer = serializerForAdapter(store, adapter, modelName);
+  var label = "DS: Handle Adapter#queryRecord of " + typeClass;
+
+  promise = Promise.cast(promise, label);
+  promise = _guard(promise, _bind(_objectIsAlive, store));
+
+  return promise.then(function(adapterPayload) {
+    var record;
+    store._adapterRun(function() {
+      var payload = normalizeResponseHelper(serializer, store, typeClass, adapterPayload, null, 'queryRecord');
+      //TODO Optimize
+      record = pushPayload(store, payload);
+    });
+
+    return record;
+
+  }, null, "DS: Extract payload of queryRecord " + typeClass);
+}


igorT:
@thaume sorry for not getting back to you sooner. Can you please rebase? I think this looks great for now. I hope we can figure out a way in the future to also allow easy passing of a custom URL and not just options.

thaume:
@igorT no problem man, the rebase is done! Let me know if you want me to open another PR to keep working on it or if it can wait!

wecc:
This should solve #1576

abobwhite:
#1326, which is closed, will also be solved by this pull request.

bakura10:
+1 on this feature. This is useful for cases like getting something by a slug instead of id (the slug is considered unique in some cases, so it's like getting by id, but needs to use a query params). Having said that, it's not really hard to workaround this.

However, it seems that in the extractQueryOne, you assume that there is only one resource, however I think it should call extractArray instead of extractSingle. The reason is that while you EXPECT only one record from your server, you are still hitting a "collection URI", not a single item URI.

thaume:
@bakura10 I don't think you'd use this method to get records through slugs, since you need to use an id to fetch.
What do you mean by "hitting a "collection URI", not a single item URI"? I think that fetchQueryOne is actually supposed to get a PromiseObject instead of an PromiseArray. Since you have to use an id, it will always return a single record.
The use case is more to add queryParams to load relationships server-side than to filter a record that you are fetching with an id (so it is unique).
What do you think?

bakura10:
What I mean by collection URI is "/users", while "item URI" is "/users/5". "/users" is expected to return a collection, so "/users?first_name=foo" is expected to return a collection too, as well as "/users?slug=bar", but in my case, as slug are unique, I can guarantee myself that I will return at most 1 element, but as this is still a collection URI, I return it as a collection (so wrapped with a "users" in plural). It would feel very strange for me to have a collection URI return things that is not a collection.

thaume:
Yes totally agree with you, but fetchQueryOne will issue a request of the
form "users/21?rel=posts" where you'll get the user with id=21 and you also
will attach all its posts.
I think you might want to try the store.query('users', { slug: bar }) to
issue "users?slug=bar" and get a collection of records.

Le mar. 10 févr. 2015 22:53, Michaël Gallego <notifications@github.com> a
écrit :

bakura10:
Haaa... Ok, I misunderstood the issue then. I thought it was actually made for use cases like a query is expected to return only one result. Sorry sorry, definitely a good idea then, I see the use case now! :)

packages/ember-data/lib/system/store.js
```diff
@@ -860,6 +860,33 @@ Store = Ember.Object.extend({
...
+    This method returns a promise, which is resolved with a `RecordArray`
+    once the server returns.
+
+    @method findQuery
+    @private
+    @param {String or subclass of DS.Model} type
+    @param {any} query an opaque query to be used by the adapter
+    @return {Promise} promise
+  */
+  findQueryOne: function(typeName, id, query) {
```
bmac: I'd be in favor of renaming the store api to `fetchQueryOne` since it looks like this will always fetch a new record. Keeping the adapter API as `findQueryOne` seems fine and consistent with the other adapter methods.

wecc:

bmac:
Marked this for discussion at tomorrow's Ember Data meeting.

packages/ember-data/lib/serializers/json_serializer.js
```diff
@@ -745,7 +745,7 @@ export default Ember.Object.extend({
     call is made to `DS.Store#findQuery`. By default this method is an
     alias for [extractArray](#method_extractArray).
 
-    @method extractFindQuery
+    @method extractFindQueryOne
```
bmac: Looks like this should stay as `extractFindQuery`

packages/ember-data/lib/serializers/json_serializer.js
```diff
@@ -757,6 +757,22 @@ export default Ember.Object.extend({
     return this.extractArray(store, type, payload, id, requestType);
   },
   /**
+    `extractFindQueryOne` is a hook into the extract method used when a
+    call is made to `DS.Store#findQueryOne`. By default this method is an
+    alias for [extractSingle](#method_extractSingle).
+
+    @method extractFindQuery
```
bmac: I think this was meant to be changed to `extractFindQueryOne`

packages/ember-data/lib/serializers/json_serializer.js
thaume: Good catch! I Guess I mixed names while changing from `findQuery` to `findQueryOne`. Thanks :)

packages/ember-data/lib/system/store.js
thaume: Then you would also rename `findQuery` -> `fetchQuery` ?

abobwhite:
Hey guys, I'm glad this is moving along. I hacked something up in my application a little over a year ago to do this. This will be a great thing to have right out of the box. Ember Data has finally moved from being powerful to powerful and flexible! I hope we can merge this soon!

iamjstates:
+1 on merging this

bmac:
@thaume I believe have this same problem in an app I maintain. However, last I chatted with @wycats and @igorT they weren't too keen on this method requiring and `id` because the opaqueness of the query params makes it hard for Ember Data to know if it can cache the resulting record. In the app I maintain, I always end up sending the same `include` params for a given model type. Is this the same problem you have? If this is the case it may make sense for Ember Data to make it easier to define custom query params on the Adapter.

thaume:
@bmac yes this is exactly the kind of use case I had in mind and your proposition totally makes sense.
I would like to know what other users think about it though, there are a couple of issues related to this PR.

abobwhite:
I think there are two cases at play here:

1. Needing to find a record by id but with a query param to get a specific version of the record or to customize what gets included on that record. Which is something that could be integrated into adapters/serializers. This is what I referred to in #1326 . Related to @bmac's point, your record may even need to pass a query param to the server to toggle whether a ```belongsTo``` or ```hasMany``` should be embedded, loaded async (traditional or via "links") which would need the per type serializer to allow for <i>dynamic configuration</i> based on these query params (has anyone discussed this? I'd be happy to work with someone on a pull request).

2. Needing to find a single record by some other property that is NOT the id. I recently came across a situation where my dynamic route segment is the name of a record and does not have the id. When accessing the route directly, I pull the record's name from the params and use it to access the single record that I want. Unfortunately, findQuery assumes that your server is going to be returning an array so depending whether your API makes that same assumption or not, you have to do some fun stuff in your serializer or utilize ```get('firstObject')```.

I think both are important but I understand that both may not get solved here. If we don't solve both, I hope we can create a new issue for the one that isn't solved.

bmac:
@abobwhite I think you are 100% correct in identifying the 2 issues that need to be solved. I suspect the first issue will need an RFC to hammer out some of the details. Like what will the API look like and when do we send the additional query parameters (always? just with get requests?). @abobwhite I'd be willing to help you with that if you'd like to work on that RFC. 

It feels like this pr is close to solving issue 2. Unfortunately, I don't think that solves @thaume's original issue. @thaume would you be interested in re working this pr to solve the case of a findQueryOne without an id? I'll try and chat with @igorT over the next few days to make sure he is on board with that direction.

thaume:
@bmac yep sure I'm 100% for finding a solution that works well on the long run.

The implementation of `findQueryOne` here is fetching one record by id and add some queryParams to it (so it solves point 1 from @abobwhite, not 2 I think). This is actually the implementation I need for a use case in my application (to be able to sideload data based on the queryParams). Regarding caching, there is no caching done with the implementation right now (so it fetches fresh record each time).

What we might be looking for is the following: 
-> `findQueryOne('user', 10, { include: 'posts' });` should register a classic `findOne` call and cache the result (the user with `id === 10`). The returned data is a **PromiseObject**.
-> At the same time (in the adapter/serializer), we would add some queryParams that could not modify the record fetched (it must be the user with `id === 10`).

**Result**: The targeted record by id is cached because the queryParams is not supposed to change it. It just adds data that comes sideloaded with it.

I know it is a pretty small use case, but many developers came to me asking for help on how they could "load a record as a **PromiseObject** and add some query params to it to add relationship that they didn't wanted to load asynchronously".

thaume:
@igorT @bmac Ok guys should I continue on this PR or maybe we should wait for the RFC to be finished ?

pearkes:
My use case, which is more along the lines of what @abobwhite has mentioned in 2, is that IDs can not always be shown the user, and or may not be the best thing for a user experience.

Given a situation where there is an identifier that is not globally unique, but unique under a resource scope, Ember data makes it hard to cleanly route with these constraints. 

A great example would be GitHub issues – if you look to the URL bar on this issue, you'll notice the ID passed through is `2584`. That's not globally unique on GitHub, but is unique on the issue scope of `emberjs/data`.

This is a pretty common pattern as I've seen it, so it make sense to have tools to support it.

Writing your own adapter is a path towards this, but something like this method are still required to pass data form the application layer into the `buildURL` method – getting `emberjs/data` out of the `Issue` is challenging. You end up doing something like this:

```javascript
this.store.find('issue', params.issue_number, {name: 'data', username: 'emberjs'});
```

`oneQuery` helps a bit, as it lets you pass this data as in the above example. I recognize that there's probably a broader solution for this type of routing, and it's not just fixed by implementing this query method. I just wanted to point out one example of it's use and need in the real world.

mar3les:
+1

joshhornby:
Big +1 from me! We are using an api that uses lots of ```?embed=foo``` and this will be a huge help

thaume:
This issue is going to be closed, this solution was notre chosen to handle
the problem it is trying to solve :)

Although I can't point you to any new work toward this problem

Maybe somebody else know where it is going ?

abobwhite:
@thaume I am confused by your comment. Is there is a solution to this now?

I don't see why this issue should be closed.

thaume:
It was stated that this pull request is doing too much and that it should
go through a RFC first, or maybe I missed something ? :)

joshhornby:
If this isn't going to be in core then does anyone have a nice solution for handling query params with regards to ember data?

sio4:
+1 please merge it.

thaume:
So @bmac what's the status of this feature ? Should I rebase changes or close the PR ?

abobwhite:
So maybe we need to open up separate issues to discuss the scenarios I listed above separately?

thaume:
@abobwhite yes that's a good idea, how should we move on ? Create a new PR for your first point and edit this PR for your second point ?

bmac:
@thaume Last I talked with @tomdale @wycats and @igorT about this issue was they thought `findQueryOne` should be used to fetch a single resource without requiring an `id`. 

Instead they were working on an alternate API to allow for passing query params into `find`/`fetch`. I'll ping them to see where that stands right now.

DevBlaaast:
@bmac roger that ! Let me know if we need to kill this PR and start something new once you have more information !

thaume:
@bmac roger that ! Let me know how this PR should continue once you have more information !

tschoartschi:
any news on this issue? I came here through https://github.com/emberjs/data/issues/1326. And I would need the possibility to find a record by ID and add query params to the URL. I explained our use case in the previous issue: https://github.com/emberjs/data/issues/1326#issuecomment-107599384

thaume:
@tschoartschi we are waiting for a response from the ember data core team to know wether or not we should continue or completely change the API. Many people are waiting for this new feature, it should land before June 12th and ED 1.0
cc @igorT @tchak @bmac

bmac:
Hey @thaume could I get your eyes on https://github.com/emberjs/rfcs/pull/61? That rfc proposes an alternate `fetchQueryOne` to the one you have proposed here. However, I would like to reserve a key on the new `adapterOptions` parameter for supplying query parameters to the adapter.

thaume:
@bmac awesome ! I'm taking a look right now !

thaume:
@igorT I think we're good to, once this is merged I'll move on to the implementation in the serializer. Let me know

packages/ember-data/lib/serializers/json-serializer.js
```diff
@@ -1224,6 +1224,22 @@ export default Serializer.extend({
...
+
+    @method extractQueryRecord
+    @param {DS.Store} store
+    @param {DS.Model} typeClass
+    @param {Object} payload
+    @param {(String|Number)} id
+    @param {String} requestType
+    @return {Object} object A hash of deserialized object
+  */
+  extractQueryRecord: function(store, typeClass, payload, id, requestType) {
```
igorT: This is the old style serializer hook, so not needed.

packages/ember-data/lib/system/store.js
```diff
@@ -993,6 +994,37 @@ Store = Service.extend({
...
+  */
+  queryRecord: function(modelName, query) {
+    Ember.assert("You need to pass a type to the store's queryRecord method", modelName);
+    Ember.assert("You need to pass a query hash to the store's queryRecord method", query);
+    Ember.assert('Passing classes to store methods has been removed. Please pass a dasherized string instead of '+ Ember.inspect(modelName), typeof modelName === 'string');
+
+    var typeClass = this.modelFor(modelName);
+    var adapter = this.adapterFor(modelName);
+
+    Ember.assert("You tried to load a queryRecord but you have no adapter (for " + typeClass + ")", adapter);
```
igorT: We are not loading a queryRecord, we are trying to do a query

packages/ember-data/tests/integration/adapter/query-record-test.js
```diff
@@ -0,0 +1,81 @@
+var Person, store, env;
+var run = Ember.run;
+
+module("integration/adapter/query-record - Query one record with a query hash", {
```
igorT: This is not an adapter integration test, as those are integration tests of adapters. I would put in a tests/integration/store or something like that

igorT:
:+1: Ready to merge after the small cleanup & squashing

thaume:
@igorT fixed !

thaume:
@igorT juste need to squash and we're good to go

DevBlaaast:
@igorT commits squashed ! Thanks for the review !

thaume:
commits squashed ! Thanks for the review !

packages/ember-data/tests/integration/store/query-record-test.js
```diff
@@ -0,0 +1,81 @@
...
+    }
+  }));
+
+  run(function() {
+    store.queryRecord('person', {}).then(null, async(function(reason) {
+      ok(true, "The rejection handler was called");
+    }));
+  });
+
+  ok(!store.hasRecordForId('person', 1), "The record has been unloaded");
```
igorT: Why is this id 1?

packages/ember-data/lib/system/adapter.js
```diff
@@ -189,6 +189,44 @@ var Adapter = Ember.Object.extend({
...
+    ```
+
+    @method queryRecord
+    @param {DS.Store} store
+    @param {subclass of DS.Model} type
+    @param {Object} query
+    @param {String} id
+    @return {Promise} promise
+  */
+  queryRecord: null,
```
bmac: Do we need a default implementation for the `RESTAdapter`?

packages/ember-data/lib/system/store.js
```diff
@@ -1022,6 +1023,37 @@ Store = Service.extend({
     language for all server-side queries, and then require all adapters to
     implement them.
 
+    This method returns a promise, which is resolved with a `RecordObject`
+    once the server returns.
+
+    @method findOne
```
bmac: `@method queryRecord`

packages/ember-data/lib/system/store.js
```diff
@@ -1022,6 +1023,37 @@ Store = Service.extend({
...
+    @private
```
bmac: This can be removed queryRecord will be public.

packages/ember-data/lib/serializers/json-serializer.js
bmac: I think its ok to leave this in to help smooth the transition to the new style hooks. @wecc is a `normalizeQueryRecordResponse` method needed?

packages/ember-data/lib/system/adapter.js
igorT: Yes but separate PR

packages/ember-data/lib/serializers/json-serializer.js
igorT: Deny, there is no transition, as this method never existed until now

packages/ember-data/tests/integration/store/query-record-test.js
thaume: Yep that was an old test from the "fetchQueryOne" method which worked with an id, removing the test

thaume:
@bmac @igorT changes have been made

igorT:
Thanks!

allthesignals:
I'm confused. I thought queryRecord [mentioned here](http://emberjs.com/blog/2015/06/18/ember-data-1-13-released.html#toc_query-and-queryrecord) enables something like `this.store.findQueryOne('post', 1, { include: 'comments' });`. Am I misunderstanding? This function builds to `/posts?1`. 

(For context, my question stems from https://github.com/emberjs/data/issues/3454).

Thanks!

DevBlaaast:
That was the old implementation. Now you cannot pass an id to queryRecord,
only a query object : queryRecord('post', { query: { slug: 'my-post' } });

Fryie:
this must be the most confusing PR i have ever seen! Can somebody clarify how to make what @allthesignals is asking for is supposed to work?

wecc:
@Fryie feel free to ping me in the Ember Community Slack, happy to help

hunterboerner:
@wecc what's the solution you came up with?

igorT:
The conversation which helped @hunterboerner for the record
```
wecc [3:03 AM] 
@hunterboerner: `queryOne` became `queryRecord`, kind of. The idea is that you shouldn't need an ID to fetch a single record, like if you only have a slug: `store.queryRecord('post', { slug: 'my-post' })`(edited)

​[3:03] 
by default this would give you `/posts?slug=my-post` iirc

​[3:04] 
`store.findRecord('post', 123)` on the other hand accepts an ID and would give you the URL `/posts/123` by default

​[3:05] 
so... what it kind of boils down to is "How do I pass query params to `findRecord`" and/or "How do I pass the ID to `queryRecord`"(edited)

​[3:07] 
> How do I pass query params to `findRecord`?
You can use the `adapterOptions` to pass additional information to your adapter, and pass a query hash in there. Here's a very simple implementation of this: https://github.com/emberjs/data/issues/3596#issuecomment-126604014(edited)

wecc [3:12 AM] 
> How do I pass the ID to `queryRecord`?
Since `queryRecord` only accepts `modelName` and `query` you would have to add `id` to the `query` hash and then customize the `urlForQueryRecord` hook to add the ID to the URL (and possibly remove the ID from the query in the `queryRecord` hook)
```

terion-name:
@igorT why is all this so complicated? 
It is a such common use case, it is a part of json-api spec, so many requests for this and still no straight way to add query parameters to 'findRecord'. Why?

jeffjarchow:
I agree. This is a very simple use case. There is no way I want to send a request like
`/posts?id=1&include=replies` when I should be able to send `/posts/1?include=replies`. I believe this is the problem that is trying to be solved here with still no solution. It is necessary to be able to send the id AND query params in a single request.

When you are dealing with a database that has hundreds of millions of records, your 10 ms query could turn into a few seconds. From an API side, most programmers don't look to see if the id is one of the many possible query params. That means if your default query limit for a GET list is 20 records, the DB Engine will search for a 'like' match to the id query param until it a) finds its 20 matches or b) reaches the end of the table. Where as a query looking for a single record sets limit to 1 and returns an exact match as soon as it finds the indexed record. Big performance difference on the db.

Luckily for me I have control over the API that I am connecting to with Ember.

terion-name:
@jeffjarchow I now use a solution provied by @igorT:
```js
export default DS.JSONAPIAdapter.extend({
  urlForFindRecord(id, modelName, snapshot) {
    let url = this._super(...arguments);
    let query = Ember.get(snapshot, 'adapterOptions.query');
    if (query) {
      url += '?' + Ember.$.param(query); // assumes no query params are present already
    }
    return url;
  }
});

this.store.findRecord('project', params.project_id, {adapterOptions: {query: {include: 'user,developer', limit: 1}}});
```
This works, but so hacky :(

jeffjarchow:
Thank you @terion-name for the snippet. I'll have to play with that to see if I can get that working. I would just be afraid of it not being compatible as things change. Currently I am on a major upgrade from 1.10 to 1.13 and all the changes that go along with that, like eliminating views.

EarvinKayonga:
@terion-name, U rock.

Subtletree:
For anyone else finding this, since Ember 2.4 it's built in: [ds-finder-include](http://emberjs.com/blog/2016/01/12/ember-data-2-3-released.html#toc_code-ds-finder-include-code)

`var article = this.store.findRecord('article', 1, { include: 'comments' });`

PR #2586, opened 2014-12-12
Work around type check issues with MODEL_FACTORY_INJECTIONS.

hjdivad:
The container subclasses types retrieved via `modelFor`, so that if A is a
subtype of B, modelFor('model:A') will not be a subtype of
modelFor('model:B'). This breaks the type check for polymorphic
associations.

This commit works around this issue by checking the superclass when doing
its type checks, if MODEL_FACTORY_INJECTIONS is enabled.

Paired with @igort


mattbeedle:
:+1: Awesome!

Issue #2600, opened 2014-12-16
ActiveModelSerializer doesn't properly translate the error keys for associations


piotrpalek:
Hey, so I've come across this weird issue, where I had an `offerCategory` property (relationship) on my model, and when I saved it, and the server responded with 422 it didn't populate the models `errors` property with the corresponding error.

As far as I've checked the response should be good, so I changed the property name to `offer_category` and suddently the errors were populated.

I think I've found the bug and fixed a test which was supposed to test this case, but wasn't actually testing it (I think..). Link here: https://github.com/piotrpalek/data/commit/15381f953c6e63b336667e11cec50f6c857ebfb8.

piotrpalek:
@sly7-7 any thoughts on this, what could be the solution?

sly7-7:
@piotrpalek To be honest, I don't know. Actually there is a bug, or at least Rails AMS and Ember-Data AMS are inconsistent. Before doing anything, I'd like to know where the truth is.
cc/ @kurko @igorT

tonycoco:
This is currently still an issue with AMS and DS.Errors, the above reference PR #2601 still fixes it too. Any idea why that was closed without a merge?

kurko:
Given Ember has an adapter for AMS, I believe the truth is likely to be AMS. IMO Ember should then conform to it. I don't use 0.9 so I don't really have enough info to help, but I'd suggest fixing this on the Ember side with the #2601 PR mentioned above.

pangratz:
I think this issue has been addressed in #3030, #3036 and #3040 ...

sly7-7:
@pangratz Thanks a lot, I'm closing it. @piotrpalek @tonycoco It would be great if you could confirm this issue is gone.

PR #2601, opened 2014-12-16
ActiveModelAdapter doesn't camelize errors

piotrpalek:
As explained in #2600, the ActiveModelAdapter doesn't camelize the errors in `ajaxResponse` which causes them to not be included in the models errors hash.
This is a first attempt at fixing the issue, as far as I've tested "it works for me" (tm)


piotrpalek:
@sly7-7 I don't have the code anymore in my repo but I'll try to update the jsbin if I got time

Issue #2588, opened 2014-12-13
EmbeddedRecordsMixin fails when using a store other then 'store:main'


jmurphyau:
I have a situation where I'm not using 'store:main' because the purpose of my application is to compare objects in different environments - so I'm using 'store:production' and 'store:test'.

The EmbeddedRecordsMixin uses ['this.store'](https://github.com/emberjs/data/blob/a83aea92457fc1c34a3277f811ea91f5b12664b8/packages/ember-data/lib/serializers/embedded_records_mixin.js#L123) which is [injected](https://github.com/emberjs/data/blob/d5f19518cb7b7e62b650ce6ce2329f9d0f1557e1/packages/ember-data/lib/initializers/store_injections.js#L11), whereas every function on a serializer (that I can see anyway) appears to use a store variable that is provided externally.

bmac:
Merged as https://github.com/emberjs/data/pull/2915

PR #2617, opened 2014-12-19
Store Managed Adapters & Serializers

jmurphyau:
Pull request to make a store instance responsible for it's own adapter and serializer instances - as suggested fix to #2588 by @tomdale.

Serializers and Adapters are no longer singletons.

Failing test included.

### DIFF
--- a/packages/ember-data/lib/system/store.js
+++ b/packages/ember-data/lib/system/store.js
@@ -26,7 +26,6 @@ import {
 } from "ember-data/system/store/common";
 
 import {
-  serializerFor,
   serializerForAdapter
 } from "ember-data/system/store/serializers";
 
@@ -1985,7 +1984,7 @@ function defaultSerializer(container) {
 function _commit(adapter, store, operation, record) {
   var type = record.constructor;
   var promise = adapter[operation](store, type, record);
-  var serializer = serializerForAdapter(adapter, type);
+  var serializer = serializerForAdapter(store, adapter, type);
   var label = "DS: Extract and notify about " + operation + " completion of " + record;
 
   Ember.assert("Your adapter's '" + operation + "' method must return a value, but it returned `undefined", promise !==undefined);
--- a/packages/ember-data/lib/system/store.js
+++ b/packages/ember-data/lib/system/store.js
@@ -205,6 +205,7 @@ Store = Ember.Object.extend({
       store: this
     });
     this._pendingSave = [];
+    this._containerCache = Ember.create(null);
     //Used to keep track of all the find requests that need to be coalesced
     this._pendingFetch = Map.create();
   },
@@ -271,7 +272,8 @@ Store = Ember.Object.extend({
 
     if (DS.Adapter.detect(adapter)) {
       adapter = adapter.create({
-        container: this.container
+        container: this.container,
+        store: this
       });
     }
 
@@ -1776,20 +1778,26 @@ Store = Ember.Object.extend({
   // ......................
 
   /**
-    Returns the adapter for a given type.
+    Returns an instance of the adapter for a given type. For
+    example, `adapterFor('person')` will return an instance of
+    `App.PersonAdapter`.
+
+    If no `App.PersonAdapter` is found, this method will look
+    for an `App.ApplicationAdapter` (the default adapter for
+    your entire application).
+
+    If no `App.ApplicationAdapter` is found, it will return
+    the value of the `defaultAdapter`.
 
     @method adapterFor
     @private
-    @param {subclass of DS.Model} type
+    @param {String or subclass of DS.Model} type
     @return DS.Adapter
   */
   adapterFor: function(type) {
-    var adapter;
-    var container = this.container;
+    type = this.modelFor(type);
 
-    if (container) {
-      adapter = container.lookup('adapter:' + type.typeKey) || container.lookup('adapter:application');
-    }
+    var adapter = this.lookupAdapter(type.typeKey) || this.lookupAdapter('application');
 
     return adapter || get(this, 'defaultAdapter');
   },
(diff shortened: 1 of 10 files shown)


igorT:
@jmurphyau apologies for not getting back to you sooner. Getting up to speed now after the holidays.  Wouldn't the simplified order prevent people from having custom adapters for custom stores?

igorT:
Can you also please rebase/squash the PR?

packages/ember-data/lib/system/store.js
```diff
@@ -250,7 +252,8 @@ Store = Ember.Object.extend({
 
     if (DS.Adapter.detect(adapter)) {
```
igorT: I think so because we just looked it up couple lines above?

igorT:
Awesome work btw @jmurphyau

jmurphyau:
@igorT For a custom adapter on a custom store, this currently exists as item 3 in the current implementation:

`defaultSerializer`: `this.adapter` (default to `-rest`)

and item 2 in what I've documented as a simplified preference/order:

`defaultAdapter`: `this.adapter` (default to `undefined`)

Also, something worth noting that may not be completely obvious to everyone reading this PR - the "simplified preference/order" I've detailed in the comments above **is not actually implemented in this PR** - this PR is only about allowing multiple stores to work with the embedded record mixin. 

Should I split out the preference/order discussion into something/somewhere else? If so, where is appropriate?

packages/ember-data/lib/system/store.js
sly7-7: This seems inconsistent to me. I mean, I thought we now should always lookup adapter/serializer via the container and not instanciating it. Also, we encourage strings everywhere, but in this case we still want to fully support a store definition with an adapter property a class ?

jmurphyau:
I think I've rebased this incorrectly - I'm not 100% sure?? 

Sorry guys don't really have much experience with Git.. I'll investigate see if I've fucked it up/need to change it..

igorT:
Can you try 
`git reset --hard 3e424a46a563083600570bf00f72759247c8bf5c` when you are in your branch
then force push the branch?

jmurphyau:
Done - that looks much better. Thanks @igorT.

I think using GitHub for Mac may have caused part of the problem

packages/ember-data/lib/system/store.js
bmac: Isn't the adapter here a result of the looked up string?

packages/ember-data/tests/integration/setup-container-test.js
```diff
@@ -38,7 +38,7 @@ test("the deprecated serializer:_default is resolved as serializer:default", fun
     deprecated = container.lookup('serializer:_default');
   });
 
-  ok(deprecated === valid, "they should resolve to the same thing");
+  ok(deprecated.constructor === valid.constructor, "they should resolve to the same thing");
```
bmac: Can you add a test that assert your config for singleton:false

packages/ember-data/tests/integration/serializers/embedded_records_mixin_test.js
```diff
@@ -1320,3 +1322,76 @@ test("serializing relationships with an embedded and without calls super when no
...
+    villains: [{
+    id: "1",
+    first_name: "Jade",
+    last_name: "John"
+    }]
+  }
+  };
+  var json, json_primary, json_secondary;
+
+  run(function(){
```
bmac: please indent

packages/ember-data/tests/integration/serializers/embedded_records_mixin_test.js
```diff
@@ -1320,3 +1322,76 @@ test("serializing relationships with an embedded and without calls super when no
   ok(calledSerializeBelongsTo);
   ok(calledSerializeHasMany);
 });
+
+test("embedded records should be created in multiple stores", function() {
```
bmac: Can you extract this into a specific module that tests support for multiple stores. I don't think this is embedded records specific

packages/ember-data/lib/system/store.js
sly7-7: hum, ok, I'm probably missing something, I thought the lookup returns an instance, not a class.

igorT: Ah maybe you're right i thought you meant class vs a string

sly7-7: Well, I think I'm just too bad explaining what I have in mind :confounded: 
So I will try again.

Historically, the user can affect a class or a string to the `adapter` property. As a result, there is this chain of checks. If the `adapter` property is a string, then the adapter is looked up via the container, or if it's a class (actually a subclass of DS.Adapter), then it is instantiated here.

My question is: do we still want to allow the user define the `adapter` property as a class. It seems to me this would be inconsistent with the "string everywhere" paradigm.

Please, tell me it's clear now... or tell me to take some intensive English training :smile_cat:

bmac:
ping @jmurphyau

jmurphyau:
@bmac - I must have missed the last few comments from you.

I'll split out those tests & add the tests you mentioned over the next day or so.

jmurphyau:
I've split the tests out - please review.

jmurphyau:
@bmac @igorT @sly7-7 - any further thoughts/suggestions on this PR?

packages/ember-data/lib/system/store.js
```diff
@@ -185,6 +185,7 @@ Store = Ember.Object.extend({
       store: this
     });
     this._pendingSave = [];
+    this._containerCache = Ember.create(null);
```
stefanpenner: why do we need a cache?

jmurphyau: It's where the store holds its Serializers and Adapters instances - since each is now one instance per store rather than a singleton.

stefanpenner: ah ok

stefanpenner: on destroy, we should be sure to also destroy these. So they can do needed cleanup

jmurphyau: Ok cool - I'll get onto that

stefanpenner: :+1:

jmurphyau:
@stefanpenner - I've made the changes discussed and a few additional ones since the serializerForAdapter function has moved.

igorT:
Thanks for the awesome work! We rebased it and merged it, but kept you as an author. Thanks!

jmurphyau:
Thanks everyone for all you help & feedback with this PR - it was my first PR ever. Was a good experience and I've since learnt a lot about Git and GitHub and moved away from GitHub for Mac in favour of the command line - so shouldn't have any issues with rebase again :)

I've created two other PRs that have stemmed from this PR

##### [#2918 Remove serializerForAdapter](https://github.com/emberjs/data/pull/2918)

This has stemmed from @bmac's issue #2916 in relation to keeping things consistent.

##### [#2919 Serializer & Adapter Lookups Preference/Order](https://github.com/emberjs/data/pull/2919)

This has stemmed from the above discussion started by @sly7-7 around the lookup order of serializers and adapters. Parts of the discussion are hidden behind outdated diffs (e.g. it started in 'sly7-7 commented on an outdated diff on Dec 20, 2014')

Serializer and adapter order/preference was discussed above but nothing really appeared to be decided - so I've started with a base - something to go off - which I hope gets things moving in this regard.
It can be changed or improved on or just closed (if now is not the time for this type of change, or if it's not a good idea)

PR #2621, opened 2014-12-22
WIP: Add options hash to model.save

knownasilya:
**WIP - Don't merge or close please**

## Use Case

Creating more specific methods apart from the basic CRUD methods.

For example:

```js
user.revoke();
// DELETE api/users/123/revoke
```
Where the endpoint checks if the user has accepted the invitation, if not it's revoked. 

Passing options to `Model#save` allows for creating methods like this, instead of writing
ajax calls and executing `model.unloadRecord()`.

## Usage

The hash is passed to `model.save({ /* .. */ });` and passed to the adapter method that is
responsible for the save, where the adapter will do what's needed, usually passing the data to the server, or to the serializer.

## Questions to Address
- [x] Should the adapter accept another argument, or pass via model as `model._options` or something.
  * pass another argument to adapter/serializer
- [ ] Should options be passed via QPs or the body, e.g. `{ meta: { /* here */ } }` or directly for delete

cc @igorT @grapho

### DIFF
--- a/packages/ember-data/lib/system/adapter.js
+++ b/packages/ember-data/lib/system/adapter.js
@@ -334,6 +334,7 @@ var Adapter = Ember.Object.extend({
     @param {DS.Store} store
     @param {subclass of DS.Model} type   the DS.Model class of the record
     @param {DS.Model} record
+    @param {Object} options
     @return {Promise} promise
   */
   createRecord: Ember.required(Function),
@@ -374,6 +375,7 @@ var Adapter = Ember.Object.extend({
     @param {DS.Store} store
     @param {subclass of DS.Model} type   the DS.Model class of the record
     @param {DS.Model} record
+    @param {Object} options
     @return {Promise} promise
   */
   updateRecord: Ember.required(Function),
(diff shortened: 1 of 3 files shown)


mphasize:
I would support a feature like this, since it makes custom actions on REST api's possible. My use case would be to build a "publish" action for a model, like `/post/[id]/publish`.

metanoic:
My use case for this feature is to provide the serializer with the options passed to model.save, as we issue a single request containing a given object graph as specified by the programmer - e.g. person.save({ graph: ["phones", "emails", "address", "address.addressDetails"]}) includes the related models in sideloaded format (these relationships are not *always* desired for person, it is a per-case basis).

Having to alter ED each time we upgrade is quite a chore; having to not worry about this particular alteration would be very nice.  

We also pass the options down to the store so we can specify query params to a sibling of findById (e.g. findByQuery) (we also force it to not hit the typeMaps, as we don't want cached results if a second request specifies diff query params).

I'm most interested in the first use case, but it would be pretty sweet if the options were passed down to all the store.find delegates as it'd reduce my chore even more.

igorT:
Every adapter method will now be receiving adapterOptions. @bmac maybe useful PR to build off of

knownasilya:
Nice!

bmac:
This is great @knownasilya!

PR #2623, opened 2014-12-23
Initial implementation of the Snapshot API

### LINKED ISSUE #2385: Async Polymorphic relations don't serialize correctly

### LINKED ISSUE #2508: `RESTSerializer.serializePolymorphicType` broken for `async` relationships

### LINKED ISSUE #2739: Model.save() errors when missing related objects
Calling `this.get('user').save();` results in

```javascript
Uncaught Error: Assertion Failed: You looked up the 'login' relationship
on a 'user' with id 3 but some of the associated records were not loaded.
Either make sure they are all loaded together with the parent record, or
specify that the relationship is async (`DS.belongsTo({ async: true })`)
```

It works when I make the login relationship async, but that's not how my app is supposed to work

### LINKED ISSUE #2750: unfulfilled async hasMany relationship serializes to an empty array
When a model record that has async hasMany relationships is saved while the hasMany relationship property has yet to be fulfilled by the promise object, the serializer produces an empty array which results in incorrect representation of the relationship to be informed to the backend.

The offending logic appears to be that the promise object which implicitly has ID references to the eventual records that will be retrieved to fulfill the relationship simply returns an empty array when the property record (the PromiseArray object) returns mapBy('id') before those resolutions have been loaded.

This is problematic for representing the relationship state of the model being saved since whether the model eventually resolved those hasMany relationships or not for use should not inform an update operation for the model record where the relationship mapping for affected hasMany property gets reset to empty.

In order to work-around this issue, I've used the below snipped as a hack by extending the RESTSerializer and having the serializer *skip* the unfulfilled hasMany relationships and not including it in the resulting PUT operation.

```coffeescript
App.ApplicationSerializer = DS.RESTSerializer.extend
    serializeHasMany: (record, json, relationship) ->
        result = Ember.get record, relationship.key
        if result?.get('isFulfilled')
            this._super.apply(this, arguments)
```
The above hack only works if the server-side backend can properly deal with PUT operations where specific properties are *not* included in the JSON payload.

I'm thinking the proper fix would be to directly alter the PromiseArray used in the context of hasMany to always return all of the IDs (whether fulfilled or not by the given model) when mapBy('id') is called on it so that it does not improperly inform state about the hasMany property being *empty*.  I've dug a little and it seems that this particular scenario may require a bit of creativity since technically there are no explicit records to iterate in a *forEach* to apply the mapping to extract out the array of record IDs... yet.

I've encountered this issue in `ember-data 1.0.0-beta.8` but upon brief scan through latest code it looks like it may still be affecting current snapshots.

Your comments/feedback will be welcome for proper workaround as necessary in this scenario.  If current behavior is actually the expected behavior, would like rationale on that as well.

wecc:
This is a first stab at the new Snapshot API, `serializer.serialize()` now receives a Snapshot instead of a record instance.

- **Caching** - Right now, attributes are cached eagerly (on instantiation) and relationships are cached on first access. Is this the design we want?
- **snapshot.belongsTo()** - Returns a new snapshot (with at least an ID) or undefined. Is this what we want or should we always return a new snapshot?
- **snapshot.hasMany()** - Always returns an array.
- **snapshot.get() deprecated** - ~~To only use `attr()`, `belongsTo()` and `hasMany()` would require some work in serializers. Should that be done in the scope of this PR or separately?~~ Done in this PR.

[@sly7-7]: Just to clean issues when that awesome work is merged: 
Closes #2551, Fixes #2385, Fixes #2508, Closes #1491 

Fixes #2750, Closes #2739

### DIFF
--- a/packages/ember-data/lib/system/snapshot.js
+++ b/packages/ember-data/lib/system/snapshot.js
@@ -1,0 +1,361 @@
+/**
+  @module ember-data
+*/
+
+var get = Ember.get;
+
+/**
+  @class Snapshot
+  @namespace DS
+  @private
+  @constructor
+  @param {DS.Model} record The record to create a snapshot from
+*/
+function Snapshot(record) {
+  this._attributes = Ember.create(null);
+  this._belongsToRelationships = Ember.create(null);
+  this._belongsToIds = Ember.create(null);
+  this._hasManyRelationships = Ember.create(null);
+  this._hasManyIds = Ember.create(null);
+
+  record.eachAttribute(function(keyName) {
+    this._attributes[keyName] = get(record, keyName);
+  }, this);
+
+  this.id = get(record, 'id');
+  this.record = record;
+  this.type = record.constructor;
+  this.typeKey = record.constructor.typeKey;
+
+  // The following code is here to keep backwards compatibility when accessing
+  // `constructor` directly.
+  //
+  // With snapshots you should use `type` instead of `constructor`.
+  //
+  // Remove for Ember Data 1.0.
+  if (Ember.platform.hasPropertyAccessors) {
+    var callDeprecate = true;
+
+    Ember.defineProperty(this, 'constructor', {
+      get: function() {
+        // Ugly hack since accessing error.stack (done in `Ember.deprecate()`)
+        // causes the internals of Chrome to access the constructor, which then
+        // causes an infinite loop if accessed and calls `Ember.deprecate()`
+        // again.
+        if (callDeprecate) {
+          callDeprecate = false;
+          Ember.deprecate('Usage of `snapshot.constructor` is deprecated, use `snapshot.type` instead.');
+          callDeprecate = true;
+        }
+
+        return this.type;
+      }
+    });
+  } else {
+    this.constructor = this.type;
+  }
+}
+
+Snapshot.prototype = {
+  constructor: Snapshot,
+
+  /**
+    The id of the snapshot's underlying record
+
+    Example
+
+    ```javascript
+    var post = store.push('post', { id: 1, author: 'Tomster', title: 'Ember.js rocks' });
+    var snapshot = post._createSnapshot();
+
+    snapshot.id; // => '1'
+    ```
+
+    @property id
+    @type {String}
+  */
+  id: null,
+
+  /**
+    The underlying record for this snapshot. Can be used to access methods and
+    properties defined on the record.
+
+    Example
+
+    ```javascript
+    var json = snapshot.record.toJSON();
+    ```
+
+    @property record
+    @type {DS.Model}
+  */
+  record: null,
+
+  /**
+    The type of the underlying record for this snapshot, as a subclass of DS.Model.
+
+    @property type
+    @type {subclass of DS.Model}
+  */
+  type: null,
+
+  /**
+    The name of the type of the underlying record for this snapshot, as a string.
+
+    @property typeKey
+    @type {String}
+  */
+  typeKey: null,
+
+  /**
+    Returns the value of an attribute.
+
+    Example
+
+    ```javascript
+    var post = store.createRecord('post', { author: 'Tomster', title: 'Ember.js rocks' });
+    var snapshot = post._createSnapshot();
+
+    snapshot.attr('author'); // => 'Tomster'
+    snapshot.attr('title'); // => 'Ember.js rocks'
+    ```
+
+    Note: Values are loaded eagerly and cached when the snapshot is created.
+
+    @method attr
+    @param {String} keyName
+    @return {Object} The attribute value or undefined
+  */
+  attr: function(keyName) {
+    if (keyName in this._attributes) {
+      return this._attributes[keyName];
+    }
+    throw new Ember.Error("Model '" + Ember.inspect(this.record) + "' has no attribute named '" + keyName + "' defined.");
+  },
+
+  /**
+    Returns all attributes and their corresponding values.
+
+    Example
+
+    ```javascript
+    var post = store.createRecord('post', { author: 'Tomster', title: 'Ember.js rocks' });
+    var snapshot = post._createSnapshot();
+
+    snapshot.attributes(); // => { author: 'Tomster', title: 'Ember.js rocks' }
+    ```
+
+    @method attributes
+    @return {Array} All attributes for the current snapshot
+  */
+  attributes: function() {
+    return Ember.copy(this._attributes);
+  },
+
+  /**
+    Returns the current value of a belongsTo relationship.
+
+    `belongsTo` takes an optional hash of options as a second parameter,
+    currently supported options are:
+
+   - `id`: set to `true` if you only want the ID of the related record to be
+      returned.
+
+    Example
+
+    ```javascript
+    var post = store.push('post', { id: 1, title: 'Hello World' });
+    var comment = store.createRecord('comment', { body: 'Lorem ipsum', post: post });
+    var snapshot = comment._createSnapshot();
+
+    snapshot.belongsTo('post'); // => DS.Snapshot of post
+    snapshot.belongsTo('post', { id: true }); // => '1'
+    ```
+
+    Calling `belongsTo` will return a new Snapshot as long as there's any
+    data available, such as an ID. If there's no data available `belongsTo` will
+    return undefined.
+
+    Note: Relationships are loaded lazily and cached upon first access.
+
+    @method belongsTo
+    @param {String} keyName
+    @param {Object} [options]
+    @return {DS.Snapshot|String|undefined} A snapshot or ID of a belongsTo relationship, or undefined
+  */
+  belongsTo: function(keyName, options) {
+    var id = options && options.id;
+    var result;
+    var relationship, inverseRecord;
+
+    if (id && keyName in this._belongsToIds) {
+      return this._belongsToIds[keyName];
+    }
+
+    if (!id && keyName in this._belongsToRelationships) {
+      return this._belongsToRelationships[keyName];
+    }
+
+    relationship = this.record._relationships[keyName];
+    if (!(relationship && relationship.relationshipMeta.kind === 'belongsTo')) {
+      throw new Ember.Error("Model '" + Ember.inspect(this.record) + "' has no belongsTo relationship named '" + keyName + "' defined.");
+    }
+
+    inverseRecord = get(relationship, 'inverseRecord');
+    if (id) {
+      if (inverseRecord) {
+        result = get(inverseRecord, 'id');
+      }
+      this._belongsToIds[keyName] = result;
+    } else {
+      if (inverseRecord) {
+        result = inverseRecord._createSnapshot();
+      }
+      this._belongsToRelationships[keyName] = result;
+    }
+
+    return result;
+  },
+
+  /**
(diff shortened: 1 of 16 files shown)


packages/ember-data/lib/system/snapshot.js
```diff
@@ -0,0 +1,299 @@
+/**
+  @module ember-data
+*/
+
+var get = Ember.get;
+
+var hasPropertyAccessors = Ember.__loader.require('ember-metal/platform')['hasPropertyAccessors'];
```
rwjblue: This are exported by Ember as `Ember.platform.hasPropertyAccessors ` since at least 1.8.

https://github.com/emberjs/ember.js/blob/v1.8.1/packages/ember-metal/lib/main.js#L183-L186

packages/ember-data/lib/system/snapshot.js
```diff
@@ -0,0 +1,299 @@
...
+var defineProperty = Ember.__loader.require('ember-metal/properties')['defineProperty'];
```
rwjblue: This is exported as `Ember.defineProperty` (for a long time) and also as `Ember.platform.defineProperty` since at least 1.8.

https://github.com/emberjs/ember.js/blob/v1.8.1/packages/ember-metal/lib/main.js#L183-L186

packages/ember-data/lib/system/snapshot.js
```diff
@@ -0,0 +1,299 @@
...
+  },
+
+  /**
+    @method get
+    @param {string} keyName
+    @return {Object} The property value
+    @deprecated Use [attr](#method_attr), [belongsTo](#method_belongsTo) or [hasMany](#method_hasMany) instead
+  */
+  get: function(keyName) {
+    Ember.deprecate('Using DS.Snapshot.get() is deprecated. Use .attr(), .belongsTo() or .hasMany() instead.');
```
rwjblue: This seems like a completely new API, why would we come out of the gates with deprecations?

Pardon me if this has already been discussed...

packages/ember-data/lib/system/snapshot.js
```diff
@@ -0,0 +1,283 @@
...
+    var callDeprecate = true;
+
+    Ember.defineProperty(this, 'constructor', {
+      get: function() {
+        /**
+          Ugly hack since accessing error.stack (done in `Ember.deprecate()`)
+          causes the internals of Chrome to access the constructor, which then
+          causes an infinite loop if accessed and calls `Ember.deprecate()`
+          again.
+        */
```
wecc: Ideas on how to improve this?

packages/ember-data/lib/system/snapshot.js
wecc: I think @igorT, @wycats and @tomdale have discussed this.

igorT: @rwjblue before when serializing something you would get a record and call `.get()` to access the properties. Now you will receive a snapshot and will call `.attr()`, `.belongsTo()`, `.hasMany()`. We still want to implement a depracated support for get in order to not break everyone's serializer

rwjblue: @igorT - I see, thanks for the clarification.

wecc:
Add test to make sure relationships aren't fetched.

packages/ember-data/lib/system/snapshot.js
```diff
@@ -0,0 +1,298 @@
...
+
+    if (!(keyName in this._hasManyRelationships)) {
+      relationship = this.record._relationships[keyName];
+
+      if (!(relationship && relationship.relationshipMeta.kind === 'hasMany')) {
+        throw new Ember.Error("Model '" + Ember.inspect(this.record) + "' has no hasMany relationship named '" + keyName + "' defined.");
+      }
+
+      members = get(relationship, 'members');
+      members.forEach(function(member) {
```
bmac: Don't create the snapshots if we only want the ids

packages/ember-data/lib/system/snapshot.js
```diff
@@ -0,0 +1,298 @@
...
+    data available, such as an ID. If there's no data available `belongsTo` will
+    return undefined.
+
+    Note: Relationships are loaded lazily and cached upon first access.
+
+    @method belongsTo
+    @param {string} keyName
+    @return {DS.Snapshot|undefined} A snapshot of a belongsTo relationship or undefined
+  */
+  belongsTo: function(keyName) {
```
bmac: Should this also support `snapshot.belongsTo(attr, { ids: true })` like `hasMany`?

packages/ember-data/lib/system/snapshot.js
```diff
@@ -0,0 +1,298 @@
...
+
+  this._id = get(record, 'id');
+
+  record.eachAttribute(function(keyName) {
+    this._attributes[keyName] = get(record, keyName);
+  }, this);
+
+  this.record = record;
+  this.type = record.constructor;
+  this.typeKey = record.constructor.typeKey;
```
bmac: We should have docs for these properties.

packages/ember-data/lib/system/snapshot.js
```diff
@@ -0,0 +1,298 @@
+/**
+  @module ember-data
+*/
+
+var get = Ember.get;
+
+/**
+  @class Snapshot
+  @namespace DS
+  @private
```
bmac: I don't think this needs to be marked private.

packages/ember-data/lib/system/snapshot.js
wecc: My reasoning for this is to cache the hasMany on first access, no matter if you want just the IDs or the snapshots.

Imagine this:

```javascript
var snapshot = post.snapshot();
var ids = snapshot.hasMany('comments', { ids: true });
var comments = snapshot.hasMany('comments');
```

If we don't create the snapshots and cache them on the first access we could have a scenario when `ids` and `comments` in the example above doesn't match.

Thoughts?

packages/ember-data/lib/system/snapshot.js
wecc: Yeah, that might be useful. There's one place that could benefit from this, see https://github.com/emberjs/data/pull/2623/files#diff-874af7e7df6c1be12e3a54caad8d859dR583

However, `ids` feels wrong to me since it's just one id? But `id` would be inconsistent to `hasMany`.

@igorT ?

packages/ember-data/lib/system/snapshot.js
igorT: You wouldn't really be serializing twice while changing the data inbetween and the main reason we added `ids:true` was for perf so you don't have to create snapshots right?

packages/ember-data/lib/system/snapshot.js
wecc: Done!

packages/ember-data/lib/system/snapshot.js
wecc: I didn't remove `@private` to make sure constructor is hidden in API docs.

wecc:
- Added docs for `record`, `type` and `typeKey`
- Added `{ id: true }` option to `snapshot.belongsTo()` + test
- Snapshots are never created if passing `id/ids: true` to `snapshot.belongsTo()/hasMany()`, IDs gets cached separately

wycats:
@wecc you are a wonderful human being. Thanks so much for this!

wecc:
Merge conflicts fixed.

knownasilya:
So what exactly is a snapshot, and how would you use it?

wecc:
@knownasilya a Snapshot is a lighter representation of a record and it's current attributes and relationships. It's initial purpose is to be passed to serializers instead of the real records. This helps for example when serializers tries to get the current value of an async relationship, instead of fetching it (record behavior) a snapshot returns the current value regardless if it's loaded of not.

tomdale:
@knownasilya As @wecc described, the high order bit for me is that a snapshot represents the frozen state of a record at a particular moment in time. As we push everything towards being all-async all the time, you want to get a way to examine the current state of that record in the moment without triggering all sorts of side-effects, like loading relationships.

knownasilya:
Thanks for the clarification @wecc and @tomdale. It sure sounds like a useful and needed API, especially in the async scenario. I've struggled numerous times trying to get state without triggering any observers/cps in complicated scenarios, and it's a battle.

sly7-7:
@tomdale @wecc From what I understand from the code, I think there are still inconsistencies in the implementation. Since the relationships are lazy and the attributes are eager, actually when you create a snapshot, and the access its data, I think problems can occur if the relationships of the underlying are changed before the snapshot's ones are accessed. Though I just may be completely wrong.

tomdale:
@sly7-7 You're absolutely correct. We discussed this at length and decided that you need to pre-emptively access the snapshot's relationships if you want to freeze them at that moment in time. If we eagerly snapshot all relationships as well as attributes, we may have to create thousands of snapshots that the user never needs. Definitely open to other suggestions here, because it's going to trip people up. But I don't know how to avoid making it easy to snapshot the entire object graph otherwise.

bakura10:
I'm not sure to completely understand this feature. Can it be used to create "forked models" (à la EPF), so hat modifications are applied locally, and then saved independently (to avoid the very famous binding problem where properties are updated visually while the model is not saved yet)

tomdale:
@bakura10 It is not that complex. It simply, at snapshot time, loops over all of the attributes of a model and copies their current values over into an immutable hash, giving you a "snapshot in time" of the state of that model.

thomasvanlankveld:
If this does what I think it will do, it's going to make me very very happy :)

packages/ember-data/lib/system/snapshot.js
```diff
@@ -0,0 +1,364 @@
...
+    var post = store.push('post', { id: 1, author: 'Tomster', title: 'Ember.js rocks' });
+    var snapshot = post._createSnapshot();
+
+    snapshot.id(); // => '1'
+    ```
+
+    @method id
+    @return {String} The ID of the snapshot's record
+  */
+  id: function() {
```
bmac: I'm nervous because now if you do `record.id` it will return the id of the record. This could break serializers that are relying on this behavior.

wecc:
Rebased, merge conflicts fixed.

packages/ember-data/lib/system/snapshot.js
```diff
@@ -0,0 +1,361 @@
...
+
+    ```javascript
+    var post = store.push('post', { id: 1, author: 'Tomster', title: 'Ember.js rocks' });
+    var snapshot = post._createSnapshot();
+
+    snapshot.id; // => '1'
+    ```
+
+    @method id
+    @return {String} The ID of the snapshot's record
```
bmac:
```
 @property id
 @type {String}
```

Issue #1829, opened 2014-03-22
Ember Data creates duplicate records


ptmoy2:
Using Ember 1.4, EmberData beta 7 with default REST adapter.  I don't have a JSBIN because problem only shows when there's a success reply from the server.  Here's the scenario:
1. create 2 routes, and transition to the 1st route.
2. create a record: `amd = store.createRecord ('myModel', myRecord);`
3. save the newly created record: `md.save();`
   note - server must return a success for problem to show.  My server returned {status, 204}
4. now transition to route 2 via link-to or something else, then hit the browser's back key to go
   back to route 1.
5. when the 'back' button is pressed, Ember makes a GET request to the server.
   (why does Ember send a GET message when it already has it in the identity map?)
   Server replies with one instance of the saved record.  Ember displays two instances of
   the saved record.  Ember inspector also shows two instances.
6. Now if the browser refresh button is pressed, Ember sends a GET to the server.  Server replies
   with 1 record, and everything looks right again.

grayt0r:
I've run into the same problem with a `belongsTo` relationship and am struggling to find a workaround. I was hoping to be able to just remove the original record (i.e. the one without the id) but it appears to still have a link to the parent record (stored in `_implicitRelationships`) which causes the link between the duplicated record (i.e. the one with the id) and the parent record to be removed when the original record is deleted.

Simple(ish) example here: http://emberjs.jsbin.com/fevepogili/1/edit?html,js,output

Steps to recreate:
- Save a new task with a description and category
- Delete the first category in the list (which shouldn't be linked to anything anymore)
- The task in the list is no longer linked to a category

Am I missing something?

jdjkelly:
I am strongly in favour of @sandstrom 's suggested approach. Given that JSON API has de-milestoned "sideposting" from 1.0, we're going to be out of luck for a while.

bmac:
This issue is rather tricky because Ember Data doesn't have enough context to know if the relationship returned from the server is the same one that already exists on the record or if the existing relationship is a new client side record that the server doesn't know about yet. 

@igorT maybe the solution is to add a new method to the store that adapters can call to mark a record in the created state as saved.

sandstrom:
@bmac Yes, you are absolutely right, that's the meat of it. But I think using a transient id, mirrored back by the server, should be pretty straightforward way of solving this.

```
{
  "type": "articles",
  "id": "1",
  "attributes": {
    "title": "Rails is Omakase"
  },
  "meta": {
[...]

PR #2635, opened 2014-12-28
[Bug] Fixed the test "create - relationships are not duplicated".

steffenbrem:
Fixed the test "create - relationships are not duplicated". Since you will not predefine the id of a comment created on the client side, it does not make very much sense.

It seems like a good idea to add this failing test so it can be fixed. I am not very deep involved in ember data so I did not manage to solve it myself.

For example in ticket #1829 they describe the same problem. Also in this discussion: http://discuss.emberjs.com/t/two-ember-objects-created-for-single-backend-object/3891.


bmac:
Hi @steffenbrem. I'm going to close this pr. I believe the existing test is correct. You may create a record with a uuid on the client side then save the record outside of Ember Data's normal save method. In that case the store needs to be able to correctly reconcile the ids when the server returns the same id form the backend.

Issue #2516, opened 2014-11-26
generateIdForRecord doesn't get "record" argument


psierks:
Docs indicate that generateIdForRecord should be supplied "store", and "record". Currently only a single argument ( the store ) is passed. The function calling / passing in the args is _generateId(...)

For reference:
http://emberjs.com/api/data/classes/DS.RESTAdapter.html#method_generateIdForRecord

PR #2637, opened 2014-12-29
generateIdForRecord gets type & object properties passed to it

psierks:
This is for #2516 

The documentation should also be updated to reflect the actual parameters being passed


bmac:
@psierks do you mind also updating the docs for [generateIdForRecord](http://emberjs.com/api/data/classes/DS.RESTAdapter.html#method_generateIdForRecord) to reflect that it gets passed an JavaScript object instead of a record instance.

PR #2641, opened 2014-12-30
Add deprecateProperty to ember-defeatureify.js

wecc:
Make sure it really works before merging


wecc:
Not using deprecateProperty

PR #2649, opened 2014-12-31
Only trigger changes for changed properties

fivetanley:
Previously, if your payload sent "firstName", but the value of "firstName" did not change, your observers would fire anyway. It seems like if we are going to do observers we should not cause observers to fire when they wouldn't if the underlying `_data` object were an `Ember.Object`. This also helps reduce zalgo when you are developing an app and wondering why observers are firing when they shouldn't be.

cc @rwjblue

### DIFF
--- a/packages/ember-data/lib/system/model/model.js
+++ b/packages/ember-data/lib/system/model/model.js
@@ -34,6 +34,33 @@ function extractPivotName(name) {
   );
 }
 
+// Like Ember.merge, but instead returns a list of keys
+// for values that fail a strict equality check
+// instead of the original object.
+function mergeAndReturnChangedKeys(original, updates) {
+  var changedKeys = [];
+
+  if (!updates || typeof updates !== 'object') {
+    return changedKeys;
+  }
+
+  var keys   = Ember.keys(updates);
+  var length = keys.length;
+  var i, val, key;
+
+  for (i = 0; i < length; i++) {
+    key = keys[i];
+    val = updates[key];
+
+    if (original[key] !== val) {
+      changedKeys.push(key);
+    }
+
+    original[key] = val;
+  }
+  return changedKeys;
+}
+
 /**
 
   The model class that all Ember Data records descend from.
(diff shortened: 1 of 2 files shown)


packages/ember-data/lib/system/model/model.js
```diff
@@ -34,6 +34,29 @@ function extractPivotName(name) {
   );
 }
 
+// Like Ember.merge, but instead returns a list of keys
+// for values that fail a strict equality check
+// instead of the original object.
+function mergeAndReturnChangedKeys(original, newInfo) {
+  var ret = [];
```
rwjblue: Tiny nit-pick: can you name this something meaningful (to make it clearer when I end up stepping the code while in *brain-fried* mode)?  Something like `changedProperties` would make this much clearer...

igorT:
I think @wecc was working on something similar, the idea was for attrs to internalize their state so we have less crazy code to write. Not sure what the status of that is

rwjblue:
It is very annoying to have to guard any CP's or observers that depend on model attributes by checking against (manually) cached inputs just because ED notifies that all properties have changed (even though they may not have).

:+1: from me...

brzpegasus:
@fivetanley I'm excited to have these changes pulled into Ember Data as it is an issue I got hit with recently as well. I am wondering, however, what the expected behavior would be if the attribute being checked for was a `hasMany` relationship. The strict equality check wouldn't work on arrays, so if a `post` record has many `comments`, and the record gets updated or reloaded, this will always flag the `comments` relationship as having changed, even if that wasn't the case.

wecc:
I haven't had the time to my PR for this yet, but yeah, we really do want this feature. My approach is a bit different and aims to simplify `_data`, `_attributes` and `_inFlightAttributes`, especially now when the approach is to merge instead of replace data.

Consider the following code:

```javascript
var post = store.push('post', { id: 1, title: 'Ember.js', author: 'Tomster' });

// this updates _attributes, not _data
post.set('author', 'Stanley');

// this updates _data['author'] with the new value, but we shouldn't
// trigger since it hasn't really changed in terms of post.get('author')
store.push('post', { id: 1, author: 'Stanley' });
```

wecc:
@fivetanley :+1:

PR #2655, opened 2015-01-02
build tool improvements

stefanpenner:
* cleanup
* windows via appveyor

- [x] confirm this works on CI

[fixes #2655]

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,15 +14,16 @@
     "lib/ember-addon"
   ],
   "scripts": {
-    "prepublish": "bower install && npm run-script dist",
+    "build": "ember build",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -14,15 +14,17 @@
     "lib/ember-addon"
   ],
   "scripts": {
+    "build": "ember build",
+    "build:production": "ember build --environment=production",
```
rwjblue: Duplicated as `dist` (we should only have one that does this)....

stefanpenner:
I'm going to merge this, as it cleans stuff up. The windows test (now that it is setup) fails regardless

PR #2658, opened 2015-01-04
[WIP] errors object with sources

tchak:
In this PR I am trying to address a few limitations of `errors` object.

- fix enumerable subclassing by using `Ember.Array` instead
- allow for several sources for errors
- this should also allow for per validator `remove` / `clear` (each validator can be a different source)

This should make it easier to integrate with third partie libraries like `ember-validations` by only triggering invalid state for internal `adapter` source. Proposed changes should also allow for per-validator errors. It should make it easier to implement live client validations, as requested by @bcardarella.

Finally, I know @bcardarella, you are not the biggest fan of how things gone in the past regarding `errors`. I really hope we can iron things out and make it work for core and third party libraries. I am looking forward to work with you on making `ember-validations` play nice with `ember-data`.

I will add some examples and use cases tomorrow. It is getting late around here and I had enough headache with this for tonight :) Any feedback is welcome, but if something is really unclear, hang tight for tomorrow, for more examples and tests.

cc @igorT

### DIFF
--- a/packages/ember-data/lib/system/model/errors.js
+++ b/packages/ember-data/lib/system/model/errors.js
@@ -1,10 +1,57 @@
 var get = Ember.get;
 var isEmpty = Ember.isEmpty;
 var map = Ember.EnumerableUtils.map;
+var makeArray = Ember.makeArray;
+var emberA = Ember.A;
+var EmberArray = Ember.Array;
+var Evented = Ember.Evented;
+var tryInvoke = Ember.tryInvoke;
+var reduceComputed = Ember.reduceComputed;
+var mapBy = Ember.computed.mapBy;
 
 import {
   MapWithDefault
 } from "ember-data/system/map";
+
+function rejectByAttribute(attribute) {
+  return function(error) {
+    return error.attribute === attribute;
+  };
+}
+
+function rejectBySource(source) {
+  return function(error) {
+    return error.source === source;
+  };
+}
+
+function rejectByAttributeAndSource(attribute, source) {
+  return function(error) {
+    return error.attribute === attribute && error.source === source;
+  };
+}
+
+function replaceWithContent(target, content, attributes) {
+  var amt = get(target, 'length'),
+      length = get(content, 'length');
+
+  if (amt !== length) {
+    target.arrayContentWillChange(0, amt, length);
+
+    get(target, 'content').setObjects(content);
+
+    for (var i = 0, lng = attributes.length; i < lng; i++) {
+      target.notifyPropertyChange(attributes[i]);
+    }
+
+    target.arrayContentDidChange(0, amt, length);
+  }
+}
+
+var ADAPTER_SOURCE = "ember-data:adapter";
+var DEFAULT_SOURCE = "ember-data:default";
+
+export { ADAPTER_SOURCE, DEFAULT_SOURCE };
 
 /**
 @module ember-data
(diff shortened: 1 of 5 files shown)


bcardarella:
To be honest I would prefer if all of this was completely removed from Ember Data. I don't think Ember Data should have any functionality handling the error state beyond a hook for other libraries to take advantage of. When libraries like ember-validations are brought in now the asset footprint is unnecessarily increased as ember-validations will discard most of what Ember Data does. If this was all extracted out to a library like an Ember Data Errors or something like that then the end-user could swap out as they desire.

tchak:
@bcardarella I am sorry, I completely disagree on this. Instead I want at some point to bring validators in to ember data. We been pulling this back and forth for 2 years now. At this point I would just love for some decision to be made. I am not sure what the process is. Should there be some kind of arbitrage from the core team? cc @igorT @stefanpenner @wycats 

Again, I am very open to have more discussions with third party library authors on how errors container should work. Can we find some compromise @bcardarella?

bcardarella:
Ok, I'll have to continue to work around this. I don't see there being much room for compromise.

bcardarella:
FWIW, adding validations to ember data is not a good idea. Validations should not be a concern for Ember Data as there are many scenarios in which validations that don't include data persistence.

ppcano:
I agree with @bcardarella

> I don't think Ember Data should have any functionality handling the error state beyond a hook for other libraries to take advantage of.

fivetanley:
@tchak I feel like before you continue putting a lot more effort into this, it needs to go through the [RFC process](https://github.com/emberjs/rfcs) first. We'd like the errors stuff to be a lot more collaborative. I will work with @tomdale and @wycats and try to get them to come comment here, but I know that tom, yehuda, and @igorT have interest in aligning with @bcardarella in this space.

I think @bcardarella may also have a point that error handling may just be a hook that other libraries can use.

For whatever we decide, let's try to flesh out these ideas together. As Ember Data is approaching 1.0, it is really important that we either decide the API or decide to remove it before the API solidifies.

tchak:
@fivetanley for me, Ember was always about "battery included". Error / Validations is a core concerne for data persistance. I would be very disappointed if the experience for `ember-data` will be: "install ember data, but before you can do something useful with it, you have to look for a third-party half assed extension".
I definitely agree with @bcardarella that they are cases where validations are not to be bound to persistance. The way I see it, we could have two separate modules `ember-errors-container` and `ember-validator`. Both should be usable on `models` or `components`. Validator package should contain only basic plumbing. Actual validators can be provided by third-party plugins.

I have not got much of support on this for the past two years. I made this PR because @igorT asked me to try to iron out concerns with `errors` object.

But if all I get is "I don't see there being much room for compromise." I will not put any more effort in this.

RFC is probably a good idea, but I am not much of a writer plus given feedback I got on all this so fare, right now my motivation is not at best. I am afraid it would take me another year to come up with something useful :) I prefer to not take this responsibility. Whatever you decide, if there going to be need for help with implementation I will gladly provide it.

bcardarella:
@tchak we just seem so far apart on this issue that I don't feel you and I will come to an understanding/compromise. It is probably best to have the core team decide on the direction rather than either of us.

tchak:
@bcardarella I agree. If it is decided that `hooks` it is, no matter how it saddens me, I am willing to provide a PR to remove `errors`.

bcardarella:
@tchak if that's the case I think we can work on a RFC to flesh out that API but let's wait for some official direction

igorT:
Closing this in favor of the new rfc

PR #2659, opened 2015-01-05
[BUGFIX] Fetching an empty record should run find instead of reload

### LINKED ISSUE #2569: Fetching an object that has been created by a reference previously throws error
An example: http://emberjs.jsbin.com/vukoli/1/edit?html,js

In this example, we load the books, which contains a reference to their author, and then fetch the author with `store.fetch`, which throws an error.

```
Error while processing route: index Attempted to handle event `reloadRecord` on <App.Author:ember306:2> while in state root.empty.
```

kimroen:
Fixes #2569

When you run `reload` on an empty record, such as one being created from a reference, you get an error about not being able to call `reloadRecord` on a record that is in state `root.empty`. This checks for that and calls `store.find` instead.

For more details on the issue, see #2569.

### DIFF
--- a/packages/ember-data/lib/system/store.js
+++ b/packages/ember-data/lib/system/store.js
@@ -771,7 +771,8 @@ Store = Ember.Object.extend({
   hasRecordForId: function(typeName, inputId) {
     var type = this.modelFor(typeName);
     var id = coerceId(inputId);
-    return !!this.typeMapFor(type).idToRecord[id];
+    var record = this.typeMapFor(type).idToRecord[id];
+    return !!record && !get(record, 'isEmpty');
   },
 
   /**
--- a/packages/ember-data/tests/integration/store_test.js
+++ b/packages/ember-data/tests/integration/store_test.js
@@ -246,3 +246,32 @@ test("Using store#fetch on non existing record calls find", function() {
     });
   });
 });
+
+test("Using store#fetch on an empty record calls find", function() {
+  expect(2);
+
+  ajaxResponse({
+    cars: [{
+      id: 20,
+      make: 'BMCW',
+      model: 'Mini'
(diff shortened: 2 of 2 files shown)


kimroen:
I'm not sure why all the tests are failing - everything was passing fine locally both before and after my changes.

fivetanley:
@kimroen they are failing on the canary build. You can use `http://localhost:4200/?emberchannel=canary` to load that build.

fivetanley:
@kimroen this looks unrelated to your build, I think the root cause is here. https://github.com/emberjs/data/issues/2661

kimroen:
@fivetanley This passes again now :)

Thoughts on the issue and fix?

sly7-7:
I wonder if we could instead add the `reloadRecord` event in the empty state. @igorT any thought ?

kimroen:
@sly7-7 That sounds like a reasonable solution too - I could go either way. Not sure what the status is on this.

bmac:
I think I'd also like to see this implemented by adding a `reloadRecord` event to the empty state.

tomdale:
@igorT believes that the correct fix is to change the `hasRecordForId` check to return `false` if we have a record but it's empty.

kimroen:
I implemented this now, but I'm not sure - `getById` also uses the check for `hasRecordForId`, and would now return `null` for empty records, where it previously returned the empty record.

kimroen:
That could be fine, as you can use `store#recordForId`, but it would be a change. I'll update this PR shortly.

kimroen:
Looks like there might be some problems with the canary-test, but this change seems fine.

bmac:
Closing. This commit was merged as part of pr #2875.

PR #2663, opened 2015-01-05
Checking superclass type for equality in addCanonicalRecord method for a belongsTo relationship

oshikryu:
When using polymorphic models in an ember app, the `addCanonicalRecord` method does not check for a model's superclass when comparing equality of a `newRecord` to the expected `type`

Because there previously was no check for a type's super class, this assertion pops up: `Ember.assert("You can only add a '" + type.typeKey + "' record to this relationship", newRecord instanceof type);`

My solution to this is to add a check for `Ember.MODEL_FACTORY_INJECTIONS` like in the `addRecord` method inside the `belongs_to` class:
```
Ember.assert("You can only add a '" + type.typeKey + "' record to this relationship", (function () {
    if (newRecord instanceof type) {
      return true;
    } else if (Ember.MODEL_FACTORY_INJECTIONS) {
      return newRecord instanceof type.superclass;
    }

    return false;
  })());
```

### DIFF
--- a/packages/ember-data/lib/system/relationships/state/belongs_to.js
+++ b/packages/ember-data/lib/system/relationships/state/belongs_to.js
@@ -35,8 +35,6 @@ BelongsToRelationship.prototype.setCanonicalRecord = function(newRecord) {
 BelongsToRelationship.prototype._super$addCanonicalRecord = Relationship.prototype.addCanonicalRecord;
 BelongsToRelationship.prototype.addCanonicalRecord = function(newRecord) {
   if (this.canonicalMembers.has(newRecord)){ return;}
-  var type = this.relationshipMeta.type;
-  Ember.assert("You can only add a '" + type.typeKey + "' record to this relationship", newRecord instanceof type);
 
   if (this.canonicalState) {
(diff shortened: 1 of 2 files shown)


igorT:
Thanks for the PR!
We actually don't even need this assert here, because changing canonical records means that we are loading data from the server, so there is no point in telling the user they did something wrong. Do you want to make a PR that just removes the assertion? Thanks

oshikryu:
Great suggestion. It makes sense that you don't need to tell the user that data from the server is bad at this point. That seems more like a server issue. I removed the assertion and all of the existing tests passed locally.

igorT:
Can you please squash as well?

oshikryu:
Squashed. The commit message may be misleading however

igorT:
Can you amend the commit message then? :)

oshikryu:
Should be good to go!

oshikryu:
Updated the commit message and the test associated with it

igorT:
Thanks!

tonywok:
Yay! Thanks. Was just about to submit this.

PR #2672, opened 2015-01-06
Polyfill Ember.Map and friends and use _values

### LINKED ISSUE #2661: ember canary build is broken due to more Ember.Map issues....
https://github.com/emberjs/ember.js/commit/685d9e36fee7a8d99f3811a9f99758a1231ffaa7

We'll need to add these to packages/ember-data/lib/system/map.js

kimroen:
Fixes #2661 

This works, but we have a failing test that I'm not sure what's all about.

### DIFF
--- a/packages/ember-data/lib/system/map.js
+++ b/packages/ember-data/lib/system/map.js
@@ -6,9 +6,172 @@
  across a few releases. As ES6 comes to a close we should have a smaller
  and smaller gap in implementations between Ember releases.
 */
-var Map            = Ember.Map;
-var MapWithDefault = Ember.MapWithDefault;
-var OrderedSet     = Ember.OrderedSet;
+var Map, MapWithDefault;
+var OrderedSet = Ember.OrderedSet;
+
+function copyNull(obj) {
+  var output = Ember.create(null);
+
+  for (var prop in obj) {
+    // hasOwnPropery is not needed because obj is Object.create(null);
+    output[prop] = obj[prop];
+  }
+
+  return output;
+}
+function copyMap(original, newObject) {
+  var keys = original._keys.copy();
+  var values = copyNull(original._values);
+
+  newObject._keys = keys;
+  newObject._values = values;
+  newObject.size = original.size;
+
+  return newObject;
+}
+
+var map = new Ember.Map();
+if (!Object.hasOwnProperty.call(map, '_keys')) {
+  Map = function ED_Map() {
+    if (this instanceof this.constructor) {
+      this._keys = OrderedSet.create();
+      this._keys._silenceRemoveDeprecation = true;
+      this._values = Ember.create(null);
+      this.size = 0;
+    } else {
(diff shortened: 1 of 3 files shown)


packages/ember-data/lib/system/map.js
```diff
@@ -6,9 +6,172 @@
...
+    this.size = 0;
+  };
+
+  Map.constructor = Map;
+
+  Map.create = function() {
+    return new this.constructor();
+  };
+
+  MapWithDefault = function(options) {
```
fivetanley: replace anonymous function with `function ED_MapWithDefault` so we can get a little better stack traces

packages/ember-data/lib/system/map.js
```diff
@@ -6,9 +6,172 @@
...
+  newObject._keys = keys;
+  newObject._values = values;
+  newObject.size = original.size;
+
+  return newObject;
+}
+
+var map = new Ember.Map();
+if (!Object.hasOwnProperty.call(map, '_keys')) {
+  Map = function() {
```
fivetanley: replace anonymous function with `function ED_Map` so we can get a little better stack traces

fivetanley:
@kimroen whatever failure you were having on canary is passing locally for me. Aside from a few comments, this looks great to me! Thank you so much!

Could you also uncomment the canary build in travis.yml?

kimroen:
@fivetanley All done!

packages/ember-data/lib/system/record_array_manager.js
```diff
@@ -261,7 +261,7 @@ export default Ember.Object.extend({
   willDestroy: function(){
     this._super();
 
-    forEach(flatten(values(this.filteredRecordArrays.values)), destroy);
+    forEach(flatten(values(this.filteredRecordArrays._values)), destroy);
```
stefanpenner: how about ember-data uses the public API for enumerating a map. `forEach`

kimroen: Fair question! That would probably render this polyfill unnecessary, correct?

I'm sure there's a reason why that's not used here, but maybe it could be rewritten to do so?

stefanpenner: i suspect it can be re-written (and should be).

At some point we will prefer a proper and full polyfil that supports the values() iterator

kimroen:
I'm not sure what the next step for this is. Some options:

1. Rewrite the line that was using the (now) private API and remove the polyfill
2. Merge this and make issues to implement `Ember.Map#values` in Ember, then use that when ready

packages/ember-data/lib/system/record_array_manager.js
bmac: re-writing this to use the map's `forEach` sounds like the best solution to me. @kimroen do you have time to update this pr to do that?

kimroen:
I submitted a new PR to rewrite the part that was using the private API: https://github.com/emberjs/data/pull/2691

PR #2675, opened 2015-01-07
Use store.push instead of store.update in test

wecc:
This removes the store.update deprecation warning when running tests


igorT:
Thanks!

PR #2676, opened 2015-01-07
Remove yuidoc warning

wecc:
Removes this warning from the build:

```
info: (yuidoc): YUIDoc Starting from: packages
warn: (docparser): Missing item type:  packages/ember-data/lib/adapters/rest_adapter.js:605
warn: (docparser):             http://stackoverflow.com/questions/417142/what-is-the-maximum-length-of-a-url-in-different-browsers
```


igorT:
Thanks!

PR #2677, opened 2015-01-07
Don't warn for polymorphic belongsTo type keys

### LINKED ISSUE #2599: Should not warn if the payload contains unknown keys that are relationship types?
When my payload has a `relationshipType` key, I'm receiving a warning that the key is unknown.

```WARNING: The payload for '(subclass of DS.Model)' contains these unknown keys: [membersType]. Make sure they've been defined in your model.```

See this [Stack Overflow question](http://stackoverflow.com/questions/27509059/where-when-should-i-remove-the-relationship-type-to-avoid-an-unknown-keys-warnin) for an example.

It seems to me that `...Type` keys should be excluded from this warning (or at least have the option). Is that possible now? Would a PR make sense for that?

### PR DESCRIPTION
**Pull request by wecc**
This is not pretty but it was the least ugly thing I could come up with. This also allows for regular attributes ending in "Type", like:

```javascript
App.Thing = DS.Model.extend({
  fakeType: DS.attr('string')
});
```

Fixes #2599

### DIFF
--- a/packages/ember-data/lib/system/store.js
+++ b/packages/ember-data/lib/system/store.js
@@ -1400,15 +1400,15 @@ Store = Ember.Object.extend({
     var type = this.modelFor(typeName);
     var filter = Ember.EnumerableUtils.filter;
 
-    // If the payload contains unused keys log a warning.
-    // Adding `Ember.ENV.DS_NO_WARN_ON_UNUSED_KEYS = true` will suppress the warning.
-    if (!Ember.ENV.DS_NO_WARN_ON_UNUSED_KEYS) {
+    // If Ember.ENV.DS_WARN_ON_UNKNOWN_KEYS is set to true and the payload
+    // contains unknown keys, log a warning.
+    if (Ember.ENV.DS_WARN_ON_UNKNOWN_KEYS) {
       Ember.warn("The payload for '" + type.typeKey + "' contains these unknown keys: " +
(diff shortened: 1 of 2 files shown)


sly7-7:
@wecc what do you think about extracting say, a `filterUnusedKeys` function and drying this thing

```
return filter(Ember.keys(data), function(key) {
  if (key.match(/Type$/)) {
    var relationship = get(type, 'relationshipsByName').get(key.replace(/Type$/, ''));
    if (relationship) {
      return relationship.kind !== 'belongsTo' && relationship.options.polymorphic !== true;
    }
  }
  return !get(type, 'fields').has(key) && key !== 'id' && key !== 'links';
})
```

Sounds possible ?

wecc:
@sly7-7 I would love to get away from the inline ugliness. The reason it's done like this is for defeatureify to be able to strip the whole `Ember.warn()` statement. I think we tried to do som `runInDebug()` voodoo in the beginning but defeatureify couldn't handle nested stuff.

sly7-7:
@wecc I should have thought about that before asking... So at this point this looks good, thanks a lot for your work.

@igorT ok with merging this + #2678 ?

wecc:
After discussing with @igorT we simplified this for better readability and less code complexity. Also switched the warning to default-off and renamed the flag to `DS_WARN_ON_UNKNOWN_KEYS`.

Rebased.

PR #2678, opened 2015-01-07
Test warnings cleanup

wecc:
Cleaned up some tests to remove all warnings when running the test suit. The only warnings left are two of these:

```
WARNING: The payload for 'comment' contains these unknown keys: [messageType]. Make sure they've been defined in your model.
```

..but they will disappear when/if #2677 is merged.


bmac:
Thanks @wecc!

PR #2679, opened 2015-01-07
Add `expect` calls when async is involved

seanpdoyle:
`expect` calls within callbacks ensure that all expectations are invoked
and passing.

### DIFF
--- a/packages/ember-data/tests/unit/model/lifecycle_callbacks_test.js
+++ b/packages/ember-data/tests/unit/model/lifecycle_callbacks_test.js
@@ -4,6 +4,8 @@ var run = Ember.run;
 module("unit/model/lifecycle_callbacks - Lifecycle Callbacks");
 
 test("a record receives a didLoad callback when it has finished loading", function() {
+  expect(3);
+
   var Person = DS.Model.extend({
     name: DS.attr(),
     didLoad: function() {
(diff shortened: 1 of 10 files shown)


packages/ember-data/tests/unit/model/errors_test.js
```diff
@@ -30,7 +30,6 @@ function unexpectedSend(eventName) {
 }
 
 test("add error", function() {
-  expect(6);
```
igorT: Is there a downside to having the expect here?

igorT:
Awesome work!

igorT:
I think we should keep the expects though, so if the test gets refactored they are kept around

packages/ember-data/tests/unit/model/errors_test.js
seanpdoyle: Synchronous expects make the test

* noisy - if an expectation fails, the subsequent failures won't be run anyway
* brittle - adding/removing an additional expectation requires both adding/removing the expectation, as well as incrementing/decrementing `expect(n)`

igorT: That is true, but I am more worried about people using tests as an example and then not adding expects, and then believing their tests pass by mistake. Or someone refactoring the test to be async. Not sure it's a reasonable fear though

seanpdoyle:
@igorT fixed :+1:

igorT:
Thanks!

PR #2682, opened 2015-01-07
Run JSHint on Test Suite

seanpdoyle:
Test suite should comply with JSHint standards and conventions

### DIFF
--- a/packages/ember-data/tests/unit/transform/number_test.js
+++ b/packages/ember-data/tests/unit/transform/number_test.js
@@ -1,3 +1,5 @@
+/* jshint -W053 */
+
 module("unit/transform - DS.NumberTransform");
 
 test("#serialize", function() {
--- a/Brocfile.js
+++ b/Brocfile.js
@@ -41,14 +41,14 @@ function testTree(packageName){
     files: [ '**/*.js' ],
     destDir: '/' + packageName
   });
-  var jshinted = jshint('packages/' + packageName + '/lib', {
+  var jshinted = jshint('packages/' + packageName + '/', {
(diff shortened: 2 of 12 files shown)


packages/ember-data/tests/unit/transform/number_test.js
```diff
@@ -20,7 +20,7 @@ test("#deserialize", function() {
   equal(transform.deserialize(undefined),       null);
   equal(transform.deserialize("1.1"),           1.1);
   equal(transform.deserialize(1.1),             1.1);
-  equal(transform.deserialize(new Number(1.1)), 1.1);
+  equal(transform.deserialize(Number(1.1)), 1.1);
```
fivetanley: These ones we want to stay the same as they are testing Number objects (not primitives) are being deserialized correctly.

seanpdoyle: This change was suggested by JSHint.

http://stackoverflow.com/questions/369220/why-should-you-not-use-number-as-a-constructor

Should I turn off the rule globally?

igorT: It's a good global rule, but I think that here we are specifically testing for people who passed in a number object.

packages/ember-data/tests/unit/transform/number_test.js
```diff
@@ -1,3 +1,5 @@
+/* jshint -W053 */
+
```
seanpdoyle: @fivetanley @igorT :+1:

bmac:
Thanks @seanpdoyle!

PR #2684, opened 2015-01-08
Lint Code with JSCS

seanpdoyle:
* For now, copy `.jscsrc` from `emberjs/ember.js`
* Integrate into `npm start`

### DIFF
--- a/package.json
+++ b/package.json
@@ -57,6 +57,6 @@
     "yuidocjs": "~0.3.46"
   },
   "dependencies": {
-    "broccoli-jscs": "0.0.13"
+    "broccoli-jscs": "0.0.12"
   }
 }
--- a/package.json
+++ b/package.json
@@ -18,7 +18,7 @@
     "build:production": "ember build --environment=production",
     "prepublish": "bower install && npm run-script build:production",
     "start": "ember serve",
-    "test": "testem -R dot ci",
+    "test": "node_modules/broccoli-jscs/node_modules/jscs/bin/jscs packages && testem -R dot ci",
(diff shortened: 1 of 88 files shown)


seanpdoyle:
For now, the violations are only displayed in the output from `npm start`.

Violations will **NOT** break the build.

Any ideas on how to integrate `jscs` into `npm test`?

rwjblue:
One way would be to change https://github.com/emberjs/data/blob/master/package.json#L21 to be `jscs packages && testem -R dot ci` or something.

package.json
```diff
@@ -18,7 +18,7 @@
     "build:production": "ember build --environment=production",
     "prepublish": "bower install && npm run-script build:production",
     "start": "ember serve",
-    "test": "testem -R dot ci",
+    "test": "node_modules/broccoli-jscs/node_modules/jscs/bin/jscs packages && testem -R dot ci",
```
seanpdoyle: @rwjblue this seems a little extreme, but `jscs` isn't a direct dependency.

Is this better or worse than adding `jscs` to the dependencies and executing it directly?

Should I add `npm install -g jscs` to `.travis.yml`?

seanpdoyle:
@igorT @fivetanley sorry to bug you, but since this PR touches almost every file in the codebase, it's somewhat blocking to future refactorings.

If I could get your :eyes: on it ASAP, I could keep cleaning up the test suite.

seanpdoyle:
Not sure how to go about making the windows CI green

stefanpenner:
@seanpdoyle it doesn't work quite yet (we could use the help if u have time)

this looks great!

PR #2685, opened 2015-01-08
Implement ES6-compliant Date string parsing

jugglinmike:
As noted in the changeset, the ECMAScript 6 specification interprets
timezone-less Date strings according to the local system's time zone.
This is in direct opposition to ECMAScript 5 specification and the
currently-implemented behavior. From section 15.9.1.15 of the ES5
specification:

> the time zone offset specified as “Z” (for UTC) [...] The value of an
> absent time zone offset is “Z”.

Update Ember Data's Date string parsing logic to comply with the latest
specification.

### DIFF
--- a/packages/ember-data/lib/ext/date.js
+++ b/packages/ember-data/lib/ext/date.js
@@ -39,11 +39,17 @@ Ember.Date.parse = function (date) {
         struct[2] = (+struct[2] || 1) - 1;
         struct[3] = +struct[3] || 1;
 
-        if (struct[8] !== 'Z' && struct[9] !== undefined) {
-            minutesOffset = struct[10] * 60 + struct[11];
+        if (struct[8] !== 'Z') {
+            if (struct[9] !== undefined) {
+                minutesOffset = struct[10] * 60 + struct[11];
 
-            if (struct[9] === '+') {
-                minutesOffset = 0 - minutesOffset;
+                if (struct[9] === '+') {
+                    minutesOffset = 0 - minutesOffset;
+                }
+            // ES6 §20.3.1.15 states "If the time zone offset is absent, the date-time is interpreted as a
+            // local time."
+            } else {
+                minutesOffset = new Date().getTimezoneOffset();
             }
         }
 
--- a/packages/ember-data/tests/unit/transform/date_test.js
+++ b/packages/ember-data/tests/unit/transform/date_test.js
@@ -1,6 +1,8 @@
 module("unit/transform - DS.DateTransform");
 
 var dateString = "2015-01-01T00:00:00.000Z";
+var dateStringNoTz = "2015-01-01";
+var localNewYear = new Date(2015, 0, 1).toISOString();
 var dateInMillis = Ember.Date.parse(dateString);
 var date = new Date(dateInMillis);
 
@@ -18,6 +20,8 @@ test("#deserialize", function() {
 
   // from String
   equal(transform.deserialize(dateString).toISOString(),        dateString);
+
+  equal(transform.deserialize(dateStringNoTz).toISOString(),    localNewYear);
 
   // from Number
   equal(transform.deserialize(dateInMillis).valueOf(),          dateInMillis);


bmac:
@igorT and I review this today and we were afraid of the potential for this to break existing apps in subtle ways. Should we be including this logic in Ember Data or is there another library that we could pull in to handle this?

cc @tomdale @wycats

jugglinmike:
Yeah, this is definitely a breaking change, and I don't think it should be merged lightly. I also know that we're discussing a feature in a developing standard, so it might not be wise for Ember Data to commit to it just yet. That said, I think it's a pretty sure thing because [the behavior specified in ECMAScript 5 has been labeled as a bug](https://bugs.ecmascript.org/show_bug.cgi?id=112).

My main goal in submitting this pull request was to raise awareness on this upcoming change. I know Ember Data is approaching 1.0, and this seems like the kind of breaking change you might rather make here and now versus waiting for 2.0 (and dealing with changing expectations as ES6 conformance increases across platforms in the interim).

tomdale:
The problem here is that the server is storing (or at least emitting) dates without a timezone. What I'd like to do is start emitting a warning for any TZ-less dates, with a specific opt-in to the highly ambiguous behavior. For example, an option on the date transform that says something like:

```js
DS.attr('date', {
  defaultTimezone: 'local'
});
```

There may be cases where you _do_ in fact want to represent TZ-less datetimes over the network, but let's at the very least make people be explicit about it. Too many potential footguns otherwise.

jugglinmike:
There are two issues at play here: the design of the specification and the importance of Ember Data's compliance.

I'm not a specification author, so I can only guess at the rationale. I think this behavior makes sense for the case where *only* date information is specified, i.e. "2015-02-06". In those cases, the lack of precision is intentional because the data is describing a date irrespective of local time zones. It's the difference between saying "The first National Yogurt Day was February 6, 2015" and "let's get free yogurt on February 6, 2015 at 21:43:32 EST".

But if that's not convincing, we should follow up with an e-mail to ES-Discuss. If the specification is finalized with that behavior, I think it will be confusing when Ember Data's abstraction of the Date object to behaves differently than the Date object itself.

jelhan:
I'm not sure what's the status here. ECMAScript 6 is now final and states that ISO 8601 strings without time zone should be interpreted as local time:
> [20.3.1.16 Date Time String Format](http://www.ecma-international.org/ecma-262/6.0/#sec-date-time-string-format)
> (...)
> If the time zone offset is absent, the date-time is interpreted as a local time.

As @jugglinmike mentioned parsing ISO 8601 date-only strings ("2015-08-02") with `Ember.Date.parse()` results in unexpected behaviour:
```js
new Date(Ember.Date.parse('2015-01-01')).toLocaleDateString();
// "31.12.2014"

new Date().getTimezoneOffset();
// 360
```

Handling time zones while dealing with a date-only format does not make any sense.

Perhaps implementation of `Ember.Date.parse()` and `date` attribute should be discussed separately? `DS.DateTransform` serialize does not support a date-only format and therefore attribute type `date` does not make much sense for date-only data. If you deal with such data, you should implement something like `DS.DateOnlyTransform`.

So to put it together:
* `Ember.Date.parse()` should implement ECMAScript 6 as soon as possible.
* `DS.DateTransform` should also change it's default behaviour in the long term. In the mean while there should be a deprecation warning when parsing ISO 8601 strings without time zone by `DS.DateTransform`.
* There should be a flag to enable `DS.DateTransform` ECMAScript 6 compatible implementation right now.

If not changing `Ember.Date.parse()` this behaviour should be well documented.

bmac:
Hi @jugglinmike. Thanks for doing this work. 

Unfortunately, I'm going to close this pr. I think now that Ember Data does not support IE8 the plan is to [remove the ISO 8601 shim](https://github.com/emberjs/data/pull/3842) for date.parse. Please feel free to re-open if you disagree.

jugglinmike:
Consistency was my goal, so that's a fine resolution by me.

It may be worthwhile to determine if any browsers have already implemented the new date parsing logic, since that would make shim removal a breaking change.

jelhan:
Tested some browsers: Firefox (41.0.1, Ubuntu), Chromium (45.0.2454.101, Ubuntu), Microsoft Edge (20.10240, Windows 10) and Internet Explorer 11 (Windows 10) does not implement ECMAScript 2015 date parse yet. But there will be some browsers implementing it for sure and then constituency is gone.

runspired:
Just an update, the spec was reverted and timezone-less dates remained UTC to avoid breaking backwards compatibility.

https://github.com/tc39/ecma262/pull/138

jugglinmike:
I missed that one; thanks for the update!

PR #2686, opened 2015-01-08
Remove unnecessary `toString` overrides

seanpdoyle:
Only re-define `toString` when the test depends on it. Otherwise, 
re-definitions add to test noise.


fivetanley:
Well, they are also very useful for debugging when the container is not creating the toString for you.

seanpdoyle:
@fivetanley that's true. If that's the case, I'd rather add the definition to a helper method and use that, or define them when necessary and remove them after the issue is debugged, but before they make it into version control.

igorT:
@seanpdoyle only defining them when necessary kinda defeats the point, you can always spend time on a particular test to figure out what the model class is, but it's very helpful to have class names around when you are refactoring and bunch of tests break for quick inspection.

seanpdoyle:
closing thanks to @igorT's comments

PR #2693, opened 2015-01-12
Sort query params in ajax calls.

cibernox:
When developing APIs, is pretty common to place a high level cachingmechanism, like Varnish, to cache requests to public API endpoints, and those tools use the URL string to determine if they have a cached
response.

As an example, if one user requests `/posts?sort=price&category=pets` and another requests `/posts?category=pets&sort=price`, the cached request won't be used in the second call.

Varnish by example has some modules to sort query parameters when caching, but it's not the default behavior and requires custom config.

Although the URL specification does not say that the order of the query parameters is meaningless, in the bast mayority (read: all) of the APIs I've seen, 2 requests are considered identical regardless of the the order.

Because of that, sorting the query params consistently is considered a good practice, so this PR introduces a `sortQueryParams` key in the RESTAdapter class, that defaults to a function to sort query params alphabetically. I decided to make this the default. I doubt this will be a breaking change for anyone.

In case you don't want this behavior, setting that property to a falsey value in the ApplicationAdapter will respect the original order.

```js
export default DS.RESTAdapter.extend({
  namespace: 'api/v1',
  sortQueryParams: false
});
```

In case you want a different order, you can pass a function that does the sorting your way.

```js
export default DS.RESTAdapter.extend({
  namespace: 'api/v1',
  sortQueryParams: function(params) {
    var sortedKeys = Object.keys(params).sort().reverse();
    var len = keys.length, newParams = {};

    for (var i = 0; i < len; i++) {
      newParams[sortedKeys[i]] = params[sortedKeys[i]];
    }
    return newParams;
  }
});
```
As other customization in ED, this can be customized on a per resource level, in case the order matters in only a few endpoints.

EDIT:

As @bakura10 mentions, this also has the side effect of improving the number of requests when working with CORS, since preflight OPTIONS requests are only valid for a given URL, so sending several requests with the query parameters in different order will require additional requests.

### DIFF
--- a/packages/ember-data/lib/adapters/rest_adapter.js
+++ b/packages/ember-data/lib/adapters/rest_adapter.js
@@ -171,6 +171,60 @@ export default Adapter.extend({
   defaultSerializer: '-rest',
 
   /**
+    By default, the RESTAdapter will send the query params sorted alphabetically to the
+    server.
+
+    For example:
+
+    ```js
+      store.find('posts', {sort: 'price', category: 'pets'});
+    ```
+
+    will generate a requests like this `/posts?category=pets&sort=price`, even if the
+    parameters were specified in a different order.
+
+    That way the generated URL will be deterministic and that simplifies caching mechanisms
+    in the backend.
+
+    Setting `sortQueryParams` to a falsey value will respect the original order.
+
+    In case you want to sort the query parameters with a different criteria, set
+    `sortQueryParams` to your custom sort function.
+
+    ```js
+    export default DS.RESTAdapter.extend({
+      sortQueryParams: function(params) {
+        var sortedKeys = Object.keys(params).sort().reverse();
+        var len = sortedKeys.length, newParams = {};
+
+        for (var i = 0; i < len; i++) {
+          newParams[sortedKeys[i]] = params[sortedKeys[i]];
+        }
+        return newParams;
+      }
+    });
+    ```
+
+    @method sortQueryParams
+    @param {Object} obj
+    @return {Object}
+  */
+  sortQueryParams: function(obj) {
+    var keys = Ember.keys(obj);
+    var len = keys.length;
+    if (len < 2) {
+      return obj;
+    }
+    var newQueryParams = {};
+    var sortedKeys = keys.sort();
+
+    for (var i = 0; i < len; i++) {
+      newQueryParams[sortedKeys[i]] = obj[sortedKeys[i]];
+    }
+    return newQueryParams;
+  },
+
+  /**
     By default the RESTAdapter will send each find request coming from a `store.find`
     or from accessing a relationship separately to the server. If your server supports passing
     ids as a query string, you can set coalesceFindRequests to true to coalesce all find requests
@@ -329,6 +383,9 @@ export default Adapter.extend({
     @return {Promise} promise
   */
   findQuery: function(store, type, query) {
+    if (this.sortQueryParams) {
+      query = this.sortQueryParams(query);
+    }
     return this.ajax(this.buildURL(type.typeKey), 'GET', { data: query });
   },
 
--- a/packages/ember-data/tests/integration/adapter/rest_adapter_test.js
+++ b/packages/ember-data/tests/integration/adapter/rest_adapter_test.js
@@ -832,6 +832,67 @@ test("metadata is accessible", function() {
   }));
 });
 
+test("findQuery - if `sortQueryParams` option is not provided, query params are sorted alphabetically", function() {
+  adapter.ajax = function(url, verb, hash) {
+    passedUrl = url;
+    passedVerb = verb;
+    passedHash = hash;
+
+    deepEqual(Object.keys(hash.data), ["in", "order", "params", "wrong"], 'query params are received in alphabetical order');
+
+    return run(Ember.RSVP, 'resolve', { posts: [{ id: 1, name: "Rails is very expensive sushi" }] });
+  };
+
+  store.findQuery('post', { "params": 1, "in": 2, "wrong": 3, "order": 4 }).then(async(function() {
+    // Noop
+  }));
+});
+
+test("findQuery - if `sortQueryParams` is falsey, query params are not sorted at all", function() {
+  adapter.ajax = function(url, verb, hash) {
+    passedUrl = url;
+    passedVerb = verb;
+    passedHash = hash;
+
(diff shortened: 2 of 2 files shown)


cibernox:
This has been waiting for appveyor 14 hours. I'll push force to see what happens.

bmac:
@cibernox I don't think appveyor is configured correctly at the moment. It seems to be failing on most prs.

@fivetanley Could you include this pr in your list of Ember Data Meeting PR Discussion items?

bakura10:
I like this idea. I'd like to not that it's also useful when using CORS, because browser sends an OPTIONS request per unique URL, making sure that query params are ordered the same make sure that options request is properly cached in the browser and not send again.

cibernox:
@bakura10 I wasn't aware of that quirk with CORS, I naively assumed that it was path-based. 
Even better then.

bakura10:
I'm not 100% sure for order, but at least CORS sends two OPTIONS requests for "api.example.com/users?name=A" and "api.example.com/users?name=B" (yeah yeah....). This spec is so strange I assume it will sends two requests based on order too ;).

cibernox:
If that is the case, it probably will. When it comes to caching this thing is very important for me. Combinatory is a bitch, and 6 query params can be sorted in 6! ways. That is 720 different urls.

cibernox:
Was this discussed on the meeting? Any feedback to address

packages/ember-data/lib/adapters/rest_adapter.js
```diff
@@ -171,6 +171,58 @@ export default Adapter.extend({
   defaultSerializer: '-rest',
 
   /**
+    By default, the RESTAdapter will send the query params sorted alphabetically to the
+    server.
+
+    For example:
+    ```js
+      store.find('posts', {sort: 'price', category: 'pets'});
+    ```
```
bmac: Do you mind adding newlines above and below this js example? The markdown parser doesn't seem to want to do code hi-lighting when the code block is so close to other text.

screen shot 2015-01-25 at 11 36 35 pm

packages/ember-data/lib/adapters/rest_adapter.js
```diff
@@ -171,6 +171,58 @@ export default Adapter.extend({
...
+      }
+    });
+    ```
+
+    @method sortQueryParams
+    @param {Object} obj
+    @return {Object}
+  */
+  sortQueryParams: function(obj) {
+    var keys = Object.keys(obj);
```
bmac: Lets use `Ember.keys` here to support older browsers that don't have `Object.keys`.

packages/ember-data/lib/adapters/rest_adapter.js
```diff
@@ -171,6 +171,58 @@ export default Adapter.extend({
...
+          newParams[sortedKeys[i]] = params[sortedKeys[i]];
+        }
+        return newParams;
+      }
+    });
+    ```
+
+    @method sortQueryParams
+    @param {Object} obj
+    @return {Object}
```
bmac: I don't think the order of iteration over keys in an objects is guaranteed to always be the order that the keys were assigned to the object. http://www.ecma-international.org/ecma-262/5.1/#sec-12.6.4 Although in practice it sounds like all browsers follow that pattern so maybe this isn't an issue. https://stackoverflow.com/questions/280713/elements-order-in-a-for-in-loop

bmac:
@cibernox it sounds like the Ember Data meeting hasn't happened for the past 2 weeks due to scheduling reasons.  I would defiantly like to see this get merged into Ember Data.

packages/ember-data/lib/adapters/rest_adapter.js
cibernox: I'm aware of that, but in practice the insertion order is always respected.

cibernox:
Rebased & feedback addressed. 

BTW, I thought that I was able to use `Object.keys` because there is an Object polyfill in Ember/ED. That is not the case?

packages/ember-data/lib/adapters/rest_adapter.js
```diff
@@ -171,6 +171,60 @@ export default Adapter.extend({
...
+  */
+  sortQueryParams: function(obj) {
+    var keys = Ember.keys(obj);
+    var len = keys.length;
+    if (len < 2) {
+      return obj;
+    }
+    var i, newQueryParams = {};
```
bmac: Looks like after the rebase this line is now failing the new jscs rules.

```
Multiple assigning variable declarations at packages/ember-data/lib/adapters/rest_adapter.js :
   216 |      return obj;
   217 |    }
   218 |    var i, newQueryParams = {};
------------^
   219 |    var sortedKeys = keys.sort();
   220 |
```

cibernox: Should be green now

cibernox:
Appveyor fails in the npm-install.

tomdale:
> I doubt this will be a breaking change for anyone.

We have some concerns about the case of hashes and arrays. For example, in Rails and PHP, you represent arrays in the query string like this:

`foo[]=a&foo[]=b`

If these were to be rearranged, any change would break the ordering of the array. In JavaScript, enumerating the non-numerical, own keys of an object is deterministically ordered in practice. Can you help us understand the cases where you are passing the same query params but they are being sent in different orders?

cibernox:
@tomdale The idea behind this is not to prevent the same query parameters to be sent in different orders because the JS engine swaps them. That is empirically impossible, and if it were possible this implementation would be flawed anyway. 

The idea is to make sure that no matter in which order you add the keys the the params object, they are sent consistently, even across different users.

Example: Imagine search form where you can check/uncheck filter options, and a query string is being generated and fired after each change. Each user will select filters in a different order, populating the object in a different order and generating a different query string that other user making the same search.
Just an example, there is plenty of reasons because the object passed to `find` can be populated differently, generating different urls, maybe leading to cache misses in my Varnish, per example.

This just ensures that it does not matter at all.

Regarding to the arrays serialization, I don't see the relation. I am sorting the object passed to `store.find/findQuery`, and only in the root level. 

Both `store.find('tshirt', { categories: [3, 1], size: 'M'})` and `store.find('tshirts', { size: 'M', categories: [3, 1] })` will generate the same url, being "categories" before "size" in the query string, but by no means will try to reorder the array.

tomdale:
Sounds good to me.

Issue #2662, opened 2015-01-05
hasMany aync relationship does not load, throws exception


aortbals:
This regression appears in Ember Data 1.0.0-beta.14. It seems that `flushCanonical` is triggering an exception in Ember when it tries to [destroy child views that don't exist](https://github.com/emberjs/ember.js/blob/master/packages/ember-views/lib/views/collection_view.js#L326)?

Regression in Ember Data 1.0.0.-beta.14: http://jsbin.com/pagotu/1/edit?html,js,console,output
Works correctly in Ember Data 1.0.0.-beta.12: http://jsbin.com/vobevevapo/3/edit?html,js,console,output
Also appears in Ember Canary (actually prints the attribute for the hasMany record, still throws exception): http://jsbin.com/pagotu/3/edit?html,js,console,output
It also gets thrown on a basic `hasMany` <-> `belongsTo` relationship with `links` and `async: true`: http://jsbin.com/pagotu/5/edit?html,js,console,output

```
"TypeError: Cannot read property 'destroy' of undefined
    at ContainerView.extend.arrayWillChange (http://builds.emberjs.com/release/ember.js:40513:20)
    at applyStr (http://builds.emberjs.com/release/ember.js:19678:29)
    at sendEvent (http://builds.emberjs.com/release/ember.js:14114:13)
    at __exports__.default.Mixin.create.arrayContentWillChange (http://builds.emberjs.com/release/ember.js:30442:9)
    at EmberObject.extend.arrangedContentArrayWillChange (http://builds.emberjs.com/release/ember.js:34309:14)
    at applyStr (http://builds.emberjs.com/release/ember.js:19678:29)
    at sendEvent (http://builds.emberjs.com/release/ember.js:14114:13)
    at __exports__.default.Mixin.create.arrayContentWillChange (http://builds.emberjs.com/release/ember.js:30442:9)
    at Ember.Object.extend.flushCanonical (http://builds.emberjs.com/canary/ember-data.js:4784:14)
    at ember$data$lib$system$relationships$state$has_many$$ManyRelationship.flushCanonical (http://builds.emberjs.com/canary/ember-data.js:6508:22)"
```

igorT:
Thanks for the bug report, it appears to be a real bug. Investigating

MarSik:
[...]

PR #2698, opened 2015-01-15
Fix: `flushCanonical` could be executed when the manyArray is destroyed

ppcano:
`flushCanonical` could have been scheduled and executed when the manyArray has been destroyed, therefore it will raise an exception. 
When the `flushCanonical` is not called on a destroyed `ManyArray`, the exception is not thrown.

I am not sure but this may be related to #2662 which is likely executing a scheduled `flushCanonical`. 
@aortbals could try to check if this different changes fixes its issue:

```js
  if (!this.manyArray.get('isDestroyed')) {
    this.manyArray.flushCanonical();
    this._super$flushCanonical();
  }
```

### DIFF
--- a/packages/ember-data/lib/system/relationships/state/has_many.js
+++ b/packages/ember-data/lib/system/relationships/state/has_many.js
@@ -65,7 +65,9 @@ ManyRelationship.prototype.removeCanonicalRecordFromOwn = function(record, idx) 
 
 ManyRelationship.prototype._super$flushCanonical = Relationship.prototype.flushCanonical;
 ManyRelationship.prototype.flushCanonical = function() {
-  this.manyArray.flushCanonical();
+  if (!this.manyArray.get('isDestroyed')) {
+    this.manyArray.flushCanonical();
+  }
   this._super$flushCanonical();
 };
 
--- a/packages/ember-data/tests/integration/relationships/has_many_test.js
+++ b/packages/ember-data/tests/integration/relationships/has_many_test.js
@@ -1234,3 +1234,42 @@ test('unloading a record with associated records does not prevent the store from
     ok(false, "store prevented from being destroyed");
   }
 });
+
+
+test('ManyRelationship can execute `flushCanonical` when the manyArray has been destroyed', function() {
+
+  expect(1);
+  var post;
+
+  Comment.reopen({
+    post: DS.belongsTo('post', { async: true })
+  });
+
+  Post.reopen({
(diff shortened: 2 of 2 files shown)


bmac:
This looks good to me :+1:

bmac:
@ppcano Do you mind rebasing this pr against the latest master?

fivetanley:
@igorT r?

ppcano:
@bmac, rebased.

packages/ember-data/lib/system/relationships/state/has_many.js
```diff
@@ -65,7 +65,9 @@ ManyRelationship.prototype.removeCanonicalRecordFromOwn = function(record, idx)
 
 ManyRelationship.prototype._super$flushCanonical = Relationship.prototype.flushCanonical;
 ManyRelationship.prototype.flushCanonical = function() {
-  this.manyArray.flushCanonical();
+  if (!this.manyArray.get('isDestroyed')) {
```
stefanpenner: get isn't needed here

packages/ember-data/lib/system/relationships/state/has_many.js
```diff
@@ -65,7 +65,9 @@ ManyRelationship.prototype.removeCanonicalRecordFromOwn = function(record, idx)
...
+    this.manyArray.flushCanonical();
+  }
   this._super$flushCanonical();
```
stefanpenner: super super also guard?

aortbals:
I wanted to leave a comment here that I tested this commit (and am also looking for an alternative solution), and it does not solve the original https://github.com/emberjs/data/issues/2662 bug for me. The `manyArray` is not marked as destroyed at this point.

bmac:
Hi @ppcano. @igorT and I reviewed this code and we suspect there is a deeper bug causing the error you are seeing. If a relationship is destroyed nothing should be calling `flushCanonical` and if something is then we need to tack down the root of that call instead of adding this check here. Can you put together a jsbin to help diagnose the root of the issue? I have my suspicions that it is related to the host of bugs that #2723 will fix.

ppcano:
@bmac, right now, I cannot reproduce this case, because the exception appeared when using a removed feature, and at that time, I did not create a `jsbin` or `integration test`.

Anyway, I have upgraded to `1.0.0-beta.15` and everything seems to work correctly.

I am closing this issue and will reopen again if the case come up again adding a `jsbin` with this case.

Many thanks

PR #2700, opened 2015-01-15
Document the signature of the callback that is passed into eachRelations...

### LINKED ISSUE #2616: Relationship descriptor does not seem to be documented (and it's parentType property isn't obvious)
The contents of the relationship descriptor passed to [DS.Model's eachRelationship callback](http://emberjs.com/api/data/classes/DS.Model.html#method_eachRelationship) (and exposed elsewhere in the public API) does not seem to be documented.

It appears to contain:

- `key`: (string) the name of this relationship on the Model
- `kind`: (string) "hasMany" or "belongsTo"
- `options`: (object) the original options hash passed when the relationship was declared
- `parentType`: (DS.Model) the type of the Model
- `type`: (DS.Model) the type of the related Model

I must be misunderstanding something, though, because it seems `parentType` isn't fully usable as I'd expect. When defining a [`serializeHasMany` method](http://emberjs.com/api/data/classes/DS.RESTSerializer.html#method_serializeHasMany) on a custom serializer, I'm not able to use the `parentType` to get any info about relationships:

    Ember.get(relationship.parentType, 'relationships')
    TypeError: Cannot call method 'modelFor' of undefined

This appears to be because `store` isn't properly defined. Possibly addressed by https://github.com/emberjs/data/issues/2528 ?
I'm avoiding this by using `record.constructor` instead

    Ember.get(record.constructor, 'relationships')
    MapWithDefault

bmac:
...hip

Closes #2616

screen shot 2015-01-15 at 10 40 21 am

### DIFF
--- a/packages/ember-data/lib/system/relationships/ext.js
+++ b/packages/ember-data/lib/system/relationships/ext.js
@@ -560,6 +560,47 @@ Model.reopen({
     invoking the callback with the name of each relationship and its relationship
     descriptor.
 
+
+    The callback method you provide should have the following signature (all
+    parameters are optional):
+
+    ```javascript
+    function(name, descriptor);
+    ```
+
+    - `name` the name of the current property in the iteration
+    - `descriptor` the meta object that describes this relationship
+
+    The relationship descriptor argument is an object with the following properties.
+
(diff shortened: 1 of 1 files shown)


wecc:
:+1:

PR #2701, opened 2015-01-15
Favor `object` over `hash` in public documentation and assertions

### LINKED ISSUE #1892: [feature request?] `hash` nomenclature & usage of JS Objects has hashes / associative arrays
I've noticed in several places in Ember Data that assertions are made, talking about Hashes.

One example:
https://github.com/emberjs/data/blob/ab589b32dbcb97640379b07be9d77388dc9d7cb8/packages/ember-data/lib/system/store.js#L1189

However, these aren't hashes, they're just regular javascript objects. This has mentally tripped me up a couple of times.

Should we either advocate the use of an actual hash instead?:
http://rick.measham.id.au/javascript/hash.htm

Or modify the assertions and references such that they refere to javascript objects rather than hashes? Calling them hashes feels very Ruby-ish to me and may make people think they're not just javascript objects.

One of the things that's bugging about using standard JS Objects is that you can't just say obj.keys or obj.values... maybe if Ember & Ember data made pervasive use of the aforementioned library when actually talking about and using hashes, it might make things easier?

bmac:
Closes #1892

### DIFF
--- a/packages/ember-data/lib/system/model/model.js
+++ b/packages/ember-data/lib/system/model/model.js
@@ -918,7 +918,7 @@ var Model = Ember.Object.extend(Ember.Evented, {
   },
 
   materializeAttributes: function(attributes) {
-    Ember.assert("Must pass a hash of attributes to materializeAttributes", !!attributes);
+    Ember.assert("Must pass an object to materializeAttributes", !!attributes);
     merge(this._data, attributes);
   },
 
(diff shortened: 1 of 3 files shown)


packages/ember-data/lib/system/model/model.js
```diff
@@ -918,7 +918,7 @@ var Model = Ember.Object.extend(Ember.Evented, {
   },
 
   materializeAttributes: function(attributes) {
-    Ember.assert("Must pass a hash of attributes to materializeAttributes", !!attributes);
+    Ember.assert("Must pass a object to materializeAttributes", !!attributes);
```
knownasilya: an object?

bmac: Fixed. Thanks.

bmac: Fixed. Thanks.

PR #2703, opened 2015-01-15
Document that ManyArrays are no longer RecordArrays in the changelog.

### LINKED ISSUE #2611: ManyArrays are no longer an array proxy
We should note this in changes because people might have been relying on internals of .content

### LINKED ISSUE #2665: Removed 'content' property from relationships
This used to work just fine in ember
```js
//each segment has many messages
var messages = segment.get("messages");

//then to get the plain array
var messagesArray = messages.get("content");
```

But now I need to do this

```js
var messagesArray = messages.get("currentState");
```

I have some old JS code that needs to see plain arrays, for indecies and things like that, Removing this feature caused me to change alot of my code

bmac:
Closes #2611
Closes #2665

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -93,6 +93,27 @@ export default DS.RESTSerializer.extend({
 Please use `store.metadataFor()` to get metadata and `store.setMetadataFor()`
 to set metadata.
 
+
+##### `ManyArray`s are no longer `RecordArray`s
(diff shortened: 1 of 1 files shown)


wecc:
I think this should be merged and #2667 reverted

bmac:
I don't think it hurts to have both this pr and #2667 in the change log since they both touch different sections of the changelog.

PR #2709, opened 2015-01-19
Fixed `@extends` and `@uses` for ManyArray doc

lolmaus:
Before: 

    @extends DS.RecordArray

But `DS.ManyArray` does not extend `DS.RecordArray`!

After:

    @extends Ember.Object
    @uses Ember.MutableArray, Ember.Evented


bmac:
Thanks!

Issue #2708, opened 2015-01-19
Should `deleted.inFlight` support `becameInvalid`?


ppcano:
Currently, when the server response of DELETE is a validation error, the record state (deleted.inFlight) does not manage the `becameInvalid` event.

I am not sure, if this implementation is missing or is a design choice, so `deletion` should only response to normal errors, avoiding validation errors.

Right now, rollback the record state may be solved as:
```
    item.destroyRecord().then(function() {
      ....
    }).catch(function(error) {
      if (error instanceof InvalidError) {
        self.store.recordWasError(item, error);
      }
      item.rollback();
   });
```

wecc:
Duplicate of #2126, open PR at #2169.

ppcano:
Closing in favour of #2710

PR #2169, opened 2014-08-08
Fixes #2126: Invalid state for root.deleted

duggiefresh:
Fixes #2126.

* Adds `becameInvalid` handler for `deleted.inFlight` state.
* Sets `Ember.run` to `run` within test.

PR #2710, opened 2015-01-19
Fix: a record in state deleted.invalid can be rollbacked

### LINKED ISSUE #2126: Invalid state for root.deleted
Apparently Ember does not allow me to invalidate a DELETE request.

If my ```App.model.destroyRecord()``` returns a default ```Errors object``` with the message 'Deletion is not allowed', embers raises a 

```
Attempted to handle event `becameInvalid` on <App.Model:ember718:2> while in state root.deleted.inFlight
```

Since its not the object error itself, I can't retrieve the error message.

But... If I try to delete again, then I get the right error response from the server with the error messages returned.

```
DEBUG: -------------------------------
DEBUG: Ember      : 1.7.0-beta.3
DEBUG: Ember Data : 1.0.0-beta.8.2a68c63a
DEBUG: Handlebars : 1.3.0
DEBUG: jQuery     : 1.11.1
DEBUG: -------------------------------
```

ppcano:
Fix #2126, #2169, #2708 

Defines the `root.deleted.invalid` state and its different events. 

Now  a record which has been invalidated during its `deletion` can be rollbacked on the same way that if it would receive an error response.

### DIFF
--- a/packages/ember-data/lib/system/model/states.js
+++ b/packages/ember-data/lib/system/model/states.js
@@ -658,6 +658,11 @@ var RootState = {
       becameError: function(record) {
         record.transitionTo('uncommitted');
         record.triggerLater('becameError', record);
+      },
+
+      becameInvalid: function(record) {
+        record.transitionTo('invalid');
+        record.triggerLater('becameInvalid', record);
       }
     },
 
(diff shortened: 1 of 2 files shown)


packages/ember-data/lib/system/model/states.js
```diff
@@ -658,6 +658,11 @@ var RootState = {
       becameError: function(record) {
         record.transitionTo('uncommitted');
         record.triggerLater('becameError', record);
+      },
+
+      becameInvalid: function(record) {
+        record.transitionTo('invalid');
+        record.triggerLater('becameInvalid', record);
```
bmac: Why do we run `triggerLater` here?

ppcano: I thought, it needed to be called because `DirtyState.inFlight.becameInvalid` calls `record.send('invokeLifecycleCallbacks')` and then:
```js
invalid: {
    invokeLifecycleCallbacks: function(record) {
      record.triggerLater('becameInvalid', record);
    }
}
```

The implementation is based on other `invalid event` implementations but honestly, I didn't understand well when `record.triggerLater('becameInvalid', record);` is necessary.

If I removed this `triggerLater`, the test pass. Do you want to me to update PR?

tomdale:
@ppcano Does this supersede #2169? If we take this PR, do we not need to take the other?

ppcano:
@tomdale yes, this one supersede #2169

jcope2013:
any timeline on this landing soon?

bmac:
Thanks

atomkirk:
Isn't the better approach to include the `DirtyState` in the deleted state? Per the docs:
>The dirty state is a abstract state whose functionality is
>shared between the `created` and `updated` states.

Issue #2405, opened 2014-10-22
beta-11 breaks belongsTo (async) relationship


shark404:
We have switched from beta-9 to beta-11 as we need some of the changes in the findHasMany, however  we've encountered issues with relationships failing to load. I believe this is very similar to #2391 but I'm not sure so raising a separate issue.

I've finally managed to recreate a simple example showing the issue. I've limited it down to having 2 different objects referencing the same underlying object. i.e. Post 1 and Post 2 both have a relationship to Post 100. Post 2 works ok but Post 1 fails to load properly. When looking in the ember inspector I notice that Post 1's child is a promise with no content, however if I look in _data the child is populated.. Post 2 is completely fine.

Here is the jsbin
http://emberjs.jsbin.com/hiciviyome/1/

Hopefully that makes sense...

Issue #2272, opened 2014-09-10
Different belongsTo behaviour beta9 vs beta10


schnoddelbotz:
I'm just getting started with ember(-data), so my approach may simply be wrong (sorry if so).
I've isolated my problem into jsbins:
http://jsbin.com/capubefemifi/2   http://jsbin.com/capubefemifi/3
http://jsbin.com/capubefemifi/2/edit   http://jsbin.com/capubefemifi/3/edit
Both are using the same code, but .../2 uses ember-data beta9 and .../3 uses beta10.
The plan was to build a nested tree... Any ideas why they give such different results (only the first/root record should report "isRoot")?

cheers

PR #2712, opened 2015-01-19
Assert of the user specifies a reflexive relationship without explicitly...

bmac:
... defining the inverse

Currently if a user specifies a reflexive belongsTo relationship in Ember Data without providing a `inverse` key Ember Data will assume the relationship is a 1 to 1 relationship. This can be confusing (see https://github.com/emberjs/data/issues/2272 and https://github.com/emberjs/data/issues/2405) since in practice 1 to 1 relationships are rarely used. 

This pr forces the user to specify a inverse key if they define a reflexive relationship. 1 to 1 relationships are still possible you just need to specify the relationship as its own inverse.

### DIFF
--- a/packages/ember-data/lib/system/relationships/ext.js
+++ b/packages/ember-data/lib/system/relationships/ext.js
@@ -228,11 +228,14 @@ Model.reopenClass({
       return null;
     }
 
+    var propertyMeta = this.metaForProperty(name);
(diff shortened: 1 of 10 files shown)


bmac:
Flagging this for review because its a breaking change for users who have 1 to 1 relationships without an `inverse` key.

PR #2715, opened 2015-01-20
Do not double include the host when it uses a protocol relative url.

### LINKED ISSUE #2714: Using protocol relative URLs in REST Adapter can cause issues
When using a protocol relative URL in a REST Adapter, issues happen when attempting to retrieve related models. Example:

Library Adapter:

```
store = DS.Store.extend
    adapter: DS.RESTAdapter.extend
      namespace: ''
      host: '//my.library.com'

  app.register 'library:main', store
  app.inject 'route', 'library', 'library:main'
```

Library Model:

```
Library = DS.Model.extend
  name: DS.attr 'string'
  groups: DS.hasMany 'group', {async: true}
```

When iterating over library.groups in a template, I get the following errors:

`Cannot load http://my.library.com//my.library.com/groups/12341234/assets`

The problem is that the URL is wrong, `http://my.library.com/groups/12341234/assets` is the correct URL.

Forcing a protocol (`host: window.location.protocol+...`) in the adapter declaration fixes the issue.

bmac:
Closes #2714

### DIFF
--- a/packages/ember-data/lib/adapters/rest_adapter.js
+++ b/packages/ember-data/lib/adapters/rest_adapter.js
@@ -562,8 +562,15 @@ export default Adapter.extend({
     var url = [];
 
     if (path) {
+      // Protocol relative url
+      //jscs:disable disallowEmptyBlocks
+      if (/^\/\//.test(path)) {
+        // Do nothing, the full host is already included. This branch
+        // avoids the absolute path logic and the relative path logic.
+
       // Absolute path
-      if (path.charAt(0) === '/') {
+      } else if (path.charAt(0) === '/') {
+        //jscs:enable disallowEmptyBlocks
         if (host) {
           path = path.slice(1);
           url.push(host);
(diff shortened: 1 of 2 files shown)


fivetanley:
LGTM, just needs style fixes for the failures on travis.

packages/ember-data/lib/adapters/rest_adapter.js
```diff
@@ -562,8 +562,14 @@ export default Adapter.extend({
     var url = [];
 
     if (path) {
+      // Protocal relative url
```
wecc: "Protocol"?

packages/ember-data/lib/adapters/rest_adapter.js
```diff
@@ -562,8 +562,14 @@ export default Adapter.extend({
...
+      //jscs:disable disallowEmptyBlocks
```
wecc: Should `disallowEmptyBlocks` be enabled again after the `if` block?

PR #2717, opened 2015-01-22
Add rootModel setting to RESTAdapter

funtusov:
This would add support for REST API's with embedded models:

If post has many comments, then 

```js
# /app/adapters/comments
export default ApplicationAdapter.extend({
  rootModel: 'post'
});
```

would generate ``/posts/:post_id/comments/:id`` as the url

Embedded models often cannot be accessed by id directly (i.e. mongodb) and many API's use something along the lines of:
```ruby
Post.find(params[:post_id]).comments.find(params[:id])
```
To access embedded models.

That's something I use and it seems it's quite widely requested:
- http://stackoverflow.com/questions/21964155/override-buildurl-method-to-include-parent-models-id
- https://github.com/emberjs/data/pull/39
- https://github.com/emberjs/data/pull/57

The alternative is the pathTemplate RFC: https://github.com/emberjs/rfcs/pull/4 

It seems that a single nested nesting covers most cases and is along the lines of how other properties are set on the adapter.

### DIFF
--- a/packages/ember-data/lib/adapters/rest_adapter.js
+++ b/packages/ember-data/lib/adapters/rest_adapter.js
@@ -113,6 +113,18 @@ var forEach = Ember.ArrayPolyfills.forEach;
   });
   ```
 
+  ### Root model customization
+
+  An adapter can scope requests to a root model by setting the rootModel
+  property on the adapter:
+
+  ```js
+  DS.RESTAdapter.reopen({
+    rootModel: 'post'
+  });
+  ```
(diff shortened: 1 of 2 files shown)


tomdale:
I like the idea in theory, but I think that in practice it does not save that much typing, and having two ways to specify a URL template seems confusing. In general, we'd prefer to have one way to do something rather than several.

Thank you so much for the PR, and let's keep the discussion going if you feel like the path template RFC does not satisfy your needs.

PR #2718, opened 2015-01-22
Add a Serializer base class

### LINKED ISSUE #1658: Serializer Method Documentation
I've spent the last few hours implementing my own serializer, and I feel as if the documentation is a bit lacking when it comes to implementing one from scratch. The JSON serializer (which is what the REST serializer extends from) has 24 methods. These methods are mostly hooks for customization, which work great when you're extending one of the build in serializers. However, if you're creating one from scratch, it can be difficult to determine which methods are necessary in a serializer (ie. which ones are called externally).

So far, I think I've figured out that only the `serialize` and `extract` methods are necessary, but I'm not 100% sure on that. Is it possible to make the documentation a little more clear in that aspect? Or, better yet, create an abstract `Serializer` class to extend from, similar to the `Adapter` class.

wecc:
`serializer.extract()` and `serializer.serialize()` should be the only methods you need to implement if you're building your own serializer but I've included `serializer.normalize()` since it's used by `store.normalize()` and there's no checking if it exists or not (as opposed to `store.pushPayload()` that asserts that `serializer.pushPayload()` has been implemented).

Closes #1658

### DIFF
--- a/packages/ember-data/lib/main.js
+++ b/packages/ember-data/lib/main.js
@@ -36,6 +36,7 @@ import {
   InvalidError,
   Adapter
 } from "ember-data/system/adapter";
+import Serializer from "ember-data/system/serializer";
 import DebugAdapter from "ember-data/system/debug";
 import {
   RecordArray,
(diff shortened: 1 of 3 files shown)


bmac:
Looks good to me. :+1:

PR #2719, opened 2015-01-22
Update JSBin link

wecc:
- Version 1 was missing ember-template-compiler.js which caused it to error out
- Link to the latest version


bmac:
Thanks

PR #2720, opened 2015-01-24
add documentation for the Store's find method

omarismail:
I was trying to do some more sophisticated querying with `find` and the docs were not clear enough as to what exactly I should do. I ended up looking in the source code at the Store's `find` method to figure out what needed to be done. This is a PR to improve the docs so that others don't need to go into the source code to figure it out. 

I hope this helps. Please let me know if I could improve it in any way. Thanks.

### DIFF
--- a/packages/ember-data/lib/system/store.js
+++ b/packages/ember-data/lib/system/store.js
@@ -440,9 +440,37 @@ Store = Ember.Object.extend({
     store.find('person', { page: 1 });
     ```
 
-    This will ask the adapter's `findQuery` method to find the records for
-    the query, and return a promise that will be resolved once the server
-    responds.
+    By passing an object `{page: 1}` as an argument to the find method, it
+    delegates to the adapter's findQuery method. The adapter then makes
+    a call to the server, transforming the object `{page: 1}` as parameters
+    that are sent along, and will return a RecordArray when the promise
+    resolves.
+
+    Exposing queries this way seems preferable to creating an abstract query
+    language for all server-side queries, and then require all adapters to
+    implement them.
+
+    The call made to the server, using a Rails backend, will look something like this:
+
+    ```
+    Started GET "/api/v1/person?page=1"
+    Processing by Api::V1::PersonsController#index as HTML
+    Parameters: {"page"=>"1"}
+    ```
+
+    If you do something like this:
+
+    ```javascript
+    store.find('person', {ids: [1, 2, 3]});
+    ```
+
+    The call to the server, using a Rails backend, will look something like this:
+
+    ```
+    Started GET "/api/v1/person?ids%5B%5D=1&ids%5B%5D=2&ids%5B%5D=3"
+    Processing by Api::V1::PersonsController#index as HTML
+    Parameters: {"ids"=>["1", "2", "3"]}
+    ```
 
     @method find
     @param {String or subclass of DS.Model} type


packages/ember-data/lib/system/store.js
```diff
@@ -440,9 +440,32 @@ Store = Ember.Object.extend({
     store.find('person', { page: 1 });
     ```
 
-    This will ask the adapter's `findQuery` method to find the records for
-    the query, and return a promise that will be resolved once the server
-    responds.
+    This will call the private method `findQuery`, which delegates the query to the adapter. The adapter then makes a call to the server, which will return a `RecordArray` when the promise resolves.
```
bmac: > This will call the private method `findQuery`

I'm not sure if we want to document the internal behavior like this. Is there a specific reason why you thought it would be useful to call out the store's `findQuery` method?

packages/ember-data/lib/system/store.js
```diff
@@ -440,9 +440,32 @@ Store = Ember.Object.extend({
...
+    Exposing queries this way seems preferable to creating an abstract query
+    language for all server-side queries, and then require all adapters to
+    implement them.
+
+    The call made to the server, depending on what backend you are using, will look something like this:
+
+    ```
+    Started GET "/api/v1/person?page=1"
+    Processing by Api::V1::PersonsController#index as HTML
+    Parameters: {"page"=>"1"}
```
bmac: Is this example form a rails backend? It might make sense to call that out explicitly.

packages/ember-data/lib/system/store.js
```diff
@@ -440,9 +440,32 @@ Store = Ember.Object.extend({
...
+    If you do something like this:
+    ```javascript
+    store.find('person', {ids: [1, 2, 3]});
+    ```
+
+    The call to the server, depending on what backend you are using, will look something like this:
+
+    ```
+    Started GET "/api/v1/person?ids%5B%5D=1&ids%5B%5D=2&ids%5B%5D=3"
+    Processing by Api::V1::PersonssController#index as HTML
```
bmac: `PersonssController` -> `PersonsController`

packages/ember-data/lib/system/store.js
omarismail: You're right, it wouldn't be good to document internal behavior. My intent was to clarify that by passing an object `{page: 1}` has an argument, it delegates to another method which ultimately makes a call to the server and transforms the object as parameters. How about this rephrasing: 

> By passing an object `{page: 1}` as an argument to the `find` method, it delegates to the adapter's `findQuery` method. The adapter then makes a call to the server, transforming the object `{page: 1}` as parameters that are sent along, and will return a `RecordArray` when the promise resolves.

packages/ember-data/lib/system/store.js
omarismail: Sure thing. I wasn't sure if it would be good to include a specific backend framework rather than trying to be agnostic, but I wanted to provide an example to aid in the clarification.

packages/ember-data/lib/system/store.js
bmac: That works for me.

omarismail:
Thanks for the feedback @bmac. Is this all good?

packages/ember-data/lib/system/store.js
```diff
@@ -440,9 +440,36 @@ Store = Ember.Object.extend({
     store.find('person', { page: 1 });
     ```
 
-    This will ask the adapter's `findQuery` method to find the records for
-    the query, and return a promise that will be resolved once the server
-    responds.
+    By passing an object {page: 1} as an argument to the find method, it
+    delegates to the adapter's findQuery method. The adapter then makes
+    a call to the server, transforming the object {page: 1} as parameters
```
bmac: Lets surround the `{page: 1}` with backticks "`" to make this easier to read when rendered. 
screen shot 2015-01-24 at 3 13 22 pm

packages/ember-data/lib/system/store.js
```diff
@@ -440,9 +440,36 @@ Store = Ember.Object.extend({
...
+    The call made to the server, using a Rails backend, will look something like this:
+
+    ```
+    Started GET "/api/v1/person?page=1"
+    Processing by Api::V1::PersonsController#index as HTML
+    Parameters: {"page"=>"1"}
+    ```
+
+    If you do something like this:
+    ```javascript
```
bmac: This needs a newline between the end of the "If you do something like this:" line and the start of the javascript code block otherwise the it doesn't seem to get rendered correctly.

screen shot 2015-01-24 at 3 15 38 pm

packages/ember-data/lib/system/store.js
omarismail: :+1:

packages/ember-data/lib/system/store.js
omarismail: :+1:

bmac:
Thanks

PR #2722, opened 2015-01-24
Do not re-add deleted records to a hasMany relationship

### LINKED ISSUE #2666: adding and removing records from hasMany relationship broken.
I am experiencing strange behavior in the latest version (1.0.0-beta.14.1) of ember data. 

It appears that items removed from a hasMany collection reappear when adding new items.

Checkout the following jsbin http://jsbin.com/boluba/13

Remove an Item then add an item, the removed item will reappear.  The strange thing is the item is not in the store when using the ember inspector.

bmac:
Closes #2666

Thanks to @dwickern for the excellent test case. 

@igorT if you have time I'd love to get your eyes on this

### DIFF
--- a/packages/ember-data/lib/system/many-array.js
+++ b/packages/ember-data/lib/system/many-array.js
@@ -5,6 +5,7 @@ import { PromiseArray } from "ember-data/system/promise-proxies";
 
 var get = Ember.get;
 var set = Ember.set;
+var filter = Ember.ArrayPolyfills.filter;
 
 /**
   A `ManyArray` is a `MutableArray` that represents the contents of a has-many
@@ -65,7 +66,10 @@ export default Ember.Object.extend(Ember.MutableArray, Ember.Evented, {
 
   flushCanonical: function() {
     //TODO make this smarter, currently its plenty stupid
-    var toSet = this.canonicalState.slice(0);
+    var toSet = filter.call(this.canonicalState, function(record) {
+      return !record.get('isDeleted');
+    });
+
     //a hack for not removing new records
     //TODO remove once we have proper diffing
     var newRecords = this.currentState.filter(function(record) {
--- a/packages/ember-data/tests/integration/relationships/has-many-test.js
+++ b/packages/ember-data/tests/integration/relationships/has-many-test.js
@@ -1255,3 +1255,60 @@ test('unloading a record with associated records does not prevent the store from
     ok(false, "store prevented from being destroyed");
   }
 });
+
+test("adding and removing records from hasMany relationship #2666", function(){
+  expect(4);
+
+  var Post = DS.Model.extend({
+    comments: DS.hasMany('comment', { async: true })
+  });
+  Post.reopenClass({
+    FIXTURES: [
+      { id: 1, comments: [ 1, 2, 3 ] }
+    ]
+  });
+
+  var Comment = DS.Model.extend({
+    post: DS.belongsTo('post')
+  });
+  Comment.reopenClass({
+    FIXTURES: [
+      { id: 1 },
+      { id: 2 },
+      { id: 3 }
+    ]
+  });
+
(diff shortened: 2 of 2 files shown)


fivetanley:
LGTM, I'm a little concerned at using slice here though instead of just filter (filter returns a new array). Is this a JS array or one of ember's array-likes?

bmac:
Good call @fivetanley. It was a JS array so I have updated the pr to use the filter function form `Ember.ArrayPolyfills`.

arenoir:
@bmac yep that works.  :+1:

tonywok:
Ah, nice, our team was investigating this as well. :+1:

tomdale:
@igorT is working on a more comprehensive solution, and will make sure to include this test case in his in-progress work.

tonywok:
@tomdale Is the fix @igorT is working on scoped to just this issue? Or something larger? If larger, is there a branch I can follow along with? Thanks :)

bmac:
@igorT anything I can do to help with move your work on this along?

bmac:
I recognize that this is a bandaid fix however I'm going to merge it as its holding people back from upgrading past Ember Data beta.12 which will soon be broken with the release version of ember (1.11) due to some changes in Ember.Map.

davidlormor:
Just want to confirm that this patch got rid of my bug, and allowed me to get rid of a poorly implemented workaround. Thanks! :+1:

drewcovi:
@bmac does this fix also account for simply removing a record from a relationship and then saving that record? I'm seeing records re-appear and checking record.get('isDeleted') will not register in that case (since we're not destroying a record)

codylettau:
@drewcovi has there been any update with the issue you are seeing in regards to only removing (not deleting) object in many-to-one relationships? We are encountering this same exact issue as well.

drewcovi:
@codylettau I've logged a new bug here: https://github.com/emberjs/data/issues/3027

From what i can tell this is definitely a bug.

denisbetsi:
This is a major bug, making a part of the project totally unusable. There's no work arounds that are clean. If using any sort of a front end tool like strongloop which depends on creating relationships between records, this creates a havoc.

denisbetsi:
Found a temporary work around, 

Prior to deleting items from hasMany I create a new field called deleted and set it to 1

```
     that.get("productBundles").forEach(function(olditem){
              olditem.set("deleted",1);
              olditem.destroyRecord();
     });
```

Then I extend DS.RESTSerializer, I also have to change the relationship key in my case to end with Id vs standard ember handling so it's not related to this hack

```
	serialize: function(record, options) {
	    record.eachRelationship(function(name, relationship) {
	      if (relationship.kind === 'hasMany'||relationship.kind === 'belongsTo') {
	      	var singular = name.singularize() + 'Id';
	      	if(name.endsWith("s")&&typeof record.get(name) != 'string' &&relationship.kind === 'hasMany'){
	        	json[singular] = (record.get(name)?record.get(name).filter(function(item){return !item.record.deleted}).mapBy('id'):[]);
	      	}else{
	      		if(record.get(name)&&record.get(name).get("id"))
		      		json[singular] = (record.get(name)?record.get(name).get('id'):'');
	      	}
	      }
	    });
	    return json;
	}
});
```

So prior to saving the record, I look for deleted and filter them out and overwrite the request

```
record.get(name).filter(function(item){return !item.record.deleted})
```

Now I am battling another yet unexplained  that.get("productBundles").invoke('destroyRecord') leaves 1 record in the array and just doesn't issue delete command on it. 

Some days I absolutely love Ember and this isn't one of those days. The simplest thing becomes a 2 day hackathon in finding a workaround to what should be a solid feature after years of development. ARGGH!!

dwickern:
@denisbetsi Are you on the latest version of ember-data? This should be fixed now.

denisbetsi:
I tried using this version //builds.emberjs.com/beta/ember-data.js

PR #2723, opened 2015-01-25
Correctly trigger arrayContentDidChange when updating hasMany relationsh...

### LINKED ISSUE #2713: Reloading a many to many relationship from server.
I've got a many to many reflexive(not sure if this matters) relationship that I want to get fresh up to date data from the server.

If I add a Record A from Record B's of the relationship, then reload Record B, Record A still has the entry for Record B, so it is then added to Record B's array again. To get the functionality I want, I've found that i have to disconnect the relationships, reload all the records, then reconnect them, like so:

```
var component = this;
this.get('model.parents.content.relationship').disconnect();
this.get('model').reload().then(function (model) {
    return model.get('parents').reload();
}).then(function () {
    component.get('model.parents.content.relationship').reconnect();
});
```

This means I have to reload every record in the many/many, and tinker with some internals that I've only read briefly.

I've got 2 of these kind of relationships one of my models, and it seems to pain to overload reload() to add in this functionality.

bmac:
...ips.

Closes #2713 

This pr still needs tests.

### DIFF
--- a/packages/ember-data/lib/system/record_arrays/many_array.js
+++ b/packages/ember-data/lib/system/record_arrays/many_array.js
@@ -74,10 +74,11 @@ export default Ember.Object.extend(Ember.MutableArray, Ember.Evented, {
       return record.get('isNew');
     });
     toSet = toSet.concat(newRecords);
-    this.arrayContentWillChange(0, this.length, this.length);
+    var oldLength = this.length;
+    this.arrayContentWillChange(0, this.length, toSet.length);
     this.set('length', toSet.length);
     this.currentState = toSet;
-    this.arrayContentDidChange(0, this.length, this.length);
+    this.arrayContentDidChange(0, oldLength, this.length);
     //TODO Figure out to notify only on additions and maybe only if unloaded
     this.relationship.notifyHasManyChanged();
     this.record.updateRecordArrays();
--- a/packages/ember-data/tests/unit/many_array_test.js
+++ b/packages/ember-data/tests/unit/many_array_test.js
@@ -66,3 +66,37 @@ test("manyArray.removeRecord() has been deprecated", function() {
     equal(tags.length, 0, 'there should not be any tags');
   });
 });
+
+
+test("manyArray trigger arrayContentChange functions with the correct values", function() {
+  expect(12);
+  var willChangeStartIdx;
+  var willChangeRemoveAmt;
+  var willChangeAddAmt;
+  var originalArrayContentWillChange = DS.ManyArray.prototype.arrayContentWillChange;
+  var originalArrayContentDidChange = DS.ManyArray.prototype.arrayContentDidChange;
+  DS.ManyArray.reopen({
+    arrayContentWillChange: function(startIdx, removeAmt, addAmt) {
+      willChangeStartIdx = startIdx;
+      willChangeRemoveAmt = removeAmt;
+      willChangeAddAmt = addAmt;
+      return this._super.apply(arguments);
+    },
+    arrayContentDidChange: function(startIdx, removeAmt, addAmt) {
+      equal(startIdx, willChangeStartIdx, 'WillChange and DidChange startIdx should match');
+      equal(removeAmt, willChangeRemoveAmt, 'WillChange and DidChange removeAmt should match');
+      equal(addAmt, willChangeAddAmt, 'WillChange and DidChange addAmt should match');
+      return this._super.apply(arguments);
+    }
+  });
+  run(function() {
+    store.push('tag', { id: 1, name: 'Ember.js' });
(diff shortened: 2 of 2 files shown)


bmac:
@igorT this is now ready for review. 

This pr fixes 2 issues from [this jsbin](http://jsbin.com/zugulu/1/edit). The first issue is the doubling of elements in the dom caused by incorrectly triggering `arrayContentWillChange` / `arrayContentDidChange`. This can be demoed by removing the first parent and clicking reload.

The second issue is the last item in the relationship will not be re-updated by `flushCanonicalLater` if its state changes but its position does not. This can be demoed by removing the last parent and clicking reload.

kimroen:
We have a few cases of this bug in our app - thanks for this.

packages/ember-data/tests/unit/many_array_test.js
```diff
@@ -66,3 +66,36 @@ test("manyArray.removeRecord() has been deprecated", function() {
     equal(tags.length, 0, 'there should not be any tags');
   });
 });
+
+
+test("manyArray trigger arrayContentChange functions with the correct values", function() {
```
sly7-7: maybe adding the expectation count would be valuable ?

packages/ember-data/lib/system/record_arrays/many_array.js
```diff
@@ -74,10 +74,11 @@ export default Ember.Object.extend(Ember.MutableArray, Ember.Evented, {
       return record.get('isNew');
     });
     toSet = toSet.concat(newRecords);
-    this.arrayContentWillChange(0, this.length, this.length);
+    var oldLength = this.length;
+    this.arrayContentWillChange(0, oldLength, toSet.length);
```
sly7-7: could'nt it be `this.arrayContentWillChange(0, this.length, toSet.length);` ?

packages/ember-data/lib/system/record_arrays/many_array.js
```diff
@@ -74,10 +74,11 @@ export default Ember.Object.extend(Ember.MutableArray, Ember.Evented, {
...
     this.set('length', toSet.length);
     this.currentState = toSet;
-    this.arrayContentDidChange(0, this.length, this.length);
+    this.arrayContentDidChange(0, oldLength, toSet.length);
```
sly7-7: could'nt it be this.arrayContentWillChange(0, oldLength, this.length); ?
I think this is ok because now, this.length refers to the 'new' length.

packages/ember-data/lib/system/relationships/state/has_many.js
```diff
@@ -139,12 +139,13 @@ ManyRelationship.prototype.computeChanges = function(records) {
...
     record = records[i];
     //Need to preserve the order of incoming records
-    if (hasManyArray.objectAt(i) === record ) {
+    if (hasManyArray.objectAt(i) === record) {
       continue;
     }
     this.removeCanonicalRecord(record);
     this.addCanonicalRecord(record, i);
   }
+  this.flushCanonicalLater();
```
sly7-7: I honestly don't know why this is needed. I mean, this makes sense to flush the canonical state, but why later ?

sly7-7:
@bmac I think you're right about the wrong length passed to arrayWill/DidChange. I'm confused about the flushCanonical things, but I trust you here :+1:

packages/ember-data/lib/system/relationships/state/has_many.js
bmac: `this.addCanonicalRecord` will also call `flushCanonicalLater `. This case is makes sure we don't run `flushCanonical` more then once it a given run loop. `this.flushCanonicalLater();` was added here because its possible for `addCanonicalRecord` to not be because a record was in the correct position but its state has been updated so the `flushCanonical` still needs to be re-run for the relationship to render correctly.

bmac: Also this line could get removed if https://github.com/emberjs/data/pull/2771 gets merged.

bmac: Remove this line.

joeruello:
:+1: Thanks alot

sly7-7:
@bmac :heart_eyes: :heart: Thanks to you I could now retry to update to latest

gmajoulet:
Thank you so much for this fix. :))

kimroen:
I have a bug that I thought this would fix, but after updating I'm still having the problem. Maybe someone could spare some time to pair or something on it? It's a bit tricky to recreate - many moving parts.

kimroen:
No never mind me - our bug is probably https://github.com/emberjs/data/pull/2722

PR #2725, opened 2015-01-26
Cleanup JSONSerializer#extract example

knownasilya:
Would love to add a nice example of overriding this method as well. Suggestions?


knownasilya:
Editing in Github no longer works, due to trailing whitespace.. blah!

bmac:
Thanks!

PR #2726, opened 2015-01-27
code doesn't seem to match comment

devinrhode2:


bmac:
Hi @devinrhode2. I think the comment is poorly worded but I believe the intention was that the date transform should only return `undefined` if there was no value in the original json. This change makes it so that if there is a value and the value is `null` then the transform will convert that value to `undefined` which is now breaking the tests. 

I don't think Ember Data is likely to accept this change as is but if you'd like to update the comment to make the intention more clear we would gladly merge that pr.

devinrhode2:
@ghempton what was the actual intention?

bmac:
Looks like @ghempton just moved that code around in that commit. The original comment came from @tomdale and @wycats in https://github.com/emberjs/data/commit/8a812c8e31b35c421679605b43beaa4af0918ee5

PR #2727, opened 2015-01-28
Remove non-breaking spaces

### LINKED ISSUE #2716: Debug version seems to have a bug
The simple page below with no components defined starts with an error, "Uncaught SyntaxError: Unexpected token typeof" on line 11348, while changing the the ember data version to prod or min shows no error. Maybe this is a minor issue because one can develop with the other two versions.

Tags didn't come through, so I'm paraphrasing the html.
doctype tag
html tag
head tag
script tag for jquery-1.11.2.min.js
script tag for handlebars-2.0.0.js
script tag for ember-1.9.1.debug.js
script tag for ember-data-1.0.0-beta.14.1.debug.js
script tag for localstorage_adapter.js

close head tag
body tag
close body tag
close html tag

wecc:
I think this is my bad to begin with, sometimes I'm not letting go of the alt key fast enough after writing a `|` followed by a space. This results in a non-breaking space being inserted instead of a regular space. Fixes #2716.

### DIFF
--- a/packages/ember-data/lib/system/relationships/belongs_to.js
+++ b/packages/ember-data/lib/system/relationships/belongs_to.js
@@ -64,7 +64,7 @@ function belongsTo(type, options) {
     type = undefined;
   }
 
-  Ember.assert("The first argument to DS.belongsTo must be a string representing a model type key, not an instance of " + Ember.inspect(type) + ". E.g., to define a relation to the Person model, use DS.belongsTo('person')", typeof type === 'string' || typeof type === 'undefined');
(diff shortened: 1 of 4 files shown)


bmac:
Thanks so much @wecc for tracking this down. Any ideas on how we can prevent this from happening in the future?

wecc:
@bmac There seem to be a `nonbsp` option (http://jshint.com/docs/options/#nonbsp) that can be enabled for JSHint. I've created a PR enabling this.

PR #2728, opened 2015-01-28
Deprecate store.find(type) and store.find(type, query)

### LINKED ISSUE #2607: Deprecate `store.find` acting as findQuery/findAll to use store.findAll and findQuery
Currently, store.find supports several function signatures:

`store.find('type')` => `store.findAll`
`store.find('type', {queryParam: 'hi'})` => `store.findQuery`
`store.find('type', '1')` => `store.find`

We should only support the last version of this form for 1.0, and deprecate until then.

wecc:
Should fix #2607

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,29 @@
 
 ### Master
 
+#### Breaking Changes
+
+##### `store.find(type)` and `store.find(type, query)` has been deprecated
+
+Using `store.find()` to find all records by type or by type and query has been
+deprecated. Use `store.find()` only to find a single record by type and id.
+
+Use `store.findAll()` to find all records by type and `store.findQuery()` to
+find all records by type and query.
+
+Example
+
+```javascript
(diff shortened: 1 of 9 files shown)


bmac:
This is going to be a big change. We should also have a note about this in the change log and be ready to update the guides in the https://github.com/emberjs/website repo.

wecc:
I've added breaking changes the CHANGELOG and created a PR to update the guides at https://github.com/emberjs/website/pull/1966

pixelhandler:
I'm curious how the default model hook in the route will behave, will it call find, findAll, or findQuery. Will here be branching based on the params?

rwjblue:
@pixelhandler - Wonderful question!

wecc:
@pixelhandler @rwjblue From my understanding the default model hook only loads data from the store when we have a model and an id, in which case the current call to find() is the right one.

wecc:

PR #2729, opened 2015-01-28
Added save() to ManyArray

### LINKED ISSUE #2669: ManyArray no longer has allows save on collection
We previously would retrieve a hasMany relation, then save the relation and all records associated with it. Now that no longer works in ember data 14. It appears the methods we used to use on RecordArray are no longer there for the ManyArray.

You can see it here in this bin. http://jsbin.com/jisufu/1 

Looking at the other bug tickets, I'm guessing it's probably related to the other tickets about many array and using .content.

wecc:
Closes #2669

### DIFF
--- a/packages/ember-data/lib/system/record_arrays/many_array.js
+++ b/packages/ember-data/lib/system/record_arrays/many_array.js
@@ -1,6 +1,7 @@
 /**
   @module ember-data
 */
+import { PromiseArray } from "ember-data/system/promise_proxies";
 
 var get = Ember.get;
 var set = Ember.set;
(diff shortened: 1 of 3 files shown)


packages/ember-data/lib/system/record_arrays/many_array.js
```diff
@@ -186,6 +187,34 @@ export default Ember.Object.extend(Ember.MutableArray, Ember.Evented, {
...
+    });
+    ```
+
+    @method save
+    @return {DS.PromiseArray} promise
+  */
+  save: function() {
+    var promiseLabel = "DS: ManyArray#save " + get(this, 'type');
+    var promise = Ember.RSVP.all(this.invoke("save"), promiseLabel).then(function(array) {
+      return Ember.A(array);
```
bmac: Should this return the original `ManyArray`?

wecc: I think that makes sense. I'm guessing `RecordArray.save()` should do that too then, to keep it consistent?

stefanpenner: it should yes

wecc: Updated `ManyArray.save()` and `RecordArray.save()` to return themselves in a separate commit.

bmac:
:+1:

PR #2732, opened 2015-01-29
Revert "generateIdForRecord gets type & object properties passed to it"

bmac:
Reverts emberjs/data#2637


igorT:
Whats going on here?

PR #2736, opened 2015-01-29
Fix to stop phantomjs from crashing with the Ember beta travis tests

bmac:


fivetanley:
sadlol

fivetanley:
thanks!

bmac:
After some trial and error I've figured out a way to make the beta tests complete without crashing phantomjs.

I'm still at a loss as to explaining why this change helps. Or why the tests were only crashing for the beta version of Ember.

PR #2741, opened 2015-01-30
getter metod for record's relationships

andrewfan:
Creating instances of model with many relationships takes too many time, it happens because all relationships are creating during _setup method. 
I found comment https://github.com/emberjs/data/blob/master/packages/ember-data/lib/system/model/model.js#L504, and decided implement getter method.

### DIFF
--- a/packages/ember-data/lib/system/model/model.js
+++ b/packages/ember-data/lib/system/model/model.js
@@ -3,7 +3,7 @@ import Errors from "ember-data/system/model/errors";
 import { PromiseObject } from "ember-data/system/promise-proxies";
 import merge from "ember-data/system/merge";
 import JSONSerializer from "ember-data/serializers/json-serializer";
-import createRelationshipFor from "ember-data/system/relationships/state/create";
+import Relationships from "ember-data/system/relationships/state/create";
 import Snapshot from "ember-data/system/snapshot";
 
 /**
@@ -483,7 +483,7 @@ var Model = Ember.Object.extend(Ember.Evented, {
     this._data = {};
     this._attributes = Ember.create(null);
     this._inFlightAttributes = Ember.create(null);
-    this._relationships = {};
+    this._relationships = new Relationships(this);
     /*
       implicit relationships are relationship which have not been declared but the inverse side exists on
       another record somewhere
@@ -505,11 +505,6 @@ var Model = Ember.Object.extend(Ember.Evented, {
       when we are deleted
     */
     this._implicitRelationships = Ember.create(null);
-    var model = this;
-    //TODO Move into a getter for better perf
-    this.constructor.eachRelationship(function(key, descriptor) {
-      model._relationships[key] = createRelationshipFor(model, descriptor, model.store);
-    });
 
   },
 
@@ -695,8 +690,8 @@ var Model = Ember.Object.extend(Ember.Evented, {
   */
   clearRelationships: function() {
     this.eachRelationship(function(name, relationship) {
-      var rel = this._relationships[name];
-      if (rel) {
+      if (this._relationships.has(name)) {
+        var rel = this._relationships.get(name);
         //TODO(Igor) figure out whether we want to clear or disconnect
         rel.clear();
         rel.destroy();
@@ -711,7 +706,7 @@ var Model = Ember.Object.extend(Ember.Evented, {
 
   disconnectRelationships: function() {
     this.eachRelationship(function(name, relationship) {
-      this._relationships[name].disconnect();
+      this._relationships.get(name).disconnect();
     }, this);
     var model = this;
     forEach.call(Ember.keys(this._implicitRelationships), function(key) {
@@ -721,7 +716,7 @@ var Model = Ember.Object.extend(Ember.Evented, {
 
   reconnectRelationships: function() {
     this.eachRelationship(function(name, relationship) {
-      this._relationships[name].reconnect();
+      this._relationships.get(name).reconnect();
     }, this);
     var model = this;
     forEach.call(Ember.keys(this._implicitRelationships), function(key) {
@@ -787,7 +782,7 @@ var Model = Ember.Object.extend(Ember.Evented, {
     });
     //We use the pathway of setting the hasMany as if it came from the adapter
     //because the user told us that they know this relationships exists already
-    this._relationships[key].updateRecordsFromAdapter(recordsToSet);
+    this._relationships.get(key).updateRecordsFromAdapter(recordsToSet);
   },
 
   _preloadBelongsTo: function(key, preloadValue, type) {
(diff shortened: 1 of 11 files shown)


tonywok:
:+1: Nice. Seems like the `_implicitRelationships` could use a similar treatment, right?

andrewfan:
@tonywok i will investigate it in the beginning of the week

packages/ember-data/lib/system/model/model.js
```diff
@@ -473,12 +473,31 @@ var Model = Ember.Object.extend(Ember.Evented, {
...
+      get: function(key) {
+        var relationships = this.initializedRelationships;
+        var relationshipsByName = get(record.constructor, 'relationshipsByName');
+        if (!relationships[key] && relationshipsByName.get(key)) {
+          try {
+            relationships[key] = createRelationshipFor(record, relationshipsByName.get(key), record.store);
+          }
+          catch(e) {
+            throw e;
+          }
```
bmac: Why the try catch throw?

packages/ember-data/lib/system/model/model.js
```diff
@@ -690,18 +704,20 @@ var Model = Ember.Object.extend(Ember.Evented, {
...
   clearRelationships: function() {
     this.eachRelationship(function(name, relationship) {
-      var rel = this._relationships[name];
-      if (rel) {
-        //TODO(Igor) figure out whether we want to clear or disconnect
-        rel.clear();
-        rel.destroy();
+      if (this._relationships.has(name)) {
+        var rel = this._relationships.get(name);
+        if (rel) {
```
bmac: Now that we have the `has` check do we need this second test on the `rel` value?

packages/ember-data/lib/system/model/model.js
```diff
@@ -777,7 +793,8 @@ var Model = Ember.Object.extend(Ember.Evented, {
     });
     //We use the pathway of setting the hasMany as if it came from the adapter
     //because the user told us that they know this relationships exists already
-    this._relationships[key].updateRecordsFromAdapter(recordsToSet);
+    this.get(key);
```
bmac: Why do we need this `get` here?

fivetanley: The relationship may not be properly instantiated at this point, using get will create it if it does not exist it seems

packages/ember-data/lib/system/model/model.js
```diff
@@ -473,12 +473,31 @@ var Model = Ember.Object.extend(Ember.Evented, {
...
 
   _setup: function() {
+    var record = this;
     this._changesToSync = {};
     this._deferredTriggers = [];
     this._data = {};
     this._attributes = Ember.create(null);
     this._inFlightAttributes = Ember.create(null);
-    this._relationships = {};
+    this._relationships = Ember.create({
```
fivetanley: I think we should move this out to its own `Relationships` or similar function / prototype. Otherwise this will begin to create more memory than it needs to as each object will have its own copy of the function.

packages/ember-data/lib/system/model/model.js
bmac: I see I had thought the `this._relationships.get(key)` in the next line would also instantiate the relationship. My mistake.

packages/ember-data/lib/system/model/model.js
```diff
@@ -473,12 +473,31 @@ var Model = Ember.Object.extend(Ember.Evented, {
...
+      initializedRelationships: {},
```
bmac: Instead of `{}` this should be `Ember.create(null)` to prevent clashes in the unlikely event that a relationship shares a name with a property on the Object prototype. You may also want to move this into an init function of you make a `Relationships` object like @fivetanley suggests above.

packages/ember-data/lib/system/model/model.js
fivetanley: Ah, maybe that's the case? I forgot `this` wasn't the `_relationships`. Yeah, I'm confused as well at the get here.

andrewfan:
@bmac @fivetanley thx for code review. i hope i fixed all typos, also as you suggested i moved `Relationships` constructor outside model file.  `get` and `has` are prototype methods now.

andrewfan:
@tonywok 
> :+1: Nice. Seems like the _implicitRelationships could use a similar treatment, right?

We don't need to do similar optimizations for `_implicitRelationships` because those relationships are creating only when (correct me if i'm wrong):

1) we do `store.push` with relationships in json, f.e 

    store.push('post', {id: 1, comments: [1,2]})

2) we add one record to another's relationships f.e

    post.get('comments').addObject(comment)

So both situations happens when we manually linking records.
I think it should not create time bottlenecks on records creation

bmac:
:+1: Looks good to me. I'd like @igorT to review this code before its merged since he knows this part of the codebase the best.

packages/ember-data/lib/system/model/model.js
```diff
@@ -690,8 +685,8 @@ var Model = Ember.Object.extend(Ember.Evented, {
   */
   clearRelationships: function() {
     this.eachRelationship(function(name, relationship) {
-      var rel = this._relationships[name];
-      if (rel) {
+      if (this._relationships.has(name)) {
+        var rel = this._relationships.get(name);
```
igorT: Minor point but if it was never accessed, no point in creating it just in order to clear it

igorT:
@andrewfan thanks for the PR!
It seems like this would only help when creating records that have relationships but have no relationship data set on them. Did you see this speed up your app? Do you have many of those records? Did you manage to run the benchmarks to see what the improvements are?
An optimization that would be potentially much wider reaching would be to not initialize the many arrays before they are need/optimize belongsTo to not use a set. I would be happy to help with that.
I am just worried about adding this complexity without the speed improvement.

packages/ember-data/lib/system/model/model.js
andrewfan: if `has` method returns `true` it means that relationship was created before, so we wont create it accidentally

andrewfan:
> Did you see this speed up your app? 

Yes, probably in 6 times (i mean creating of record) on hugest model (6 belongsTo relationships, 8 hasMany relationships)
> Do you have many of those records?

Yes, we have about 60 models and nearly each of them has relationships, plus we are using http://jsonapi.org/format/#fetching-sparse-fieldsets API, so sometimes even huge models come with 1-2 non-relationship attributes. 
> Did you manage to run the benchmarks to see what the improvements are?

I ran the same code with this wrapper for many times:
   
    var start = Date.now();
    // creating record code 
    var end = Date.now() - start;

and then compared average time, in my case for the hugest model it differed in 6 times, but ofc everything depends on number of relationships in model.

andrewfan:
@bmac @igorT should i resolve conflicts?

packages/ember-data/lib/system/model/model.js
igorT: :+1:

igorT:
@andrewfan this makes a lot more sense after the internal model refactor was merged, so we rebased and merged as #3185 
we kept you as an author though. thanks!

PR #2755, opened 2015-02-04
Fix JS code block for coalesceFindRequests

wecc:
Minor fix for JS code block at http://emberjs.com/api/data/classes/DS.RESTAdapter.html#property_coalesceFindRequests


bmac:
Thanks!

PR #2756, opened 2015-02-05
Fix tests on canary. They broke when Ember.js removed the meta.descs

bmac:
property.

See https://github.com/emberjs/ember.js/pull/10323

cc @ebryn @igorT

### DIFF
--- a/packages/ember-data/tests/integration/relationships/belongs_to_test.js
+++ b/packages/ember-data/tests/integration/relationships/belongs_to_test.js
@@ -7,6 +7,21 @@ var hasMany = DS.hasMany;
 var belongsTo = DS.belongsTo;
 var hash = Ember.RSVP.hash;
 
+// Before https://github.com/emberjs/ember.js/pull/10323 the computed
+// property descriptor was stored on the ember meta object. After that
+// pr it was moved to the ember object. This code normalized that
+// lookup because the Ember Data ci tests run against diferent version
+// of Ember. Once that code reaches the release branch this code can
+// be removed.
+function getComputedPropertyDesc(model, key) {
+  if (Ember.meta(model).descs) {
+    return Ember.meta(model).descs[key];
+  }
+  var possibleDesc = model[key];
+  var desc = (possibleDesc !== null && typeof possibleDesc === 'object' && possibleDesc.isDescriptor) ? possibleDesc : undefined;
+  return desc;
+}
+
 module("integration/relationship/belongs_to Belongs-To Relationships", {
   setup: function() {
     User = DS.Model.extend({
@@ -362,15 +377,16 @@ test("relationshipsByName is cached in production", function() {
   var oldTesting = Ember.testing;
   //We set the cacheable to true because that is the default state for any CP and then assert that it
   //did not get dynamically changed when accessed
-  var oldCacheable = Ember.meta(model).descs.relationshipsByName._cacheable;
-  Ember.meta(model).descs.relationshipsByName._cacheable = true;
+  var relationshipsByName = getComputedPropertyDesc(model, 'relationshipsByName');
+  var oldCacheable = relationshipsByName._cacheable;
+  relationshipsByName._cacheable = true;
   Ember.testing = false;
   try {
     equal(get(model, 'relationshipsByName'), get(model, 'relationshipsByName'), 'relationshipsByName are cached');
     equal(get(model, 'relationshipsByName'), get(model, 'relationshipsByName'), 'relationshipsByName are cached');
   } finally {
     Ember.testing = oldTesting;
-    Ember.meta(model).descs.relationshipsByName._cacheable = oldCacheable;
+    relationshipsByName._cacheable = oldCacheable;
   }
 });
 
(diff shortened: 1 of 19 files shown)


bmac:
I'm not sure why the relationship integration tests are complaining about "run loop at end of test". They all are green for me locally in both the browser and the command line. I will investigate some more tonight.

wecc:
This has happened for me locally, running it again usually solves it. No idea why it's happening though. Force push?

bmac:
I've also hit this sporadic "run loop at end of test" error locally as well. But that doesn't seem to be the issue here as I've re-triggered the build several times and it seems to consistently be failing at the same spot on travis.

igorT:
cc @stefanpenner

packages/ember-data/tests/integration/relationships/belongs_to_test.js
```diff
@@ -7,6 +7,10 @@ var hasMany = DS.hasMany;
 var belongsTo = DS.belongsTo;
 var hash = Ember.RSVP.hash;
 
+function getComputedPropertyDesc(model, key) {
+  return Ember.meta(model).descs ? Ember.meta(model).descs[key] : model[key];
```
stefanpenner: the check is a bit more specific.

cc @ebryn I know this is violating a private API, but it is working around some annoying cross version issues and a serious performance hack. The proper solution is a hefty refactor that will happen sometime soon. Hence the triage fix.

bmac: @stefanpenner what should I test for to be more specific?

stefanpenner: https://github.com/emberjs/ember.js/pull/10323/files#diff-ca4a7728f6d3a705f5a202589e46d85cR123

bmac:
I've spent a few hours on this pr today and Thursday attempting to try down why travis is reporting a "run loop at end of test" in canary and I haven't made much progress. If any one has any theories or insight I'm all ears.

bmac:
Cleared the cache on travis and now everything is green.

packages/ember-data/tests/integration/relationships/belongs_to_test.js
```diff
@@ -7,6 +7,15 @@ var hasMany = DS.hasMany;
 var belongsTo = DS.belongsTo;
 var hash = Ember.RSVP.hash;
 
+function getComputedPropertyDesc(model, key) {
```
igorT: Can you add a comment here why we need to do this.

packages/ember-data/tests/integration/relationships/belongs_to_test.js
```diff
@@ -451,7 +463,7 @@ test("relationship changes shouldn’t cause async fetches", function() {
   env.adapter.deleteRecord = function(store, type, record) {
     ok(record instanceof type);
     equal(record.id, 1, 'should first comment');
-    return record;
+    return record.toJSON({ includeId: true });
```
igorT: Why do we need to do this change?

packages/ember-data/tests/integration/store_test.js
```diff
@@ -171,7 +171,6 @@ test("destroying the store correctly cleans everything up", function() {
   Ember.run(store, 'destroy');
 
   equal(car.get('person'), null, "expected car.person to no longer be present");
-  equal(person.get('cars'), undefined, "expected person.cars to be empty");
```
igorT: Why did we remove this?

packages/ember-data/tests/unit/record_array_test.js
```diff
@@ -49,6 +49,11 @@ test("stops updating when destroyed", function() {
   expect(3);
   var store = createStore();
 
+  // TODO remove once
+  // https://github.com/emberjs/ember.js/commit/c3f13e85a62069295965dd49ca487fe6ddba1188
+  // is on the release branch
+  var emptyLength = Ember.meta(store).descs ? undefined : 0;
```
igorT: Why is this related to meta.descs?

packages/ember-data/tests/integration/relationships/belongs_to_test.js
bmac: An adapter is suppose to return JSON not a record instance. Before the meta change this worked ok because if you tried to read all the keys out of a record instance you would get the id and undefined for every other key. Not that computed properties are stored directly on the record object instead of the meta object [this code](https://github.com/emberjs/data/blob/master/packages/ember-data/lib/system/store.js#L1788) gets the computed property descriptor which blows up later when it tries to normalize it into a relationship.

packages/ember-data/tests/integration/store_test.js
bmac: Computed properties are no longer removed when an object is destroy now that they are not stored on the meta object.

packages/ember-data/tests/unit/record_array_test.js
bmac: `length` on an ArrayProxy is a computed property and they are no longer removed when an object is destroy now that they are not stored on the meta object.

igorT: That seems crazyish?

packages/ember-data/tests/integration/store_test.js
igorT: Can you verify this? I tried it on jsbin and I couldn't find the change in behavior between release/canary.

igorT: Nm, @rwjblue and I verified there was a change in behavior

PR #2759, opened 2015-02-06
Added failing test for hasMany add/remove bug

slindberg:
When a record with a `hasMany` relationship has a record removed and another added, then is persisted, the added record does not end up in the relationship's canonical list. Nearest I can tell, it's because [`ManyRelationship#computeChanges`](https://github.com/emberjs/data/blob/v1.0.0-beta.14.1/packages/ember-data/lib/system/relationships/state/has_many.js#L108-L142) performs a remove, but never adds new records. However, my attempts at adding the records there result in another test failing indicating that relationships are duplicated on record create.  :confused: 

Here's a [jsbin](http://jsbin.com/fetoqomopi/1/edit?html,js,output) that demonstrates the issue.

Possibly related to #2666

### DIFF
--- a/packages/ember-data/tests/integration/adapter/rest_adapter_test.js
+++ b/packages/ember-data/tests/integration/adapter/rest_adapter_test.js
@@ -622,6 +622,37 @@ test("update - a serializer's primary key and attributes are consulted when buil
   }));
 });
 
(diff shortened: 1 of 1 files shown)


tonywok:
I think it's also the same as PR #2722

PR #2763, opened 2015-02-07
ManyArray.flushCanonical should use set method for currentState

MarSik:
I was getting an "should use set" exception from ManyArray's
flushCannonical. This patch fixes the issue and works properly
in my application.

### DIFF
--- a/packages/ember-data/lib/system/record_arrays/many_array.js
+++ b/packages/ember-data/lib/system/record_arrays/many_array.js
@@ -76,7 +76,7 @@ export default Ember.Object.extend(Ember.MutableArray, Ember.Evented, {
     toSet = toSet.concat(newRecords);
     this.arrayContentWillChange(0, this.length, this.length);
     this.set('length', toSet.length);
-    this.currentState = toSet;
+    this.set('currentState', toSet);
     this.arrayContentDidChange(0, this.length, this.length);
     //TODO Figure out to notify only on additions and maybe only if unloaded
     this.relationship.notifyHasManyChanged();


bmac:
@MarSik the fact that you are getting a "should use set" error likely means you have a `.get('currentState')` somewhere in your codebase. Since currentState is a private property and isn't meant to be observable I believe the current code is correct. 

@igorT knows this code best so he should make the final call on this pr.

MarSik:
I do not call get like that anywhere, but git grep on ember-data tells me:

packages/ember-data/lib/system/model/model.js:19:  return get(get(this, 'currentState'), key);
packages/ember-data/lib/system/model/model.js:518:    var currentState = get(this, 'currentState');
packages/ember-data/lib/system/model/model.js:537:    var currentState = get(this, 'currentState');
packages/ember-data/lib/system/model/states.js:39:  record.get('currentState.stateName');
packages/ember-data/lib/system/store.js:1187:      if (get(record, 'currentState.stateName') === 'root.deleted.saved') {

So I supposed that meant flushCanonical should have used set.

8. 2. 2015 v 17:47, Brendan McLoughlin <notifications@github.com>:

bmac:
It looks like all of those lines turned up by grep are acting on the [currentState](http://emberjs.com/api/data/classes/DS.Model.html#property_currentState) property of a Model instance and not the `currentState` property of a hasMany relationship.

MarSik:
Yes, you are right. I was not accessing currentState directly anywhere though. Except from the Ember Firefox plugin when I was trying to debug my application. That might have triggered it as it probably uses getters for displaying fields.

8. 2. 2015 v 20:19, Brendan McLoughlin <notifications@github.com>:

bmac:
I'm going to close this because it sounds like the original issue was caused by using the firefox plugin using a getter. Feel free to comment if there is still an issue and we can reopen.

PR #2766, opened 2015-02-09
Ember Data should delegate to the container to normalize the type key.

### LINKED ISSUE #2761: Custom Adapter fails to resolve in tests.
When executing tests, custom adapters with dashes fail to resolve and instead return the application adapter.

https://github.com/cah-briangantzler/ember-cli-custom-adapter.git

is a repo with a failing test demonstrating this.

Originally opened on ember-cli as [Issue 3178](https://github.com/ember-cli/ember-cli/issues/3178).

bmac:
Closes #2761 

cc @stefanpenner @igorT

### DIFF
--- a/packages/ember-data/lib/serializers/rest_serializer.js
+++ b/packages/ember-data/lib/serializers/rest_serializer.js
@@ -723,7 +723,7 @@ var RESTSerializer = JSONSerializer.extend({
     @param {Object} options
   */
   serializeIntoHash: function(hash, type, record, options) {
-    hash[type.typeKey] = this.serialize(record, options);
+    hash[this.typeForRoot(type.typeKey)] = this.serialize(record, options);
   },
 
   /**
--- a/packages/ember-data/lib/system/store.js
+++ b/packages/ember-data/lib/system/store.js
@@ -9,7 +9,6 @@ import {
   InvalidError,
   Adapter
 } from "ember-data/system/adapter";
-import { singularize } from "ember-inflector/system/string";
 import {
   Map
 } from "ember-data/system/map";
@@ -76,8 +75,6 @@ var map = Ember.EnumerableUtils.map;
 var Promise = Ember.RSVP.Promise;
 var copy = Ember.copy;
 var Store;
-
-var camelize = Ember.String.camelize;
 
 // Implementors Note:
 //
@@ -1777,7 +1774,11 @@ Store = Ember.Object.extend({
     @return {String} if the adapter can generate one, an ID
   */
   _normalizeTypeKey: function(key) {
-    return camelize(singularize(key));
+    // Delegate to the container for normalization. The container
+    // requires a ':' to normalize so we add `model:` then slice off
+    // the first 6 characters to remove the `model:` in the return
+    // value
+    return this.container.normalize('model:' + key).slice(6);
   }
 });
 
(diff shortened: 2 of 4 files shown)


packages/ember-data/lib/serializers/rest_serializer.js
```diff
@@ -723,7 +723,7 @@ var RESTSerializer = JSONSerializer.extend({
     @param {Object} options
   */
   serializeIntoHash: function(hash, type, record, options) {
-    hash[type.typeKey] = this.serialize(record, options);
+    hash[this.typeForRoot(type.typeKey)] = this.serialize(record, options);
```
bmac: We should make a `rootForType` function to generate this root namespace based on the type key.

packages/ember-data/lib/system/store.js
```diff
@@ -1777,7 +1774,11 @@ Store = Ember.Object.extend({
     @return {String} if the adapter can generate one, an ID
   */
   _normalizeTypeKey: function(key) {
-    return camelize(singularize(key));
+    // Delegate to the container for normalization. The container
+    // requires a ':' to normalize so we add `model:` then slice off
+    // the first 6 characters to remove the `model:` in the return
+    // value
+    return this.container.normalize('model:' + key).slice(6);
```
bmac: @stefanpenner Is there a better way go get the normalized value for `key`?

bmac:
cc @mixonic @igorT pointed out that this looks like the inverse of this change you made https://github.com/bmac/data/commit/766f0c9a8cdc8d21b99d80378c01678f8c404aa3 can you review this pr as well?

tomdale:
Instead of delegating to the resolver, let's make dasherized the canonical form. To help Ember Data users migrate if they are using the camelized form, can we detect capital letters (perhaps using a Regular Expression™) and fall back to the camelized behavior but issue a deprecation warning? @bmac

packages/ember-data/lib/system/store.js
stefanpenner: not currently :(

stefanpenner:
> Instead of delegating to the resolver, let's make dasherized the canonical form.

this would be lovely, but can be transition globals users from it as well?
I wonder if we should still send it through the resolver, which can add those validations you speak but globally for all resolved entities.

mixonic:
Fwiw I think this PR was on the right track. @igorT asked me to review it in detail and I left no comments. My bad.

A few thoughts on this mess:

* I don't think it matters what the format of typeKey is, or who/what decides it. It does matter that the typeKey is *consistent*.
* I do not think the format of typeKey should be coupled to the format a user users. If a user types `find('fooBar')` it should not matter if the typeKey is dasherized, camelcased, or decided by the container. It should Just Work.
  * This means the container setting the typeKey (if we do that) should have no impact on a user transitioning between globals/ember-cli, and that all documentation can use one format for multiple words that works everywhere.
* The places we have been bit this have been where user-input (whether via `find` or `hasMany` or any other function) is trusted to have the correct formatting. We should mitigate this, perhaps with a `model.isType(userInputString)` function or another method that removes direct typeKey string comparisons from the internal code (`isType` can perform normalization).

stefanpenner:
> Fwiw I think this PR was on the right track

yes

PR #2771, opened 2015-02-10
Fix hasMany simultaneous adds and removes bug

bmac:
Includes @slindberg test case from #2759.

cc @igorT

### DIFF
--- a/packages/ember-data/lib/system/model/states.js
+++ b/packages/ember-data/lib/system/model/states.js
@@ -417,6 +417,11 @@ createdState.uncommitted.rollback = function(record) {
   record.transitionTo('deleted.saved');
 };
 
+createdState.uncommitted.pushedData = function(record) {
+  record.transitionTo('loaded.updated.uncommitted');
+  record.triggerLater('didLoad');
+};
+
 createdState.uncommitted.propertyWasReset = Ember.K;
 
 function assertAgainstUnloadRecord(record) {
--- a/packages/ember-data/lib/system/relationships/state/has_many.js
+++ b/packages/ember-data/lib/system/relationships/state/has_many.js
@@ -128,8 +128,6 @@ ManyRelationship.prototype.computeChanges = function(records) {
 
   this.removeCanonicalRecords(recordsToRemove);
 
-  var hasManyArray = this.manyArray;
-
   // Using records.toArray() since currently using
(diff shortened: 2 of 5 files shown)


packages/ember-data/lib/system/model/states.js
```diff
@@ -417,6 +417,10 @@ createdState.uncommitted.rollback = function(record) {
   record.transitionTo('deleted.saved');
 };
 
+createdState.uncommitted.pushedData = function(record) {
+  record.transitionTo('loaded.saved');
+};
+
```
slindberg: Interesting. I'm curious why this doesn't end up putting records created with data directly into the `loaded.saved` state, since `buildRecord` should call `pushedData`, which (seems to) call this transition. I'm actually unsure when this transition get invoked at all since the only two paths I can identify are the the one above and `push` => `_load` => `setupData` => `pushedData`, which shouldn't affect the created state.

Damn the `DS.Model` state machine is difficult to reason about :stuck_out_tongue_closed_eyes:

bmac: Looks like inside `buildRecord` the newly created record is in the `root.loading` state so this code doesn't get hit.

slindberg: Heh, all my tests show that records inside `buildRecord` leave in the `root.empty` state... which means that your hook won't be invoked anyway. On a related note, it doesn't look like `buildRecord` is ever invoked with its third param, so [this line](https://github.com/emberjs/data/blob/v1.0.0-beta.14.1/packages/ember-data/lib/system/store.js#L1602) will never get executed anyway.

packages/ember-data/lib/system/model/states.js
```diff
@@ -417,6 +417,12 @@ createdState.uncommitted.rollback = function(record) {
   record.transitionTo('deleted.saved');
 };
 
+createdState.uncommitted.pushedData = function(record) {
+  record.transitionTo('loaded.saved');
```
bmac: `loaded.saved` may be the wrong state here. If there are unsaved changes in `_attributes` we likely want to transition to `loaded.updated.uncommitted`

bmac: We need a test for this change.

PR #2775, opened 2015-02-11
Adds support for using mixins in polymorphic relationships

igorT:
Now it's possible to reference mixins in polymorphic relationships.

For example:

 ```
  var Comment = DS.Model.extend({
     owner: belongsTo('commentable'. { polymorphic: true })
  });
  var Commentable = Ember.Mixin.create({
    comments: hasMany('comment')
  });
```

### DIFF
--- a/packages/ember-data/lib/system/relationships/state/has_many.js
+++ b/packages/ember-data/lib/system/relationships/state/has_many.js
@@ -86,13 +86,13 @@ ManyRelationship.prototype.removeRecordFromOwn = function(record, idx) {
 ManyRelationship.prototype.notifyRecordRelationshipAdded = function(record, idx) {
   var type = this.relationshipMeta.type;
   Ember.assert("You cannot add '" + record.constructor.typeKey + "' records to the " + this.record.constructor.typeKey + "." + this.key + " relationship (only '" + this.belongsToType.typeKey + "' allowed)", (function () {
-    if (record instanceof type) {
-      return true;
-    } else if (Ember.MODEL_FACTORY_INJECTIONS) {
-      return record instanceof type.superclass;
+    if (type.__isMixin) {
+      return type.__mixin.detect(record);
     }
-
-    return false;
+    if (Ember.MODEL_FACTORY_INJECTIONS) {
+      type = type.superclass;
+    }
+    return record instanceof type;
   })());
 
   this.record.notifyHasManyAdded(this.key, record, idx);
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

**Comment by lolmaus** [comment]
That's awesome, thank you!

**Comment by quaertym** [comment]
@igorT Thank you, I was waiting this for ages.

File: packages/ember-data/lib/system/relationships/state/has_many.js [hunk 1]
```diff
@@ -86,13 +86,14 @@ ManyRelationship.prototype.removeRecordFromOwn = function(record, idx) {
...
   Ember.assert("You cannot add '" + record.constructor.typeKey + "' records to the " + this.record.constructor.typeKey + "." + this.key + " relationship (only '" + this.belongsToType.typeKey + "' allowed)", (function () {
-    if (record instanceof type) {
-      return true;
-    } else if (Ember.MODEL_FACTORY_INJECTIONS) {
-      return record instanceof type.superclass;
+    if (Ember.MODEL_FACTORY_INJECTIONS) {
+      type = type.superclass;
+    }
+    if (type.__isMixin) {
+      type = type.superclass;
```
igorT: should just use the mixin instead of superclass hack

MiracleBlue:
This is the thing I've been needing all my life.  Well.. for the last 3 weeks of it anyway.  Thank you so much.

sheldonnbbaker:
I'm trying to use this but consistently get an error (1.0.0-beta.17) - `findPossibleInverses(this, inverseType)` returns `undefined`. Specifying inverses manually simple causes another related error.

```
// activity.js
  trackable: belongsTo('trackable', { polymorphic: true })
});

// trackable.js
export default Ember.Mixin.create({
  activity: belongsTo('activity')
});
```

igorT:
Can you post a JSBin tha reproduces the issue?

sheldonnbbaker:
[Here it is](http://emberjs.jsbin.com/pohukineve/1/edit). Thanks!

igorT:
@sheldonnbbaker if you are using globals renaming it to `App.TrackableMixin` works. Not sure whether this is a bug on our side or not.

sheldonnbbaker:
I'm using modules with ember-cli, and couldn't get renaming the mixin to work. Also tried playing with `MODEL_FACTORY_INJECTIONS` but no dice.

Do you think it could be something with the resolver - or should I just focus on ember-data code if I wanted to fix this?

igorT:
is you mixin in the mixins folder? In '/mixins/trackable' ? In the JSBin it seems to work.

sheldonnbbaker:
Ah, that did it! Many thanks!

igorT:
Where did you place it before? Is there anything in the docs that could've helped you avoid this mistake?

sheldonnbbaker:
I assumed that where I saved mixins had no bearing on the final output, so long as I `import`ed them correctly. Perhaps just another section on [this page](http://guides.emberjs.com/v1.11.0/models/defining-models/) about where to save mixins would help?

PR #2781, opened 2015-02-12
Unload record instead of dematerialize

### LINKED ISSUE #2780: Record arrays should remove dematerialize records
https://github.com/emberjs/data/pull/2467#issuecomment-72606486

wecc:
- Use store.unloadRecord() in favor of store.dematerializeRecord() (Fixes #2780)
- Deprecate store.dematerializeRecord()

### DIFF
--- a/packages/ember-data/lib/system/model/states.js
+++ b/packages/ember-data/lib/system/model/states.js
@@ -671,7 +671,7 @@ var RootState = {
 
       setup: function(record) {
         var store = get(record, 'store');
-        store.dematerializeRecord(record);
+        store._dematerializeRecord(record);
       },
 
       invokeLifecycleCallbacks: function(record) {
--- a/packages/ember-data/lib/system/store.js
+++ b/packages/ember-data/lib/system/store.js
@@ -1912,7 +1912,7 @@ function _find(adapter, store, type, id, record) {
     if (record) {
       record.notFound();
       if (get(record, 'isEmpty')) {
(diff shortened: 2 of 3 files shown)


wecc:
Since `dematerializeRecord()` was never intended for public use, should we just assert instead of deprecate + proxy to `_dematerializeRecord`?

sly7-7:
@wecc Do you think we need a test before merging, to make sure it fixes https://github.com/emberjs/data/pull/2467#issuecomment-72606486 ?

wecc:
@sly7-7 There's already a test to make sure record arrays are updated when records are unloaded: https://github.com/emberjs/data/blob/master/packages/ember-data/tests/integration/records/unload_test.js#L77

sly7-7:
Sorry, I should have known you already looked at this :smile: 
So, maybe the only missing thing is to update https://github.com/emberjs/data/blob/master/packages/ember-data/tests/integration/adapter/find_test.js#L124 and using `unload` rather than `dematerialize`

wecc:
@sly7-7 nice catch, test name updated!

bmac:
@sly7-7 I see you added the Needs Discussion label. What concerns do you have with this pr that you think should be discussed?

sly7-7:
@bmac sorry for the delay, I labelled it after @wecc told me there was a discussion about the deprecation vs error on the private _dematarializeRecord method. Since it's merged, I think deprecation is fine for everyone :+1:

PR #2785, opened 2015-02-13
An unloaded belongsTo should be 'refetched' if it was unloaded

sly7-7:
Currently, [the seat](https://github.com/sly7-7/data/compare/async-belongsTo-should-refetch-unloaded-record?expand=1#diff-c666930f1c4d7d319d55084a86985ae3R306) is null.


sly7-7:
Ok my bad, I realize what I want here makes no sense at all. Maybe a belongsTo relationship could have a reload() method. Anyway, closing this. Sorry for the noise.

PR #2790, opened 2015-02-15
Embedded records mixin should use the correct serialization key when deserialize configuration is set, Fixes #2556

### LINKED ISSUE #2556: EmbeddedRecordsMixin issue with belongsTo deserialize: 'records' + serialize: 'id'
When you want to load embedded records but save relation with id on a model with the ActiveRecord adapter, the EmbeddedRecordsMixin seems to break the expected behaviour of ActiveRecord.

It should serialize `provider_id: 3` but it serialize `provider: 3` and it seems to be because it only relies ont the `deserialize` attr to set the key during the serialize process.

Specifically, it seems to be because the [`EmbeddedRecordsMixin#serializeBelongsTo][l1] method rely on [`EmbeddedRecordsMixin#keyForRelationship`][l2] method which in turn rely on the [`EmbeddedRecordsMixin#hasDeserializeRecordsOption`][l3] option value to set the key for the relation. But the key set during the serialization should solely depend on the `serialize` option, not the `deserialize`.

See demo here : http://emberjs.jsbin.com/teweposowu

cc @jbourassa

[l1]: https://github.com/emberjs/data/blob/v1.0.0-beta.12/packages/ember-data/lib/serializers/embedded_records_mixin.js#L183-L194

[l2]: https://github.com/emberjs/data/blob/v1.0.0-beta.12/packages/ember-data/lib/serializers/embedded_records_mixin.js#L126-L132

[l3]: https://github.com/emberjs/data/blob/v1.0.0-beta.12/packages/ember-data/lib/serializers/embedded_records_mixin.js#L374

agrobbin:
This is my first attempted contribution to Ember Data, so if I'm approaching anything the wrong way, please let me know!

When extending the `DS.EmbeddedRecordsMixin`, if you configure a `belongsTo` relationship as `{ serialize: 'id', deserialize: 'records' }`, it won't actually pay attention to the `serialize: 'id'` option, but use the `deserialize: 'records'` option. This is because of [`serializeBelongsTo`](https://github.com/emberjs/data/blob/f1be2af71d7402d034bc034d9502733647cad295/packages/ember-data/lib/serializers/embedded_records_mixin.js#L194) and the reimplementation of [`keyForRelationship`](https://github.com/emberjs/data/blob/f1be2af71d7402d034bc034d9502733647cad295/packages/ember-data/lib/serializers/embedded_records_mixin.js#L126-L132), which notices the `deserialize` option instead of the `serialize` option.

Fixes #2556.

### DIFF
--- a/packages/activemodel-adapter/lib/system/active-model-serializer.js
+++ b/packages/activemodel-adapter/lib/system/active-model-serializer.js
@@ -262,7 +262,7 @@ var ActiveModelSerializer = RESTSerializer.extend({
       type.eachRelationship(function(key, relationship) {
         var payloadKey, payload;
         if (relationship.options.polymorphic) {
-          payloadKey = this.keyForAttribute(key);
+          payloadKey = this.keyForAttribute(key, "deserialize");
           payload = hash[payloadKey];
           if (payload && payload.type) {
             payload.type = this.typeForRoot(payload.type);
@@ -273,7 +273,7 @@ var ActiveModelSerializer = RESTSerializer.extend({
             });
           }
         } else {
-          payloadKey = this.keyForRelationship(key, relationship.kind);
+          payloadKey = this.keyForRelationship(key, relationship.kind, "deserialize");
           if (!hash.hasOwnProperty(payloadKey)) { return; }
           payload = hash[payloadKey];
         }
--- a/packages/ember-data/lib/serializers/embedded-records-mixin.js
+++ b/packages/ember-data/lib/serializers/embedded-records-mixin.js
@@ -123,11 +123,11 @@ var EmbeddedRecordsMixin = Ember.Mixin.create({
     return extractEmbeddedRecords(this, this.store, type, normalizedHash);
   },
 
-  keyForRelationship: function(key, type) {
-    if (this.hasDeserializeRecordsOption(key)) {
-      return this.keyForAttribute(key);
+  keyForRelationship: function(key, type, method) {
+    if ((method === 'serialize' && this.hasSerializeRecordsOption(key)) || (method === 'deserialize' && this.hasDeserializeRecordsOption(key))) {
+      return this.keyForAttribute(key, method);
     } else {
-      return this._super(key, type) || key;
+      return this._super(key, type, method) || key;
     }
   },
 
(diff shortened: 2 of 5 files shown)


joeruello:
Hit this today, thanks for the fix!

packages/ember-data/lib/serializers/embedded_records_mixin.js
```diff
@@ -123,8 +123,10 @@ var EmbeddedRecordsMixin = Ember.Mixin.create({
     return extractEmbeddedRecords(this, this.store, type, normalizedHash);
   },
 
-  keyForRelationship: function(key, type) {
-    if (this.hasDeserializeRecordsOption(key)) {
+  keyForRelationship: function(key, type, method) {
+    if (method === undefined) { method = 'deserialize'; }
```
bmac: I think the convention is to put the conditional body on its own line so its easier to inset a breakpoint if needed,

bmac:
For symmetry, I'd like to see the 'deserialize' argument included on the `keyForRelationship` relationship call at line https://github.com/agrobbin/ember-data/blob/2556-embedded-records-mixin-serialize-ids/packages/ember-data/lib/serializers/embedded_records_mixin.js#L340

bmac:
@pixelhandler and @igorT would one of you like to review this pr since you have more familiarity with the EmbeddedMixin and (presumably) with ActiveRecord?

pixelhandler:
@bmac I can take a look in the next day or two

agrobbin:
@bmac done and done!

packages/ember-data/lib/serializers/embedded_records_mixin.js
```diff
@@ -123,8 +123,12 @@ var EmbeddedRecordsMixin = Ember.Mixin.create({
     return extractEmbeddedRecords(this, this.store, type, normalizedHash);
   },
 
-  keyForRelationship: function(key, type) {
-    if (this.hasDeserializeRecordsOption(key)) {
+  keyForRelationship: function(key, type, method) {
+    if (method === undefined) {
```
igorT: This isn't needed anymore right?

igorT:
Thanks for the PR! Looks pretty good. If we are gonna add 'serialize'/'deserialize', could you do it throughout all the serializers and not just the EmbeddedRecords one?

agrobbin:
@igorT I can certainly do that, but won't that cause jsHint warnings about unused arguments in the `keyForRelationship` function definitions when it happens to not be used?

agrobbin:
@igorT or were you thinking of just adding the argument to the call-sites and since we only happen to use it in the `EmbeddedRecordsMixin` mixin, leave the other `keyForRelationship` definitions alone?

agrobbin:
@igorT I went with the latter. Let me know if I can do anything else!

agrobbin:
@igorT is there anything remaining to get this in before beta 16 goes out?

bmac:
@agrobbin This issue was discussed in the last Ember Data meeting and it was decided that if this change is made for `keyForRelationship` the method param should also be passed into `keyForAttribute` as well.

agrobbin:
@bmac :+1: just pushed up a second commit to this branch, happy to squash them into 1 assuming you like how this looks!

pixelhandler:
Seems good :+1: 

To support getting the right key, it seems necessary to know whether the key will be used to serialize or deserialize, nice.

agrobbin:
Squashed!

igorT:
Thank you @agrobbin

agrobbin:
No problem!

PR #2796, opened 2015-02-16
[DOC] Clarify when didCreate is fired

joeruello:
Previous doc was ambiguous and could be interpreted to mean when the record is created using createRecord(). Clarifies that the event is fired when a new record is saved.


bmac:
Thanks

PR #2798, opened 2015-02-18
update hasRecordForId to check model state

jordpo:
return false when id is present but record empty from an ED relationship


sly7-7:
I think this is something like [@tomdale's comment](https://github.com/emberjs/data/pull/2659#issuecomment-72246238), but checking for the empty state instead of loaded state. 
This will need tests too.
cc/ @igorT

sly7-7:
closing as duplicate of https://github.com/emberjs/data/pull/2659/

PR #2803, opened 2015-02-19
Mutating a `DS.attr()` array causes dirtying

seanpdoyle:
This is currently a failing test case.

Is this missing behavior a bug, or is it deliberate, since the attribute is
arbitrary?

The canonical use case:

```js
App.Post = DS.Model.extend({
 tags: DS.attr({ defaultValue: [] })
});

// current behavior var post = this.store.createRecord("post");

post.get("tags").pushObject("emberjs"); post.get("isDirty") // false

post.get("tags").setObjects(["ember-data"]); post.get("isDirty") // false

post.set("tags", ["emberconf"]); post.get("isDirty") // true

// new behavior var post = this.store.createRecord("post");

post.get("tags").pushObject("emberjs"); post.get("isDirty") // true

post.get("tags").setObjects(["ember-data"]); post.get("isDirty") // true

post.set("tags", ["emberconf"]); post.get("isDirty") // true
```

### DIFF
--- a/packages/ember-data/tests/unit/model_test.js
+++ b/packages/ember-data/tests/unit/model_test.js
@@ -10,7 +10,8 @@ module("unit/model - DS.Model", {
 
     Person = DS.Model.extend({
       name: DS.attr('string'),
-      isDrugAddict: DS.attr('boolean')
+      isDrugAddict: DS.attr('boolean'),
+      addictions: DS.attr({ defaultValue: [] })
     });
   },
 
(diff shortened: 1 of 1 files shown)


seanpdoyle:
@bmac where do we stand on this? Is it something I should look into implementing or is it a Bad Idea :tm:

bmac:
@seanpdoyle If this is something you'd like to attempt to implement as an addon then go for it! (and please share). For Ember Data at the moment @igorT is worried about how deep into the array Ember Data would need to track changes (e.g. what should happen if a property of an object inside an array changes?).

seanpdoyle:
@bmac I agree, it couldget pretty hairy.

Closing this to reconsider.

Issue #2802, opened 2015-02-19
BelongsTo wrong type assertion should be more clear


igorT:
The hasMany one is much nicer. We should make the belongsTo one follow it's format.

omarismail:
Hey @igorT, I would like to work on this. Could I perhaps have more direction?

wecc:
@omarismail I think it's [this](https://github.com/emberjs/data/blob/v1.0.0-beta.15/packages/ember-data/lib/system/relationships/state/belongs_to.js#L63) vs. [this](https://github.com/emberjs/data/blob/v1.0.0-beta.15/packages/ember-data/lib/system/relationships/state/has_many.js#L88).

omarismail:
thanks @wecc, working on it now.

bmac:
Closing because it looks like this was fixed by #2810. Feel free to reopen if I'm wrong.

PR #2810, opened 2015-02-20
improve error message for belongsTo

omarismail:
Improved the error message for ensuring belongsTo has proper relationship. 

This is for #2802


omarismail:
@igorT what do you think?

wecc:
I think the property part is missing, right now we're getting:
_You cannot add a 'comment' record to the **'post'**. You can only add a 'user' record to this relationship'_

We probably want to add the property (like in the hasMany assert):
_You cannot add a 'comment' record to **'post.user'**. You can only add a 'user' record to this relationship'_

omarismail:
@wecc Thanks! Updated

sly7-7:
@omarismail Thank you :+1: 
Before merging, Could complete the assertion in the related tests please ? https://github.com/emberjs/data/blob/master/packages/ember-data/tests/integration/relationships/belongs_to_test.js#L131

omarismail:
@sly7-7 all set!

fivetanley:
Thanks!

PR #2812, opened 2015-02-21
Move buildURL and related methods to a mixin

amiel:
This is my first step towards emberjs/rfcs#4.

The intent here is to separate url building from the RESTAdapter so that
it can eventually be included in other adapters and hold the logic for the
path template concept described in emberjs/rfcs#4.

Even though it's a fairly small change, I've written this pull request to:

1. Get my foot in the door. Since this is my first pull request to ember-data, I figure starting small is a good idea.
2. Make a place for the url-templates feature before starting a longer-lived branch. I think this will help prevent conflict management overhead as I work on the url-templates feature.

### DIFF
--- a/packages/ember-data/lib/adapters/build_url_mixin.js
+++ b/packages/ember-data/lib/adapters/build_url_mixin.js
@@ -1,0 +1,111 @@
+var get = Ember.get;
+
+export default Ember.Mixin.create({
+  /**
+    Builds a URL for a given type and optional ID.
+
+    By default, it pluralizes the type's name (for example, 'post'
+    becomes 'posts' and 'person' becomes 'people'). To override the
+    pluralization see [pathForType](#method_pathForType).
+
+    If an ID is specified, it adds the ID to the path generated
+    for the type, separated by a `/`.
+
+    @method buildURL
+    @param {String} type
+    @param {String} id
+    @param {DS.Model} record
+    @return {String} url
+  */
+  buildURL: function(type, id, record) {
+    var url = [];
+    var host = get(this, 'host');
+    var prefix = this.urlPrefix();
+
+    if (type) { url.push(this.pathForType(type)); }
+
+    //We might get passed in an array of ids from findMany
+    //in which case we don't want to modify the url, as the
+    //ids will be passed in through a query param
+    if (id && !Ember.isArray(id)) { url.push(encodeURIComponent(id)); }
+
+    if (prefix) { url.unshift(prefix); }
+
+    url = url.join('/');
+    if (!host && url) { url = '/' + url; }
+
+    return url;
+  },
+
+  /**
+    @method urlPrefix
+    @private
+    @param {String} path
+    @param {String} parentUrl
+    @return {String} urlPrefix
+  */
+  urlPrefix: function(path, parentURL) {
+    var host = get(this, 'host');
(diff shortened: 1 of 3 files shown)


fivetanley:
This seems great to me! I wonder how this will affect the website. I think we still want the methods to show up on the documentation for DS.Model.

Do you have ruby installed? If so can you try building this in the website repo and post some screenshots of the doc output?

amiel:
I do have Ruby installed, but have never generated the API docs before. I'll try following the instructions at https://github.com/emberjs/website/blob/master/README.md and post the results.

bmac:
Should we be exporting the `BuildUrlMixin` on the `DS` global?

bmac:
@amiel I suspect @fivetanley is correct and the `buildURL` and `urlPrefix` will not show up on the `RESTAdapter` class in the docs. I you will likely need to define a `BuildUrlMixin` for your new mixin and use the YUIDoc `@uses` syntax to show that the `RESTAdapter` mixes in the `BuildUrlMixin`'s methods.

amiel:
Hmmm, I can't get the docs to generate:

```
$ bundle exec rake generate_docs
Generating docs data from /Users/amiel/src/github/ember/ember.js... Could not locate Gemfile or .bundle/ directory
You are missing dependencies for /Users/amiel/src/github/ember/ember.js. Attempting to install now.
bundle install
Could not locate Gemfile or .bundle/ directory
rake aborted!
```

Should there be a `Gemfile` in the `ember.js` repo?

I'm using ruby 2.1.0.

amiel:
@bmac @fivetanley I've tried for a while to generate the API docs to verify. Since I couldn't get anything working, I've just made the changes you suggested.

I wasn't going to export `BuildURLMixin` on the `DS` global initially, but if it helps the docs, I'm all for it.

packages/ember-data/lib/adapters/build_url_mixin.js
```diff
@@ -0,0 +1,111 @@
+var get = Ember.get;
+
+export default Ember.Mixin.create({
```
bmac: This will need a doc block at the top with an `@class BuildURLMixin` annotation for the methods below to get correctly attributed to the correct class.

packages/ember-data/lib/main.js
```diff
@@ -50,6 +50,7 @@ import {
   RESTAdapter,
   FixtureAdapter
 } from "ember-data/adapters";
+import BuildURLMixin from "emmber-data/adapters/build_url_mixin";
```
bmac: `emmber` -> `ember`

amiel: OMG, so embarrassing

packages/ember-data/lib/adapters/build_url_mixin.js
amiel: @bmac ok, I've added this, and some documentation for how to use `BuildURLMixin`. It is not very thorough, but my goal is to completely change the way url building is done (with url templates), and I don't want to spend too much time improving documentation for something that will soon be obsolete.

amiel:
Ok, I've got the docs building now, thanks @rwjblue.

Here's what it looks like now:

screen shot 2015-03-04 at 10 02 03 am

screen shot 2015-03-04 at 10 01 34 am

PR #2815, opened 2015-02-24
Declare `store` property on DS.Model

seanpdoyle:
Without explicitly declaring `store` on a `DS.Model` instance,
[`setUnknownProperty`][unknown] will be invoked, breaking any
user-overridden behavior.

[unknown]: https://github.com/emberjs/ember.js/blob/c9af09524a191ba155d98de8b9bbabaf802245e2/packages/ember-runtime/lib/system/core_object.js#L157-L159

### DIFF
--- a/packages/ember-data/lib/system/model/model.js
+++ b/packages/ember-data/lib/system/model/model.js
@@ -74,6 +74,9 @@ function mergeAndReturnChangedKeys(original, updates) {
 var Model = Ember.Object.extend(Ember.Evented, {
   _recordArrays: undefined,
   _relationships: undefined,
+
+  store: null,
+
   /**
     If this property is `true` the record is in the `empty`
     state. Empty is the first state all records enter after they have
--- a/packages/ember-data/tests/unit/model_test.js
+++ b/packages/ember-data/tests/unit/model_test.js
@@ -283,6 +283,26 @@ test("a DS.Model can have a defaultValue", function() {
   equal(get(tag, 'name'), null, "null doesn't shadow defaultValue");
 });
 
+test("a DS.model can define 'setUnknownProperty'", function() {
+  var tag;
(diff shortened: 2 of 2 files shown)


seanpdoyle:
For the time being, users can create an initializer with the following workaround:

```js
DS.Model.reopen({
  store: null
});
```

fivetanley:
I'm not sure what the value add of this commit is. Also, when records are instantiated by the store the store attaches itself on the model.

seanpdoyle:
@fivetanley if you remove the

```js
App.Post = DS.Model.extend({
  store: null,
  container: null
});
```

From http://emberjs.jsbin.com/dicera/2/edit?html,js,output, you'll see the errors, as 

https://github.com/emberjs/ember.js/blob/c9af09524a191ba155d98de8b9bbabaf802245e2/packages/ember-runtime/lib/system/core_object.js#L157-L159 will be triggered with the keys `"store"` and `"container"` as `key` arguments

seanpdoyle:
Arguably, the `store` and `container` keys could be left to models overriding `unknownProperty` and `setUnknownProperty`, but setting the keys to `null` initially in the `DS.Model` declaration would clear things up a bit

stefanpenner:
this actually indicates store is being set to late.

stefanpenner:
https://github.com/emberjs/data/blob/master/packages/ember-data/lib/system/store.js#L1410 shouldn't be causing what you demonstate hmm..

stefanpenner:
oh, i mis-read that. Thats the stupid put the store on the factory code.... Ug we gotta burn that

stefanpenner:
i am totally on-crack, this is the correct solution.

stefanpenner:
This is the correct solution, sorry about that apparently I am on crack.

Issue #2818, opened 2015-02-25
Expose the store as a Ember.Service


martndemus:
It would be very neat to be able to do this:
```javascript
App.FooObject = Ember.Object.extend({
  store: Ember.inject.service(),
  getSomething: function() {
    return this.get('store').find('something');
  }
});
```

Kuzirashi:
+1 on that. It would be nicer than using `targetObject.store` in components.

rwjblue:
Yes, I would like to be able to do this also.  I've discussed this previously with @tomdale, and I believe he is :+1: on this conceptually.  I am not sure what will be required to make it work properly though (haven't spent time looking into it much).

quaertym:
:+1:  This would be great if added.

martndemus:
I just played around get it to work, but this works for me:

```javascript
// initializers/store-as-a-service.js
export default {
  name: 'store-as-a-service',
  after: 'ember-data',
  initializer: function(container, application) {
    var store = container.lookup('store:main')
    container.register('service:store', store, { instantiate: false });
  }
}

bmac:
@martndemus Would you like to submit a pr adding similar logic to the end of this file?
https://github.com/emberjs/data/blob/master/packages/ember-data/lib/initializers/store.js#L36

martndemus:
Done, see #2820

bmac:
Fixed by #2820.

PR #2820, opened 2015-02-25
Register the store as a service

martndemus:
I propably want to add a test.
And also people want to feedback on this.

#2818

### DIFF
--- a/packages/ember-data/lib/initializers/store.js
+++ b/packages/ember-data/lib/initializers/store.js
@@ -33,5 +33,6 @@ export default function initializeStore(container, application) {
 
   // Eagerly generate the store so defaultStore is populated.
   // TODO: Do this in a finisher hook
-  container.lookup('store:main');
+  var store = container.lookup('store:main');
+  container.register('service:store', store, { instantiate: false });
 }
--- a/packages/ember-data/lib/system/store.js
+++ b/packages/ember-data/lib/system/store.js
@@ -99,6 +99,11 @@ var Store;
 
 var camelize = Ember.String.camelize;
 
+var Service = Ember.Service;
+if (!Service) {
+  Service = Ember.Object;
+}
+
 // Implementors Note:
 //
 //   The variables in this file are consistently named according to the following
@@ -189,9 +194,9 @@ function coerceId(id) {
 
   @class Store
   @namespace DS
-  @extends Ember.Object
+  @extends Ember.Service
 */
-Store = Ember.Object.extend({
+Store = Service.extend({
 
   /**
     @method init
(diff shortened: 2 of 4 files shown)


packages/ember-data/lib/initializers/store.js
```diff
@@ -33,5 +33,6 @@ export default function initializeStore(container, application) {
 
   // Eagerly generate the store so defaultStore is populated.
   // TODO: Do this in a finisher hook
-  container.lookup('store:main');
+  var store = container.lookup('store:main');
+  container.register('service:store', store,  { instantiate: false });
```
bmac: Travis failed because there are 2 spaces after `store,` which made jscs unhappy.

martndemus: I'll fix it :)

martndemus:
The Ember used in the integration tests is 1.9.1, so I can't add a test to check if the actual injection works. (Though it works on my application with Ember 1.10.0)

bmac:
On travis tests get run against the latest release, beta and canary channels. If you are testing locally you can change the ember version using the `emberchannel` query parameter (e.g. `http://localhost:4200/?hidepassed&emberchannel=beta` or `http://localhost:4200/?hidepassed&emberchannel=canary`).

It should be ok to wrap the test in a check for the existence of `Ember.inject` or `Ember.inject.service` that way the test will only get run in the versions of Ember that support that API. We can remove the check later once `Ember.inject.service` makes it into a release version of Ember.

martndemus:
@bmac Thanks for the pointer!

bmac:
This looks good to me. @martndemus do you mind squashing your commits?

@tomdale would you like to review this pr?

martndemus:
@bmac I squashed the commits.

slindberg:
Yay!

It might also make sense to [extend `DS.Store`](https://github.com/emberjs/data/blob/v1.0.0-beta.15/packages/ember-data/lib/system/store.js#L174) from [`Ember.Service`](https://github.com/emberjs/ember.js/blob/v1.11.0-beta.4/packages/ember-runtime/lib/system/service.js#L38) if it is available.

bmac:
@martndemus I chatted with @tomdale at Ember Conf and he is on board with this change. I also like @slindberg's suggestion. Could you extend the store form `Ember.Service` when available?

packages/ember-data/lib/system/store.js
```diff
@@ -99,6 +99,11 @@ var Store;
 
 var camelize = Ember.String.camelize;
 
+var Service = Ember.Service;
+if (!Service) {
+  Service = Ember.Object;
+}
+
```
martndemus: This was necessary to be backward compatible with  `ember-data <= 1.9.1`

bmac:
Thanks @martndemus!

PR #2821, opened 2015-02-25
Change _normalizeTypeKey to defer normalization to the container

mixonic:
Let's talk about Ember-Data model resolution works in this PR.

First, remember that [isolated containers have no normalization](https://github.com/emberjs/ember.js/blob/5d853cefaec16a58af72f380bbc927474eb88c63/packages/container/lib/registry.js#L314-L323). When testing with isolated containers (unit tests), **the user-entered model name becomes the factory name requested from the container as well as the typeKey**.

For example, in a unit test modelFor works like this:

* `modelFor("SomeItem")`
* Looks for `"model:SomeItem"` in the container
* This is normalized to `"model:SomeItem"` (unit tests have no
  normalization)
* The factory is returned. In Ember-CLI, an error (?). In globals Ember,
  `App.SomeItem`.
* The typeKey is set to the normalized value: `SomeItem`

In an acceptance or app environment the container has a different resolver. In these cases normalization changes things. For example, [with the Ember-CLI resolver](https://github.com/ember-cli/ember-resolver/blob/64fca02c571057a93891a16cc3d20f7b38533e4f/packages/ember-resolver/lib/core.js#L168-L179):

* `modelFor("SomeItem")`
* Looks for `"model:SomeItem"` in the container
* This is normalized to `"model:some-item"` (Ember-CLI is dasherizing)
* The factory is returned. In Ember-CLI, that from `app/models/some-item`.
* The typeKey is set to the normalized value: `some-item`

There are several points of confusion IMO. I'd like to suggest two followup issues:

* Most developers do not realize that unit testing uses a different resolver/normalization than apps and acceptance tests. Acceptance tests are more "forgiving". If you camelCase in an Ember-CLI unit test you will get model-not-found errors. IMO (and I've PR this before, got it merged, and it was ripped out), Ember-Data should normalize model names **before** passing them to the container. This doesn't mean the typeKey should not be based on the container normalization (should not matter, should be an internal anyway), but that an additional normalization step would be introduced at the [beginning of `modelFor`](https://github.com/emberjs/data/blob/f71e90a76f8290f98315a1776a5d158b9c463694/packages/ember-data/lib/system/store.js#L1413). This normalization would ensure that Ember-Data, when there is no normalization provided by the container, always normalizes its own requests to camelCase or some other consistent format instead of blindly trying to trust user input.
* RESTAdapter defers to the `typeKey` for model keys. **If you change your resolver, you will change the requests sent to your server and the way data is parsed**.
  * [serializing a hash presumes the typeKey is the payload key](https://github.com/emberjs/data/blob/f71e90a76f8290f98315a1776a5d158b9c463694/packages/ember-data/lib/serializers/rest_serializer.js#L726). [bmac suggested](https://github.com/emberjs/data/pull/2766/files#diff-e9dfda7f1f589005862728db5d21e6d4R726) using `typeForRoot` here (which is definitely just a happens to work fix, it camelizes) or introducing a `rootForType`. `rootForType` with a default of camelization (based on [other typeKey normalization in that file](https://github.com/emberjs/data/blob/f71e90a76f8290f98315a1776a5d158b9c463694/packages/ember-data/lib/serializers/rest_serializer.js#L746)) is correct and I believe necessary.
  * Embedded records mixin also seems to presume the type can be read from a server response [during extract](https://github.com/emberjs/data/blob/f71e90a76f8290f98315a1776a5d158b9c463694/packages/ember-data/lib/serializers/embedded_records_mixin.js#L440). This should maybe go through `typeForRoot` as well.

See also:

* https://github.com/emberjs/data/pull/2766
* https://github.com/bantic/data/pull/1

### DIFF
--- a/packages/ember-data/lib/serializers/json_serializer.js
+++ b/packages/ember-data/lib/serializers/json_serializer.js
@@ -487,17 +487,32 @@ export default Serializer.extend({
   },
 
   /**
-    You can use this method to customize how a serialized record is added to the complete
-    JSON hash to be sent to the server. By default the JSON Serializer does not namespace
-    the payload and just sends the raw serialized JSON object.
-    If your server expects namespaced keys, you should consider using the RESTSerializer.
-    Otherwise you can override this method to customize how the record is added to the hash.
+    Serialize a model snapshot into a server request payload.
 
-    For example, your server may expect underscored root objects.
+    By default the JSON Serializer does not namespace
+    payloads, instead a "raw" serialized object is sent. For example:
+
+    ```js
+    {
+      id: 'car-one',
+      speed: 'fast'
+    }
+    ```
+
+    Note the lack of a namespace, and of any information about what
+    model this payload refers to.
+
+    If your server expects namespaced payloads, you should consider using the RESTSerializer.
+
+    You can also override this method to customize serialization behavior.
+    For example, your server may expect underscored model namespaces:
 
     ```js
     App.ApplicationSerializer = DS.RESTSerializer.extend({
       serializeIntoHash: function(data, type, snapshot, options) {
+        // type.typeKey is a model name with formatting decided by your
+        // application container. In globals mode, ususally camelCase. In
+        // Ember-CLI, usually dasherize-case.
         var root = Ember.String.decamelize(type.typeKey);
         data[root] = this.serialize(snapshot, options);
       }
@@ -678,7 +693,10 @@ export default Serializer.extend({
         }
       }
     });
-   ```
+    ```
+
+    The default behavior of this method is to not serialize
+    type data for polymorphic relationships.
 
     @method serializePolymorphicType
     @param {DS.Snapshot} snapshot
--- a/packages/ember-data/lib/serializers/json_serializer.js
+++ b/packages/ember-data/lib/serializers/json_serializer.js
@@ -1,4 +1,6 @@
 import Serializer from "ember-data/system/serializer";
+var camelize = Ember.String.camelize;
+import { singularize } from "ember-inflector/system/string";
 
 var get = Ember.get;
 var isNone = Ember.isNone;
@@ -1100,5 +1102,38 @@ export default Serializer.extend({
     var transform = this.container.lookup('transform:' + attributeType);
     Ember.assert("Unable to find transform for '" + attributeType + "'", skipAssertion || !!transform);
     return transform;
+  },
+
+  /**
+    Convert an Ember-Data model type string into a type string for a request
+    payload. For example, your server may expect types to be camelCased:
+
+    ```js
+    // a request payload
+    {
+      "fastCar": {
+        "id": "1",
+        "name": "corvette"
+      }
+    }
+    ```
+
+    The default behavior is to camelCase and singularize model keys.
+    This method is used in `serializeIntoHash`, `serializePolymorphicType`
+    and in the embedded records mixin.
+
+    If you customize this behavior, keep in mind that Model naming
+    conventions may vary between applications and environements, but
+    will most often follow these standards:
+
+    * Ember-CLI applications will follow the pattern: `dasherized-model-name`.
+    * Globals-mode Ember applications will follow the pattern `camelCaseName`.
+
+    @method typeForPayload
+    @param {String} key
+    @return {String} the model's name in your request payloads
+  */
+  typeForPayload: function(key) {
+    return camelize(singularize(key));
   }
 });
--- a/packages/ember-data/lib/serializers/rest_serializer.js
+++ b/packages/ember-data/lib/serializers/rest_serializer.js
@@ -505,15 +505,13 @@ var RESTSerializer = JSONSerializer.extend({
   },
 
   /**
-    This method is used to convert each JSON root key in the payload
-    into a typeKey that it can use to look up the appropriate model for
-    that part of the payload. By default the typeKey for a model is its
-    name in camelCase, so if your JSON root key is 'fast-car' you would
-    use typeForRoot to convert it to 'fastCar' so that Ember Data finds
-    the `FastCar` model.
+    This method is used to convert each JSON root key (from a payload)
+    into a model name. Model naming conventions may vary between
+    applications and environements, but will most often follow
+    these standards:
 
-    If you diverge from this norm you should also consider changes to
-    store._normalizeTypeKey as well.
+    * Ember-CLI applications will follow the pattern: `dasherized-model-name`.
+    * Globals-mode Ember applications will follow the pattern `camelCaseName`.
 
     For example, your server may return prefixed root keys like so:
 
@@ -527,25 +525,43 @@ var RESTSerializer = JSONSerializer.extend({
     ```
 
     In order for Ember Data to know that the model corresponding to
-    the 'response-fast-car' hash is `FastCar` (typeKey: 'fastCar'),
-    you can override typeForRoot to convert 'response-fast-car' to
-    'fastCar' like so:
+    the `response-fast-car` property is the Globals-mode class
+    `App.FastCar` the `typeForRoot` must be overridden. For example:
 
     ```js
     App.ApplicationSerializer = DS.RESTSerializer.extend({
       typeForRoot: function(root) {
         // 'response-fast-car' should become 'fast-car'
         var subRoot = root.substring(9);
+        // normalizes 'fast-car' to 'fastCar'
+        // singularize for good measure
+        return subRoot.singularize().camelize();
+      }
+    });
+    ```
 
-        // _super normalizes 'fast-car' to 'fastCar'
-        return this._super(subRoot);
+    The default implementation of `typeForRoot` is to
+    `singularize` then `camelize` a payload property name.
+    This does not technically match the conventions of
+    Ember-CLI, and although it will work well in apps
+    and acceptance tests, may be problematic in unit tests.
+
+    A suggested implementation for Ember-CLI users would
+    be:
+
+    ```js
+    // app/serializers/application.js
+    import DS from "ember-data";
+    export default DS.RESTSerializer.extend({
+      typeForRoot: function(root) {
+        return subRoot.singularize().dasherize();
       }
     });
     ```
 
     @method typeForRoot
     @param {String} key
-    @return {String} the model's typeKey
+    @return {String} the model's name in the container
   */
   typeForRoot: function(key) {
     return camelize(singularize(key));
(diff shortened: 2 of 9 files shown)


packages/ember-data/tests/integration/serializers/rest_serializer_test.js
```diff
@@ -284,19 +284,6 @@ test("serialize polymorphicType", function() {
...
-  run(function() {
-    tom = env.store.createRecord(YellowMinion, { name: "Alex", id: "124" });
-    ray = env.store.createRecord(DoomsdayDevice, { evilMinion: tom, name: "DeathRay" });
-  });
-
-  var json = env.restSerializer.serialize(ray._createSnapshot());
-
-  deepEqual(json["evilMinionType"], "yellowMinion");
-});
-
```
mixonic: I believe this test is incorrect.

* You should never manually set a `typeKey`.
* The `YellowMinion` was registered into the container with camelCase.
* This is a unit test, there is no normalization of model names

Looking into polymorphics, I am confused by the type usage. Sometimes it seems like we trust the typeKey ([serialization](https://github.com/emberjs/data/blob/f71e90a76f8290f98315a1776a5d158b9c463694/packages/ember-data/lib/serializers/rest_serializer.js#L726)) other times we camelize it ([relationType](https://github.com/emberjs/data/blob/f71e90a76f8290f98315a1776a5d158b9c463694/packages/ember-data/lib/serializers/rest_serializer.js#L746)).

bmac:
I'm also in favor of adding a `rootForType` with a default of camelization for backwards compatibility. I now believe the `typeForRoot` that I used in #2766 was a mistake. 


Overall I like this pr. I wanted to bring up a that one of @tomdale's reasons for suggesting Ember Data issues a deprecation warning when it detects caps (that may not have been well communicated in the comment in #2766) was the fear that users would have to change their type argument the `store.find` call when migrating from globals to the Ember CLI resolver.

mixonic:
@bmac I believe a user can move app code between globals and Ember-CLI smoothly today. The [default ember application container normalization](https://github.com/emberjs/ember.js/blob/73d49ffa092018557bb185d0aae41531e0a87e82/packages/ember-application/lib/system/resolver.js#L119-L147) should be sufficient.

Moving unit tests is the mess: There is no normalization, but the model names in the container have changed. Thus you end up being forced into dasherized format in unit tests.

I'll add `rootForType` to this PR.

mixonic:
Slight change of plans: I did not implement a `rootForType`, as it is would only be used in `serializeIntoHash`. Instead I just `camelize` in `serializeIntoHash`.

Need to add a test for new `serializeIntoHash` behavior. It should assert that the current behavior in Ember-CLI (that a payload namespace is camelCased) occurs even when dasherized names are used in the container.

Did some cleanup of embedded records mixin. The tests make a lot of assumptions about camelCasing. Need to add some for serialization.

mixonic:
* More tests
* Introduce [typeForPayload](https://github.com/mixonic/data/blob/468ad0f39af5f6c18859176fb14b6e2cb6dfddb4/packages/ember-data/lib/serializers/json_serializer.js#L1107-L1138) in the JSONSerializer. I added it to the JSONSerializer because embedded records mixin is tested against that serializer, and this method is required for polymorphic serialization. `typeForPayload` allows us to not leak the typeKey. An Ember-CLI app (with dasherized keys) and a globals app (with camelCase keys) will send the same polymorphic type string to the server via `typeForPayload` (camelCase by default for backwards compat).
* Importantly, I believe this PR changes *nothing* about how polymorphic type are sent to the server. It also changes *nothing* about how mixin keys are read from response payloads. So it should be backwards compatible.

@bmac @igorT @stefanpenner I'd appreciate a look. I think this is good to go.

fivetanley:
@mixonic I think @tomdale and @wycats thought this kind of logic should always delegate to a Resolver. Am I way off track here tomhuda?

mixonic:
This is rebased and passing. Waiting on a +1 or further discussion.

packages/ember-data/lib/serializers/rest_serializer.js
```diff
@@ -527,25 +525,43 @@ var RESTSerializer = JSONSerializer.extend({
...
+
+    A suggested implementation for Ember-CLI users would
+    be:
+
+    ```js
+    // app/serializers/application.js
+    import DS from "ember-data";
+    export default DS.RESTSerializer.extend({
+      typeForRoot: function(root) {
+        return subRoot.singularize().dasherize();
```
bmac: Does this mean Ember CLI users will still need to use a custom Serializer after this pr is merged?

mixonic: I want to stay really specific here. I don't think most devs know that a dasherizing `typeForRoot` can make their ember-cli unit tests work properly (with multi-word model names). The current code will pass a camelized name to the store (and then container, where there is no normalization), and not find the model. That behavior is not changed in this PR.

I would recommend that they use a singularize then dasherize `typeForRoot`, yeah.

mixonic:
We discussed this a bit at the ember-data meeting.

* [ ] Let's introduce a property on classes called `modelName`, which is always dasherized
* [ ] Let's deprecate getting `typeKey`
* [ ] We need both `typeForRoot` and `payloadForType`, though perhaps the names need more bikeshedding (`modelNameFromPayload`, `modelNameForPayload` I suggest)
* [ ] Before the use-dashes-everywhere work can land Ember's default container, the isolated container, and the Ember-CLI container must raise a warning or deprecation notice for `camelCase` lookups.

stefanpenner:
> Let's introduce a property on classes called modelName, which is always dasherized

can you also explain what this property will be used for.

> Before the use-dashes-everywhere work can land Ember's default container, the isolated container, and the Ember-CLI container must raise a warning or deprecation notice for camelCase lookups.

only the resolver needs these deprecation warnings

mixonic:
> can you also explain what this property will be used for.

It is the new public interface for model names. Always dasherized, which matches our conventions across the board (once the deprecations are added to the resolvers)

> only the resolver needs these deprecation warnings

Ah, so the normalization is different in the three places I listed. It seems correct that the resovler may be the only things that needs to throw, but the normalization should probably be reviewed.

stefanpenner:
> It is the new public interface for model names. Always dasherized, which matches our conventions across the board (once the deprecations are added to the resolvers)

What are the public usages of this api? (i am just curious why a name placed on a factory is ever needed)

stefanpenner:
> Ah, so the normalization is different in the three places I listed. It seems correct that the resovler may be the only things that needs to throw, but the normalization should probably be reviewed.

all normalization must go through the resolver

mixonic:
> What are the public usages of this api? (i am just curious why a name placed on a factory is ever needed)

when building a url in the adapter. during serialization of a polymorphic model. when determining the root property name of a create payload.

> all normalization must go through the resolver

This has changed recently. Looks like you're right. I see at least these:

* normalize on the default resolver https://github.com/emberjs/ember.js/blob/master/packages/ember-application/lib/system/resolver.js#L119
* normalize on the ember-cli resolver https://github.com/ember-cli/ember-resolver/blob/master/packages/ember-resolver/lib/core.js#L139

So "the resolver" translates back to my original statement. "Ember's default container, the isolated container, and the Ember-CLI container must raise a warning or deprecation notice for camelCase lookups."

It seems a non sequitur to raise that normalization goes through the resolver. The behavior still needs to change in three situations, correct?

stefanpenner:
> So "the resolver" translates back to my original statement. "Ember's default container, the isolated container, and the Ember-CLI container must raise a warning or deprecation notice for camelCase lookups."

i want to clarify further, to make sure we remove all mis-behaving code. The resolver is the only thing that knows about naming conventions, and as such it is the only place we should put these deprecations. With one exception, and that is plural/singlular inflections. If other exceptions exist, I strongly suspect that is demonstrating a bug, that we must resolve.

packages/ember-data/lib/serializers/json_serializer.js
```diff
@@ -1082,5 +1102,38 @@ export default Serializer.extend({
...
+
+    * Ember-CLI applications will follow the pattern: `dasherized-model-name`.
+    * Globals-mode Ember applications will follow the pattern `camelCaseName`.
+
+    @method typeForPayload
+    @param {String} key
+    @return {String} the model's name in your request payloads
+  */
+  typeForPayload: function(key) {
+    return camelize(singularize(key));
```
stefanpenner: is this plucking from the payload ?

packages/ember-data/lib/serializers/rest_serializer.js
stefanpenner: do API's return dasherized responses?

packages/ember-data/lib/system/store.js
```diff
@@ -1864,7 +1861,11 @@ Store = Service.extend({
     @return {String} if the adapter can generate one, an ID
   */
   _normalizeTypeKey: function(key) {
-    return camelize(singularize(key));
+    // Delegate to the container for normalization. The container
+    // requires a ':' to normalize so we add `model:` then slice off
+    // the first 6 characters to remove the `model:` in the return
+    // value
+    return this.container.normalize('model:' + key).slice(6);
```
stefanpenner: we should add `normalizeType('model')` to the container, which then delegates to the resolver.

packages/ember-data/tests/integration/serializers/embedded_records_mixin_test.js
```diff
@@ -1321,3 +1321,164 @@ test("serializing relationships with an embedded and without calls super when no
...
+    secretWeapons: DS.hasMany("secret-weapon", { polymorphic: true })
+  });
+
+  env.registry.register('adapter:super-villain', DS.ActiveModelAdapter);
+  env.registry.register('serializer:super-villain', DS.ActiveModelSerializer.extend(DS.EmbeddedRecordsMixin, {
+    attrs: {
+      secretWeapons: { embedded: 'always' }
+    },
+    typeForRoot: function(type) {
+      return Ember.String.dasherize(type);
```
stefanpenner: is this dash or underscore, the keys in the payload are all underscored.

fivetanley:
> only the resolver needs these deprecation warnings

The idea here is that you shouldn't be able to register anything in the container that is not dasherized either.

stefanpenner:
> The idea here is that you shouldn't be able to register anything in the container that is not dasherized either.

registration itself is some-what of an anti-pattern, most app scenarios should ideally be resolved.

cah-briangantzler:
Each version of Ember data that comes out I check to see if perhaps this has been completed :) Any idea on when this may make it into a beta? 1.0 is coming quickly and hoping this does not get deferred to 2.x I currently can not use the mock http facility and would very much like to.

cah-briangantzler:
Looking at the code more closely, this particular problem is indeed fixed.(Tried beta.18, could be fixed earlier though) The correct adapter is being called. My test is still failing because the http-mock is not being installed. I have found that the tests work using http:/localhost:4200/tests but does not work with ember test -s. Any ideas?

mixonic:
I believe much of this has been addressed in beta.18 in other PRs by @fivetanley (thank you sir). I'm closing, though I'd like to port my tests from here forward to master now that his code has landed.

fivetanley:
@mixonic much of the code left is in #3074

cah-briangantzler:
Are mocks suppose to be installed when running ember test -s?

bmac:
@cah-briangantzler I asked @rwjblue about mocks with `ember test -s` and he reports that they currently do not work.

PR #2835, opened 2015-03-06
Copy the attr's defaultValue before assigning it.

### LINKED ISSUE #2833: Return copy of object specified in defaultValue
For example I have next model:
```
var Model = DS.Model.extend({
  myArray: DS.attr({ defaultValue: [] })
})
```

When I use:
```
var object1 = this.get('store').createRecord('model');
var object2 = this.get('store').createRecord('model');
object1.get('myArray').addObject('foo');
console.log(object2.get('myArray'));
```
I receive: `['foo']`

I know that I should use `myArray: DS.attr({ defaultValue: function () { return []; })`
But this moment isn't explicitly described in documentation.
I would like to see default behaviour to return copy of object if `defaultValue` function isn't specified.

bmac:
Closes #2833 

This is a breaking change for anyone relying on the old behavior. Also this pr just performs a shallow copy. Should ED make a deep copy or simply require users to define their own function if they want a deep copy?

### DIFF
--- a/packages/ember-data/lib/system/model/attributes.js
+++ b/packages/ember-data/lib/system/model/attributes.js
@@ -211,7 +211,7 @@ function getDefaultValue(record, options, key) {
   if (typeof options.defaultValue === "function") {
     return options.defaultValue.apply(null, arguments);
   } else {
-    return options.defaultValue;
+    return Ember.copy(options.defaultValue);
   }
 }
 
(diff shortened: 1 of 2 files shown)


yokree:
I would like to see deep copy, shallow copy could be a source of bugs for people who doesn't know defaultValue behaviour.

stefanpenner:
i think this is misleading, if someone has a complex object they should use the function form of `defaultValue`

bmac:
Thanks for the feedback @stefanpenner. I'll close this and we can just update the docs to call out that developers will need to use the function syntax if they want to return unique objects.

Issue #2836, opened 2015-03-06
pushPayload and meta data


hcatlin:
I'm using websockets to send data back to my application, and so I'm using `pushPayload` to update the store. My data looks something like this:

````
{
  "user": { 
    "id": "hcatlin"
  },
  "meta": {
    "online": 12
  }
}
````

That response works great for `store.find` and the like, but totally fails on `pushPayload`.

````
this.store.pushPayload('user', data);
````

I get the error, "can't find model metum", which is pushPayload adorably thinking that `meta` isn't something special.

So! I have to do this to get it working.

````
this.store.setMetadataFor('user', data.meta);
delete data.meta;
this.store.pushPayload('user', data);
````

Seems like `pushPayload` should know how to deal with metadata... if that's the go-to entry point for things like WS.

Am I crazy?

fivetanley:
@igorT says this should be fixed by the new serializer API in https://github.com/emberjs/data/pull/2837, please re-open if it's still an issue.

PR #2837, opened 2015-03-06
store.pushPayload handles meta keys, and serializer.pushPayload deep copies payload

hcatlin:
Added tests, and code to allow pushPayload to accept meta keys.

First, the `DS.Store.pushPayload` method was updated to call `DS.Store.setMetadataFor` if the property is found.

Secondly, the RESTSerializer was taking the 'meta' key and thinking it was referencing a model and attempting to update 'metum'. The RESTSerializer now silently ignores any meta keys it sees, since meta keys are a very special type of key name that will never map to a DS.Model.


hcatlin:
Sorry, this also closes issue #2836

packages/ember-data/lib/serializers/rest_serializer.js
```diff
@@ -487,20 +490,22 @@ var RESTSerializer = JSONSerializer.extend({
...
 
     for (var prop in payload) {
-      var typeName = this.typeForRoot(prop);
-      if (!store.modelFactoryFor(typeName, prop)) {
-        Ember.warn(this.warnMessageNoModelForKey(prop, typeName), false);
-        continue;
-      }
-      var type = store.modelFor(typeName);
-      var typeSerializer = store.serializerFor(type);
+      if (prop !== "meta") {
```
bmac: Instead of doing this check could you call this.extractMeta before this.normalizePayload to mirror how meta gets extracted in the normal flow?

hcatlin: Good call. Totally agree. I didn't see that method before.

hcatlin: As I'm looking at it, `extractMeta` doesn't just remove the meta key, it actually goes and sets the metadata on the store. Which is fine, it means the store could be simpler, but the downside is that it requires a type, and we don't know the type at this point. So, I'm going to replace the code in store with a call to `serializer.extractMeta`, before we call `serializer.pushPayload`, since that's the only place where we have all the ingredients required for a call to `JSONSerializer.extractMeta`

hcatlin: Actually `extractMeta` mutates the original object passed in, by calling `delete payload.meta`, which means that the caller of the `pushPayload()` method, would be surprised to find that their payload object was mutated after calling the function.

The only way to avoid this is to either duplicate the payload object, then remove the key, or use the technique I use here to use comparisons to ignore the key. I think the original approach (which I did to avoid use of `delete`, which `extractMeta` uses) is the less of two evils. It's not ideal to run comparisons each iteration, but duplicating an object is even less ideal. 

Thoughts?

bmac: I think mutating the payload object is ok. There are other paths through the serializer already end up deleting properties from the payload. Ideally we could clone the payload at the start of the serialization process but that ends up being really expensive with large payload in older browsers.

hcatlin: I strongly disagree with that... the following code failing would just be... unacceptable.

````
this.store.pushPayload("message", data);
if(data.meta !== undefined) {
  this.flash.notice("Only " + data.meta.total + " messages left!");
}
````

That's the kind of bug that makes developers pull their hair out. I don't think saying that this side-effect exists in other parts of the system is a good defense for why it should get worse.

The "if-statement" method in my code works well, because the internal serializer code can mutate it's own data object all it wants, and, while it's slightly slower than deletion, the `pushPayload` method of the serializer is there for public API usage, and mutations would be highly unexpected. This seems like a good trade off to me. 

Also, I haven't observed any mutations in the `pushPayload` serializer in my usage of it, yet. I think those parts of the code might be closer to the non-manual methods.

bmac: Thanks for your concern @hcatlin. I have flagged this for discussion at the Ember Data weekly meeting.

hcatlin:
Any movement on this?

bmac:
We ran out of time in the last Ember Data meeting before we got to this issue. I'll try bring it up again at the next meeting on Friday.

bmac:
@hcatlin As Tom mentioned in #2952, it sounds like the plan is to add a deep copy to the beginning of `pushPayload` that way we can still modify the object without breaking user's expectations.

hcatlin:
Cool. I'll update this to match that behaviour.

hcatlin:
I've updated this pull request to also close #2952, by doing a deep copy before we do any mutations.

Specifically, the deep copy happens before normalizePayload gets called, and the meta key is deleted after that. Why that order? If I were implementing my own normalizePayload (which seems like a thing we recommend users to do), I'd expect to see *everything* I passed in in there. Even the meta key. Because, we want it to seem that we are ignoring the meta key, not messing with it. Putting the delete *after* the normalization means that the overall effect to the user is that it was ignored, and we did it by mutating, but only after the developer never sees this particular copy of the payload again.

/cc @tomdale

fivetanley:
Hey Hampton,

Sorry for taking so long to review this. We've been discussing making the serializer API more functional and user friendly for 1.0. While this PR works for the default case where you pass both a type and payload to `pushPayload`, it will set the meta on the `payload` type. Also, I think we can use `extractMeta` here instead of using `delete` in other places.

This PR has a great direction, but we don't want to send you on a goose chase while @wecc and @igor rapidly iterate on the serializer API. You've spent a lot of time patiently working and communicating with us, and we don't really want to waste any more of your time. We would love to keep the test for @wecc and @igorT's refactor. Once they open a branch on this repo (it's currently on wecc's I think) I will merge this commit's test over on that branch. I'll close this when that happens.

igorT:
I believe this should be fixed with the new serializer api,

PR #2857, opened 2015-03-11
Fire didDelete event/callback when in uncommited state

cibernox:
Now newly created and uncommited records will fire its livecycle events
when deleted.


cibernox:
Fixes #2857 

Yay!

cibernox:
force-pushed to trigger app-veyor again

fivetanley:
thanks!

PR #2867, opened 2015-03-12
Don't remove records from record arrays before commiting

drogus:
Before this commit records were deleted from record arrays just after
deleting them, but before they were commited. This can be an issue if
you want to keep records visible in the UI until they're actually
removed.


ianstarz:
This should be closed in favor of #3247

Issue #2862, opened 2015-03-12
We do not destroy records that have been unloaded


igorT:
We probably should?

bmac:
Sounds like the correct move. It also looks like we don't remove the `unloaded` record from the store unless we are calling `store.unloadAll`. Is that something we should be doing?

aaronbhansen:
As far as I have seen, if you delete a record, but don't destroy it, when you need to save that delete later there isn't an easy way to get all records that have been deleted in ember and save the deletes to the server. The only easy way is to keep a reference of the deleted models somewhere. In our app, we try to save everything at once before the page transition, and this is one issue we have. Seems like it's related to this ticket. It would be great if there was a way to specifically get deleted records, or on a has many save, save the deleted records first.

drogus:
@aaronbhansen I think this is not quite that case. This is only talking about unloading records, so when you run `record.unloadRecord()`. Deleted records should be unloaded as well, but even in this case it will happen only when the change is persisted, ie. when the record goes in `deleted.saved` state

aaronbhansen:
Okay, maybe mine is a separate issue then. It sounded similar but I wasn't sure. I didn't see another issue about saving deleted records at a later point in time, but this sounded close. If this issue if purely about unloading, I can create a separate issue.

givanse:
I think that destroying records that have been unloaded would break the workaround that I have for this scenario: https://github.com/emberjs/data/issues/2982 If you can tell me how to push this forward I'll do some more tinkering.

igorT:
tracking this under https://github.com/emberjs/data/issues/3296

PR #2870, opened 2015-03-12
Fix unload record

### LINKED ISSUE #2859: Unloading a record sets isDeleted to true
This might be a problem if someone unloads a record without deleting it.

drogus:
This targets #2859 and #2862. When I started my work on this I only wanted to fix #2859, but both issues are connected and I believe that it's better to fix them at once. You can't easily change transitioning to `deleted.saved` without doing something else that removes record from record arrays, which can be achieved by destroying a record. Additionally, destroying a record on unloading without dematerializing it doesn't really make sense, so I also added that.

One more thing that should be a motivation for this change is that `dematerializeRecord` and it mentions using `unloadRecord`, which isn't an equivalent at the moment.

### DIFF
--- a/packages/ember-data/lib/system/model/states.js
+++ b/packages/ember-data/lib/system/model/states.js
@@ -455,10 +455,9 @@ var RootState = {
   // you out of the in-flight state.
   rolledBack: Ember.K,
   unloadRecord: function(record) {
-    // clear relationships before moving to deleted state
-    // otherwise it fails
-    record.clearRelationships();
-    record.transitionTo('deleted.saved');
+    record.destroy();
+    var store = get(record, 'store');
+    store._dematerializeRecord(record);
   },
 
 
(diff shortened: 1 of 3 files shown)


drogus:
I should probably add that currently the tests fail, because we don't remove records from record arrays when a record is destroyed or when it's destroying. It will pass once #2867 is fixed.

drogus:
@igorT is this thing still needed? I haven't followed ED development lately, so I'm not sure how it looks like in context of other changes.

tchak:
@drogus I would say that yes it is still needed. But now there a record and internal-record to deal with.

@krisselden was asking about exactly this today in slack :)

fivetanley:
@drogus do you have the bandwidth to pick this back up?

drogus:
@fivetanley yup, although I would need to get familiar with ED internals again. I can look at it over the weekend. Do you know if the algorithm that @IgorT outlined [here](https://github.com/emberjs/data/issues/3296#issuecomment-111003976) is still up to date?

sly7-7:
There was also a work done by @igorT https://github.com/emberjs/data/pull/3301, and I've added a test. Not sure how this is related, and I can't remember what was missing for the test to pass.

bmac:
@drogus Looks like https://github.com/emberjs/data/pull/2867 is now fixed. What needs to happen to get this pr into a place where it can be merged?

bmac:
@drogus  Do you know needs to happen to get this pr into a place where it can be merged?

bmac:

PR #2875, opened 2015-03-13
Store.hasRecordForId should true only if a record for a given type and ID is already loaded.

bmac:
This pr is built on top of #2659

### DIFF
--- a/packages/ember-data/tests/unit/store/has_record_for_id_test.js
+++ b/packages/ember-data/tests/unit/store/has_record_for_id_test.js
@@ -1,0 +1,67 @@
+var env, store, Person, PhoneNumber;
+var attr = DS.attr;
+var hasMany = DS.hasMany;
+var belongsTo = DS.belongsTo;
+var run = Ember.run;
+
+module("unit/store/hasRecordForId - Store hasRecordForId", {
+  setup: function() {
+
+    Person = DS.Model.extend({
+      firstName: attr('string'),
+      lastName: attr('string'),
+      phoneNumbers: hasMany('phone-number')
+    });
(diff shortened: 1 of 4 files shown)


packages/ember-data/tests/unit/store/has_record_for_id_test.js
```diff
@@ -0,0 +1,55 @@
...
+    store = env.store;
+
+  },
+
+  teardown: function() {
+    Ember.run(store, 'destroy');
+  }
+});
+
+test("hasRecordForId should return false for records in the empty state ", function() {
```
igorT: How about a test for true while we are at it?

bmac: Will do

bmac: Added a test for for the true case.

packages/ember-data/tests/unit/store/has_record_for_id_test.js
```diff
@@ -0,0 +1,67 @@
...
+      lastName: "Katz",
+      phoneNumbers: [1]
+    });
+
+    equal(false, store.hasRecordForId(PhoneNumber, 1), 'hasRecordForId only returns true for loaded records');
+
+  });
+});
+
+test("hasRecordForId should return false for records in the loaded state ", function() {
```
sly7-7: I think you meant `true` here :hamster:

bmac: Good catch :dancers:

PR #2879, opened 2015-03-14
Have snapshots respect the order of items in hasMany relationships

### LINKED ISSUE #2852: Order of hasMany relationship is not maintained when serializing
Consider the model:

````javascript
App.Post = DS.Model.extend({
  comments: DS.hasMany('comment') 
})
````

And the payload:

````json
{
  "post": {
    "comment_ids": [1, 2, 3]
  }
}
````

When *deserializing*, the order of the comment IDs is respected. For example:

````javascript
var ids = post.get('comments').map(function(comment) {
  return comment.get('id')
})
console.log(ids.join(', ')) // 1, 2, 3
````

If you modify the order of the comments using methods such as `removeObject` and `insertAt`, the order of the IDs is *not* respected during serialization. For example:

````javascript
store.find('comment', 3).then(function(comment) {
  post.get('comments').removeObject(comment)
  post.get('comments').insertAt(0, comment)
  post.save()
})
````

In previous versions of Ember Data the resulting payload would have been:

````json
{
  "post": {
    "comment_ids": [3, 1, 2]
  }
}
````

In the latest version the order of the IDs is seemingly random. I suspect the previous functionality may have been an unsupported side effect but was quite useful. When using a NoSQL store such as Mongodb the comment_ids property is an actual document property stored in the database and the order has meaning. 

From digging through the source code it appears that the serializer uses the `relationship.members` property and the order of the comments attribute does not have any affect on serialization.

wecc:
Fixes #2852

### DIFF
--- a/packages/ember-data/lib/system/map.js
+++ b/packages/ember-data/lib/system/map.js
@@ -1,5 +1,5 @@
 /*
- The Map/MapWithDefault/OrderedSet code has been in flux as we try
+ The Map/MapWithDefault code has been in flux as we try
  to catch up with ES6. This is difficult as we support multiple
  versions of Ember.
  This file is currently here in case we have to polyfill ember's code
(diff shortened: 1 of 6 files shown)


igorT:
cc @stefanpenner 
We need an ability to insertAt to an orderedSet. We are subclassing for now, until we figure out the correct approach

Issue #2877, opened 2015-03-14
We should rename the local files to use dashes instead of underscores


igorT:

fivetanley:
if you do this, please use git mv

stefanpenner:
:+1:

jakesjews:
I made an attempt at this in pr #2880

sly7-7:
Closing since #2880 is merged. Just out of curiosity, is this a "new" convention for file naming ? Or is it a choice made in the ember-cli ecosystem ?

PR #2880, opened 2015-03-14
Rename the local files to use dashes instead of underscores. Relates to issue #2877

jakesjews:
Hopefully I got everything. I made sure to use git mv in my script and all tests are passing after building.


fivetanley:
#yolomerge thanks!

PR #2881, opened 2015-03-15
Rollback Relationships

mmpestorich:
This commit allows relationships to be rolledback and also reintroduces `dirtyRecordFor*Change` hooks on the adapter.

### DIFF
--- a/packages/ember-data/lib/system/adapter.js
+++ b/packages/ember-data/lib/system/adapter.js
@@ -397,6 +397,26 @@ var Adapter = Ember.Object.extend({
   */
   groupRecordsForFindMany: function(store, snapshots) {
     return [snapshots];
+  },
+
+  dirtyRecordForAttrChange: function (record, context) {
+    return context.value !== context.originalValue;
+  },
+
+  dirtyRecordForBelongsToChange: function (record, context) {
+    return context.value !== context.originalValue;
+  },
+
+  dirtyRecordForHasManyChange: function (record, context) {
+    var relationshipType = record.constructor.determineRelationshipType({ key: context.key, kind: context.kind });
+
+    if (relationshipType === 'manyToMany' || relationshipType === 'manyToNone') {
+      if (context.recordAdded) {
+        return !context.originalValue.has(context.recordAdded);
+      }
+      return context.originalValue.has(context.recordRemoved);
+    }
+    return false;
   }
 });
 
--- a/packages/ember-data/lib/system/model/attributes.js
+++ b/packages/ember-data/lib/system/model/attributes.js
@@ -292,7 +292,9 @@ export default function attr(type, options) {
 
   var meta = {
     type: type,
+    kind: 'attr',
     isAttribute: true,
+    key: null,
     options: options
   };
 
@@ -307,8 +309,9 @@ export default function attr(type, options) {
         this._attributes[key] = value;
 
         this.send('didSetProperty', {
-          name: key,
-          oldValue: oldValue,
+          key: key,
+          kind: 'attr',
+          isAttribute: true,
           originalValue: this._data[key],
           value: value
         });
--- a/packages/ember-data/lib/system/model/model.js
+++ b/packages/ember-data/lib/system/model/model.js
@@ -767,6 +767,15 @@ var Model = Ember.Object.extend(Ember.Evented, {
     });
   },
 
+  rollbackRelationships: function() {
+    this.eachRelationship(function(name, relationship) {
+      this._relationships[name].rollback();
+    }, this);
+    var model = this;
+    forEach.call(Ember.keys(this._implicitRelationships), function(key) {
+      model._implicitRelationships[key].rollback();
+    });
+  },
 
   /**
     @method updateRecordArrays
@@ -1002,15 +1011,14 @@ var Model = Ember.Object.extend(Ember.Evented, {
       set(this, 'isError', false);
     }
 
-    //Eventually rollback will always work for relationships
-    //For now we support it only out of deleted state, because we
-    //have an explicit way of knowing when the server acked the relationship change
-    if (get(this, 'isDeleted')) {
-      this.reconnectRelationships();
-    }
+    var isDeleted = get(this, 'isDeleted');
+    var isNew = get(this, 'isNew');
 
-    if (get(this, 'isNew')) {
-      this.clearRelationships();
+    if (isDeleted || isNew) {
+      if (isDeleted) { this.reconnectRelationships(); }
+      if (isNew) { this.clearRelationships(); }
+    } else {
+      this.rollbackRelationships();
     }
 
     if (!get(this, 'isValid')) {
--- a/packages/ember-data/lib/system/model/states.js
+++ b/packages/ember-data/lib/system/model/states.js
@@ -4,6 +4,8 @@
 
 var get = Ember.get;
 var set = Ember.set;
+var classify = Ember.String.classify;
+
 /*
   This file encapsulates the various states that a record can transition
   through during its lifecycle.
@@ -174,11 +176,21 @@ var set = Ember.set;
 */
 
 function didSetProperty(record, context) {
-  if (context.value === context.originalValue) {
-    delete record._attributes[context.name];
-    record.send('propertyWasReset', context.name);
-  } else if (context.value !== context.oldValue) {
+  var adapter = get(record, 'store').adapterFor(record.constructor);
+  var fn = adapter['dirtyRecordFor' + classify(context.kind) + 'Change'];
+
+  if (fn(record, context)) {
+    if (context.isRelationship) {
+      record._relationships[context.key].isDirty = true;
+    }
     record.send('becomeDirty');
+  } else {
+    if (context.isRelationship) {
+      record._relationships[context.key].isDirty = false;
+    } else {
+      delete record._attributes[context.key];
+    }
+    record.send('propertyWasReset', context.key);
   }
 
   record.updateRecordArraysLater();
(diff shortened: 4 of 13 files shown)


fivetanley:
Thanks! I will bring this up at friday's meeting.

jsedlacek:
Thanks @mmpestorich! This solves a very painful point of ember data. Are there any news on having this merged?

topaxi:
What is the state on this? We're currently managing our relation rollback by hand and it gets very confusing sometimes.

dallasread:
+1

bmac:
I'm going to close this pr. @igorT is working on an RFC to describe what needs to happen to for Ember Data to support relationship rollback.

mmpestorich:
@bmac This pull request implements rollback for both belongsTo and hasMany relationships. @igorT work in #3698 only addresses belongsTo relationships. 

If it's a matter of more discussion, conforming to an RFC, cleaning up the code, etc... please point me in the right direction. I'll gladly help in anyway that I can. This PR has been sitting out here for over a year now with no feedback from the ember-data team... slightly frustrating. Prior to this, I had pushed up #1631 - which I closed after being told that rolling back relationships would be addressed by the single-source-of-truth branch - ulitmately it wasn't. 

The bottom line is that this PR works and has a handful of passing tests and it's been over 2 years since ember-data was able to rollback relationships (when this behavoir was removed from ember-data, it was sold as a temporary thing). I'd gladly contribute more time and energy to this in order to put something together that you all would feel comfortable accepting... anything I can do please let me know.

There's up to date version of this commit [here](https://github.com/swarmbox/data/tree/master%2Brollback-relationships)

steffenbrem:
@mmpestorich It looks really nice, but I am wondering if it is also working correctly after you've saved a record. I've also implemented something like this myself, but `canonicalState` of a relationship isn't updated with the new saved records after a successful save. It might also be a good idea to test that, since I think it is broken in this PR too.

mmpestorich:
Hm... our UI doesn't allow a rollback after a record has been saved so I don't think I've ever encountered that. I'll create a test and look into what it would take to handle that properly.

steffenbrem:
@mmpestorich I think it is a common scenario, lets say you have a `parent` record and you add a `child` record on it's `children` relationship and save `parent`. Now you have a saved state of `parent` that contains one `child`. Next, you edit the `parent` record and add another `child`, it now has two children. This time, we do not save it, but we rollback the `parent` record. Right there, the issue exposes itself, since you now have **zero** children after the rollback (because of the canonicalState I think).

A code example to illustratie this use-case:
```js
parent.get(‘children’); // []

// save record
parent.get(‘children’).pushObject(child1);

// we save parent and it's linked child record
parent.save().then(() => {
    
    // add another child
    parent.get(‘children’).pushObject(child2);
    parent.rollback();

    // this will be empty, but you would expect it to contain `child1`
    parent.get(‘children’); // []
});
```

steffenbrem:
A way to solve it is to let the API return the ids of the `children` relationship of the `parent` record. It will than update the `canonicalState` appropriately. So having a `related` link for the relationship will break proper rollback of saved records. I do think that it might be good to make the `currentState` the `canonicalState` of all relationships if the save was successful, unregarded if the `PATCH` request includes the ids in the response payload.

mmpestorich:
@steffenbrem Ok, couple things....

I added a [test](https://github.com/emberjs/data/compare/master...swarmbox:master+rollback-relationships#diff-20e6ca00f88c093aea017f7286938567) that passes the parent-child scenario you outlined above. Both `currentState` and `canonicalState` are maintained properly.

The [dirtyRecordFor](https://github.com/emberjs/data/compare/master...swarmbox:master+rollback-relationships#diff-bb48652aef7fcc2b4d9fdec294cb1b01) hooks that I added back to `DS.Adapter`, by default, won't dirty the `parent` when any of its `hasMany` relationships change. The `parent` didn't actually change, but rather the `child` that points to the `parent` did. I'm making an assumption here (that I probably shouldn't be) about the type of backend/database in use. If you're using an RDMS, there's no reason to dirty the `parent` in this instance as there's nothing to actually persist to the `parent`. It's the `child` that actually maintains all the relational data in an RDMS type system. So, along this line of thinking, you would `save` or `rollback` the `child` and not the `parent` when the relationship between the two change. Using this child-up approach properly updates the `canonicalState` of the relationship.

With that said, I realize that there are a lot of people using some form or fashion of a NoSQL database in which it might make more sense to simply `save` or `rollback` the `parent`. To handle that, the `dirtyRecordFor` hooks could be easily modified to dirty the `parent` when a child is added or removed from a `hasMany` relationship. However, I believe that you might be right in that the `canoncialState` is not kept up to date when you take this parent-down approach. Though I haven't really tested this scenario...

steffenbrem:
@mmpestorich Thanks for taking time to look into this :smiley:! You are indeed right that people manage their relationships differently in Ember Data, but I think there should be a mechanism build in to update the canonicalState of relationships when the parent is saved. I am using a RDMS btw (not that important), and do persist relationship changes through the parent, simply because it requires much less requests. I don't think saving every relationship independently of the parent is a good solution (at least not in my apps). Right now I use the `include` trick when saving, to let ember data reload the relationships and so the canonical state.

Do you have an idea on how you can update the canonicalState of all relationships of a parent record that has been saved (without including every relationship in the response)? I have tried to fix this, but I lack the knowledge for it.

mmpestorich:
@steffenbrem 

> people manage their relationships differently in Ember Data

:stuck_out_tongue_winking_eye: Well it just wouldn't be fun if everyone did everything the same way!

> [I] do persist relationship changes through the parent, simply because it requires much less requests. I don't think saving every relationship independently of the parent is a good solution

I actually agree. On our end we added a `commitRecords` method to `DS.Store` for just that reason. You can look at that code [here](https://github.com/emberjs/data/compare/master...swarmbox:commit-records) if you're interested. Of course, you'd also have to implement something on the backend as well that can handle the request it sends off.

To handle commits and rollbacks for multiple records, we rolled our own `TransactionMixin`. In conjunction with the `commitRecords` code, commiting and rollingback multiple records at one time is seamless and transparent. Here's the gist of our transaction related code:

Basically, we extend `DS.Model` to notify the current `Route` when it becomes dirty or clean (our models all extend this instead of `DS.Model`):
```js
var TransactionEnabledModel = DS.Model.extend({

  _isDirtyDidChange: function () {
    var router = Ember.getOwner(this).lookup('router:main');
    var activeTransition = router.get('router.activeTransition');
    var record = this;

    if (!router.router.currentHandlerInfos) { return; }

    handleDirtyChange(activeTransition || router);

    function handleDirtyChange(handler) {
      if (record.get('isDirty')) {
        handler.send('modelBecameDirty', record);
      } else {
        handler.send('modelBecameClean', record);
      }
    }
  }.observes('currentState.isDirty'),
});
```

Then we add our `TransactionMixin` to routes that manage persistent data. Its purpose is to respond to the `modelBecameDiry` and `modelBecameClean` notifications that originate from our models. In addition, it handles `commit` and `rollback` notifications triggered by a user action on the current page:
```js
var TransactionMixin = Ember.Mixin.create({

  transaction: Transaction.create(),

  commitPromise: null,

  commitTransaction: function () {
    var self = this;

    this.get('commitPromise', this.get('transaction').commit(this.store))
                                                     .then(saveSuccess)
                                                     .catch(saveFail)
                                                     .finally(clearPromise);

    function clearPromise() {
      self.set('commitPromise', null);
    }
  },

  actions: {

    modelBecameDirty: function (model) {
      this.get('transaction').addDirty(model);
    },

    modelBecameClean: function (model) {
      this.get('transaction').removeClean(model);
    },

    rollback: function () {
      this.get('transaction').rollback();
    },

    commit: function () {
      this.commitTransaction();
    },

    willTransition: function (transition) {
      if (this.get('transaction.dirty').length) {
        transition.abort();
        return false;
      }
      return true;
    }
  }
});

var TransactionAwareRoute = Ember.Route.extend(TransactionMixin, { ... });
```

And of course the actual `Transaction` that manages the dirty objects:
```js
var Transaction = Ember.Object.extend({

  deleted: [],
  created: [],
  updated: [],

  dirty: Ember.computed.union('deleted', 'created', 'updated'),

  isDirty: Ember.computed.notEmpty('dirty.[]'),

  addDirty: function (model) {
    var type = get(model, 'isNew') ? 'created' : get(model, 'isDeleted') ? 'deleted' : 'updated';

    if (type === 'deleted') {
      get(this, 'updated').removeObject(model);
    }
    get(this, type).addObject(model);
  },

  removeClean: function (model) {
    get(this, 'deleted').removeObject(model);
    get(this, 'created').removeObject(model);
    get(this, 'updated').removeObject(model);
  },

  commit: function (store) {
    return store.commitRecords(get(this, 'dirty'));
  },

  rollback: function () {
    return get(this, 'dirty').slice().reverse().invoke('rollback');
  }
});
```

Ember Data used to have a transaction api and IMHO it worked pretty well. I was disappointed when it was removed. I don't know if any of this helps, but along with the code in this PR it made working with Ember Data a lot more productive. Without it, I don't think that we could have continued using Ember Data in its current state. There's just too much overhead in manually managing dirty objects on data-heavy pages (again IMHO).

mmpestorich:
@steffenbrem 
> Do you have an idea on how you can update the canonicalState of all relationships of a parent record that has been saved (without including every relationship in the response)? I have tried to fix this, but I lack the knowledge for it.

I am currently looking into this and have written a few tests for it, but haven't come up with a good solution yet. If I figure something out, i'll put it up here.

bmac:
> If it's a matter of more discussion, conforming to an RFC, cleaning up the code, etc... please point me in the right direction. I'll gladly help in anyway that I can. This PR has been sitting out here for over a year now with no feedback from the ember-data team... slightly frustrating. Prior to this, I had pushed up #1631 - which I closed after being told that rolling back relationships would be addressed by the single-source-of-truth branch - ulitmately it wasn't.

> The bottom line is that this PR works and has a handful of passing tests and it's been over 2 years since ember-data was able to rollback relationships (when this behavoir was removed from ember-data, it was sold as a temporary thing). I'd gladly contribute more time and energy to this in order to put something together that you all would feel comfortable accepting... anything I can do please let me know.

Thanks for your comments @mmpestorich. I understand the frustration. I think I may have been wrong to close this pull request. Unfortunately, github wont let me reopen this pr. Do you mind submitting a new one and I will get it on the agenda for the next meeting?

mmpestorich:
@bmac Sorry for the late response, I've been on vacation these past couple weeks. I'd be more than happy to submit this as a new pull request in order to gather more input from the community and your team. I should be able to get something up by tomorrow.

PR #2882, opened 2015-03-15
bump ember-cli to 2.0 & remove sourcemapping comments in production

### LINKED ISSUE #2871: ember-data breaks all production sourcemaps
Any app that includes ember-data will end up with broken production sourcemaps, because `ember-data.prod.js` includes this sourcemap reference:

```js
//# sourceMappingURL=ember-data.js.map
```

which points to a sourcemap file that's only valid for `ember-data.js`, the dev build. This attempt in ember-data's index.js to hide the dev sourcemap in production:

```js
    this.app.import({
      development: app.bowerDirectory + '/ember-data/ember-data.js',
      production: app.bowerDirectory + '/ember-data/ember-data.prod.js'
    }, options);
    // Source maps
    this.app.import({
      development: app.bowerDirectory + '/ember-data/ember-data.js.map'
    }, {destDir: 'assets'});
```

does not work, because sourcemap files do not need to be imported explicitly, they are discovered via relative URLs.

ember-data needs to ship with a valid sourceMappingURL in ember-data.prod.js, or leave it off entirely. Pointing to an invalid file breaks the sourcemap for any file that appears in the vendor bundle later than ember-data.

fivetanley:
fixes #2871, although I think that https://github.com/ember-cli/ember-cli-uglify/issues/4 has not actually been fixed.

cc @stefanpenner @ef4 @bmac

### DIFF
--- a/Brocfile.js
+++ b/Brocfile.js
@@ -30,6 +30,7 @@ function minify(tree, name){
     srcFile: name + '.js',
     destFile: '/' + name + '.prod.js'
   });
+  tree = removeSourceMappingURL(tree);
   var uglified = moveFile(uglify(tree, {mangle: true}),{
     srcFile: name + '.prod.js',
     destFile: '/' + name + '.min.js'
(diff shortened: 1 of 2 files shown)


fivetanley:
I am no longer able to reproduce the error with either ember data or es5-shim... seems good!

stefanpenner:
:+1:

PR #2883, opened 2015-03-15
Pass snapshots to adapters instead of records

wecc:
- [x] Add tests to make sure the store pass snapshots to adapter methods (createRecord, updateRecord, deleteRecord, find, findMany, findHasMany, findBelongsTo)
- [x] Write release notes/breaking change guide. There are deprecation warnings for using `.get()` but a heads-up seems appropriate.
- [X] ~~Go through website/guides and update as needed~~ Couldn't find any related guides

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,50 @@
 
 ### Master
 
+#### Breaking Changes
+
+##### The store now passes snapshots instead of records to adapter methods
+
+In 1.0.0-beta.15 serializers were updated to be passed snapshots instead of
+records to prevent side-effects like fetching data when inspecting
+relationships. This has now been extended to also include adapters methods.
+
+The following adapter methods are now passed snapshots instead of records:
+
+- `find(store, type, id, snapshot)`
+- `findMany(store, type, ids, snapshots)`
+- `findHasMany(store, snapshot, url, relationship)`
+- `findBelongsTo(store, snapshot, url, relationship)`
+- `createRecord(store, type, snapshot)`
+- `updateRecord(store, type, snapshot)`
+- `deleteRecord(store, type, snapshot)`
+
+The signature of `buildURL(type, id, snapshot)` has also been updated to receive
+snapshots instead of records.
+
+This change removes the need for adapters to create snapshots manually using the
+private API `record._createSnapshot()` to be able to pass snapshots to
+serializers.
+
+Snapshots are backwards-compatible with records (with deprecation warnings) and
+it should be pretty straight forward to update current code to the public
+Snapshot API:
+
+```js
+post.get('id')           => postSnapshot.id
+post.get('title')        => postSnapshot.attr('title')
+post.get('author')       => postSnapshot.belongsTo('author')
+post.get('comments')     => postSnapshot.hasMany('comments')
+post.constructor         => postSnapshot.type;
+post.constructor.typeKey => postSnapshot.typeKey
+```
+
+If you need to access the underlying record of a snapshot you can do so by
+accessing `snapshot.record`.
+
+The full API reference of `DS.Snapshot` can be found [here](http://emberjs.com/api/data/classes/DS.Snapshot.html).
+
+
 ### Release 1.0.0-beta.15 (February 14, 2015)
 
 #### Breaking Changes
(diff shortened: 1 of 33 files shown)


bmac:
Currently some custom adapters are calling `record._createSnapshot()` should Ember Data do anything to help these adapters transition (e.g. making `snapshot._createSnapshot` an assert or deprecation?)

wecc:
@bmac We'll probably have to cover these changes thoroughly in the release notes :) I've added a warning if you're trying to call `_createSnapshot` on a DS.Snapshot, see https://github.com/emberjs/data/pull/2883/files#diff-d8a3fdbb89d6cf0bb2b162ed7d3b76ffR360

CHANGELOG.md
```diff
@@ -2,6 +2,51 @@
 
 ### Master
 
+#### Breaking Changes
+
+##### The store no longer passes records to adapter methods
```
igorT: Maybe we want to rephrase this, so people don't freak out

wecc: Updated!

wecc:
Rebased. Fixed merge conflicts.

packages/ember-data/lib/system/snapshot.js
```diff
@@ -355,6 +355,14 @@ Snapshot.prototype = {
...
   unknownProperty: function(keyName) {
     return this.get(keyName);
+  },
+
+  /**
+    @method _createSnapshot
+    @private
+  */
+  _createSnapshot: function() {
+    Ember.assert("You called _createSnapshot on what's already a DS.Snapshot. You shouldn't manually create snapshots in your adapter since the store passes snapshots to adapters by default.");
```
bmac: @wecc Sorry for changing my mind here. I chatted with Igor about this and it might be a nicer migration path for this method to give a deprecation warning and just return the instance of the snapshot itself.

wecc: I agree, that makes sense. Updated with deprecation.

wecc:
I've added 72e5c3e508bdda5ddc5b3deb31951a1904ed3138. Let me know if this is more suited in a separate PR.

packages/ember-data/lib/system/store.js
```diff
@@ -748,8 +748,24 @@ Store = Service.extend({
...
+      // We want to provide records to all store methods and snapshots to all
+      // adapter methods. To make sure we're doing that we're providing an array
+      // of snapshots to adapter.groupRecordsForFindMany(), which in turn will
+      // return grouped snapshots instead of grouped records.
+      //
+      // But since the _findMany() finder is a store method we need to get the
+      // records from the grouped snapshots even though the _findMany() finder
+      // will once again convert the records to snapshots for adapter.findMany()
+
+      var snapshots = Ember.A(map(records, function(record) {
```
igorT: Might be easier to call invoke on the array of records

packages/ember-data/lib/system/store/finders.js
```diff
@@ -41,7 +43,10 @@ export function _find(adapter, store, type, id, record) {
 
 
 export function _findMany(adapter, store, type, ids, records) {
-  var promise = adapter.findMany(store, type, ids, records);
+  var snapshots = Ember.A(map(records, function(record) {
```
igorT: same here

igorT:
:+1:

packages/ember-data/lib/system/store/finders.js
wecc: Changed these to `Ember.A(records).invoke('_createSnapshot');`

nolanlawson:
> Currently some custom adapters are calling `record._createSnapshot()

For ember-pouch, I'm also detecting the `_createSnapshot` method in order to determine if I'm running in a pre-beta15 environment. Definitely a nasty hack. :/

Will there be a recommended way for adapters to to support all three of 14, 15, and 16? Or should we just ditch pre-16 support entirely and bump the major version?

wecc:
@nolanlawson I'm sorry that I don't have any recommendations other than if you want to support 14, 15 and 16 your current solution seems OK. It will work as is in all three version but give a deprecation warning in 16.

If you want to ditch pre-16 that's fine too :metal:

PR #2884, opened 2015-03-15
Remove container deprecation warnings in Ember Data tests

bmac:


packages/ember-data/lib/system/store.js
```diff
@@ -1446,11 +1447,7 @@ Store = Service.extend({
   },
 
   modelFactoryFor: function(key) {
-    if (this.container.has('model:' + key)) {
-      return this.container.lookupFactory('model:' + key);
-    } else {
-      return null;
-    }
+    return this.container.lookupFactory('model:' + key);
```
bmac: @igorT getting rid of the `has` check did not break any tests. You added this for the polymorphic mixins so I wanted to get your eyes on this change.

packages/ember-data/lib/system/store.js
```diff
@@ -1397,10 +1397,11 @@ Store = Service.extend({
   */
 
   _modelForMixin: function(key) {
-    var mixin = this.container.resolve('mixin:' + key);
+    var registry = this.container._registry;
```
bmac: Not super happy with grabbing `_registry` here. Accessing private members like this has caused problems for ED in the past. Is there a better way to do this @dgeb / @stefanpenner?

packages/ember-data/lib/system/store.js
```diff
@@ -1397,10 +1397,11 @@ Store = Service.extend({
   */
 
   _modelForMixin: function(key) {
-    var mixin = this.container.resolve('mixin:' + key);
+    var registry = this.container._registry ? this.container._registry : this.container;
```
bmac: Posting again because Github though my previous comment was resolved: Not super happy with grabbing _registry here. Accessing private members like this has caused problems for ED in the past. Is there a better way to do this @dgeb / @stefanpenner?

PR #2885, opened 2015-03-15
Make _stripIDFromURL() understand URLs like something/:id.format

### LINKED ISSUE #2690: coalesceFindRequests: true got ignored when trying to get a hasMany relationshipts
After upgrading to Ember-data 1.0.0-beta.14, the one request to server from

`parentModel.get(childrenModel)` where just

``` 
parentModel = DS.Model.extend
  childrenModel: DS.hasMany('child', async: true)
```

Ember-data now sends instead of one, n-requests, where n - is a length of childrenModel array of ids. 
I have set `coalesceFindRequests: true` on my adapter.

wecc:
Having URLs like `/posts/1.json` is not an uncommon scenario. This PR makes `_stripIDFromURL()` understand these URLs so coalesced requests can be grouped correctly.

Unsure if RegExp is the right way to go, feedback welcome.

- [ ] Add more asserts to check the actual URLs

Fixes #2690

### DIFF
--- a/packages/ember-data/lib/adapters/rest-adapter.js
+++ b/packages/ember-data/lib/adapters/rest-adapter.js
@@ -591,6 +591,13 @@ export default Adapter.extend(BuildURLMixin, {
     } else if (endsWith(lastSegment, '?id=' + id)) {
       //Case when the url is of the format ...something?id=:id
       expandedURL[expandedURL.length - 1] = lastSegment.substring(0, lastSegment.length - id.length - 1);
+    } else if (expandedURL.length > 1) {
+      // Case when the url is of the format ...something/:id.format
+      var pattern = '^' + id + '\\.\\w+$';
+      if (new RegExp(pattern).test(lastSegment)) {
+        expandedURL[expandedURL.length - 1] = "";
+        expandedURL[expandedURL.length - 2] += lastSegment.substring(id.length, lastSegment.length);
+      }
     }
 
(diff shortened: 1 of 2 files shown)


fivetanley:
I think it would be great if we could reuse the work in tildeio/route-recognizer.

wecc:
@amiel I know you've been looking at URL related libraries in your work with URL Templates, do you have any suggestions on how to handle this a bit nicer?

Basically we want to check if the URL returned from `buildURL()` ends with `.../{id}`, `.../{id}.format` or `...?id={id}` where `{id}` is the only known part. `...` (and `format`) can be anything.

amiel:
@wecc If we were going to include geriantluff/uri-templates, you could use the ["de-substitution"](https://github.com/geraintluff/uri-templates#guess-variables-from-uri-de-substitution) feature, but I think we're planning to make a uri-template library an optional dependency...

wecc:
Thanks @amiel!

I've looked into [geraintluff/uri-templates](https://github.com/geraintluff/uri-templates) and the de-substitution feature would indeed make this a lot cleaner.

The following patterns match exactly what we're looking for:
- `{/path*,id}`
- `{/path*,id}{.format}`
- `{/path*}{?id}`.

@fivetanley do we want/have the technology to include geraintluff/uri-templates in ED?

igorT:
Closing this for now

PR #2888, opened 2015-03-16
Pass Snapshot As Parameter To _canSerialize

ryanpatrickcook:
By passing the snapshot/record to _canSerialize, you can decide which attributes or relationships get serialized based on the state of the record or the attributes in flight.

We would like the ability to perform a partial update for a record on only attributes that are dirty.

### DIFF
--- a/packages/ember-data/lib/serializers/json-serializer.js
+++ b/packages/ember-data/lib/serializers/json-serializer.js
@@ -308,9 +308,10 @@ export default Serializer.extend({
     @method _canSerialize
     @private
     @param {String} key
+    @param {Object} snapshot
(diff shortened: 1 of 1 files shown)


bmac:
Not too crazy about encouraging users to override a private method like `_canSerialize`. Maybe we should drop the underscore and make it public? Thoughts @wycats @igorT?

bmac:
We talked about this pr in the Ember Data weekly meeting today and decided that we don't want to make `_canSerialize` public. Instead we would recommend you override [serializeAttribute](http://emberjs.com/api/data/classes/DS.JSONSerializer.html#method_serializeAttribute). Let me know if you have any questions.

Issue #2806, opened 2015-02-20
DEPRECATION: has should be called on the registry instead of the container


twokul:
Should this [lookup](https://github.com/emberjs/data/blob/f49c2a7fe5fb5483b72999cc8ccffb87d8cf4c91/packages/ember-data/lib/system/store.js#L1416) be replaced with `this.container._registry.lookupFactory` given the changes in Ember introduced in this [PR](https://github.com/emberjs/ember.js/pull/9981)? Otherwise, a lot of deprecation warnings are shown (`DEPRECATION: has should be called on the registry instead of the container`)

bmac:
@twokul Sounds like a good change to make. Although I'm curious if there is a better way to access the registry then `_registry` Ember Data has been bitten before by relying on private apis like that. Would you be interested in making a pr?

igorT:
ping @dgeb

twokul:
@bmac I can look into it.

dgeb:
So `lookup` and `lookupFactory` are available on the container without deprecation. In fact, they should be accessed from the container. 

I would guess that an instance of the registry is being injected into the store as `this.container`. This would still allow `lookup` and `lookupFactory` to work, but with deprecations, since the registry will proxy them to its default container (the first container created by the registry).

I would look into the initializers because they are now being passed a registry instead of a container.

The end game here is to have an Application initializer, which gets passed an Application instead of a registry. You can call `register` and `inject` directly on this, which will flow through to the Application's internal registry.

Furthermore, we will soon also have ApplicationInstance initializer, which will get passed an ApplicationInstance. You'll be able to call `lookup` and `lookupFactory` directly on this, which will flow through to the instance's internal container.

I believe @tomdale is currently working on this and can probably provide more details / status. Hopefully we can provide some workarounds to avoid these deprecations.

bmac:
ping @twokul Do you have time to work on a pr for this?

igorT:
Seems like we don't need to do anything anything here? Because @dgeb said using it on the container is correct? Or maybe I am very confused

bmac:
@igorT it sounds like many of the places in ember data that use the variable `container` are now actually getting assigned a `registry` instance in Ember 1.11.

twokul:
I'll look into this tonight

bmac:
@twokul any progress?

twokul:
[...]

PR #2892, opened 2015-03-17
First pass on deprecations

twokul:
related to #2806

### DIFF
--- a/packages/ember-data/lib/initializers/store.js
+++ b/packages/ember-data/lib/initializers/store.js
@@ -11,15 +11,15 @@ import Store from "ember-data/system/store";
   @param {Ember.Container} container
   @param {Object} [application] an application namespace
 */
-export default function initializeStore(container, application) {
+export default function initializeStore(registry, application) {
   Ember.deprecate('Specifying a custom Store for Ember Data on your global namespace as `App.Store` ' +
                   'has been deprecated. Please use `App.ApplicationStore` instead.', !(application && application.Store));
 
-  container.register('store:main', container.lookupFactory('store:application') || (application && application.Store) || Store);
+  registry.register('store:main', registry.lookupFactory('store:application') || (application && application.Store) || Store);
 
   // allow older names to be looked up
 
-  var proxy = new ContainerProxy(container);
+  var proxy = new ContainerProxy(registry);
   proxy.registerDeprecations([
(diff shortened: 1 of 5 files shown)


packages/ember-data/lib/initializers/store.js
```diff
@@ -27,12 +27,12 @@ export default function initializeStore(container, application) {
...
-  container.register('adapter:-rest', RESTAdapter);
+  registry.register('serializer:-default', JSONSerializer);
+  registry.register('serializer:-rest', RESTSerializer);
+  registry.register('adapter:-rest', RESTAdapter);
 
   // Eagerly generate the store so defaultStore is populated.
   // TODO: Do this in a finisher hook
-  var store = container.lookup('store:main');
-  container.register('service:store', store, { instantiate: false });
+  var store = registry.lookup('store:main');
```
rwjblue: Lookup on a registry will trigger a deprecation.

packages/ember-data/tests/integration/setup-container-test.js
```diff
@@ -63,7 +67,7 @@ test("the deprecated adapter:_rest is resolved as adapter:rest", function() {
     deprecated = container.lookup('adapter:_rest');
   });
 
-  ok(deprecated === valid, "they should resolve to the same thing");
+  ok(deprecated._debugContainerKey === valid._debugContainerKey, "they should resolve to the same thing");
```
rwjblue: Seems odd to compare the key like this, given that it will always just be the value that you passed to `container.lookup` above (making the test not useful).

rwjblue:
This needs to be updated to properly handle all supported Ember versions (not just 1.11+).

packages/ember-data/tests/integration/setup-container-test.js
twokul: comparing keys like this is just wrong, I know. `deprecated` and `valid` are classes so we likely should use `deepEqual`

rwjblue: The idea here is that both versions are the same actual class returned. That is why it used `===` before (and likely should still do).

twokul: I'm not disagreeing with you, still figuring it out.

PR #2895, opened 2015-03-17
Update the InvalidError docs to make it more clear about where the

### LINKED ISSUE #2887: Documention clarification
Here: http://emberjs.com/api/data/classes/DS.InvalidError.html

It says: 
>This function should return the entire payload as received from the server. Error object extraction and normalization of model errors should be performed by extractErrors on the serializer.

The example:
```js
App.ApplicationAdapter = DS.RESTAdapter.extend({
  ajaxError: function(jqXHR) {
    var error = this._super(jqXHR);

    if (jqXHR && jqXHR.status === 422) {
      var jsonErrors = Ember.$.parseJSON(jqXHR.responseText);
      return new DS.InvalidError(jsonErrors);
    } else {
      return error;
    }
  }
});
```

Requiring format:
```js
return new DS.InvalidError({
  length: ['Must be less than 15'],
  name: ['Must not be blank']
});
```

1. What exactly does Ember do with what I return from my overridden `extractErrors`?
2. "Error object extraction and normalization of model errors should be performed by extractErrors on the serializer." makes it sound like I should not be doing that here, but then how do I pass the extracted and normalized errors into `DS.InvalidError` in the format that it expects?
3. If `extractErrors` will extract and normalize the payload that is returned `ajaxError`, assuming the status code is NOT 422 and `DS.InvalidError` is not returned, how do the errors get associated with the model when `DS.InvalidError` is designated as the mechanism for that purpose?
4. "This function should return the entire payload as received from the server." needs clarification if not correction. In either case in the example above, `ajaxError` does not return the entire payload as returned from the server. Nor does `new DS.InvalidError(…)`, of which this documentation is for.
5. This same confusing phrase "This function should return the entire payload as received from the server." appears here as well: http://emberjs.com/api/data/classes/DS.RESTAdapter.html#method_ajaxError
So what is "This function" referring to?

Hopefully these questions help illustrate how confusing these particular docs are. Most Ember docs are great, but this one has me totally confused. Those are the best questions I could come up with, there are more questions I don't even know how to ask. Please clean up these docs.

bmac:
server payload gets normalized.

Closes #2887

@atomkirk I'd like your feedback on these changes to the docs

screen shot 2015-03-17 at 4 01 32 pm

<hr>

screen shot 2015-03-17 at 4 01 40 pm

### DIFF
--- a/packages/ember-data/lib/adapters/rest-adapter.js
+++ b/packages/ember-data/lib/adapters/rest-adapter.js
@@ -671,11 +671,10 @@ export default Adapter.extend(BuildURLMixin, {
 
     Returning a `DS.InvalidError` from this method will cause the
     record to transition into the `invalid` state and make the
-    `errors` object available on the record.
-
-    This function should return the entire payload as received from the
-    server.  Error object extraction and normalization of model errors
-    should be performed by `extractErrors` on the serializer.
+    `errors` object available on the record. When returning an
+    `InvalidError` the store will attempt to normalize the error data
+    returned from the server using the serializer's `extractErrors`
+    method.
 
     Example
 
--- a/packages/ember-data/lib/system/adapter.js
+++ b/packages/ember-data/lib/system/adapter.js
@@ -23,9 +23,39 @@ var errorProps = [
   transition to the `invalid` state and the errors will be set to the
   `errors` property on the record.
 
-  This function should return the entire payload as received from the
-  server.  Error object extraction and normalization of model errors
-  should be performed by `extractErrors` on the serializer.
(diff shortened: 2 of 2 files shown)


jbrown:
@bmac Does this work out of the box with `RESTAdapter` if you structure your JSON a certain way and use a specific response code?

atomkirk:
Yes, thats much better. It makes clear that extractErrors is called before ajaxError is called. Thanks for doing that. The other helpful correction is that it's much more clear why I might want to use extractErrors.

bmac:
Thanks @atomkirk

PR #2896, opened 2015-03-17
Add doc example for defaultValue as a function on DS.attr

### LINKED ISSUE #2833: Return copy of object specified in defaultValue

bmac:
closes #2833

@yokree Does this address your concerns?

### DIFF
--- a/packages/ember-data/lib/system/model/attributes.js
+++ b/packages/ember-data/lib/system/model/attributes.js
@@ -259,6 +259,21 @@ function getValue(record, key) {
   });
   ```
 
+  Default value can also be a function. This is useful it you want to return
+  a new object for each attribute.
+
+  ```javascript
+  var attr = DS.attr;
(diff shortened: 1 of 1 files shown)


yokree:
@bmac Yeah, nice to see more detailed explanation in docs, cheers.

PR #2898, opened 2015-03-18
Pass requestType to buildURL

amiel:
This is another step toward emberjs/rfcs#4.

Others have requested ways to handle different urls per request type (see emberjs/data#1078).

Once url-templates lands, most users extending `BuildURLMixin` will not need to handle
`requestType` explicitly, as it will be possible to define a template for each request type.

### DIFF
--- a/packages/ember-data/lib/adapters/build-url-mixin.js
+++ b/packages/ember-data/lib/adapters/build-url-mixin.js
@@ -14,7 +14,7 @@ var get = Ember.get;
   ```javascript
   export default DS.Adapter.extend(BuildURLMixin, {
     find: function(store, type, id, snapshot) {
-      var url = this.buildURL(type.typeKey, id, snapshot);
+      var url = this.buildURL(type.typeKey, id, snapshot, 'find');
       return this.ajax(url, 'GET');
     }
   });
@@ -45,9 +45,10 @@ export default Ember.Mixin.create({
     @param {String} type
     @param {String|Array} id single id or array of ids
     @param {DS.Snapshot|Array} snapshot single snapshot or array of snapshots
+    @param {String} requestType
     @return {String} url
   */
-  buildURL: function(type, id, snapshot) {
+  buildURL: function(type, id, snapshot, requestType) {
     var url = [];
     var host = get(this, 'host');
     var prefix = this.urlPrefix();
--- a/packages/ember-data/lib/adapters/rest-adapter.js
+++ b/packages/ember-data/lib/adapters/rest-adapter.js
@@ -344,7 +344,7 @@ export default Adapter.extend(BuildURLMixin, {
     @return {Promise} promise
   */
   find: function(store, type, id, snapshot) {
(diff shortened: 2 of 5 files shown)


igorT:
Can you rebase? Seems like it's not merging cleanly

amiel:
Rebased, sorry about that.

packages/ember-data/lib/adapters/build-url-mixin.js
```diff
@@ -13,9 +13,15 @@ var get = Ember.get;
...
   ```javascript
   export default DS.Adapter.extend(BuildURLMixin, {
-    find: function(store, type, id, record) {
-      var url = this.buildURL(type.typeKey, id, record);
-      return this.ajax(url, 'GET');
+    createRecord: function(store, type, record) {
+      var data = {};
+      var serializer = store.serializerFor(type.typeKey);
+      var snapshot = record._createSnapshot();
+      var url = this.buildURL(type.typeKey, null, snapshot, 'createRecord');
```
wecc: The code for `createRecord` pass `'create'` to buildURL, but in this example it's `'createRecord'`, typo?

amiel: Good catch. Which do you think it should be?

wecc: I think it makes sense to keep it consistent with the method names?

amiel: Ok, I'll go with that.

I think I was thinking that the method names were an implementation detail, but I think the consistency would be worth it.

Thanks for your input :)

amiel: Ok, I've changed these all to `createRecord`, `updateRecord`, etc

packages/ember-data/lib/adapters/rest-adapter.js
```diff
@@ -362,13 +362,15 @@ export default Adapter.extend(BuildURLMixin, {
...
   findAll: function(store, type, sinceToken) {
-    var query;
+    var query, url;
 
     if (sinceToken) {
       query = { since: sinceToken };
     }
 
-    return this.ajax(this.buildURL(type.typeKey), 'GET', { data: query });
+    url = this.buildURL(type.typeKey, undefined, undefined, 'findAll');
```
bmac: Could these missing arguments be `null` instead of `undefined`? This will make it more consistent with https://github.com/emberjs/data/pull/2898/files#diff-24089ab4caa677e410cbc5c13588c012L533

bmac:
:+1:

packages/ember-data/lib/adapters/rest-adapter.js
amiel: Done. Good idea

amiel:
I've made changes suggested by @bmac, but I need to rebase again.

amiel:
@igorT this should be ready now.

bmac:
Thanks @amiel!

amiel:
Thank you!

igorT:
:+1:

PR #2904, opened 2015-03-19
Implement basic JSON-API adapter

wecc:
A basic working JSON-API adapter.

### DIFF
--- a/packages/ember-data/lib/adapters/json-api-adapter.js
+++ b/packages/ember-data/lib/adapters/json-api-adapter.js
@@ -1,0 +1,84 @@
+/**
+  @module ember-data
+*/
+
+import RESTAdapter from "ember-data/adapters/rest-adapter";
+
+/**
+  @class JSONAPIAdapter
+  @constructor
+  @namespace DS
+  @extends DS.RESTAdapter
+*/
+export default RESTAdapter.extend({
+  defaultSerializer: '-json-api',
+
+  /**
+    @method ajaxOptions
+    @private
+    @param {String} url
+    @param {String} type The request type GET, POST, PUT, DELETE etc.
+    @param {Object} options
+    @return {Object}
+  */
+  ajaxOptions: function(url, type, options) {
+    let hash = this._super(...arguments);
+
+    if (hash.contentType) {
+      hash.contentType = 'application/vnd.api+json';
+    }
+
+    let beforeSend = hash.beforeSend;
+    hash.beforeSend = function(xhr) {
+      xhr.setRequestHeader('Accept', 'application/vnd.api+json');
+      if (beforeSend) {
+        beforeSend(xhr);
+      }
+    };
+
+    return hash;
+  },
+
+  /**
+    @method findMany
+    @param {DS.Store} store
+    @param {DS.Model} type
+    @param {Array} ids
+    @param {Array} snapshots
+    @return {Promise} promise
+  */
+  findMany: function(store, type, ids, snapshots) {
+    var url = this.buildURL(type.modelName, ids, snapshots, 'findMany');
+    return this.ajax(url, 'GET', { data: { filter: { id: ids.join(',') } } });
+  },
+
+  /**
+    @method pathForType
+    @param {String} modelName
+    @return {String} path
+  **/
+  pathForType: function(modelName) {
+    var dasherized = Ember.String.dasherize(modelName);
+    return Ember.String.pluralize(dasherized);
+  },
+
+  // TODO: Remove this once we have a better way to override HTTP verbs.
+  /**
+    @method updateRecord
+    @param {DS.Store} store
+    @param {DS.Model} type
+    @param {DS.Snapshot} snapshot
+    @return {Promise} promise
+  */
+  updateRecord: function(store, type, snapshot) {
+    var data = {};
+    var serializer = store.serializerFor(type.modelName);
+
+    serializer.serializeIntoHash(data, type, snapshot, { includeId: true });
+
+    var id = snapshot.id;
+    var url = this.buildURL(type.modelName, id, snapshot, 'updateRecord');
+
+    return this.ajax(url, 'PATCH', { data: data });
+  }
+});
--- a/packages/ember-data/lib/serializers/json-api-serializer.js
+++ b/packages/ember-data/lib/serializers/json-api-serializer.js
@@ -1,0 +1,310 @@
+/**
+  @module ember-data
+*/
+
+import JSONSerializer from 'ember-data/serializers/json-serializer';
+import normalizeModelName from 'ember-data/system/normalize-model-name';
+import { pluralize, singularize } from 'ember-inflector/lib/system/string';
+
+var dasherize = Ember.String.dasherize;
+var map = Ember.EnumerableUtils.map;
+
+/**
+  @class JSONAPISerializer
+  @namespace DS
+  @extends DS.JSONSerializer
+*/
+export default JSONSerializer.extend({
+
+  /*
+    @method _normalizeRelationshipDataHelper
+    @param {Object} relationshipDataHash
+    @return {Object}
+    @private
+  */
+  _normalizeRelationshipDataHelper: function(relationshipDataHash) {
+    let type = this.modelNameFromPayloadKey(relationshipDataHash.type);
+    relationshipDataHash.type = type;
+    return relationshipDataHash;
+  },
+
+  /*
+    @method _normalizeResourceHelper
+    @param {Object} resourceHash
+    @return {Object}
+    @private
+  */
+  _normalizeResourceHelper: function(resourceHash) {
+    let modelName = this.modelNameFromPayloadKey(resourceHash.type);
+    let modelClass = this.store.modelFor(modelName);
+    let serializer = this.store.serializerFor(modelName);
+    let { data } = serializer.normalize(modelClass, resourceHash);
+    return data;
+  },
+
+  /**
+    @method _normalizeResponse
+    @param {DS.Store} store
+    @param {DS.Model} primaryModelClass
+    @param {Object} payload
+    @param {String|Number} id
+    @param {String} requestType
+    @param {Boolean} isSingle
+    @return {Object} JSON-API Document
+    @private
+  */
+  _normalizeResponse: function(store, primaryModelClass, payload, id, requestType, isSingle) {
+
+    if (Ember.typeOf(payload.data) === 'object') {
+      payload.data = this._normalizeResourceHelper(payload.data);
+    } else {
+      payload.data = map(payload.data, this._normalizeResourceHelper, this);
+    }
+
+    if (Ember.typeOf(payload.included) === 'array') {
+      payload.included = map(payload.included, this._normalizeResourceHelper, this);
+    }
+
+    return payload;
+  },
+
+  /*
+    @method extractAttributes
+    @param {DS.Model} modelClass
+    @param {Object} resourceHash
+    @return {Object}
+  */
+  extractAttributes: function(modelClass, resourceHash) {
+    var attributes = {};
+
+    if (resourceHash.attributes) {
+      modelClass.eachAttribute((key) => {
+        let attributeKey = this.keyForAttribute(key, 'deserialize');
+        if (resourceHash.attributes.hasOwnProperty(attributeKey)) {
+          attributes[key] = resourceHash.attributes[attributeKey];
+        }
+      });
+    }
+
+    return attributes;
+  },
+
+  /*
+    @method extractRelationship
+    @param {Object} relationshipHash
+    @return {Object}
+  */
+  extractRelationship: function(relationshipHash) {
+
+    if (Ember.typeOf(relationshipHash.data) === 'object') {
+      relationshipHash.data = this._normalizeRelationshipDataHelper(relationshipHash.data);
+    }
+
+    if (Ember.typeOf(relationshipHash.data) === 'array') {
+      relationshipHash.data = map(relationshipHash.data, this._normalizeRelationshipDataHelper, this);
+    }
+
+    return relationshipHash;
+  },
+
+  /*
+    @method extractRelationships
+    @param {Object} modelClass
+    @param {Object} resourceHash
+    @return {Object}
+  */
+  extractRelationships: function(modelClass, resourceHash) {
+    let relationships = {};
+
+    if (resourceHash.relationships) {
+      modelClass.eachRelationship((key, relationshipMeta) => {
+        let relationshipKey = this.keyForRelationship(key, relationshipMeta.kind, 'deserialize');
+        if (resourceHash.relationships.hasOwnProperty(relationshipKey)) {
+
+          let relationshipHash = resourceHash.relationships[relationshipKey];
+          relationships[key] = this.extractRelationship(relationshipHash);
+
+        }
+      });
+    }
+
+    return relationships;
+  },
+
+  /*
+    @method extractType
+    @param {DS.Model} modelClass
+    @param {Object} resourceHash
+    @return {String}
+    @private
+  */
+  _extractType: function(modelClass, resourceHash) {
+    return this.modelNameFromPayloadKey(resourceHash.type);
+  },
+
+  /**
+    @method modelNameFromPayloadKey
+    @param {String} key
+    @return {String} the model's modelName
+  */
+  modelNameFromPayloadKey: function(key) {
+    return singularize(normalizeModelName(key));
+  },
+
+  /**
+    @method payloadKeyFromModelName
+    @param {String} modelName
+    @return {String}
+  */
+  payloadKeyFromModelName: function(modelName) {
+    return pluralize(modelName);
+  },
+
+  /*
+    @method normalize
+    @param {DS.Model} modelClass
+    @param {Object} resourceHash
+    @return {String}
+  */
+  normalize: function(modelClass, resourceHash) {
+    this.normalizeUsingDeclaredMapping(modelClass, resourceHash);
+
+    let data = {
+      id:            this.extractId(resourceHash),
+      type:          this._extractType(modelClass, resourceHash),
+      attributes:    this.extractAttributes(modelClass, resourceHash),
+      relationships: this.extractRelationships(modelClass, resourceHash)
+    };
+
+    this.applyTransforms(modelClass, data.attributes);
+
+    return { data };
+  },
+
+  /**
+   @method keyForAttribute
+   @param {String} key
+   @param {String} method
+   @return {String} normalized key
+  */
+  keyForAttribute: function(key, method) {
+    return dasherize(key);
+  },
+
+  /**
+   @method keyForRelationship
+   @param {String} key
+   @param {String} typeClass
+   @param {String} method
+   @return {String} normalized key
+  */
+  keyForRelationship: function(key, typeClass, method) {
+    return dasherize(key);
+  },
+
+  /**
+    @method serialize
+    @param {DS.Snapshot} snapshot
+    @param {Object} options
+    @return {Object} json
+  */
+  serialize: function(snapshot, options) {
+    let data = this._super(...arguments);
+    data.type = this.payloadKeyFromModelName(snapshot.modelName);
+    return { data };
+  },
+
+  /**
+   @method serializeAttribute
+   @param {DS.Snapshot} snapshot
+   @param {Object} json
+   @param {String} key
+   @param {Object} attribute
+  */
+  serializeAttribute: function(snapshot, json, key, attribute) {
+    var type = attribute.type;
+
+    if (this._canSerialize(key)) {
+      json.attributes = json.attributes || {};
+
+      var value = snapshot.attr(key);
+      if (type) {
+        var transform = this.transformFor(type);
+        value = transform.serialize(value);
+      }
+
+      var payloadKey =  this._getMappedKey(key);
+      if (payloadKey === key) {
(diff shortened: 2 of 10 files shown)


packages/ember-data/lib/adapters/json-api-adapter.js
```diff
@@ -0,0 +1,188 @@
...
+  deleteRecord: function(store, type, record) {
+    var id = get(record, 'id');
+
+    return this.ajax(this.buildURL(type.typeKey, id, record), "DELETE");
+  },
+
+
+  pathForType: function(typeKey) {
+    return Ember.String.dasherize(typeKey);
+  },
```
bmac: :+1:

packages/ember-data/lib/serializers/json-api-serializer.js
```diff
@@ -0,0 +1,294 @@
...
+    return Ember.String.dasherize(key);
+  },
+
+  keyForRelationship: function(key, kind) {
+    return Ember.String.dasherize(key);
+  },
+
+  normalizeTypeKey: function(typeKey) {
+    return Ember.String.camelize(typeKey);
+  },
```
bmac: Could this be changed to `Ember.String. dasherize(typeKey)`? If that doesn't work today I expect it will be something we want to do in the future as we move to dasherized type keys internality.

packages/ember-data/lib/serializers/json-api-serializer.js
```diff
@@ -0,0 +1,294 @@
...
+  @extends DS.Serializer
+*/
+export default Serializer.extend({
+
+  keyForAttribute: function(key) {
+    return Ember.String.dasherize(key);
+  },
+
+  keyForRelationship: function(key, kind) {
+    return Ember.String.dasherize(key);
```
bmac: I believe JSONAPI dropped the recommendation that these keys be dasherized?

For both `keyForAttribute` and `keyForRelationship` I think we should no-op these methods and allow users to override them to best suit their api.

packages/ember-data/lib/adapters/json-api-adapter.js
```diff
@@ -0,0 +1,188 @@
...
+
+  updateRecord: function(store, type, record) {
+    var serializer = store.serializerFor(type.typeKey);
+
+    var snapshot = record._createSnapshot();
+    var data = serializer.serialize(snapshot, { includeId: true });
+
+    var id = get(record, 'id');
+
+    return this.ajax(this.buildURL(type.typeKey, id, record), "PUT", { data: data });
```
bmac: jsonapi has changed updates to always use a `PATCH` verb http://jsonapi.org/format/#crud-updating

packages/ember-data/lib/serializers/json-api-serializer.js
```diff
@@ -0,0 +1,294 @@
...
+  serializeHasMany: function(snapshot, json, relationship) {
+    var key = relationship.key;
+    var hasMany = snapshot.hasMany(key);
+
+    var links = json['links'] = json['links'] || {};
+
+    var payloadKey = this.keyForRelationship(key, 'hasMany');
+
+    if (hasMany.length === 0) {
+      links[payloadKey] = { ids: [] };
```
bmac: `ids` -> `id` https://github.com/json-api/json-api/pull/392

packages/ember-data/lib/serializers/json-api-serializer.js
```diff
@@ -0,0 +1,294 @@
...
+        link = hash.links[payloadKey];
+
+        if (relationship.kind === 'belongsTo' && link.id) {
+
+          hash[key] = { id: link.id, type: this.normalizeTypeKey(link.type) };
+          delete hash.links[payloadKey];
+
+        } else if (relationship.kind === 'hasMany') {
+
+          if (link.ids) {
```
bmac: `ids` -> `id` https://github.com/json-api/json-api/pull/392

packages/ember-data/lib/serializers/json-api-serializer.js
nwjsmith: Nope, dasherized keys are still recommended http://jsonapi.org/recommendations/

bmac: Thanks @nwjsmith, I missed that.

packages/ember-data/lib/serializers/json-api-serializer.js
wecc: I think that will fail for global apps right now? I'm a bit unsure of changing this before #2821 gets merged.

bmac: My main concern is I'd like to get the default serialization of the `type` property correct now, rather then have to change it in the future. Ideally, this would not be coupled to globals vs modules versions of an Ember App.

packages/ember-data/lib/serializers/json-api-serializer.js
```diff
@@ -0,0 +1,294 @@
...
+    var links = json['links'] = json['links'] || {};
+
+    var payloadKey = this.keyForRelationship(key, 'hasMany');
+
+    if (hasMany.length === 0) {
+      links[payloadKey] = { ids: [] };
+    } else {
+
+      // TODO: if all items in hasMany is of the same type, provide
+      // { type: "type", ids: [..] } instead of { data: [...] }
```
nwjsmith: I don't think this `TODO` is valid as of https://github.com/json-api/json-api/pull/447. Additionally, the key is `linkage` now, instead of `data`

tests/ember-configuration.js
```diff
@@ -119,11 +122,11 @@
...
-    DS.JSONSerializer.reopen({
-      transformFor: function(attributeType) {
-        return this._super(attributeType, true) || transforms[attributeType];
-      }
-    });
+    var transformFor = function(attributeType) {
+      return this._super(attributeType, true) || transforms[attributeType];
+    };
+    DS.JSONSerializer.reopen({ transformFor: transformFor });
+    DS.JSONAPISerializer.reopen({ transformFor: transformFor });
```
chnn: Looks like `transformFor` is never defined on the serializer and adapter, but the tests pass because of this patching (would not work in an actual app). I think?

packages/ember-data/lib/serializers/json-api-serializer.js
```diff
@@ -0,0 +1,295 @@
...
+var get = Ember.get;
+var forEach = Ember.ArrayPolyfills.forEach;
+var map = Ember.ArrayPolyfills.map;
+
+/**
+  @class JSONAPISerializer
+  @namespace DS
+  @extends DS.Serializer
+*/
+export default Serializer.extend({
```
bmac: Should this extend from `JSONSerializer` so it can inherit the `transformFor` method?

wwwdata:
hi, this is still in the RC3 format and not the final specification of jsonapi 1.0 right? Are you planning to finish this for 1.0? Can I help out in any way? Please give a small status update, thanks :)

wecc:
@wwwdata I'm hard at work updating this to both 1.0 and the new Serializer API refactor we did recently. I hope I can update this PR later tonight with the new code! Happy to have an extra set of :eyes: reviewing it!

wwwdata:
@wecc awesome! I will review it carefully and test it against my backend once you push the new commits

wecc:
@wwwdata ping

packages/ember-data/lib/adapters/json-api-adapter.js
```diff
@@ -0,0 +1,80 @@
...
+/**
+  @class JSONAPIAdapter
+  @constructor
+  @namespace DS
+  @extends DS.RESTAdapter
+*/
+export default RESTAdapter.extend({
+  defaultSerializer: '-json-api',
+
+  concatenatedProperties: ['headers'],
```
igorT: If we are gonna do this, should add it to the rest adapter, but it seems like they can also be a CP as per the comments in the rest adapter, so might not work

packages/ember-data/lib/adapters/json-api-adapter.js
```diff
@@ -0,0 +1,80 @@
...
+    var url = this.buildURL(type.modelName, ids, snapshots, 'findMany');
+    return this.ajax(url, 'GET', { data: { filter: { id: ids.join(',') } } });
+  },
+
+  /**
+    @method pathForType
+    @param {String} modelName
+    @return {String} path
+  **/
+  pathForType: function(modelName) {
```
igorT: don't need this

packages/ember-data/lib/serializers/json-api-serializer.js
```diff
@@ -0,0 +1,135 @@
...
+    @param {DS.Store} store
+    @param {DS.Model} primaryModelClass
+    @param {Object} payload
+    @param {String|Number} id
+    @param {String} requestType
+    @param {Boolean} isSingle
+    @return {Object} JSON-API Document
+    @private
+  */
+  _normalizeResponse: function(store, primaryModelClass, payload, id, requestType, isSingle) {
```
igorT: don't need this

packages/ember-data/lib/serializers/json-api-serializer.js
```diff
@@ -0,0 +1,135 @@
...
+          payloadKey = this.keyForRelationship(key, 'belongsTo', 'serialize');
+        }
+
+        let data;
+        if (belongsTo) {
+          data = {
+            type: belongsTo.modelName,
+            id: belongsTo.id
+          };
+        } else {
```
igorT: don't need this, can just say = null on 89

packages/ember-data/lib/serializers/json-api-serializer.js
```diff
@@ -0,0 +1,135 @@
...
+  /**
+   @method serializeHasMany
+   @param {DS.Snapshot} snapshot
+   @param {Object} json
+   @param {Object} relationship
+  */
+  serializeHasMany: function(snapshot, json, relationship) {
+    var key = relationship.key;
+
+    if (this._canSerialize(key)) {
```
igorT: Should change a bit probably with slys PR

packages/ember-data/lib/serializers/json-api-serializer.js
```diff
@@ -0,0 +1,135 @@
...
+
+        json.relationships = json.relationships || {};
+
+        var payloadKey = this._getMappedKey(key);
+        if (payloadKey === key && this.keyForRelationship) {
+          payloadKey = this.keyForRelationship(key, 'hasMany', 'serialize');
+        }
+
+        let data = hasMany.map((item) => {
+          return {
```
igorT: can rewrite without return and {

igorT:
:+1:

bjubinville:
I can't express the amount of awesome there is in the ember community!  Thanks for all the amazing work. :+1:

packages/ember-data/tests/integration/adapter/json-api-adapter-test.js
```diff
@@ -0,0 +1,831 @@
...
+        id: '5',
+        attributes: {
+          text: 'This is the second comment'
+        }
+      }]
+    }]);
+
+    run(function() {
+      store.find('post').then(function(posts) {
+        equal(passedUrl[0], '/post');
```
wwwdata: the url should be /posts here. They are always pluralized.

SeyZ:
Awesome work @wecc @wwwdata @igorT ! :tada:

packages/ember-data/tests/integration/adapter/json-api-adapter-test.js
```diff
@@ -0,0 +1,831 @@
...
+        id: '1',
+        attributes: {
+          title: 'Ember.js rocks'
+        }
+      }
+    }]);
+
+    run(function() {
+      store.find('post', 1).then(function(post) {
+        equal(passedUrl[0], '/post/1');
```
pangratz: @wwwdata same here as down below: this should be `/posts/1`?

wwwdata: yes

packages/ember-data/lib/serializers/json-api-serializer.js
```diff
@@ -0,0 +1,135 @@
...
+      json.attributes = json.attributes || {};
+
+      var value = snapshot.attr(key);
+      if (type) {
+        var transform = this.transformFor(type);
+        value = transform.serialize(value);
+      }
+
+      var payloadKey =  this._getMappedKey(key);
+      if (payloadKey === key && this.keyForAttribute) {
```
bmac: Do we need to check `this.keyForAttribute` if we know the `JSONSerializer` has it defined?

packages/ember-data/tests/integration/adapter/json-api-adapter-test.js
pangratz: Ok, so @wecc  it seems that most/all URLs in the tests are singular at the moment ...

wecc: This is currently intentional as JSON-API does not state anything about URLs being singular/plural. Right now we only pass through from whatever your model names are. If your model is named `Post` the URL will be `/post/1`. If it's named `Posts` the URL will be `/posts/1`.

I'm very open to give Ember Data stronger opinions on this. Like plural for URLs and types and dasherized keys in payloads to match the examples from http://jsonapi.org/. There simply needs to be a decision made. I will bring this up at our meeting later today.

packages/ember-data/lib/serializers/json-api-serializer.js
wecc: forEach and `data.push()`?

packages/ember-data/tests/integration/adapter/json-api-adapter-test.js
pangratz: You are completely right @wecc. Thanks for the heads up!

packages/ember-data/lib/adapters/json-api-adapter.js
```diff
@@ -0,0 +1,84 @@
...
+    @param {String} url
+    @param {String} type The request type GET, POST, PUT, DELETE etc.
+    @param {Object} options
+    @return {Object}
+  */
+  ajaxOptions: function(url, type, options) {
+    let hash = this._super(...arguments);
+
+    if (hash.contentType) {
+      hash.contentType = 'application/vnd.api+json; charset=utf-8';
```
nonsensery: I don't think parameters, like `charset=utf-8` are allowed by the spec right now. See [content negotiation](http://jsonapi.org/format/#content-negotiation).

packages/ember-data/tests/integration/adapter/json-api-adapter-test.js
wwwdata: hmm, but it is very confusing to name a single model in plural like `posts`. If we do not want to enforce a pluralized url we need some mechanism to configure the url I guess. On the jsonapi.org website, all the URLs are plural. A lot of backends that generate an API with URLs will generate them in that recommended format, so it should be easy for an ember-data user to use this recommended url scheme.

SeyZ: Totally agree with the inflection confusion. For me:

- Url should be plural: `/api/posts` or `/api/posts/5` (REST compliant)
- Model should be singular: `Post`
- the `type` key in a jsonapi payload can be singular or plural (according to the spec). In my opinion, we need an option in the JSONAPIAdapter to specify the singular or plural form.

packages/ember-data/lib/adapters/json-api-adapter.js
wecc: Thanks!

packages/ember-data/lib/serializers/json-api-serializer.js
wecc: We do not, updated!

packages/ember-data/tests/integration/adapter/json-api-adapter-test.js
wecc: @pangratz @wwwdata @SeyZ this PR has now been updated to match the format on jsonapi.org as closely as possible.

- **types** are serialized to, and expected to be **plural dasherized** when received from the server
- **member names** are serialized to, and expected to be **dasherized** when received from the server

However, the internal JSON-API format for ED has **singular dasherized** for types and **camelized** for member names. This should not be an issue as the `JSONAPISerializer` normalizes the format mentioned above to the internal ED format automatically.

All of these will be easily overridden using hooks (`modelNameFromPayloadKey`, `payloadKeyFromModelName`, `extractType`, `keyForAttribute`, `keyForRelationship`...)

URLs are now also built to be **plural dasherized** by default.

pangratz: :heart_eyes: awesome work @wecc!

packages/ember-data/lib/adapters/json-api-adapter.js
```diff
@@ -0,0 +1,85 @@
...
+    @param {String} modelName
+    @return {String} path
+  **/
+  pathForType: function(modelName) {
+    var dasherized = Ember.String.dasherize(modelName);
+    return Ember.String.pluralize(dasherized);
+  },
+
+  /**
+    TODO: Remove this once we have a better way to override HTTP verbs.
```
bmac: Lets move this comment out of the doc block.

packages/ember-data/lib/serializers/json-api-serializer.js
```diff
@@ -0,0 +1,299 @@
...
+    @param {Object} payload
+    @param {String|Number} id
+    @param {String} requestType
+    @param {Boolean} isSingle
+    @return {Object} JSON-API Document
+    @private
+  */
+  _normalizeResponse: function(store, primaryModelClass, payload, id, requestType, isSingle) {
+
+    let normalizeResourceHelper = (resourceHash) => {
```
fivetanley: we should probably move this to the prototype to avoid re-creating a closure every time.

packages/ember-data/lib/serializers/json-api-serializer.js
```diff
@@ -0,0 +1,299 @@
...
+    };
+
+    if (Ember.typeOf(payload.data) === 'object') {
+      payload.data = normalizeResourceHelper(payload.data);
+    } else {
+      payload.data = payload.data.map(normalizeResourceHelper);
+    }
+
+    if (Ember.typeOf(payload.included) === 'array') {
+      payload.included = payload.included.map(normalizeResourceHelper);
```
fivetanley: need to use Ember.EnumerableUtils if it's available, otherwise this will break IE8

packages/ember-data/lib/serializers/json-api-serializer.js
```diff
@@ -0,0 +1,299 @@
...
+
+  /**
+    @method serialize
+    @param {DS.Snapshot} snapshot
+    @param {Object} options
+    @return {Object} json
+  */
+  serialize: function(snapshot, options) {
+    let data = this._super(...arguments);
+    data.type = pluralize(snapshot.modelName);
```
bmac: Should this use `this.payloadKeyFromModelName` instead of `pluralize`?

packages/ember-data/lib/serializers/json-api-serializer.js
fivetanley: needs EnumerableUtils for IE8.

fivetanley:
Overall looks pretty good. You, bmac, and igor did all the hard work which is why this is a relatively small change :dancer:

packages/ember-data/lib/serializers/json-api-serializer.js
```diff
@@ -0,0 +1,313 @@
...
+    return relationships;
+  },
+
+  /*
+    @method extractType
+    @param {DS.Model} modelClass
+    @param {Object} resourceHash
+    @return {String}
+  */
+  extractType: function(modelClass, resourceHash) {
```
igorT: underscore

packages/ember-data/lib/serializers/json-api-serializer.js
```diff
@@ -0,0 +1,313 @@
...
+  /*
+    @method normalize
+    @param {DS.Model} modelClass
+    @param {Object} resourceHash
+    @return {String}
+  */
+  normalize: function(modelClass, resourceHash) {
+    let data = null;
+
+    if (resourceHash) {
```
igorT: remove

wwwdata:
this is merged now, which is awesome but did anybody actually use this adapter against his backend and can help me? When I want to load entities from my backend, the request is made, I can see that in chrome debug tools but I always have an empty array instead of a list of my `question` entity.

I am doing

```js
this.store.findAll('question');
```

in my route

question model

```js
import DS from 'ember-data';

  text: DS.attr('string')
});
```

the server returns this very minimalistic json

```json
{
    "data": [
        {
            "attributes": {
                "text": "wow?"
            },
            "id": "ddb72a88-3004-4c30-a14c-71f50084233b",
            "type": "questions"
        }
    ]
}	
```

I am using ember-cli and registered the JSONAPIAdapter in my application adapter and the JSONAPISerializer  as application serializer. 

What could I have possibly done wrong that this does not work?

wecc:
@wwwdata I'm able to replicate that the JSONSerializer doesn't automatically use the new Serializer API when extending. Can you confirm if it works when you change your application serializer to:

```js
  isNewSerializerAPI: true
});
```

wwwdata:
aaah, thank you so much, yes that was the problem!

wecc:
@wwwdata thank you for discovering and confirming! This should be fixed with #3375 (as in not needing to provide `isNewSerializerAPI: true`)

PR #2907, opened 2015-03-20
Make all model lookups go through the container

fivetanley:
move factories off constructor and delegate to the container for lookups

This commit changes the way factories are located by Ember Data.
Previously, factories were stuffed onto the constructor of a
DS.Model subclass.

This was problematic for a few reasons:

- Classes for relationships defined via hasMany(Object) (instead of
  hasMany('string-name') were different than what was in the container.
  This caused several issues for Ember CLI and Globals Mode users alike.
- Inconsistency with the rest of the way lookups work in Ember.
- Zalgo

Now, relationship information can be accessed off the "type" in the
relationships "meta" hash. If you need to lookup a factory in your app,
you need to use this.store.modelFor, or this.store.modelFactoryFor if
you prefer to check that the model exists without throwing an exception.
This change makes all model factory lookups go through the container.
This should hopefully kill `Ember.MODEL_FACTORY_INJECTIONS = true;`.


fivetanley:
cc @igorT @mixonic @stefanpenner

packages/ember-data/lib/serializers/json-serializer.js
```diff
@@ -158,13 +158,14 @@ export default Serializer.extend({
   */
   normalize: function(type, hash) {
     if (!hash) { return hash; }
+    var model = this.store.modelFor(type);
```
fivetanley: not sure at this point if serializers should receive type or typeKey.

packages/ember-data/lib/system/relationships/state/belongs-to.js
```diff
@@ -60,14 +60,18 @@ BelongsToRelationship.prototype._super$addRecord = Relationship.prototype.addRec
...
+  var modelClass = this.store.modelFor(type);
+  Ember.assert("You cannot add a '" + newRecord.constructor.typeKey + "' record to the '" + this.record.constructor.typeKey + "." + this.key +"'. " + "You can only add a '" + modelClass.typeKey + "' record to this relationship.", (function () {
+    if (modelClass.__isMixin) {
+      return modelClass.__mixin.detect(newRecord);
     }
-    if (Ember.MODEL_FACTORY_INJECTIONS) {
-      type = type.superclass;
+
+    // TODO: There seems to be a weird cylic error with loader js here.
+    if (modelClass.superclass && modelClass.superclass !== DS.Model) {
```
fivetanley: Need to talk to stef here. This inheritance check is wacky. Not sure what to check for wrt injections.

packages/ember-data/lib/system/relationships/state/has-many.js
```diff
@@ -84,12 +85,12 @@ ManyRelationship.prototype.removeRecordFromOwn = function(record, idx) {
...
 ManyRelationship.prototype.notifyRecordRelationshipAdded = function(record, idx) {
-  var type = this.relationshipMeta.type;
-  Ember.assert("You cannot add '" + record.constructor.typeKey + "' records to the " + this.record.constructor.typeKey + "." + this.key + " relationship (only '" + this.belongsToType.typeKey + "' allowed)", (function () {
+  var type = this.store.modelFor(this.relationshipMeta.type);
+  Ember.assert("You cannot add '" + record.constructor.typeKey + "' records to the " + this.record.constructor.typeKey + "." + this.key + " relationship (only '" + this.belongsToType + "' allowed)", (function () {
     if (type.__isMixin) {
       return type.__mixin.detect(record);
     }
-    if (Ember.MODEL_FACTORY_INJECTIONS) {
+    if (type.superclass && type.superclass !== DS.Model) {
```
fivetanley: wacky inheritance check here too.

packages/ember-data/lib/system/store.js
```diff
@@ -1421,12 +1421,14 @@ Store = Service.extend({
   */
   modelFor: function(key) {
     var factory;
+    var singularizedKey;
 
     if (typeof key === 'string') {
-      factory = this.modelFactoryFor(key);
+      singularizedKey = singularize(key);
```
fivetanley: this can go away since modelFactoryFor normalizes.

packages/ember-data/lib/system/relationships/state/belongs-to.js
stefanpenner: i believe once we have this commit inplace, we can begin to remove then `extend` in the container, which breaks alot of inheritance checks, This that i believe we can clean ^^ up.

stefanpenner:
This looks great.

so to confirm, no more instances (like store) now exist on the factories?

fivetanley:
@bmac I'm removing this from the milestone. It needs some battletesting before we bring it in.

packages/ember-data/lib/serializers/json-serializer.js
igorT: While we are at it, can we maybe figure out a better name than model for the actual class. Something like using type & typeKey or typeClass. Model implies an actual record, very easy to get confused

packages/ember-data/lib/adapters/rest-adapter.js
```diff
@@ -502,7 +502,7 @@ export default Adapter.extend(BuildURLMixin, {
   */
   findBelongsTo: function(store, record, url, relationship) {
     var id   = get(record, 'id');
-    var type = record.constructor.typeKey;
+    var type = store.modelFor(record.constructor.typeKey);
```
stefanpenner: `record.constructor` is the type

packages/ember-data/lib/system/relationships/ext.js
```diff
@@ -277,9 +278,9 @@ Model.reopenClass({
       var possibleRelationships = relationshipsSoFar || [];
 
       var relationshipMap = get(inverseType, 'relationships');
-      if (!relationshipMap) { return; }
+      if (!relationshipMap) { return possibleRelationships; }
 
-      var relationships = relationshipMap.get(type);
+      var relationships = relationshipMap.get(type.typeKey);
```
stefanpenner: we should try and make `type` always the string and if its something else call it the factory/modelConstructor or something

packages/ember-data/lib/system/relationship-meta.js
```diff
@@ -8,7 +8,7 @@ export function typeForRelationshipMeta(store, meta) {
     if (meta.kind === 'hasMany') {
       typeKey = singularize(typeKey);
     }
-    type = store.modelFor(typeKey);
```
stefanpenner: `store` is now no longer needed as an argument to this function. which means  `relatedTypesDescriptor`  and `relationshipsByNameDescriptor` should no longer need the store either.

packages/ember-data/lib/system/many-array.js
```diff
@@ -226,7 +226,9 @@ export default Ember.Object.extend(Ember.MutableArray, Ember.Evented, {
     var type = get(this, 'type');
     var record;
 
-    Ember.assert("You cannot add '" + type.typeKey + "' records to this polymorphic relationship.", !get(this, 'isPolymorphic'));
+    var klass = store.modelFactoryFor(type);
+
+    Ember.assert("You cannot add '" + klass.typeKey + "' records to this polymorphic relationship.", !get(this, 'isPolymorphic'));
```
kball: Do we know that in this situation modelFactoryFor will never return null?  Because if not, this will break...

packages/ember-data/lib/system/store.js
```diff
@@ -1445,8 +1447,9 @@ Store = Service.extend({
   },
 
   modelFactoryFor: function(key) {
-    if (this.container.has('model:' + key)) {
-      return this.container.lookupFactory('model:' + key);
+    var singularized = singularize(key);
```
kball: In some places we're using singularized, and other places fully normalizing with this._normalizeTypeKey.  Shouldn't we normalize the same way throughout?

packages/ember-data/lib/system/many-array.js
fivetanley: Yeah, modelFactoryFor won't raise an exception. modelFor will.

fivetanley: Oh wait, I read that wrong. You sir, are correct!

packages/ember-data/lib/system/store.js
fivetanley: That is a good point! looking into it now. Thanks for taking the time to review! Makes me more confident that we'll catch more things before this gets merged in.

packages/ember-data/lib/system/relationships/ext.js
```diff
@@ -76,8 +76,8 @@ var relationshipsByNameDescriptor = Ember.computed(function() {
   this.eachComputedProperty(function(name, meta) {
     if (meta.isRelationship) {
       meta.key = name;
-      var relationship = relationshipFromMeta(this.store, meta);
-      relationship.type = typeForRelationshipMeta(this.store, meta);
+      var relationship = relationshipFromMeta(meta);
+      relationship.type = typeForRelationshipMeta(meta);
```
kball: I believe this is extraneous.  Looking at relationshipFromMeta it appears to already define type using typeForRelationshipMeta

packages/ember-data/lib/system/relationships/ext.js
```diff
@@ -207,6 +207,7 @@ Model.reopenClass({
     @method inverseFor
     @static
     @param {String} name the name of the relationship
+    @param {DS.Store} store the current store
```
kball: I don't think this comment belongs here...

packages/ember-data/lib/system/relationships/state/has-many.js
```diff
@@ -84,12 +85,12 @@ ManyRelationship.prototype.removeRecordFromOwn = function(record, idx) {
 };
 
 ManyRelationship.prototype.notifyRecordRelationshipAdded = function(record, idx) {
-  var type = this.relationshipMeta.type;
-  Ember.assert("You cannot add '" + record.constructor.typeKey + "' records to the " + this.record.constructor.typeKey + "." + this.key + " relationship (only '" + this.belongsToType.typeKey + "' allowed)", (function () {
+  var type = this.store.modelFor(this.relationshipMeta.type);
```
kball: Should this also use typeForRelationshipMeta instead of trying to directly access the type?

packages/ember-data/lib/system/relationships/state/belongs-to.js
kball: I'm new to Ember Data so I don't really understand the backstory here, but is the situation of multiple layers of inheritance supported?  And if so, does that cause issues?  Because this appears to only explicitly handle one layer (both before and now)

E.g.
```
var Person = DS.Model.extend(/*..define some relationship..*/)
var Employee = Person.extend(/*...*/)
var Contractor = Employee.extend(/*...*/)
```

fivetanley:
ping @igorT @stefanpenner

packages/ember-data/lib/system/relationships/state/belongs-to.js
fivetanley: Hey, sorry for the late reply. Yah, that use case should be and is currently supported.

fivetanley:
you can try this out by using the following in your bower.json file:

```javascript
{
  "ember-data": "components/ember-data#always-lookup-models-through-container"
}
```

fivetanley:
@stefanpenner i see you are reading your email :P

stefanpenner:
i'll do one pass now, then another later :)

stefanpenner:
got distracted by es7 decorators, will look later :P

packages/ember-data/lib/serializers/rest-serializer.js
```diff
@@ -282,7 +282,7 @@ var RESTSerializer = JSONSerializer.extend({
 
       // legacy support for singular resources
       if (isPrimary && Ember.typeOf(value) !== "array" ) {
-        primaryRecord = this.normalize(primaryType, value, prop);
+        primaryRecord = this.normalize(type, value, prop);
```
igorT: Why this change? Shouldn't this be primary type?

packages/ember-data/lib/system/relationship-meta.js
```diff
@@ -1,14 +1,14 @@
...
   var typeKey, type;
 
   typeKey = meta.type || meta.key;
   if (typeof typeKey === 'string') {
     if (meta.kind === 'hasMany') {
-      typeKey = singularize(typeKey);
+      typeKey = normalizeTypeKey(typeKey);
     }
-    type = store.modelFor(typeKey);
+    type = typeKey;
```
igorT: what is going on here, didn't we standardize the naming?

packages/ember-data/lib/serializers/rest-serializer.js
fivetanley: i think so, yeah. i'll change it back.

packages/ember-data/lib/system/relationships/ext.js
```diff
@@ -179,7 +178,7 @@ Model.reopenClass({
   */
   typeForRelationship: function(name) {
     var relationship = get(this, 'relationshipsByName').get(name);
-    return relationship && relationship.type;
+    return relationship && this.store.modelFor(relationship.type);
```
stefanpenner: this `store` shouldn't exist, as the goal of this work is to remove instance state from the factories. `this` here I believe is the factory.

packages/ember-data/lib/system/normalize-type-key.js
```diff
@@ -0,0 +1,17 @@
+import {singularize} from 'ember-inflector/lib/system/string';
+
+var camelize = Ember.String.camelize;
```
fivetanley: not sure why this came up in the diff.

packages/ember-data/lib/system/many-array.js
```diff
@@ -230,7 +230,13 @@ export default Ember.Object.extend(Ember.MutableArray, Ember.Evented, {
     var type = get(this, 'type');
     var record;
 
-    Ember.assert("You cannot add '" + type.typeKey + "' records to this polymorphic relationship.", !get(this, 'isPolymorphic'));
+    var klass = store.modelFactoryFor(type);
```
mixonic: It is pretty spooky that what `type` contains is unknown this far into the code. Shouldn't this normalization step happen earlier, like on `init`?

packages/ember-data/lib/system/relationships/ext.js
```diff
@@ -273,13 +272,13 @@ Model.reopenClass({
       inverseKind = possibleRelationships[0].kind;
     }
 
-    function findPossibleInverses(type, inverseType, relationshipsSoFar) {
+    function findPossibleInverses(modelClass, inverseType, relationshipsSoFar) {
```
mixonic: the factory was called `typeClass` earlier, and here is `modelClass`. Should probably pick one.

Issue #2916, opened 2015-03-22
should serializerForAdapter take the store as an argument?


bmac:
Before https://github.com/emberjs/data/pull/2617 the helpers in https://github.com/emberjs/data/blob/4205061a1ce72a61a01dab0f34c2df4ac25f6341/packages/ember-data/lib/system/store/serializers.js did not need access to the store. Now they both need to use the store. `serializerFor` got moved back onto the store while `serializerForAdapter` stayed as a helper method. This feels a little bit inconsistent. Should `serializerForAdapter` be moved onto the store or should `serializerFor` be refactored back into a helper that receives `store` as an argument.

jmurphyau:
@bmac I've expanded on the lists that were created in #2617 below to give a clearer overview of the current situation:

## Current preference/order

### Adapters:

1 `adapterFor`: `typeKey`
2 `adapterFor`: `'application'`
3 `defaultAdapter`: `this.adapter` (default to `-rest`)
4 `defaultAdapter`: `'application'`
5 `defaultAdapter`: `'-rest'`

### Serializers:

1 `serializerFor`: `typeKey`
2 `serializerFor`: `'application'`
3 `serializerFor`: `adapterFor(typeKey).defaultSerializer` (this could return either the type specific adapter, the application wide adapter or the -rest adapter)
4 `serializerFor`: `'-default'`

### `serializerForAdapter` function

1 `serializerForAdapter `: `adapter.serializer`
2 `serializerFor`: `typeKey`
3 `serializerFor`: `'application'`
4 `serializerFor`: `adapterFor(typeKey).defaultSerializer` (this could return either the type specific adapter, the application wide adapter or the -rest adapter)
5 `serializerFor`: `'-default'`

So with these lists in mind - serializerForAdapter is only used to access the adapters `serializer` property as a first preference, then goes through the standard `serializerFor` logic.

`serializerForAdapter` is used [here](https://github.com/emberjs/data/blob/174fc94488d7e66ebb8b61e3bef142c0d051edca/packages/ember-data/lib/system/store/finders.js#L17) and [here](https://github.com/emberjs/data/blob/174fc94488d7e66ebb8b61e3bef142c0d051edca/packages/ember-data/lib/system/store.js#L1996). I think the best bet would be to change both of these from:

`var serializer = serializerForAdapter(store, adapter, type);`

to

`var serializer = adapater.serializer || store.serializerFor(type);`

jmurphyau:
[...]

PR #2918, opened 2015-03-23
Remove serializerForAdapter

jmurphyau:
`adapater.serializer || store.serializerFor(type)` is used in place

Seems like a logical way to close issue #2916.


bmac:
@stefanpenner could you review this (and #2916 for context) since I believe the helpers in `packages/ember-data/lib/system/store/serializers.js` where a result of you suggestion?

igorT:
@stefanpenner thought it's cleaner to keep them separate to simplify store.js and that it's ok to pass store in. I don't feel too strongly about it, so seems alright by me. @jmurphyau thanks for the PR though

PR #2919, opened 2015-03-23
Serializer & Adapter Lookups Preference/Order

jmurphyau:
I've put together this PR after the discussion in PR #2617. Although nothing was decided I figure if a PR exists and defines something to start with and a place for discussion it can start there?

No tests have broken but thats likely because no tests exist for the purpose of serializer/adapter lookup order/preference..

The commits have been ordered in a particular way and the messages help explain the overall changes in this PR in smaller pieces - so having a look at each commit might give more of an understanding.

### Current Preference/Order

1 `adapterFor`: `'adapter:[typeKey]'`
2 `adapterFor`: `'adapter:application'`
3 `defaultAdapter`: `'adapter:[this.adapter]'` (`this.adapter` default to `'-rest'`)
4 `defaultAdapter`: `'adapter:application'`
5 `defaultAdapter`: `'adapter:-rest'`

### Proposed Preference/Order

1 `adapterFor`: `'adapter:[typeKey]'`
2 `defaultAdapter`: `'adapter:[this.adapter]'` (`this.adapter` default to `null`)
3 `defaultAdapter`: `'adapter:application'`
4 `defaultAdapter`: `'adapter:-default'`

- A `-default` adapter has been added, which is registered on the registry as the RESTAdapter.

Adapters largely retains the same lookup order but in a clearer to understand way. I can see two scenarios that break:

##### 1. When a user has defined `store.adapter` to something other than `'-rest'` and they have an `application` adapter

By defaulting `store.adapter` to `null` we keep the same order but break users who have explicitly defined `store.adapter` in subclasses.

For example if someone has defined `store.adapter` as `'custom'`, the result of this change would be:

*Current:* `adapter:[typekey]`, `adapter:application`, `adapter:custom`, 

*Proposed:* `adapter:[typekey]`, `adapter:custom`, `adapter:application`

### Current Preference/Order

1 `serializerFor`: `'serializer:[typeKey]'`
2 `serializerFor`: `'serializer:application'`
3 `serializerFor`: `'serializer:[defaultSerializer]'` using the `defaultSerializer` property of the adapter returned in `adapterFor` (could either be: `'adapter:[typeKey]'`, `'adapter:application'` or `'adapter:-rest'`)
4 `serializerFor`: `'serializer:-default'`

### Proposed Preference/Order

1 `serializerFor`: `serializer:[typeKey]`
2 `serializerFor`: `'serializer:[defaultSerializer]'` using the `defaultSerializer` property of `'adapter:[typeKey]'` (if `'adapter:[typeKey]'` exists)
3 `defaultSerializer`: `'serializer:[this.serializer]'` (default to `null`)
4 `defaultSerializer`: `'serializer:application'`
5 `defaultSerializer`: `'serializer:-default'`

Serializers are where the most changes are happening:

- `store.defaultSerializer` and `store.serializer` have been added for consistency with the adapter lookup
- `store.serializerFor` is more explicit in which adapter is used when retrieving the `defaultSerializer` property - it no longer uses `adapterFor` and therefor would not return the 'default' adapters (such as `application` or `-rest`)

##### 1. Where `serializer:[typeKey]` does not exist, `serializer:application` is defined and the `defaultSerializer` from `adapter:[typeKey]` is defined and is not `serializer:application` 

For example if you have lookup 'superVillain' and you have `serializer:application` and `adapter:superVillain.defaultSerializer` is set to `serializer:aliens`

*Current:* `serializer:superVillain` (fail), `serializer:application` (found), `serializer:aliens` 

*Proposed:* `serializer:superVillain` (fail), `serializer:aliens` (found), `serializer:application`

There are a few others that are quite convoluted and I started typing them out but deleted them. They are mainly around the use of the `defaultSerializer` property on the 'default' adapter (`application` and `-rest`)

### DIFF
--- a/packages/ember-data/lib/initializers/store.js
+++ b/packages/ember-data/lib/initializers/store.js
@@ -32,6 +32,7 @@ export default function initializeStore(registry, application) {
   // new go forward paths
   registry.register('serializer:-default', JSONSerializer);
   registry.register('serializer:-rest', RESTSerializer);
+  registry.register('adapter:-default', RESTAdapter);
   registry.register('adapter:-rest', RESTAdapter);
 
   // Eagerly generate the store so defaultStore is populated.
--- a/packages/ember-data/lib/system/store.js
+++ b/packages/ember-data/lib/system/store.js
@@ -216,19 +216,19 @@ Store = Service.extend({
   /**
     The adapter to use to communicate to a backend server or other persistence layer.
 
-    This can be specified as an instance, class, or string.
+    This must be specified as a string.
 
-    If you want to specify `App.CustomAdapter` as a string, do:
+    If you want to specify `App.CustomAdapter`, do:
 
     ```js
     adapter: 'custom'
     ```
 
     @property adapter
-    @default DS.RESTAdapter
-    @type {DS.Adapter|String}
+    @default null
+    @type {String}
   */
-  adapter: '-rest',
+  adapter: null,
 
   /**
     Returns a JSON representation of the record using a custom
@@ -253,13 +253,6 @@ Store = Service.extend({
     This property returns the adapter, after resolving a possible
     string key.
 
-    If the supplied `adapter` was a class, or a String property
-    path resolved to a class, this property will instantiate the
-    class.
-
-    This property is cacheable, so the same instance of a specified
-    adapter class should be used for the lifetime of the store.
-
     @property defaultAdapter
     @private
     @return DS.Adapter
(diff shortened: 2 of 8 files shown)


jmurphyau:
Turns out there are a whole heap of tests that fail - `npm test`, `ember test` and `ember test -s` didn't show me - `ember server` then http://localhost:4200/ did though.. Thought I was really lucky but no..

I'll continue on this tomorrow unless overnight discussion leads to it's closure.

bmac:
Currently with Ember Data a 3rd party Adapter can specify its own 3rd party serializer as a default. However the user can override this 3rd party serializer globally by setting an application adapter. It seems like these changes would break that workflow.

jmurphyau:
Wow it's been a while since I've looked at this.. 

Are you specifically talking about how in `Serializers: Current Preference/Order`, item 2 (`serializer:application`) overrides item 3 (`serializer:[defaultSerializer]` where `defaultSerializer` is the property on the following (in order): 1. `adapter:[typeKey]`, 2. `adapter:application`, 3. `adapter:-rest`)

Your sentence said `the user can override this 3rd party serializer globally by setting an application adapter` but I'm thinking this is a typo and you meant to say application serializer?

If I've understood correctly, than yes that workflow will break. Is it common?

This is something that could also be deprecated if thats a way to go? For example as a second lookup for a serializer it would still check and return the `application` serializer if it exists but log a deprecation if future behaviour will instead return a different serializer.

igorT:
@bmac @fivetanley whats the status here, closing?

bmac:
Hi @jmurphyau. Sorry for the late response. Yes I was talking about the workflow you identified. 

I think it is a common workflow. A quick look at some popular 3rd party adapters shows them all using this pattern. [localforage](https://github.com/genkgo/ember-localforage-adapter/blob/master/addon/adapters/localforage.js#L7), [sails](https://github.com/huafu/ember-data-sails/blob/master/addon/adapters/sails-base.js#L32), [firebase](https://github.com/firebase/emberfire/blob/master/addon/adapters/firebase.js#L38) and [active-model-adapter](https://github.com/ember-data/active-model-adapter/blob/master/addon/active-model-adapter.js#L113).

Unfortunately, because of semver I think we missed the opportunity to make this change so I'm going to close this pr. Please feel free to re-open if you disagree.

PR #2920, opened 2015-03-23
Revert reserved properties assertion

bmac:


igorT:
maybe we should keep the original commit, but just remove some of the props, while we work towards migrating people off of them

bmac:
What would be a reasonable subset of properties to restrict?

igorT:
Not sure, I would guess we don't want to restrict state ones, errors, at least. Also clientId doesn't seem to be actually used anywhere?

PR #2932, opened 2015-03-24
Fix doc for Snapshot.attributes()

### LINKED ISSUE #2923: Snapshot doc say snapshot.attributes() returns an array, but sample indicates object
Doc: http://emberjs.com/api/data/classes/DS.Snapshot.html#method_attributes
Source: https://github.com/emberjs/data/blob/5a47172a36e3a71951f9f0fda46f7b3ffc064de8/packages/ember-data/lib/system/snapshot.js#L146

I think the return type is supposed to be Object instead of Array.

tp:
Method returns an object, not an array.

This will fix #2923

### DIFF
--- a/packages/ember-data/lib/system/snapshot.js
+++ b/packages/ember-data/lib/system/snapshot.js
@@ -143,7 +143,7 @@ Snapshot.prototype = {
     ```
 
     @method attributes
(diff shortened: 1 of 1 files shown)


wecc:
Thanks @tp!

PR #2934, opened 2015-03-24
Fix deprecation warning in store test

wecc:
`Using store.fetch() has been deprecated. Use store.fetchById for fetching individual records or store.fetchAll for collections`


bmac:
Thanks

PR #2935, opened 2015-03-24
Use ember-inflector 1.5

### LINKED ISSUE #2927: ember data 1.0.0-beta.16 with ember 1.10.0-beta.4 fails
Using the newly released beta.16  with ember 1.10.0-beta.4 is not working because of a few lines like this:

Ember.HTMLBars._registerHelper 

 I am guessing its because this method ( _registerHelper ) on HTMLBars is only in ember 1.11.0 ?? 
because in my environment, Ember.HTMLBars exists, just not that method.

I am using npm: 
 ember-cli@0.2.0
 ember-cli-htmlbars@0.7.4

and bower: 
ember#1.10.0-beta.4 
handlebars#3.0.0

bmac:
Closes #2927 

cc @rwjblue @stefanpenner


Brocfile.js
```diff
@@ -89,10 +90,26 @@ function package(packagePath, vendorPath) {
   });
 }
 
+function packageAddon(packagePath, vendorPath) {
+  return renameFiles(pickFiles(vendorPath + packagePath, {
+    files: [ '**/*.js' ],
```
stefanpenner: `stew.rename` might be simpler: https://github.com/stefanpenner/broccoli-stew/blob/master/lib/rename.js#L4-L16

bmac: Will do

stefanpenner: if you run into problems with it, let me know I'll fix them.

PR #2936, opened 2015-03-24
Introduce relationship.hasData

wecc:
**This is a first stab at making relationships aware if data exists for the relationship or not.**

There's currently no way of knowing if the contents of a relationship is _unknown_, _empty_ or _set_ just by looking at a snapshot (which is what serializers must do). This PR lays the groundwork of having snapshots returning an _unknown_ state for calls to `.belongsTo()` and `.hasMany()` when we don't know the contents of the relationship yet.

_This should most likely be reviewed and thought through at least once before we merge it._

### DIFF
--- a/packages/ember-data/lib/system/store.js
+++ b/packages/ember-data/lib/system/store.js
@@ -331,6 +331,10 @@ Store = Service.extend({
     // Set the properties specified on the record.
     record.setProperties(properties);
 
+    record.eachRelationship(function(key, descriptor) {
+      record._relationships[key].setHasData(true);
+    });
+
     return record;
   },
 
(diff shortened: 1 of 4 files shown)


packages/ember-data/tests/integration/relationships/has-many-test.js
```diff
@@ -1312,3 +1314,95 @@ test("adding and removing records from hasMany relationship #2666", function() {
...
+
+  run(function() {
+    store.find('chapter', 1).then(function(chapter) {
+      var relationship = chapter._relationships['pages'];
+      equal(relationship.hasData, true, 'relationship has data');
+    });
+  });
+});
+
+test("hasMany hasData sync loaded", function () {
```
igorT: This doesn't seem like a sync relationship, is the title wrong?

packages/ember-data/tests/integration/relationships/has-many-test.js
```diff
@@ -1312,3 +1314,95 @@ test("adding and removing records from hasMany relationship #2666", function() {
...
+
+  run(function() {
+    store.find('chapter', 1).then(function(chapter) {
+      var relationship = chapter._relationships['pages'];
+      equal(relationship.hasData, false, 'relationship does not have data');
+    });
+  });
+});
+
+test("hasMany hasData sync not loaded", function () {
```
igorT: same here

packages/ember-data/lib/system/store.js
```diff
@@ -331,6 +331,10 @@ Store = Service.extend({
     // Set the properties specified on the record.
     record.setProperties(properties);
 
+    record.eachRelationship(function(key, descriptor) {
+      record._relationships[key].setHasData(true);
```
pangratz: The `true` argument is not used within the `relationship`s `setHasData` method ...

wecc: @pangratz this is intentional, "future proofing"... we should probably add the parameter to `setHasData()` method though, even if we currently only support setting it to true.

PR #2939, opened 2015-03-25
Snapshots unknown relationships

wecc:
This is the second part of solving the problem described in #2936.

Snapshots are now aware of unknown relationships and can return `undefined` from `belongsTo()` and `hasMany()` if that is the case.

~~Ping @igorT re: f67ab9c13cd4cf20deb494a85826aa2468636a28~~

### DIFF
--- a/packages/ember-data/lib/system/relationships/state/relationship.js
+++ b/packages/ember-data/lib/system/relationships/state/relationship.js
@@ -85,6 +85,7 @@ Relationship.prototype = {
       }
     }
     this.flushCanonicalLater();
+    this.setHasData(true);
   },
 
   removeCanonicalRecords: function(records, idx) {
@@ -125,6 +126,7 @@ Relationship.prototype = {
       }
       this.record.updateRecordArraysLater();
     }
+    this.setHasData(true);
   },
 
   removeRecord: function(record) {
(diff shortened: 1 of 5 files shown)


packages/ember-data/lib/system/relationships/state/relationship.js
```diff
@@ -228,12 +228,19 @@ Relationship.prototype = {
...
   },
 
   notifyRecordRelationshipAdded: Ember.K,
   notifyRecordRelationshipRemoved: Ember.K,
 
-  setHasData: function() {
+  setHasData: function(value) {
+    var inverseKey = this.inverseKey;
+    if (inverseKey) {
+      this.canonicalMembers.forEach(function(member) {
```
igorT: This breaks if your inverse is a many relationship, because even though you only added one record, you are saying it has all of them. Not really sure what to do about this though. Also this seems like the wrong place, probably should happen automatically as the setHasData gets called by addCanonicalRecord

igorT: After consulting with @stefanpenner we think it's ok to just send what we have

packages/ember-data/lib/system/relationships/state/relationship.js
```diff
@@ -228,12 +230,13 @@ Relationship.prototype = {
...
     self.computeChanges(records);
+    self.setHasData(true);
   },
 
   notifyRecordRelationshipAdded: Ember.K,
   notifyRecordRelationshipRemoved: Ember.K,
 
-  setHasData: function() {
+  setHasData: function(value) {
     this.hasData = true;
```
fivetanley: seems like we should do this.hasData = value instead

wecc: :+1:

Issue #2844, opened 2015-03-09
Allow empty/null type path ?


vinilios:
The current default behaviour of rest adapter buildURL method appends the result of pathForType for the given type and joins it along with other params in order to construct the final resource request url. In case pathForType returns an empty or null value (this could be a case for an API which serve a resource results from the root url) the resulted url will contain an additional slash causing issues if such urls are not handled server-side.

```javascript
[host, typePath, id].join("/") // result is host.com/api/v1//id for null/empty typePath
```

Is this a case that you would consider ember-data should deal with ? The change is trivial, i could provide a PR including the corresponding tests, but i'm not sure what kind of side-effects and/or backwards compatible issues such a change might cause.

bmac:
@vinilios if you'd like to submit a pr for this feature then go for it.

vinilios:
do i have to branch ?

bmac:
Closed by #2946.

PR #2946, opened 2015-03-27
Handle null/empty type paths in build url mixin

vinilios:
handle empty values returned by pathForType in buildURL method of
the build url mixin.

refs emberjs/data#2844

### DIFF
--- a/packages/ember-data/lib/adapters/build-url-mixin.js
+++ b/packages/ember-data/lib/adapters/build-url-mixin.js
@@ -52,8 +52,12 @@ export default Ember.Mixin.create({
     var url = [];
     var host = get(this, 'host');
     var prefix = this.urlPrefix();
+    var path;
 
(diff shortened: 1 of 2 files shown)


packages/ember-data/lib/adapters/build-url-mixin.js
```diff
@@ -52,8 +52,11 @@ export default Ember.Mixin.create({
     var url = [];
     var host = get(this, 'host');
     var prefix = this.urlPrefix();
+    var path;
 
-    if (type) { url.push(this.pathForType(type)); }
+    if (type && (path = this.pathForType(type))) {
```
igorT: Can you take the assignment out of the if statement? I don't think it's needed there.

igorT:
Thanks for the PR!

vinilios:
PR updated @igorT

igorT:
Thanks!

PR #2949, opened 2015-03-27
Fix building steps in README

hibariya:
`npm run dist` was `ember build --environment=production`.


bmac:
Good catch. Thanks @hibariya!

PR #2943, opened 2015-03-27
DirtyState.invalid handle pushedData event

turboMaCk:
I think that this should by handled similar as in uncommited.
It's currently causing this error:

```javascript
route: xz Attempted to handle event `pushedData` on <app@model:class::instance> while in state root.loaded.updated.invalid.
```

### DIFF
--- a/packages/ember-data/lib/system/model/states.js
+++ b/packages/ember-data/lib/system/model/states.js
@@ -345,7 +345,7 @@ var DirtyState = {
 
     rolledBack: function(record) {
       get(record, 'errors').clear();
-      record.triggerLater('ready');
+      record.transitionTo('loaded.saved');
     },
 
     becameValid: function(record) {
--- a/packages/ember-data/lib/system/model/states.js
+++ b/packages/ember-data/lib/system/model/states.js
@@ -336,6 +336,8 @@ var DirtyState = {
 
     becomeDirty: Ember.K,
 
+    pushedData: Ember.K,
+
     willCommit: function(record) {
       get(record, 'errors').clear();
       record.transitionTo('inFlight');
--- a/packages/ember-data/lib/system/model/states.js
+++ b/packages/ember-data/lib/system/model/states.js
@@ -345,7 +345,13 @@ var DirtyState = {
 
     rolledBack: function(record) {
       get(record, 'errors').clear();
-      record.transitionTo('loaded.saved');
+
+      if (record.get('isNew')) {
+        DirtyState.uncommitted.rollback.apply(this, arguments);
+        record.transitionTo('deleted.saved');
+      } else {
+        record.transitionTo('loaded.saved');
+      }
     },
 
     becameValid: function(record) {
--- a/packages/ember-data/tests/unit/model/merge-test.js
+++ b/packages/ember-data/tests/unit/model/merge-test.js
@@ -102,6 +102,31 @@ test("When a record is dirty, pushes are overridden by local changes", function(
   equal(person.get('city'), "Portland", "if there are no local changes, the new data applied");
 });
 
+test("When a record is invalid, pushes are overridden by local changes", function() {
+  var store = createStore({ adapter: DS.Adapter });
+  var person;
+
+  run(function() {
+    person = store.push(Person, { id: 1, name: "Brendan McLoughlin", city: "Boston" });
+    person.set('name', "Brondan McLoughlin");
+    person.send('becameInvalid');
+  });
+
+  equal(person.get('isDirty'), true, "the person is currently dirty");
+  equal(person.get('isValid'), false, "the person is currently invalid");
+  equal(person.get('name'), "Brondan McLoughlin", "the update was effective");
+  equal(person.get('city'), "Boston", "the original data applies");
+
+  run(function() {
+    store.push(Person, { id: 1, name: "bmac", city: "Prague" });
+  });
+
+  equal(person.get('isDirty'), true, "the local changes are reapplied");
+  equal(person.get('isValid'), false, "record is still invalid");
+  equal(person.get('name'), "Brondan McLoughlin", "the local changes are reapplied");
+  equal(person.get('city'), "Prague", "if there are no local changes, the new data applied");
+});
+
 test("A record with no changes can still be saved", function() {
   expect(1);
 
--- a/packages/ember-data/tests/unit/model/rollback-test.js
+++ b/packages/ember-data/tests/unit/model/rollback-test.js
@@ -215,10 +215,12 @@ test("invalid record can be rollbacked", function() {
 
   run(function() {
     dog.save().then(null, async(function() {
+      equal(dog.get('isValid'), false);
       dog.rollback();
 
       equal(dog.get('name'), "Pluto");
       ok(dog.get('isValid'));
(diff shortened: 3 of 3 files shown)


turboMaCk:
Sorry but I've found one more issue with rollbacking invalid records. It should not be dirty after rollback, righ?

turboMaCk:
Check [this](https://github.com/github/hub) tool for using git + github api. And [there](http://stackoverflow.com/questions/4528869/how-do-you-attach-a-new-pull-request-to-an-existing-issue-on-github) is stackoverflow thread about this feature.

turboMaCk:
OT: @bmac Don't you know if there is some way to use my fork with ember-cli?

devinus:
I'm having this problem as well. But how does this happen when my code does `model.get('errors').clear()` by itself?

devinus:
I actually don't understand at all how I'm getting this error when I use `this.store.find('type')`. I've cleared the errors on my model and I've even tried doing `this.store.unloadAll('type')` and I'm still getting this.

turboMaCk:
@devinus The problem is in root state maneger of ember data. Cleaning errors doesn't change record state which cause this issue. Try to use my fork of ember-data https://github.com/turboMaCk/data until this will be merged.

@bmac Can you please review this pull request and merge it?

devinus:
@turboMaCk Is there any workaround until it's merged? Why can't I forcefully change the models state? I thought `model.get('errors').clear()` would change the state.

devinus:
For anybody else coming to this, you unfortunately might have to do what I did for now:

```js
  beforeModel() {
    // https://github.com/emberjs/data/pull/2943
    this.store.unloadAll('type');
  },

  model() {
    return this.store.find('type');
  }
```

turboMaCk:
@devinus I don't want to hack this in application layer. For mo It make more sense to use custom build of ember data and than switch to official one after this will be merged – no need to changes app code. And I'm also not sure if unloadAll do not cause revert on all record in collection which I do not want to happened. I want to keep all state as it and only handle fetching on invalid records right way.

packages/ember-data/lib/system/model/states.js
```diff
@@ -343,7 +345,7 @@ var DirtyState = {
 
     rolledBack: function(record) {
       get(record, 'errors').clear();
-      record.triggerLater('ready');
+      record.transitionTo('loaded.saved');
```
pangratz: This will break the case when a new record is saved, has errors and is then rolled back. By transitioning into the `loaded.saved` state, this record is now marked as created, but it should be deleted instead...

Failing test case:

``` javascript
test("invalid new record can be rollbacked", function() {
  var person;
  var adapter = DS.RESTAdapter.extend({
    ajax: function(url, type, hash) {
      var adapter = this;

      return new Ember.RSVP.Promise(function(resolve, reject) {
        /* If InvalidError is passed back in the reject it will throw the
           exception which will bubble up the call stack (crashing the test)
           instead of hitting the failure route of the promise.
           So wrapping the reject in an Ember.run.next makes it so save
           completes without failure and the failure hits the failure route
           of the promise instead of crashing the save. */
        Ember.run.next(function() {
          reject(adapter.ajaxError({ name: 'is invalid' }));
        });
      });
    },

    ajaxError: function(jqXHR) {
      return new DS.InvalidError(jqXHR);
    }
  });

  env = setupStore({ person: Person, adapter: adapter });

  run(function() {
    person = env.store.createRecord('person', { id: 1 });
  });

  equal(person.get('isNew'), true, "must be new");
  equal(person.get('isDirty'), true, "must be dirty");

  run(function() {
    person.save().then(null, async(function() {
      equal(person.get('isValid'), false);
      person.rollback();

      // assertions taken from the test 'new record can be rollbacked' http://git.io/vTsKR
      equal(person.get('isNew'), false, "must not be new");
      equal(person.get('isDirty'), false, "must not be dirty");
      equal(person.get('isDeleted'), true, "must be deleted");
    }));
  });
});
```

pangratz: Actually it looks like the above test fails on current `master` as well. I will investigate and suspect the test setup is not correct.

But still, getting the state of the new, invalid person after it is rolled back via `person.currentState.stateName` is `root.loaded.saved`, which should be `root.deleted.saved`...

pangratz: > or the behavior of rolling back an invalid record is currently not supported

This is indeed the current behavior: rolling back a new record which is marked as invalid puts the record in the `loaded.created.uncommitted` state. Is this the intended behavior? There are currently no tests for this case in [`rollback-test.js`](https://github.com/emberjs/data/blob/master/packages/ember-data/tests/unit/model/rollback-test.js) so it needs to be discussed what should be the intended behavior here. Sorry for hijacking this PR for this issue, I will gladly open a new one if this is desired ...

devinus:
Anything that allows me to use the store after I have a `DS.InvalidError` would be welcome. I'm surprised this hasn't been caught yet, considering so many people must be using the `ActiveModelAdapter`.

packages/ember-data/lib/system/model/states.js
turboMaCk: I agree that we need add test case for rollbacking new uncommitted invalid record.
For me it make sense to handle it same as rollbacking uncommitted valid record.

pangratz: > For me it make sense to handle it same as rollbacking uncommitted valid record.

yep, that would be my vote too

turboMaCk:
@devinus You can try to build my fork yourself, and replace ember-data build in your app with this custom build untill this will be finaly fixed and merged to master.

```
$ git clone -b push-invalid-record git@github.com:turboMaCk/data.git
$ npm install
$ npm run build:production
```

The builds will be placed in the dist/ directory.

packages/ember-data/lib/system/model/states.js
turboMaCk: @pangratz I fix it this way but I still think that this should be done in some better way but my solution might be useful for anybody who needs to fix this right know. We should always refactor and change logic after final will be made.

pangratz: @turboMaCk not needed, I have already a fix available, just haven't made the PR yet.

devinus:
@turboMaCk Indeed, I'm just concerned about the issues @pangratz raised.

turboMaCk:
@devinus my last commit solve it. But you should merge this commit https://github.com/emberjs/data/pull/2943#issuecomment-103613223 which is more clean.

turboMaCk:
@devinus https://github.com/turboMaCk/data on master has both my and @pangratz fixes merged if you are interested....

jcope2013:
any timeline on this being merged in?

tchak:
@turboMaCk we merged @pangratz's PR. Can you leave only the `pushedData` part and the test in this PR? I can do it but if you can get to it before I do it would be nice.

bmac:
Hi @turboMaCk I've rebased and committed this pr as a5e4df6d5fe272664a2353cdb28a66110df2fd75. I kept you as the author. Thanks for pull request!

turboMaCk:
@bmac @tchak sorry for late response and thank you!

PR #2954, opened 2015-03-28
Fix: remove self reference

patricksimpson:
The reference to `self` is breaking FastBoot. 

The issue was outlined here: https://github.com/tildeio/ember-cli-fastboot/issues/11

### DIFF
--- a/packages/ember/lib/main.js
+++ b/packages/ember/lib/main.js
@@ -1,2 +1,2 @@
 // Shim Ember module
-export default self.Ember;
+export default Ember;


stefanpenner:
So ultimately this is just a limitation of how the two packages are bundled today, they should really just be using module references to each other. Also, the limitation is that the the resolveExternal hook for the module transpiler is either not available yet, or not obvious how to use.

i wonder if the node stuff, should just provide `self` as `global` or if this should do?

stefanpenner:
this also indicates we don't run ember-data's tests in node, which we should likely start doing.

patricksimpson:
@stefanpenner Thank you for the explanation. :-) 

I like the idea of providing `self` as a global, especially if that's something going to be used or expected in the future.

rwjblue:
This is good for now though, and unblocks usage of fastboot with Ember Data.  

@stefanpenner - Do you have objections to landing this for now?

PR #2956, opened 2015-03-29
Clear meta data when unloading all records for a type. Fixes #2772

### LINKED ISSUE #2772: store.unloadAll does not reset meta data
Using ember-cli, given type `conversation-preview`, after calling `this.get('store').unloadAll('conversationPreview')` a call to `this.get('store').metadataFor('conversation-preview');` returns the stale meta data.

runspired:


packages/ember-data/lib/system/store.js
```diff
@@ -1073,6 +1073,7 @@ Store = Service.extend({
     }
 
     typeMap.findAllCache = null;
+    this.typeMapFor(type).metadata = {};
```
wecc: Probably want to do `Ember.create(null)` instead of `{}` here

runspired: Changed.  Are there docs somewhere on Ember.create ? It's not in the website api docs. I'm assuming this is the underlying instantiation method (similar to Ember.set or Ember.get) that get's added to Ember.Object?

packages/ember-data/lib/system/store.js
```diff
@@ -1073,6 +1073,7 @@ Store = Service.extend({
     }
 
     typeMap.findAllCache = null;
+    this.typeMapFor(type).metadata = Ember.create(null);
```
wecc: You already have `typeMap` available here, another call to `typeMapFor()` should not be needed.

wecc:
Had a brief chat with @bmac about maybe clearing the typeMap for the specified type all together instead of nullifying separate properties (`findAllCache`, `metadata`). Thoughts @igorT?

runspired:
@wecc I personally think that makes sense.  I'm assuming if desired doing so would involve doing this after unloading the records?

```
var typeMaps = get(this, 'typeMaps');
var guid = Ember.guidFor(type);
if (typeMaps[guid]) {
    delete typeMaps[guid];
}
```

runspired:
@wecc also, any idea why the build can't complete? Can't get the tests to run locally, just hangs forever, seems same issue is happening with the CI tests.

wecc:
@runspired failing build should be fixed on master, can you rebase?

igorT:
Thanks for the PR! Can you squash/rebase and I'll or @wecc will merge

runspired:
@igorT @wecc squashed and rebased

PR #2961, opened 2015-03-31
Use new getter/setter for computed if available

wecc:
This fixes the ~3400 deprecation warnings `Using the same function as getter and setter is deprecated` for beta and canary builds.

### DIFF
--- a/packages/ember-data/lib/system/relationships/belongs-to.js
+++ b/packages/ember-data/lib/system/relationships/belongs-to.js
@@ -1,5 +1,7 @@
 import Model from 'ember-data/system/model';
 
+import computedPolyfill from "ember-data/utils/computed-polyfill";
+
 /**
   `DS.belongsTo` is used to define One-To-One and One-To-Many
   relationships on a [DS.Model](/api/data/classes/DS.Model.html).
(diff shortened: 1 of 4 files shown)


rwjblue:
:+1:

packages/ember-data/lib/system/relationships/belongs-to.js
```diff
@@ -86,9 +91,16 @@ function belongsTo(type, options) {
...
+      return this._relationships[key].getRecord();
     }
+  };
 
-    return this._relationships[key].getRecord();
+  return Ember.computed(supportsComputedGetterSetter ? computedConfig : function(key, value) {
+    if (arguments.length > 1) {
+      return computedConfig.set.call(this, key, value);
+    }
+    return computedConfig.get.call(this, key);
```
bmac: Could we move this logic into a shared util helper?

wecc: I considered using https://github.com/rwjblue/ember-new-computed/blob/master/addon/index.js from @rwjblue but didn't know if it was worth it since we only have settable CPs in these two places in ED. If we use a helper like the one linked we should probably make sure that we use it everywhere within ED.

wecc: That said, I'm happy to make it a shared helper too.

igorT: If we are already putting `supportsComputedGetterSetter` into a shared util, but it's only used as part of this function, we should probably move this into the util as well.

wecc:
1. Does the name `computedPolyfill` make sense?
2. Do we want all other CPs (getters only) to use this too for consistency?

igorT:
I think the getter only cps are gonna stay only functions in 2.0 as well, so I don't think there is a need. `computedPolyfill` does make sense

PR #2965, opened 2015-04-01
Make version check for Ember.js strict

tricknotes:
When `Ember.VERSION` contains some version suffix (such as '1.11.0.72d2409e'),
This check will reject this version of Ember unexpectedly.

### DIFF
--- a/packages/ember-data/lib/main.js
+++ b/packages/ember-data/lib/main.js
@@ -8,7 +8,7 @@
 Ember.RSVP.Promise.cast = Ember.RSVP.Promise.cast || Ember.RSVP.resolve;
 
 Ember.runInDebug(function() {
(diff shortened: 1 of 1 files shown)


rwjblue:
:+1:

wecc:
Thanks @tricknotes!

truenorth:
@rwjblue any chance we could get this brought into the ember-data beta channel on https://github.com/components/ember-data ? Right now, using the release/beta/canary streams of ember w/ ember-data doesn't work, due to the bug that this PR fixes

CC @stefanpenner

rwjblue:
It should already be in the ember-data canary builds.

igorT:
I think the beta.17 should be out pretty soon, so maybe not worth doing a beta.16.2

PR #2979, opened 2015-04-07
Add fullMessages property to DS.Errors

HeroicEric:
This adds a `fullMessages` property to a model's `errors` object, similar to the [#full_messages](http://apidock.com/rails/ActiveModel/Errors/full_messages) that Rails provides.

This is similar to the `errors.messages` that already exists but makes the error messages more useful by prefixing the error messages with the name of the attribute that it's for.

```javascript
user.get('errors').add('firstName', 'is required');
user.get('errors').add('lastName', 'is required');

user.get('errors.fullMessages');
// ["First name is required", "Last name is required"]
```

```handlebars
{{#each model.errors.fullMessages as |message|}}
  <p>{{message}}</p>
{{/each}}
```

### DIFF
--- a/packages/ember-data/lib/system/model/errors.js
+++ b/packages/ember-data/lib/system/model/errors.js
@@ -1,6 +1,8 @@
+var capitalize = Ember.String.capitalize;
 var get = Ember.get;
 var isEmpty = Ember.isEmpty;
 var map = Ember.EnumerableUtils.map;
+var underscore = Ember.String.underscore;
 
 import {
   MapWithDefault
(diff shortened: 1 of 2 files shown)


packages/ember-data/lib/system/model/errors.js
```diff
@@ -169,6 +181,41 @@ export default Ember.Object.extend(Ember.Enumerable, Ember.Evented, {
...
+    ```
+
+    @property fullMessages
+    @type {Array}
+  */
+  fullMessages: Ember.computed('content', function() {
+    var errors = get(this, 'content');
+
+    return map(errors, function(error) {
+      var attribute = underscore(error.attribute).replace('_', ' ');
```
HeroicEric: I'm not sure if there's a way to make this more flexible without having it use some function defined in your adapter.

johnnyshields:
FYI in Rails there's an I18n key which controls the ordering of attribute and error message:

```
  errors:
    format: "%{attribute} %{message}"
    messages:
      accepted: must be accepted
      blank: can't be blank
      present: must be blank
```

bmac:
cc @tchak

HeroicEric:
@tchak should I close this?

tchak:
@bmac I am against this, because it makes assumptions about attributes naming and leaves no place for i18n. So I am +1 to close.

igorT:
It seems like you could do this manually pretty easily if you wanted to. Closing

PR #2980, opened 2015-04-07
[DOC] Use block params style {{#each}}

HeroicEric:
Replaces  `{{#each ... in ...}}` syntax with block params style `{{#each ... as |...|}}`.


rwjblue:
:+1:

bmac:
Thanks @HeroicEric

PR #2983, opened 2015-04-08
Do not prepend a `/` if namespace is absolute.

rwjblue:
If you provide `namespace` as `/api/v1` it will end up using `//api/v1`.

As of this change, if `namespace` is absolute it will not prepend a slash.


fivetanley:
:+1:

Issue #2988, opened 2015-04-11
v1.0.0-beta.16.1 (and canary) - error after willDestroy 


svox1:
Ember-cli 0.2.3
Ember 1.12.0-beta.1

With v1.0.0-beta.15 everything works finde.

With v1.0.0-beta.16.1 I get an error after call
```javascript
this.store.willDestroy()
```

If I try to make a request:
```javascript
this.store.find("user", 1);
```

I get the error:

Uncaught TypeError: Cannot read property 'adapter:user' of undefinedstore.js:1901 ember$data$lib$system$store$$Service.extend.retrieveManagedInstancestore.js:1914 ember$data$lib$system$store$$Service.extend.lookupAdapterstore.js:1831 ember$data$lib$system$store$$Service.extend.adapterForstore.js:702 ember$data$lib$system$store$$Service.extend._flushPendingFetchForTypeember.debug.js:13590 Map.forEach.cbember.debug.js:13387 OrderedSet.forEachember.debug.js:13598 Map.forEachstore.js:696 ember$data$lib$system$store$$Service.extend.flushAllPendingFetchesember.debug.js:878 Queue.invokeember.debug.js:943 Queue.flushember.debug.js:748 DeferredActionQueues.flushember.debug.js:173 Backburner.endember.debug.js:576 (anonymous function)

igorT:
You are trying to find something after you destroyed the store? It seems ok that it errors out. Why are you trying to do this?

svox1:
I use `this.store.willDestroy()` to logged out current user. - to clean all models

After this I use the Login for a new user and get this error (I call this.store.find('user') after login).

With v1.0.0-beta.15 this works as expected.

igorT:
`this.store.willDestroy` is meant to be called once the store is being destroyed, it's not a public method. What you probably want is `store.unloadAll`
Can you try that?

svox1:
Thanks for the quick response.

If I change it to unloadAll I get this error:

Uncaught TypeError: Cannot read property 'typeKey' of undefinedstore.js:1462 
ember$data$lib$system$store$$Service.extend.model  Forstore.js:1062 
ember$data$lib$system$store$$Service.extend.unloadAll  auth-manager.js:185 
logout  application.js:180 
logout  ember.debug.js:23670 
triggerEvent  ember.debug.js:44818 
trigger  ember.debug.js:43542 
Router.trigger  ember.debug.js:23165 
[...]

Issue #2914, opened 2015-03-22
willDestroy missed in documentation?


svox1:
Hi,

I use store.willDestroy() to clean my models after the user is logged out. This works already fine.
I find this method after reading:
https://github.com/emberjs/data/blob/1.0.0-beta.15/packages/ember-data/lib/system/store.js#L1766

If this method is for public use, I was surprised that this method isnt in the documentation:
http://emberjs.com/api/data/classes/DS.Store.html

Or is this method not for public use?

Regards and thanks all Ember contributors!

bmac:
The willDestroy comes form Ember.Object which the store extends (http://emberjs.com/api/classes/Ember.Object.html#method_willDestroy). It looks like when the Ember Data API doc are built they aren't pulling in classes from Ember itself. I wonder if it would be possible for the API docs to know about Ember classes they depend on without those classes showing up in the Ember Data sidebar.

wecc:
@svox1 `store.willDestroy()` is not public API and not intended to be called manually. It's called by Ember internals when the store is being destroyed.

Could you explain more about your use-case? `store.unloadAll(type)` should be able to unload the models for you?

svox1:
>> I use store.willDestroy() to clean my models after the user is logged out.

I dont want to manually remove every model for himself. 
So that is why I need a foreach like in willDestroy:
```javascript
forEach(types, this.unloadAll, this);
```

I have create a pull request:
https://github.com/emberjs/data/pull/2995
with this change, the method is still working in ember-data v1.0.0-beta.16.1

Or are there better alternatives with public methods?

wecc:
Thanks for the PR! I'm afraid messing with `willDestroy` is not the right way to solve this problem, it's not supposed to be called at all.

Perhaps a better way would be to provide a public API for unloading all records of all types. We currently have methods for unloading a single record as well as all records for a specific type. I think a method to unload all records of all types would make more sense.

@bmac @fivetanley @igorT thoughts?

igorT:
@wecc I just proposed the exact same solution in the other thread :) 
`willDestroy` is a hook thats called when the store is being destroyed, it's not meant to be called directly.

svox1:
I have create a pull request for this. 
https://github.com/emberjs/data/pull/2999
I hope everything is well.

PR #2995, opened 2015-04-13
fix issue #2988 (v1.0.0-beta.16.1 (and canary) - error after willDestroy)

svox1:
After that, everything is clean and I can use willDestroy(); to teardown my App after the user is loggedout.


wecc:
Related discussion, see: #2914

PR #2997, opened 2015-04-13
jspm support

MajorBreakfast:
Currently it's possible to install ember via jspm, now this adds jspm support to ember-data.


igorT:
cc @rwjblue @fivetanley

fivetanley:
How does this work??

MajorBreakfast:
@fivetanley It tells SystemJS that ember-data depends on the ember module and defines the DS global as default export. I didn't specify ember as a dependency, this config expects an module called "ember" to be present but it can be a release, beta or canary build.

fivetanley:
seems great, thanks! :+1:

PR #2998, opened 2015-04-13
Improve property change notifications for records

### LINKED ISSUE #2900: No model update from PUT request result in beta.15
ember-data 1.0.0-beta.15 is not updating the model from the result of a PUT request anymore.

I downgraded ember-data to 1.0.0-beta.14 and the bug is not present in this version.

### LINKED ISSUE #2937: Model's property observer is triggered after save although value isn't changed by response
I have an observer on a model's property. 

When the model is saved, and only if the API returns the property (unchanged) as part of the payload, then the observer is being triggered.

Since the property hasn't changed - same type, same value - I'm left confused as to why my observer is being triggered after save.

wecc:
This PR improves how we trigger property changes when pushing data to the store/a record.

Previously we only cared about comparing new data with `model._data`, this resulted in weird behavior in a couple of scenarios:

1. If you modify an attribute locally and push new canonical data that matched the current local attribute we got a false trigger.
2. If you modify an attribute locally and save, and the server return new canonical data that matches the existing canonical data, we did not get a trigger (hence the old locally changed value stayed).

Fixes #2937, fixes #2900

### DIFF
--- a/packages/ember-data/lib/system/model/model.js
+++ b/packages/ember-data/lib/system/model/model.js
@@ -39,33 +39,6 @@ function extractPivotName(name) {
   );
 }
 
-// Like Ember.merge, but instead returns a list of keys
-// for values that fail a strict equality check
-// instead of the original object.
-function mergeAndReturnChangedKeys(original, updates) {
-  var changedKeys = [];
-
-  if (!updates || typeof updates !== 'object') {
-    return changedKeys;
-  }
-
-  var keys   = Ember.keys(updates);
-  var length = keys.length;
-  var i, val, key;
-
-  for (i = 0; i < length; i++) {
-    key = keys[i];
-    val = updates[key];
-
-    if (original[key] !== val) {
-      changedKeys.push(key);
-    }
-
-    original[key] = val;
-  }
-  return changedKeys;
-}
-
 /**
 
   The model class that all Ember Data records descend from.
@@ -844,6 +817,35 @@ var Model = Ember.Object.extend(Ember.Evented, {
   },
 
   /**
+    @method _changedKeys
+    @private
+  */
+  _changedKeys: function(updates) {
+    var changedKeys = [];
+
+    if (updates && typeof updates === 'object') {
+      var original, i, value, key;
+      var keys = Ember.keys(updates);
+      var length = keys.length;
+
+      original = merge({}, this._data);
+      original = merge(original, this._attributes);
+      original = merge(original, this._inFlightAttributes);
+
+      for (i = 0; i < length; i++) {
+        key = keys[i];
+        value = updates[key];
+
+        if (original[key] !== value) {
+          changedKeys.push(key);
+        }
+      }
+    }
+
+    return changedKeys;
+  },
+
+  /**
     @method _notifyProperties
     @private
   */
@@ -907,11 +909,12 @@ var Model = Ember.Object.extend(Ember.Evented, {
     @method adapterDidCommit
   */
   adapterDidCommit: function(data) {
-    var changedKeys;
     set(this, 'isError', false);
 
+    var changedKeys = this._changedKeys(data);
+
     if (data) {
-      changedKeys = mergeAndReturnChangedKeys(this._data, data);
+      merge(this._data, data);
     } else {
       merge(this._data, this._inFlightAttributes);
     }
@@ -956,7 +959,9 @@ var Model = Ember.Object.extend(Ember.Evented, {
   setupData: function(data) {
     Ember.assert("Expected an object as `data` in `setupData`", Ember.typeOf(data) === 'object');
 
-    var changedKeys = mergeAndReturnChangedKeys(this._data, data);
+    var changedKeys = this._changedKeys(data);
+
+    merge(this._data, data);
 
     this.pushedData();
 
--- a/packages/ember-data/tests/integration/records/property-changes-test.js
+++ b/packages/ember-data/tests/integration/records/property-changes-test.js
@@ -1,0 +1,115 @@
+var env, store, Person;
+var attr = DS.attr;
+var run = Ember.run;
+
+module('integration/records/property-changes - Property changes', {
+  setup: function() {
+    Person = DS.Model.extend({
+      firstName: attr('string'),
+      lastName: attr('string')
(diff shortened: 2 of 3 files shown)


packages/ember-data/lib/system/model/model.js
```diff
@@ -844,6 +817,35 @@ var Model = Ember.Object.extend(Ember.Evented, {
...
+  _changedKeys: function(updates) {
+    var changedKeys = [];
+
+    if (updates && typeof updates === 'object') {
+      var original, i, value, key;
+      var keys = Ember.keys(updates);
+      var length = keys.length;
+
+      original = merge({}, this._data);
+      original = merge(original, this._attributes);
```
igorT: not sure this is correct, as if you consider user applying changes after sending to the server, even if server sends a different value, it will not show up once you get it as an attr, as the local change will win

wecc: if we're in the middle of a save we have values in `_inFlightAttributes` and that will win over `_attributes`.

```
var person = store.push('person', { id: 1, name: 'Tomster' });
person.set('name', 'Yehuda');
person.save();
// => request { name: 'Yehuda' }
person.set('name', 'Tom');
// <= response { name: 'Yehuda' }
```
`_data` will contain `{ name: 'Tomster' }`
`_attributes` will contain `{ name: 'Tom' }`
`_inFlightAttributes` will contain `{ name: 'Yehuda' }`

We will compare payload (`{ name: 'Yehuda' }`) to the merged original (`{ name: 'Yehuda' }`) and trigger no change.

Isn't this the correct behavior? We have `{ name: 'Tom' }` locally (since we changed that) and we still do, without triggering observer. If we want to get back to `{ name: 'Yehuda' }` we should do a rollback.

wecc: However, if the server were to return something completely different, say `{ name: 'Igor' }`, we would get a trigger, but the value would remain `{ name: 'Tom' }`. That might be an issue though.

igorT: Yes, but we would get even if you didn't merge in attributes, no?

wecc: This is required for the scenario in test https://github.com/emberjs/data/pull/2998/files#diff-56dbbdcd2b2c38cfe44c52bec2b99646R55

packages/ember-data/lib/system/model/model.js
```diff
@@ -907,11 +909,12 @@ var Model = Ember.Object.extend(Ember.Evented, {
...
   adapterDidCommit: function(data) {
-    var changedKeys;
     set(this, 'isError', false);
 
+    var changedKeys = this._changedKeys(data);
+
     if (data) {
-      changedKeys = mergeAndReturnChangedKeys(this._data, data);
+      merge(this._data, data);
     } else {
```
igorT: We should always merge inflight while we are at it, it's related to the other bug we looked at

packages/ember-data/lib/system/model/model.js
igorT: Hm, then the order seems wrong, it should be inflight then attributes?

packages/ember-data/lib/system/model/model.js
wecc: Ah, yes. Do you remember the issue?

igorT: I think it's one of the two you linked to, sorry can't look in depth right now

packages/ember-data/tests/integration/records/property-changes-test.js
```diff
@@ -0,0 +1,115 @@
...
+  });
+
+  person.addObserver('lastName', function() {
+    ok(true, 'lastName observer should be triggered');
+  });
+
+  run(function() {
+    person.save();
+  });
+});
```
bmac: @wecc What should happen when store.push is called with a different value from a modified attribute?

Example:

```js
test('store.push should not override a modified attribute', function() {
  expect(1);
  var person;

  run(function() {
    person = store.push('person', {
      id: 'wat',
      firstName: 'Yehuda',
      lastName: 'Katz'
    });

    person.set('lastName', 'Katz!');
  });

  person.addObserver('firstName', function() {
    ok(true, 'firstName observer should be triggered');
  });

  person.addObserver('lastName', function() {
    ok(false, 'lastName observer should not be triggered');
  });

  run(function() {
    person = store.push('person', {
      id: 'wat',
      firstName: 'Tom',
      lastName: 'Dale'
    });
  });
});
```

wecc: That's a good question... thinking about it a bit more I _think_ this is how we want it:

Pushing should only update canonical values (`_data`). If an attribute is present in `_attributes` or `_inFlightAttributes` we should not trigger, no matter what. If an attribute is present only in `_data` and the value differs from the new value, then we should trigger.

Saving should work more or less the same way but only take `_data` and `_inFlightAttributes` into consideration. If the attribute is present in `_attributes` we should not trigger.

Does this make sense?

bmac: Makes sense to me.

igorT: Yup thats the current intended behavior, we might have tests for it even, but not sure

packages/ember-data/lib/system/model/model.js
```diff
@@ -844,6 +817,35 @@ var Model = Ember.Object.extend(Ember.Evented, {
...
+  */
+  _changedKeys: function(updates) {
+    var changedKeys = [];
+
+    if (updates && typeof updates === 'object') {
+      var original, i, value, key;
+      var keys = Ember.keys(updates);
+      var length = keys.length;
+
+      original = merge({}, this._data);
```
fivetanley: Ember.create(null) instead of `{}`

bmac:
@wecc what needs to happen for this to get merged?

wecc:
@bmac needs to address the comments in here to make sure it's working as intended.

igorT:
@bmac didn't you move this into a new pr?

wecc:
@igorT it's in #3216

PR #2999, opened 2015-04-13
Make unloadAll() unload all records, deprecate unloadAll(type) in favor ...

svox1:
Make unloadAll() unload all records, deprecate unloadAll(type) in favor of unloadType(type)

Big thx to @wecc for the support!

### DIFF
--- a/packages/ember-data/lib/system/store.js
+++ b/packages/ember-data/lib/system/store.js
@@ -1054,13 +1054,13 @@ Store = Service.extend({
     This method unloads all of the known records for a given type.
 
     ```javascript
-    store.unloadAll('post');
+    store.unloadType('post');
     ```
 
-    @method unloadAll
+    @method unloadType
     @param {String or subclass of DS.Model} type
   */
-  unloadAll: function(type) {
+  unloadType: function(type) {
     var modelType = this.modelFor(type);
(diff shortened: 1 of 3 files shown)


fivetanley:
LGTM but we'll plus one / merge it at friday's meeting

svox1:
Sounds good, but was there any issues on this?

fivetanley:
Hey this looked great to us. We weren't solid on `unloadType`. Igor and I thought this could be construed as unloading an entire type (as in you can't make new records with it) from the store.

@wecc suggested  `unloadRecordsByType(type)`, `unloadAllRecords()` as new method names, I think those are good. If you have any alternative names we're totally open to it! If not, if you can update the PR with the renames I will merge it ASAP. Thanks!

fivetanley:
After talking at the meeting, we'd like to switch on type. So for example `unloadAll()` and `unloadAll(type)`. We can always switch these to different methods later in a backwards compatible way.

svox1:
I hope I have understood you correctly. 

I have change the method so we have:
unloadAll(); and
unloadAll(type);

fivetanley:
:+1: I'll merge it once Travis and Appveyor give us the green light. Thanks!

quaertym:
:+1: Great.

PR #3003, opened 2015-04-15
make store.serialize a public method

fivetanley:
Sometimes, it is useful to have hooks for serializing records for an endpoint that has a custom endpoint that does not match a simple CRUD operation. An example of this is creating bulk records, where you may want to map over a list of records:

```javascript

var records = this.store.all('post');

var array = records.map( (record) => this.store.serialize(record) );

$.ajax({type: 'POST', data: array, url: '/bulk_update_posts'});
```

This seems to me like a necessary escape hatch for people who can't adapt to JSONAPI in the future for some endpoints as well.

### DIFF
--- a/packages/ember-data/lib/system/store.js
+++ b/packages/ember-data/lib/system/store.js
@@ -241,7 +241,6 @@ Store = Service.extend({
       the JSON representation
 
     @method serialize
-    @private
     @param {DS.Model} record the record to serialize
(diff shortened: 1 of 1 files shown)


igorT:
isn't `records.invoke('serialize')` faster/easier than ^^?

fivetanley:
c

PR #3004, opened 2015-04-17
Fix serialization results in the documentation.

nathanhammond:
I believe that this is a mistake in the description, but I'm unfamiliar enough with Ember Data to be unsure. :)

### DIFF
--- a/packages/ember-data/lib/serializers/json-serializer.js
+++ b/packages/ember-data/lib/serializers/json-serializer.js
@@ -87,6 +87,8 @@ export default Serializer.extend({
 
     ```javascript
     {
+      "firstName": "Harry",
+      "lastName": "Houdini",
       "career": "magician"
     }
     ```


bmac:
I believe `career` is correct in this case because the attribute is remapped by the {key: 'career'} option: https://github.com/emberjs/data/pull/3004/files#diff-cce3ad76c2e3801ed1a94bf5276dbbbbR81.

Here is a simple jsbin that demonstrates the example in the docs:
http://emberjs.jsbin.com/lisuyahoma/1/edit?html,css,js,console

It does appear that the `firstName` and `lastName` are missing from the serialized example. If you would like to change this pr to add those I would be happy to merge it.

nathanhammond:
I read it as a `deserialize` instead of a `serialize`. :) Patch modified to include additional details.

bmac:
Thanks.

PR #3008, opened 2015-04-17
url-templates inital implementation

amiel:
This is an initial implementation for https://github.com/emberjs/rfcs/pull/4

Currently
---------

* This is fully backwards compatible with `buildURL`.
* The template is compiled with a very incomplete inline implementation
  that does not handle lists or reserved words, and only supports the
  `/`, `?`, and `+` operators.
* A context for filling in the templates is created using `urlSegments`,
  which is defined on the adapter.

Example Usage
-------------

```javascript
// adapters/invoice
extend default Adapter.extend({
  urlTemplate: '/api/users/{userId}/invoices{/id}',

  urlSegments: {
    userId: function(type, id, snapshot, requestType) {
      return snapshot.belongsTo('user', { id: true });
    }
  }
});
```

TODO
----

* [x] Create an ember-cli addon with this functionality to get it out there and
  get feedback.
* [ ] Either use a 3rd-party url-template library, or support a larger subset
  of the rfc6570 spec. Possibly create an adapter layer so that a
  lightweight version can be included in ember-data and users can include
  a more full-featured library if they want.
* [ ] Allow for defining a template for specific `requestType`s.
* [ ] Support `findBelongsTo` and `findHasMany` without using `urlPrefix`.
  These methods use `buildURL` and then `urlPrefix` to generate a prefix
  suitable for prepending to the value from `links`. This seems
  unnecessary with url-templates, but needs to be backwards compatible.

### DIFF
--- a/packages/ember-data/lib/adapters/build-url-mixin.js
+++ b/packages/ember-data/lib/adapters/build-url-mixin.js
@@ -1,4 +1,6 @@
 var get = Ember.get;
+var isArray = Ember.isArray;
+var sanitize = encodeURIComponent;
 
 /**
 
@@ -28,6 +30,10 @@ var get = Ember.get;
   @namespace DS
 */
 export default Ember.Mixin.create({
+  // Default template should mimick original behaviour
+  urlTemplate: '{+host}{/namespace}{/pathForType}{/id}',
+  mergedProperties: ['urlSegments'],
+
   /**
     Builds a URL for a given type and optional ID.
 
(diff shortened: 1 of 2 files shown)


amiel:
I've created an addon that I'm currently using (pretty limited use though): https://github.com/amiel/ember-data-url-templates

It depends on ember-data canary (see notes on which specific commits in the [README](https://github.com/amiel/ember-data-url-templates#requirements)).

bmac:
Thanks for publishing the addon @amiel.

I'm going to close this pr because that seems to be working as a good solution for now. Feel free to re-open if you disagree.

PR #3021, opened 2015-04-22
Remove unnecessary URL check in findHasMany

wecc:
This check is already done in [urlPrefix()](https://github.com/emberjs/data/blob/v1.0.0-beta.16.1/packages/ember-data/lib/adapters/build-url-mixin.js#L90-L95), no need to have it duplicated.


bmac:
:+1:

PR #3024, opened 2015-04-24
[refactor] Prune code that does nothing

koriroys:
Original intent of the code is in commit 42d966e43f9ea13d94e61f62fa72b7b8ffe33178.
It was changed in fbabbda32cccf43d79782bbcaa3192fbeecaeb18.

If there is no adapterPayload, why set payload = adapterPayload? Just let it stay empty.


fivetanley:
thanks!

PR #3030, opened 2015-04-27
ActiveModelAdapter camelizes keys for errors on 422

bdvholmes:
Fields are not currently camelized in the model's errors object e.g.

    pilot.errors.first_name

should be

    pilot.errors.firstName

There is already test for this behavior but because of the way `InvalidError` and `Ember.inspect` are implemented the test passes when it shouldn't do.

ActiveModelAdapter now passes the errors hash directly to `InvalidError` as is expected according to the docs. This assumes that the 422 provides JSON of the form

    {
      errors: {
        first_name: ['is too long']
      }
    }

as is standard in active model. If `errors` is not present in the JSON response, active model adapter will continue to behave as it did before.

### DIFF
--- a/packages/activemodel-adapter/lib/system/active-model-adapter.js
+++ b/packages/activemodel-adapter/lib/system/active-model-adapter.js
@@ -141,7 +141,8 @@ var ActiveModelAdapter = RESTAdapter.extend({
     var error = this._super.apply(this, arguments);
 
     if (jqXHR && jqXHR.status === 422) {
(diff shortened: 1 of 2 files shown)


fivetanley:
awesome, thanks!

pangratz:
@fivetanley @bdvholmes shouldn't camelCasing they keys be a concern of the serializer, and be done inside `extractErrors` (see http://git.io/vJ0Ey)?

igorT:
@pangratz is correct, this PR seems wrong

PR #3031, opened 2015-04-28
Added pushedData hook to root.deleted.uncommitted state.

### LINKED ISSUE #3023: Attempted to handle event `pushedData` on model while in state root.deleted.uncommitted
If you have a request to the server in-flight and you delete the model then you get an exception `Attempted to handle event `pushedData` on model while in state root.deleted.uncommitted`. I suggest that `pushedData` should be supported in this state - simply add the new data and leave the model in the `root.deleted.uncommitted` state.

aexmachina:
Resolves #3023.

### DIFF
--- a/packages/ember-data/lib/system/model/states.js
+++ b/packages/ember-data/lib/system/model/states.js
@@ -633,6 +633,7 @@ var RootState = {
         record.triggerLater('ready');
       },
 
+      pushedData: Ember.K,
       becomeDirty: Ember.K,
       deleteRecord: Ember.K,
 
--- a/packages/ember-data/tests/unit/model-test.js
+++ b/packages/ember-data/tests/unit/model-test.js
@@ -260,6 +260,19 @@ test("Calling attr(), belongsTo() or hasMany() throws a warning", function() {
   });
 });
 
+test("supports pushedData in root.deleted.uncommitted", function() {
+  var record;
+  var hash = { id: 1 };
+  run(function() {
+    record = store.push(Person, hash);
+    record.deleteRecord();
+    store.push(Person, hash);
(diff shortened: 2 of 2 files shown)


packages/ember-data/tests/unit/states-test.js
```diff
@@ -87,3 +87,14 @@ test("the deleted.saved state", function() {
...
+module("unit/states - hooks for record states", {
+  setup: function() {
+    rootState = DS.RootState;
+  }
+});
+
+test("root.deleted.uncommitted supports pushedData", function() {
+  equal(typeof get(rootState, "deleted.uncommitted.pushedData"), 'function',
+    "root.deleted.uncommitted should be a function");
+});
```
bmac: Instead of testing that the `deleted.uncommitted.pushedData` function exists could you add a new test to the `model-test.js` file that creates a record with an id using `store.push`, calls `record.deleteRecord()` to move it into the `deleted.uncommitted` state then call `store.push` again with the same id then assert that the record is still in the `root.deleted.uncommitted` state and no errors are thrown?

This should help make the test more resilient against future refactors.

igorT:
No idea why Appveyor failed, I think it's due to a npm issue, will try and restart

packages/ember-data/tests/unit/states-test.js
aexmachina: Yep good call, will do. I didn't like this approach but wasn't sure how to correctly test it.

danbell:
@igorT Any movement on this PR?

aexmachina:
I've improved the test based on the feedback from @bmac.

bmac:
Looks good to me.

PR #3032, opened 2015-04-29
use typeKey and typeClass instead of using `type` inconsistently

fivetanley:
Before, we exchanged type and typeKey intermittently. Sometimes it
was one or the other. This changes most usages of `type` to
either `typeKey` or `typeClass` for internal consistency.

The API docs were also changed to reflect this.

The Store API docs have been changed to suggest using a string
for a future patchset where we don't take factories to the store
finder methods.


igorT:
:heart: :heart: :heart: I went through most of it, look good

PR #3033, opened 2015-04-29
dasherize ALL the things: use dasherized model names everywhere

fivetanley:
use dasherized model names everywhere / change typeKey -> modelName

Previously, we allowed container keys to be loosy-goosy and normalized
them in several places. This commit consolidates those efforts into
one place `normalizeTypeKey`.

typeKey is now referred to as modelName on snapshot instances and Model
classes. typeKey remains on both of these but is deprecated.

modelName is now a dasherized string, where previously it was
camelCased.

### DIFF
--- a/packages/activemodel-adapter/lib/system/active-model-serializer.js
+++ b/packages/activemodel-adapter/lib/system/active-model-serializer.js
@@ -1,5 +1,6 @@
 import { singularize } from "ember-inflector";
 import RESTSerializer from "ember-data/serializers/rest-serializer";
+import normalizeModelName from "ember-data/system/normalize-model-name";
 /**
   @module ember-data
 */
@@ -146,7 +147,7 @@ var ActiveModelSerializer = RESTSerializer.extend({
     @param {Object} options
   */
   serializeIntoHash: function(data, typeClass, snapshot, options) {
-    var root = underscore(decamelize(typeClass.typeKey));
+    var root = underscore(decamelize(typeClass.modelName));
     data[root] = this.serialize(snapshot, options);
   },
 
@@ -166,7 +167,7 @@ var ActiveModelSerializer = RESTSerializer.extend({
     if (Ember.isNone(belongsTo)) {
       json[jsonKey] = null;
     } else {
-      json[jsonKey] = classify(belongsTo.typeKey).replace(/(\/)([a-z])/g, function(match, separator, chr) {
+      json[jsonKey] = classify(belongsTo.modelName).replace(/(\/)([a-z])/g, function(match, separator, chr) {
         return match.toUpperCase();
       }).replace('/', '::');
     }
@@ -290,9 +291,10 @@ var ActiveModelSerializer = RESTSerializer.extend({
     }
   },
   typeForRoot: function(key) {
-    return camelize(singularize(key)).replace(/(^|\:)([A-Z])/g, function(match, separator, chr) {
+    var convertedFromRubyModule = camelize(singularize(key)).replace(/(^|\:)([A-Z])/g, function(match, separator, chr) {
       return match.toLowerCase();
     }).replace('::', '/');
+    return normalizeModelName(convertedFromRubyModule);
   }
 });
 
--- a/packages/ember-data/lib/serializers/rest-serializer.js
+++ b/packages/ember-data/lib/serializers/rest-serializer.js
@@ -3,12 +3,12 @@
 */
 
 import JSONSerializer from "ember-data/serializers/json-serializer";
+import normalizeModelName from "ember-data/system/normalize-model-name";
+import {singularize} from "ember-inflector/lib/system/string";
 
 var forEach = Ember.ArrayPolyfills.forEach;
 var map = Ember.ArrayPolyfills.map;
 var camelize = Ember.String.camelize;
-
-import { singularize } from "ember-inflector/lib/system/string";
 
 function coerceId(id) {
   return id == null ? null : id + '';
@@ -263,7 +263,7 @@ var RESTSerializer = JSONSerializer.extend({
   */
   extractSingle: function(store, primaryTypeClass, rawPayload, recordId) {
     var payload = this.normalizePayload(rawPayload);
-    var primaryTypeClassName = primaryTypeClass.typeKey;
+    var primaryTypeClassName = primaryTypeClass.modelName;
     var primaryRecord;
 
     for (var prop in payload) {
@@ -274,7 +274,7 @@ var RESTSerializer = JSONSerializer.extend({
         continue;
       }
       var type = store.modelFor(typeName);
-      var isPrimary = type.typeKey === primaryTypeClassName;
+      var isPrimary = type.modelName === primaryTypeClassName;
       var value = payload[prop];
 
       if (value === null) {
@@ -417,26 +417,26 @@ var RESTSerializer = JSONSerializer.extend({
   */
   extractArray: function(store, primaryTypeClass, rawPayload) {
     var payload = this.normalizePayload(rawPayload);
-    var primaryTypeClassName = primaryTypeClass.typeKey;
+    var primaryTypeClassName = primaryTypeClass.modelName;
     var primaryArray;
 
     for (var prop in payload) {
-      var typeKey = prop;
+      var modelName = prop;
       var forcedSecondary = false;
 
       if (prop.charAt(0) === '_') {
         forcedSecondary = true;
-        typeKey = prop.substr(1);
+        modelName = prop.substr(1);
       }
 
-      var typeName = this.typeForRoot(typeKey);
+      var typeName = this.typeForRoot(modelName);
       if (!store.modelFactoryFor(typeName)) {
         Ember.warn(this.warnMessageNoModelForKey(prop, typeName), false);
         continue;
       }
       var type = store.modelFor(typeName);
       var typeSerializer = store.serializerFor(type);
-      var isPrimary = (!forcedSecondary && (type.typeKey === primaryTypeClassName));
+      var isPrimary = (!forcedSecondary && (type.modelName === primaryTypeClassName));
 
       /*jshint loopfunc:true*/
       var normalizedArray = map.call(payload[prop], function(hash) {
@@ -488,12 +488,12 @@ var RESTSerializer = JSONSerializer.extend({
     var payload = this.normalizePayload(rawPayload);
 
     for (var prop in payload) {
-      var typeKey = this.typeForRoot(prop);
-      if (!store.modelFactoryFor(typeKey, prop)) {
-        Ember.warn(this.warnMessageNoModelForKey(prop, typeKey), false);
+      var modelName = this.typeForRoot(prop);
+      if (!store.modelFactoryFor(modelName, prop)) {
+        Ember.warn(this.warnMessageNoModelForKey(prop, modelName), false);
         continue;
       }
-      var type = store.modelFor(typeKey);
+      var type = store.modelFor(modelName);
       var typeSerializer = store.serializerFor(type);
 
       /*jshint loopfunc:true*/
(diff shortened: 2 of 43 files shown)


igorT:
Tests fail?

fivetanley:
> Tests fail?

Yeah, I rebased without re running, fixed the tests that broke. They were expecting a camelCased type when extracting, but it should be dasherized.

packages/ember-data/lib/system/store.js
```diff
@@ -1902,12 +1902,13 @@ Store = Service.extend({
...
-    @param {String} type the object type
+    @param {String} type the object typeKey
     @param {String} type the object name
     @return {Ember.Object}
   */
-  retrieveManagedInstance: function(type, name) {
-    var key = type+":"+name;
+  retrieveManagedInstance: function(typeKey, name) {
+    var normalizedTypeKey = this._normalizeTypeKey(typeKey);
+    var key = normalizedTypeKey +":"+name;
```
pangratz: tiny nitpick: spaces around `+`

igorT:
Can we add tests that assert this is backwards compatible?

igorT:
And by we, I meant the awesome @fivetanley :D

packages/ember-data/lib/system/normalize-type-key.js
```diff
@@ -0,0 +1,16 @@
+import {singularize} from 'ember-inflector/lib/system/string';
+
+var dasherize = Ember.String.dasherize;
+var camelize  = Ember.String.camelize;
+
+/**
+  All typeKeys are camelCase internally. Changing this function may
```
wecc: camelCase => dasherized?

packages/ember-data/lib/system/normalize-type-key.js
```diff
@@ -0,0 +1,16 @@
...
+/**
+  All typeKeys are dasherized internally. Changing this function may
+  require changes to other normalization hooks (such as typeForRoot).
+  @method normalizeTypeKey
+  @private
+  @param {String} type
+  @return {String} if the adapter can generate one, an ID
+*/
+export default function normalizeTypeKey(typeKey) {
+  return singularize(dasherize(camelize(typeKey)));
```
igorT: Why do we have to singularize here?

packages/ember-data/lib/serializers/rest-serializer.js
```diff
@@ -549,7 +548,7 @@ var RESTSerializer = JSONSerializer.extend({
     @return {String} the model's typeKey
   */
   typeForRoot: function(key) {
-    return camelize(singularize(key));
+    return normalizeTypeKey(key);
```
igorT: Seems like we don't expose normalizeTypeKey as a public api/on the DS namespace, yet we use it in the REST serializer? I am kinda dubious of the same function being used both internally and externally.

packages/ember-data/lib/system/relationship-meta.js
```diff
@@ -6,7 +6,7 @@ export function typeForRelationshipMeta(store, meta) {
   typeKey = meta.type || meta.key;
   if (typeof typeKey === 'string') {
     if (meta.kind === 'hasMany') {
-      typeKey = singularize(typeKey);
+      typeKey = normalizeTypeKey(typeKey);
```
igorT: don't we already do this in the relationship CPs themselves?

packages/ember-data/lib/system/store.js
```diff
@@ -1952,7 +1953,7 @@ Store = Service.extend({
     @return {String} if the adapter can generate one, an ID
   */
   _normalizeTypeKey: function(key) {
-    return camelize(singularize(key));
+    return normalizeTypeKey(key);
```
igorT: Should this not be a store method anymore?

packages/ember-data/tests/integration/adapter/build-url-mixin-test.js
```diff
@@ -147,7 +147,7 @@ test('buildURL - with camelized names', function() {
   adapter.setProperties({
     pathForType: function(type) {
       var decamelized = Ember.String.decamelize(type);
-      return Ember.String.pluralize(decamelized);
+      return Ember.String.underscore(Ember.String.pluralize(decamelized));
```
igorT: Is this a breaking change?

igorT:
I am pretty scared of dasherizing typeKey, gonna break people's stuff a bunch. Maybe we should go with the modelName strategy?

packages/ember-data/lib/system/normalize-type-key.js
fivetanley: we don't have to, but it's nice

fivetanley:
>  I am pretty scared of dasherizing typeKey, gonna break people's stuff a bunch.

It'd be great to know what people are using typeKey for. Most of the changes in your app code would be comparing against dasherized strings. If we're going to remove this, I'd rather remove it this week than right before 1.0.

fivetanley:
> Maybe we should go with the modelName strategy?

Yeah, just takes more tedious work, but I can switch this over to that.

typeoneerror:
> It'd be great to know what people are using typeKey for.

I'm using typeKey in a handful of cases, specifically for some very hacky polymorhphic relationship handling. One example is the "attachable" polymorphic relationship. I have an action that updates the "attached" model when an image is removed from it. I use typeKey here (I'm sure there's a much better way to do this, but I just did a search in my project and that's where I'm using it.

    imageRemoved: function(response, model) {
      let type = model.constructor.typeKey;
      if (response) {
        if (response.hasOwnProperty(type) && !Ember.isEmpty(response[type])) {
          let data = this.store.normalize(type, response[type]);
          this.store.push(type, data);
        }
      }
    }

packages/activemodel-adapter/lib/system/active-model-serializer.js
```diff
@@ -290,9 +290,10 @@ var ActiveModelSerializer = RESTSerializer.extend({
     }
   },
   typeForRoot: function(key) {
-    return camelize(singularize(key)).replace(/(^|\:)([A-Z])/g, function(match, separator, chr) {
+    var normalized = camelize(singularize(key)).replace(/(^|\:)([A-Z])/g, function(match, separator, chr) {
       return match.toLowerCase();
     }).replace('::', '/');
+    return this._super(normalized);
```
mixonic: If `_super` is not `RESTAdapter` the return would not be present. It is, seems ok, but is not safe if the parent class is not as expected.

packages/ember-data/lib/system/relationship-meta.js
```diff
@@ -6,7 +6,7 @@ export function typeForRelationshipMeta(store, meta) {
   typeKey = meta.type || meta.key;
```
mixonic: I might say `typeKey` here is inappropriate and misleading. This is `userEnteredSomeKey`

packages/ember-data/lib/system/relationships/belongs-to.js
```diff
@@ -56,24 +57,32 @@ import computedPolyfill from "ember-data/utils/computed-polyfill";
...
-  @param {String} type (optional) type of the relationship
+  @param {String} modelTypeKey (optional) type of the relationship
   @param {Object} options (optional) a hash of options
   @return {Ember.computed} relationship
 */
-function belongsTo(type, options) {
-  if (typeof type === 'object') {
-    options = type;
-    type = undefined;
+function belongsTo(modelTypeKey, options) {
```
mixonic: Again, not really a formal `typeKey`.

packages/ember-data/lib/system/store.js
mixonic: Indeed seems good to remove

packages/ember-data/tests/integration/backwards-compat/non-dasherized-lookups.js
```diff
@@ -0,0 +1,121 @@
...
+    });
+  });
+
+  run(function() {
+    store.find('long_model_name', 1).then(function(longModelName) {
+      deepEqual(longModelName.get('postNotes').toArray(), [store.getById('postNote', 1)]);
+    });
+  });
+
+});
```
mixonic: I worry these tests are only that the API is consistent with the way to user passes type to the store API. Can we add some that mix it up?

packages/ember-data/tests/unit/store/model-for-test.js
```diff
@@ -26,24 +26,25 @@ module("unit/store/model_for - DS.Store#modelFor", {
...
 test("when fetching factory from string and dashing normalizer, sets a normalized key as typeKey", function() {
-  env.replaceContainerNormalize(function(fullName) {
-    return dasherize(camelize(fullName));
+  env.replaceContainerNormalize(function(key) {
+    return dasherize(camelize(key));
   });
-
   equal(registry.normalize('some.post'), 'some-post', 'precond - container dasherizes');
-  equal(store.modelFor("blog.post").typeKey, "blogPost", "typeKey is normalized to camelCase");
+  equal(store.modelFor("blog.post").typeKey, "blog-post", "typeKey is normalized to camelCase");
```
mixonic: assertion comment is not correct here.

packages/ember-data/tests/unit/store/model-for-test.js
```diff
@@ -26,24 +26,25 @@ module("unit/store/model_for - DS.Store#modelFor", {
...
 });
 
 test("when returning passed factory, sets a normalized key as typeKey", function() {
   var factory = { typeKey: 'some-thing' };
-  equal(store.modelFor(factory).typeKey, "someThing", "typeKey is normalized to camelCase");
+  equal(store.modelFor(factory).typeKey, "some-thing", "typeKey is normalized to camelCase");
```
mixonic: assertion comment is not correct

packages/activemodel-adapter/lib/system/active-model-serializer.js
fivetanley: :+1:

packages/ember-data/lib/system/relationship-meta.js
fivetanley: This gets used by some of the relationship macros.

packages/ember-data/lib/system/relationship-meta.js
fivetanley: :+1:

packages/ember-data/tests/integration/adapter/build-url-mixin-test.js
fivetanley: No, the default adapter has no normalization, but the tests expected underscored strings.

packages/ember-data/lib/system/snapshot.js
```diff
@@ -25,7 +25,7 @@ function Snapshot(record) {
   this.id = get(record, 'id');
   this.record = record;
   this.type = record.constructor;
-  this.typeKey = record.constructor.typeKey;
+  this.modelName = record.constructor.modelName;
```
fivetanley: needs deprecation

packages/ember-data/lib/system/store.js
```diff
@@ -1466,21 +1466,39 @@ Store = Service.extend({
...
+      if (factory.modelName) {
+        factory.modelName = normalizeTypeKey(factory.modelName);
       }
     }
 
+    Ember.runInDebug(function() {
+      // deprecate typeKey
+      if (!('typeKey' in factory)) {
+        Ember.defineProperty(factory, 'typeKey', {
+          enumerable: false,
```
igorT: Isn't typeKey supposed to be enumerable?

fivetanley: typeKey is on the class. Do people do for in on the model class?

raytiley:
So this breaks everyone's apps that where `model-name` results in `/modelname` for the url.

igorT:
@fivetanley I thought we kept the existing behavior for RestAdapter? None of the tests were changed

raytiley:
@igorT looks like my app is customizing `pathForType` fixing it on my end.

igorT:
@raytiley kinda surprised it broke, would you mind sharing what the problem/fix was?

raytiley:
@igorT 

I had this code which broke whith `type='myModel'` would now be `my-models` instead of `mymodels`.
```js
RESTAdapter.reopen({
  pathForType: function(type) {
    return Ember.String.pluralize(type.toLowerCase());
  }
});
```

I changed it to 

```js
RESTAdapter.reopen({
  pathForType: function(type) {
    var path = this._super(type);
    return path.toLowerCase();
  }
});
```

PR #3034, opened 2015-04-29
introduce modelNameFromPayloadKey and deprecate typeForRoot

fivetanley:
Introduced in this commit is payloadKeyFromModelName, which allows you
to specify the outgoing root key for a model name in the JSON payload.

For example:

```javascript
RESTSerializer.extend({modelNameFromPayloadKey: underscore});
```

```json
{
  "taco-party": { }
}
```

becomes

```json
{
  "taco_party": {}
}
```

For symmetry, modelNameFromPayloadKey has been introduced to
look up the ember data record in the store given a JSON root
key. This used to be called pathForType, which has been deprecated.

Part of the refactor stuff covered in https://github.com/emberjs/data/pull/2821#issuecomment-79183508

### DIFF
--- a/TRANSITION.md
+++ b/TRANSITION.md
@@ -712,13 +712,13 @@ response to be camelized.
 ```
 
 If your server uses underscored root objects you can define the
-`typeForRoot` method in your `ApplicationSerializer`.
+`modelNameFromPayloadKey` method in your `ApplicationSerializer`.
 
 ```js
 App.ApplicationSerializer = DS.RESTSerializer.extend({
-  typeForRoot: function(root) {
-    var camelized = Ember.String.camelize(root);
-    return Ember.String.singularize(camelized);
+  modelNameFromPayloadKey: function(root) {
+    var dasherize = Ember.String.dasherize(root);
+    return Ember.String.singularize(dasherize);
   }
 });
 ```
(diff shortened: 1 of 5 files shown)


bmac:
Looks good to me. @igorT thoughts?

TRANSITION.md
```diff
@@ -712,11 +712,11 @@ response to be camelized.
...
 
 If your server uses underscored root objects you can define the
-`typeForRoot` method in your `ApplicationSerializer`.
+`modelNameFromPayloadKey` method in your `ApplicationSerializer`.
 
 ```js
-  typeForRoot: function(root) {
+  modelNameFromPayloadKey: function(root) {
     var camelized = Ember.String.camelize(root);
```
mixonic: I'm unsure of the context in this transition doc, but this suggests returning modelName as non-dasherized. Is incorrect.

packages/ember-data/lib/serializers/rest-serializer.js
```diff
@@ -529,12 +529,12 @@ var RESTSerializer = JSONSerializer.extend({
 
     In order for Ember Data to know that the model corresponding to
     the 'response-fast-car' hash is `FastCar` (modelName: 'fastCar'),
-    you can override typeForRoot to convert 'response-fast-car' to
+    you can override modelNameFromPayloadKey to convert 'response-fast-car' to
     'fastCar' like so:
```
mixonic: This line suggest returning `fastCar`, which is incorrect. Should only suggest returning dasherized names.

PR #3035, opened 2015-04-30
Add method argument to key serialization docs

Dremora:
In `JSONSerializer`, both `keyForAttribute` and `keyForRelationship` get an extra undocumented parameter, `method` (which can be either `serialize` or `deserialize`). This PR adds the missing bits.
Also there was a call to `keyForAttribute` with a `method` argument missing — 'serialize` is now explicitly being passed there.

### DIFF
--- a/packages/ember-data/lib/serializers/rest-serializer.js
+++ b/packages/ember-data/lib/serializers/rest-serializer.js
@@ -40,7 +40,7 @@ function coerceId(id) {
 
   ```js
   App.ApplicationSerializer = DS.RESTSerializer.extend({
-    keyForAttribute: function(attr) {
+    keyForAttribute: function(attr, method) {
       return Ember.String.underscore(attr).toUpperCase();
     }
   });
--- a/packages/ember-data/lib/serializers/json-serializer.js
+++ b/packages/ember-data/lib/serializers/json-serializer.js
@@ -554,7 +554,7 @@ export default Serializer.extend({
       var payloadKey =  this._getMappedKey(key);
 
       if (payloadKey === key && this.keyForAttribute) {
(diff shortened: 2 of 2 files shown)


packages/ember-data/lib/serializers/rest-serializer.js
```diff
@@ -40,8 +40,12 @@ function coerceId(id) {
...
   ```js
-    keyForAttribute: function(attr) {
-      return Ember.String.underscore(attr).toUpperCase();
+    keyForAttribute: function(attr, method) {
+      if (method === 'serialize') {
+        return Ember.String.underscore(attr).toUpperCase();
+      } else {
+        return Ember.String.camelize(attr.toLowerCase());
+      }
```
bmac: Lets leave this example as it was (although feel free to add the `method` parameter to the function declaration). The new implementation doesn't match the text above it.

Dremora: The original example would work fine if the data is never saved back to the server. If it is, though, we do need to perform the reverse conversion. Perhaps I should instead expand the text above and and a bit clarifying this reverse procedure?

bmac: The `keyForAttribute` method's `attr` parameter is always the name of the attribute as defined on the Model. So if the API you are working with consistently uses the same keys when serializing and deserializing then you should not need to include a check on the method argument.

Using `method` method is usually reserved for the rare case where an API returns data using one key, but then expects it to be listed under a different key when updating.

Dremora: That makes sense now, thanks! I'll update the PR.

bmac:
Thanks!

PR #3036, opened 2015-05-02
Remove logic for camelCase-ing error keys in ActiveModelAdapter

pangratz:
This addresses my comment made in https://github.com/emberjs/data/pull/3030#issuecomment-98396379, where I stated that the made changes of camelCase-ing the error keys are a serializer concern:

Since c947e25 (introduced in #3030) the keys of the error object are camelCased in the adapter, which is actually a serializer concern. The JSONSerializer, from which the ActiveModelSerializer extends, does a normalization of the errors object in the `extractErrors` hook (see http://git.io/vJ0A6 & http://git.io/vJ0AS) and this hook gets called by the store when the adapter rejects (see http://git.io/vJ0AI), the camelCase can be removed from inside the adapter.

cc @fivetanley @bdvholmes

### DIFF
--- a/packages/activemodel-adapter/lib/system/active-model-adapter.js
+++ b/packages/activemodel-adapter/lib/system/active-model-adapter.js
@@ -6,7 +6,6 @@ import {pluralize} from "ember-inflector";
   @module ember-data
 */
 
-var camelize = Ember.String.camelize;
 var decamelize = Ember.String.decamelize;
 var underscore = Ember.String.underscore;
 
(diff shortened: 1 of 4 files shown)


fivetanley:
Oh interesting. So it already camelCased, the tests were just wrong?

bdvholmes:
Looks like it. I should have checked to see how `extractErrors` was being done here.

pangratz:
There were no tests for `extractErrors`, so @bdvholmes raised an important issue.

igorT:
Can you rebase please?

pangratz:
@igorT done

igorT:
Thanks

PR #3040, opened 2015-05-04
JSONSerializer#extractErrors respects custom key mappings

pangratz:
Custom key mappings defined in the `attrs` hash of the JSONSerializer
are now considered when extracting errors.


igorT:
Thanks!

Issue #3013, opened 2015-04-20
isLoading `true` after failed response from server


g13013:
I have two Models
Model1-key --- hasMany -- Model2 (async)

when ember-data loads the records and one of the records request fail, its model `isLoading` property still always be `true`

igorT:
Hi,
could you write a test for this please? It also seems related to #3011

pangratz:
Is this the test case we are talking about?

```js
var get = Ember.get;
var hasMany = DS.hasMany;
var Post, Comment, env;
var run = Ember.run;

module("integration/load - Loading Records", {
  setup: function() {
    Post = DS.Model.extend({
      comments: hasMany({ async: true })
    });

    Comment = DS.Model.extend();

    Post.toString = function() { return "Post"; };
    Comment.toString = function() { return "Comment"; };

    env = setupStore({ post: Post, comment: Comment });
  },

  teardown: function() {
    run(env.container, 'destroy');
  }
});

test("isLoading is false when request failed", function() {
  env.adapter.find = function(store, type, id, snapshot) {
    if (type === Post) {
      return Ember.RSVP.resolve({ id: id, comments: [1, 2] });
    }

    // reject second find call for comment
    if (type === Comment && id === '2') {
      return Ember.RSVP.reject();
    }

    return Ember.RSVP.resolve({ id: id });
  };

  run(function() {
    var post;

    env.store.find('post', 1).then(function(lePost) {
      post = lePost;

      equal(post.get("isLoading"), false, "post is not loading anymore");
      return post.get('comments');
    }).then(null, function() {
      equal(post.get("isLoading"), false, "post is not loading");
      equal(post.get('comments').objectAt(0).get("isLoading"), false, "first comment is no more loading");

      // fails here: isLoading is true but should be false
      equal(post.get('comments').objectAt(1).get("isLoading"), false, "second comment is no more loading");
    });
  });
});

```

sly7-7:
@pangratz LGTM, did you plan to fix this ?

pangratz:
@sly7-7 yup, I give it a try.

pangratz:
[...]

PR #3043, opened 2015-05-04
Fix bug where record rejected via `find` stayed in loading state

pangratz:
`store.getById(typeClass, id)` is used inside the rejection handler of
`find` to get the record and invoke the corresponding hooks. This is
buggy since the record is not loaded yet into the store - a successful
`find` would mark it as such - and hence the `store.getById` returned
`null`. That's why the hooks are not called and the record stays in the
`loading` state, having the `isLoaded` flag still being set to `false`.

The fix is to reuse the `record` parameter, which is passed to the
`find` method. By this, the hooks are invoked and the record is passed
transitioned out of the `loading` state.

---

This partially addresses #3013

### DIFF
--- a/packages/ember-data/lib/system/store/finders.js
+++ b/packages/ember-data/lib/system/store/finders.js
@@ -29,13 +29,11 @@ export function _find(adapter, store, typeClass, id, record) {
       return store.push(typeClass, payload);
     });
   }, function(error) {
-    var record = store.getById(typeClass, id);
-    if (record) {
-      record.notFound();
-      if (get(record, 'isEmpty')) {
-        store.unloadRecord(record);
-      }
+    record.notFound();
+    if (get(record, 'isEmpty')) {
+      store.unloadRecord(record);
     }
+
     throw error;
   }, "DS: Extract payload of '" + typeClass + "'");
 }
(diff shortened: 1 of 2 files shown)


packages/ember-data/lib/system/store/finders.js
```diff
@@ -29,13 +29,11 @@ export function _find(adapter, store, typeClass, id, record) {
       return store.push(typeClass, payload);
     });
   }, function(error) {
-    var record = store.getById(typeClass, id);
```
pangratz: This line has been added in [63795a4](https://github.com/emberjs/data/commit/63795a40c833eb23b6f3e78841b65bee4418278c#diff-1636bea36bbe333a834ea11fbfbde48dR1385), where the `record` parameter was not available. I guess that's the reason why a `var record = store.getById(typeClass, id)` has been used here; the `record` parameter is added in [51728a8](https://github.com/emberjs/data/commit/51728a806426309f494de5d8930b9f8951ffac6a#diff-1636bea36bbe333a834ea11fbfbde48dR1869). So I think reusing the param down below is save...

packages/ember-data/tests/integration/records/load-test.js
```diff
@@ -0,0 +1,39 @@
...
+});
+
+test("When loading a record fails, the isLoading is set to false", function() {
+  env.adapter.find = function(store, type, id, snapshot) {
+    return Ember.RSVP.reject();
+  };
+
+  run(function() {
+    env.store.find('post', 1).then(null, function() {
+      // store.recordForId is private, but there is currently no other way to
```
igorT: Yup, it's ok to use private methods like this in the test

igorT:
Thanks!

PR #2687, opened 2015-01-08
Run through Ember Build Pipeline

seanpdoyle:
* Share build process found in [emberjs/emberjs-build][1]
* Easiest approach was to copy paste from [emberjs/ember.js][2]
* In the long term, we should move common build tasks / setup to
[emberjs/emberjs-build][1]

[1]: https://github.com/emberjs/emberjs-build
[2]: https://github.com/emberjs/ember.js

PR #3053, opened 2015-05-08
Extract util/computed-polyfill to it's own addon

cibernox:
I'm tired of receiving deprecation warnings from 3rd party addons, and this solution can make extremely easy to addon authors to support new and old versions of ember easily.

My personal cruise now is get this addon used in the most popular addons and get my app free of warnings except those comming from my own code.


rwjblue:
Is this different than 'ember-new-computed' addon? I made that when the syntax  landed initially so authors could always use the new syntax.

cibernox:
Oh, no. It's exactly the same. Damn, you're too fast!

Ok, I'm closing

cibernox:
Changed my mind. I'll keep the PR open and include your addon instead.

rwjblue:
Sounds good, also your implementation may be better (you are the initial author of the feature ;) )....

wecc:
I think there was a reason for not using `ember-new-computed` directly in ED, but I cannot remember what it was...

cibernox:
@wecc I can't think in any reason for not using it. The code is almost identical, just lives somewhere else.

wecc:
@cibernox Yeah, current code was just copied from the addon. Let's do it!

cibernox:
@wecc @rwjblue I discovered the reason. `ember-new-computed` uses ES6 syntax and ED doesn't uses babel in the compile process.

Do you prefer to add babel to the build process of ED or convert the addon to ES5?

wecc:
@cibernox Ah, nice catch! There's work being done in #2687 to use ember-build for ED. Not sure of the current status though.

cibernox:
Seems that the PR is being holded by Node0.12/io.js clunkyness (because LOL javascript).

Seems that the quickest solution would be to make the addon ES5, but that probably it's only a problem in ED. Most other addons will use the standard ember-cli stack, which includes babel.

fivetanley:
@cibernox Ember Data has babel now. Does that help unblock this?

cibernox:
@fivetanley Probably yes. I'll check it.

cibernox:
@fivetanley confirm. Ready to go.

fivetanley:
thanks!

PR #3054, opened 2015-05-08
Fix for deleting from ManyArray

wagenet:
Basically, if you delete the final object from a ManyArray elsewhere the following will happen:

* The ManyArray will report the correct length of 0.
* The ManyArray's `objectAt(0)` will still return a value.

The issue appears to be that the `currentState` is correctly empty, but the `canonicalState` still retains the deleted item. Because `objectAt` has a fallback, we incorrectly use the value from the `canonicalState`.

### DIFF
--- a/packages/ember-data/tests/integration/relationships/has-many-test.js
+++ b/packages/ember-data/tests/integration/relationships/has-many-test.js
@@ -821,7 +821,7 @@ test("Only records of the same base type can be added to a polymorphic hasMany r
 });
 
 test("A record can be removed from a polymorphic association", function() {
-  expect(3);
+  expect(4);
 
   run(function() {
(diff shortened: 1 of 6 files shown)


packages/ember-data/tests/integration/relationships/has-many-test.js
```diff
@@ -1406,3 +1406,25 @@ test("hasMany hasData sync created", function () {
     equal(relationship.hasData, true, 'relationship has data');
   });
 });
+
+test("destroying an item properly updates a hasMany", function() {
+  expect(4);
+
+  env.adapter.deleteRecord = function() {
+    return { id: 1 };
```
igorT: you can just return RSVP.resolve() as deletion doesn't usually return data

PR #3055, opened 2015-05-08
Merge `attrs` from superclasses into their subclasses.

omghax:
This PR defines `attrs` as a `mergedProperty`, so it can be inherited from superclasses. This is useful if you have a polymorphic type and want to define a base serializer for common attributes.

### DIFF
--- a/packages/ember-data/lib/serializers/json-serializer.js
+++ b/packages/ember-data/lib/serializers/json-serializer.js
@@ -98,6 +98,7 @@ export default Serializer.extend({
     @property attrs
     @type {Object}
   */
+  mergedProperties: ['attrs'],
 
   /**
    Given a subclass of `DS.Model` and a JSON object this method will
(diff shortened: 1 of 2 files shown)


igorT:
Thanks

pangratz:
Nice

igorT:
Would be nice to point out somewhere in the docs

pangratz:
What about merging hashes?

``` js
var BaseSerializer = DS.JSONSerializer.extend({
  attrs: {
    comments: {
      key: "le_comments"
    }
  }
});

var PostSerializer = BaseSerializer.extend({
  attrs: {
    comments: {
      anotherOptionForCustomSerializer: true
    }
  }
});
```

results currently in the following `attrs` hash:

```
{
  comments: {
    anotherOptionForCustomSerializer: true
  }
}
```

Do we want the result to be:

```
{
  comments: {
    key: "le_comments",
    anotherOptionForCustomSerializer: true
  }
}
```

??

igorT:
I feel like having to look at multiple classes to be sure of what the hash
is pretty error prone. Not sure this is a really common use case and would
be confusing. Open to being persuaded otherwise

pangratz:
Yeah, you might be right that this is not a common use case ...

omghax:
My mental model of `attrs` was that it behaved like `actions` in Ember, so I was a little surprised when that wasn't the case. Granted `actions` is flat, unlike `attrs`. I'm not in favor of recursively merging though, it seems like that would add more surprises.

PR #3056, opened 2015-05-08
Add more assertions for merged attributes from superclass

pangratz:
This adds assertions that defined keys of attributes in the superclass
can be overwritten by subclasses.

This adds more assertions for the merge-able `attrs` has introduced in #3055.


igorT:
thanks!

PR #3058, opened 2015-05-10
Fix changelog script to point to emberjs/data

tonywok:
Looks like a simple typo. It was pointing to the ember.js repo instead.

cc @fivetanley


fivetanley:
hah, thanks.

PR #3017, opened 2015-05-11
normalizeErrors with nested errors breaks for polymorphic associations

jbourassa:
Calling `normalizeRelationships` from `normalizeErrors` breaks for polymorphic relationships in ActiveModelSerializer (introduced in #2392, [see this diff] (https://github.com/emberjs/data/pull/2392/files#diff-874af7e7df6c1be12e3a54caad8d859dR270)). Here's a [failing test](https://github.com/didacte/data/blob/692febe825f67aa5f82d52661e04530606f9a5bd/packages/activemodel-adapter/tests/integration/active-model-serializer-test.js#L329-#L342).

This happens because ActiveModelSerializer's `normalizeRelationships` will inspect the payload for a `type` property. Since this is an error payload, the `type` property won't be set. Here's the (albeit not so pretty) workaround we have:

```javascript
{
  // stuff...
  normalizeErrors: function(type, hash) {
    this.normalizeId(hash);
    this.normalizeAttributes(type, hash);

    if (this.keyForRelationship) {
      type.eachRelationship(function(key, relationship) {
        var payloadKey = this.keyForRelationship(key, relationship.kind);
        if (key !== payloadKey && hash.hasOwnProperty(payloadKey)) {
          hash[key] = hash[payloadKey];
          delete hash[payloadKey];
        }

        if(!hash || !hash[key]) return;
        // Array.isArray => basically checking for .kind === 'hasMany'
        if(Array.isArray(hash[key])) {
          hash[key].forEach(function(hash) {
            if(hash) this.normalizeErrors(relationship.type, hash);
          }.bind(this))
        }
        else {
          this.normalizeErrors(relationship.type, hash[key]);
        }
      }, this);
    }
  }
}
```
Would you like me to improve this code and submit a PR?

### DIFF
--- a/packages/ember-data/tests/integration/serializers/embedded-records-mixin-test.js
+++ b/packages/ember-data/tests/integration/serializers/embedded-records-mixin-test.js
@@ -1572,3 +1572,33 @@ test("serializing belongsTo correctly removes embedded foreign key", function() 
     }
   });
 });
+
+test('extractErrors extracts embedded errors', function() {
+  SuperVillain.reopen({
+    evilMinions: DS.hasMany('secret-weapon', { polymorphic: true })
+  });
+
+  env.registry.register('serializer:super-villain', DS.ActiveModelSerializer.extend(DS.EmbeddedRecordsMixin, {
+    attrs: {
+      evilMinions: { embedded: 'always' }
+    }
+  }));
+
+  var serializer = env.container.lookup('serializer:super-villain');
+  var payload = {
+    errors: {
+      evil_minions: [
+        { name: ['required'] }
+      ]
+    }
+  };
+
+  serializer.extractErrors(env.store, SuperVillain, payload, null); // id (last param) is not relevant here
+  deepEqual(payload, {
+    errors: {
+      evilMinions: [
+        { name: ['required'] }
+      ]
+    }
+  });
+});


jbourassa:
Hey @bmac,

I just updated the issue with a PR. Let me know what you think!

packages/ember-data/lib/serializers/json-serializer.js
```diff
@@ -274,11 +276,42 @@ export default Serializer.extend({
...
+        hash[key] = hash[payloadKey];
+        delete hash[payloadKey];
+      }
+
+      if (!hash || !hash[key]) {
+        return;
+      }
+      if (relationship.kind === 'hasMany' && isArray(hash[key])) {
+        forEach.call(hash[key], function(hash) {
+          if (hash) {
```
bmac: `hash` as an argument name is a little confusing because there already is a `hash` argument from the parent scope. Lets rename this to something else. Maybe `error`?

jbourassa: Agreed! I used `errors` since any key can hold multiple errors.

bmac: sounds good.

packages/activemodel-adapter/tests/integration/active-model-serializer-test.js
```diff
@@ -341,3 +341,18 @@ test("extractErrors camelizes keys", function() {
...
+      evil_minions: [
+        { name: ['required'] }
+      ]
+    }
+  };
+
+  var serializer = env.container.lookup('serializer:application');
+  serializer.extractErrors(env.store, MediocreVillain, payload, null); // id (last param) is not relevant here
+
+  ok(true);
```
bmac: This test should assert that the record error object was extracted correctly. It might look something like:  

```js
var extractedError = serializer.extractErrors(env.store, MediocreVillain, payload, null); // id (last param) is not relevant here

deepEqual(extractedError, {
  evilMinions: [{ name: ['required'] }]
});
```

bmac:
@igorT @tchak Does Ember Data's Error handling mechanism handle the case were the error object for a hasMany relationship be an array of error objects or do these errors get silently ignored? e.g. 

```js
{
    errors: {
      evil_minions: [
        { name: ['required'] },
        { name: ['required'] }
      ]
}
```

packages/activemodel-adapter/tests/integration/active-model-serializer-test.js
jbourassa: Also very correct, my bad. I didn't look very far and kept my original test that was flat out erroring. That is also fixed!

jbourassa:
@bmac 
> Does Ember Data's Error handling mechanism handle the case were the error object for a hasMany relationship be an array of error objects or do these errors get silently ignored? [...]
I know you didn't ask me — I have no idea if support is intended or not — but it works and we do use it in our app, keeping it would be very nice :pray:

bmac:
Sorry @jbourassa that question was meant to be directed at @igorT. I posted it to this issue to give him some context.

jbourassa:
The tests were passing locally but not from a fresh install: I'll investigate.

packages/activemodel-adapter/tests/integration/active-model-serializer-test.js
jbourassa: Actually, it's a bit more complicated than that. The serializer will _not_ change the key from `evil_minion` to `evilMinions` because it is looking for `evil_minion_ids`.

So there are two things that are relevant in this test:
1) No matter what you pass in, `extractErrors` should crash (eg: _not_ look for a `.type` property on an error payload)
2) If the serializer is configured to rename the `evil_minions` association to `evilMinions`, errors should also be renamed.

I updated the test again to cover (2), but I had to bring in EmbeddedRecord.

packages/ember-data/lib/serializers/json-serializer.js
```diff
@@ -274,11 +276,42 @@ export default Serializer.extend({
...
+    if (!this.keyForRelationship) {
+      return;
+    }
+
+    typeClass.eachRelationship(function(key, relationship) {
+      var payloadKey = this.keyForRelationship(key, relationship.kind, 'deserialize');
+      if (key !== payloadKey && hash.hasOwnProperty(payloadKey)) {
+        hash[key] = hash[payloadKey];
+        delete hash[payloadKey];
+      }
```
pangratz: Since #3040 the `normalizeUsingDeclaredMapping` method is invoked inside `normalizeErrors`: this makes the above lines `293-297` obsolete as the hash already contains the normalized relationship keys. The call for `this. normalizeRelationshipErrors` inside `normalizeErrors` above needs to moved before though, so this works correctly.

packages/ember-data/lib/serializers/json-serializer.js
```diff
@@ -274,11 +276,42 @@ export default Serializer.extend({
...
+
+      if (!hash || !hash[key]) {
+        return;
+      }
+      if (relationship.kind === 'hasMany' && isArray(hash[key])) {
+        forEach.call(hash[key], function(errors) {
+          if (errors) {
+            this.normalizeErrors(relationship.type, errors);
```
pangratz: `this` refers to the serializer of the parent; if I am not mistaken `store.serializerFor(relationship.type).normalizeErrors` should be used here, so the correct serializer is used.

Also, I think we should add an assertion/warning for the case when the length of the errors array is not the same length of the array of records, which would indicate that there are possible errors missing.

packages/ember-data/lib/serializers/json-serializer.js
```diff
@@ -274,11 +276,42 @@ export default Serializer.extend({
...
+          }
+        }, this);
+      } else {
+        this.normalizeErrors(relationship.type, hash[key]);
```
pangratz: same as above with `store.serializerFor(relationship.type).normalizeErrors`

packages/ember-data/lib/serializers/json-serializer.js
```diff
@@ -274,11 +276,42 @@ export default Serializer.extend({
...
+    typeClass.eachRelationship(function(key, relationship) {
+      var payloadKey = this.keyForRelationship(key, relationship.kind, 'deserialize');
+      if (key !== payloadKey && hash.hasOwnProperty(payloadKey)) {
+        hash[key] = hash[payloadKey];
+        delete hash[payloadKey];
+      }
+
+      if (!hash || !hash[key]) {
+        return;
+      }
```
pangratz: Another approach could be to check if the `typeClass` has the `DS.EmbeddedRecordsMixin` mixed in, and if so, check if there would be embedded errors via [`this.hasSerializeRecordsOption(key)`](http://git.io/vTLpG).

jbourassa:
@pangratz Thanks for the feedback / sorry for not nailing it better. This is taking longer than I want but I'll update the PR at some point this week.

pangratz:
> sorry for not nailing it better

@jbourassa I hope I didn't leave the impression you need to apologize and if so I am very sorry! That was absolutely not my intention :worried: . I am very happy you took the time and effort to add this to the core of ember-data. I am using a quite similar implementation within an app but never had the time to submit a patch. Once this lands I can remove that bit :dancers: 

Feel free to ping me if you want further feedback or need anything else. Cheers!

packages/ember-data/lib/serializers/json-serializer.js
jbourassa: >Also, I think we should add an assertion/warning for the case when the length of the errors array is not the same length of the array of records [...]

While I agree with the idea, I don't think it's the serializer role to do that—and we don't have an instance of the record there— it would be at the record level in the current implementation (`adapterDidInvalidate`), which I feel is out of the scope of this PR.

packages/ember-data/lib/serializers/json-serializer.js
jbourassa: I might be missing something but I don't think that's correct. `normalizeUsingDeclaredMapping` only normalize _declared_ keys, not association defaults defined with `keyForRelationship`.

packages/ember-data/lib/serializers/json-serializer.js
jbourassa: Actually, I'd like to avoid referencing anything embedded-record related in the JSON serializer. I just pushed what I think is a more robust implementation.

packages/ember-data/lib/serializers/json-serializer.js
pangratz: Good catch! You're right, I haven't thought about `keyForRelationship` :+1:

tchak:
We are moving internal representation of errors to a son-api format. It means there is no such thing as nested errors or nested relationships on internal representation. We are going to need to normalize to something flat with the right pointer.

Something like ?
``` json
{
  errors: [
    {
      title: 'required',
      source: { pointer: 'included/3/attributes/name' }
    }
  ]
}

```

jbourassa:
Got it! We're not transitioning our errors to json-api in in the next few weeks so I'll go ahead and close this PR.

igorT:
@jbourassa why close if you are not transitioning, this is still a bug we should fix?

jbourassa:
@igorT Well, would you rather me to reopen and revert the last commit that handles nested relationships? If I understand correctly, this code and the original bug will be obsolete when json-api errors are implemented, so I don't see why you'd want to merge this.

igorT:
we still want the failing test though?

jbourassa:
I might be missing something but if ember-data drops support for non-json-api error payload, the failing test will also be a irrelevant, no?

igorT:
We are switching our internal error format to json api, but active model serializer should still work.

jbourassa:
Gotcha! I'll reopen and leave the failing test.

jbourassa:
Done — I forced-push an updated test and rebased it on master.

fivetanley:
@jbourassa Sorry for the late follow-up. Can you update the PR or close it if it's no longer needed?

jbourassa:
This is still an issue with 1.13 (the version I'm on), but I think it's gone in 2.3 — I can't find the offending code in master. That said, I feel that this use case is pretty far from where ember data is going with JSON API, I don't think it's a bug at this point but just something that is not supported.

Let's close this.

Issue #3069, opened 2015-05-14
Deprecation warnings with Ember 1.12.0


SaladFork:
I've updated to `ember-data@1.0.0-beta.17` but am still seeing deprecation warnings:

- `DEPRECATION: lookupFactory was called on a Registry. The initializer API no longer receives a container, and you should use an instanceInitializer to look up objects from the container.` from: [ember-data/lib/initializers/store.js:21](https://github.com/emberjs/data/blob/174fc94488d7e66ebb8b61e3bef142c0d051edca/packages/ember-data/lib/initializers/store.js#L21)

- `DEPRECATION: lookup was called on a Registry. The initializer API no longer receives a container, and you should use an instanceInitializer to look up objects from the container.` from: [ember-data/lib/initializers/store.js:39](https://github.com/emberjs/data/blob/174fc94488d7e66ebb8b61e3bef142c0d051edca/packages/ember-data/lib/initializers/store.js#L39)

bmac:
Looks like this is a duplicate of https://github.com/emberjs/data/issues/3051

PR #3070, opened 2015-05-14
Refactor the ember initializers to use instanceInitializer

bmac:
This fixes deprecation warnings with Ember 1.12. (https://github.com/emberjs/data/issues/3069)

This pr is currently blocked on https://github.com/emberjs/ember.js/issues/11172.

### DIFF
--- a/packages/ember-data/lib/initializers/store.js
+++ b/packages/ember-data/lib/initializers/store.js
@@ -11,14 +11,22 @@ import Store from "ember-data/system/store";
   @param {Ember.Registry} registry
   @param {Object} [application] an application namespace
 */
-export default function initializeStore(registry, application) {
+export default function initializeStore(paramOne, paramTwo) {
+  // initializerStore be called by an Ember.initializer
+  // when used with ember 1.11.0 or below.
+  var params = getRegistryAndContainer(paramOne, paramTwo);
+  var application = params.application;
+  var registry = params.registry;
+  var container = params.container;
+
   Ember.deprecate('Specifying a custom Store for Ember Data on your global namespace as `App.Store` ' +
                   'has been deprecated. Please use `App.ApplicationStore` instead.', !(application && application.Store));
 
   registry.optionsForType('serializer', { singleton: false });
   registry.optionsForType('adapter', { singleton: false });
 
-  registry.register('store:main', registry.lookupFactory('store:application') || (application && application.Store) || Store);
+  // Change to `application.register` once we drop support for Ember pre 1.12
+  registry.register('store:main', container.lookupFactory('store:application') || (application && application.Store) || Store);
 
   // allow older names to be looked up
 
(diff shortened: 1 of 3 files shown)


igorT:
cc I know @twokul had some thoughts about this stuff

packages/ember-data/lib/initializers/store.js
```diff
@@ -12,13 +12,20 @@ import Store from "ember-data/system/store";
...
+  var container = params.container;
+
   Ember.deprecate('Specifying a custom Store for Ember Data on your global namespace as `App.Store` ' +
                   'has been deprecated. Please use `App.ApplicationStore` instead.', !(application && application.Store));
 
   registry.optionsForType('serializer', { singleton: false });
   registry.optionsForType('adapter', { singleton: false });
 
-  registry.register('store:main', registry.lookupFactory('store:application') || (application && application.Store) || Store);
+  registry.register('store:main', container.lookupFactory('store:application') || (application && application.Store) || Store);
```
twokul: should it be `application.register`?

packages/ember-data/lib/initializers/store.js
```diff
@@ -12,13 +12,20 @@ import Store from "ember-data/system/store";
   @param {Object} [application] an application namespace
 */
 export default function initializeStore(registry, application) {
+  // initializerStore be called by an Ember.initializer
+  // when used with ember 1.11.0 or below.
+  var params = getRegistryAndContainer(registry, application);
+  application = params.application;
```
twokul: it's a bit confusing to override passed in arguments, perhaps name variables differently?

twokul:
@bmac @igorT looks good I think. ember bug looks odd

bmac:
I had a conversation this morning with @mixonic about: 
https://github.com/emberjs/data/blob/master/packages/ember-data/lib/initializers/store.js#L21. He suggested an alternate approach that Ember Data could use to register `store:main` as an initializer without the deprecation warning.

```
Matthew Beale [10:02 AM]
the idea of the initializer split

Matthew Beale [10:02 AM]
is that you get one pass to set everything up

Matthew Beale [10:02 AM]
make all the registers, all the injections

Matthew Beale [10:03 AM]
the world is ready to go

Matthew Beale [10:03 AM]
nobody has cut the line to instantiate themselves

Matthew Beale [10:03 AM]
so any initializer can inject on to any other factory

Matthew Beale [10:03 AM]
then in the instance initializers you look things up if need be

Matthew Beale [10:04 AM]
having a line where you look something up and register it doesn't make a lot of sense in that world

Matthew Beale [10:05 AM]
factories in the container are just create or extendable

Matthew Beale [10:05 AM]
they don't need to be classes, technically

Matthew Beale [10:05 AM]
they just need to return a factory or instance

Matthew Beale [10:05 AM]
I wonder if the easiest path isn't just to write a shim and register that for store:main

Matthew Beale [10:05 AM]
and when extend or create is called on the shim

Matthew Beale [10:06 AM]
do the lookup steps for application || App.store, etc

Matthew Beale [10:07 AM]
the way the container works, if we just called this store:application everything would be fine

Matthew Beale [10:07 AM]
since a looked-up factory will override a registered factory

Matthew Beale [10:07 AM]
(not what you expect, but this is exactly why)

Brendan McLoughlin [10:10 AM] 
I think I like the idea of using store:application

Brendan McLoughlin [10:11 AM]
Will need to do some research to find out what might break if ED moves away from store:main

Matthew Beale [10:11 AM] 
aye

Matthew Beale [10:11 AM]
ideally I think we do the shim with a deprecation warning
```

btecu:
Could we get this in `beta 18`?

igorT:
@bmac whats the status of this?

packages/ember-data/lib/initializers/store.js
bmac: I think you are right and `application.register` is correct in the new model. However, that will break when ED is used with older version of Ember. I'll add a comment to the codebase to take this out when we drop support for old Ember.

bmac:
Closing this in favor of https://github.com/emberjs/data/pull/3136

Issue #3072, opened 2015-05-16
Assert against passing classes to public methods on the store


igorT:

igorT:
I believe this was done

PR #3074, opened 2015-05-16
remove passing factories to store methods

fivetanley:
Previously, we allowed users to either pass factories (subclasses of
DS.Model), or a string to store methods. For more consistency in the
container-based world where things are looked up through strings,
support for looking up via passing a factory has been removed.

This is part of a refactor that will remove state (like the store)
from Ember Data's factories in order to remove
Ember.MODEL_FACTORY_INJECTIONS.


fivetanley:
closes #3072

igorT:
I really wish I waited a day before rebasing references, this has been a source of major pain. :heart: :heart: Awesome work!

packages/ember-data/lib/system/store.js
```diff
@@ -1847,15 +1862,12 @@ Store = Service.extend({
...
     @return DS.Adapter
   */
-  adapterFor: function(type) {
-    if (type !== 'application') {
-      type = this.modelFor(type);
-    }
-
-    var adapter = this.lookupAdapter(type.modelName) || this.lookupAdapter('application');
+  adapterFor: function(modelName) {
+    Ember.assert('Passing classes to store methods has been removed. Please pass a dasherized string instead of '+ Ember.inspect(modelName), typeof modelName === 'string');
```
igorT: I think we might want to deprecate here if our own code was doing it.

packages/ember-data/lib/system/store.js
```diff
@@ -1886,18 +1898,16 @@ Store = Service.extend({
...
-    @param {String or subclass of DS.Model} type the record to serialize
+    @param {String} modelName the record to serialize
     @return {DS.Serializer}
   */
-  serializerFor: function(type) {
-    if (type !== 'application') {
-      type = this.modelFor(type);
-    }
+  serializerFor: function(modelName) {
+    Ember.assert('Passing classes to store methods has been removed. Please pass a dasherized string instead of '+ Ember.inspect(modelName), typeof modelName === 'string');
```
igorT: Same here

packages/ember-data/lib/system/store.js
fivetanley: I fixed the places our own code was doing it.. Seems odd to ship a version that triggers deprecation warnings?

igorT: I am saying we should fix our own code as well, yes, but if we were doing it, seems like other people could've as well, and it would be nicer not to break their apps.

fivetanley: Their apps would be broken by everything else asserting anyway.

fivetanley: The fix is also pretty easy for others to incorporate.

fivetanley:
before the merge of #3075 :

https://github.com/emberjs/data/compare/2f8024be6863b7a2b3768588760686239b7208d0...b213ce18e2cb99ce87e754c5bb47929e3db578f9

You can look at the diff in #3075 for removing the store from factories.

packages/ember-data/lib/system/debug/debug-adapter.js
```diff
@@ -42,7 +42,11 @@ export default Ember.DataAdapter.extend({
     return columns;
   },
 
-  getRecords: function(modelName) {
+  getRecords: function(modelNameOrFactory) {
+    // TODO: Ask Teddy what we should do here.
+    // Ideally this should always get passed a string.
+
+    var modelName = typeof modelNameOrFactory === 'string' ? modelNameOrFactory : modelNameOrFactory.modelName;
```
stefanpenner: maybe deprecate?

stefanpenner:
SGB

packages/ember-data/lib/system/debug/debug-adapter.js
fivetanley: I don't think this gives user's an action as it's in the debug adapter for the Ember Inspector. We try to avoid deprecations for code that isn't the user's.

igorT:
Seems great except for the passing of the store to inverseFor

packages/ember-data/lib/system/store/factory-locators.js
```diff
@@ -0,0 +1,74 @@
...
+
+  if (!factory) {
+    //Support looking up mixins as base types for polymorphic relationships
+    factory = modelForMixin(container, modelName);
+  }
+  if (!factory) {
+    throw new Ember.Error("No model was found for '" + modelName + "'");
+  }
+  factory.modelName = factory.modelName || normalizeModelName(modelName);
+  factory.__container__ = factory.__container__ || container;
```
igorT: I am very suspicious of this. Seems like it would break for multiple app runs server side. cc @stefanpenner

stefanpenner: Static methods just can't return concrete references to other factories.

stefanpenner: Although this may appear drastic, this appears like the last lynch-pin preventing ED from being a good citizen in this area.

igorT: We need a factory in order to be able to reflect on it. To say if I am a `user` and I have `message`, does that message have any relationships that point to a `user`

stefanpenner: Can you show me that code? Or provide an expanded scenario by which i can continue to think about this?

stefanpenner: in slack we discussed this further. I provided the following example:

As nothing prevents two apps from having two different set of resolvers, factory level reflection without polluting instance state isn't really available to us.

The solution is to continue to purge instance state (the container/ store etc) from factories, and instead enable the needing wiring when instance information is present.

fivetanley:
:tada:

bmac:
:dancers:

PR #3079, opened 2015-05-18
restore backwards compatibility of buildURL

### LINKED ISSUE #3064: DS.RESTAdapter.buildURL receives query instead of id in 1.0.0-beta.17
buildURL signature is documented as ```buildURL (typeKey, id, snapshot, requestType)```, but in beta 17 query object is passed instead of id. See findQuery() in rest-adapter.js.

Beta 16.1 works fine.

fivetanley:
closes #3064


fivetanley:
cc @amiel you'll need to update this for your URL addon.

amiel:
@fivetanley thanks a ton for the notice

amiel:
Also, thanks for this change. As I mentioned in https://github.com/emberjs/data/pull/2966#discussion-diff-28162729, I prefer this :)

Issue #2991, opened 2015-04-12
Bring back references


igorT:
A reference is a javascript object that has data for a particular record object, particularly the id and the foreign key data. It is used so we do not have to instantiate a DS.Model for every single foreign key, and to support more advanced unloading cases. 

We used to have them, but they were removed to simplify the code base and because they were a massive pain in the old non promise based ED. It is time to bring them back, as foreign key creation of records is a huge perf bottleneck.

The main issue when considering implementing them is to make sure all internal ED methods *either* deal with record objects or with references but never with both, as that makes code very hard to reason about.

wycats:
:+1: assuming the last paragraph of the description:

> The main issue when considering implementing them is to make sure all internal ED methods either deal with record objects or with references but never with both, as that makes code very hard to reason about.

igorT:
Work ongoing at https://github.com/emberjs/data/tree/references

mphasize:
Would this also support some kind of lazy casting of polymorphic relationships?

I have a situation where an API (which I cannot change) has polymorphic relationships, but neither embeds the records nor supplies a type alongside the ID. Since we already have a "parent" type for polymorphic relationships, I was wondering if the related record could be known to Ember Data just as the ID and the parent type until it gets fetched from the API. When we finally have the complete information, the type could be recast to the child type. (I should also mention that all records are accessible through the same endpoint, so it's not necessary to build a specific URL.)

Does that fit here or should this go into a separate issue?

igorT:
Yes, do you want to write tests for that feature?

mphasize:
Where would I start? (I've been using Ember a lot, but I haven't done much development for Ember yet... but I can give it a shot.)

igorT:
Look at polymorphic tests, like https://github.com/emberjs/data/blob/master/packages/ember-data/tests/integration/relationships/has-many-test.js#L630 
and try to write new tests that use the api you proposed, and assert that it works. 
This would be a great chance to also clean up the tests, and move polymorphic stuff into their own files.
I think the rest adapter integration tests have a bunch of polymorphic stuff that should be moved out

mphasize:
Ok, I will see what I can do there and take a look what I can learn from the rest adapter tests.

mphasize:
Quick question: this behavior is expected to be independent of the adapter/serializer being used, right?

igorT:
Yup. Thats why I was saying that there are probably tests in integration
adapter/serialzier that don't belong there

[...]

PR #3087, opened 2015-05-21
Add a test for async polymorphic relationships with lazy type inference

mphasize:
This test describes a situation where an API has polymorphic relationships, but neither embeds the records nor supplies a type alongside the ID. Since we already have a "parent" type for polymorphic relationships, I was wondering if the related record could be known to Ember Data just as the ID and the parent type until it gets fetched from the API. When we finally have the complete information, the type could be recast to the child type. 

There would be 2 requirements: 
- all the polymorphic records have to be accessible on the endpoint of the parent type
- all polymorphic records need to have unique IDs

A bit of related discussion:
https://github.com/emberjs/data/issues/2991

I should mention that this comes from a real-world example. I was trying to build an Ember Data integration for the [Contentful](https://www.contentful.com/developers/documentation/content-management-api/) content management API.

There should be more tests to verify that the records have been recast to their correct type after loading, but I will wait with adding them until I have some first feedback on this test.


packages/ember-data/tests/integration/relationships/has-many-test.js
```diff
@@ -1407,3 +1407,40 @@ test("hasMany hasData sync created", function () {
...
+      id: 1,
+      messages: [2, 3]
+    });
+
+    env.store.find('user', 1)
+      .then(function (user) {
+        return user.get('messages');
+      })
+      .then(function (messages) {
+        equal(messages.get('length'), 2, "The contacts relationship is correctly set up");
```
igorT: You would need to assert the type and the data in the messages here, otherwise the test could pass even if polymorphism didn't work

mphasize:
@igorT I added tests for the data in the records, but I am not sure how I would test for the type/class of the record. (During a quick look at the other tests I couldn't find an example.)

mphasize:
@igorT Anything else I could do to improve this test?

igorT:
@mphasize we should have this working shortly. Do you want to write the same tests for belongsTo?

mphasize:
@igorT Sounds great. I can add the test for `belongsTo`.

mphasize:
@igorT Added test for polymorphic belongsTo. Anything else I could do to help?

igorT:
Can you clean up the PR a bit please, take only your commits and rebase them off master?

igorT:
We are blocked on some api design, but it shouldn't be too hard

igorT:
We are blocked on some api design, but it shouldn't be too hard

mphasize:
Closing in favor of https://github.com/emberjs/data/pull/3304

PR #3088, opened 2015-05-22
Deprecation test: pathForType -> typeForRoot

Dremora:
Just a test name fix, but it is currently quite confusing given that `pathForType` hasn't been deprecated.


sly7-7:
Thanks :+1:

PR #3094, opened 2015-05-24
Lazily materialize DS.Models for app code, use InternalModel inside ED otherwise

igorT:


packages/ember-data/lib/system/many-array.js
```diff
@@ -143,7 +148,7 @@ export default Ember.Object.extend(Ember.MutableArray, Ember.Evented, {
       this.get('relationship').removeRecords(records);
     }
     if (objects) {
-      this.get('relationship').addRecords(objects, idx);
+      this.get('relationship').addRecords(map(objects, function(obj) { return obj.reference; }), idx);
```
igorT: Could use mapBy

packages/ember-data/lib/system/reference.js
```diff
@@ -0,0 +1,684 @@
...
+  materializeRecord: function() {
+    // lookupFactory should really return an object that creates
+    // instances with the injections applied
+    this.record = this.type._create({
+      id: this.id,
+      store: this.store,
+      container: this.container
+    });
+    this.record.reference = this;
+    //TODO Probably should call deferred triggers here
```
igorT: Blocker, make sure deferred triggers are called upon instantiating? Double check with wycats.

packages/ember-data/lib/system/reference.js
```diff
@@ -0,0 +1,684 @@
...
+  },
+
+  reload: function() {
+    this.startedReloading();
+    var record = this;
+    var promiseLabel = "DS: Model#reload of " + this;
+    return new Promise(function(resolve) {
+      record.send('reloadRecord', resolve);
+    }, promiseLabel).then(function() {
+      //TODO FIXMEMEMEMEME
```
igorT: blocker

packages/ember-data/lib/system/reference.js
```diff
@@ -0,0 +1,684 @@
...
+    if (this._deferredTriggers.push(args) !== 1) {
+      return;
+    }
+    Ember.run.schedule('actions', this, '_triggerDeferredTriggers');
+  },
+
+  //TODO double check whether we care about having the record
+  _triggerDeferredTriggers: function() {
+    //TODO Worry about didLoad etc. that people might be defining/using
+    //Queue up, to be ran once there is a record
```
igorT: Blocker

packages/ember-data/lib/system/reference.js
```diff
@@ -0,0 +1,684 @@
...
+  _preloadBelongsTo: function(key, preloadValue, type) {
+    var recordToSet = this._convertStringOrNumberIntoRecord(preloadValue, type);
+
+    //We use the pathway of setting the hasMany as if it came from the adapter
+    //because the user told us that they know this relationships exists already
+    this._relationships[key].setRecord(recordToSet);
+  },
+
+  //TODO Rename to reference
+  _convertStringOrNumberIntoRecord: function(value, type) {
```
igorT: Blocker

packages/ember-data/lib/system/reference.js
```diff
@@ -0,0 +1,684 @@
...
+  },
+
+  becameReady: function() {
+    var self = this;
+    Ember.run.schedule('actions', function() {
+      self.store.recordArrayManager.recordWasLoaded(self);
+    });
+  },
+
+  didIinitalizeData: function() {
```
pangratz: typo

packages/ember-data/lib/system/record-arrays/record-array.js
```diff
@@ -90,8 +90,11 @@ export default Ember.ArrayProxy.extend(Ember.Evented, {
...
   objectAtContent: function(index) {
     var content = get(this, 'content');
-
-    return content.objectAt(index);
+    if (content.objectAt(index)) {
+      return content.objectAt(index).getRecord();
+    } else {
+      return content.objectAt(index);
+    }
   },
```
pangratz: `content.objectAt()` is invoked 3 times...

``` js
var reference = content.objectAt(index);
return reference && reference.getRecord();
```

packages/ember-data/lib/system/model/model.js
```diff
@@ -1061,15 +683,9 @@ var Model = Ember.Object.extend(Ember.Evented, {
...
-    this.store.scheduleSave(this, resolver);
-    this._inFlightAttributes = this._attributes;
-    this._attributes = Ember.create(null);
-
-    return PromiseObject.create({
-      promise: resolver.promise
+    var model = this;
+    return this.reference.save().then(function() {
+      return model;
     });
```
pangratz: Previously a `PromiseObject` has been returned here, where now the promise itself is returned. Is this intended?

packages/ember-data/lib/system/relationships/belongs-to.js
```diff
@@ -89,19 +89,21 @@ function belongsTo(modelName, options) {
...
+      return this.reference._relationships[key].getRecord();
     },
     set: function(key, value) {
       if (value === undefined) {
         value = null;
       }
       if (value && value.then) {
-        this._relationships[key].setRecordPromise(value);
+        this.reference._relationships[key].setRecordPromise(value);
+      } else if (value) {
```
pangratz: should this be a check for `if (value.reference)`?

packages/ember-data/tests/integration/adapter/find-all-test.js
```diff
@@ -10,6 +10,9 @@ module("integration/adapter/find_all - Finding All Records of a Type", {
       firstName: DS.attr('string'),
       lastName: DS.attr('string')
     });
+    //TODO mock for now
+    Person.typeKey = 'person';
+    Person.modelName = 'person';
```
pangratz: Since #3033 is merged and `references` is rebased onto `master`, this can be removed.

packages/ember-data/tests/integration/adapter/find-test.js
```diff
@@ -9,6 +9,9 @@ module("integration/adapter/find - Finding Records", {
       firstName: DS.attr('string'),
       lastName: DS.attr('string')
     });
+    Person.typeKey = 'person';
+    Person.modelName = 'person';
+
```
pangratz: Since #3033 is merged and `references` is rebased onto `master`, this can be removed.

packages/ember-data/tests/unit/model-test.js
```diff
@@ -151,6 +152,7 @@ test("it should use `_reference` and not `reference` to store its reference", fu
     });
   });
 });
+*/
```
pangratz: I guess this test can be removed.

packages/ember-data/lib/system/reference.js
igorT: Thanks! Luckily internal method.

packages/ember-data/lib/system/model/model.js
igorT: awesome catch

packages/ember-data/lib/system/relationships/belongs-to.js
igorT: No, because either you passed a promise, record or null/undefined. Could add an assert though

igorT:
Thanks @pangratz

packages/ember-data/lib/system/model/model.js
pangratz: Should I PR this branch with a fix?

packages/ember-data/lib/system/relationships/belongs-to.js
pangratz: I was just wondering since inside this branch the `value.reference` is accessed; since a record has a `reference` this seems alright ...

packages/ember-data/lib/system/model/model.js
igorT: Sure, that would be awesome, for the other things you found as well.

packages/ember-data/lib/system/model/model.js
```diff
@@ -887,8 +599,8 @@ var Model = Ember.Object.extend(Ember.Evented, {
       and value is an [oldProp, newProp] array.
   */
   changedAttributes: function() {
-    var oldData = get(this, '_data');
-    var newData = get(this, '_attributes');
+    var oldData = get(this.reference, '_data');
```
fivetanley: it seems like this logic should move into the reference object rather than reaching into its private api.

something like `Reference.prototype.changedAttributes`. Then this function becomes `this.reference.changedAttributes()`

packages/ember-data/tests/unit/model-test.js
igorT: I think we should _ whatever we end up calling references, so probably want to keep this

packages/ember-data/lib/system/model/model.js
```diff
@@ -1001,41 +647,17 @@ var Model = Ember.Object.extend(Ember.Evented, {
...
-  /**
+
+  //TODO not sure what to do
+  /*
     @method _createSnapshot
     @private
   */
   _createSnapshot: function() {
-    return new Snapshot(this);
+    return this.reference._createSnapshot();
```
fivetanley: if references are private objects we should be able to call methods on it without using its private api, so it seems like it would be ok to remove the underscore?

packages/ember-data/lib/system/model/states.js
```diff
@@ -329,7 +326,7 @@ var DirtyState = {
     },
 
     didSetProperty: function(record, context) {
-      get(record, 'errors').remove(context.name);
+      record.getErrors().remove(context.name);
```
fivetanley: i know these are probably just some methods moved around, but it feels weird to mutate another object's property's state. perhaps we could call record.removeError(context.name) ?

packages/ember-data/lib/system/model/states.js
```diff
@@ -337,12 +334,12 @@ var DirtyState = {
     becomeDirty: Ember.K,
 
     willCommit: function(record) {
-      get(record, 'errors').clear();
+      record.getErrors().clear();
```
fivetanley: same here, record.clearErrors()

packages/ember-data/lib/system/model/model.js
pangratz: :+1:

packages/ember-data/lib/system/record-array-manager.js
```diff
@@ -79,7 +79,7 @@ export default Ember.Object.extend({
 
   //Don't need to update non filtered arrays on simple changes
   _recordWasChanged: function (record) {
-    var typeClass = record.constructor;
+    var typeClass = record.type;
```
fivetanley: this will be a string after my PR for removing passing factories to store type gets merged.

packages/ember-data/lib/system/record-array-manager.js
```diff
@@ -153,7 +153,7 @@ export default Ember.Object.extend({
     for (var i = 0, l = records.length; i < l; i++) {
       record = records[i];
 
-      if (!get(record, 'isDeleted') && !get(record, 'isEmpty')) {
+      if (!record.isDeleted() && !record.isEmpty()) {
```
fivetanley: This will be a breaking change for people currently using .get in their code. One we could probably make an Ember Watson transform for, but still very important to take into consideration and deprecate the getter form.

igorT: It won't because record here is in fact a reference. Anything you touch internally is a reference. I just need to go and rename everything once we have a name.

packages/ember-data/tests/unit/model-test.js
pangratz: Ah, right :+1:

packages/ember-data/lib/system/many-array.js
```diff
@@ -57,19 +58,23 @@ export default Ember.Object.extend(Ember.MutableArray, Ember.Evented, {
...
+      return undefined;
+    }
+    return this.currentState[index].getRecord();
   },
 
   flushCanonical: function() {
     //TODO make this smarter, currently its plenty stupid
     var toSet = filter.call(this.canonicalState, function(record) {
-      return !record.get('isDeleted');
+      return !record.isDeleted();
```
bmac: Lets use `internalModel` here instead of `record` because `record` is already used both internally and externally to refer to an instance of a `DS.Model`.

packages/ember-data/lib/system/model/internal-model.js
```diff
@@ -0,0 +1,696 @@
...
+  save: function() {
+    var promiseLabel = "DS: Model#save " + this;
+    var resolver = Ember.RSVP.defer(promiseLabel);
+
+    this.store.scheduleSave(this, resolver);
+    return resolver.promise;
+  },
+
+  startedReloading: function() {
+    this.isReloading = true;
```
bmac: Should `this.isReloading` be set in the initializer to make it easy to see all of the `InternalModel` state at a glance?

packages/ember-data/lib/system/model/internal-model.js
```diff
@@ -0,0 +1,696 @@
...
+    var record = this;
+    var promiseLabel = "DS: Model#reload of " + this;
+    return new Promise(function(resolve) {
+      record.send('reloadRecord', resolve);
+    }, promiseLabel).then(function() {
+      record.didCleanError();
+      return record;
+    }, function(reason) {
+      record.didError();
+      throw reason;
```
bmac: We have a rejection handler here but it looks like this promise can never reject since we only pass around the resolve function.

packages/ember-data/lib/system/record-arrays/adapter-populated-record-array.js
```diff
@@ -42,13 +42,15 @@ export default RecordArray.extend({
     var records = store.pushMany(type, data);
     var meta = store.metadataFor(type);
 
+    //TODO Optimize
+    var refs = Ember.A(records).mapBy('_internalModel');
```
tchak: `var internalModels`

packages/ember-data/lib/system/model/internal-model.js
```diff
@@ -0,0 +1,696 @@
...
+  );
+}
+
+function retrieveFromCurrentState(key) {
+  return function() {
+    return get(this.currentState, key);
+  };
+}
+
+/*
```
bmac: Can we make this a real doc block starting with `/**` and having a `@class InternalModel` otherwise yuidoc will try to stick the `@method` docblocks below on the wrong class.

packages/ember-data/lib/system/relationships/state/belongs-to.js
```diff
@@ -50,7 +50,7 @@ BelongsToRelationship.prototype._super$flushCanonical = Relationship.prototype.f
 BelongsToRelationship.prototype.flushCanonical = function() {
   //temporary fix to not remove newly created records if server returned null.
   //TODO remove once we have proper diffing
-  if (this.inverseRecord && this.inverseRecord.get('isNew') && !this.canonicalState) {
+  if (this.inverseRecord && this.inverseRecord.isNew() && !this.canonicalState) {
```
tchak: `inverRecord` here is an `internalModel`?

packages/ember-data/lib/system/snapshot.js
```diff
@@ -11,21 +11,23 @@ var get = Ember.get;
...
+  this.record = record;
   record.eachAttribute(function(keyName) {
     this._attributes[keyName] = get(record, keyName);
   }, this);
 
-  this.id = get(record, 'id');
-  this.record = record;
-  this.type = record.constructor;
-  this.modelName = record.constructor.modelName;
+  this.id = get(internalModel, 'id');
```
tchak: don't need `get`

packages/ember-data/lib/system/store.js
```diff
@@ -611,14 +646,14 @@ Store = Service.extend({
       record._preloadData(preload);
     }
 
-    if (get(record, 'isEmpty')) {
+    if (record.isEmpty()) {
```
tchak: record => internalModel

packages/ember-data/lib/system/model/internal-model.js
```diff
@@ -0,0 +1,696 @@
...
+  a performance optimization.
+
+  `InternalModel` should never be exposed to application code. At the boundaries of the system, in places
+  like `find`, `push`, etc. we convert between Models and InternalModels.
+
+  We need to make sure that the properties from `InternalModel` are correctly exposed/proxied on `Model`
+  if they are needed.
+*/
+
+var InternalModel = function(type, id, store, container, data) {
```
stefanpenner: `function InternalModel` so that cross platform we get good naming on stacks and debugging.

packages/ember-data/lib/system/model/internal-model.js
```diff
@@ -0,0 +1,696 @@
...
+  this.type = type;
+  this.id = id;
+  this.store = store;
+  this.container = container;
+  this._data = data || {};
+  this.modelName = type.modelName;
+  this.errors = null;
+  this.dataHasInitialized = false;
+  this._deferredTriggers = [];
```
stefanpenner: in theory, if any of:

* this._deferredTriggers
* this._attributes = Ember.create(null);
* this._inFlightAttributes = Ember.create(null);
* this._relationships = {};

are not always needed on init, making them lazy would increase the alloc/s of this object. note, only do this if they are in-fact not always needed on init.

packages/ember-data/lib/system/model/internal-model.js
```diff
@@ -0,0 +1,696 @@
...
+  this._data = {};
+  this._attributes = Ember.create(null);
+  this._inFlightAttributes = Ember.create(null);
+  this._relationships = {};
+  this.currentState = RootState.empty;
+  /*
```
stefanpenner: kinda strange to have a comment block in the middle of a constructor? Maybe split the algorithim out into a method, and document it?

packages/ember-data/lib/system/model/internal-model.js
stefanpenner: yes, and to keep the same shape

packages/ember-data/lib/system/model/internal-model.js
```diff
@@ -0,0 +1,696 @@
...
+      })
+    ```
+
+    would have a implicit post relationship in order to be do things like remove ourselves from the post
+    when we are deleted
+  */
+  this._implicitRelationships = Ember.create(null);
+  var model = this;
+  //TODO Move into a getter for better perf
+  this.eachRelationship(function(key, descriptor) {
```
stefanpenner: this should likely be a `this._initializerRelationships()`

packages/ember-data/lib/system/model/internal-model.js
```diff
@@ -0,0 +1,696 @@
...
+  isDirty: retrieveFromCurrentState('isDirty'),
+  isSaving: retrieveFromCurrentState('isSaving'),
+  isDeleted: retrieveFromCurrentState('isDeleted'),
+  isNew: retrieveFromCurrentState('isNew'),
+  isValid: retrieveFromCurrentState('isValid'),
+  dirtyType: retrieveFromCurrentState('dirtyType'),
+
+  constructor: InternalModel,
+  materializeRecord: function() {
+    // lookupFactory should really return an object that creates
```
stefanpenner: should we guard + throw if a re-materialization occurs? That seems like an error we would like to fail fast on.

packages/ember-data/lib/system/model/internal-model.js
```diff
@@ -0,0 +1,696 @@
...
+
+  finishedReloading: function() {
+    this.isReloading = false;
+    if (this.record) {
+      set(this.record, 'isReloading', false);
+    }
+  },
+
+  reload: function() {
+    this.startedReloading();
```
stefanpenner: we likely don't want 2 concurrent reloads to happen on the same record, as they will result in hard to reason about plan interference.

packages/ember-data/lib/system/model/internal-model.js
```diff
@@ -0,0 +1,696 @@
...
+    var promiseLabel = "DS: Model#reload of " + this;
+    return new Promise(function(resolve) {
+      record.send('reloadRecord', resolve);
+    }, promiseLabel).then(function() {
+      record.didCleanError();
+      return record;
+    }, function(reason) {
+      record.didError();
+      throw reason;
+    }, "DS: Model#reload complete, update flags")['finally'](function () {
```
stefanpenner: if we use babel or es3-safe, we don't need to make finally a ES computed property

tchak:
@igorT there is lot's of places where `record` should be `internalModel`. It seems important to make the naming consistent.

packages/ember-data/lib/system/model/internal-model.js
```diff
@@ -0,0 +1,696 @@
...
+    var changedKeys = mergeAndReturnChangedKeys(this._data, data);
+    this.pushedData();
+    if (this.record) {
+      this.record._notifyProperties(changedKeys);
+    }
+    this.didInitalizeData();
+  },
+
+  becameReady: function() {
+    var self = this;
```
stefanpenner: Self is a real thing in the bowser / webWorkers. we should likely not shadow it. Instead lets call this `internalModel`

Or even better, use the non-closure retaining form of schedule

```
Ember.run.schedule('actions', this, this.store.recordArrayManager.recordWasLoaded, this)
```

packages/ember-data/lib/system/model/internal-model.js
```diff
@@ -0,0 +1,696 @@
...
+    if (relationshipMeta.kind === 'hasMany') {
+      this._preloadHasMany(key, preloadValue, type);
+    } else {
+      this._preloadBelongsTo(key, preloadValue, type);
+    }
+  },
+
+  _preloadHasMany: function(key, preloadValue, type) {
+    Ember.assert("You need to pass in an array to set a hasMany property on a record", Ember.isArray(preloadValue));
+    var record = this;
```
bmac: `record` -> `internalModel`

packages/ember-data/lib/system/model/internal-model.js
```diff
@@ -0,0 +1,696 @@
...
+
+  _preloadBelongsTo: function(key, preloadValue, type) {
+    var recordToSet = this._convertStringOrNumberIntoInternalModel(preloadValue, type);
+
+    //We use the pathway of setting the hasMany as if it came from the adapter
+    //because the user told us that they know this relationships exists already
+    this._relationships[key].setRecord(recordToSet);
+  },
+
+  _convertStringOrNumberIntoInternalModel: function(value, type) {
```
bmac: `_convertIdIntoInternalModel`? `_convertRecordOrIdIntoInternalModel`?

packages/ember-data/lib/system/model/internal-model.js
```diff
@@ -0,0 +1,696 @@
...
+      if (state.exit) { state.exit(this); }
+      state = state.parentState;
+    } while (!state.hasOwnProperty(pivotName));
+
+    var path = splitOnDot(name);
+    var setups = [];
+    var enters = [];
+    var i, l;
+
+    for (i=0, l=path.length; i<l; i++) {
```
stefanpenner: might be interesting to split this method into several smaller ones, especially ones the isolate the tight loops.

Now that `this` is stable, we would benefit from betting jitting, assuming the function size isn't to large.

Clearly this should actually be the result of some testing, as right now it is just a hunch.

packages/ember-data/lib/system/model/internal-model.js
```diff
@@ -0,0 +1,696 @@
...
+    var args = new Array(length);
+
+    for (var i = 0; i < length; i++) {
+      args[i] = arguments[i];
+    }
+
+    if (this._deferredTriggers.push(args) !== 1) {
+      return;
+    }
+    Ember.run.schedule('actions', this, '_triggerDeferredTriggers');
```
stefanpenner: seems like something we can use scheduleOnce for.

packages/ember-data/lib/system/model/internal-model.js
```diff
@@ -0,0 +1,696 @@
...
+    this.eachRelationship(function(name, relationship) {
+      var rel = this._relationships[name];
+      if (rel) {
+        //TODO(Igor) figure out whether we want to clear or disconnect
+        rel.clear();
+        rel.destroy();
+      }
+    }, this);
+    var model = this;
+    forEach.call(Ember.keys(this._implicitRelationships), function(key) {
```
stefanpenner: call isn't needed if imported from EnumerableUtil

packages/ember-data/lib/system/model/internal-model.js
```diff
@@ -0,0 +1,696 @@
...
+      model._implicitRelationships[key].destroy();
+    });
+  },
+
+  disconnectRelationships: function() {
+    this.eachRelationship(function(name, relationship) {
+      this._relationships[name].disconnect();
+    }, this);
+    var model = this;
+    forEach.call(Ember.keys(this._implicitRelationships), function(key) {
```
stefanpenner: ^

packages/ember-data/lib/system/model/internal-model.js
```diff
@@ -0,0 +1,696 @@
...
+      model._implicitRelationships[key].disconnect();
+    });
+  },
+
+  reconnectRelationships: function() {
+    this.eachRelationship(function(name, relationship) {
+      this._relationships[name].reconnect();
+    }, this);
+    var model = this;
+    forEach.call(Ember.keys(this._implicitRelationships), function(key) {
```
stefanpenner: ^

packages/ember-data/lib/system/model/internal-model.js
```diff
@@ -0,0 +1,696 @@
...
+  _preloadData: function(preload) {
+    var record = this;
+    //TODO(Igor) consider the polymorphic case
+    forEach.call(Ember.keys(preload), function(key) {
+      var preloadValue = get(preload, key);
+      var relationshipMeta = record.type.metaForProperty(key);
+      if (relationshipMeta.isRelationship) {
+        record._preloadRelationship(key, preloadValue);
+      } else {
+        get(record, '_data')[key] = preloadValue;
```
stefanpenner: isn't _data just a normal property ?

packages/ember-data/lib/system/model/internal-model.js
```diff
@@ -0,0 +1,696 @@
...
+  _preloadBelongsTo: function(key, preloadValue, type) {
+    var recordToSet = this._convertStringOrNumberIntoInternalModel(preloadValue, type);
+
+    //We use the pathway of setting the hasMany as if it came from the adapter
+    //because the user told us that they know this relationships exists already
+    this._relationships[key].setRecord(recordToSet);
+  },
+
+  _convertStringOrNumberIntoInternalModel: function(value, type) {
+    if (Ember.typeOf(value) === 'string' || Ember.typeOf(value) === 'number') {
```
stefanpenner: just use `typeof` for strings and numbers, the naked form is correct.

packages/ember-data/lib/system/model/internal-model.js
```diff
@@ -0,0 +1,696 @@
...
+    @method updateRecordArrays
+    @private
+  */
+  updateRecordArrays: function() {
+    this._updatingRecordArraysLater = false;
+    this.store.dataWasUpdated(this.type, this);
+  },
+
+  setId: function(id) {
+    this.id = id;
```
stefanpenner: seems like it might be risky to keep this state in two places. Although that isn't true, if ID is now immutable.

packages/ember-data/lib/system/model/internal-model.js
```diff
@@ -0,0 +1,696 @@
...
+  },
+  // FOR USE DURING COMMIT PROCESS
+
+  /**
+    @method adapterDidInvalidate
+    @private
+  */
+  adapterDidInvalidate: function(errors) {
+    var recordErrors = this.getErrors();
+    for (var key in errors) {
```
stefanpenner: Ember.keys ?

packages/ember-data/tests/integration/relationships/belongs-to-test.js
```diff
@@ -602,7 +602,7 @@ test("belongsTo hasData async loaded", function () {
 
   run(function() {
     store.find('book', 1).then(function(book) {
-      var relationship = book._relationships['author'];
+      var relationship = book._internalModel._relationships['author'];
```
stefanpenner: test reaching two levels into private API seems dubious, at the very least this should be factored into a test helper.

packages/ember-data/lib/system/store.js
```diff
@@ -86,6 +88,20 @@ if (!Backburner.prototype.join) {
 }
 
 
+//Get the materialized model from the internalModel/promise that returns
+//an internal model and return it in a promiseObject. Useful for returning
+//from find methods
+function promiseRecord(internalModel, label) {
```
stefanpenner: this method seems dubious :P

packages/ember-data/lib/system/relationships/belongs-to.js
```diff
@@ -89,19 +89,21 @@ function belongsTo(modelName, options) {
 
   return computedPolyfill({
     get: function(key) {
-      return this._relationships[key].getRecord();
+      return this._internalModel._relationships[key].getRecord();
```
stefanpenner: this pattern is also used in the tests, It seems pretty brittle and looks like it would anger the law of demeter

packages/ember-data/lib/system/store.js
igorT: It's a wrapper between the internal world and the outside world so not sure how to make it nicer. But i don't like it

packages/ember-data/lib/system/model/model.js
```diff
@@ -1091,64 +691,14 @@ var Model = Ember.Object.extend(Ember.Evented, {
...
-      throw reason;
-    }, "DS: Model#reload complete, update flags")['finally'](function () {
-      record.set('isReloading', false);
-      record.updateRecordArrays();
-    });
-
+    var model = this;
     return PromiseObject.create({
-      promise: promise
+      promise: this._internalModel.reload().then(function() {
```
stefanpenner: what is reloads normal fulfillment value, seems like `model` would be a good default ?

packages/ember-data/lib/system/relationships/belongs-to.js
igorT: As soon as I merge this, the lazy relationships refactor is next, there is a pr for it, it actually makes sense now that models are lazy, so that will make this be nicer

packages/ember-data/lib/system/model/model.js
```diff
@@ -1054,13 +656,11 @@ var Model = Ember.Object.extend(Ember.Evented, {
...
   save: function() {
-    var promiseLabel = "DS: Model#save " + this;
-    var resolver = Ember.RSVP.defer(promiseLabel);
-
-    this.store.scheduleSave(this, resolver);
-
+    var model = this;
     return PromiseObject.create({
-      promise: resolver.promise
+      promise: this._internalModel.save().then(function() {
```
stefanpenner: seems like save fulfilling with model is a good default, any reason why internalModel.save does not?

stefanpenner:
I provided some superficial feedback. It is a bit tricky to provide something more in-depth for such a large set of changes. I will try to do another pass in the morning.

igorT:
@stefanpenner the annoying part is that the large part is just moving methods from model to internalModel but not sure how to make it easier to review. You did review some of old code and now we can improve it ;)

packages/ember-data/lib/system/model/model.js
igorT: if you call it inside ED you want to be dealing with internalModels, and not materialized versions, and only expose materialized ones at the boundaries. `save` is kinda special are you guaranteed to have the materializer record, but better to be consistent.

stefanpenner: seems reasonable

packages/ember-data/lib/system/model/model.js
igorT: method's on `internalModel` fulfill with `internalModel`. There are cases where you can reload something without materializing(through relationships for example)

packages/ember-data/lib/system/model/internal-model.js
igorT: Yes, see TODO, also there is a separate PR for that

packages/ember-data/lib/system/model/internal-model.js
igorT: good point, can you open a separate issue for that, i just copied the method around

packages/ember-data/lib/system/store.js
igorT: nice catch was a leftover method

packages/ember-data/lib/system/model/internal-model.js
igorT: :+1:

packages/ember-data/lib/system/model/internal-model.js
igorT: seems good, will file a separate issue about it.

packages/ember-data/lib/system/model/internal-model.js
igorT: not sure what you mean here

packages/ember-data/lib/system/model/internal-model.js
igorT: Will create an issue as this is just copied code

packages/ember-data/lib/system/model/internal-model.js
igorT: `Ember.run.schedule('actions', this.store.recordArrayManager, this.store.recordArrayManager.recordWasLoaded, this)`

packages/ember-data/lib/system/model/internal-model.js
igorT: yes, could convert to a getter in model now that ie8 is going away

packages/ember-data/lib/system/model/internal-model.js
igorT: done

packages/ember-data/lib/system/relationships/state/belongs-to.js
igorT: yes, will need to cleanup later

packages/ember-data/lib/system/model/internal-model.js
igorT: will profile later to measure win vs code clarity

packages/ember-data/lib/system/model/internal-model.js
igorT: https://github.com/emberjs/data/issues/3142

packages/ember-data/lib/system/model/internal-model.js
igorT: :+1:

packages/ember-data/lib/system/model/internal-model.js
igorT: :+1:

packages/ember-data/lib/system/model/states.js
igorT: @tchak is gonna work on this tomorrow

PR #3095, opened 2015-05-24
Minor fixes for references

pangratz:


packages/ember-data/lib/system/reference.js
```diff
@@ -117,7 +117,10 @@ Reference.prototype = {
     this._attributes = Ember.create(null);
 
     return PromiseObject.create({
-      promise: resolver.promise
+      promise: resolver.promise.then(() => {
```
pangratz: :raised_hands: thanks @fivetanley for #3092 :raised_hands:

packages/ember-data/lib/system/model/model.js
```diff
@@ -683,10 +683,7 @@ var Model = Ember.Object.extend(Ember.Evented, {
     successfully or rejected if the adapter returns with an error.
   */
   save: function() {
-    var model = this;
-    return this.reference.save().then(function() {
-      return model;
-    });
+    return this.reference.save();
```
igorT: reload wraps in the record file, so if we want to be consistent, maybe makes more sense to wrap here? Reference deals with internals, and model.js deals with exposing itself to the world?

pangratz: Sounds very reasonable. This also means bye bye PromiseObject from `system/reference.js` :+1:

igorT:
Thanks!

packages/ember-data/lib/system/reference.js
```diff
@@ -194,7 +191,7 @@ Reference.prototype = {
     });
   },
 
-  didIinitalizeData: function() {
+  didInitalizeData: function() {
```
sly7-7: seems like a typo is still here

pangratz:
lol, fixing typos, how do you even work? Good catch @sly7-7 :disappointed:

packages/ember-data/lib/system/reference.js
pangratz: #3107 thanks again :wink:

PR #3100, opened 2015-05-24
Ember.copy() resolved value for ajaxResponse

wecc:
This prevents the original hash provided to ajaxResponse to be modified between multiple adapter runs for the same payload.

### DIFF
--- a/packages/ember-data/tests/integration/store-test.js
+++ b/packages/ember-data/tests/integration/store-test.js
@@ -192,7 +192,7 @@ function ajaxResponse(value) {
     passedVerb = verb;
     passedHash = hash;
 
-    return Ember.RSVP.resolve(value);
+    return run(Ember.RSVP, 'resolve', Ember.copy(value, true));
   };
 }
 
--- a/packages/ember-data/tests/integration/adapter/build-url-mixin-test.js
+++ b/packages/ember-data/tests/integration/adapter/build-url-mixin-test.js
@@ -36,7 +36,7 @@ function ajaxResponse(value) {
   adapter.ajax = function(url, verb, hash) {
     passedUrl = url;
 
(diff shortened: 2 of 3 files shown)


packages/ember-data/tests/integration/store-test.js
```diff
@@ -192,7 +192,7 @@ function ajaxResponse(value) {
     passedVerb = verb;
     passedHash = hash;
 
-    return Ember.RSVP.resolve(value);
+    return run(Ember.RSVP, 'resolve', Ember.copy(value, true));
```
igorT: why the run?

wecc: Just for consistency with the other `ajaxResponse()` methods

fivetanley:
Seems like we might want some kind of perf numbers for this for a fairly large number of records. The way I see it we have a few options:

1) deep copy (this pr)
2) shallow copy (Ember.copy without passing true as last arg)
3) leave it the way it is (:( :( :()

wecc:
This is my extremely non-scientific benchmark running `npm test` on current master vs this (freshly rebased) branch:

**master**
```
886 tests complete (8108 ms)
886 tests complete (7929 ms)
886 tests complete (7912 ms)
886 tests complete (7867 ms)
```

**ajaxResponse-copy-value**

```
886 tests complete (7950 ms)
886 tests complete (7967 ms)
886 tests complete (8009 ms)
886 tests complete (7928 ms)
```

The reason for this PR is because my serializer refactor work caused one of the tests to fail when converting the payload between ED format and JSON-API format during serialization. The store mutates the payload and replaces loaded relationships with the model classes and one test reused the same payload for two different responses, making the payload in the second respons contain the records instead of just the IDs. Therefor I don't think shallow copying would work.

PR #3110, opened 2015-05-26
Cleanup unused 'materialize' records in model

tonywok:
I came across three functions and was wondering what they were for. On closer inspection, they seem to be unused. Perhaps left over from pre-modern adapter/serializer refactors?


fivetanley:
Nice catch! Can we deprecate these instead of removing? We can remove them soon as the 1.0 rcs approach.

igorT:
No reason to deprecate, they were internal. Thanks!

PR #3099, opened 2015-05-24
[FEATURE ds-improved-ajax] Finer control over customizing a request

pangratz:
Though `ajax()` (and `ajaxOptions()`) of `DS.RESTAdapter` are marked as
private, they have been overwritten in many applications, since there is
currently no other way to customize the request.

This feature adds new public methods to `DS.RESTAdapter`, which allow to
customize the properties of a request:

- `methodForRequest` to get the HTTP verb
- `urlForRequest` to get the URL
- `headersForRequest` to get the headers
- `dataForRequest` to get the data (query params or request body)

The `params` hash passed to those methods has all the properties with
which the corresponding `find`, `createRecord`, `findQuery`, ...  call
have been invoked: store, type, snapshot(s), id(s) and query. The
`requestType` property indicates which method is requested; the possible
values are:

- `createRecord`
- `updateRecord`
- `deleteRecord`
- `query`
- `queryRecord`
- `findRecord`
- `findAll`
- `findMany`
- `findHasMany`
- `findBelongsTo`

Performing the actual AJAX request is handled by the `makeRequest`
method, which is similar to the existing `ajax` method: it makes the
request using `jQuery.ajax` and attaches success and failure handlers.

---

Say your API handles creation of resources via PUT, this can now be
customized as follows:

``` js
// adapters/application.js
import DS from 'ember-data';

export default DS.RESTAdapter.extend({
  methodForRequest: function(params) {
    if (params.requestType === 'createRecord') {
      return "PUT";
    }

    return this._super(...arguments);
  }
});
```

---

Credit for the initial idea goes to @igorT in https://github.com/emberjs/data/issues/3047#issuecomment-99989094.

PR #3119, opened 2015-05-28
maxUrlLength -> maxURLLength

### LINKED ISSUE #3101: Rest Adapter's maxUrlLength is not properly cased
It should be maxURLLength

sly7-7:
Fixes #3101
@igorT Let me know if I can remove the deprecation warning (it seems to me that people may override groupsForHasMany and use this property, like in the test)

@pangratz When this is merged, you will have to rebase your PR #3099 :sweat_smile:

### DIFF
--- a/packages/ember-data/lib/adapters/rest-adapter.js
+++ b/packages/ember-data/lib/adapters/rest-adapter.js
@@ -10,6 +10,7 @@ import {
   MapWithDefault
 } from "ember-data/system/map";
 var get = Ember.get;
+var set = Ember.set;
 var forEach = Ember.ArrayPolyfills.forEach;
 
 import BuildURLMixin from "ember-data/adapters/build-url-mixin";
@@ -170,7 +171,7 @@ import BuildURLMixin from "ember-data/adapters/build-url-mixin";
   @extends DS.Adapter
   @uses DS.BuildURLMixin
 */
-export default Adapter.extend(BuildURLMixin, {
+var RestAdapter = Adapter.extend(BuildURLMixin, {
   defaultSerializer: '-rest',
 
   /**
@@ -598,7 +599,7 @@ export default Adapter.extend(BuildURLMixin, {
   },
 
   // http://stackoverflow.com/questions/417142/what-is-the-maximum-length-of-a-url-in-different-browsers
-  maxUrlLength: 2048,
+  maxURLLength: 2048,
 
   /**
     Organize records into groups, each of which is to be passed to separate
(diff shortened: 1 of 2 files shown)


pangratz:
@sly7-7 Thanks for the heads up!

packages/ember-data/lib/adapters/rest-adapter.js
```diff
@@ -598,7 +598,12 @@ export default Adapter.extend(BuildURLMixin, {
   },
 
   // http://stackoverflow.com/questions/417142/what-is-the-maximum-length-of-a-url-in-different-browsers
-  maxUrlLength: 2048,
+  maxURLLength: 2048,
+
+  maxUrlLength: Ember.computed(function() {
```
igorT: Shouldn't turn it into a computed if even we weren't calling get on it. You can deprecate using deprecateProperty/making a prop getter similar to https://github.com/emberjs/data/blob/master/packages/ember-data/lib/system/snapshot.js#L39

sly7-7:
@igorT Done, AppVeyor is angry with npm though...

packages/ember-data/lib/adapters/rest-adapter.js
```diff
@@ -838,3 +838,13 @@ function endsWith(string, suffix) {
     return string.endsWith(suffix);
   }
 }
+
+Ember.defineProperty(RESTAdapter, 'maxUrlLength', {
```
igorT: Might have needed something more for IE8, but I think we are gonna drop it today

sly7-7: Sorry, I missed that. What could we do for IE8 ? I can at least put  this defineProperty inside a  Ember.platform.hasPropertyAccessors block, but how could I do it for IE8 ? (Am I missing something or Ember.deprecateProperty is just a noop on IE8 ?)

igorT: We are dropping IE8 so no worries

igorT: Also wouldn't this break if someone subclassed the adapter to provide their own limit as they would override this getter? cc @rwjblue whats the proper way of doing this

bmac: Here is the code Ember uses to deprecate a property. https://github.com/emberjs/ember.js/blob/1b5adaa65d20af4f4e57702ea34aa0dcc42e1489/packages/ember-metal/lib/deprecate_property.js#L24-L43. It seems to work even if a user overrides this getter. Its also usually called on the classes's `prototype` object.

sly7-7: @bmac I saw this too, so that basically mean that people developping in a world without property accessors, they just don't have the deprecation warning, right ?

bmac: Correct

sly7-7: Ok, so @igorT Do you want me to wrap that the same way Ember does ?

bmac: @sly7-7 yes lets do it the way Ember does it.

sly7-7: @bmac Ok, Will do this in couple hours

sly7-7:
@bmac Should be good

packages/ember-data/lib/adapters/rest-adapter.js
```diff
@@ -838,3 +838,15 @@ function endsWith(string, suffix) {
     return string.endsWith(suffix);
   }
 }
+
+if (Ember.platform.hasPropertyAccessors) {
+  Ember.defineProperty(RestAdapter, 'maxUrlLength', {
```
bmac: This should be defined on `RestAdapter.prototype`.

packages/ember-data/lib/adapters/rest-adapter.js
```diff
@@ -838,3 +838,15 @@ function endsWith(string, suffix) {
...
+    enumerable: false,
+    get: function() {
+      Ember.deprecate('maxUrlLength has been deprecated (wrong casing). You should use maxURLLength instead.');
+      return this.maxURLLength;
+    }
```
bmac: Lets also add a setter otherwise a type error will get thrown if a user tires to extend an adapter using `maxUrlLength`.

sly7-7: Ok, I added it at first, but thought it was a kind of read only property as this seems to me it was a static integer.

bmac:
Tested it locally and everything looks good. :+1: 
screen shot 2015-06-04 at 3 36 36 pm

PR #3120, opened 2015-05-28
Update tests to always pass strings to store methods

wecc:


pangratz:
Here are [some](http://git.io/vkWsE) [more](http://git.io/vkWsK) in [unit/adapters/rest-adapter/ajax.js](http://git.io/vkWsb) ... :gift:

PR #3121, opened 2015-05-28
Fix docparser warnings, @returns => @return

wecc:
```
warn: (docparser): replacing incorrect tag: returns with return:  packages/activemodel-adapter/lib/system/active-model-serializer.js:140
warn: (docparser): replacing incorrect tag: returns with return:  packages/ember-data/lib/serializers/rest-serializer.js:746
warn: (docparser): replacing incorrect tag: returns with return:  packages/ember-data/lib/serializers/rest-serializer.js:793
```


igorT:
Thanks!

Issue #1834, opened 2014-03-24
[Bug] Sideloading relations (hasMany) on models not working!


steffenbrem:
Hi,

**NOTE: I am using Ember App Kit**

Following the docs at emberjs.com you can sideload hasMany relations with one request. For me this is not working, I've tried almost every type of response and I keep getting the following error in Ember:

<code>Uncaught Error: Assertion Failed: You looked up the 'users' relationship on
'<(subclass of DS.Model):ember605:1>' but some of the associated records were not
loaded. Either make sure they are all loaded together with the parent record, or specify
t...<omitted>...`)</code>

This is the code I use to load the data:

```
// routes/customers.js
var customers = this.store.find('customer');
controller.set("customers", customers);
```

```
// templates/customers.hbs
{{#each customers}}
    <!-- this is working fine -->
    <p>{{firstName}}</p>

    <!-- Bad things happen! -->
    <p>Users:</p>
    {{#each users}}
        <p>Why am I getting errors here?</p>
    {{/each}}
{{/each}}
```

1) Scalar fields are working and I can print them out for every customer that is returned.
2) When I iterate over a relation, I receive errors.

This is the response Ember Data receives:

```
{
    "users": [
        {
            "id": 1,
            "username": "Steffen",
            "email": "email1@test.com"
        },
        {
            "id": 2,
            "username": "Krantenwijk",
            "email": "email2@test.com"
        },
        {
            "id": 3,
            "username": "Tan",
            "email": "email3@test.com"
        }
    ],
    "customers": [
        {
            "id": 1,
            "company": "Acme BV",
            "firstName": "Steffen",
            "lastName": "Brem",
            "lastModified": "2014-03-21T12:00:51+0100"
            "users": [
                1,
                2,
                3
            ]
        }
    ]
}
```

Is this a bug?

Steffen

igorT:
@sly7-7 I am pretty sure we support both links and ids at the same time. Can you verify?

sly7-7:
@igorT Sure, I will do that in the next couple of days

sly7-7:
@igorT I've just added a passing test about that (here)[https://github.com/sly7-7/data/commit/f8624bf4acdb83797582d21e6d709b69a05c6017], so it seems like the error is gone, but I don't know if this behavior is expected or not. (The link takes precedence over the ids)

steffenbrem:
[...]

PR #3122, opened 2015-05-29
Ensure getting async belongsTo relationship does not fetch link if it is already loaded

csantero:
We were experiencing an infinite loop of AJAX requests when getting a belongsTo relationship when the inverse record was already in the store. We changed `findLink` to return the inverse record if it exists, to match the behavior of [findRecord](https://github.com/emberjs/data/blob/master/packages/ember-data/lib/system/relationships/state/belongs-to.js#L105)

Paired with @raycohen

### DIFF
--- a/packages/ember-data/lib/system/relationships/state/belongs-to.js
+++ b/packages/ember-data/lib/system/relationships/state/belongs-to.js
@@ -125,10 +125,14 @@ BelongsToRelationship.prototype.getRecord = function() {
   if (this.isAsync) {
     var promise;
     if (this.link) {
-      var self = this;
-      promise = this.findLink().then(function() {
-        return self.findRecord();
-      });
+      if (this.hasData) {
+        promise = this.findRecord();
+      } else {
+        var self = this;
+        promise = this.findLink().then(function() {
+          return self.findRecord();
+        });
+      }
     } else {
       promise = this.findRecord();
     }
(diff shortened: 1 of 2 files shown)


packages/ember-data/lib/system/relationships/state/belongs-to.js
```diff
@@ -110,6 +110,15 @@ BelongsToRelationship.prototype.findRecord = function() {
   }
 };
 
+BelongsToRelationship.prototype._super$findLink = Relationship.prototype.findLink;
+BelongsToRelationship.prototype.findLink = function() {
```
igorT: You should probably be using https://github.com/csantero/data/blob/dont-fetch/packages/ember-data/lib/system/relationships/state/relationship.js#L18 instead of the find link, because that tells us whether we already loaded data for the relationship. Can you put this on the super class and thus make it work for many relationships as well?

csantero:
@igorT I pushed a new diff that checks `this.hasData` in `getRecord`, and if true, it bypasses `this.findLink()`.

I wasn't sure what you meant about putting something on the super class. I tried adding an equivalent check to `ManyRelationship.prototype.getRecords`, but it caused tests to fail, and I'm not really familiar enough with the issue in #1834 to know how to fix it.

raycohen:
@igorT Are we on the right track with these updates?  I know there's a lot of work behind the scenes on next versions of ember-data.  Is this PR possibly in conflict with that work?

bmac:
@wecc Do you mind taking a look at this pr?

wecc:
I'm closing this PR since it's been superseded by #3762. Thanks for the work though @csantero!

PR #3123, opened 2015-05-29
fix build issues

fivetanley:
This restores ES6 build output for tests. We let the assertion for
Ember.VERSION run in production builds due to a weird bug with
defeautreify where it removes one too many characters in the case of
Ember.runInDebug.

demo of working .prod.js build here: http://emberjs.jsbin.com/litilaxuro/1/edit

working .min.js build here: http://emberjs.jsbin.com/litilaxuro/2/edit


fivetanley:
cc @amiel

amiel:
Thanks!

PR #3124, opened 2015-05-29
refactor store managed instances

### LINKED ISSUE #3050: [Proposal] Resolving an adapter-less model should cache the 'application adapter' resolution.
I'm not sure if this belongs here, in ember-cli, or in ember-resolver.  Currently I have the problem where I load 3 conversations with 20 messages each.  These 63 records all use the default `application` adapter.  Since no explicit adapter is ever found, each one triggers the resolver to seek the adapter at each possible path each time, leading to 3 lookups per record and a minimum of 189 lookups.

This is a lot of excess when the first lookup attempt could have stashed the application resolver.

The bigger issue arises in safari which seems fragile to this and can get stuck in a perpetual loop trying to resolve the adapters.

fivetanley:
In this commit, we refactor store managed instance lookups to have its
own class that takes the store's container. This moves a large portion
of code out of the store and instead uses delegation.

There's also a performance benefit:

If you look up `adapter:post` and fallback to `adapter:application`,
this lookup is cached. So, if you look up `adapter:post` again, that
instance will actually be cached. Previously, we did the lookup for
`adapter:post` every time.

This also removes passing factories as the `adapter` property to
DS.Store.

closes #3050


fivetanley:
cc @runspired

runspired:
:D

PR #3125, opened 2015-05-30
Cleanup unneeded if after  ManyArray refactor

tonywok:
Was familiarizing myself with the record array manager while looking into the memory leak issue with filter. This condition looks like it can be removed now that `ManyArray` isn't a `RecordArray`, and all possible traffic through that method should be derived from `RecordArray` and therefore have a `manager`.


igorT:
Thanks!

PR #3126, opened 2015-05-31
Guard for embedded unknown hasMany relationship

### LINKED ISSUE #3071: Uncaught TypeError: Cannot read property 'map' of undefined
Receiving: `Uncaught TypeError: Cannot read property 'map' of undefined | embedded-records-mixin.js:485` on beta.17 with Ember 1.12.0.

This prevents the `model.save()` request from being sent to the server.

wecc:
Closes #3071

### DIFF
--- a/packages/ember-data/lib/serializers/embedded-records-mixin.js
+++ b/packages/ember-data/lib/serializers/embedded-records-mixin.js
@@ -298,13 +298,17 @@ var EmbeddedRecordsMixin = Ember.Mixin.create({
     }
     var includeIds = this.hasSerializeIdsOption(attr);
     var includeRecords = this.hasSerializeRecordsOption(attr);
-    var key;
+    var key, hasMany;
     if (includeIds) {
       key = this.keyForRelationship(attr, relationship.kind, 'serialize');
       json[key] = snapshot.hasMany(attr, { ids: true });
     } else if (includeRecords) {
       key = this.keyForAttribute(attr, 'serialize');
-      json[key] = snapshot.hasMany(attr).map(function(embeddedSnapshot) {
+      hasMany = snapshot.hasMany(attr);
+
+      Ember.warn("The embedded relationship '" + key + "' is undefined for '" + snapshot.modelName + "' with id '" + snapshot.id + "'. Please include it in your original payload.", Ember.typeOf(hasMany) !== 'undefined');
+
+      json[key] = Ember.A(hasMany).map(function(embeddedSnapshot) {
         var embeddedJson = embeddedSnapshot.record.serialize({ includeId: true });
         this.removeEmbeddedForeignKey(snapshot, embeddedSnapshot, relationship, embeddedJson);
         return embeddedJson;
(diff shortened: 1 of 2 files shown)


packages/ember-data/tests/integration/serializers/embedded-records-mixin-test.js
```diff
@@ -609,6 +609,31 @@ test("serialize with embedded objects (hasMany relationship)", function() {
...
+  var serializer, json;
+  run(function() {
+    serializer = env.container.lookup("serializer:home-planet");
+
+    json = serializer.serialize(league._createSnapshot());
+  });
+
+  deepEqual(json, {
+    name: "Villain League",
+    villains: []
```
wecc: Is this really what we want? (since the relationship is "unknown")

Excluding unknown relationships from the payload might not be correct when it comes to embedded since it _should_ be present, correct?

ping @igorT @jakesjews

jakesjews: I do agree that an embedded relationship should always be present but a developer might not necessarily be in control of the API. An alternative could perhaps be an assert so it's easier to debug.

packages/ember-data/lib/serializers/embedded-records-mixin.js
```diff
@@ -304,8 +304,9 @@ var EmbeddedRecordsMixin = Ember.Mixin.create({
       key = this.keyForRelationship(attr, relationship.kind, 'serialize');
       json[key] = snapshot.hasMany(attr, { ids: true });
     } else if (includeRecords) {
+
       key = this.keyForAttribute(attr, 'serialize');
-      json[key] = snapshot.hasMany(attr).map(function(embeddedSnapshot) {
+      json[key] = Ember.A(snapshot.hasMany(attr)).map(function(embeddedSnapshot) {
```
igorT: Can you add a comment as to why we do this, also maybe a warning?

wecc:
@igorT @jakesjews updated with a warning, still embedding it as an empty array.

packages/ember-data/lib/serializers/embedded-records-mixin.js
```diff
@@ -299,13 +299,17 @@ var EmbeddedRecordsMixin = Ember.Mixin.create({
...
+    var key, hasMany;
     if (includeIds) {
       key = this.keyForRelationship(attr, relationship.kind, 'serialize');
       json[key] = snapshot.hasMany(attr, { ids: true });
     } else if (includeRecords) {
       key = this.keyForAttribute(attr, 'serialize');
-      json[key] = snapshot.hasMany(attr).map(function(embeddedSnapshot) {
+      hasMany = snapshot.hasMany(attr);
+
+      Ember.warn("The relationship '" + key + "' is unknown for '" + snapshot.modelName + "' with id '" + snapshot.id + "'. Please include it in your original payload.", Ember.typeOf(hasMany) !== 'undefined');
```
igorT: should probably mention that it's an embedded relationship, and say undefined instead of unknown

igorT:
Thanks!

packages/ember-data/lib/serializers/embedded-records-mixin.js
wecc: Done!

ChrisHonniball:
Will this be tagged soon? This is the only thing preventing me from updating my application to Ember v1.12.1. Thanks!

esbanarango:
Same here!

bmac:
Yeah, there have been a lot of changes recently. The plan is to get a release out this week.

PR #3127, opened 2015-05-31
rename Reference to Ghost

fivetanley:
Originally we planned on renaming to `modelDataPOJO`. However, the name
is misleading because it does more than just store data; it has its own
constructor and prototype chain.

I chose Ghost from Martin Fowler's 'Patterns of Enterprise Application
Architecture':

> A ghost is the real object in a partial state. When you load the object
> from the database it contains just its ID. Whenever you try to access a
> field it loads its full state. Think of a ghost as an object, where
> every field is lazy-initialized in one fell swoop, or as a virtual
> proxy, where the object is its own virtual proxy. Of course, there's no
> need to load all the data in one go; you may group it in groups that are
> commonly used together.

### DIFF
--- a/packages/ember-data/lib/serializers/embedded-records-mixin.js
+++ b/packages/ember-data/lib/serializers/embedded-records-mixin.js
@@ -116,7 +116,7 @@ var EmbeddedRecordsMixin = Ember.Mixin.create({
    @method normalize
    @param {subclass of DS.Model} typeClass
    @param {Object} hash to be normalized
-   @param {String} key the hash has been referenced by
+   @param {String} key the hash has been ghostd by
    @return {Object} the normalized hash
   **/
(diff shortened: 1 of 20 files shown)


packages/ember-data/lib/serializers/embedded-records-mixin.js
```diff
@@ -116,7 +116,7 @@ var EmbeddedRecordsMixin = Ember.Mixin.create({
    @method normalize
    @param {subclass of DS.Model} typeClass
    @param {Object} hash to be normalized
-   @param {String} key the hash has been referenced by
+   @param {String} key the hash has been ghostd by
```
rwjblue: ghosted

wycats:
What's wrong with "Reference"?

PR #3132, opened 2015-06-01
Remove ember.debug warning from the browser console when running tests

bmac:


igorT:
Don't we want the debug version of release as well, in order to make debugging easier?

igorT:
Nm, I brainfarted

PR #3133, opened 2015-06-01
Create snapshots on save not on flush

tchak:
alows us to have access to changed attributes in snapshot

cc @igorT @wecc

### DIFF
--- a/packages/ember-data/lib/system/store.js
+++ b/packages/ember-data/lib/system/store.js
@@ -1254,8 +1254,9 @@ Store = Service.extend({
     @param {Resolver} resolver
   */
   scheduleSave: function(record, resolver) {
+    var snapshot = record._createSnapshot();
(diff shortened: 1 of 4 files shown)


packages/ember-data/lib/system/store.js
```diff
@@ -2040,7 +2042,7 @@ function _commit(adapter, store, operation, record) {
 
     store._adapterRun(function() {
       if (adapterPayload) {
-        payload = serializer.extract(store, type, adapterPayload, get(record, 'id'), operation);
+        payload = serializer.extract(store, type, adapterPayload, get(snapshot, 'id'), operation);
```
wecc: Since `snapshot` is a POJO we're doing `snapshot.id` most places in the codebase.

igorT:
Can you add a test for the save and attribute change case?

PR #3136, opened 2015-06-01
Lookup the store using store:application instead of store:main

bmac:
This simplifes the process of registering the store when a user
creates a custom one.  This change also fixes deprecations warning
introduces in Ember 1.12. The store service is now injected as a
instanceInitializer.

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -2,7 +2,7 @@
   "name": "ember-data",
   "private": true,
   "dependencies": {
-    "ember": "~1.11.1"
+    "ember": "~1.12.1"
   },
   "devDependencies": {
     "qunit": "~1.17.0",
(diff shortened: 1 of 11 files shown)


igorT:
Can @rwjblue or @stefanpenner review?

rwjblue:
* store:main will still work, with a deprecation in favor of store:application. 

* store:application worked in prior versions


Looks good to me. We should probably update any docs that reference 'app/store.js' in favor of 'app/stores/application.js' (both here and in the guides).

bmac:
@rwjblue I searched through the guides and the API docs and couldn't find any instance of 'app/store.js' (which should have already given a deprecation message). 

`app/stores/application.js` has been the recommend path for a while, however most of the docs that describe how to do it are still using globals so it looks like `App.ApplicationStore`.

PR #3137, opened 2015-06-01
minor adjustments and tests for snapshot creation refactor

tchak:
@igorT requested changes and test

### DIFF
--- a/packages/ember-data/lib/system/store.js
+++ b/packages/ember-data/lib/system/store.js
@@ -1262,12 +1262,13 @@ Store = Service.extend({
 
     @method scheduleSave
     @private
-    @param {DS.Model} record
+    @param {InternalModel} record
     @param {Resolver} resolver
   */
-  scheduleSave: function(record, resolver) {
-    var snapshot = record._createSnapshot();
-    record.adapterWillCommit();
+  scheduleSave: function(internalModel, resolver) {
+    var snapshot = internalModel._createSnapshot();
+    internalModel.flushChangedAttributes();
+    internalModel.adapterWillCommit();
     this._pendingSave.push([snapshot, resolver]);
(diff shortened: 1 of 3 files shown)


tchak:
@igorT I rebased and updated after your refactor

packages/ember-data/lib/system/store.js
```diff
@@ -257,11 +257,11 @@ Store = Service.extend({
 
     @method serialize
     @private
-    @param {DS.Model} record the record to serialize
+    @param {InternalModel} record the record to serialize
```
sly7-7: should we also change the description from record to internalModel ?

igorT: :+1:

igorT: actuallly doesn't take an internalModel

packages/ember-data/tests/unit/model/merge-test.js
```diff
@@ -40,6 +40,43 @@ test("When a record is in flight, changes can be made", function() {
...
 });
 
+test("Make sure snapshot is created at save time not at flush time", function() {
+  expect(5);
+
+  var adapter = DS.Adapter.extend({
+    updateRecord: function(store, type, snapshot) {
+      equal(snapshot.attr('name'), 'Thomas Dale');
+
+      return new Ember.RSVP.Promise(function(resolve, reject) {
```
igorT: you can just return Ember.RSVP.resolve() here

tchak: updated

packages/ember-data/lib/system/store.js
tchak: fixed

Issue #2793, opened 2015-02-16
Embedded polymorphic record with custom id doesn't work.


mayatskiy:
If I have embedded polymorphic records and a custom primaryKey (for example '_id'). 
When I use EmbeddedRecordsMixin this code works incorrectly:

<pre>
function extractEmbeddedHasManyPolymorphic(store, key, hash) {
  ...
  forEach(hash[key], function(data) {
    var typeKey = data.type;
    var embeddedSerializer = store.serializerFor(typeKey);
    var embeddedType = store.modelFor(typeKey);
    var primaryKey = get(embeddedSerializer, 'primaryKey'); // returns '_id'

    var embeddedRecord = embeddedSerializer.normalize(embeddedType, data, null);
    store.push(embeddedType, embeddedRecord);
    ids.push({ id: embeddedRecord[primaryKey], type: typeKey }); // the error is here
  });
</pre>

next line:

<pre>
   ids.push({ id: embeddedRecord[primaryKey], type: typeKey });
</pre>

sets undefined id. Because embeddedRecord has no '_id' key after deserialization and have only 'id' key.

lolmaus:
Hey @mayatskiy, i'm using polymorphic embedded records successfully. Can you please reproduce the issue on http://emberjs.jsbin.com/ ?

mayatskiy:
@lolmaus with custom id?

lolmaus:
@mayatskiy, my polymorphic records are treated by the backend as nested attributes rather than child records. They arrive without ids. 

I override the `extractArray`, `extractSingle` and `serialize` methods of the parent model serializer and slip ids into the polymorphic items.

bmac:
Thanks for the bug report @mayatskiy.  Would you like to open a pr to fix this issue?

igorT:
@pangratz wanna take a stab at this? Shouldn't be too hard, we in fact don't need to worry about the primary key here because we already normalized, so if you just remove those couple lines + a test would be good to go

pangratz:
@igorT I will take stab at it tomorrow

igorT:
Fixed by #3138

PR #3138, opened 2015-06-01
Custom primaryKey for embedded polymorphic relations work

pangratz:
Since the hashs of the polymorphic embedded records are normalized, the - possible - customized primaryKey can be accessed via the `id` property.

---

This addresses #2793.

cc @igorT

### DIFF
--- a/packages/ember-data/tests/integration/serializers/embedded-records-mixin-test.js
+++ b/packages/ember-data/tests/integration/serializers/embedded-records-mixin-test.js
@@ -1165,6 +1165,63 @@ test("extractSingle with polymorphic hasMany", function() {
 
 });
 
+test("extractSingle with polymorphic hasMany and custom primary key", function() {
+  SuperVillain.reopen({
+    secretWeapons: DS.hasMany("secretWeapon", { polymorphic: true })
+  });
+
+  env.registry.register('adapter:super-villain', DS.ActiveModelAdapter);
+  env.registry.register('serializer:light-saber', DS.ActiveModelSerializer.extend({
+    primaryKey: 'custom'
+  }));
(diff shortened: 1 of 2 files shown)


packages/ember-data/tests/integration/serializers/embedded-records-mixin-test.js
```diff
@@ -1165,6 +1165,63 @@ test("extractSingle with polymorphic hasMany", function() {
...
+      last_name: "Dale",
+      secret_weapons: [
+        {
+          custom: "1",
+          type: "LightSaber",
+          name: "Tom's LightSaber",
+          color: "Red"
+        },
+        {
+          id: "1",
```
igorT: shouldn't this be custom?

igorT:
Thanks!

packages/ember-data/tests/integration/serializers/embedded-records-mixin-test.js
pangratz: :hushed: – The `primaryKey` is only defined for the `serializer:light-saber`. If both embedded records should have `custom` as `primaryKey`, it would/could be defined in `serializer:secret-weapon`.

Am I missing something? :thought_balloon: :pensive:

pangratz: I could add a comment since I guess this would make reading the test easier ... c/d?

igorT: NM, I scrwed up while reading

pangratz:
@mayatskiy since you originally reported the issue in #2793, can you check the test cases in this PR and confirm that they are the ones you had/have issues with?

igorT:
Thanks!

pangratz:
:dancers:

mayatskiy:
@pangratz yes, I can.

PR #3139, opened 2015-06-01
Extend adapter instead of reopening it

ZenCocoon:
Update documentation to recommend extended the adapter instead of reopening it.


Turbo87:
shouldn't we use `export default DS.RESTAdapter.extend(...)` instead, now that `ember-cli` is the recommended way of doing things?

rwjblue:
Yes, that makes sense to me.

igorT:
Yeah we need to go through all the docs and update

PR #3143, opened 2015-06-02
Refactor the Serializer API

wecc:
This PR refactors the way serializers normalizes adapter responses. `serializer.extract()` is now deprecated for the new `serializer.normalizeResponse()`.

`normalizeResponse()` is expected to return a JSON-API Document with primary and, if any, secondary/sideloaded/included as `{ data, included }`.

`normalize()` is also expected to return a JSON-API Document having the primary data be the normalized record in `{ data }` and, if any, secondary data (e.g. embedded) in `{ included }`.

This change should be fully backwards compatible. For users to opt-in to the new Serializer API they should convert their existing `extract()` hooks to `normalizeResponse()` hooks and make it return JSON-API. Then add `isNewSerializerAPI: true` in their serializer(s).

## Todos

- [x] Rename `REST2Serializer`, `EmbeddedRecords2Mixin` and `ActiveModel2Serializer` to something that makes sense
- [x] Figure out what and how to deprecate the old API
- [x] Feature flag / set `isNewSerializerAPI` for non-extended serializers
- [x] Create hooks to make it easier to customize `normalizeResponse()`
- [x] Break up `normalizeResponse()` to reflect the old `extractArray/extractSingle` methods
- [x] Add deprecation notices


igorT:
@wecc you forgot to mention JSON2Serializer

sly7-7:
I'm probably missing something, but why not naming those classes as JSONAPIXxx ?

bmac:
I think the fear is that `JSONAPI` in the name will be confusing because the new "2" serializers are not intended to be used with a JSONAPI backend. They simply transform a non JSONAPI data into JSON API format, with the eventual goal of Ember Data using json api natively, allowing 3rd party serializer authors to target JSON API instead of the unspecified "normalized" format that Ember Data uses today.

sly7-7:
@bmac Thanks, that makes sense

packages/ember-data/lib/serializers/rest2-serializer.js
```diff
@@ -0,0 +1,138 @@
...
+  _normalizeSerializerPayloadItem,
+  pushPayload
+} from "ember-data/system/store/serializer-response";
+
+var forEach = Ember.ArrayPolyfills.forEach;
+//var map = Ember.ArrayPolyfills.map;
+
+function coerceId(id) {
+  return id == null ? null : id + '';
+}
```
bmac: Looks like this method exists is repeated in the `rest2-serializer` `rest-serializer` and `store` module. Lets move it into its own file and import it in all 3 places.

packages/ember-data/lib/serializers/rest2-serializer.js
```diff
@@ -0,0 +1,138 @@
...
+
+  normalize: function(typeClass, hash, prop) {
+    return {
+      data: _normalizeSerializerPayloadItem(typeClass, this._super(typeClass, hash, prop)),
+      included: []
+    };
+  },
+
+  normalizeResponse: function(store, primaryTypeClass, payload, id, requestType) {
+    var document = {
```
bmac: Github is hi-lighting `document` in a different color because it assumes its the global document. Maybe we should use a different name instead?

wecc: Yeah, I'm not too fond of `document` but I'd like to have a more distinct name than `payload` or something like that. Any suggestions?

packages/ember-data/lib/serializers/rest2-serializer.js
wecc: Done!

packages/ember-data/lib/serializers/json-serializer.js
```diff
@@ -124,6 +132,101 @@ export default Serializer.extend({
...
+        return this.normalizeFindManyResponse(...arguments);
+      case 'findQuery':
+        return this.normalizeFindQueryResponse(...arguments);
+      case 'createRecord':
+        return this.normalizeCreateRecordResponse(...arguments);
+      case 'deleteRecord':
+        return this.normalizeDeleteRecordResponse(...arguments);
+      case 'updateRecord':
+        return this.normalizeUpdateRecordResponse(...arguments);
+    }
```
wecc: Might want to default to `return this._normalizeResponse(...arguments)`

fivetanley: could also dry this up to `this[`normalize${capitalize(requestType)}`] || this._normalizeResponse)`

wecc: I have this vague memory that we wanted to change this to switch to make it easier for people to see what actual `requestType`s are getting passed. We just never got around to change it here. The discussion was originally around https://github.com/emberjs/data/blob/master/packages/ember-data/lib/adapters/build-url-mixin.js#L52-L75

igorT:
:+1: on structuring the commits

packages/ember-data/lib/serializers/json-serializer.js
```diff
@@ -172,6 +492,134 @@ export default Serializer.extend({
...
+      if (relationshipHash.id) {
+        relationshipHash.id = coerceId(relationshipHash.id);
+      }
+      return relationshipHash;
+    }
+    return { id: coerceId(relationshipHash), type: relationshipModelName };
+  },
+
+  /**
+    Returns the resource's relationships formatted as an JSON-API "relationships object".
```
tomdale: s/an/a/

packages/activemodel-adapter/tests/integration/active-model-serializer-namespaced-modelname-new-test.js
```diff
@@ -0,0 +1,155 @@
+var SuperVillain, EvilMinion, YellowMinion, DoomsdayDevice, MediocreVillain, TestSerializer, env;
```
sly7-7: not sure, but maybe a missing dash in the file name ? `active-model-serializer-namespaced-model*-*name-new-test.js`?

packages/ember-data/lib/system/store/serializer-response.js
```diff
@@ -0,0 +1,236 @@
+var forEach = Ember.EnumerableUtils.forEach;
+var map = Ember.EnumerableUtils.map;
+
+/**
+  This is a helper method that always returns a JSON-API Document.
+
+  If the serializer doesn't have the `normalizeResponse` method `extract` will
```
sly7-7: Maybe a missing comma here `the normalizeResponse method**,**`extract`

packages/ember-data/lib/serializers/embedded-records-mixin.js
```diff
@@ -393,63 +400,124 @@ var EmbeddedRecordsMixin = Ember.Mixin.create({
...
+      return hash;
+    }
+
+    var embeddedSerializer = store.serializerFor(embeddedTypeClass.modelName);
+    var embeddedRecord = embeddedSerializer.normalize(embeddedTypeClass, hash[key], null);
+    store.push(embeddedTypeClass.modelName, embeddedRecord);
 
-  var ids = [];
+    hash[key] = embeddedRecord.id;
+    //TODO Need to add a reference to the parent later so relationship works between both `belongsTo` records
```
igorT: Don't need this todo anymore

igorT:
:+1:

PR #3145, opened 2015-06-02
Move coerceId to a separate file

wecc:


bmac:
:+1:

packages/ember-data/lib/system/coerce-id.js
```diff
@@ -0,0 +1,9 @@
+// Used by the store to normalize IDs entering the store.  Despite the fact
```
sly7-7: So, not used only by the store. Maybe something like `Internally ED normalizes id to String`?

PR #3147, opened 2015-06-02
more cleanup on internalModel vs record

tchak:


packages/ember-data/lib/serializers/json-serializer.js
```diff
@@ -115,7 +115,7 @@ export default Serializer.extend({
   */
   applyTransforms: function(typeClass, data) {
     typeClass.eachTransformedAttribute(function applyTransform(key, typeClass) {
-      if (!data.hasOwnProperty(key)) { return; }
+      if (!(key in data)) { return; }
```
igorT: This data comes from jqxhr afaik, so can't do this

tchak: You right let's leave serializable hash alone for now. I removed the changes.

packages/ember-data/lib/system/model/internal-model.js
```diff
@@ -660,6 +659,14 @@ InternalModel.prototype = {
       }
     }
     this._inFlightAttributes = Ember.create(null);
+  },
+
+  toString: function() {
+    if (this.record) {
+      return this.record.toString();
```
igorT: :+1:

Issue #3152, opened 2015-06-02
Remove deprecated `pushRecord` method on RecordArray


bmac:
https://github.com/emberjs/data/blob/6d6c6462fe597be556d358585e9445513cbc662d/packages/ember-data/lib/system/record-arrays/record-array.js#L142-L153

sly7-7:
On it

sly7-7:
Woops, was not autoclosed :disappointed:

PR #3156, opened 2015-06-02
Remove deprecated push record

### LINKED ISSUE #3153: RecordArray#addRecord is unused
@igorT can this code be removed?

https://github.com/emberjs/data/blob/6d6c6462fe597be556d358585e9445513cbc662d/packages/ember-data/lib/system/record-arrays/record-array.js#L121-L136

sly7-7:
Closes #3153, #3152

### DIFF
--- a/packages/ember-data/lib/system/record-array-manager.js
+++ b/packages/ember-data/lib/system/record-array-manager.js
@@ -124,7 +124,7 @@ export default Ember.Object.extend({
 
     if (shouldBeInArray) {
       if (!recordArrays.has(array)) {
-        array._pushRecord(record);
+        array.addRecord(record);
         recordArrays.add(array);
       }
     } else if (!shouldBeInArray) {
--- a/packages/ember-data/lib/system/record-arrays/record-array.js
+++ b/packages/ember-data/lib/system/record-arrays/record-array.js
@@ -135,22 +135,6 @@ export default Ember.ArrayProxy.extend(Ember.Evented, {
     }
   },
 
-  _pushRecord: function(record) {
-    get(this, 'content').pushObject(record);
-  },
-
-  /**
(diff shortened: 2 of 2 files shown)


sly7-7:
Ooops, seems like something is wrong with my branch...

bmac:
Looks good to me. @igorT can you review?

packages/ember-data/lib/system/record-arrays/record-array.js
```diff
@@ -135,22 +135,6 @@ export default Ember.ArrayProxy.extend(Ember.Evented, {
...
-  /**
-    Adds a record to the `RecordArray`, but allows duplicates
-
-    @deprecated
-    @method pushRecord
-    @private
-    @param {DS.Model} record
-  */
-  pushRecord: function(record) {
-    Ember.deprecate('Usage of `recordArray.pushRecord` is deprecated, use `recordArray.addObject` instead');
```
igorT: I would still keep the deprecation until 1.0

igorT: but i bet this breaks with references refactor. Should call this.pushObject instead of on the content

packages/ember-data/lib/system/record-array-manager.js
```diff
@@ -124,7 +124,7 @@ export default Ember.Object.extend({
 
     if (shouldBeInArray) {
       if (!recordArrays.has(array)) {
-        array._pushRecord(record);
+        array.addRecord(record);
```
igorT: :+1:

packages/ember-data/lib/system/record-arrays/record-array.js
igorT: Actually nm, many arrays are not record arrays anymore, so we can remove this

PR #3158, opened 2015-06-02
[WIP] Only change isReloading to false if all reloads have settled

### LINKED ISSUE #3141: There might be potential problems with doing `reload` twice on a model

mikehollis:
This pull request fixes #3141

This test calls `.reload()` on a model twice before any response is received. Then, it checks both the `isReloading` flag as well as the data to make sure that it is correct after returning.

If you double issue reloads like this, the first one to come back changes the `isReloading` flag to be false even though the second reload is still in flight, which this test shows as currently failing.

### DIFF
--- a/packages/ember-data/lib/system/model/internal-model.js
+++ b/packages/ember-data/lib/system/model/internal-model.js
@@ -63,6 +63,8 @@ var InternalModel = function InternalModel(type, id, store, container, data) {
   this._relationships = new Relationships(this);
   this.currentState = RootState.empty;
   this.isReloading = false;
+  this._lastReloadingPromise = Ember.RSVP.resolve();
+  this._currentReloadCount = 0;
   /*
     implicit relationships are relationship which have not been declared but the inverse side exists on
     another record somewhere
@@ -135,6 +137,7 @@ InternalModel.prototype = {
   },
 
   startedReloading: function() {
+    this._currentReloadCount++;
     this.isReloading = true;
     if (this.record) {
       set(this.record, 'isReloading', true);
@@ -142,9 +145,12 @@ InternalModel.prototype = {
   },
 
   finishedReloading: function() {
-    this.isReloading = false;
-    if (this.record) {
-      set(this.record, 'isReloading', false);
+    this._currentReloadCount--;
+    if (this._currentReloadCount === 0) {
+      this.isReloading = false;
+      if (this.record) {
+        set(this.record, 'isReloading', false);
+      }
     }
   },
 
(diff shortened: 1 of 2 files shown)


mikehollis:
@igorT Here's my first try at fixing the reload() issue. I added a chained promise to the `internalModel` so that promises could be linked. Then I added a check so that if more promises had been added, it just leaves the reloading flags true.

Do you think this is a good approach? Does the code seem ugly?

packages/ember-data/lib/system/model/internal-model.js
```diff
@@ -135,9 +135,12 @@ InternalModel.prototype = {
   },
 
   finishedReloading: function() {
-    this.isReloading = false;
-    if (this.record) {
-      set(this.record, 'isReloading', false);
+    var record = this;
```
igorT: internalModel, not record

packages/ember-data/lib/system/model/internal-model.js
```diff
@@ -135,9 +135,12 @@ InternalModel.prototype = {
...
+    if (record._loadingPromiseChain._state) {
```
igorT: where does `_state` come from?

mikehollis: @igorT -- I thought that it was inherent in promises.

`_loadingPromiseChain` is a promise that keeps getting chained with new promises when a reload is called. A promise class sets the `_state` value when it completes either with an error or not. Since I'm always updating the `_loadingPromiseChain` to be the last promise on the chain, if that hasn't successfully finished (because someone else called `reload()`, then it doesn't change the flags back to false.

Did you want me to comment that in the code?

igorT: _state is not a public promises method, especially as it's underscored. cc @stefanpenner
you might have to do this manually for now

mikehollis:
@igorT, I switched over to use a PromiseObject Proxy rather than a straight promise, which allows me to query the status of a promise. Let me know what you think now.

Also, I know there are a ton of commits. I just wanted to keep track of what I was changing. If you like the solution, I'll squash all the commits into one, rebase, and force push before I remove the `[WIP]` on the name.

igorT:
Shouldn't use a promise proxy, lets look over it together later if you are free to pair for a bit, ping me on slack

mikehollis:
@igorT -- much simpler fix than I was doing. Thanks for the help!

packages/ember-data/lib/system/model/internal-model.js
```diff
@@ -136,15 +137,18 @@ InternalModel.prototype = {
...
     if (this.record) {
       set(this.record, 'isReloading', true);
     }
   },
 
   finishedReloading: function() {
-    this.isReloading = false;
-    if (this.record) {
-      set(this.record, 'isReloading', false);
+    if (--this._currentReloadCount === 0) {
```
igorT: can you split this into two lines please

igorT:
@mikehollis we still want the functionality of chaining them, and only resolving when they are all done, that seems to have disappeared.

igorT:
I can help over the weekend if you get stuck

mikehollis:
@igorT I added the reload chaining back in and squashed everything into a single commit.

Let me know what you think!

igorT:
Seems much better. The only tricky bit left is to not resolve the original reload until all the relaods have resolved. There are multiple ways to do this, but not sure what the best one is. Do you want to add the test case first?

mikehollis:
Sounds great! I'll try to tackle this this weekend

mikehollis:
@igorT I've added the test that checks that the first promise waits to resolve for the second one.

Although, I'm somewhat stumped as to how this would actually be implemented. You mentioned that there were a couple of ways. Any chance you could point me in the right direction?

bmac:
Hi @mikehollis, sorry this has been opened for so long. Any chance you have some time to rebase this pr?

bmac:
@mikehollis do you have some time to rebase this pr?

bmac:

Issue #2542, opened 2014-12-02
Store.find/all are created as filteredRecordArrays


igorT:
Not sure this is appropriate, it leads to confusion when reading the code and to subtle bugs.

igorT:
Does someone want to tackle this, @pangratz or @sly7-7 maybe?

pangratz:
I gave it a shot and changing `findAll` to return a `RecordArray` [breaks this assertion](http://git.io/vTAHK) (all other tests pass). As pointed out by @stefanpenner in https://github.com/emberjs/data/pull/518#issuecomment-17819789 this might be ok; if the `all()` functionality is desired, then [`store.all()`](http://git.io/vTAHZ) should be used...

igorT:
`findAll` should still cache the array and return the same cached array each time. It's just that we don't want it to be a `filteredRecordArray` as there is really no filtering except for deletion happening. So this would refactor the recordArrayManager to care about filteredRecordArrays and something like liveRecordArrays

sly7-7:
@pangratz I'm currently working on rebasing/fixing https://github.com/emberjs/data/pull/2491, which causes some tests to fail (store.all() would become run-loop sensible). Maybe it would be possible to include those changes https://github.com/sly7-7/data/commit/0435b64e2b81ef14a6dc70d85b25ab20472b5ea2 in your current work ? So both issues could be resolved ?

pangratz:
@sly7-7 I try to tackle this later today ...

sly7-7:
Closing since 3167 is merged

PR #3167, opened 2015-06-02
store.all() is not treated as a filteredRecordArray anymore

pangratz:
This addresses #2542 

---

I haven't squashed the commits so the changes are more comprehensible. If this is accepted, I will squash em all!

### DIFF
--- a/packages/ember-data/lib/system/record-array-manager.js
+++ b/packages/ember-data/lib/system/record-array-manager.js
@@ -27,6 +27,12 @@ export default Ember.Object.extend({
       defaultValue: function() { return []; }
     });
 
+    this.liveRecordArrays = MapWithDefault.create({
+      defaultValue: (typeClass) => {
+        return this.createRecordArray(typeClass);
+      }
+    });
+
     this.changedRecords = [];
     this._adapterPopulatedRecordArrays = [];
   },
@@ -77,7 +83,6 @@ export default Ember.Object.extend({
   },
 
 
-  //Don't need to update non filtered arrays on simple changes
   _recordWasChanged: function (record) {
     var typeClass = record.type;
     var recordArrays = this.filteredRecordArrays.get(typeClass);
@@ -85,9 +90,7 @@ export default Ember.Object.extend({
 
     forEach(recordArrays, function(array) {
       filter = get(array, 'filterFunction');
-      if (filter) {
-        this.updateRecordArray(array, filter, typeClass, record);
-      }
+      this.updateFilterRecordArray(array, filter, typeClass, record);
     }, this);
   },
 
@@ -99,37 +102,54 @@ export default Ember.Object.extend({
 
     forEach(recordArrays, function(array) {
       filter = get(array, 'filterFunction');
-      this.updateRecordArray(array, filter, typeClass, record);
+      this.updateFilterRecordArray(array, filter, typeClass, record);
     }, this);
+
+    if (this.liveRecordArrays.has(typeClass)) {
+      var liveRecordArray = this.liveRecordArrays.get(typeClass);
+      this._addRecordToRecordArray(liveRecordArray, record);
+    }
   },
   /**
     Update an individual filter.
 
-    @method updateRecordArray
+    @method updateFilterRecordArray
     @param {DS.FilteredRecordArray} array
     @param {Function} filter
     @param {DS.Model} typeClass
     @param {InternalModel} record
   */
-  updateRecordArray: function(array, filter, typeClass, record) {
-    var shouldBeInArray;
-
-    if (!filter) {
-      shouldBeInArray = true;
-    } else {
-      shouldBeInArray = filter(record.getRecord());
-    }
-
+  updateFilterRecordArray: function(array, filter, typeClass, record) {
+    var shouldBeInArray = filter(record.getRecord());
     var recordArrays = this.recordArraysForRecord(record);
 
     if (shouldBeInArray) {
-      if (!recordArrays.has(array)) {
-        array.addInternalModel(record);
-        recordArrays.add(array);
-      }
-    } else if (!shouldBeInArray) {
+      this._addRecordToRecordArray(array, record);
+    } else {
       recordArrays.delete(array);
       array.removeInternalModel(record);
+    }
+  },
+
+  _addRecordToRecordArray: function(array, record) {
+    var recordArrays = this.recordArraysForRecord(record);
+    if (!recordArrays.has(array)) {
+      array.addInternalModel(record);
+      recordArrays.add(array);
+    }
+  },
+
+  populateLiveRecordArray: function(array, modelName) {
+    var typeMap = this.store.typeMapFor(modelName);
+    var records = typeMap.records;
+    var record;
+
+    for (var i = 0, l = records.length; i < l; i++) {
+      record = records[i];
+
+      if (!record.isDeleted() && !record.isEmpty()) {
+        this._addRecordToRecordArray(array, record);
+      }
     }
   },
 
@@ -154,13 +174,25 @@ export default Ember.Object.extend({
       record = records[i];
 
       if (!record.isDeleted() && !record.isEmpty()) {
-        this.updateRecordArray(array, filter, modelName, record);
+        this.updateFilterRecordArray(array, filter, modelName, record);
       }
     }
   },
 
   /**
-    Create a `DS.RecordArray` for a type and register it for updates.
+    Get the `DS.RecordArray` for a type, which contains all loaded records of
+    given type.
+
+    @method liveRecordArrayFor
+    @param {Class} typeClass
+    @return {DS.RecordArray}
+  */
+  liveRecordArrayFor: function(typeClass) {
+    return this.liveRecordArrays.get(typeClass);
+  },
+
+  /**
+    Create a `DS.RecordArray` for a type.
 
     @method createRecordArray
     @param {Class} typeClass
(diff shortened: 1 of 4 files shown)


packages/ember-data/lib/system/record-array-manager.js
```diff
@@ -101,6 +107,13 @@ export default Ember.Object.extend({
       filter = get(array, 'filterFunction');
       this.updateRecordArray(array, filter, typeClass, record);
     }, this);
+
+    if (this.liveRecordArrays.has(typeClass)) {
+      var liveRecordArray = this.liveRecordArrays.get(typeClass);
+      if (!liveRecordArray.isDestroyed && !liveRecordArray.isDestroying) {
```
pangratz: The check for the record array not being destroyed is needed as otherwise [this test](http://git.io/vk5zB) fails.

I am not satisfied about this check; how should this be tackled? Check if the `RecordArray` which is about to be destroyed is a `liveRecordArray` and if so, remove it from the map?

sly7-7: it seems like previously it was handled by unregisterFilteredRecordArray, called when in the recordArray's destroy method.
Maybe this could be rename as unregisterRecordArray which could handle both filtered/live record arrays ?

packages/ember-data/lib/system/record-array-manager.js
```diff
@@ -27,6 +27,12 @@ export default Ember.Object.extend({
       defaultValue: function() { return []; }
     });
 
+    this.liveRecordArrays = MapWithDefault.create({
+      defaultValue: (typeClass) => {
+        return this.createRecordArray(typeClass);
```
sly7-7: could this pattern be applied to filteredRecordArrays as well ?

packages/ember-data/lib/system/record-array-manager.js
pangratz: Sounds good :+1:

packages/ember-data/lib/system/store.js
```diff
@@ -1045,18 +1045,11 @@ Store = Service.extend({
...
-    var findAllCache = typeMap.findAllCache;
 
-    if (findAllCache) {
-      this.recordArrayManager.updateFilter(findAllCache, typeClass);
-      return findAllCache;
-    }
-
-    var array = this.recordArrayManager.createRecordArray(typeClass);
+    var liveRecordArray = this.recordArrayManager.liveRecordArrayFor(typeClass);
+    this.recordArrayManager.updateFilter(liveRecordArray, typeClass);
```
sly7-7: is the call to updateFilter really necessary here ? It seems like it will end up to a noop for liveRecordArrays

pangratz: This call is needed so the initially created `RecordArray` is filled with the currently loaded records in the store.

pangratz: This call *could* move inside the `defaultValue` function for the `liveRecordArrays` map, but then [this spec](https://github.com/emberjs/data/blob/58c2cf65eb492165c56ba561803357b99020305a/packages/ember-data/tests/integration/all-test.js#L36-L46) fails...

sly7-7: I don't understand why the test fail, but... then I guess we should let the updateFilter here :smile:

pangratz: > It seems like it will end up to a noop for liveRecordArrays

The name of `updateFilter` does not correctly reflect what's going on here: it calls `updateRecordArray` for each currently loaded record of type `typeClass`; and since there is a [hack](https://github.com/emberjs/data/blob/58c2cf65eb492165c56ba561803357b99020305a/packages/ember-data/lib/system/record-array-manager.js#L118) inside `updateRecordArray` at the moment, it also adds all records to the array. I will address in this PR ...

sly7-7: Sorry I was unclear, I understood the call was necessary, but I don't understand why calling it in the default value make the test fail.

packages/ember-data/lib/system/record-array-manager.js
```diff
@@ -160,6 +171,17 @@ export default Ember.Object.extend({
...
+    @method liveRecordArrayFor
+    @param {Class} typeClass
+    @return {DS.RecordArray}
+  */
+  liveRecordArrayFor: function(typeClass) {
+    return this.liveRecordArrays.get(typeClass);
+  },
+
+  /**
     Create a `DS.RecordArray` for a type and register it for updates.
```
pangratz: This comment needs to be updated since the *register it for updates* is no more valid ...

packages/ember-data/lib/system/record-array-manager.js
```diff
@@ -160,6 +171,17 @@ export default Ember.Object.extend({
   },
 
   /**
+    Get the `DS.RecordArray` for a type.
```
pangratz: Add comment that the `RecordArray` returned by this function holds *all* loaded records in the store of the specifc type.

packages/ember-data/lib/system/store.js
pangratz: I think it's because in the test creating a record eventually would trigger a `recordWasLoaded` which in turn updates the record arrays – but this event is triggered [in the `actions` queue](https://github.com/emberjs/data/blob/37401a8380cc9e6c26e8ad701c4b2f92d34edf3d/packages/ember-data/lib/system/model/internal-model.js#L200), which would be executed in the next step of the run loop... I guess :thought_balloon:

packages/ember-data/lib/system/record-array-manager.js
pangratz: Unsure how this would be achieved since `filteredRecordArrays` has an array as value ...

pangratz:
I refactored the whole thing a little more and kept the commits so the changes are more comprehensible.

packages/ember-data/lib/system/record-array-manager.js
```diff
@@ -99,37 +102,54 @@ export default Ember.Object.extend({
...
+      array.addRecord(record);
+      recordArrays.add(array);
+    }
+  },
 
-    if (shouldBeInArray) {
-      if (!recordArrays.has(array)) {
-        array.addRecord(record);
-        recordArrays.add(array);
+  updateLiveRecordArray: function(array, modelName) {
```
igorT: This doesn't really update anymore, just initially populates it, so a different name/location might be better

packages/ember-data/lib/system/record-array-manager.js
```diff
@@ -245,16 +275,24 @@ export default Ember.Object.extend({
...
     @param {DS.RecordArray} array
   */
-  unregisterFilteredRecordArray: function(array) {
+  unregisterRecordArray: function(array) {
+    var typeClass = array.type;
+
+    // unregister filtered records arrays
     var recordArrays = this.filteredRecordArrays.get(array.type);
     var index = indexOf(recordArrays, array);
     recordArrays.splice(index, 1);
```
igorT: array.splice(-1,1) trolls you here pretty badly

packages/ember-data/tests/integration/record-array-manager-test.js
```diff
@@ -118,12 +118,12 @@ test("Should not filter a stor.all() array when a record property is changed", f
     });
   });
 
-  equal(filterdSummary.called.length, 1);
+  equal(filterdSummary.called.length, 0);
```
igorT: We want to assert that updateRecordArray is also called only once before and after

packages/ember-data/lib/system/record-array-manager.js
pangratz: Uff, good catch!

igorT:
Can you squash please

packages/ember-data/lib/system/record-array-manager.js
```diff
@@ -245,16 +275,26 @@ export default Ember.Object.extend({
...
+    @method unregisterRecordArray
     @param {DS.RecordArray} array
   */
-  unregisterFilteredRecordArray: function(array) {
-    var recordArrays = this.filteredRecordArrays.get(array.type);
+  unregisterRecordArray: function(array) {
+    var typeClass = array.type;
+
+    // unregister filtered records arrays
+    var recordArrays = this.filteredRecordArrays.get(typeClass);
```
igorT: Seems like we could just know which one it is, and don't need to try both?

igorT:
Thanks @pangratz !
Looks good after that one comment and squashing

fivetanley:
This looks like it makes it so if you call `store.all('post')` more than once, you always get the same record array back?

fivetanley:
Looks like it. :+1:

igorT:
@fivetanley that was the case before as well,this just refactors the code

sly7-7:
@pangratz Great Work !! Also next time instead of `adresses`, could you use `fixes`or `closes` the referencing issue ?

pangratz:
> @pangratz Great Work !! Also next time instead of adresses, could you use fixesor closes the referencing issue ?

@sly7-7 I used to do this in my commit messages but I started to refrain from this since otherwise the referenced issue is cluttered with `@pangratz referenced this pull request in commit xxx`, every time I do a rebase and (re)push to GitHub.

I should get into the habit of adding the `closes #xxx` once the PR is ready to be merged ...

sly7-7:
`I should get into the habit of adding the closes #xxx once the PR is ready to be merged ...`
:+1: 
For me adding the `closes`thing just after my first commit/push, but in the comment part does the trick

PR #3170, opened 2015-06-03
Ensure snapshot.belongsTo() and hasMany() do not return deleted records

### LINKED ISSUE #3081: Snapshot returns deleted ids in hasMany call
We have a situation where we delete a record, and then the main model is saved (that contains a hasMany relation for the record that was just deleted). When it goes to serialize the hasMany ids in that relation, the deleted records (but not yet saved), are included in that list. 

It seems like if a record is marked as deleted, it shouldn't be included in the snapshot hasMany along with the live records. The snapshot when gathering the ids doesn't do any filtering, but feels like it should check isDeleted before including it in the snapshot.

https://github.com/emberjs/data/blob/v1.0.0-beta.17/packages/ember-data/lib/system/snapshot.js#L276

https://www.dropbox.com/s/lijhqe0gy37t0q7/Screenshot%202015-05-18%2016.31.29.png?dl=0

ianstarz:
Even though we remove records from their opposite relationships once they are deleted, this commit places this functionality under test and ensures that only non deleted records are returned from snapshot.belongsTo() and snapshot.hasMany().

Closes #3081

### DIFF
--- a/packages/ember-data/tests/integration/snapshot-test.js
+++ b/packages/ember-data/tests/integration/snapshot-test.js
@@ -166,6 +166,22 @@ test("snapshot.belongsTo() returns a snapshot if relationship is set", function(
   });
 });
 
+test("snapshot.belongsTo() returns null if relationship is deleted", function() {
+  expect(1);
+
+  run(function() {
+    var post = env.store.push('post', { id: 1, title: 'Hello World' });
+    var comment = env.store.push('comment', { id: 2, body: 'This is comment', post: 1 });
+
+    post.deleteRecord();
+
+    var snapshot = comment._createSnapshot();
+    var relationship = snapshot.belongsTo('post');
+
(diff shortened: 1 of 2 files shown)


ianstarz:
@igorT said we may want to ensure this functionality even though tests pass without the implementation in the snapshot package. At the very least, we do want the test coverage I believe.

packages/ember-data/tests/integration/snapshot-test.js
```diff
@@ -246,7 +262,7 @@ test("snapshot.belongsTo() and snapshot.hasMany() returns correctly when setting
   });
 });
 
-test("snapshot.hasMany() returns ID if option.id is set", function() {
+test("snapshot.belongsTo() returns ID if option.id is set", function() {
```
ianstarz: Test was incorrectly named

PR #3173, opened 2015-06-03
add tests for error conditions in snapshot methods

### LINKED ISSUE #3157: Snapshot#attr's error condition is untested
There should be a test asserting that the error is thrown.

https://github.com/emberjs/data/blob/master/packages/ember-data/lib/system/snapshot.js#L133

yratanov:
* snapshot.attr()
* snapshot.belongsTo()
* snapshot.hasMany()

fixes #3157


sly7-7:
Thanks :+1:

PR #3175, opened 2015-06-03
add missing tests for record arrays

### LINKED ISSUE #3154: FilteredRecordArray#replace and AdapterPopulatedRecordArray#replace needs a test.
It should assert that an error is thrown when it is called.

https://github.com/emberjs/data/blob/87dec3e56b6b1744c49103417cc8cacc8f1edb23/packages/ember-data/lib/system/record-arrays/filtered-record-array.js#L49-L52

https://github.com/emberjs/data/blob/master/packages/ember-data/lib/system/record-arrays/adapter-populated-record-array.js#L29-L32

yratanov:
fixes #3154

### DIFF
--- a/packages/ember-data/tests/unit/record-arrays/filtered-record-array-test.js
+++ b/packages/ember-data/tests/unit/record-arrays/filtered-record-array-test.js
@@ -1,0 +1,13 @@
+var filteredArray;
+
+module("unit/record-arrays/filtered-record-array - DS.FilteredRecordArray", {
+  setup: function() {
+    filteredArray = DS.FilteredRecordArray.create({ type: 'recordType' })
+  }
+});
+
+test('recordArray.replace() throws error', function() {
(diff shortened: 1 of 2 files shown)


packages/ember-data/tests/unit/record-arrays/filtered-record-array-test.js
```diff
@@ -0,0 +1,13 @@
+var filteredArray;
+
+module("unit/record-arrays/filtered-record-array - DS.FilteredRecordArray", {
+  setup: function() {
+    filteredArray = DS.FilteredRecordArray.create({type: 'recordType'})
```
sly7-7: linter is angry here, `DS.FilteredRecordArray.create({ type: 'recordType' })` (missing spaces)

bmac:
Thanks @yratanov

PR #3176, opened 2015-06-03
Fix JSDoc issues

Turbo87:
see commits

### DIFF
--- a/packages/ember-data/lib/adapters/build-url-mixin.js
+++ b/packages/ember-data/lib/adapters/build-url-mixin.js
@@ -43,8 +43,8 @@ export default Ember.Mixin.create({
 
     @method buildURL
     @param {String} modelName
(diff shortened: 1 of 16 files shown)


bmac:
This looks awesome! Do you mind squashing it down to 1 commit?

Turbo87:
sure, no problem. may I ask why though?

bmac:
Actually looking at these commit messages and the fact that this is mostly doc changes, I'm going to merge it as is. 

By convention, Ember Data usually squashes down most prs to make it easier to git bisect thought the code base (lading on a commit with broken tests makes it hard to determine when a bug was introduced).

Turbo87:
I mostly agree, though often I find it easier to bisect when the commits are smaller and that's why I tend to make only one change per commit. Obviously that only works when there are no failing tests, but then again the tests shouldn't fail in the first place. Thanks for merging!

PR #3179, opened 2015-06-03
Use object literal methods

Turbo87:
Babel will transpile them into the previous syntax
(see https://babeljs.io/docs/learn-es2015/#enhanced-object-literals)

### DIFF
--- a/packages/activemodel-adapter/lib/system/active-model-adapter.js
+++ b/packages/activemodel-adapter/lib/system/active-model-adapter.js
@@ -115,7 +115,7 @@ var ActiveModelAdapter = RESTAdapter.extend({
     @param {String} modelName
     @return String
   */
-  pathForType: function(modelName) {
+  pathForType(modelName) {
     var decamelized = decamelize(modelName);
     var underscored = underscore(decamelized);
     return pluralize(underscored);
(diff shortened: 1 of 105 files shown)


igorT:
Thanks for the PR. I feel that object literal methods make searching through the code base significantly harder, so would not want to switch to them for a while, until our tooling becomes better. Thanks!

Turbo87:
could you elaborate on why you find it harder with object literal methods?

igorT:
If you want to search for where serialize is defined normally you can just search for 'serailize:`, with object literal methods you need to use a javascript parser in command line/editor, and good luck in chrome scripts if you use source maps

Turbo87:
ah, I see... so the reason is that method definition and method calls would then look similar to a plain text search.

I guess one could take advantage of the fact that methods are usually indented and method calls are prefixed with something like `this.someMethod()`. In theory that would allow searching for <code>&nbsp;&nbsp;serialize(</code> to find the method definition, right?

Please don't take my questions the wrong way. I'm not trying to convince you of anything, I'm just trying to understand why you don't want to use that feature.

rwjblue:
yes, you could search for `/\s+serialize[(:]/g` to find either instances, but that is a ton more work than just searching for `serialize:` :smile: 

Demo: http://regexr.com/3b4ph

Turbo87:
lol, it's obviously more complicated if you want to search for **both** styles ;)

PR #3180, opened 2015-06-03
add tests for restAdapter#ajaxOptions method

### LINKED ISSUE #3151: Sending headers from the RESTAdapter needs a test
https://github.com/emberjs/data/blob/4d5a5a5135b016d900b4c4ab3692e85fc340eb4c/packages/ember-data/lib/adapters/rest-adapter.js#L821-L827

yratanov:
* rename adapters/rest-adapter/ajax to ajax-test. QUnit didn't run this file before.
 * fixes #3151


igorT:
Too late now, but usually if you are doing changes like these, it's good to first move the file using git mv so that blame will work in the future

igorT:
Thanks!

PR #3181, opened 2015-06-03
system/model: Simplify trigger() method with slice()

Turbo87:
work in progress


Turbo87:
nevermind, just read https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/arguments ;)

igorT:
Keep in mind this is a very hot path, needs to be pretty optimized

Turbo87:
yep, that's why I thought slice() might be a good idea, but it seems that
doing it manually is indeed faster.

PR #3182, opened 2015-06-03
initialize lives on the application instance not the Ember namespace

bmac:
Fixes Ember Data canary with version of Ember 1.11 or lower. Credit goes to @fivetanley for reporting this issue.


igorT:
Do we even support 1.11? Whats the lowest we support

bmac:
Yes we support 1.11. The lowest version of Ember that Ember Data supports is 1.8 source http://emberjs.com/blog/2014/11/24/ember-data-1-0-beta-12-released

PR #3191, opened 2015-06-04
refactor store managed instances

### LINKED ISSUE #3050: [Proposal] Resolving an adapter-less model should cache the 'application adapter' resolution.

fivetanley:
In this commit, we refactor store managed instance lookups to have its
own class that takes the store's container. This moves a large portion
of code out of the store and instead uses delegation.

There's also a performance benefit:

If you look up `adapter:post` and fallback to `adapter:application`,
this lookup is cached. So, if you look up `adapter:post` again, that
instance will actually be cached. Previously, we did the lookup for
`adapter:post` every time.

This also removes passing factories as the `adapter` property to
DS.Store.

closes #3050

### DIFF
--- a/packages/ember-data/lib/system/store.js
+++ b/packages/ember-data/lib/system/store.js
@@ -7,8 +7,7 @@
 
 import normalizeModelName from "ember-data/system/normalize-model-name";
 import {
-  InvalidError,
-  Adapter
+  InvalidError
 } from "ember-data/system/adapter";
 import {
   Map
@@ -41,6 +40,7 @@ import {
 import coerceId from "ember-data/system/coerce-id";
 
 import RecordArrayManager from "ember-data/system/record-array-manager";
+import ContainerInstanceCache from 'ember-data/system/store/container-instance-cache';
 
 import InternalModel from "ember-data/system/model/internal-model";
 import Model from "ember-data/system/model";
@@ -214,7 +214,7 @@ Store = Service.extend({
       store: this
     });
     this._pendingSave = [];
-    this._containerCache = Ember.create(null);
+    this._instanceCache = new ContainerInstanceCache(this.container);
     //Used to keep track of all the find requests that need to be coalesced
     this._pendingFetch = Map.create();
   },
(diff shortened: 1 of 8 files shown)


packages/ember-data/lib/system/store.js
```diff
@@ -1958,30 +1943,28 @@ Store = Service.extend({
     @private
     @param {String} modelName the object modelName
     @param {String} name the object name
+    @param {Object} fallback the fallback object to return if the lookup for modelName or 'application' fails
```
bmac: `{Array} fallbacks the list of the fallback objects`...

packages/ember-data/lib/system/store/container-instance-cache.js
```diff
@@ -0,0 +1,86 @@
...
+*/
+function ContainerInstanceCache(container) {
+  this._container  = container;
+  this._cache      = Ember.create(null);
+}
+
+ContainerInstanceCache.prototype = Ember.create(null);
+
+Ember.merge(ContainerInstanceCache.prototype, {
+  get(type, preferredKey, fallbacks) {
```
bmac: I think @igorT  wants to avoid ES6 concise methods. https://github.com/emberjs/data/pull/3179

fivetanley: seems like we should do whatever ember does

fivetanley: then again i don't care enough about this particular bikeshed ;)

igorT:
:+1:

packages/ember-data/lib/system/store/container-instance-cache.js
igorT: https://github.com/emberjs/data/pull/3179
They make searching much worse for no benefit other than minor prettyness. Last time I talked to Ember core team, they seemed pretty divided but decided to go ahead.

PR #3193, opened 2015-06-04
Fix Store#findHasMany (?)

sozuuuuu:
#### Problem

1. The doc says Store#findHasMany accepts type as string
2. Store#findHasMany passes the type to Store#_findHasMany and that references type of type (type.type) => returns `undefined` and it causes an error

#### Fix

- Store#findHasMany looks up the container to get the factory if a string passed
- On Store#findHasMany the methods pass `relationship` instead of `relationship.type`


sozuuuuu:
failing tests

PR #3198, opened 2015-06-04
Allow Ember Data to update the id of a record if the record is in the…

bmac:
… loading state

### DIFF
--- a/packages/ember-data/lib/system/store.js
+++ b/packages/ember-data/lib/system/store.js
@@ -1384,12 +1384,30 @@ Store = Service.extend({
   updateId: function(internalModel, data) {
     var oldId = internalModel.id;
     var id = coerceId(data.id);
+    var isLoading = internalModel.isLoading();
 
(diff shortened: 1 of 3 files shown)


packages/ember-data/lib/system/store.js
```diff
@@ -1384,12 +1384,30 @@ Store = Service.extend({
...
+    var typeMap = this.typeMapFor(internalModel.type);
+
+    if (oldId && oldId !== id) {
+      // Old ID and new ID do not match. The internalModel should be
+      // moved in the identity map
+      if (typeMap.idToRecord[id]) {
+        // Record already exists for that id. Destroy this
+        // internalModel so we can merge it with the existing
+        // internalModel.
+        this._dematerializeRecord(internalModel);
```
igorT: We need to merge the data between them, or assert that they don't conflict, in terms of relationships

igorT:
@bmac I realized that we shouldn't just move it, as it will mess with our adapter hooks. In fact we should let it live in both places, but only after people opt in(for example via a `slug` field)

PR #3202, opened 2015-06-04
use the esperanto module transpiler

fivetanley:
don't merge til the next release; this breaks ie8

### DIFF
--- a/Brocfile.js
+++ b/Brocfile.js
@@ -1,7 +1,5 @@
 /* jshint node: true */
 
-var es6             = require('broccoli-es6-module-transpiler');
-var PackageResolver = require('es6-module-transpiler-package-resolver');
 var concat          = require('broccoli-sourcemap-concat');
 var uglify          = require('broccoli-uglify-js');
 var es3SafeRecast   = require('broccoli-es3-safe-recast');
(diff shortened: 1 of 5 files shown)


stefanpenner:
why does this break IE8?

fivetanley:
https://github.com/esperantojs/esperanto/wiki/Converting-a-single-module#options

> _evilES3SafeReExports - will cause re-exported bindings to be done with direct property assignment rather than Object.defineProperty(). In most cases, the resulting behaviour will be no different, but it could result in undefined bindings in cases of cyclical dependencies, and values are fixed at the time of re-export rather than live

we don't use that option

stefanpenner:
> we don't use that option

why not? 

I am confused why ember uses esperanto while supporting IE8 but this does not

fivetanley:
we won't be supporting ie8 as of the next release

stefanpenner:
> we won't be supporting ie8 as of the next release

ah, so no reason other then we are dropping support anyways

fivetanley:
yeah, apparently @igort has something that requires cyclical deps?

igorT:
@fivetanley I could refactor but you said to just wait 10 min :)

stefanpenner:
Esperanto is no more

PR #3203, opened 2015-06-04
Use string model names in debug adapter

teddyzeenny:
Related to https://github.com/emberjs/ember.js/pull/11346

### DIFF
--- a/packages/ember-data/lib/system/debug/debug-adapter.js
+++ b/packages/ember-data/lib/system/debug/debug-adapter.js
@@ -5,6 +5,7 @@ import Model from "ember-data/system/model";
 var get = Ember.get;
 var capitalize = Ember.String.capitalize;
 var underscore = Ember.String.underscore;
+const { assert } = Ember;
 
 /**
   Extend `Ember.DataAdapter` with ED specific code.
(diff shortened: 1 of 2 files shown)


packages/ember-data/lib/system/debug/debug-adapter.js
```diff
@@ -42,11 +43,15 @@ export default Ember.DataAdapter.extend({
...
-    // TODO: Ask Teddy what we should do here.
-    // Ideally this should always get passed a string.
-
-    var modelName = typeof modelNameOrFactory === 'string' ? modelNameOrFactory : modelNameOrFactory.modelName;
+  getRecords: function(klass, modelName) {
+    if (arguments.length < 2) {
+      // Legacy Ember.js < 1.13 support
+      let containerKey = klass._debugContainerKey;
+      if (containerKey) {
+        modelName = containerKey.match(/model:(.*)/)[1];
```
fivetanley: if there is no match won't this return null, and thus error out when trying to use `[1]` to access a property on null?

teddyzeenny: Yes, but I couldn't think of a reason why it wouldn't match. Do you think there's a case where the model factory's container key doesn't contain `model:`?

teddyzeenny: I added a check just in case.

packages/ember-data/lib/system/debug/debug-adapter.js
```diff
@@ -42,11 +43,18 @@ export default Ember.DataAdapter.extend({
     return columns;
   },
 
-  getRecords: function(modelNameOrFactory) {
-    // TODO: Ask Teddy what we should do here.
-    // Ideally this should always get passed a string.
-
-    var modelName = typeof modelNameOrFactory === 'string' ? modelNameOrFactory : modelNameOrFactory.modelName;
+  getRecords: function(klass, modelName) {
```
igorT: i think we use modelClass for this case

PR #3204, opened 2015-06-04
Add test for normalizing attrs inside links.

### LINKED ISSUE #3188: Add tests for normalizing `links` in Json and Rest serializer
We don't test what happens if you set a custom key in attrs for a relationship and receive that relationship with a link

Serabe:
This tests checkes that the attrs object in serializers are taking into account
inside the links object too.

Fix #3188

### DIFF
--- a/packages/ember-data/tests/integration/serializers/json-serializer-test.js
+++ b/packages/ember-data/tests/integration/serializers/json-serializer-test.js
@@ -237,6 +237,27 @@ test('Serializer should respect the attrs hash when extracting records', functio
   deepEqual(post.comments, [1,2]);
 });
 
+test('Serializer should respect the attrs hash in links', function() {
+  env.registry.register("serializer:post", DS.JSONSerializer.extend({
+    attrs: {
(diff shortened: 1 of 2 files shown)


igorT:
cc @wecc

igorT:
The test looks good, thanks @Serabe, I think this should pass on master once @wecc rebases this/wraps in feature flags

wecc:
This is not passing on master even with `normalizeSingleResponse`. The only place we actually take `attrs` into consideration is in `normalizeUsingDeclaredMapping()`.

Serabe:
@wecc should I change the test or just ignore the PR?

wecc:
@Serabe completely missed this, sorry I haven't responded. If we want the keys in `links` to be normalized using the `attrs` object I think we need to modify the behavior of `normalizeUsingDeclaredMapping`.

bmac:
@Serabe Thanks for this pr. I have cleaned up your tests and used them in this pr that fixes the original issue. https://github.com/emberjs/data/pull/4268

PR #3211, opened 2015-06-05
lookup JSONSerializer instance through store instead of manual instan…

fivetanley:
…tiation

### DIFF
--- a/packages/ember-data/lib/system/model/model.js
+++ b/packages/ember-data/lib/system/model/model.js
@@ -1,5 +1,4 @@
 import { PromiseObject } from "ember-data/system/promise-proxies";
-import JSONSerializer from "ember-data/serializers/json-serializer";
 
 /**
   @module ember-data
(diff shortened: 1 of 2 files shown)


packages/ember-data/lib/system/model/model.js
```diff
@@ -375,7 +374,7 @@ var Model = Ember.Object.extend(Ember.Evented, {
   */
   toJSON: function(options) {
     // container is for lazy transform lookups
-    var serializer = JSONSerializer.create({ container: this.container });
+    var serializer = this.store.serializerFor('-default');
```
sandstrom: Would it be possible to do something like `store.serializerFor(this.constructor.typeKey)` instead? To get the record-specific serializer when one is available.

We've had this work-around for a long time (code not aligned with recent ED, but it shows the main idea):
```
toJSON: function(options) {
  var serializer = this.store.serializerFor(this.constructor.typeKey);
  return serializer.serialize(this, options);
}
```

igorT: no, if you want the record specific serializer, you `record.serialize()`. `toJSON` is using the json serializer for debugging and similar purposes

sly7-7: I have a doubt, is the '-default' serializer always JsonSerializer ? Is'nt it possible to override this ?

sandstrom: You are right, thanks! This part of the code is old, I see now that the issue I worked around has been solved for ~2 years :smile:

PR #3214, opened 2015-06-05
serialize:true takes priority over the OneToMany check for relationships

### LINKED ISSUE #3189: Serialize:true should take priority over the OneToMany check for relationships
https://github.com/emberjs/data/blob/ae6355af8235ca42d7dba0b3cf7a400732aa3c99/packages/ember-data/lib/serializers/json-serializer.js#L644 we should have a check for something like `mustSerialize` and account for it https://github.com/emberjs/data/blob/ae6355af8235ca42d7dba0b3cf7a400732aa3c99/packages/ember-data/lib/serializers/json-serializer.js#L656

sly7-7:
Fixes #3189 

Apologize in advance for my frenglished documentation
cc/ @igorT

### DIFF
--- a/packages/ember-data/lib/serializers/json-serializer.js
+++ b/packages/ember-data/lib/serializers/json-serializer.js
@@ -758,6 +758,41 @@ export default Serializer.extend({
     return !attrs || !attrs[key] || attrs[key].serialize !== false;
   },
 
+  /**
+    When attrs.key.serialize is set to true then
+    it takes priority over the other checks and the related
+    attribute/relationship will be serialized
+
+    @method _mustSerialize
+    @private
+    @param {String} key
+    @return {boolean} true if the key must be serialized
+  */
+  _mustSerialize: function(key) {
+    var attrs = get(this, 'attrs');
+
+    return attrs && attrs[key] && attrs[key].serialize === true;
+  },
+
+  /**
+    Check if the given hasMany relationship should be serialized
+
+    @method _shouldSerializeHasMany
+    @private
+    @param {DS.Snapshot} snapshot
+    @param {String} key
+    @param {String} relationshipType
+    @return {boolean} true if the hasMany relationship should be serialized
+  */
+  _shouldSerializeHasMany: function (snapshot, key, relationship) {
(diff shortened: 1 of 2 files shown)


packages/ember-data/lib/serializers/json-serializer.js
```diff
@@ -111,6 +111,19 @@ export default Serializer.extend({
...
+    The `_mustSerialize` property is set to true when the user
+    explicitely use `{serialize: true}` when defining the attrs hash.
+
+    Specifically it takes priority over the other checks and the related
+    attribute/relationship will be serialized
+
+    @property _mustSerialize
+    @private
+  */
+  _mustSerialize: false,
```
igorT: This doesn't make sense as a flag, it's a value that changes for each key, so needs to be a function, but we can just make it a part of canSerialize

igorT:
@sly7-7 we should implement `shouldSerializeRelationship` and that can implement this logic + `oneToMany` check and then we call that instead of the check on 692 and the

sly7-7:
@igorT I think I understand, I don't know when you want this, but I won't be able to do before tomorrow evening (trip today, soccer tomorrow). If it's not to late I'm happy to take that.

packages/ember-data/lib/serializers/json-serializer.js
sly7-7: Did'nt thought about that, definitely I will remove that.

igorT:
No rush, enjoy the game

sly7-7:
@igorT Ping ?

igorT:
Seems to fail on Travis due to jshint

sly7-7:
@igorT I'm confused, I can't reproduce the jshint error locally. I just rebase and force push... let's see whan happen

sly7-7:
@igorT It's ready if you don't have any other (welcome) advices :)

igorT:
@sly7-7 I think we talked about moving the 
```
+      if (this._shouldSerializeHasMany(key, relationshipType)) {
```
to be next to the canSerialize line right?

igorT:
Thanks!

PR #3215, opened 2015-06-05
remove Map/MapWithDefault polyfills, use Ember's ones

### LINKED ISSUE #2512: Remove Map/MapWithDefault/Set polyfills after beta.12 is released

sly7-7:
Closes #2512


fivetanley:
Need to make sure with @stefanpenner these aren't likely to change anytime soon.

sly7-7:
@fivetanley @stefanpenner any news on that ?

stefanpenner:
@fivetanley is ED relying on the public API's of Map/MapWithDefault ?

stefanpenner:
I suspect as long as,

* size
* set
* get
* forEach
* delete
* has
* clear

are the only methods used, everything should be fine.

sly7-7:
@stefanpenner Thanks for the investigation :+1: 

As a side note, is it safe to use Ember.Map/MapWithDefault in client app ?

stefanpenner:
> As a side note, is it safe to use Ember.Map/MapWithDefault in client app ?
>  Show all checks

I would say so, but I didn't release they are marked as private. Someone should investigate :P

sly7-7:
@stefanpenner It seems like Map has been marked private since the beggining, at least since the first move to yuidoc, see: https://github.com/emberjs/ember.js/commit/1325935d1396fd9a37b378d5216b9980e2acf83a#diff-7d75f034e93c4304edfd8944392ef2e0R60

Also, [here](https://github.com/emberjs/ember.js/pull/11362) @rwjblue took that in account to mark all the method as private, but it seems like there is no pb to make them public if wanted/needed

sly7-7:
@fivetanley ping ?

wecc:
@sly7-7 I think this would be great to get merged, would you be able to sort out the conflicts?

sly7-7:
@wecc My pleasure, Should be ok now :smile:

bmac:
Thanks @sly7-7.

PR #3216, opened 2015-06-05
Record property changes

### LINKED ISSUE #2900: No model update from PUT request result in beta.15

### LINKED ISSUE #2937: Model's property observer is triggered after save although value isn't changed by response

bmac:
This is a rebased and cleaned version of https://github.com/emberjs/data/pull/2998.

Closes #2937.
Closes #2900.
Closes #2998.

### DIFF
--- a/packages/ember-data/lib/system/model/internal-model.js
+++ b/packages/ember-data/lib/system/model/internal-model.js
@@ -190,7 +190,8 @@ InternalModel.prototype = {
   },
 
   setupData: function(data) {
-    var changedKeys = mergeAndReturnChangedKeys(this._data, data);
+    var changedKeys = this._changedKeys(data);
+    merge(this._data, data);
     this.pushedData();
     if (this.record) {
       this.record._notifyProperties(changedKeys);
@@ -584,11 +585,11 @@ InternalModel.prototype = {
     @method adapterDidCommit
   */
   adapterDidCommit: function(data) {
-    var changedKeys;
     this.didCleanError();
+    var changedKeys = this._changedKeys(data);
 
     if (data) {
-      changedKeys = mergeAndReturnChangedKeys(this._data, data);
+      merge(this._data, data);
     } else {
       merge(this._data, this._inFlightAttributes);
     }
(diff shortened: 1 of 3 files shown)


packages/ember-data/lib/system/model/internal-model.js
```diff
@@ -663,6 +663,42 @@ InternalModel.prototype = {
...
+
+        // A value in _attributes means the user has a local change to
+        // this attributes. We never override this value when merging
+        // updates from the backend so we should not sent a change
+        // notification if the server value differs from the original.
+        if (this._attributes[key] !== undefined) {
+          continue;
+        }
+
+        if (original[key] !== value) {
```
fivetanley: Can we switch this to use Ember.isEqual while we're at it? http://emberjs.com/api/#method_isEqual

packages/ember-data/lib/system/model/internal-model.js
```diff
@@ -663,6 +663,42 @@ InternalModel.prototype = {
...
   },
 
+  /**
+    @method _changedKeys
+    @private
+  */
+  _changedKeys: function(updates) {
+    var changedKeys = [];
+
+    if (updates && typeof updates === 'object') {
```
igorT: why typeof updates === 'object')

packages/ember-data/lib/system/model/internal-model.js
```diff
@@ -663,6 +663,42 @@ InternalModel.prototype = {
...
+      var original, i, value, key;
+      var keys = Ember.keys(updates);
+      var length = keys.length;
+
+      original = merge({}, this._data);
```
igorT: isn't it `Object.create(null)`?

packages/ember-data/lib/system/model/internal-model.js
```diff
@@ -663,6 +663,42 @@ InternalModel.prototype = {
     this._inFlightAttributes = Ember.create(null);
   },
 
+  /**
+    @method _changedKeys
+    @private
+  */
+  _changedKeys: function(updates) {
```
bmac: Add a comment here to describe this algorithm because it can be tricky to understand when revisiting.

packages/ember-data/lib/system/model/internal-model.js
```diff
@@ -663,6 +663,42 @@ InternalModel.prototype = {
...
+    var changedKeys = [];
+
+    if (updates && typeof updates === 'object') {
+      var original, i, value, key;
+      var keys = Ember.keys(updates);
+      var length = keys.length;
+
+      original = merge({}, this._data);
+      original = merge(original, this._inFlightAttributes);
```
igorT: Can't we also just merge this._attribtues here as well?

bmac:
@igorT I have updated this pr to address your comments.

packages/ember-data/lib/system/model/internal-model.js
bmac: `_attributes` do not need to be merged because we don't include it in the changed keys.

packages/ember-data/lib/system/model/internal-model.js
```diff
@@ -663,6 +663,80 @@ InternalModel.prototype = {
     this._inFlightAttributes = Ember.create(null);
   },
 
+  /**
+    @method _changedKeys
+
+    Ember Data has 3 buckets for storing the value of an attribute on an internalModel.
```
igorT: :heart:

PR #3217, opened 2015-06-05
Remove the 'container' property from model instances

slindberg:
From the commit:

> Models have always been created/managed by a store instance, and
> support for multiple stores makes accessing the container from a model
> inconsistent (and potentially dangerous), given that all adapter/serializer
> instances are looked up through a store's ContainerInstanceCache.

I think this makes sense, but I don't know if removing the `container` property from `DS.Model` instances would be considered a breaking change. If so, I can create a deprecated property accessor. @igorT, @fivetanley thoughts?

### DIFF
--- a/packages/ember-data/lib/system/model/model.js
+++ b/packages/ember-data/lib/system/model/model.js
@@ -373,7 +373,6 @@ var Model = Ember.Object.extend(Ember.Evented, {
     @return {Object} A JSON representation of the object.
   */
   toJSON: function(options) {
-    // container is for lazy transform lookups
     var serializer = this.store.serializerFor('-default');
     var snapshot = this._internalModel.createSnapshot();
 
--- a/packages/ember-data/lib/system/model/internal-model.js
+++ b/packages/ember-data/lib/system/model/internal-model.js
@@ -47,11 +47,10 @@ function retrieveFromCurrentState(key) {
   @class InternalModel
 */
 
-var InternalModel = function InternalModel(type, id, store, container, data) {
+var InternalModel = function InternalModel(type, id, store, data) {
   this.type = type;
   this.id = id;
   this.store = store;
-  this.container = container;
   this._data = data || Ember.create(null);
   this.modelName = type.modelName;
   this.errors = null;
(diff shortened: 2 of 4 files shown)


fivetanley:
Container is currently private. I don't think it's ever been documented as public.

slindberg:
Don't know what to make of the appveyor error, seems like a configuration error? @fivetanley is this familiar you?

igorT:
@slindberg appveyor has been crap recently, i wouldn't worry about it

igorT:
@slindberg i would put in an error message that tells you why it is wrong and what you can do instead, and that it is private(though you'll be able to `this.store.container`

packages/ember-data/lib/system/model/model.js
```diff
@@ -33,6 +33,11 @@ var Model = Ember.Object.extend(Ember.Evented, {
 
   store: null,
 
+  // Remove support for `record.container`
+  get container() {
+    Ember.assert("A model's private reference to the container has been removed for consistency reasons. Instances of a store (i.e. `this.store`) have access to the container, although usage from within a model is discouraged and can lead to unintended consequences.");
+  },
+
```
slindberg: @igorT I read that beta.19 is the last ED release to support IE8 :trollface: 

Seriously though, I started by doing a `Ember.defineProperty` on each model instance right after it's created, but it seemed like a bad choice for perf.

Everything passes when running tests in the browser, but for some reason destroys all tests in node. Debugging in node is proving to be fruitless, is this something I should even be trying to do?

igorT: You should create the getter dynamically inside an Ember.runInDebug so it gets stripped in production. However apparently phantom 1.9 doesn't implement defineProperty, but 2.0 does so @fivetanley is the person to ask about that

fivetanley: I'm in favor of removing. Telling people that we moved a private API and encouraging them to use another private API seems bad.

igorT: We should tell them to use a public api, but Ember is lacking public container apis, so it's an Ember bug for now

slindberg:
@igorT's comment about public container APIs made me realize that even though the `container` property has never been public, its existence is what [enables service injections](https://github.com/emberjs/ember.js/blob/bfcc15ee3ac2b2b3e4dad6aa0d7a447936302f5a/packages/ember-metal/lib/injected_property.js#L29). This has the potential to break seemingly innocent app code.

I still think that injecting services into models is an anti-pattern and should be actively disallowed, but it's probably worth an assertion explaining why.

@mixonic brought up some good points when I was working on injected properties, I'd like to hear his opinion on using services from within models.

fivetanley:
Removing the container from models does not prevent dependency injection. If you really need the container you can get it off routes or controllers. Otherwise you can use the store for looking up ember data stuff

igorT:
Not really, but you can get it off the store, and the store will always need it.

slindberg:
Sorry, I meant that it prevents lazy injections via `Ember.inject.service()`, [which requires the object to have a `container` property](https://github.com/emberjs/ember.js/blob/bfcc15ee3ac2b2b3e4dad6aa0d7a447936302f5a/packages/ember-metal/lib/injected_property.js#L31).

```js
  thingService: Ember.inject.service('thing')
});
```

PR #2996, opened 2015-04-13
Added failing test for hasMany metadata scenario

wecc:
Failing test for https://github.com/emberjs/data/pull/2878#discussion_r26906750

PR #2878, opened 2015-03-14
Has many metadata

wecc:

PR #3221, opened 2015-06-06
Metadata for hasmany

igorT:
Happy birthday @wecc 
Copied commit message:

Rebased #2878 and added a test from #2996
Added metadata support behind a feature flag, for hasMany relationships.
For now, until we refactor the internals, the store internally pushes hasMany
metadata around in the `meta` key next to `links` and `data`.
This might lead to overriding meta attribtues with meta relationships,
but will be solved by the upcoming change of the internals to JSONAPI format


wecc:
:+1:

PR #3226, opened 2015-06-06
Warnings for incorrect options in belongsTo, resolves #3187

### LINKED ISSUE #3187: If someone passes serialize:true or embedded:true to the relationship helper, warn them it's not the right place
If someone does something like
```
var User = DS.Model({
   friend: DS.belongsTo('friends', { serialize: true })
})
```
we should warn them

zoeesilcock:
Display a warning when the user specifies ```serialize``` or ```embedded``` in the options for ```belongsTo```. These settings belong in the serializer now, I have linked to the relevant documentation pages in order to help the user to do what they meant to do. Resolves #3187.

### DIFF
--- a/packages/ember-data/lib/system/relationships/belongs-to.js
+++ b/packages/ember-data/lib/system/relationships/belongs-to.js
@@ -102,6 +102,9 @@ function belongsTo(modelName, options) {
 
   return computedPolyfill({
     get: function(key) {
+      Ember.warn('You provided a serialize option on the "' + key + '" property in the "' + this._internalModel.modelName + '" class, this belongs in the serializer. See DS.Serializer and it\'s implementations http://emberjs.com/api/data/classes/DS.Serializer.html', !opts.hasOwnProperty('serialize'));
+      Ember.warn('You provided an embedded option on the "' + key + '" property in the "' + this._internalModel.modelName + '" class, this belongs in the serializer. See DS.EmbeddedRecordsMixin http://emberjs.com/api/data/classes/DS.EmbeddedRecordsMixin.html', !opts.hasOwnProperty('embedded'));
+
       return this._internalModel._relationships.get(key).getRecord();
     },
     set: function(key, value) {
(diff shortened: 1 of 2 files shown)


packages/ember-data/lib/system/relationships/belongs-to.js
```diff
@@ -93,6 +93,14 @@ function belongsTo(modelName, options) {
 
   opts = opts || {};
 
+  if (opts.hasOwnProperty('serialize')) {
+    Ember.warn('You provided a serialize option, this belongs in the serializer. See DS.Serializer and it\'s implementations http://emberjs.com/api/data/classes/DS.Serializer.html');
```
igorT: Can you tell the user for which model they provided it, that would make the warning much less confusing. Something like: In the `modelName` class,  passed a serialize option to the belongsTo relationship `key`. You most likely want to be configuring that in the serializer...

zoeesilcock: Good idea, I'll change the error message.

zoeesilcock: Sorry @igorT, I can't figure out how to provide the information you mentioned. I can only get ahold of the ```modelName``` which is the name of the model that the relationship is to. I can't find the name of the model or the property on where the ```belongsTo``` was used.

igorT: If you move the warning into the getter, you should be able to do this. Perf isn't as important as anything inside Ember.warn/assert etc. gets stripped in production

packages/ember-data/lib/system/relationships/belongs-to.js
```diff
@@ -102,6 +102,14 @@ function belongsTo(modelName, options) {
 
   return computedPolyfill({
     get: function(key) {
+      if (opts.hasOwnProperty('serialize')) {
```
igorT: Can you put the `opts.hasOwnProperty('serialize')` inside the warn, so it will get stripped in production?

PR #3233, opened 2015-06-06
raise errors on deprecation

fivetanley:
We shouldn't be committing code that causes deprecation warnings. :(

### DIFF
--- a/tests/ember-configuration.js
+++ b/tests/ember-configuration.js
@@ -14,6 +14,7 @@
 
   // Handle testing feature flags
   ENV['ENABLE_OPTIONAL_FEATURES'] = !!QUnit.urlParams.enableoptionalfeatures;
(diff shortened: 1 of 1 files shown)


rwjblue:
:+1: - But this will likely cause a bunch of issues when testing against canary of Ember during the ramp up to Ember 2.0. :disappointed:

pangratz:
> But this will likely cause a bunch of issues when testing against canary of Ember during the ramp up to Ember 2.0.

One (interim) solution could be to [allow failing builds in `.travis.yml`](http://docs.travis-ci.com/user/customizing-the-build/#Rows-that-are-Allowed-To-Fail) when testing against canary ...

bmac:
@fivetanley do you mind rebasing this pr. I suspect it will pass now.

bmac:
Superseded by https://github.com/emberjs/data/pull/3834

Issue #3227, opened 2015-06-06
`store.findAll(type)` should be made public


bmac:
store.[findAll](https://github.com/emberjs/data/blob/master/packages/ember-data/lib/system/store.js#L1000) is currently private. It should be made public and the docs should be updated to use this method.

`store.find(type)` should be deprecated and suggest the use of `store.findAll` instead.

`fetchAll` should be deprecated and suggest the use of `store.findAll` instead.

HeroicEric:
I'm on it

PR #3234, opened 2015-06-06
Make store.findAll(type) public

HeroicEric:
closes https://github.com/emberjs/data/issues/3227

- [x] Make `store.findAll` public
- [x] Make `adapter.findAll` public
- [x] Deprecate `store.find(type)` in favor of `store.findAll(type)`
- [x] Deprecate `store.fetchAll(type)` in favor of `store.findAll(type)`

### DIFF
--- a/packages/ember-data/lib/system/store.js
+++ b/packages/ember-data/lib/system/store.js
@@ -462,10 +462,10 @@ Store = Service.extend({
 
     ---
 
-    To find all records for a type, call `find` with no additional parameters:
+    To find all records for a type, call `findAll`:
 
     ```javascript
-    store.find('person');
+    store.findAll('person');
     ```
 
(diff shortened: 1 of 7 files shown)


HeroicEric:
this is for https://github.com/emberjs/data/issues/3227

packages/ember-data/lib/system/store.js
```diff
@@ -584,9 +585,8 @@ Store = Service.extend({
   */
   fetchAll: function(modelName) {
     Ember.assert('Passing classes to store methods has been removed. Please pass a dasherized string instead of '+ Ember.inspect(modelName), typeof modelName === 'string');
```
sly7-7: maybe this assert could be removed, (it's in findAll as well)

bmac:
:+1:

PR #3235, opened 2015-06-06
rename store.findQuery to store.query

### LINKED ISSUE #3229: Add store.query method
Rename [store.findQuery](https://github.com/emberjs/data/blob/master/packages/ember-data/lib/system/store.js#L980) to `query`. 

Update internal references to `findQuery` and make sure they use the new `query` method.

Update [store.find(type, {query: 'object'})](https://github.com/emberjs/data/blob/master/packages/ember-data/lib/system/store.js#L534) to log a deprecation warning that recommends the use of `store.query`.

Add a new `store.findQuery` method which logs a deprecation warning and delegates to store.query.

csantero:
Closes #3229

### DIFF
--- a/packages/ember-data/lib/system/store.js
+++ b/packages/ember-data/lib/system/store.js
@@ -40,7 +40,7 @@ import {
   _findHasMany,
   _findBelongsTo,
   _findAll,
-  _findQuery
+  _query
 } from "ember-data/system/store/finders";
 
 import coerceId from "ember-data/system/coerce-id";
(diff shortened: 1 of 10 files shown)


packages/ember-data/lib/system/store.js
```diff
@@ -983,16 +943,38 @@ Store = Service.extend({
...
+    ```
+    Started GET "/api/v1/person?page=1"
+    Processing by Api::V1::PersonsController#index as HTML
+    Parameters: {"page"=>"1"}
+    ```
+
+    If you do something like this:
+
+    ```javascript
+    store.find('person', {ids: [1, 2, 3]});
```
HeroicEric: I think you want to use `store.query` here

packages/ember-data/lib/system/store.js
```diff
@@ -983,16 +943,38 @@ Store = Service.extend({
...
+    Processing by Api::V1::PersonsController#index as HTML
+    Parameters: {"ids"=>["1", "2", "3"]}
+    ```
+
     This method returns a promise, which is resolved with a `RecordArray`
     once the server returns.
 
-    @method findQuery
+    @method query
     @private
```
HeroicEric: I think this should not be private if it's going to be recommended to be used

packages/ember-data/lib/system/store.js
```diff
@@ -1003,7 +984,30 @@ Store = Service.extend({
...
+
+    Exposing queries this way seems preferable to creating an abstract query
+    language for all server-side queries, and then require all adapters to
+    implement them.
+
+    This method returns a promise, which is resolved with a `RecordArray`
+    once the server returns.
+
+    @method query
+    @private
```
fivetanley: not that it matters too much for a deprecated method but i think this should have been public in the past.

fivetanley:
@csantero this looks good to me, can you make sure the guides get updated on the website repository? https://github.com/emberjs/data

csantero:
@fivetanley Removed the `@private` from the deprecated method, and submitted a [guides PR](https://github.com/emberjs/guides/pull/345).

fivetanley:
thanks!

Issue #3231, opened 2015-06-06
Deprecation warnings from RESTSerializer in canary


bmac:
In Canary I'm seeing the RESTSerialzer trigger the "Passing classes to store methods has been removed. Please pass a dasherized string instead of App.Author" deprecation warning. 

http://emberjs.jsbin.com/gexuhevefi/3

The offending line is 
https://github.com/emberjs/data/blob/43710498765c02f6dc2ad65f43003811c51978ad/packages/ember-data/lib/serializers/rest-serializer.js#L454

It should be passing `modelName` instead of `type`.

wecc:
I thought I fixed this :confused:

PR #3236, opened 2015-06-07
Use dasherized modelName to fetch serializer

HeroicEric:
fixes https://github.com/emberjs/data/issues/3231


fivetanley:
thanks!

PR #3237, opened 2015-06-07
Use Map.prototype.delete instead of Map.prototype.remove

HeroicEric:
`Map.prototype.remove` is deprecated and causes:

```
DEPRECATION: Calling `Map.prototype.remove` has been deprecated, please use `Map.prototype.delete` instead.
```

https://github.com/emberjs/ember.js/commit/39bf8b23a8482eb9fffeda00d8beb914c3b8c444


fivetanley:
thanks!

PR #3242, opened 2015-06-07
Always use dasherized modelName to fetch serializer

HeroicEric:
Found two more instances of https://github.com/emberjs/data/issues/3231

This brings the deprecations that come up when running the tests from 70 to 3 :smiley:


fivetanley:
:cake:

PR #3247, opened 2015-06-07
Dont remove deleted but uncommited records from record arrays

ianstarz:
Before this commit records were deleted from record arrays just after
deleting them, but before they were committed. This can be an issue if
you want to keep records visible in the UI until they're actually
removed.

### DIFF
--- a/packages/ember-data/lib/system/record-array-manager.js
+++ b/packages/ember-data/lib/system/record-array-manager.js
@@ -61,6 +61,9 @@ export default Ember.Object.extend({
   updateRecordArrays: function() {
     forEach.call(this.changedRecords, function(record) {
       if (record.isDeleted()) {
+      // TODO: it can be refactored after #2862 && #2859 are closed
+      // if (get(record, 'isDestroyed') || get(record, 'isDestroying') ||
+      //     (get(record, 'isDeleted') && !get(record, 'isDirty'))) {
         this._recordWasDeleted(record);
       } else {
         this._recordWasChanged(record);
--- a/packages/ember-data/lib/system/record-array-manager.js
+++ b/packages/ember-data/lib/system/record-array-manager.js
@@ -61,9 +61,16 @@ export default Ember.Object.extend({
   updateRecordArrays: function() {
     forEach.call(this.changedRecords, function(record) {
       if (record.isDeleted()) {
-      // TODO: it can be refactored after #2862 && #2859 are closed
+
+      // master ^
+
+      // ianstarz' attempt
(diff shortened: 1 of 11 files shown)


packages/ember-data/lib/system/record-array-manager.js
```diff
@@ -60,7 +60,8 @@ export default Ember.Object.extend({
   */
   updateRecordArrays: function() {
     forEach(this.changedRecords, function(record) {
-      if (record.isDeleted()) {
+      // TODO: it can be refactored after #2862 && #2859 are closed
+      if (get(record, 'record.isDestroyed') || get(record, 'record.isDestroying') || (record.isDeleted() && !record.isDirty())) {
```
ianstarz: @igor @drogus I'm still having trouble figuring out how to form this conditional, but only 4 failing tests now...

ianstarz:
There's still 4 failing tests for this, but otherwise a lot of the work to rebase and cleanup/fix other tests is done. Taking a break now, so I just wanted to open this and update you guys.

screen shot 2015-06-07 at 15 30 33

ianstarz:
This is the updated PR for the work done in #2867

packages/ember-data/lib/system/record-array-manager.js
```diff
@@ -61,6 +61,16 @@ export default Ember.Object.extend({
...
+
+      // master ^
+
+      // ianstarz' attempt
+      // if (get(record, 'isDestroyed') || get(record, 'isDestroying') ||
+      //     (get(record, 'isDeleted') && !get(record, 'isDirty'))) {
+
+      // drogus' attempt
+      // if (get(record, 'record.isDestroyed') || get(record, 'record.isDestroying') || (record.isDeleted() && !record.isDirty())) {
+
```
ianstarz: @igorT here is where I was talking about

bmac:
Hi @ianstarz. Thanks for this work. I ended up taking a stab at this too. https://github.com/emberjs/data/pull/3539

For the `updateRecordArrays` code in `record-array-manager.js` I ended up checking if the record was in the `root.deleted.saved` state to determine if the record should be removed from the record array.


I also left the `disconnectRelationships`/`reconnectRelationships` on the model and moved the places where it was called such that it is now only called when the record transitions into the deleted.saved state.

ianstarz:
@bmac it seemed like inspecting the state path was the right way to go. Thanks for picking this up! I'm still getting familiar with the code base, and was feeling pretty stuck.

Issue #3223, opened 2015-06-06
Convert internal tests to use json api format for `store.push`


igorT:
Everywhere where we do,
`store.push(type, data)` in tests, we need to convert to `store.push(jsonApiFormat)`
We can also break this up in multiple PRs and merge to a branch.

igorT:
Instead of pushing something like
```
store.push('user', { 
  id:1, 
  name: 'bmac', 
  friends: [1,2,3], 
  links: { 
    other: 'hello' 
  }
})
```
should push

```
```
store.push('user', { 
  id:1, 
  name: 'bmac', 
  locations: [1,2,3], 
  links: { 
    other: 'hello' 
  }
})
```
store.push({ 
  id:1, 
  type: 'user',
  attributes: {
    name: 'bmac'
  },
  relationships: { 
    friends: {
       data: [{ type: 'location', id:1}, { type: 'location', id:2}, { type: 'location', id:3}]
     },
     other: {
       links: { 
         related: 'hello' 
       }
    }
})
```

igorT:
Will need to temporarily comment out  `Ember.assert('Passing classes to store methods has been removed. Please pass a dasherized string instead of '+ Ember.inspect(modelName), typeof modelName === 'string');`

fivetanley:
Why?

Serabe:
Sorry, I've been in a plane for most of my weekend. I see @stevenwu has started to tackle the issue. If it is too big for one person and you decide to split it in several, or if @stevenwu needs some help, just let me know.

PR #3251, opened 2015-06-08
Convert unload test to json api format for `store.push`.. #3223

andrejunges:
This PR will break some tests, but now the the calls to store.push are
using jsonApiFormat

### DIFF
--- a/packages/ember-data/tests/integration/filter-test.js
+++ b/packages/ember-data/tests/integration/filter-test.js
@@ -16,8 +16,34 @@ var shouldNotContain = function(array, item) {
 
 module("integration/filter - DS.Model updating", {
   setup: function() {
-    array = [{ id: 1, name: "Scumbag Dale", bestFriend: 2 }, { id: 2, name: "Scumbag Katz" }, { id: 3, name: "Scumbag Bryn" }];
-    Person = DS.Model.extend({ name: DS.attr('string'), bestFriend: DS.belongsTo('person', { inverse: null, async: false }) });
+    array = [{
+      id: '1',
+      type: 'person',
+      attributes: {
+        name: 'Scumbag Dale'
+      },
+      relationships: {
+        bestFriend: {
+          data: {
+            id: '2',
+            type: 'person'
+          }
+        }
+      }
+    }, {
+      id: '2',
+      type: 'person',
+      attributes: {
+        name: 'Scumbag Katz'
+      }
+    }, {
+      id: '3',
+      type: 'person',
+      attributes: {
+        name: 'Scumbag Bryn'
+      }
+    }];
+    Person = DS.Model.extend({ name: DS.attr('string'), bestFriend: DS.belongsTo('person', { inverse: null }) });
 
     env = setupStore({ person: Person });
     store = env.store;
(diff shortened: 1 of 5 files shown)


packages/ember-data/tests/integration/filter-test.js
```diff
@@ -18,7 +18,7 @@ var shouldNotContain = function(array, item) {
 
 module("integration/filter - DS.Model updating", {
   setup: function() {
-    array = [{ id: 1, name: "Scumbag Dale", bestFriend: 2 }, { id: 2, name: "Scumbag Katz" }, { id: 3, name: "Scumbag Bryn" }];
+    array = [{ "id": "1", "type": "person", "attributes": { "name": "Scumbag Dale", "bestFriend": "2" } }, { "id": "2", "type": "person", "attributes": { "name": "Scumbag Katz" } }, { "id": 3, "type": "person", "attributes": { "name": "Scumbag Bryn" } }];
```
wecc: I think bestFriend should be a relationship here

wecc:
Would you do squashing the commits and remove "WIP"?

packages/ember-data/tests/integration/filter-test.js
```diff
@@ -16,8 +16,34 @@ var shouldNotContain = function(array, item) {
...
   setup: function() {
-    array = [{ id: 1, name: "Scumbag Dale", bestFriend: 2 }, { id: 2, name: "Scumbag Katz" }, { id: 3, name: "Scumbag Bryn" }];
-    Person = DS.Model.extend({ name: DS.attr('string'), bestFriend: DS.belongsTo('person', { inverse: null, async: false }) });
+    array = [{
+      "id": "1",
+      "type": "person",
+      "attributes": {
+        "name": "Scumbag Dale"
+      },
+      relationships: {
```
wecc: Could you encapsulate all of the keys in quotes, for consistency?

packages/ember-data/tests/integration/filter-test.js
```diff
@@ -16,8 +16,34 @@ var shouldNotContain = function(array, item) {
...
+        }
+      }
+    }, {
+      "id": "2",
+      "type": "person",
+      "attributes": {
+        "name": "Scumbag Katz"
+      }
+    }, {
+      "id": 3,
```
wecc: Nitpicking here but as we want to conform to JSON-API as closely as possible IDs should be strings (http://jsonapi.org/format/#document-resource-object-identification).

There are a couple of places below too where IDs are numbers.

wecc:
Can you prefix the commit with [CLEANUP beta]?

packages/ember-data/tests/integration/filter-test.js
mixonic: Perhaps the quoted keys should be unquoted instead @wecc?

packages/ember-data/tests/integration/filter-test.js
mixonic: Staying really close to JSON-API seems very important. There is great power in just telling people to check the spec when they have a question about formatting :clap:

wecc:
Thank you so much!

pangratz:
:+1: Awesome work @andrejunges!

PR #2352, opened 2014-10-06
check for empty string

calvinmetcalf:
treats empty strings in ids the same as null or undefined, WIP needs tests

PR #3253, opened 2015-06-08
Treat empty strings in ids the same as null or undefined

pangratz:
Treats empty strings in ids the same as null or undefined.

---

This is the rebased version of #2352


igorT:
Thanks!

pangratz:
Fixing the failing tests ...

PR #3209, opened 2015-06-05
Rename and deprecate finders (store.all -> store.peekAll and store.getById -> store.peekRecord)

HeroicEric:
closes #3205

PR #3254, opened 2015-06-08
prefix private methods with underscore - store.js

### LINKED ISSUE #3250: prefix private methods with underscore - store.js
Steps to fixing this issue:

1) deprecate the method. You can use `Ember.deprecate('serialize was documented as private and will be removed in the next version of Ember Data.')`
2) move the method to an underscored method (for example `_serialize`) and update usages in the codebase

If the item is labeled in parenthesis as `(public)`, the `@private` yuidoc tag should be removed. Otherwise, if it is `private`, take the steps above to deprecate and rename the method.
- [ ] packages/ember-data/lib/system/store.js
  - [ ] init (private, but not sure how to document)
  - [ ] serialize (private)
  - [ ] findById (public)
  - [ ] findByIds (public? ask @igorT)
  - [ ] fetchRecord (private)
  - [ ] reloadRecord (private)
  - [ ] recordForId (deprecate?? ask @bmac)
  - [ ] findMany (private)
  - [ ] findHasMany (private)
  - [ ] findBelongsTo (private)
  - [ ] findAll (public)
  - [ ] didUpdateAll (private)
  - [ ] dataWasUpdated (private)
  - [ ] scheduleSave (private)
  - [ ] flushPendingSave (private)
  - [ ] didSaveRecord (private)
  - [ ] recordWasInvalid (private)
  - [ ] recordWasError (private)
  - [ ] updateId (private)
  - [ ] typeMapFor (private)
  - [ ] buildRecord (private)
  - [ ] dematerializeRecord (should be public, but is deprecated)
  - [ ] adapterFor (should be public - fivetanley)
  - [ ] serializerFor (should be public - fivetanley)
  - [ ] retrieveManagedInstance (should be private - fivetanley)

gevious:
Closes #3250

### DIFF
--- a/packages/ember-data/lib/system/store.js
+++ b/packages/ember-data/lib/system/store.js
@@ -263,9 +263,12 @@ Store = Service.extend({
     @param {Object} options an options hash
   */
   _serialize: function(record, options) {
-    Ember.deprecate('_serialize was documented as private and will be removed in the next version of Ember Data.');
     var snapshot = record._internalModel.createSnapshot();
     return snapshot.serialize(options);
+  },
+  serialize: function(record, options) {
+    Ember.deprecate('serialize was documented as private and will be removed in the next version of Ember Data.');
+    return this._serialize(record, options);
   },
 
   /**
@@ -576,8 +579,11 @@ Store = Service.extend({
   */
   _findById: function(modelName, id, preload) {
     Ember.deprecate('Using store._findById() has been deprecated. Use store.findRecord() to return a record for a given type and id combination.');
-    Ember.deprecate('_findById was documented as private and will be removed in the next version of Ember Data.');
     return this.findRecord(modelName, id, preload);
+  },
+  findById: function(modelName, id, preload) {
+    Ember.deprecate('findById was documented as private and will be removed in the next version of Ember Data.');
+    return this._findById(modelName, id, preload);
   },
 
   /**
@@ -623,13 +629,16 @@ Store = Service.extend({
     @return {Promise} promise
   */
   _findByIds: function(modelName, ids) {
-    Ember.deprecate('_findByIds was documented as private and will be removed in the next version of Ember Data.');
     Ember.assert('Passing classes to store methods has been removed. Please pass a dasherized string instead of '+ Ember.inspect(modelName), typeof modelName === 'string');
     var store = this;
 
     return promiseArray(Ember.RSVP.all(map(ids, function(id) {
       return store.findRecord(modelName, id);
     })).then(Ember.A, null, "DS: Store#_findByIds of " + modelName + " complete"));
+  },
+  findByIds: function(modelName, ids) {
+    Ember.deprecate('_findByIds was documented as private and will be removed in the next version of Ember Data.');
+    return this._findByIds(modelName, ids);
   },
 
   /**
@@ -643,7 +652,6 @@ Store = Service.extend({
     @return {Promise} promise
   */
   _fetchRecord: function(internalModel) {
-    Ember.deprecate('_fetchRecord was documented as private and will be removed in the next version of Ember Data.');
     var typeClass = internalModel.type;
     var id = internalModel.id;
     var adapter = this.adapterFor(typeClass.modelName);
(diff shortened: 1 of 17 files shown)


gevious:
any way I can reproduce these failures locally? `npm test` ran without erros

gevious:
My tests pass sporadically. I get some issues relating to semaphore every now and then. I'm not sure what to do about that.

packages/ember-data/lib/system/store.js
```diff
@@ -1377,11 +1391,12 @@ Store = Service.extend({
...
     is rejected (with anything other than a `DS.InvalidError`).
 
-    @method recordWasError
+    @method _recordWasError
     @private
     @param {InternalModel} internalModel
   */
-  recordWasError: function(internalModel) {
+  _recordWasError: function(internalModel) {
+    Ember.deprecate('_recordWasError was documented as private and will be removed in the next version of Ember Data.');
```
fivetanley: We want to keep the underscore method working, and create a new method called `recordWasError` that calls the previous one:

```javascript
recordWasError: function(internalModel) {
  Ember.deprecate('message');
  return this._recordWasError(internalModel);
}
```

fivetanley:
@gevious It seems my instructions weren't clear, sorry. When we deprecate a method, we still want people to be able to call it until it gets removed. I left an example here https://github.com/emberjs/data/pull/3254/files#r31911784

gevious:
Ah, yes. I clearly wasn't thinking this morning. What should I do with the tests, @fivetanley. I've updated them for the new methods. Should I create parallel tests for the deprecated ones?

gevious:
I've kept the tests the same, but reinserted the deprecated methods. This means the deprecated methods aren't used in the test suite, but the new private methods are.If the new api is used internally, this shouldn't cause any problems going forward.

packages/ember-data/lib/system/store.js
```diff
@@ -566,17 +570,21 @@ Store = Service.extend({
...
     @return {Promise} promise
   */
-  findById: function(modelName, id, preload) {
-    Ember.deprecate('Using store.findById() has been deprecated. Use store.findRecord() to return a record for a given type and id combination.');
+  _findById: function(modelName, id, preload) {
+    Ember.deprecate('Using store._findById() has been deprecated. Use store.findRecord() to return a record for a given type and id combination.');
     return this.findRecord(modelName, id, preload);
   },
+  findById: function(modelName, id, preload) {
+    Ember.deprecate('findById was documented as private and will be removed in the next version of Ember Data.');
```
fivetanley: I think we want to change this deprecation message to the one in _findById.

packages/ember-data/lib/system/store.js
```diff
@@ -850,25 +870,29 @@ Store = Service.extend({
...
     @param {String} modelName
     @param {(String|Integer)} id
     @return {DS.Model} record
   */
-  recordForId: function(modelName, id) {
+  _recordForId: function(modelName, id) {
     Ember.assert('Passing classes to store methods has been removed. Please pass a dasherized string instead of '+ Ember.inspect(modelName), typeof modelName === 'string');
     return this._internalModelForId(modelName, id).getRecord();
   },
+  recordForId: function(modelName, id) {
```
fivetanley: @bmac what should people be using here instead?

bmac: `getById` / soon to be `peekRecord`

gevious:
I've made the changes. I've included a deprecation warning for `recordForId` to point to `getById`. I think we can always change that once the RFC is approved and `peekRecord` is official.

Also, I've left `init` as is. I wasn't sure what to do with it. I'm happy to change it if need be.

HeroicEric:
@gevious `peekRecord` is now in master https://github.com/emberjs/data/pull/3209

bmac:
Hi @gevious. Thanks for doing all the work on this pr. 

Unfortunately, I'm going to close it because since Ember Data 2.0 is released, I think we missed the opportunity to make this change in a semver compatible way.

Feel free to re-open if you disagree.

PR #3262, opened 2015-06-08
Pass _internalModel into a record when it is created

### LINKED ISSUE #3255: _internalModel is undefined on 'init' of a record.
Testcase: http://jsbin.com/cegapapigo/2/edit?html,css,js,output

```js
DS.Model.extend({
  foo: DS.attr('string'),
  initFoo: Ember.on('init', function() {
    this.get('foo');
  })
})._create();
```

bmac:
Closes #3255.

### DIFF
--- a/packages/ember-data/tests/unit/model-test.js
+++ b/packages/ember-data/tests/unit/model-test.js
@@ -869,3 +869,23 @@ test('toJSON looks up the JSONSerializer using the store instead of using JSONSe
   ok(!errorThrown, 'error not thrown due to missing store');
   deepEqual(json, {});
 });
+
+
+test('accessing attributes in the initializer should not throw an error', function() {
+  expect(1);
+  var Person = DS.Model.extend({
+    name: DS.attr('string'),
+
+    init: function() {
+      this._super.apply(this, arguments);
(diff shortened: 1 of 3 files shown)


igorT:
Any way we could have a test please?

bmac:
Will do

bmac:
I added a test. @igorT let me know if there is a better way to test this without directly asserting around internals.

packages/ember-data/tests/unit/model-test.js
```diff
@@ -869,3 +869,23 @@ test('toJSON looks up the JSONSerializer using the store instead of using JSONSe
...
+
+
+test('_internalModel should be set by the store when a record is created', function() {
+  expect(1);
+  var Person = DS.Model.extend({
+    name: DS.attr(),
+
+    init: function() {
+      this._super.apply(this, arguments);
+      ok(this._internalModel);
```
igorT: ok(this.get('name')) is the thing we actually care about

bmac: updated

igorT:
cc @fivetanley for beta.19.1

Issue #3171, opened 2015-06-03
Implement `filterBy` on results of `store.find` that listens for changes and recomputes


igorT:

wecc:
#3263 was reverted in #3648

From https://github.com/emberjs/data/pull/3263#issuecomment-129531675:

> We talked about this issue in a recent team meeting and it sounds like this is a regression that we want to fix. The current fixing is we will back out the filterBy changes and replace it with a new name or a more limited API.

Unsure if this has been re-implemented with another name? ping @bmac @pangratz

pangratz:
This has been implemented in an addon https://github.com/ember-data/ember-data-live-filter-by

PR #3263, opened 2015-06-08
Add `RecordArray#filterBy` which contains a live, filtered subset

pangratz:
This method takes the parameters `key` and an optional `value` and
returns an updating list of records of the `RecordArray`, which'
property value of `key` matches `value`.

---

This addresses #3171

### DIFF
--- a/packages/ember-data/lib/system/record-arrays/record-array.js
+++ b/packages/ember-data/lib/system/record-arrays/record-array.js
@@ -8,6 +8,21 @@ import SnapshotRecordArray from "ember-data/system/snapshot-record-array";
 var get = Ember.get;
 var set = Ember.set;
 
+var FilteredSubset = Ember.ArrayProxy.extend({
+  init: function() {
+    this._super(...arguments);
+
+    var { filterByArgs, recordArray } = this.getProperties('filterByArgs', 'recordArray');
+    var [key] = filterByArgs;
+
+    var path = `recordArray.@each.${key}`;
+    Ember.defineProperty(this, 'content', Ember.computed(path, function() {
+      return this.filterBy.apply(recordArray, filterByArgs);
+    }));
+  }
+});
+
+
 /**
   A record array is an array that contains records of a certain type. The record
   array materializes records as needed when they are retrieved for the first
@@ -94,6 +109,44 @@ export default Ember.ArrayProxy.extend(Ember.Evented, {
     var content = get(this, 'content');
     var internalModel = content.objectAt(index);
     return internalModel && internalModel.getRecord();
+  },
+
+  /**
+    Get a filtered subset of the underlying `RecordArray`.
+    The subset updates when a record would match or mismatch the
+    specified filter parameters.
+
+    Example
+
+    ```javascript
+    var allToms = store.all('person').filterBy('name', 'Tom');
+
+    allToms.get('length'); // 0, since no toms yet in store
+
+    var tom = store.push('person', { id: 1, name: 'Tom' });
+    allToms.get('length'); // Tom is added
+
+    tom.set('name', 'Thomas');
+    allToms.get('length'); // 0, since no more records with name === 'Tom'
+    ```
+
+    @method filterBy
+    @param {String} key property path
+    @param {*} value optional
+
+  */
+  filterBy: function(key, value) {
+    // only pass value to the arguments if it is present; this mimics the same
+    // behavior for `filterBy`: http://git.io/vIurH
+    var filterByArgs = [key];
+    if (arguments.length === 2) {
+      filterByArgs.push(value);
+    }
+
+    return FilteredSubset.create({
+      filterByArgs,
+      recordArray: this
+    });
   },
 
   /**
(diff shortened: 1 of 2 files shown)


packages/ember-data/lib/system/record-arrays/record-array.js
```diff
@@ -95,6 +95,47 @@ export default Ember.ArrayProxy.extend(Ember.Evented, {
...
+  filterBy: function(key, value) {
+    var content = get(this, 'content');
+
+    var FilteredRecordArray = Ember.ArrayProxy.extend({
+      // since content is an array of InternalModel's and we need to observe
+      // the property at path `key`, the records are mapped to
+      // `internalRecords`, so we can use the `Ember.computed.filterBy` macro,
+      // which listens to changes of this property
+      internalRecords: Ember.computed.mapBy('content', 'record'),
+      arrangedContent: Ember.computed.filterBy('internalRecords', key, value)
```
pangratz: I'm very unsure about this implementation, since `record` on the `InternalModel` should be accessed via [`getRecord()`](http://git.io/vIEk3) I guess...

igorT: Ember.computed.filterBy('internalRecords') should be filterBy content

packages/ember-data/lib/system/record-arrays/record-array.js
```diff
@@ -95,6 +95,47 @@ export default Ember.ArrayProxy.extend(Ember.Evented, {
...
+
+    @method filterBy
+    @param {String} key property path
+    @param {*} value optional
+
+  */
+  filterBy: function(key, value) {
+    var content = get(this, 'content');
+
+    var FilteredRecordArray = Ember.ArrayProxy.extend({
```
igorT: Can we call it something else, as FilteredRecordArray is already a thing

packages/ember-data/lib/system/record-arrays/record-array.js
pangratz: I tried that, but then `store.all('person').filterBy('name', xxx)` is always an empty array. So I guess observing `InternalModel`'s and properties on them doesn't work as expected yet ...

packages/ember-data/lib/system/record-arrays/record-array.js
pangratz: :+1: haven't thought about that

pangratz: `UpToDateFilteredSubset` ?

pangratz:
@igorT I am not sure about naming the method `filterBy`, since the `RecordArray#filterBy` method (inherited by `ArrayProxy`) is hereby overwritten with a different implementation: an auto-updating filtered subset. Maybe it should be renamed into something like `updatingFilterBy`, `liveFilterBy` or so ...

cibernox:
IMO the name it's ok as it is. 
Seems reasoable to me that the `filterBy` method of a live collection returns another live collection.

fivetanley:
Can we move the ArrayProxy class definition out of the function? It doesn't need to be in its own file (might be nice), but it seems like it could get expensive to create a new class every time a filter is created.

pangratz:
@fivetanley done

bmac:
Looks good. Should this also be implemented on a `ManyArray`?

bmac:
Thanks @pangratz

bendemboski:
This change seems problematic to me.

First of all, it's a breaking change. Previously a call to ```filterBy``` on a ```RecordArray``` would call ```ArrayProxy```'s ```filterBy``` method which returns a native array. So, the following code would behave as expected and would call ```doSomething```:

```javascript
let recordArray = getRecordArraySomehow(),
    newRecords = recordArray.filterBy('isNew');
if (newRecords.length > 0) {
    doSomething();
}
```

After this update, the same code will not run ```doSomething``` because ```newRecords``` is no longer a native array, so ```length``` is a computed property and ```newRecords.length``` is not an integer, so ```newRecords.length > 0``` will always be false.

Now before updating to Ember Data 1.13.8 I have to audit all of my usages of ```RecordArrays``` to make sure I don't treat the return value like a native array anywhere :(

Secondly, it seems wrong to override ```filterBy``` but not override ```filter```. This means that ```recordArray.filterBy('isNew')``` will return an ArrayProxy but ```recordArray.filter(function (record) { return record.get('isNew'); })``` will return a native array. Same inconsistency applies to ```reject``` and to ```rejectBy```.

Perhaps the decision to name this method ```filterBy``` could be re-visited? Or some kind of warning could be added to the release notes (I know it's already released...)?

bmac:
Hi @bendemboski. Thanks for bringing up this issue. We talked about this issue in a recent team meeting and it sounds like this is a regression that we want to fix. The current fixing is we will back out the `filterBy` changes and replace it with a new name or a more limited API.

cibernox:
I've also faced the same. 
I was chaining `this.get('manyRelation').filterBy('isNew', false).sort(compareFn)` and that broke too, since there is no `sort` function in this new kind of liveArray.

I'm explicitly converting the relation to an array before filter/sort `this.get('manyRelation').toArray().filt....`

Issue #3248, opened 2015-06-07
Add adapterOptions hash to find methods


igorT:
Make sure to deprecate the passing of preloaded to find right now, and to move preloaded into a preloaded key.

PR #3264, opened 2015-06-08
Deprecate the preload argument to `find`/`fetchById` and move in into the `…

bmac:
…preload` key on the options argument.

Closes https://github.com/emberjs/data/issues/3248

### DIFF
--- a/packages/ember-data/lib/system/store.js
+++ b/packages/ember-data/lib/system/store.js
@@ -445,7 +445,7 @@ Store = Service.extend({
     without fetching the post you can pass in the post to the `find` call:
 
     ```javascript
-    store.find('comment', 2, { post: 1 });
+    store.find('comment', 2, { preload: { post: 1 } });
     ```
 
     If you have access to the post model you can also pass the model itself:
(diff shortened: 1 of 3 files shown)


packages/ember-data/lib/system/store.js
```diff
@@ -584,21 +584,43 @@ Store = Service.extend({
...
+    if (options) {
+      var modelProperties = [];
+      var fields = Ember.get(internalModel.type, 'fields');
+      fields.forEach(function(fieldType, key) {
+        modelProperties.push(key);
+      });
+      var preloadDetected = modelProperties.reduce(function(memo, key) {
+        return typeof options[key] !== 'undefined' || memo;
+      }, false);
+      if (preloadDetected) {
```
igorT: This should not be on `findRecord` as findRecord is a new api, we should add the deprecation only on `find` and `findById` if findById was public

bmac: Updated.

igorT:
Rebase please? :(

bmac:
Rebased

igorT:
Can you please add a test that confirms that the deprecated form still works?

PR #3267, opened 2015-06-09
Pad single-line objects with white-space

HeroicEric:
https://github.com/emberjs/ember.js/blob/master/STYLEGUIDE.md#objects

I think it's important, especially in the docs/examples, to provide consistency.


igorT:
Thanks!

PR #3272, opened 2015-06-09
Convert internals to json api - WIP needs cleanup

igorT:
We will want to refactor this once we are in 2.0 land and don't care about backcompat

### DIFF
--- a/packages/ember-data/lib/system/store.js
+++ b/packages/ember-data/lib/system/store.js
@@ -27,7 +27,7 @@ import {
 import {
   convertResourceObject,
   normalizeResponseHelper,
-  pushPayload
+  _normalizeSerializerPayload
 } from "ember-data/system/store/serializer-response";
 
 import {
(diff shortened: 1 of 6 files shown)


packages/ember-data/lib/system/store.js
```diff
@@ -1608,9 +1611,21 @@ Store = Service.extend({
     @return {DS.Model|Array} the record(s) that was created or
       updated.
   */
-  push: function(modelName, data) {
+  push: function(modelNameArg, dataArg) {
+    var data, modelName;
+
+    if (Ember.typeOf(modelNameArg) === 'object' && Ember.typeOf(dataArg) === 'undefined') {
+      data = modelNameArg;
+      modelName = data.data.type;
```
wecc: Couldn't `data.data` be an array here?

igorT: Yes, but not in the backcompat mode. Need some tests for that case

bmac:
This landed as https://github.com/emberjs/data/pull/3365

PR #3279, opened 2015-06-10
Deprecation if id is Model in deserializeRecordId

### LINKED ISSUE #3271: Deprecate pushing real models in relationships into store.push
https://github.com/emberjs/data/blob/master/packages/ember-data/lib/system/store.js#L2059

this should deprecate

joelalejandro:
Fixes #3271

### DIFF
--- a/packages/ember-data/lib/system/store.js
+++ b/packages/ember-data/lib/system/store.js
@@ -2058,6 +2058,7 @@ function deserializeRecordId(store, data, key, relationship, id) {
 
   //If record objects were given to push directly, uncommon, not sure whether we should actually support
   if (id instanceof Model) {
+    Ember.deprecate("Passing a record object to push directly into the store has been deprecated. Pass a number or string as a value to represent the record.");
     data[key] = id._internalModel;
     return;
   }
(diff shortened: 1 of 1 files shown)


packages/ember-data/lib/system/store.js
```diff
@@ -2058,6 +2058,7 @@ function deserializeRecordId(store, data, key, relationship, id) {
 
   //If record objects were given to push directly, uncommon, not sure whether we should actually support
   if (id instanceof Model) {
+    Ember.deprecate("Passing a record object to push directly into the store has been deprecated. Pass a number or string as a value to represent the record.");
```
igorT: Can you make the deprecation message a bit more descriptive to the user. Think about if you were to see it pop up in the console. Something with more context like:
You tried pushing a record 'recordType' with id 'id'  and passed a DS.Model as a value for the relationship 'relationshipKey' etc.

igorT:
thanks!

packages/ember-data/lib/system/store.js
joelalejandro: Sure, I'll change it in a bit.

bmac:
:+1:

igorT:
@joelalejandro Can you add a test for this please

joelalejandro:
@igorT sure, I'll get to it.

igorT:
We already did it, thanks though!

joelalejandro:
Great @igorT, was out of town during the weekend  ^^'

PR #3278, opened 2015-06-09
Add a way to pass options to an adapter 

bmac:
`DS.Store#findAll`, `DS.Store#findRecord`, `DS.Store#query`, `DS.Model#save` and `DS.Model#destroyRecord` now all take an options parameter. If this `options` parameter has a value on the `adapeterOptions` key it will be passed to the adapter method used to retrieve or update from the backend.

This pr currently include the code from #3264 because that is a prerequisites for this work. I will rebase this pr after #3264 is merged.

This pr is built on top of the work @knownasilya started in pr https://github.com/emberjs/data/pull/2621

PR #3281, opened 2015-06-10
Add a way to pass options to an adapter (rebase)

HeroicEric:
rebasing https://github.com/emberjs/data/pull/3278


igorT:
Maybe we should put the options on the snapshot, otherwise not sure how coalescing will work

bmac:
Would options go on the snapshot and be passed as an argument into the adapter's `find` method?

igorT:
yes, see also the discussion in slack

HeroicEric:
@bmac should I close this now since I see https://github.com/emberjs/data/commit/059ec88cbc3a71a698f7a8952e2e8022d927c36e?

bmac:
yes, thanks

Issue #3220, opened 2015-06-06
Make async relationships default


igorT:
In current ember data throw a deprecation on any relationship that is not `async:false`. Write an Ember Watson helper to add `{async:false}` to each one

igorT:
@tomdale @wycats **we are keeping the sync relationships, but changing the defaults

abuiles:
@igorT just released https://github.com/abuiles/ember-watson/releases/tag/v0.5.1 to help with this, thanks to @Serabe

Serabe:
This is still open. Anything I can help with?

sly7-7:
Hm, I'm closing since https://github.com/emberjs/data/pull/3488 has been merged. 
@Serabe Thanks for the heads up :+1:

sly7-7:
Well, I thinks this could probably be closed, the Ember Watson exists, and https://github.com/emberjs/data/pull/3488 has been merged. 
@igorT c/d ?

bmac:
I agree @sly7-7 this issue can be closed.

PR #3283, opened 2015-06-10
Add deprecation to relationships without { async: false }

HeroicEric:
closes https://github.com/emberjs/data/issues/3220

Adds a deprecation to relationships defined without `{ async: false }` that warns to add `{ async: false }` if you wish for the relationship to remain synchronous.

### DIFF
--- a/packages/ember-data/lib/system/relationships/belongs-to.js
+++ b/packages/ember-data/lib/system/relationships/belongs-to.js
@@ -92,6 +92,10 @@ function belongsTo(modelName, options) {
 
   opts = opts || {};
 
+  if (typeof opts.async === 'undefined') {
+    Ember.deprecate('In the future, relationships will be asynchronous by default. You must set { async: false } if you wish for a relationship remain synchronous.');
+  }
+
   var meta = {
     type: userEnteredModelName,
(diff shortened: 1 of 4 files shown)


packages/ember-data/lib/system/relationships/belongs-to.js
```diff
@@ -92,6 +92,10 @@ function belongsTo(modelName, options) {
 
   opts = opts || {};
 
+  if (typeof opts.async === 'undefined') {
+    Ember.deprecate('In the future, relationships will be asynchronous by default. You must set { async: false } if you wish for a relationship remain synchronous.');
```
igorT: Can you add which relationship to the deprecation notice please. This way it will be much more annoying to track down.

igorT: Something like model type, relationship key, etc.

igorT:
Can you also add async:false to all our internal tests?

HeroicEric:
just working on updating the tests now. Should be able to get it done today

HeroicEric:
closed by https://github.com/emberjs/data/pull/3366

PR #3284, opened 2015-06-10
Add failing test on PromiseArray.createRecord when called before hasM…

Serabe:
…any is loaded

Rebases #2412


bmac:
Hi @Serabe. Thanks for this pr. Unfortunately I'm going to close it because I don't think Ember Data is going to proxy createRecord to an unresolved relationship any time soon. Because the serialization process is synchronous this functionality may accidentally cause records to be deleted if a user tires to save the newly created record before loading in the rest of the records in the relationship.

PR #3286, opened 2015-06-10
Deprecate `store:application` and `store:main` in favor of `service:store` 

bmac:
Right now Ember Data supports 3 namespaces where you can define the store `store:main`, `store:application` and `service:store`. I want to reduce this to 1 place for Ember Data 2.0 because it simplifies our initializer code and if a user were to define a custom store for more then 1 of those namespaces it can result in multiple stores instances in their app which is confusing to debug.

I'd like to recommend defining a custom store as `service:store` (`app/services/store.js`) as the path going forward. Is this a good idea/bad idea?

I would also like to see this pr become a beta-19.2 patch release to help people upgrading who currently have a custom store defined in `app/store.js`. Let me know what you think about this change?

### DIFF
--- a/packages/ember-data/lib/ember-initializer.js
+++ b/packages/ember-data/lib/ember-initializer.js
@@ -13,7 +13,7 @@ var K = Ember.K;
   This code initializes Ember-Data onto an Ember application.
 
   If an Ember.js developer defines a subclass of DS.Store on their application,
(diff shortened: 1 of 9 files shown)


igorT:
Can we add a test for the duplicate store bug, or is it too hard?

bmac:
@igorT I have updated this pr to add deprecation warnings.

igorT:
:+1: merge whenever travis likes it

PR #3287, opened 2015-06-10
use the esperanto module transpiler / vendor backburner

fivetanley:
**This commit drops support for IE8!**. Esperanto by default uses
Object.defineProperty to support cyclical dependencies.

This allows us to use iojs/node.js v0.12 as well.


stefanpenner:
Esperanto is no more. We will likely keep using embers backburner until ember is also properly moduled and resolvable

Issue #3285, opened 2015-06-10
Consider renaming DS.Snapshot


igorT:
Snapshots are becoming a primary tool for communication between the store and adapter. However `Snapshot` is a pretty confusing name. Can we come up with a better one. I tried AdapterSnapshot to indicated it's meant to be passed to adapter/serializer only, but it's a bad name as it's not a Snapshot of an Adapter

fivetanley:
We decided to not move on this.

PR #3294, opened 2015-06-10
Rename Snapshot to AdapterSnapshot

HeroicEric:
fixes https://github.com/emberjs/data/issues/3285

- [x] Rename `Snapshot` to `AdapterSnapshot`

### DIFF
--- a/packages/activemodel-adapter/lib/system/active-model-serializer.js
+++ b/packages/activemodel-adapter/lib/system/active-model-serializer.js
@@ -165,7 +165,7 @@ var ActiveModelSerializer = RESTSerializer.extend({
     Serializes a polymorphic type as a fully capitalized model name.
 
     @method serializePolymorphicType
-    @param {DS.Snapshot} snapshot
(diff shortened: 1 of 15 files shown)


igorT:
record._createSnapshot is private no need to deprecate

HeroicEric:
@igorT I was just thinking it might be nice since there are some popular addons that are using it.

- https://github.com/lytics/ember-data.model-fragments
- https://github.com/kurko/ember-json-api
- https://github.com/clintjhill/ember-parse-adapter

to name a few

igorT:
well end up with a public hook later, so would rather not rename fo rnow

HeroicEric:
You'd rather not rename `Snapshot` or `Model#_createSnapshot`?

igorT:
_createsnapshot

bmac:
Thanks for the pr @HeroicEric. Unfortunately, the Ember Data team has decided to stick with the name `Snapshot`.

PR #2491, opened 2014-11-20
Async filter

H1D:
This fixes #2181
Almost done. Need your help.

- [x] implement the stuff
- [x] handle possible race conditions
- [ ] check is it causes performance degradation
- [ ] fix failing tests (looks like it changes records order in array sometimes)

P.S. I won't touch internet till 30 of november. On vacation :tropical_drink:

PR #3300, opened 2015-06-11
added promises support for filter()

sly7-7:
Rebase of https://github.com/emberjs/data/pull/2491

Still failing tests, need to investigate


igorT:
Tests were failing because of a missing runloop. Pushed as #3302

PR #3301, opened 2015-06-11
WIP-better unload records fixes #3296

### LINKED ISSUE #3296: Unload issue with relationships
It is not clear what unloading should do with regards to relationships. It seems to me that it should dematerialize a record, but keep the internalModel intact if it has SSOT foreign key references(removing foreign keys seems obviously incorrect).
This is problematic though, because internalModels will still leak memory. How do we let users unload a bunch of related records? Does this mean we have to write a mini Garbage Collector, that will anytime you unload a record, check the foreign key dependencies and unload things it can. This doesn't sound too bad, can you confirm it is the right thing to do?

igorT:
@sly7-7

### DIFF
--- a/packages/ember-data/lib/system/model/internal-model.js
+++ b/packages/ember-data/lib/system/model/internal-model.js
@@ -122,6 +122,14 @@ InternalModel.prototype = {
     this.record = null;
   },
 
+  dematerializeRecord: function() {
+    if (this.record) {
+      this.record.destroy();
+      this.updateRecordArrays();
+      this.record = null;
+    }
+  },
+
   deleteRecord: function() {
     this.send('deleteRecord');
   },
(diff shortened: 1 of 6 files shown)


packages/ember-data/lib/system/model/internal-model.js
```diff
@@ -173,8 +181,63 @@ InternalModel.prototype = {
     return this.record;
   },
 
+  directlyRelatedRecords: function() {
```
sly7-7: maybe s/record/internalModel ?

igorT: please PR to this branch

igorT:
Fixes https://github.com/emberjs/data/issues/3296

raytiley:
@igorT is this still the approach you had in mind?

bmac:
@igorT do you have some time to write up what needs to happen before this pr can be merged?

bmac:
@igorT do you have some time to write up what needs to happen before this pr can be merged?

bmac:

hjdivad:
@igorT rebased at https://github.com/hjdivad/data/tree/better-unload

I'll try to find some time to finish this

PR #3304, opened 2015-06-11
Add a test for async polymorphic relationships with lazy type inference

mphasize:
These test describes a situation where an API has polymorphic relationships, but neither embeds the records nor supplies a type alongside the ID. Since we already have a "parent" type for polymorphic relationships, I was wondering if the related record could be known to Ember Data just as the ID and the parent type until it gets fetched from the API. When we finally have the complete information, the type could be recast to the child type.

There would be 2 requirements:

all the polymorphic records have to be accessible on the endpoint of the parent type
all polymorphic records need to have unique IDs
A bit of related discussion:
#2991

I should mention that this comes from a real-world example. I was trying to build an Ember Data integration for the Contentful content management API.

### DIFF
--- a/packages/ember-data/tests/integration/relationships/belongs-to-test.js
+++ b/packages/ember-data/tests/integration/relationships/belongs-to-test.js
@@ -754,3 +754,28 @@ test("Model's belongsTo relationship should be created during 'get' method", fun
     ok(user._internalModel._relationships.has('favouriteMessage'), "Newly created record with relationships in params passed in its constructor should have relationships");
   });
 });
+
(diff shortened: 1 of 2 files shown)


mphasize:
@igorT Done

igorT:
Thanks!

runspired:
Closing this both because it is stale but also because the API in use did not survive into the 1.x 2.x and now 3.x eras.  For relationships, we follow json-api guidelines. With that in mind, achieving the above today would require using `links`, sending the correct `type` information along with the IDs, or in limited situations using the base `type` until the subclass is known.

PR #3307, opened 2015-06-11
Fix relationship being named `belongsTo` instead of `book`

pangratz:
Looks like this relationship isn't used in any test though ...


bmac:
Thanks

PR #3314, opened 2015-06-11
Extract polymorphic check

pangratz:
Currently the logic to check if a record can be added to a (polymorphic) relationship has been duplicated in two places: within [`belongs-to relationship state](http://git.io/vIHxH) and [`has-many relationship state](http://git.io/vIHxh).

This PR extracts the functionality into a `assertPolymorphicType` function, so it can be reused elsewhere. Currently, also the assertion messages are slightly different; by using the same function they are the same for both calls.

---

Caveats: (currently) only the `Ember.assert` within the `assertPolymorphicType` function is stripped for production builds, so a little noise is left. Is this something we can ignore?

### DIFF
--- a/packages/ember-data/lib/system/relationships/state/belongs-to.js
+++ b/packages/ember-data/lib/system/relationships/state/belongs-to.js
@@ -2,6 +2,8 @@ import {
   PromiseObject
 } from "ember-data/system/promise-proxies";
 
+import { assertPolymorphicType } from "ember-data/utils";
+
 import Relationship from "ember-data/system/relationships/state/relationship";
 
 var BelongsToRelationship = function(store, record, inverseKey, relationshipMeta) {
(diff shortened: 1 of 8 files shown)


fivetanley:
>  only the Ember.assert within the assertPolymorphicType function is stripped for production builds, so a little noise is left. Is this something we can ignore?

Yeah, code stripping doesn't work right now anyway.

fivetanley:
Seems good, seems weird that it passed appveyor but not travis

pangratz:
> seems weird that it passed appveyor but not travis

crazy times!

---

Travis is choking on:

```
  973 tests complete (14138 ms)
  973) unit/transform - DS.StringTransform: global failure
     Error: Could not find module ember-data/utils
```

Do you know what the problem could be here?

fivetanley:
You can't currently use import in tests because of the way it gets compiled for production mode. I'll be fixing that soon (we'll always use an AMD build even for production builds)

pangratz:
Ok, thanks for the heads up. I updated the tests ...

iamjstates:
This is causing my application to fail with my relationships - how do I either fix this or remove the assertion?

pangratz:
@iamjstates can you provide more details on what models you are using? On which response do you get the error? What versions are you using?

brendanoh:
I have the exact same issue as @iamjstates 

this.record is my object and this.relationshipMeta has the type info but the function seems to think it is in newRecord.
 
``` var ember$data$lib$utils$$assertPolymorphicType = function (record, relationshipMeta, addedRecord) {
      debugger
      var addedType = addedRecord.type.modelName;
```

Issue #3327, opened 2015-06-13
[Deprecation] usage of reduce computed in errors.js


stefanpenner:
https://github.com/emberjs/data/blob/97f39acbe5b67157a3077e4ed124290625f00576/packages/ember-data/lib/system/model/errors.js#L112-L132

source: https://github.com/emberjs/ember.js/issues/11435

igorT:
I believe this has been fixed

pangratz:
@igorT yep

yonida:
Still getting it with ember 1.13.0 and ember data 1.0.0-beta.19
How can I fix it?

pangratz:
@yonida This should be fixed in `ember-data` `v1.13.0`

PR #3328, opened 2015-06-13
Suppress deprecation warnings for Ember.reduceComputed

pangratz:
This suppresses the deprecation warnings reported in #3327 (originally reported in emberjs/ember.js#11435), by using the same fix as in emberjs/ember.js#11428.

### DIFF
--- a/packages/ember-data/lib/system/model/errors.js
+++ b/packages/ember-data/lib/system/model/errors.js
@@ -110,6 +110,8 @@ export default Ember.Object.extend(Ember.Enumerable, Ember.Evented, {
     @private
   */
   errorsByAttributeName: Ember.reduceComputed("content", {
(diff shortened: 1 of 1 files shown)


stefanpenner:
should we just remove this RC?

rwjblue:
We will be removing RC from Ember 2.0.0, if we can replace usage here with plain array methods it will save us future time (and help ensure ED versions work better for newer ember versions)...

pangratz:
Convinced. Will update the PR for removing the RC completely...

stefanpenner:
> Convinced. Will update the PR for removing the RC completely...
>  Show all checks

:)

pangratz:
To keep the traceability of this issue intact, I decided against updating the code in this PR and created a new one instead: #3329.

PR #3329, opened 2015-06-13
Refactor `Errors#errorsByAttributeName` to not use Ember.reduceComputed

pangratz:
`Ember.reduceComputed` has been deprecated in emberjs/ember.js#11403. Once this is merged, #3327 can be closed.

### DIFF
--- a/packages/ember-data/lib/system/model/errors.js
+++ b/packages/ember-data/lib/system/model/errors.js
@@ -109,26 +109,12 @@ export default Ember.Object.extend(Ember.Enumerable, Ember.Evented, {
     @type {Ember.MapWithDefault}
     @private
   */
-  errorsByAttributeName: Ember.reduceComputed("content", {
-    initialValue: function() {
-      return MapWithDefault.create({
-        defaultValue: function() {
-          return Ember.A();
-        }
-      });
-    },
-
-    addedItem: function(errors, error) {
-      errors.get(error.attribute).pushObject(error);
-
-      return errors;
-    },
-
-    removedItem: function(errors, error) {
-      errors.get(error.attribute).removeObject(error);
-
-      return errors;
-    }
+  errorsByAttributeName: Ember.computed(function() {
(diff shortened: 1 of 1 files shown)


packages/ember-data/lib/system/model/errors.js
```diff
@@ -297,6 +284,7 @@ export default Ember.Object.extend(Ember.Enumerable, Ember.Evented, {
 
     var content = get(this, 'content').rejectBy('attribute', attribute);
     get(this, 'content').setObjects(content);
+    get(this, 'errorsByAttributeName').remove(attribute);
```
fivetanley: isn't map.remove also deprecated?

pangratz: Lemme check

stefanpenner: `delete` https://people.mozilla.org/~jorendorff/es6-draft.html#sec-map.prototype.delete

pangratz: [Your're absolutely right](https://github.com/emberjs/ember.js/blob/v1.13.0/packages/ember-metal/lib/map.js#L356), nice catch! Will update to use `delete`...

pangratz: So this would have been caught by Travis we would need to run the tests using a debug build of Ember, right?

stefanpenner: c, it seems like we should do that.

pangratz: Hmm, as far as I can see the debug builds [are already used ...](https://github.com/emberjs/data/blob/73cb9247228b86b8829fff3756278a02d39941c6/tests/index.html#L47-L49)

pangratz: This is because currently the `RAISE_ON_DEPRECATION` env is not set to `true`, which is tracked in #3233

PR #3302, opened 2015-06-11
WIP async filter rebase and cleanup of #3300

igorT:

PR #3332, opened 2015-06-14
Clear else branch of RecordArrayManager#filter

pangratz:
The `else if (!shouldBeInArray)` has been replaced with a simple `else`
in during cleanup in b9044fd978c2ff4fc6e6acda1800c57e59351454.

This addresses https://github.com/emberjs/data/pull/3302#discussion_r32233897.


igorT:
thanks!

PR #3333, opened 2015-06-14
Update Ember.js to v1.13

pangratz:


igorT:
cc @fivetanley

fivetanley:
This only updates the local version you run against while testing; it doesn't introduce a hard dependency. I think we're good to support >= 1.8 for 1.13, so this is fine.

PR #3334, opened 2015-06-14
deprecate ActiveModelAdapter being bundled with core

fivetanley:


wecc:
LGTM

igorT:
Can someone setup active model stuff using initializers/container and not hit this deprecation?

fivetanley:
yes. if you are using ember-cli it works. otherwise it should just work.

fivetanley:
https://github.com/ember-data/active-model-adapter/blob/master/lib/globals.js

PR #3325, opened 2015-06-13
Add Deprecation test for store.push with DS.Record in relationship

mikehollis:
Add a test that checks for deprecation message when calling store.push and using a record in a relationship instead of an id.

This tests PR #3279

PR #3335, opened 2015-06-14
Raise error on unhandled deprecation warnings

pangratz:
This is my stab at enabling the `ENV['RAISE_ON_DEPRACATION']` flag when the tests are run. By this, unhandled deprecation messages throw an error and hence it is not possible to use deprecated features  within the tests of `ember-data` (an example being not using the deprecated `Ember.Map.remove` and use `Ember.Map.delete` instead).

Some minor deprecation warnings are fixed in the course of this PR, but there are two areas, which are addressed differently:

- `store.pushMany` is deprecated in favor of `store.push(object)`; since this is used throughout many tests, this deprecation warning is changed to be soft-deprecated, which means that the deprecation warning are logged but they do not result in a failed test/build

- the same is true for `Serializer#extract`, which is deprecated when the `ds-new-serializer-api` feature is enabled.

Soft deprecating those features allows us to migrate the tests step-by-step migrate and make adaptions in smaller chunks.

Changing the log level for deprecations has been implemented in emberjs/ember.js#11419.

### DIFF
--- a/packages/ember-data/tests/unit/store/push-test.js
+++ b/packages/ember-data/tests/unit/store/push-test.js
@@ -184,7 +184,7 @@ test("Calling push on normalize allows partial updates with raw JSON", function 
   equal(person.get('lastName'), "Jackson", "existing fields are untouched");
 });
 
-test("Calling push with a normalized hash containing related records returns a record", function() {
+test("DEPRECATED - Calling push with a normalized hash containing related records returns a record", function() {
   var number1, number2, person;
   run(function() {
     number1 = store.push('phone-number', {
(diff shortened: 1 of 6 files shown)


packages/ember-data/tests/unit/store/push-test.js
```diff
@@ -178,15 +178,18 @@ test("Calling push with a normalized hash containing related records returns a r
...
 
+  run(function() {
+    expectDeprecation(function() {
+      person = store.push('person', {
+        id: 'wat',
+        firstName: 'John',
+        lastName: 'Smith',
+        phoneNumbers: [number1, number2]
+      });
+    }, /You tried to push a record 'Person'' with id '<PhoneNumber:.*>' and passed a DS.Model instance as a value for the relationship '1'. You should instead pass a numerical or string id to represent the record./);
```
pangratz: This deprecation warning is updated in #3325 and needs to be updated once the other PR is merged.

pangratz:
As pointed out by @mixonic in https://github.com/emberjs/ember.js/pull/11419#issuecomment-111874799, this PR uses a private API for changing the log level of certain deprecation messages. This might go through an RFC before it can be used in `ember-data` here ...

bmac:
@pangratz do you mind rebasing this pr? (or starting it over if thats easier). I suspect we should be able to set `RAISE_ON_DEPRECATION` now on master.

pangratz:
@bmac There is an open PR #3233 by @fivetanley which only sets the `RAISE_ON_DEPRECATION`. So I would say you go ahead and merge that one, since it has the same effect, without the need for a new PR ...

PR #3312, opened 2015-06-11
Deprecate fetch and replace it with store.findRecord(type, id, { reload: true})

bmac:
This pr is build on top of pr #3310 and it is intended to be merged after that one is merged.

PR #3340, opened 2015-06-15
Use 'findRecord' instead of 'find' in doc

sly7-7:
cc/ @bmac, I think it was a small bad copy paste in https://github.com/emberjs/data/pull/3312/files#diff-1636bea36bbe333a834ea11fbfbde48dR569


bmac:
:+1:

PR #3343, opened 2015-06-15
Fix store register as a service

MichaelVdheeren:
Do not register if there is a store service registered


bmac:
Closing by this issue has been fixed by https://github.com/emberjs/data/pull/3349

PR #3354, opened 2015-06-15
Replace Ember.EnumerableUtils with Ember.ArrayPollyfills.

bmac:
Ember.EnumerableUtils is deprecated. Ember.ArrayPollyfills is
private. They are both going away in Ember 2.0 but
Ember.ArrayPollyfills doesn't introduce console noise in user's code

cc @fivetanley @igorT 

@igorT this would just be for the 1.13 release as I've seen a bunch of confusion in slack about Ember deprecation warnings originating from Ember Data's use of `EnumerableUtils`.

### DIFF
--- a/packages/ember-data/lib/adapters/fixture-adapter.js
+++ b/packages/ember-data/lib/adapters/fixture-adapter.js
@@ -3,7 +3,7 @@
 */
 var get = Ember.get;
 var fmt = Ember.String.fmt;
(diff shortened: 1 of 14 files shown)


igorT:
Thank you @bmac

stefanpenner:
these are also gone, ES5 has all these functions.

igorT:
@stefanpenner the commit message agrees with you

stefanpenner:
:+1:

PR #3356, opened 2015-06-16
Remove handlebars from the test file 

bmac:
since it is no longer a dependency of ember and is not installed by bower.


pangratz:
This block can be removed as well I guess: [index.html#L34-L40](https://github.com/emberjs/data/blob/0948b3aa19b204ef397f1df05e9f9da792ae096d/tests/index.html#L34-L40)

PR #3358, opened 2015-06-16
Don't include Handlebars in the tests

pangratz:
This is a followup cleanup to #3356, which removes no more needed Handlebars.js in `tests/index.html` (see https://github.com/emberjs/data/pull/3356#issuecomment-112247792).


bmac:
Thanks

PR #3361, opened 2015-06-16
Use MapWithDefault forEach instead of Array#forEach

bmac:
This fixes a deprecation warning that was introduces in https://github.com/emberjs/data/pull/3354


stefanpenner:
This should be covered by my other pr. Which i believe should land after 1.13 branches

bmac:
Agreed.

bmac:
Going to merge this now since it removes about 1400 deprecation warnings when the Ember 1.13 tests are run.

PR #3366, opened 2015-06-16
Log a deprecation warning if users do not specify `async` on a relationship

bmac:


miguelcobain:
@bmac I don't know how, but I'm getting these deprecation warnings even when I set explicitly `{ async: true }`. A bit annoying. :cry: 

Using ED 1.13.5 with Ember 1.13.3.

PR #3370, opened 2015-06-17
Allow overwrite for deprecated adapters

tricknotes:
Currently, the following assignment is ignored.
``` js
DS.ActiveModelAdapter = null;
Ember.typeOf(DS.ActiveModelAdapter); //=> 'class'
```

To fix this issue, `set` function is required for `Ember.defineProperty`.

Should this commit be backport to 1.13.x release?


bmac:
> Should this commit be backport to 1.13.x release?

Yes. I think it should.

PR #3374, opened 2015-06-18
fix jquery error handler

### LINKED ISSUE #3371: Errors not loaded into model's errors attribute
When I submit an invalid request to my API, it returns some JSON with a 422 status.
```js
var user = this.store.createRecord("user");
user.save(); // This triggers a 422 error
```

The JSON returned is along the lines of:

```json
{"errors":{"login":["can't be blank"]}}
```

In 1.0.0.beta-19.1 and earlier, this would allow me to use `user.get("errors")` to iterate over each error. As of 1.13.0, however, after the 422 response is received, `user.errors` is still empty.

I've been using Ember Data's built-in Active Model Adapter/Serializer; I also tried using the ember-data/active-model-adapter addon but ran into the same problem.

Has the approach for accessing a model's errors changed as of 1.13.0?

tchak:
closes #3371

### DIFF
--- a/packages/ember-data/lib/adapters/rest-adapter.js
+++ b/packages/ember-data/lib/adapters/rest-adapter.js
@@ -894,10 +894,6 @@ var RestAdapter = Adapter.extend(BuildURLMixin, {
           error = adapter.ajaxError(jqXHR, textStatus, errorThrown);
         }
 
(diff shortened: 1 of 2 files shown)


tchak:
This is critical :( Needs a patch release. So sorry.
cc @igorT @fivetanley

PR #3375, opened 2015-06-18
JSONAPISerializer only supports the new Serializer API

wecc:
Using the `JSONAPISerializer` didn't make it opt in to the new Serializer API. This PR makes `JSONAPISerializer` always use the new Serializer API.


bmac:
:+1:

wwwdata:
:+1: thank you!

PR #3379, opened 2015-06-18
store.push should support arrays in the JSONAPI data property

bmac:
This fixes a regression that was introduces in eef10be


wecc:
LGTM

PR #3383, opened 2015-06-18
update ember-inflector to 1.6.2

artych:
There are some important fixes in 1.6.2
https://github.com/stefanpenner/ember-inflector/commits/master


bmac:
@artych should this be included in a patch release to Ember Data 1.13 ?

fivetanley:
> @artych should this be included in a patch release to Ember Data 1.13 ?

I just checked and this was the version that was installed at the time of publishing. Thanks!

Issue #3377, opened 2015-06-18
Wiered error. data.data is undefined


novtor:
Hello,
I have a wiered error with ember-data 1.13.1
ember-data.js, line 12671: data.data is undefined, the line in question is:

modelName = data.data.type;

It happens when I call 
store.findRecord('record1', id);

my record1 has 3 hasMany relations that are side loaded in the same response. I am using RESTAdapter

wecc:
Are you pushing JSON-API data to `store.push()` manually? Have you opted in to the new Serializer API?

novtor:
@wecc, no I do not push data manually. And for the new Serializer API I have not yet seen the documentation. Tried to create a jsbin to reproduce but in jsbin everything works fine. Here is my approximate usecase
http://emberjs.jsbin.com/dideza/edit?html,js,output

wecc:
@novtor I think we've tracked down the issue, @bmac is working on it.

novtor:
thank you @wecc

novtor:
I have modified jsbin
http://emberjs.jsbin.com/dideza/edit?html,js,output
Now it throws an error not the same as I had but similar at the same line:
ember-data/ember-data.js:12675:11
Cannot read property 'type' of undefined TypeError: Cannot read property 'type' of undefined
In this jsbin ember-data v1.13.0 (could not find 1.13.1 on net, link on emberjs.com yeilds 404)

bmac:
@wecc I tested @novtor's jsbin with https://github.com/emberjs/data/pull/3379 and it helps the situation but it seems there is another error now where `undefined` is getting added to the `includes` array. I'm still investigating.

bmac:
I figured it out. AdressSerializer needs to opt into the new serializer api.

We should likely assert for that case and give a better warning.

```js
App.AdressSerializer = DS.RESTSerializer.extend(DS.EmbeddedRecordsMixin, {
    isNewSerializerAPI: true,
    attrs: {
        gorods: { embedded: 'always' }
    }
});
```

novtor:
Thank you @bmac, that's do the deal. I'd been setting isNewSerializerAPI on my rest adapter not on rest serializers, my error, thank you for your help

jcope2013:
ran into this also https://github.com/emberjs/data/pull/3379/files fixed my problem

bmac:
I just published ember-data 1.13.3 with a fix for this bug.

PR #3384, opened 2015-06-18
Assert if the RESTSerializers when using the new format calls another

bmac:
serializer that does not use the new format.


bmac:
cc @wecc

wecc:
Merge conflicts but LGTM :+1:

wecc:
Fixes #3377

PR #3390, opened 2015-06-18
Update the jsbin link so it has no deprecation warnings

bmac:


wecc:
Do we want to promote the `JSONAPIAdapter` instead of `RESTAdapter`?

igorT:
Can you set  the html,js flags, as css is not needed

bmac:
@wecc @igorT I have updated this pr to exclude the css page on the jsbin and added a link for the JSONAPIAdapter example.

wecc:
@bmac really nice! getting a deprecation for `shouldReloadAll` in the JSONAPIAdapter one

bmac:
Thanks @wecc the deprecation warning have been fixed.

PR #3394, opened 2015-06-19
[DOC] Deprecate store.pushMany in documentation

mikehollis:
When I pushed up the deprecation for the `store.pushMany` function, I missed adding the deprecation in the comments (and thus the documentation).

This just adds that deprecation to the documents. I'm assuming it's just the `@deprecation` line that will fix this.


bmac:
Thanks

PR #3396, opened 2015-06-19
make sure deprecate error doesnt get swallowed when using store.push(type, data)

jcope2013:
right now it will get swallowed as it will error out before it gets to the deprecation message which is confusing to users trying to consume the api

```
Assertion Failed: You must include an `id` for author in an object passed to `push` 
```

http://emberjs.jsbin.com/hiludosuza/1/edit?html,js,console,output


wecc:
Thanks!

wecc:
ping @bmac, release

bmac:
I have cherry-picked this pr into the release branch for the future 1.13.4 release.

PR #3397, opened 2015-06-20
[BUGFIX release] Use typeOf instead of isArray in _normalizeSerializerPayload

### LINKED ISSUE #3387: Odd behavior with payload object with length attribute in Normalize Serializer Payload
### Current Behavior

The array check (using `Ember.isArray`) in the SerializeResponse [_normalizeSerializerPayload](https://github.com/emberjs/data/blob/v1.13.2/packages/ember-data/lib/system/store/serializer-response.js#L48) returns true for a `payload` object (`[object Object]`) with a `length` data attribute. This yielded a `data` object with several attributes with `undefined` as the value.

`Ember: 1.13.2 &  Ember Data: 1.13.2`

### Expected Behavior

A `payload` object with `length` data attribute should NOT be evaluated as an `Array` and go through the `true` [block](https://github.com/emberjs/data/blob/v1.13.2/packages/ember-data/lib/system/store/serializer-response.js#L49-L51).

cc/ @bmac

---

The JSON API we were using returned a resource with several attributes one of which was `length`.
```json
{
    "id": 123,
    "width": 10,
    "length": 15
}
```

_Note: Don't use `length` or any possible reserved keywords in your API. We have changed our API to now return `size_length` vs. `length`. It goes along the same line as using `lng` vs. `long` for longitude._

wecc:
Because of `Ember.isArray({ length: 1 }) => true`.

`Array.isArray()` used in master is not affected by this.

Fixes #3387

### DIFF
--- a/packages/ember-data/lib/system/store/serializer-response.js
+++ b/packages/ember-data/lib/system/store/serializer-response.js
@@ -45,7 +45,7 @@ export function _normalizeSerializerPayload(modelClass, payload) {
   let data = null;
 
   if (payload) {
-    if (Ember.isArray(payload)) {
+    if (Ember.typeOf(payload) === 'array') {
       data = map.call(payload, (payload) => {
         return _normalizeSerializerPayloadItem(modelClass, payload);
       });


bmac:
:+1:

PR #3398, opened 2015-06-20
[REGRESSION release] Make meta available with new Serializer API

### LINKED ISSUE #3391: Metadata lost in RESTAdapter with isNewSerializerAPI
I talked with @wecc a bit about this in Slack, and it looks like metadata is extracted but not saved when the `RESTAdapter` has its `isNewSerializerAPI` set to true. Here is an example of the problem: 

http://emberjs.jsbin.com/moxuno/3/edit?html,js,output

You can see it not working or working by setting the `isNewSerializerAPI` attribute on the `ApplicationSerializer`.

wecc:
We forgot to deprecate `store.metadataFor` and `store.setMetadataFor` when releasing 1.13. We also failed to keep backwards compatibility with metadata for users that are preparing for 2.0 and switches over to the new Serializer API.

This PR fixes that regression and correctly deprecates `store.metadataFor` and `store.setMetadataFor`.

Fixes #3391

### DIFF
--- a/packages/ember-data/lib/system/store.js
+++ b/packages/ember-data/lib/system/store.js
@@ -1388,8 +1388,20 @@ Store = Service.extend({
     @method metadataFor
     @param {String} modelName
     @return {object}
+    @deprecated
   */
   metadataFor: function(modelName) {
+    Ember.deprecate("`store.metadataFor()` has been deprecated. You can use `.get('meta')` on relationships and arrays returned from `store.query()`.");
+    return this._metadataFor(modelName);
+  },
+
+  /**
+    @method _metadataFor
+    @param {String} modelName
+    @return {object}
+    @private
+  */
+  _metadataFor: function(modelName) {
     Ember.assert('Passing classes to store methods has been removed. Please pass a dasherized string instead of '+ Ember.inspect(modelName), typeof modelName === 'string');
     var typeClass = this.modelFor(modelName);
     return this.typeMapFor(typeClass).metadata;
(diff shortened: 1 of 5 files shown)


packages/ember-data/lib/system/store.js
```diff
@@ -1388,8 +1388,20 @@ Store = Service.extend({
     @method metadataFor
     @param {String} modelName
     @return {object}
+    @deprecated
   */
   metadataFor: function(modelName) {
+    Ember.deprecate("`store.metadataFor()` has been deprecated. You can use `.get('meta')` on relationships and arrays returned from `store.query()`.");
+    return this._metadataFor(...arguments);
```
igorT: can just do modelName

packages/ember-data/lib/system/store.js
```diff
@@ -1402,8 +1414,20 @@ Store = Service.extend({
     @param {String} modelName
     @param {Object} metadata metadata to set
     @return {object}
+    @deprecated
   */
   setMetadataFor: function(modelName, metadata) {
+    Ember.deprecate("`store.setMetadataFor()` has been deprecated. Please return meta from your serializer's `extractMeta` hook.");
+    this._setMetadataFor(...arguments);
```
igorT: same here

JKGisMe:
if it's being deprecated, what's the new way?

wecc:
@JKGisMe You can get meta off of relationships plus arrays returned from `query` by doing `.get('meta')` (when this has been released).

kottenator:
What about "meta" on the top level (not inside relationships)? What is the correct replacement of `this.store.metadataFor('smth')`?

I've provided the full example/question in [another ticket](https://github.com/emberjs/data/issues/3419)

PR #3413, opened 2015-06-22
[BUGFIX release] Sideloaded records are not pushed when saving

### LINKED ISSUE #3406: createRecord response with sideloaded hasMany records doesn't populate store with those records
JSBin:

http://emberjs.jsbin.com/hovisewafa/edit?html,js,console,output

In short: let's say I have a Parent model which has many Child records (async false, because I want to do sideloading). If I do a 'find' on the store for a parent and the adapter returns some sideloaded Child records, then all is fine (the template iterates through the children, rendering the ids).

If I instead do a store.createRecord('parent') and then a save() and tell the adapter's createRecord method to return those same sideloaded Child records, I get an error ("Uncaught Error: Assertion Failed: You looked up the 'childs' relationship on a 'parent' with id 2 but some of the associated records were not loaded."). A check of the store after the adapter response seems to show that no child records were loaded into the store.

In the JSBin you can swap the code in the index route's model hook to see it working for the 'find' case.

In my 'real' app my tests are using jQuery Mockjax to stub the server POST call, so I hope I've done it correctly in the JSBin with overriding the createRecord method to simulate this.

Sorry if I've missed something here (I'm relatively new)...perhaps this is by design and it's not "best practice" to have a server return sideloaded records on a POST response? I'd be happy to explain our use case if there's any interest in it.

wecc:
Before the new Serializer API the serializers pushed sideloaded records to the store. Now sideloaded records are returned in `included` and it's the store's responsibility to push them.

Fixes #3406

### DIFF
--- a/packages/ember-data/lib/system/store.js
+++ b/packages/ember-data/lib/system/store.js
@@ -2292,6 +2292,9 @@ function _commit(adapter, store, operation, snapshot) {
       var payload, data;
       if (adapterPayload) {
         payload = normalizeResponseHelper(serializer, store, typeClass, adapterPayload, snapshot.id, operation);
+        if (payload.included) {
+          store.push({ data: payload.included });
+        }
         data = convertResourceObject(payload.data);
       }
       store.didSaveRecord(internalModel, _normalizeSerializerPayload(internalModel.type, data));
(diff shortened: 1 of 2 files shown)


bmac:
:+1:

PR #3420, opened 2015-06-22
[CLEANUP] Remove ember-new-computed dependency

pangratz:
Since v1.13.0 we use Ember.js-v1.13.0, so this addon is no more needed
since the new computed property syntax is supported out of the box with
this Ember.js version.

### DIFF
--- a/Brocfile.js
+++ b/Brocfile.js
@@ -88,7 +88,6 @@ function packageAddon(packagePath, vendorPath) {
 
 var packages = merge([
   packageAddon('ember-inflector', 'node_modules/'),
-  packageAddon('ember-new-computed', 'node_modules/'),
   package('ember-data'),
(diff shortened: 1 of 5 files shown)


wecc:
I find it weird that this pass :) There's a reference to `ember-new-computed` in _Brocfile.js_ and `computedPolyfill` is used in:

- _packages/ember-data/lib/system/model/attributes.js_
- _packages/ember-data/lib/system/relationships/belongs-to.js_
- _packages/ember-data/lib/system/relationships/has-many.js_

pangratz:
Hmm, good catch. Crazy. Good thing AppVeyor was failing (for a changed :wink: ) Haven't seen those. I will clean the references and update the PR...

wecc:
Thanks!

PR #3430, opened 2015-06-24
[CLEANUP beta] Remove old Serializer API

wecc:
This removes all use of the `isNewSerializerAPI` flag as well as backward compatibility with the old Serializer API.


bmac:
Looks good to me. :+1:

PR #3432, opened 2015-06-24
[CLEANUP beta] Use Object.keys/Object.create instead of Ember.keys/Ember.create

jmurphyau:
Since Ember.create and Ember.keys have been deprecated, ember-data has become the cause of quite a few deprecation warnings in the console.


wecc:
Thanks!

PR #3433, opened 2015-06-24
Fix deprecation warnings

pangratz:
I kept each commit separately to see which deprecations are fixed how. If this is :+1:, I will squash them and prefix with `[CLEANUP beta]`.

### DIFF
--- a/tests/index.html
+++ b/tests/index.html
@@ -62,7 +62,7 @@
 
         var skipPackageRegexp = new RegExp('^('+skipPackages.join('|')+')/');
 
-        var keys = Ember.keys(requirejs.entries);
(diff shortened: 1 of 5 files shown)


wecc:
LGTM

tests/index.html
```diff
@@ -62,7 +62,7 @@
 
         var skipPackageRegexp = new RegExp('^('+skipPackages.join('|')+')/');
 
-        var keys = Ember.keys(requirejs.entries);
+        var keys = Object.keys(requirejs.entries);
```
pangratz: This one has been missed in #3432

wecc: Nice catch!

bmac:
Looks good.

PR #3437, opened 2015-06-24
Deprecate normalizePayload and normalizeHash

### LINKED ISSUE #3428: Deprecate `normalizePayload` and `normalizeHash`
`normalizePayload` and `normalizeHash` is implicitly going away in 2.0 so we have to deprecate them in the next 1.13.x release.

wecc:
I've duplicated the `Ember.deprecate` calls for `normalizePayload` to prevent the deprecation from being silent when users override `normalizePayload` and don't call super. The example in the docs doesn't even call super.

`RESTSerializer` does not override `normalizePayload`, that's why it's comparing with `JSONSerializer.prototype.normalizePayload `.

Closes #3428

### DIFF
--- a/packages/ember-data/lib/serializers/json-serializer.js
+++ b/packages/ember-data/lib/serializers/json-serializer.js
@@ -656,6 +656,7 @@ var JSONSerializer = Serializer.extend({
     @method normalizePayload
     @param {Object} payload
     @return {Object} the normalized payload
+    @deprecated
   */
   normalizePayload: function(payload) {
     return payload;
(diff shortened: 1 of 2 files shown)


packages/ember-data/lib/serializers/json-serializer.js
```diff
@@ -1459,6 +1460,7 @@ var JSONSerializer = Serializer.extend({
     @return {Object} json The deserialized payload
   */
   extractSingle: function(store, typeClass, payload, id, requestType) {
+    Ember.deprecate('`serializer.normalizePayload` has been deprecated. Please use `serializer.normalizeResponse` with the new Serializer API to modify the payload.', this.normalizePayload === JSONSerializer.prototype.normalizePayload);
```
igorT: This is a bit confusing, should we say extractSingle has been deprecated?

wecc: I put the deprecation here instead of inside the `normalizePayload` method since overriding would lose the deprecation. We're only displaying it if the user has overridden `normalizePayload`:

```js
'...modify the payload.', this.normalizePayload === JSONSerializer.prototype.normalizePayload);
```

bcardarella: This is causing thousands of deprecations to be thrown for my application that has a large payload. The logging of all the deprecations is slow and the application load time is impacted. Before this change application load time was nearly instant. Now it is almost 10 seconds.

Issue #3451, opened 2015-06-26
Failure on null has-one payloads


sirvine:
My API generates a payload for a collection, where a single record looks like this:

        {
          "id": "16",
          "type": "contracts",
          "links": {
            "self": "/api/v1/contracts/16"
          },
          "attributes": {
            "title": "Example Title",
            "current-cycle-id": 40,
            "created-at": "2015-05-22T09:38:26.418-04:00",
            "updated-at": "2015-05-22T09:38:26.418-04:00"
          },
          "relationships": {
            "cycles": {
              "links": {
                "self": "/api/v1/contracts/16/relationships/cycles",
                "related": "/api/v1/contracts/16/cycles"
              }
            },
            "supplier": {
              "links": {
                "self": "/api/v1/contracts/16/relationships/supplier",
                "related": "/api/v1/contracts/16/supplier"
              },
              "data": null
            },
            "tower": {
              "links": {
                "self": "/api/v1/contracts/16/relationships/tower",
                "related": "/api/v1/contracts/16/tower"
              },
              "data": {
                "type": "towers",
                "id": "1"
              }
            },
            "client": {
              "links": {
                "self": "/api/v1/contracts/16/relationships/client",
                "related": "/api/v1/contracts/16/client"
              },
              "data": {
                "type": "clients",
                "id": "1"
              }
            }
          }
        }

Note that two of the has-one relationships (tower and client) include a value under the data key, but the supplier relationship is null.

The JSONAPIAdapter is chasing down all of the relationship links correctly. In the case of supplier, the resulting payload looks like:

      {
        "data": null
      }

When the JSONAPISerializer attempts to normalize this payload, it applies `map` to null, which causes a failure.

sirvine:
That did the trick.  Thanks!

PR #3452, opened 2015-06-26
[BUGFIX release] handle case where linked relationship data is null

pangratz:
This addresses #3451.

### DIFF
--- a/packages/ember-data/lib/serializers/json-api-serializer.js
+++ b/packages/ember-data/lib/serializers/json-api-serializer.js
@@ -25,7 +25,7 @@ export default JSONSerializer.extend({
 
     if (Ember.typeOf(documentHash.data) === 'object') {
       documentHash.data = this._normalizeResourceHelper(documentHash.data);
-    } else {
+    } else if (Ember.typeOf(documentHash.data) === 'array') {
       documentHash.data = documentHash.data.map(this._normalizeResourceHelper, this);
     }
 
--- a/packages/ember-data/tests/integration/adapter/json-api-adapter-test.js
+++ b/packages/ember-data/tests/integration/adapter/json-api-adapter-test.js
@@ -875,3 +875,37 @@ test('update record - serialize hasMany', function() {
     });
   });
 });
+
+test('fetching a belongsTo relationship link that returns null', function() {
+  expect(3);
+
+  ajaxResponse([{
+    data: {
+      type: 'post',
+      id: '1',
+      attributes: {
+        title: 'Ember.js rocks'
+      },
+      relationships: {
+        author: {
+          links: {
+            related: 'http://example.com/post/1/author'
+          }
+        }
+      }
+    }
+  }, {
+    data: null
+  }]);
+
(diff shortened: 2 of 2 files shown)


packages/ember-data/lib/serializers/json-api-serializer.js
```diff
@@ -25,7 +25,7 @@ export default JSONSerializer.extend({
 
     if (Ember.typeOf(documentHash.data) === 'object') {
       documentHash.data = this._normalizeResourceHelper(documentHash.data);
-    } else {
+    } else if (Ember.typeOf(documentHash.data) === 'array') {
```
pangratz: @wecc would you prefer a more readable way without the `else if`, like we do in [`json-api-serializer/extractRelationship`](http://git.io/vtCxf)?

packages/ember-data/tests/integration/adapter/json-api-adapter-test.js
```diff
@@ -875,3 +875,34 @@ test('update record - serialize hasMany', function() {
     });
   });
 });
+
+test('fetching a relationship via link works with data = null - #3451', function() {
```
pangratz: Not so sure about the test name but couldn't come up with a better one. Suggestions?

packages/ember-data/lib/serializers/json-api-serializer.js
wecc: This seems fine

packages/ember-data/tests/integration/adapter/json-api-adapter-test.js
```diff
@@ -875,3 +875,37 @@ test('update record - serialize hasMany', function() {
...
+    data: {
+      type: 'post',
+      id: '1',
+      attributes: {
+        title: 'Ember.js rocks'
+      },
+      relationships: {
+        author: {
+          links: {
+            related: 'http://example.com/author/1'
```
wecc: Can we change this URL to `http://example.com/post/1/author`? It really doesn't make sense to provide a URL to a specific author with an ID and then return `null`.

Sorry for the OCD.

packages/ember-data/tests/integration/adapter/json-api-adapter-test.js
wecc: Maybe something like `fetching a belongsTo relationship link that returns null`

wecc:
`[BUGFIX beta]`?

:+1:

pangratz:
> Edit: Hm, maybe should to into release too while we're at it.

You mean a PR into the `release` branch, right?

wecc:
> You mean a PR into the release branch, right?

No, this is fine, against `mater`. But `[BUGFIX release]` and we can cherry-pick to both `release` and `beta`.

pangratz:
> No, this is fine, against mater. But [BUGFIX release] and we can cherry-pick to both release and beta.

Gotcha. Done.

wecc:
Thanks @pangratz!

PR #3458, opened 2015-06-29
[DOC] createRecord relationships close #3421

### LINKED ISSUE #3421: createRecord doesn't keep relationship attributes
`createRecord` somehow removes the `user` attribute when it does `record.setProperties(properties);`
https://github.com/emberjs/data/blob/v1.13.3/packages/ember-data/lib/system/store.js#L336

After `createRecord` I expected:

    title: "foobar", user: 1

got:

    title: "foobar", user: null

test case: http://emberjs.jsbin.com/jatayoviva/edit?html,js,console,output

givanse:


packages/ember-data/lib/system/store.js
```diff
@@ -300,6 +300,17 @@ Store = Service.extend({
     });
     ```
 
+    To create a new instance of `App.Post` that has a relationship with `App.User`
```
bmac: I think we want to move away from references to globals Ember Apps in the API docs (`App.foo`). Do you mind changing this line to "To create a new instance of a `Post` that has a relationship with a `User` record"? and updating line 295 to say " To create a new instance of a `Post`:"

givanse: Ok, I will.

bmac:
Awesome :+1:. Thanks.

PR #3460, opened 2015-06-29
[BUGFIX release] Do not register the service store when it has already been registered

ppcano:
The following change remove the need of the `service:store` registration after the `store` initializer like:

```
import ApplicationStore from 'app/data/store';

export default {
  name: 'setup-store',
  after: ['store'],

  initialize: function(registry, application) {
    registry.register('service:store', ApplicationStore);
  }
};
```

Now, the `service:store` could be registered either before or after the store initializer.

### DIFF
--- a/packages/ember-data/lib/initializers/store.js
+++ b/packages/ember-data/lib/initializers/store.js
@@ -58,7 +58,7 @@ export default function initializeStore(registry, application) {
 
   if (store) {
     registry.register('service:store', store, { instantiate: false });
-  } else {
+  } else if (!registry.has('service:store')) {
     registry.register('service:store', application && application.Store || Store);
   }
 }
--- a/packages/ember-data/tests/integration/application-test.js
+++ b/packages/ember-data/tests/integration/application-test.js
@@ -166,6 +166,30 @@ test("ember-data initializer is run", function() {
   ok(ran, 'ember-data initializer was found');
 });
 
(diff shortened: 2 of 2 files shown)


bmac:
Hi @ppcano. Do you mind sharing a little more info about your usecase? 

In Ember Data 1.13 the supported way to define a custom  `ApplicationStore` is to define it in `app/services/store.js` so it gets registered as `service:store` automatically.

ppcano:
@bmac, my use case is that we are NOT still using `ember-cli. So our initializer/registration process differs a little of the standard  `ember-cli`.

I know the ED/EM position and understand/agree if this change is not included. I wanted to share our specific case.

bmac:
That seems reasonable @ppcano. Do you mind adding a test for this functionality? 

Here is an existing test that uses an `initializer` to help you get started https://github.com/emberjs/data/blob/9022250b73aca03747352c45b064bdf38b27def3/packages/ember-data/tests/integration/application-test.js#L154-L167.

ppcano:
@bmac, done.

bmac:
Awesome. Do you mind also updating the commit message to start with [BUGFIX release] so we can make sure this is included in the next patch release for Ember Data 1.13?

ppcano:
Commit and pull request title changed.

bmac:
Thanks

PR #3468, opened 2015-07-01
[BUGFIX release] shouldReloadRecord and shouldBackgroundReloadRecord …

bmac:
…should not be called when the store is retrieving an async relationship from a record.


wecc:
LGTM :+1:

Issue #3472, opened 2015-07-01
Model.modelName api docs are missing `modelName`


bmac:
![screen shot 2015-07-01 at 11 00 20 am](https://cloud.githubusercontent.com/assets/54056/8457670/d14e6ba6-1fe0-11e5-860b-76efb22a54f6.png)

https://github.com/emberjs/data/blob/v1.13.4/packages/ember-data/lib/system/model/model.js#L883

`modelName` needs to be added after `@property`.

bmac:
@XrXr is on top of this in pr https://github.com/emberjs/data/pull/3473.

PR #3473, opened 2015-07-01
[DOC] Fill @property of DS.Model.modelName

XrXr:
Before this commit, the property would should up as not having a name
in the docs.

invisible property


bmac:
Awesome thanks @XrXr. I just notice this too https://github.com/emberjs/data/issues/3472.

Do you mind amending the commit message to start with [DOC release] so we can make sure this is fixed in the `1.13` branch?

XrXr:
@bmac amended :)

bmac:
Thanks

Issue #3471, opened 2015-07-01
JSONSerializer.extractId does not get model argument


bmac:
I'd like to write a generic `ApplicationSerializer` that computes it id key using the model name. However the current implementation of `extractId` only receives the resourceHash. https://github.com/emberjs/data/blob/3c6bc3356ffaa0d052db9ccd43e96fef11042f8a/packages/ember-data/lib/serializers/json-serializer.js#L528-L532

For example the id key that the server responds with for the `Author` model would be `author_id` and the id key for a `Book` model would be `book_id`.

```js
JSONSerializer.extend({
  extractId: function(resourceHash, model) {
    var primaryKey = Ember.String.underscore(model.modelName) + '_id';
    var id = resourceHash[primaryKey];
   // coerce id to a string
    return '' + id;
  }
});
```

bmac:
Fixed by https://github.com/emberjs/data/pull/3477

PR #3477, opened 2015-07-02
Add modelClass argument to JSONSerializer.extractId

pdud:
This should take care of #3471 :)

### DIFF
--- a/packages/ember-data/lib/serializers/json-api-serializer.js
+++ b/packages/ember-data/lib/serializers/json-api-serializer.js
@@ -190,7 +190,7 @@ export default JSONSerializer.extend({
     this.normalizeUsingDeclaredMapping(modelClass, resourceHash);
 
     let data = {
-      id:            this.extractId(resourceHash),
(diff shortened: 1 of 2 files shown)


wecc:
I'm :+1: this change but I suspect it's be a breaking change for people already overriding `extractId`?

pdud:
Yes it would be a breaking change for those overriding `extractId`.

It is possible to not make it breaking by making the argument order `resourceHash` then `modelClass`.
Doing that would be the opposite to the other extract methods.

https://github.com/pdud/data/blob/extract-id/packages/ember-data/lib/serializers/json-api-serializer.js#L193

bmac:
@wecc was `extractId` introduced in 1.13? Also it currently does not show up in the API docs because the doc block doesn't start with `/**`.

wecc:
LGTM

wecc:
Thanks @pdud!

wecc:
@bmac ping `release`

pdud:
Glad to help out :) Thanks for marking issues as Good for New Contributors!

PR #3482, opened 2015-07-02
[BUGFIX release] Update the version of ember-new-computed to fix an E…

bmac:
…mber.keys deprecation warning.


bmac:
This should fix the remaining deprecation warnings on the release branch. http://output.jsbin.com/losona

Issue #3495, opened 2015-07-04
pushObject in hasMany/belongsTo doesn't set type and id


lessless:
It's and idiomatic hasMany/belongsTo polymorphic example where email can belongs to the many types of Contactable ('profile' is one of them) http://emberjs.jsbin.com/sibecasiga/1/edit?js,output, which was perfectly described in the http://www.toptal.com/emberjs/a-thorough-guide-to-ember-data

The problem is that after profile.get('emails').pushObject(email); email's 'contactable' and 'contactableType' are empty. This is not how it is supposed to be, right?

Also issue may be related to #2985

bmac:
In your jsbin I'm getting an error "Uncaught TypeError: Cannot read property 'modelFor' of undefined" because this line is passing the undefined into `determineRelationshipType`. https://github.com/emberjs/data/blob/cf428d4f5d1cda0414a96cb7daaf2eabb7cd5d39/packages/ember-data/lib/serializers/json-serializer.js#L806

This looks defiantly like a bug to me, although I'm unsure if its the same bug you are describing.

sly7-7:
Hm, it seems like there has something to do with the serializer itself. Using the JSONAPI one to serialize, all seems to be ok.
http://emberjs.jsbin.com/cigimif/2/edit?js,output

Maybe @wecc could be able to quick debug this one.

sly7-7:
I cleaned up a bit the example, and even use the JSONAPI format.
http://emberjs.jsbin.com/cigimif/edit?js,output

I don't know if it's a new bug, or it it has ever worked. That beeing said, as a workaround, you can set the inverse, and that works.

lessless:
thank you @sly7-7   for the tip! will keep an eye on the topic.

wecc:
Is this still an issue on 2.x?

#3500 was closed in favor of #3975

pete-the-pete:
I don't think this is an issue anymore: [https://ember-twiddle.com/0cf9034428370e344f6d0cda8eb529ff](.)  If not, can we close this?

runspired:
@pete-the-pete Relationships "are complicated" BAHAHAHAHAHAHAHA 😂

lessless:
@pete-the-pete shouldn't `email.set('contactable', profile);` and `profile.get('emails').pushObject(email);` produce identical results?

sly7-7:
@pete-the-pete I could probably update my failing PR to if it pass now. Will try that

[...]

PR #3500, opened 2015-07-06
failing test when pushing a polymorphic belongsTo in a has many relat…

sly7-7:
…ionship

Reproducing https://github.com/emberjs/data/issues/3495#issuecomment-118869585

I thought it was only a mistake about the implicitInverseKey, but it's not enough to fix.

cc/ @igorT I'm a bit lost a this point, so many (inverse)relationships created on the fly, I'm not sure what the key should be in order to make DS.belongsTo() working well here.


bmac:
@sly7-7 do you mind reopening this pr with https://github.com/sly7-7/data/commit/3e724df227945ef52726edcd84236e6f1cc7b440 targeted against the master branch?

PR #3506, opened 2015-07-07
[BUGFIX release] There is no `store` property on a serializer

bmac:
Looks like our tests were injecting a `store` property into the serializers, this does not happen in a real app.

### DIFF
--- a/packages/ember-data/lib/serializers/embedded-records-mixin.js
+++ b/packages/ember-data/lib/serializers/embedded-records-mixin.js
@@ -94,6 +94,7 @@ var camelize = Ember.String.camelize;
 */
 var EmbeddedRecordsMixin = Ember.Mixin.create({
 
+  _store: Ember.inject.service('store'),
   /**
     Normalize the record and recursively normalize/extract all the embedded records
     while pushing them into the store as they are encountered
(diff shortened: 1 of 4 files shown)


packages/ember-data/lib/serializers/embedded-records-mixin.js
```diff
@@ -121,9 +121,9 @@ var EmbeddedRecordsMixin = Ember.Mixin.create({
    @param {String} prop the hash has been referenced by
    @return {Object} the normalized hash
   **/
-  normalize: function(typeClass, hash, prop) {
-    var normalizedHash = this._super(typeClass, hash, prop);
-    return this._extractEmbeddedRecords(this, this.store, typeClass, normalizedHash);
+  normalize: function(typeClass, hash, prop, store) {
```
bmac: I'm not really happy about adding a `store` argument here. @wecc is there a better way to give store to `_extractEmbeddedRecords`?

tests/ember-configuration.js
```diff
@@ -99,8 +99,6 @@
     registry.register('adapter:-json-api', DS.JSONAPIAdapter);
     registry.register('serializer:-json-api', DS.JSONAPISerializer);
 
-    registry.injection('serializer', 'store', 'store:main');
```
bmac: Looks like this line was causing the test to pass previously `registry.injection('serializer', 'store', 'store:main');`

wecc:
:+1:

PR #3510, opened 2015-07-07
Assert serializers are using the same API

### LINKED ISSUE #3424: Serializer issue with embedded relationships.
Got a lot of `you must include an 'id' for some-model in an object passed to 'push'` errors since I upgraded to 1.13.4. Tracked it down and it seems to come from my embedded relationships. When I do not define a custom serializer for the embedded relationship, above error is thrown. When I do create a custom serializer (just an empty RESTSerializer), above error disappears.

wecc:
This makes it a little bit easier to track down issues where serializers collaborates and are using different versions (old/new) of the Serializer API.

Closes #3424


wecc:
This should go in `release` not `master`, closing.

PR #3512, opened 2015-07-07
Revert "[BUGFIX release] There is no `store` property on a serializer"

bmac:
Reverts emberjs/data#3506

Looks like I was wrong and a `store` property gets set by `retrieveManagedInstance` https://github.com/bmac/data/blob/18a3c3a356995d1d8f95f02152185699cd99063b/packages/ember-data/lib/system/store.js#L2078

### DIFF
--- a/tests/ember-configuration.js
+++ b/tests/ember-configuration.js
@@ -99,7 +99,7 @@
     registry.register('adapter:-json-api', DS.JSONAPIAdapter);
     registry.register('serializer:-json-api', DS.JSONAPISerializer);
 
-    registry.register('service:store', DS.Store);
+    registry.injection('serializer', 'store', 'store:main');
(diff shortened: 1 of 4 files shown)


tests/ember-configuration.js
```diff
@@ -99,7 +99,7 @@
     registry.register('adapter:-json-api', DS.JSONAPIAdapter);
     registry.register('serializer:-json-api', DS.JSONAPISerializer);
 
-    registry.register('service:store', DS.Store);
```
sly7-7: Maybe this one could have been kept ? I think it's better to register the factories and let the framework do the injection.

sly7-7:
Did you find what was causing the issue in the jsbin you showed me ?

PR #3539, opened 2015-07-13
Only remove deleted records form record arrays when saved

bmac:


packages/ember-data/lib/system/record-array-manager.js
```diff
@@ -57,11 +57,12 @@ export default Ember.Object.extend({
     @method updateRecordArrays
   */
   updateRecordArrays: function() {
-    this.changedRecords.forEach((record) => {
-      if (record.isDeleted()) {
-        this._recordWasDeleted(record);
+    this.changedRecords.forEach((internalModel) => {
+      if (get(internalModel, 'record.isDestroyed') || get(internalModel, 'record.isDestroying') ||
+           (get(internalModel, 'currentState.stateName') === 'root.deleted.saved')) {
```
ianstarz: :+1:

packages/ember-data/lib/system/model/states.js
```diff
@@ -661,6 +661,7 @@ var RootState = {
       // TODO: More robust semantics around save-while-in-flight
       willCommit: Ember.K,
       didCommit: function(internalModel) {
+        internalModel.disconnectRelationships();
```
igorT: Entering the deleted.saved state should handle this most likely

bmac:
@igorT I have updated this pr based on your feedback.

bmac:
@ianstarz Igor explained why its a good idea to remove `disconnectRelationships`/`reconnectRelationships` so I ended up pulling your commit from https://github.com/emberjs/data/pull/3247 into this pr.

ianstarz:
:+1: that's great! I'm glad it could be useful

packages/ember-data/lib/system/model/states.js
```diff
@@ -685,6 +681,7 @@ var RootState = {
       isDirty: false,
 
       setup: function(internalModel) {
+        internalModel.clearRelationships();
```
igorT: We also do this on dematerialize, so it will be called twice now?

packages/ember-data/lib/system/relationships/state/has-many.js
```diff
@@ -156,8 +156,10 @@ ManyRelationship.prototype.findRecords = function() {
   //TODO CLEANUP
   return this.store.findMany(this.manyArray.toArray().map((rec) => rec._internalModel)).
     then(() => {
-      //Goes away after the manyArray refactor
-      this.manyArray.set('isLoaded', true);
+      if (!this.manyArray.get('isDestroyed')) {
```
igorT: This seems scary, why is this needed?

packages/ember-data/lib/system/model/states.js
bmac: Should I remove the `internalModel.clearRelationships();` from the `unloadRecord` methods?

packages/ember-data/lib/system/relationships/state/has-many.js
bmac: I copied it from this code https://github.com/bmac/data/blob/rebase-3247/packages/ember-data/lib/system/relationships/state/has-many.js#L187-L189. 

I think with the async code its possible to get into a state where the relationship is cleaned up before the `findMany` resolves.

bmac:
ping @igorT

packages/ember-data/lib/system/model/states.js
igorT: Hm, I would rely on it there, and not here, but I think we can fix that later when I merge the unload refactor

PR #3540, opened 2015-07-13
[DOC release] Update store.js, additionnal doc

gabriel-letarte:
Added a use case for `Store.find()`.

It was unclear in the documentation that you could pass an object as the second parameter of the `Store.find()` that is translated to query parameters.

### DIFF
--- a/packages/ember-data/lib/system/store.js
+++ b/packages/ember-data/lib/system/store.js
@@ -938,18 +938,28 @@ Store = Service.extend({
     language for all server-side queries, and then require all adapters to
     implement them.
 
+    ---
+
+    If you do something like this:
+
+    ```javascript
+    store.query('person', { page: 1 });
+    ```
+
     The call made to the server, using a Rails backend, will look something like this:
 
     ```
     Started GET "/api/v1/person?page=1"
     Processing by Api::V1::PersonsController#index as HTML
-    Parameters: {"page"=>"1"}
+    Parameters: { "page"=>"1" }
     ```
+
+    ---
 
     If you do something like this:
 
     ```javascript
-    store.query('person', {ids: [1, 2, 3]});
+    store.query('person', { ids: [1, 2, 3] });
     ```
 
     The call to the server, using a Rails backend, will look something like this:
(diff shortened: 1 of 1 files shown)


wecc:
`store.find(type, query)` has been deprecated in favor of `store.query(type, query)`

wecc:
btw, thanks and sorry if I came off short. we really appreciate doc PRs but this should be added to `store.query` instead.

packages/ember-data/lib/system/store.js
```diff
@@ -439,6 +439,16 @@ Store = Service.extend({
 
     ---
 
+    To pass additionnal query parameter, you can pass an object as the second parameter
```
wecc: `additionnal` => `additional`

packages/ember-data/lib/system/store.js
```diff
@@ -439,6 +439,16 @@ Store = Service.extend({
...
+
+    ```javascript
+    store.find('person', {page: 2})
```
wecc: missing spaces, `{ page: 2 }`

gabriel-letarte:
Will take a look and update my pull request, thanks for the quick input

gabriel-letarte:
The current documentation for `store.query` says it's deprecated.
http://emberjs.com/api/data/classes/DS.Store.html#method_query

I guess it will be fixed in next version?
The following line was erroneous but it got remove from current master.
https://github.com/emberjs/data/blob/v1.13.5/packages/ember-data/lib/system/store.js#L1126

gabriel-letarte:
After reading the `store.query` documentation, I found out that a piece of what I wanted to add to the `store.query` doc was still useful. 

Let me know what you think about this new commit.

bmac:
Looks great @gabriel-letarte. Do you mind squashing this down to 1 commit and prefixing it with [DOC release] so we can make sure to cherry-pick it into the 1.13 branch?

gabriel-letarte:
@bmac Should I create a new pull request with the squashed commit?

bmac:
Thanks @gabriel-letarte

PR #3544, opened 2015-07-13
Add error message in _pushInternalMessage when unknown type

### LINKED ISSUE #3542: Assert that type is found when doing store.push
We should add an assert in `store._pushInternalModel` checking that `type` is found.

Serabe:
When pushing data with an unknown type, an error is raised.
Fix #3542

### DIFF
--- a/packages/ember-data/tests/unit/store/push-test.js
+++ b/packages/ember-data/tests/unit/store/push-test.js
@@ -476,6 +476,20 @@ test('Calling push with a link for a non async relationship should warn', functi
   }, /You have pushed a record of type 'person' with 'phoneNumbers' as a link, but the association is not an async relationship./);
 });
 
+test('Calling push with an unknown model name throws an assertion error', function() {
+
+  expectAssertion(function() {
+    run(function() {
+      store.push({
+        data: {
+          id: '1',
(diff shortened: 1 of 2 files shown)


packages/ember-data/tests/unit/store/push-test.js
```diff
@@ -476,6 +476,18 @@ test('Calling push with a link for a non async relationship should warn', functi
...
 });
 
+test('Calling push with an unknown model name throws an assertion error', function() {
+
+  expectAssertion(function() {
+    run(function() {
+      store.push({
+        id: '1',
+        type: 'unknown'
+      });
```
wecc: This should be valid JSON API and has to be wrapped in `data`:

```js
{
  data: {
    type: 'unknown',
    id: '1'
  }
}
```

Serabe:
Thank you, @wecc. Solved.

wecc:
LGTM, @bmac?

bmac:
Looks good. @Serabe Could you amend the commit message to start with [BUGFIX release] so we can include this in a future 1.13 release?

Serabe:
Done!

bmac:
Thanks @Serabe

Issue #3549, opened 2015-07-15
JSONSerializer does not push payload's `included` records into the store


tstirrat:
I am using `DS.JSONSerializer` with `DS.EmbeddedRecordsMixin` while trying to get emberfire compatible with 1.13.x (firebase/emberfire#283).

The embedded records are being correctly extracted into the JSON-API `included` property, but they are not being pushed into the store. This results in the error: 

```
You looked up the 'posts' relationship on a 'author' with id author_1 but some of the associated records were not loaded.
```

Initial payload given to `serializer.normalize`:

```json
{
  "id": "author_1",
  "name": "Tim",
  "posts": [
    {
      "id": "post_1",
      "title": "Post 1"
    }
  ]
}
```

Serializer normalizes this into:

```json
{
  "data": {
    "id": "author_1",
    "type": "author",
    "attributes": {
      "name": "Tim"
    },
    "relationships": {
      "posts": {
        "data": [
          {
            "id": "post_1",
            "type": "post"
          }
        ]
      }
    }
  },
  "included": [
    {
      "id": "post_1",
      "type": "post",
      "attributes": {
        "title": "Post 1"
      },
      "relationships": {}
    }
  ]
}
```

Here is a minimal test to show you what is happening:

http://jsfiddle.net/3eff3bsz/

The culprit is [_normalizeResponse #443](https://github.com/emberjs/data/blob/v1.13.5/packages/ember-data/lib/serializers/json-serializer.js#L443) not passing through the `included` property.

I'm willing to make a PR for this but do I branch off the `release` branch to get it into 1.13.x?

bmac:
Hi @tstirrat. Thanks for tracking down the issue. Go ahead and open a pr against the master branch and prefix your commit message with [BUGFIX release], then the Ember Data team will cherry-pick the commit onto the release branch.

tstirrat:
ok, thanks, will do

wecc:
Thanks for the issue @tstirrat!

This is indeed a bug and should be easily solved by take `included` into account [here](https://github.com/emberjs/data/blob/v1.13.5/packages/ember-data/lib/serializers/json-serializer.js#L453) and [here](https://github.com/emberjs/data/blob/v1.13.5/packages/ember-data/lib/serializers/json-serializer.js#L457).

fivetanley:
closed via #3550

PR #3550, opened 2015-07-15
Fix JSONSerializer to pass through `payload.included`

tstirrat:
Respect an underlying JSON-API payload's `included` property.

Prior to this, records extracted via `EmbeddedRecordsMixin` were being truncated and forgotten.

See #3549

### DIFF
--- a/packages/ember-data/lib/serializers/json-serializer.js
+++ b/packages/ember-data/lib/serializers/json-serializer.js
@@ -449,11 +449,17 @@ var JSONSerializer = Serializer.extend({
     }
 
     if (isSingle) {
-      let { data } = this.normalize(primaryModelClass, payload);
+      let { data, included } = this.normalize(primaryModelClass, payload);
       documentHash.data = data;
+      if (included) {
+        documentHash.included = included;
+      }
     } else {
       documentHash.data = payload.map((item) => {
-        let { data } = this.normalize(primaryModelClass, item);
+        let { data, included } = this.normalize(primaryModelClass, item);
+        if (included) {
+          documentHash.included.push(...included);
+        }
         return data;
       });
     }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ### Master
 
+- [#3550](https://github.com/emberjs/data/pull/3550) JSONSerializer now works with EmbeddedRecordsMixin by respecting the `included` property in normalizeResponse [@tstirrat](https://github.com/tstirrat)
+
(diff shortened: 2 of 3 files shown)


tstirrat:
The tests pass locally for me but I am sure my test is not actually running!! 

is there some way to isolate just this test? or do integration tests need separate command?

fivetanley:
If you use `npm start`, then visit on `localhost:4200`, you can click on the individual test, or use the dropdown menu, to isolate tests or suites of tests.

tstirrat:
@fivetanley ok thanks, got it!

My tests are not running with `npm test`. is this a problem?

fivetanley:
It looks like it is, at least on travis: https://travis-ci.org/emberjs/data/builds/71138302#L552

tstirrat:
They run with `npm start` and they run on travis.. but not locally.

as a test, I deleted the entire `/packages/ember-data/tests/unit` folder and it made no difference to the total test count.

```
975 tests complete (11519 ms)
```

surely that should have dropped by at least 100..

Anyway, I can get by using `npm start` but this shows something is weird with `npm test`

fivetanley:
npm test doesn't do a full build, so you're likely getting an old build. we should probably fix that.

tstirrat:
ok, np, tests are passing now.

kmiyashiro:
Would love for embedded records to not be broken :+1:

packages/ember-data/lib/serializers/json-serializer.js
```diff
@@ -449,11 +449,13 @@ var JSONSerializer = Serializer.extend({
     }
 
     if (isSingle) {
-      let { data } = this.normalize(primaryModelClass, payload);
+      let { data, included } = this.normalize(primaryModelClass, payload);
       documentHash.data = data;
+      documentHash.included = included;
```
wecc: `included` could be `undefined` here, we should add a check

packages/ember-data/lib/serializers/json-serializer.js
```diff
@@ -449,11 +449,13 @@ var JSONSerializer = Serializer.extend({
...
     } else {
       documentHash.data = payload.map((item) => {
-        let { data } = this.normalize(primaryModelClass, item);
+        let { data, included } = this.normalize(primaryModelClass, item);
+        documentHash.included.push(...included);
```
wecc: ...same here

tstirrat:
@wecc fixed, added isolating test

wecc:
Awesome, thanks @tstirrat!

ping @bmac, release

PR #3554, opened 2015-07-17
Add queryRecord method to the RESTAdapter

thaume:
I just added the queryRecord to the RESTAdapter, it still needs more tests (coming soon)

### DIFF
--- a/packages/ember-data/lib/adapters/rest-adapter.js
+++ b/packages/ember-data/lib/adapters/rest-adapter.js
@@ -424,6 +424,34 @@ var RESTAdapter =  Adapter.extend(BuildURLMixin, {
   },
 
   /**
+    Called by the store in order to fetch a JSON object for
+    the record that matches a particular query.
+
+    The `queryRecord` method makes an Ajax (HTTP GET) request to a URL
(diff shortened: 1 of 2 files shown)


bmac:
Awesome. Thanks for getting to this @thaume

wecc:
@thaume what's the status of this? anything I can do to help?

thaume:
@wecc I just need to write a couple more tests, I'll try to do that today so that we can merge it before I leave for 2 weeks away from keyboard

thaume:
@wecc @bmac I just added more tests, do you think we need more ? I feel like there are already many integration tests in the store/query-record-test file

bmac:
@thaume What happens of `posts` is an array instead of a single object?

thaume:
@bmac  It picks the record at index 0, but it still saves all the other records in the store. Do you think this is the intended result ? :)

bmac:
I do. The tests look good. Can you squash these commits and prefix the commit with [BUGFIX release]?

thaume:
@bmac there you go

bmac:
Thanks

wecc:
:tada:

thaume:
cool :)

PR #3560, opened 2015-07-18
Deprecate error property

tchak:
@bmac as discussed, here is the deprecation for `error` property

### DIFF
--- a/packages/ember-data/lib/system/model/model.js
+++ b/packages/ember-data/lib/system/model/model.js
@@ -6,6 +6,9 @@ import {
   keysFunc
 } from 'ember-data/system/object-polyfills';
 
+import computedPolyfill from "ember-new-computed";
+var errorDeprecationShown = false;
+
 /**
   @module ember-data
 */
@@ -128,6 +131,22 @@ var Model = Ember.Object.extend(Ember.Evented, {
     Ember.deprecate('DS.Model#isDirty has been deprecated please use hasDirtyAttributes instead');
     return this.get('currentState.isDirty');
   }),
+
+  /**
+    @property error
+    @type {Boolean}
+    @deprecated
+  */
+  error: computedPolyfill('adapterError', {
+    get() {
+      if (!errorDeprecationShown) {
+        Ember.deprecate('DS.Model#error has been deprecated please use adapterError instead');
+        errorDeprecationShown = true;
+      }
+      return Ember.get(this, 'adapterError');
+    }
+  }),
+
   /**
     If this property is `true` the record is in the `dirty` state. The
     record has local changes that have not yet been saved by the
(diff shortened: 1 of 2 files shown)


packages/ember-data/lib/system/model/model.js
```diff
@@ -128,6 +128,20 @@ var Model = Ember.Object.extend(Ember.Evented, {
...
+  /**
+    @property error
+    @type {Boolean}
+    @deprecated
+  */
+  error: Ember.computed('currentState.isDirty', function(key, error) {
+    if (arguments.length === 2) {
+      this._error = error;
+    }
+    Ember.deprecate('DS.Model#error has been deprecated please use adapterError instead');
```
fivetanley: This seems pretty frequent and i imagine users will disable warnings altogether as their apps will get super noisy. We should probably deprecate once.

tchak: Make sense, I will update later tonight.

tchak:
@fivetanley updated for only show deprecation once

tchak:
come on appveyor...

packages/ember-data/lib/system/model/model.js
```diff
@@ -901,4 +902,19 @@ Model.reopenClass({
...
+    enumerable: false,
+    get: function() {
+      if (!errorDeprecationShown) {
+        Ember.deprecate('DS.Model#error has been deprecated please use adapterError instead');
+        errorDeprecationShown = true;
+      }
+      return Ember.get(this, 'adapterError');
+    }
+  });
+}
```
bmac: Lets just make this a normal computed property so it can be overridden. Part of the original reason for renaming error to `adapterError` was because people were defining an `error` attribute on their models. That should still be allowed.

tchak: right, but now that I am thinking, if someone defined `error` attribute on the model, the deprecation will not work anyway. Not sure what to do about it. Maybe it is double, but I have no idea how to define a property that can be overridden but will still run some code on get...

bmac: If someone defines an error attribute I don't think we should log the deprecation because presumably they want to use the error property as an attribute and are not accidentally trying to reference `adapterError`.

tchak: @bmac this is what I assumed. PR updated.

packages/ember-data/lib/system/model/model.js
```diff
@@ -128,6 +131,22 @@ var Model = Ember.Object.extend(Ember.Evented, {
...
+  */
+  error: computedPolyfill({
+    get() {
+      if (!errorDeprecationShown) {
+        Ember.deprecate('DS.Model#error has been deprecated please use adapterError instead');
+        errorDeprecationShown = true;
+      }
+      return Ember.get(this, 'adapterError');
+    }
+  }),
```
bmac: Does this computed need a dependency on `adapterError` or should it be marked volatile?

tchak: off cours... Sorry for the mess with this thing. Should have been an easy one. Updated...

bmac:
Tested and it works great. :+1:

PR #3571, opened 2015-07-22
[DOC release] Update README examples

HeroicEric:
- Adds missing types for `DS.attr`s
- Use dasherized model names
- Use `store.findRecord` to fetch a single record


wecc:
Nice catch! Can you prefix the commit? I think `[DOC release]` makes sense since all of these changes apply to `release` too.

bmac:
Thanks

PR #3577, opened 2015-07-23
Use the correct modelClass and serializer to extract a polymorphic ty…

### LINKED ISSUE #3547: [Bug] Polymorphic won't work
Hello, I'm trying to work with polymorphic relations, the problem seems to be that it ignores the payload "type" attribute, It will use the payload primary key as modelType instead.

Here is a JsBin: http://emberjs.jsbin.com/wilavu/edit?html,js,output

Let me know what am I doing wrong, if is not a bug I'm really sorry, please close this issue.

Thanks

bmac:
…pe when normalizing async relationships.

Closes #3547

### DIFF
--- a/packages/ember-data/lib/serializers/rest-serializer.js
+++ b/packages/ember-data/lib/serializers/rest-serializer.js
@@ -149,7 +149,7 @@ var RESTSerializer = JSONSerializer.extend({
     let serializer = store.serializerFor(modelName);
 
     arrayHash.forEach((hash) => {
-      let { data, included } = serializer.normalize(modelClass, hash, prop);
+      let { data, included } = this._normalizePolymorphicRecord(store, hash, prop, modelClass, serializer);
       documentHash.data.push(data);
       if (included) {
         documentHash.included.push(...included);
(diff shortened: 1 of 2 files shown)


bmac:
@marcemira I believe this fixes https://github.com/emberjs/data/issues/3547.


@igorT and @wecc can you review?

wecc:
LGTM but we should implement the same thing [here](https://github.com/bmac/data/blob/bug-3547/packages/ember-data/lib/serializers/rest-serializer.js#L142-L160) or it will only work for singular object root keys.

Maybe a private helper method wrapping this would be nice?

marcemira:
:+1:  great! Thank you guys :grin:

jrhe:
I can confirm that this fixes #3547 but only for singular keys. Needs the fix @wecc mentioned.

bmac:
@wecc, I have updated this pr. Do you mind reviewing it again?

wecc:
I would like to DRY this up a bit and move the duplicate code to a helper but we can definitely do that later.

:shipit:

PR #3582, opened 2015-07-24
[BUGFIX release] Fix unconsistent behavior in Model.changedAttributes

### LINKED ISSUE #3579: model.changedAttributes() behaves inconsistently
I've observed that `DS.Model.changedAttributes()` behaves differently on new records than on saved ones (the behaviour on the saved models looks correct to me, the one in the new records seems as bug)

If you create a new record and change some of its attributes, `model.changedAttributes()` will return an object with the properties that have changed and the current and previous value, and `model.isDirty` is true. So far so good.

But when you save it, while the request is pending and the record is still being saved, on the new record the changed attributes is cleared (an empty object), while on the already saved record it the changed attributes don't change until the save operation completes.

The behaviour of the saved records seems the right one. The dirty attributes should remain dirty until we get confirmation from the API).

In both cases the record's `isDirty` property is true, is just the `changedAttributes()` function the one that returns the wrong value.

Reproducible live example: http://development.dummy-app.divshot.io/ember-data-dirty-bug
Code: https://github.com/cibernox/dummy-app-for-benchmark

Also I'm curious about why `changedAttributes` is a function and not a computedProperty.

cibernox:
closes #3579

### DIFF
--- a/packages/ember-data/lib/system/model/model.js
+++ b/packages/ember-data/lib/system/model/model.js
@@ -6,6 +6,8 @@ import Errors from "ember-data/system/model/errors";
 */
 
 var get = Ember.get;
+var merge = Ember.merge;
+var copy = Ember.copy;
 
 function intersection (array1, array2) {
   var result = [];
@@ -611,7 +613,9 @@ var Model = Ember.Object.extend(Ember.Evented, {
   */
   changedAttributes: function() {
     var oldData = get(this._internalModel, '_data');
-    var newData = get(this._internalModel, '_attributes');
+    var currentData = get(this._internalModel, '_attributes');
+    var inFlightData = get(this._internalModel, '_inFlightAttributes');
+    var newData = merge(copy(inFlightData), currentData);
     var diffData = Object.create(null);
 
     var newDataKeys = Object.keys(newData);
--- a/packages/ember-data/tests/unit/model-test.js
+++ b/packages/ember-data/tests/unit/model-test.js
@@ -249,7 +249,6 @@ test("it should cache attributes", function() {
       run(store, 'destroy');
     });
   });
-
 });
 
 test("changedAttributes() return correct values", function() {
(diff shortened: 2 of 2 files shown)


packages/ember-data/lib/system/model/model.js
```diff
@@ -611,7 +613,9 @@ var Model = Ember.Object.extend(Ember.Evented, {
   */
   changedAttributes: function() {
     var oldData = get(this._internalModel, '_data');
-    var newData = get(this._internalModel, '_attributes');
+    var currentData = get(this._internalModel, '_attributes');
```
cibernox: BTW, I wonder why here we have to use `get` instead of just `this._internalModel._whatever`

packages/ember-data/tests/unit/model-test.js
```diff
@@ -298,6 +297,68 @@ test("changedAttributes() return correct values", function() {
   equal(Object.keys(mascot.changedAttributes()).length, 0, 'after rollback attributes there are no changes');
 });
 
+test("changedAttributes() works while the record is being saved", function() {
+  var cat;
```
bmac: Since the assert is in an adapter method could you add `expect(1)` to the start of this test.

packages/ember-data/tests/unit/model-test.js
```diff
@@ -298,6 +297,68 @@ test("changedAttributes() return correct values", function() {
...
+
+  run(function() {
+    cat = store.createRecord('mascot');
+    cat.setProperties({ name: 'Argon', likes: 'Cheese' });
+    cat.save();
+  });
+});
+
+test("changedAttributes() works while re record is being updated", function() {
+  var cat;
```
bmac: Also add `expect(1)` to this test.

cibernox:
Updated

bmac:
Looks good. :+1:

cibernox:
Fixed a typo. Now it's done

bmac:
Thanks @cibernox

Issue #3493, opened 2015-07-03
Ember Data 1.13.4: Rejected promise error on model.save() has no responseJSON


sarus:
When using the following versions:

```
ember.debug.js:4888DEBUG: Ember                     : 1.12.0
ember.debug.js:4888DEBUG: Ember Data                : 1.0.0-beta.18
ember.debug.js:4888DEBUG: jQuery                    : 1.11.3
ember.debug.js:4888DEBUG: Ember Simple Auth         : 0.8.0-beta.3
ember.debug.js:4888DEBUG: Ember Simple Auth Testing : 0.8.0-beta.3
```

A failed call (Bad Request HTTP Code 400) to `model.save` returns an error object in the promise that contains the responseJSON so the following works:

```javascript
        this.model.save().then(function (model) {
          // save worked
        }, function (error) {
          // error object contains responseJSON so I can get access to the server response 
          self._showErrorMessage(error.responseJSON.userMessage);
        })
```

When upgrading to v1.13:

```
ember.debug.js:4888DEBUG: Ember                     : 1.13.2
ember.debug.js:4888DEBUG: Ember Data                : 1.13.4
ember.debug.js:4888DEBUG: jQuery                    : 1.11.3
ember.debug.js:4888DEBUG: Ember Simple Auth         : 0.8.0
ember.debug.js:4888DEBUG: Ember Simple Auth Testing : 0.8.0
```

The `error` object returned by the rejected promise is a totally different object and I don't see the servers responseJSON anywhere in it.

Is this intentional?  How can I get access to the server response in the latest ember data?  Thank you!

sarus:
To update @tchak mentioned in slack that Ember Data used to reject with the xhr object but now rejects with a normalized error.  The suggestion was look in `error.errors[0]`.

```
        this.model.save().then(function (model) {
          // save worked
        }, function (error) {
          console.info(error.errors[0]);
        })
```
However that just contains the following including the string `[object Object]` which seems like a serialization issue?

```
details: "[object Object]"
status: "400"
title: "The backend responded with an error"
```

tchak:
@sarus yes `normalizeErrorResponse` will assume that if the error payload is not in JSON-API format (`errors: []` form) the payload is a string. I am open to discuss what the default should be. In the mean time, you can override `parseErrorResponse` in order to return a string that will be interpreted as error detail.

tchak:
[...]

Issue #3491, opened 2015-07-03
ember-data error does not contain http status code


jsedlacek:
After upgrade to ember-data `1.13.4` I am not able to find the http status code of failed request.

Here is an example when an API does not exist:

```javascript
// models/user.js:
export default DS.Model.extend({});
```

```javascript
// routes/index.js:
export default Ember.Route.extend({
  model() {
    return this.store.find('user', 1);
  },
  actions: {
    error(err) {
      assert.equal(err.status, 404);
    }
  }
});
```

Result:
```
expected undefined to equal 404
```

Note 1: This works correctly in `1.0.0-beta.18`.

Note 2: This behavior is documented on http://guides.emberjs.com/v1.12.0/routing/loading-and-error-substates/

tchak:
We need to fix the documentation, thank's for the pointer.
We are now rejecting with normalized `DS.AdapterError` instead of `xhr` object
If you want to access some information about the error, you need to use `err.errors[0]`.

jsedlacek:
Thanks @tchak. However `err.errors[0]` does not contain the status code.

Do I understand it correctly that there's no way to get the status code any more?

tchak:
@jsedlacek what kind of error is the backend returning? It may be possible that you will need to customize the adapter hook `handleResponse`. This is intended. In most cases it is a bad idea to leek http details to you app. You probably want to reject with a specific type of error.

jsedlacek:
@tchak I'd like to handle `401` (Unauthorized) error to transition to login route when the auth token expires.

The error body looks like this:
```json
{
    "errors": {
        "base": ["You're not authorized to access this resource"]
}
```

tchak:
@jsedlacek you should override `handleResponse(status, headers, body)` on the adapter and reject with `UnauthorizedError`.

jsedlacek:
Perfect, let me try that. Thanks for quick responses.

tchak:
Something like this

```
handleResponse(status, headers, body) {
  if (status === 401) {
    return new UnauthorizedError();
  } else {
    return this._super(status, headers, body);
  }
}

jsedlacek:
The proposed solution works like a charm. Thanks for your help @tchak.

albertpak:
[...]

Issue #3585, opened 2015-07-25
Add API docs and info in the guides about `ds-extended-errors`


tchak:
see #3493 for an example of confusion

zyllorion:
Please also document how to handle timeouts from requests. E.g. hook, retry 3 times and redirect to an error page.

So far the only documentation I've ever been able to find is around 422 field level validations, not 4xx and 5xx errors or timeouts in the newer ED builds.

Thank you.

zyllorion:
Please also document how to handle timeouts from requests. E.g. hook, retry 3 times and redirect to an error page.

So far the only documentation I've ever been able to find is around 422 field level validations, not 4xx and 5xx errors or timeouts in the newer ED builds.

Thank you

locks:
I'm closing this since https://github.com/emberjs/data/pull/4632 has been merged for a while. Reopen if I missed something!

PR #3586, opened 2015-07-25
[FEATURE ds-extended-errors] Make adapter error extendable and add more error types

tchak:
This is going to help with questions like #3493 or #3491

### DIFF
--- a/addon/adapters/rest.js
+++ b/addon/adapters/rest.js
@@ -7,6 +7,10 @@ import Adapter from "ember-data/adapter";
 import {
   AdapterError,
   InvalidError,
+  UnauthorizedError,
+  ForbiddenError,
+  NotFoundError,
+  ConflictError,
   TimeoutError,
   AbortError
 } from 'ember-data/adapters/errors';
@@ -905,6 +909,19 @@ var RESTAdapter = Adapter.extend(BuildURLMixin, {
 
     let errors          = this.normalizeErrorResponse(status, headers, payload);
     let detailedMessage = this.generatedDetailedMessage(status, headers, payload, requestData);
+
+    if (isEnabled('ds-extended-errors')) {
+      switch (status) {
+      case 401:
+        return new UnauthorizedError(errors, detailedMessage);
+      case 403:
+        return new ForbiddenError(errors, detailedMessage);
+      case 404:
+        return new NotFoundError(errors, detailedMessage);
+      case 409:
+        return new ConflictError(errors, detailedMessage);
+      }
+    }
 
     return new AdapterError(errors, detailedMessage);
   },
--- a/FEATURES.md
+++ b/FEATURES.md
@@ -66,3 +66,18 @@ entry in `config/features.json`.
 
   This is particularly useful for polymorphic relationships not backed by STI when just including the id
   of the records is not enough.
+
+- `ds-extended-errors`
+
+  Enables `extend` method on errors. It means you can extend from `DS.AdapterError`.
+
+  ```js
+    const MyCustomError = DS.AdapterError.extend({ message: "My custom error." });
+  ```
+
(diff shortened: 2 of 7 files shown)


btecu:
This could probably be labeled as [BUGFIX release].

tchak:
Strictly this is not a bug fix. I am not sure where is the line of what we back porting in 1.13 series. @bmac  you have an opinion?

bmac:
I think we should treat it as a new feature and feature flag it for a future 2.x release. If we find after it is release that it would provide value and help with the transition for users on the 1.13 series we can always backport it in the future.

bmac:
@tchak What is the status of this pr?

lindyhopchris:
@tchak  @bmac can I also ask what the status of this is? This is really required because `AdapterError` is just too generic to handle effectively in route transitions. Plus because it doesn't give access to the HTTP status code, it's difficult to handle.

For info, I do think that `AdapterError` should have the original HTTP status code on it. The reason being that the JSON API spec says:

> Error objects provide additional information about problems encountered while performing an operation. 

The key word here being **additional**. There's no requirement for a server to send a response that does include error objects within the JSON body - a simple HTTP status response with no content is valid. I know you've said in #3491 that the original HTTP response should not leak into the application, but it seems to me that the status code is required information considering that there is no obligation under the spec for the server to return error objects in its response body.

bmac:
@tchak  Do you mind rebasing this pr and wrapping it in a [feature flag](https://github.com/emberjs/data/pull/3979)?

tchak:
@bmac done

addon/adapters/rest.js
```diff
@@ -906,7 +910,18 @@ var RESTAdapter = Adapter.extend(BuildURLMixin, {
     let errors          = this.normalizeErrorResponse(status, headers, payload);
     let detailedMessage = this.generatedDetailedMessage(status, headers, payload, requestData);
 
-    return new AdapterError(errors, detailedMessage);
+    switch (status) {
```
bmac: Should this be wrapped in a `isEnabled('ds-extended-errors');`?

lindyhopchris:
@tchak @bmac 

are there any plans to get this out from behind a feature flag? if so, what release might that happen in?

bmac:
Currently the only think blocking this feature flag from being enabled is a lack of documentation. We need some more fleshed out API docs for all of the errors and a pr on the guides repo explaining the new errors.

I haven't had any time to add docs for this feature but @lindyhopchris if you are up for the challenge a pr updating the error's [doc blocks](https://github.com/emberjs/data/blob/c9d8212c857ca78218ad98d11621819b38dba98f/addon/adapters/errors.js#L119-L164) to describe when they trigger (and their [status codes](https://github.com/emberjs/data/blob/master/FEATURES.md#feature-flags)) would be a great place to start.

wecc:
Related: #3585

lindyhopchris:
@bmac happy to give it a go, will probably have some time later this week to look at it and submit a PR.

sandstrom:
@bmac Now that it's documented (https://github.com/emberjs/data/issues/4632), can the flag be removed? 😄

sandstrom:
friendly ping @alexlafroscia 😄

lindyhopchris:
@sandstrom 

re: your second point, `AbortError` and `TimeoutError` are included:
https://github.com/emberjs/data/blob/master/addon/adapters/errors.js#L214-L226

and are used here:
https://github.com/emberjs/data/blob/master/addon/adapters/rest.js#L1518-L1520

sandstrom:
@lindyhopchris Thanks, you are correct! Thanks for correcting me + this is great news! 😄

lindyhopchris:
ping @bmac 

would be great to see this feature taken out from behind the feature flag. any update on when this might happen?

Issue #3607, opened 2015-07-29
urlForQueryRecord is missing in BuildURLMixin


wecc:
We should add and make sure we call `urlForQueryRecord` [here](https://github.com/emberjs/data/blob/v1.13.7/packages/ember-data/lib/adapters/build-url-mixin.js#L52) when `requestType` is `queryRecord`.

PR #3610, opened 2015-07-29
[BUGFIX release] Add missing urlForFindQueryRecord to BuildUrlMixin

HeroicEric:
fixes https://github.com/emberjs/data/issues/3607

~~I wasn't sure which branch this should be opened against. What's the protocol when something needs to go into master as well as release?~~

### DIFF
--- a/packages/ember-data/lib/adapters/build-url-mixin.js
+++ b/packages/ember-data/lib/adapters/build-url-mixin.js
@@ -57,6 +57,8 @@ export default Ember.Mixin.create({
         return this.urlForFindAll(modelName);
       case 'query':
         return this.urlForQuery(query, modelName);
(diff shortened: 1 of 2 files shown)


HeroicEric:
What is the intended behavior for `store.queryRecord()`?

Should this be:

```javascript
this.store.queryRecord('comment', { id: 11, postId: 42 }); 
// => /comments?id=11&post_id=42
```

or

```javascript
this.store.queryRecord('comment', { postId: 42 }, 11); 
// => /comments/11?post_id=42
```

The example implementation in the docs is not clear about this and uses makes use of an `id` param that isn't passed into the actual function but is documented as coming after the `query` param.

wecc:
Thanks @HeroicEric!

PR #3615, opened 2015-07-30
[DOC release] Update misleading example for Adapter.queryRecord()

HeroicEric:
- Uses ember-cli style modules
- Removes `id` param

### DIFF
--- a/packages/ember-data/lib/system/adapter.js
+++ b/packages/ember-data/lib/system/adapter.js
@@ -199,13 +199,16 @@ var Adapter = Ember.Object.extend({
 
     Example
 
-    ```javascript
-    App.ApplicationAdapter = DS.Adapter.extend({
-      queryRecord: function(store, typeClass, query) {
-        var url = [type.typeKey, id].join('/');
+    ```app/adapters/application.js
+    import DS from 'ember-data';
+    import Ember from 'ember';
+
+    export default DS.Adapter.extend(DS.BuildURLMixin, {
+      queryRecord: function(store, type, query) {
+        var urlForQueryRecord = this.buildURL(type.modelName, null, null, 'queryRecord', query);
 
         return new Ember.RSVP.Promise(function(resolve, reject) {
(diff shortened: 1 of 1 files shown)


packages/ember-data/lib/system/adapter.js
```diff
@@ -199,10 +199,13 @@ var Adapter = Ember.Object.extend({
...
-    App.ApplicationAdapter = DS.Adapter.extend({
-      queryRecord: function(store, typeClass, query) {
-        var url = [type.typeKey, id].join('/');
+    ```app/adapters/application.js
+    import DS from 'ember-data';
+    import Ember from 'ember';
+
+    export default DS.Adapter.extend(DS.BuildURLMixin, {
+      queryRecord: function(store, type, query) {
+        var url = this.buildURL(type.modelName, null, null, 'queryRecord', query);
```
wecc: Should this be `urlForQueryRecord` instead to make it a bit cleaner?

bmac:
Thanks @HeroicEric

packages/ember-data/lib/system/adapter.js
```diff
@@ -199,13 +199,16 @@ var Adapter = Ember.Object.extend({
...
-    App.ApplicationAdapter = DS.Adapter.extend({
-      queryRecord: function(store, typeClass, query) {
-        var url = [type.typeKey, id].join('/');
+    ```app/adapters/application.js
+    import DS from 'ember-data';
+    import Ember from 'ember';
+
+    export default DS.Adapter.extend(DS.BuildURLMixin, {
+      queryRecord: function(store, type, query) {
+        var urlForQueryRecord = this.buildURL(type.modelName, null, null, 'queryRecord', query);
```
wecc: Just for the record I didn't mean to change the variable name but to call urlForQueryRecord instead of buildURL :)

HeroicEric: :smile:

PR #3624, opened 2015-08-03
Implement filterBy on DS.ManyArray

bmac:


packages/ember-data/lib/system/record-arrays/filtered-subset.js
```diff
@@ -0,0 +1,15 @@
+var FilteredSubset = Ember.ArrayProxy.extend({
+  init: function() {
+    this._super(...arguments);
+
+    var { filterByArgs, recordArray } = this.getProperties('filterByArgs', 'recordArray');
+    var [key] = filterByArgs;
+
+    var path = `recordArray.@each.${key}`;
+    Ember.defineProperty(this, 'content', Ember.computed(path, function() {
```
igorT: I find this init highly suspect but it was already in there so fine for now

packages/ember-data/lib/system/many-array.js
```diff
@@ -277,5 +278,37 @@ export default Ember.Object.extend(Ember.MutableArray, Ember.Evented, {
...
+    ```
+
+    @method filterBy
+    @param {String} key property path
+    @param {*} value optional
+
+  */
+  filterBy: function(key, value) {
+    // only pass value to the arguments if it is present; this mimics the same
+    // behavior for `filterBy`: http://git.io/vIurH
```
igorT: seems weird to not error out on filterBy with no args, I would certainly not call that behavior supported. Don't feel strongly about it though

igorT:
:+1: other than comments

dustinfarris:
Thanks for this.

Issue #3635, opened 2015-08-10
Ember data 1.13.8 should have removed _registerHelper deprecations warnings but didn't


adam-knights:
This is continueing on the conversationat the end of issue https://github.com/stefanpenner/ember-inflector/issues/75.

@jcope2013 was of the opinion that something has gone wrong with the build of 1.13.8 on bower.

To summarise - if you compare http://builds.emberjs.com/tags/v1.13.8/ember-data.js which does not give deprecation warnings and https://cdnjs.cloudflare.com/ajax/libs/ember-data.js/1.13.8/ember-data.js which does cause deprecation warnings then theres a large difference in the diff in that the cdnjs file has this section (from about line 2322) while the other does not:

    var ember$inflector$lib$lib$utils$register$helper$$default = ember$inflector$lib$lib$utils$register$helper$$registerHelper;

    function ember$inflector$lib$lib$utils$register$helper$$registerHelperIteration1(name, helperFunction) {
      //earlier versions of ember with htmlbars used this
      ember$lib$main$$default.HTMLBars.helpers[name] = helperFunction;
    }

    function ember$inflector$lib$lib$utils$register$helper$$registerHelperIteration2(name, helperFunction) {
      //registerHelper has been made private as _registerHelper
      //this is kept here if anyone is using it
      ember$lib$main$$default.HTMLBars.registerHelper(name, helperFunction);
    }

    function ember$inflector$lib$lib$utils$register$helper$$registerHelperIteration3(name, helperFunction) {
      //latest versin of ember uses this
      ember$lib$main$$default.HTMLBars._registerHelper(name, helperFunction);
    }
    function ember$inflector$lib$lib$utils$register$helper$$registerHelper(name, helperFunction) {
      if (ember$lib$main$$default.HTMLBars) {
        var fn = ember$lib$main$$default.HTMLBars.makeBoundHelper(helperFunction);

        if (ember$lib$main$$default.HTMLBars._registerHelper) {
          if (ember$lib$main$$default.HTMLBars.helpers) {
            ember$inflector$lib$lib$utils$register$helper$$registerHelperIteration1(name, fn);
          } else {
            ember$inflector$lib$lib$utils$register$helper$$registerHelperIteration3(name, fn);
          }
        } else if (ember$lib$main$$default.HTMLBars.registerHelper) {
          ember$inflector$lib$lib$utils$register$helper$$registerHelperIteration2(name, fn);
        }
      } else if (ember$lib$main$$default.Handlebars) {
[...]

PR #3636, opened 2015-08-10
Add ember-cli-dependency-checker.

rwjblue:
Since we are using ember-cli to build ember-data we can take advantage of the work done in ember-cli-dependency-checker to ensure that the deps in `node_modules` and `bower_components` actually match what is specified in the `package.json`/`bower.json`.

With this added, running `ember` commands will error if matching versions are not found.

This will help prevent issues like #3635 in the future.

### DIFF
--- a/package.json
+++ b/package.json
@@ -55,6 +55,7 @@
     "defeatureify": "~0.1.4",
     "ejs": "^1.0.0",
     "ember-cli": "^0.1.15",
+    "ember-cli-dependency-checker": "^1.0.1",
     "ember-inflector": "^1.9.0",
     "ember-publisher": "0.0.7",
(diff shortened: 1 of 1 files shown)


fivetanley:
:+1: just waiting on travis to make sure it likes it.

fivetanley:
Build failure is not the fault of this commit. Merging. We have some work to do to fix on canary.

quaertym:
:+1:

bmac:
Thanks @rwjblue

stefanpenner:
@fivetanley FYI this appears to have been red when merged. And is blocking other PR now, going forward can I recommend we avoid this?

PR #3641, opened 2015-08-12
Improve InternalModel (2x - 3x faster) 

stefanpenner:

PR #3646, opened 2015-08-12
Replace instances of Object.create(null) with new EmptyObject().

### LINKED ISSUE #3642: Switch Object.create(null) to new EmptyObject
once https://github.com/emberjs/data/pull/3641 lands

it is considerably faster for instantiate (in v8 8x or more), with the same benefits.

karanjthakkar:
Fixes #3642 and prepares for #3641

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -831,7 +831,7 @@ correctly will need a shim for Object.create.
 * Refactor JSON serializer to use _getMappedKey
 * Don't normalize the key if not present in the hash
 * Add serializeIntoHash to the JSON Serializer
-* prefer Object.create(null) for typeMap backing stores. Cache misses are faster, and won’t collide with prototype pollution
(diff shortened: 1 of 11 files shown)


stefanpenner:
a rebase should pass the tests: https://github.com/emberjs/data/pull/3641

stefanpenner:
after rebasing, this appears to be failing to do a bug in QUnit, that prevents EmptyObjects from being compared to `{  }`

https://github.com/jquery/qunit/blob/d359314e0ad6a21d998bed450f6fc289c5be3490/src/equiv.js#L185-L186

stefanpenner:
opened a QUnit issue: https://github.com/jquery/qunit/issues/851

stefanpenner:
rebased and worked around a QUnit bug here: https://github.com/emberjs/data/pull/3649

PR #3647, opened 2015-08-12
Adapter.query should be part of the public overrideable interface.

yaymukund:
[The docs tell us to override `query`][override-docs] so it should be public.

This was made private in a commit in 2013 (ec2e040a) by @rwjblue. Many of the other private methods in that commit are now public.

[override-docs]: https://github.com/emberjs/data/blob/v1.13.8/packages/ember-data/lib/system/adapter.js#L47


bmac:
Thanks

PR #3652, opened 2015-08-13
Don’t crash \w Ember 2.0

stefanpenner:
There may be more, but this at-least makes my one app work correctly.

- [ ] try getting tests working to get actual confidence.
- [ ] fix other things
- [ ] instead of poluting the Ember object, we should instead add a level of indirection by moving the polifills to a local module, and doing the funky work there.

Goal:
------

to ease upgrade pain for people


stefanpenner:
the GH ui decided after a force push that this was to go to master, unsure why... 

new PR is correct again https://github.com/emberjs/data/pull/3667

PR #3659, opened 2015-08-15
If using {{input value=date}},  should got the right date instead of null.

Eric-Guo:
To avoid writing below code in Ember app.

```javascript
export default Ember.Component.extend({
  actions: {
    save: function(item) {
      let self = this;
      item.set('start', new Date(item.get('start'))); # not necessary
      item.set('end', new Date(item.get('end')));   # after apply this commit.
      item.save().then(function() {
        self.set('isEditing', false);
      });
    }
  }
});
```

### DIFF
--- a/packages/ember-data/lib/transforms/date.js
+++ b/packages/ember-data/lib/transforms/date.js
@@ -38,8 +38,12 @@ export default Transform.extend({
   },
 
   serialize: function(date) {
+    var type = typeof date;
+
     if (date instanceof Date) {
       return date.toISOString();
+    } else if (type === "string") {
+      return new Date(Ember.Date.parse(date)).toISOString();
     } else {
       return null;
     }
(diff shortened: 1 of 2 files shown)


Eric-Guo:
failed in `npm run-script test:canary` not my fail, maybe it's just not right time...

Eric-Guo:
@bmac can now accept my first PR?

bmac:
Hi @Eric-Guo. I agree it's frustrating not being able to bind a date attribute directly to a date input. However, I'm not sure that changing the date transform to serialize strings is the best solution to this problem. 

While it is a simple fix, it means a date attribute can contain 2 types of values (strings and date object). This can cause problems with code that consumes the date attribute (like computed properties or template helpers) as it now must work with both strings and date objects.

Eric-Guo:
Sounds reasonable, thanks.

PR #3660, opened 2015-08-17
Added documentation to modify the hash by reference when working with `serializeIntoHash`

rtablada:
When working with `serializeIntoHash` if you want to modify the hash with a full set of other results, you are forced to either overwrite each individual key OR use something like `_.extend` to modify the `hash` argument in place using pass by reference.

This is a common use case when creating adapters for APIs that do not have a rootKey (though strongly discouraged).

I think this addition (or something to the effect should be added to make sure that this functionality is documented.

### DIFF
--- a/packages/ember-data/lib/serializers/json-serializer.js
+++ b/packages/ember-data/lib/serializers/json-serializer.js
@@ -988,6 +988,7 @@ var JSONSerializer = Serializer.extend({
     the payload and just sends the raw serialized JSON object.
     If your server expects namespaced keys, you should consider using the RESTSerializer.
     Otherwise you can override this method to customize how the record is added to the hash.
+    The hash property should be modified by reference (possibly using something like _.extend)
 
     For example, your server may expect underscored root objects.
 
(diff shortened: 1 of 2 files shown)


packages/ember-data/lib/serializers/json-serializer.js
```diff
@@ -988,6 +988,7 @@ var JSONSerializer = Serializer.extend({
     the payload and just sends the raw serialized JSON object.
     If your server expects namespaced keys, you should consider using the RESTSerializer.
     Otherwise you can override this method to customize how the record is added to the hash.
+    The hash property should be modified by reference (possibly using something like _.extend)
```
bmac: I like this addition to the docs but I'd prefer to see "(possibly using something like _.extend)" removed. "possibly using something like" doesn't fit the "voice" of the docs.

rtablada: Agreed, just a first pass. Let me commit and smash

bmac:
ping @rtablada. Do you have time to update this pr?

bmac:
This pr was merged as https://github.com/emberjs/data/commit/63e94324a468a113b4ade8d3385689fb9c29f9b6

PR #3665, opened 2015-08-17
[BUGFIX beta] Fix usage of registry for 2.1.0+.

rwjblue:
Using `registry.resolve` is generally private API, and `container.lookupFactory` provides roughly the same result.

We still need the registry to actually register the model we create, so I added a small matrix inline (this was the only usage of accessing the registry off of `this.container` that I found in the codebase).

### DIFF
--- a/packages/ember-data/lib/system/store.js
+++ b/packages/ember-data/lib/system/store.js
@@ -1475,8 +1475,11 @@ Store = Service.extend({
 
   _modelForMixin: function(modelName) {
     var normalizedModelName = normalizeModelName(modelName);
-    var registry = this.container._registry ? this.container._registry : this.container;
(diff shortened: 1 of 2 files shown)


rwjblue:
I added another commit that is unrelated to the registry usage issue here.  It was required to fix the beta/canary test suite...

bmac:
Thanks @rwjblue.

thec0keman:
@bmac would it be possible to get this ported back into the 1.13.x branch?

bmac:
@thec0keman Just so I understand would the purpose of back porting this fix to the 1.13.x branch be so it can work with Ember 2.1 or is there another issue here?

thec0keman:
Correct, so that 1.13.x could work on Ember 2.1.

PR #3670, opened 2015-08-18
[DOC beta] Change to decorator-style CPs in RESTAdapter API doc samples

greyhwndz:
Proposing a change to decorator-style CPs in an example in the RESTAdapter api docs.


bmac:
Thanks.

greyhwndz:
You're welcome

PR #3675, opened 2015-08-19
Fix: IE11 compatibility of JSONAPI adapter

spieker:
I know that the [JSON API](http://jsonapi.org/format/#crud-updating) spec uses the `PATCH` method for updating a record. Since the IE is still not supporting the `PATCH` method, I suggest to use the `PUT` method instead of `PATCH` but giving the user the ability to change this behaviour by setting a property on the `ApplicationAdapter`.

In this pull request I have introduced the property `patchMethod` which is set to `PUT` by default.

What do you think, is this a suitable way to deal with this problem?

### DIFF
--- a/packages/ember-data/lib/adapters/json-api-adapter.js
+++ b/packages/ember-data/lib/adapters/json-api-adapter.js
@@ -14,6 +14,16 @@ export default RESTAdapter.extend({
   defaultSerializer: '-json-api',
 
   /**
+    Regarding to the JSON API (and HTTP) spec, changes should be sent to the
+    server using the `PATCH` method. Since IE11 is still not supporting `PATCH`
+    it is necessary to use `PUT` to make an app work with the IE.
+
+    This property enables the user to change this incorrect behavior if an
+    IE compatibility is not necessary.
+  */
+  patchMethod: 'PUT',
+
+  /**
     @method ajaxOptions
     @private
     @param {String} url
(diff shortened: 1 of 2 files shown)


wecc:
> Since the IE is still not supporting the PATCH method 

From what I've read IE _does_ support `PATCH` (under the correct circumstances) and in this case it's probably more of a jQuery issue (switching between XHR and ActiveX) than a IE issue. `PATCH` seem to be working find in IE using the jQuery Compat Edge branch.

We do not want `PUT` as a default since that would not only break people's apps but also make our default implementation not follow the JSON API specification.

I know @pangratz has been working on making it easier to override and customize the HTTP verb for all AJAX calls made by Ember Data. I think that's a better approach to solve this than a specific override just for the `PATCH` request in `JSONAPIAdapter`.

pangratz:
Yep, the PR is at #3099. I try to find some time in the evening to rebase the PR onto latest master...

spieker:
I see your point.

If others run into the same problem, this is how we solved it for now

```javascript
ApplicationAdapter = DS.JSONAPIAdapter.extend({
  ajax: function(url, type, options) {
    if (type === 'PATCH') {
      type = 'PUT';
    }
    return this._super(url, type, options);
  }
});
```

bmac:
Hi @spieker. I'm going to close this pr as the Ember Data team would like the defaults to encourage users to follow the JSON API spec. If users need to use "PUT" they can override the `ApplicationAdapter` as you have shown in your example.

Feel free to re-open this issue if I'm missing something.

spieker:
It's fine, thanks!

Issue #3644, opened 2015-08-12
complex InternalModel attributes may want to become lazy


stefanpenner:
* https://github.com/emberjs/data/blob/master/packages/ember-data/lib/system/model/internal-model.js#L51
* https://github.com/emberjs/data/blob/master/packages/ember-data/lib/system/model/internal-model.js#L56-L58
* https://github.com/emberjs/data/blob/master/packages/ember-data/lib/system/model/internal-model.js#L90

stefanpenner:
some more details.

currently we have:

```js
export default function InternalModel(type, id, store, container, data) {
  // ...
  this._data = data || new EmptyObject();
  this._attributes = new EmptyObject();
  this._deferredTriggers = [];
  this._attributes = new EmptyObject();
  this._inFlightAttributes = new EmptyObject();
  this._relationships = new Relationships(this);
  //..
}
```

If for complex attributes that not every InternalModel uses upfront, we made them lazy. 

First step, shape the object:

```js
export default function InternalModel(type, id, store, container, data) {
  // ...
  this._data = data || undefined;
  this._attributes = undefined;
  this._deferredTriggers = undefined;
  this._attributes = undefined;
  this._inFlightAttributes = undefined;
  this._relationships = undefined;
  //..
}
```

then make accessors for each property, something like that.

```js
InternalModel.prototype.getData = function() {
  if (this._data) { return this._data;}
  return (this._data = new EmptyObject());
};
```

Then update internal usage accordingly.

We can mitigate unused early allocations.

pangratz:
Closing due to the work done by @runspired in #4562 🚀

PR #3689, opened 2015-08-22
Backport empty-object performance improvements from 2.1.0-beta1

jnfisher:
We found these changes substantially improved loading of large datasets in our app.

This PR contains code from the following 2.1.0-beta1 PRs:
#3641 Improve InternalModel (2x - 3x faster)
#3649 Empty object

### DIFF
--- a/packages/ember-data/lib/adapters/rest-adapter.js
+++ b/packages/ember-data/lib/adapters/rest-adapter.js
@@ -12,6 +12,7 @@ import {
 import {
   MapWithDefault
 } from "ember-data/system/map";
+import EmptyObject from "ember-data/system/empty-object";
 import ArrayPolyfills  from 'ember-data/ext/ember/array';
 
 var get = Ember.get;
@@ -19,7 +20,6 @@ var set = Ember.set;
 var forEach = ArrayPolyfills.forEach;
 
 import {
-  create,
   keysFunc
 } from 'ember-data/system/object-polyfills';
 
@@ -1042,7 +1042,7 @@ var RestAdapter = Adapter.extend(BuildURLMixin, {
 });
 
 function parseResponseHeaders(headerStr) {
-  var headers = create(null);
+  var headers = new EmptyObject();
   if (!headerStr) { return headers; }
 
   var headerPairs = headerStr.split('\u000d\u000a');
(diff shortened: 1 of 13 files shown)


stefanpenner:
> We found these changes substantially improved loading of large datasets in our app.

I would be curious to know the #s you have seen in your app with your data.

jnfisher:
In our app we have a large, hierarchical tree of data that is pulled and deserialized when users visit the site.  

To test, we monkey patched these two PRs into the latest 1.13 build of Ember-Data in our app and found loading times improved by 50% (for a dataset containing on the order of hundreds of complex DS.Models with relationships the time to load went from 40s to 20s, roughly).  We still need to make it faster but this was a nice bump for a simple change!

stefanpenner:
@jnfisher good to hear, those numbers still aren't very but our headroom for improvement here is still very high.

It is also worth noting, those two optimizations had a third counter part, one that made OrderSet and Set insertion much faster.
https://github.com/emberjs/ember.js/pull/12061

Which was back-ported to 1.13-stable (but not yet part of a 1.13 release). 

I am curious if you are also utilizing that.

stefanpenner:
@jnfisher when i took a look at this, i also noticed InternalModel tends to allocate many members upfront, event thought not all of them are always needed.

Related issue (I also added some ideas on how to implement) https://github.com/emberjs/data/issues/3644

jnfisher:
I don't recall if we were using stable or release in our testing (we're still on a branch until we clear out some remaining problems with stained-by-children and fragments)-- but we'll keep an eye on it.  

Agree that 10s+ load times are not great.  Fortunately, our users accept waiting for an upfront loading time if the app is fast afterwards.  It's a work in progress.

rwjblue:
> emberjs/ember.js#12061 which was back-ported to 1.13-stable (but not yet part of a 1.13 release)

It was included in 1.13.8.

stefanpenner:
@bmac @igorT @fivetanley @wecc r?

igorT:
I'm a bit concerned we leak NullObjects to user space. People might be surprised by that? @stefanpenner any thoughts? Should be simple to wrap `changedAttributes` to avoid that, and the perf impact would be minimal

PR #3701, opened 2015-08-26
[BUGFIX release] fix regression where dynamically set id is not serialized in a create…

### LINKED ISSUE #3694: Dynamically set DS.Model id's don't seem to be serialized in a createRecord POST request
[JSBin example](http://emberjs.jsbin.com/tijuwoveyu/1/edit?html,js,console,output)

I was working on one of my applications and it appears, since ember-data 1.0.0-beta.19, dynamically set id's are no longer being serialized with a POST request.

In that JSBin example, the input field value is mapped directly to the model ID, and when "save" is clicked, the mocked request outputs the POST data. In Ember-Data version 1.0.0-beta.18, "id" is included. Switching it to any version above that (I tried release, beta, and canary versions of ember-data) results in the same issue.

I tried messing around with the internal ember-data code in my applications and as I was able to find out it has something to do with the internal model's ID not being set correctly before a save occurs.

If I did the model creation/save process wrong in the JSBin, I would appreciate feedback so that I can fix it in my real application.

acburdine:
… save request

closes #3694
- converts the 'id' property to a computed property alias of internalmodel.id

### DIFF
--- a/packages/ember-data/lib/system/model/internal-model.js
+++ b/packages/ember-data/lib/system/model/internal-model.js
@@ -112,7 +112,6 @@ InternalModel.prototype = {
     // lookupFactory should really return an object that creates
     // instances with the injections applied
     this.record = this.type._create({
-      id: this.id,
       store: this.store,
       container: this.container,
       _internalModel: this,
@@ -542,9 +541,9 @@ InternalModel.prototype = {
   },
 
   setId: function(id) {
+    var oldId = this.id;
+    Ember.assert('A record\'s id cannot be changed once it is in the loaded state', oldId === null || oldId === id || this.isNew());
     this.id = id;
-    //TODO figure out whether maybe we should proxy
-    set(this.record, 'id', id);
   },
 
   didError: function(error) {
--- a/packages/ember-data/lib/system/model/model.js
+++ b/packages/ember-data/lib/system/model/model.js
@@ -292,7 +292,6 @@ var Model = Ember.Object.extend(Ember.Evented, {
     @property id
     @type {String}
   */
-  id: null,
 
   /**
     @property currentState
@@ -769,6 +768,7 @@ var Model = Ember.Object.extend(Ember.Evented, {
   willMergeMixin: function(props) {
     var constructor = this.constructor;
     Ember.assert('`' + intersection(Object.keys(props), RESERVED_MODEL_PROPS)[0] + '` is a reserved property name on DS.Model objects. Please choose a different property name for ' + constructor.toString(), !intersection(Object.keys(props), RESERVED_MODEL_PROPS)[0]);
+    Ember.assert("You may not set `id` as an attribute on your model. Please remove any lines that look like: `id: DS.attr('<type>')` from " + constructor.toString(), Object.keys(props).indexOf('id') === -1);
   },
 
   attr: function() {
@@ -838,6 +838,21 @@ Model.reopenClass({
    @readonly
   */
   modelName: null
+});
+
+Object.defineProperty(Model.prototype, 'id', {
+  configurable: true,
+  enumerable: false,
+  set(id) {
+    if (this._internalModel) {
+      this._internalModel.setId(id);
+    }
+  },
+  get() {
+    if (this._internalModel) {
+      return this._internalModel.id;
+    }
+  }
 });
 
 export default Model;
--- a/packages/ember-data/lib/system/model/attributes.js
+++ b/packages/ember-data/lib/system/model/attributes.js
@@ -323,7 +323,6 @@ export default function attr(type, options) {
       }
     },
     set: function(key, value) {
-      Ember.assert("You may not set `id` as an attribute on your model. Please remove any lines that look like: `id: DS.attr('<type>')` from " + this.constructor.toString(), key !== 'id');
       var internalModel = this._internalModel;
       var oldValue = getValue(internalModel, key);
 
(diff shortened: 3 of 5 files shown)


packages/ember-data/lib/system/model/model.js
```diff
@@ -295,7 +295,7 @@ var Model = Ember.Object.extend(Ember.Evented, {
     @property id
     @type {String}
   */
-  id: null,
+  id: Ember.computed.alias('_internalModel.id'),
```
bmac: I'm worried about this being a breaking changed. It is possible to access a record's id without using a getter (`record.id`).  Maybe we want to use an ES5 getter to grab the id out of the internal model?

acburdine: Would that work for setting the id as well? The main cause of the regression is that the id is not being set correctly when it is modified after creation.

wecc:
Won't this mess up the typeMaps in store? I know there are issues related to this and it would be great to get it fixed, just want to make sure we're solving it in a reliable way.

Ping @igorT for review.

bmac:
@wecc I believe the store already has code to update the typeMaps after the `createRecord` record request is resolved. @acburdine's usecase of setting the id of a new record on the client side use to work before Ember Data beta.19.

packages/ember-data/lib/system/model/model.js
bmac: I think it would work for setting the id as well if we also implemented an ES5 setter.

acburdine:
@bmac Anything else I need to do for this or is it just waiting on review?

packages/ember-data/lib/system/model/internal-model.js
```diff
@@ -541,9 +540,8 @@ InternalModel.prototype = {
   },
 
   setId: function(id) {
+    Ember.assert('Object is a newly created model instance', this.isNew);
```
bmac: `isNew` is a function. `Ember.assert('Object is a newly created model instance', this.isNew());`

bmac: Also lets make the assertion message say something like "A record's id can not be chanced once it is in the loaded state."

packages/ember-data/lib/system/model/internal-model.js
```diff
@@ -541,9 +540,8 @@ InternalModel.prototype = {
   },
 
   setId: function(id) {
+    Ember.assert('A record\'s id cannot be changed once it is in the loaded state', this.isNew());
```
bmac: Looks like there are a few code paths that try to set a record with an id to the same id. Changing the assertion logic to the following should work.

```js
var oldId = this.id;
Ember.assert('A record\'s id cannot be changed once it is in the loaded state', oldId === null || id === oldId || this.isNew());
```

packages/ember-data/lib/system/model/model.js
```diff
@@ -295,7 +295,13 @@ var Model = Ember.Object.extend(Ember.Evented, {
...
     @type {String}
   */
-  id: null,
+  get id () {
+    return this._internalModel.id;
+  },
+
+  set id (id) {
+    this._internalModel.setId(id);
+  },
```
bmac: It looks the Ember object model doesn't like it when you use an enumerable computed getter. Instead of using the shorthand get and set syntax we will need to use `Object.defineProperty` at the bottom of this file to set the non enumerable computed id property.

```js
Object.defineProperty(Model.prototype, 'id', {
  configurable: true,
  enumerable: false,
  set(id) {
    if (this._internalModel) {
      this._internalModel.setId(id);
    }
  },
  get() {
    if (this._internalModel) {
      return this._internalModel.id;
    }
  }
});
```

packages/ember-data/lib/system/model/model.js
```diff
@@ -840,4 +839,19 @@ Model.reopenClass({
...
+    if (this._internalModel) {
+      this._internalModel.setId(id);
+    }
+  },
+  get() {
+    if (this._internalModel) {
+      return this._internalModel.id;
+    }
+  }
+})
```
bmac: jshint wants a semicolon here.

bmac:
@acburdine Looks like these changes have caused [this test](https://github.com/bmac/data/blob/9fb8e4319cf0d820a29b2fac028a611ba0ff1c5c/packages/ember-data/tests/unit/model-test.js#L153-L177) to fail. 

I think the simplest fix would be to add this assertion to the end of the model's [willMergeMixin method](https://github.com/emberjs/data/blob/71f8cbdc8e09f1df30aa8709045d350abb685109/packages/ember-data/lib/system/model/model.js#L771). 

```js
Ember.assert("You may not set `id` as an attribute on your model. Please remove any lines that look like: `id: DS.attr('<type>')` from " + constructor.toString(), Object.keys(props).indexOf('id') === -1);
```

bmac:
ping @igorT I'd like you to review this pr if you have time.

packages/ember-data/lib/system/model/model.js
```diff
@@ -840,4 +840,19 @@ Model.reopenClass({
...
+Object.defineProperty(Model.prototype, 'id', {
+  configurable: true,
+  enumerable: false,
+  set(id) {
+    if (this._internalModel) {
+      this._internalModel.setId(id);
+    }
+  },
+  get() {
+    if (this._internalModel) {
```
igorT: This should never be false

PR #3713, opened 2015-08-27
Document the return value of DS.Adapter#updateRecord

### LINKED ISSUE #3712: it's unclear what the resolved value that DS.Adapter#updateRecord suppose to return
i notice it works with the promise when resolved with `undefined`, but it's unclear what it suppose to resolve if some data changed in the model because of the save

bmac:
Closes #3712

### DIFF
--- a/packages/ember-data/lib/system/adapter.js
+++ b/packages/ember-data/lib/system/adapter.js
@@ -330,6 +330,14 @@ export default Ember.Object.extend({
 
     Serializes the record update and sends it to the server.
 
+    The updateRecord method is expected to return a promise that will
+    resolve with the serialized record. This allows the backend to
+    inform the Ember Data store the current state of this record after
+    the update. If it is not possible to return a serialized record
+    the updateRecord promise can also resolve with `undefined` and the
+    Ember Data store will assume all of the updates were successfully
+    applied on the backend.
+
     Example
 
     ```app/adapters/application.js


packages/ember-data/lib/system/adapter.js
```diff
@@ -330,6 +330,14 @@ var Adapter = Ember.Object.extend({
 
     Serializes the record update and sends it to the server.
 
+    The updateRecord method is expected to return a promise that will
+    resolve with the serialized record. This allows the backend to
```
oriSomething: hey, it's actually worked for me with **pre** serialized data instead of serialized data (the raw object before normalizing the model by serializer)? am i did something wrong or is it mistake in the docs?

oriSomething:
also it seems relations are not effected from the process... again i'm not sure it's me or maybe it's undocumented stuff

packages/ember-data/lib/system/adapter.js
bmac: That is correct. Maybe I'm using the wrong terminology. I think of the raw object before normalizing the model by serializer as the "serialized data" and the json object after it has been run through `normalizeResponse` to be the "normalized data".

bmac:
@oriSomething. Relationships can be effected by this process if the backend supports sideloading in its serialized payload.

For example with json API any relationships returned by the backend in the `included` property would also be updated in the store.

Issue #3691, opened 2015-08-24
FixtureAdapter still being imported despite being made as an addon


greyhwndz:
The FixtureAdapter is still being imported from this line:
- https://github.com/emberjs/data/blob/master/packages/ember-data/lib/main.js#L60
despite already having this:
- https://github.com/emberjs/ember-data-fixture-adapter
Would like to clarify if this needs to be removed or is being retained for a transition towards total deprecation...

PR #3717, opened 2015-08-30
Remove Fixture Adapter

greyhwndz:
@bmac: Would be glad to hear some feedback regarding this PR. Thanks!


bmac:
Could you also remove this file? https://github.com/emberjs/data/blob/master/packages/ember-data/lib/adapters/fixture-adapter.js

greyhwndz:
@bmac: Done.

bmac:
Oops, looks like the fixture adapter also needs to be emoved from this file: https://github.com/emberjs/data/blob/7b2c70d5e454a485eae4d64c7cdabfa020208b9d/packages/ember-data/lib/adapters.js

greyhwndz:
@bmac: Just finished removing the references to FixtureAdapter in lib/adapter.js

greyhwndz:
@bmac: This should address issue #3691

PR #3722, opened 2015-09-03
Shape and cleanup

stefanpenner:


stefanpenner:
ah, more things are needed... 1 sec

PR #3725, opened 2015-09-03
[BUGFIX release-1-13] Disable polymorphic deserialization when a mode…

### LINKED ISSUE #3702: DS.Model type property name in non-polymorphic models
I upgraded Ember data from 1.13.x to **1.13.11** in my app and got a weird error in my serializer (REST) in decimalizing a number property's value.

I debugged it and saw, that it's related to the fact that I use a property named `type` in the model. It's just a numeric value, it's not a  polymorphic subclass identifier.

    type: DS.attr('number')

The same code used to work in a prior Ember data version. Is this a bug? If not, is there some documentation on this? I have a lot of model which have a `type` property.

### LINKED ISSUE #3724: _normalizePolymorphicRecord in RESTSerializer breaks old REST API when model has type attribute
I am using ember-data 2.0.0

When using the old REST API,  if a model has an attribute named type which can have a value that matches the name of another model, then the model instance may be loaded with the wrong model.

_normalizePolymorphicRecord tries to detect if a payload is using  JSON API or REST API, by checking if the hash has a type attribute that matches a model name.  But, it is possible for this to occur when using the REST API as well.  So, payloads from the REST API are sometimes erroneously parsed as JSON API payloads, causing them to be loaded into the store with the wrong model.

Since I don't use the JSON API,  I have worked around this issue in my application by adding the below to my application serializer.

```js
    _normalizePolymorphicRecord: function (store, hash, prop, modelClass, serializer) {
        return serializer.normalize(modelClass, hash, prop);
    }
```

XrXr:
…l expects a type attribute

b7f7b7a introduced a bug where if there is a type key in a payload that is part
of an array, it would be used for polymorphic deserialization even when the
model expects an attribute that is named "type".

Interestingly, for such payloads, `arrayHash` passed into `normalizeArray()` in
rest-serializer.js contains Ember.Object instances as opposed to plain objects.
This causes the code to throw, since `hash.type` would be a computed property
in that case instead of a string that `dasherize` expects. This is probably because
of extending from JSONAPISerializer.

Fixes #3702.

It probably also fixes #3724, but needs a port to 2.0

### DIFF
--- a/packages/ember-data/lib/serializers/rest-serializer.js
+++ b/packages/ember-data/lib/serializers/rest-serializer.js
@@ -228,9 +228,10 @@ var RESTSerializer = JSONSerializer.extend({
 
     Ember.assert(`${this.toString()} has opted into the new serializer API and expects the ${serializer.toString()} it collaborates with to also support the new serializer API by setting its \`isNewSerializerAPI\` property to true.`, get(serializer, 'isNewSerializerAPI'));
 
+    const primaryHasTypeAttribute = get(modelClass, 'attributes').get('type');
     /*jshint loopfunc:true*/
     forEach.call(arrayHash, (hash) => {
-      let { data, included } = this._normalizePolymorphicRecord(store, hash, prop, modelClass, serializer);
+      let { data, included } = this._normalizePolymorphicRecord(store, hash, prop, modelClass, serializer, primaryHasTypeAttribute);
       documentHash.data.push(data);
       if (included) {
         documentHash.included.push(...included);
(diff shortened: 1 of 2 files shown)


bmac:
Thanks @XrXr. This pr looks great. @wecc or @igorT could you also review this?

wecc:
LGTM :+1:

jmurphyau:
Something worth considering that _looks like_ it could still be an issue: consider the scenario where the a `Product` has a `productType` attribute but it's coming from the server as `type` - the serializer would know to pull the payload datas `type` property into the models `productType` attribute.

It looks like `primaryHasTypeAttribute` will be false here a user may hit this same issue..

PR #3732, opened 2015-09-08
follow up fixes for #3701

acburdine:
cc @stefanpenner @igorT This fixes the two comments that were made on the PR after it was merged.


bmac:
Thanks @acburdine

PR #3744, opened 2015-09-10
[BUGFIX beta] Fix Model `id` handling, take 2

acburdine:
Fixes the thing that should have been fixed in #3701

### DIFF
--- a/packages/ember-data/lib/system/model/model.js
+++ b/packages/ember-data/lib/system/model/model.js
@@ -781,6 +781,25 @@ var Model = Ember.Object.extend(Ember.Evented, {
 
   hasMany: function() {
     Ember.assert("The `hasMany` method is not available on DS.Model, a DS.Snapshot was probably expected. Are you passing a DS.Model instead of a DS.Snapshot to your serializer?", false);
+  },
+
+  // handlers for the id property
+  unknownProperty: function (key) {
+    if (key === 'id') {
+      return this._internalModel.id;
+    } else {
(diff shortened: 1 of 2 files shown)


acburdine:
cc @bmac @igorT 

I have no clue why the tests are failing, they worked locally

packages/ember-data/lib/system/model/model.js
```diff
@@ -781,6 +781,25 @@ var Model = Ember.Object.extend(Ember.Evented, {
 
   hasMany: function() {
     Ember.assert("The `hasMany` method is not available on DS.Model, a DS.Snapshot was probably expected. Are you passing a DS.Model instead of a DS.Snapshot to your serializer?", false);
+  },
+
+  // handlers for the id property
+  unknownProperty: function (key) {
```
bmac: I think this way requires uses to do `record.get('id')` to access the id. I was hoping to support `record.id` because it previously worked before the internal model refactor but maybe thats not possible :(.

acburdine: Well, this unknown property only needs to be in place until ember 2.2, as I think the object.defineProperty way should work then via rwjblue's fix

acburdine:
@bmac if I add an observer to the `id` property instead of using an unknown property, that's the other way in which I was able to get it to work.

bmac:
Superseded by https://github.com/emberjs/data/pull/3781.

Issue #3704, opened 2015-08-26
findRecord with { reload: true } does not reload cached record


irnc:
This regression was introduced in https://github.com/emberjs/data/commit/0ac6ba4f331eecf3a3ddddccd76808520fbc7a45 for 2.0 and in https://github.com/emberjs/data/commit/75bec5ec1b1b842ee87eb4cb2fb0b6dbe1b93d85 for 1.13 bracnhes.

In current code, `_findRecord` function checks `reload` option and schedules fetch if it is `true`, but this function is never called if record is already in cache.

You can see that `Store#findRecord` would never reach the call to `_findRecord` for cached records and would return result of `_findByInternalModel` which does not check for `reload` option.

bmac:
Oops. How did I miss that? Thanks for reporting this issue @irnc.

visoft:
Is there a fix for this? I discovered the same bug that @irnc found. I'm on the 1.13.13 release of Ember Data.

graup:
I just ran into this problem and would be willing to contribute a fix. But I wonder where that should go? Just add a check for `&& !options.reload` in https://github.com/emberjs/data/blob/75bec5ec1b1b842ee87eb4cb2fb0b6dbe1b93d85/packages/ember-data/lib/system/store.js#L624 ?

visoft:
@graup - I tried that fix locally. It does go through and call `scheduleFetch`, but it seems to still grab it from the cache. At least in my testing, it doesn't go out and call the server again.

tuomole:
We're currently facing this problem, as we need to reload a cached record before editing it. As a dirty workaround we do `model.reload()` in route's `afterModel` hook, which causes 2 requests to the server when there's no record in the cache already.

Any updates about the proposed fix?

Ember 1.13.7
Ember Data 1.13.8

DLiblik:
We implemented the same workaround (calling reload), but get a double-tap to the server as well... hoping somebody has nothing better to do this weekend ;)

pangratz:
I cannot reproduce using `ember-data` `v1.13.15`: https://ember-twiddle.com/5e7aad8a008b29029d06?numColumns=1&openFiles=application.route.js

Someone still hit this bug?

vvainio:
I bumped into this issue on the aforementioned `ember-data v1.13.15`. Resorting to calling `model.reload()` until this is resolved.

pangratz:
[...]

PR #3746, opened 2015-09-10
Improve test coverage for store#findRecord()

HeroicEric:
Started working on fixing https://github.com/emberjs/data/issues/3704 by trying to write a failing test but was not able to reproduce the issue.

Am I missing something?

### DIFF
--- a/packages/ember-data/tests/integration/store-test.js
+++ b/packages/ember-data/tests/integration/store-test.js
@@ -211,37 +211,35 @@ function ajaxResponse(value) {
 
 
 
-module("integration/store - findRecord { reload: true }", {
-  setup: function() {
-    initializeStore(DS.RESTAdapter.extend());
-  }
+module("integration/store - findRecord", {
 });
 
-test("Using store#findRecord on non existing record fetches it from the server", function() {
+test("store#findRecord fetches record from server when cached record is not present", function() {
   expect(2);
+
+  initializeStore(DS.RESTAdapter.extend());
 
   env.registry.register('serializer:application', DS.RESTSerializer);
   ajaxResponse({
     cars: [{
       id: 20,
-      make: 'BMCW',
+      make: 'BMC',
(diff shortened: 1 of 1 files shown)


wecc:
LGTM

HeroicEric:
@irnc am I misunderstanding the bug you were running into?

packages/ember-data/tests/integration/store-test.js
```diff
@@ -254,22 +253,58 @@ test("Using store#findRecord on existing record reloads it", function() {
...
+    });
+  });
+
+  run(function() {
+    let car = store.peekRecord('car', 1);
+    equal(car.get('model'), 'Princess', 'car record was reloaded');
+  });
+});
+
+test("store#findRecord { reload: true } ignores cached record and reloads record from server", function() {
```
bmac: @HeroicEric in this test could you also assert that `shouldReloadRecord` is never called when `{ reload: true }` is passed into findRecord?

bmac:
Thanks @HeroicEric this looks great!

packages/ember-data/tests/integration/store-test.js
HeroicEric: added assertion at https://github.com/emberjs/data/pull/3746/files#diff-a40f0b2297a7d2fb4022eb1fcae21af9R280

bmac:
Thanks @HeroicEric

irnc:
@HeroicEric Tests look good. I should have misread code, so my report is incorrect, based on how I read it right now.

But there is definitely an issue for my app, which I can reproduce, but can not debug it right now because Chrome does not stop on breakpoint in `store.js` (which is server in ES6 from source maps for some reason).

bmac:
@irnc try disabling sourcemaps in your chrome dev tools.

screen shot 2015-09-11 at 2 37 39 pm

Issue #3736, opened 2015-09-09
JSONSerializer.extractRelationships() is now failing when a relationship is not set in the model


sebweaver:
Hello,

I added a commented on the PR which has introduced this issue (worked previously in ED 1.13). Please see https://github.com/emberjs/data/pull/3345#diff-for-comment-38619627 for more detail.

Sébastien

wecc:
I'm unsure how you can end up in that code path when the key is missing in the payload, there's a check for this here: https://github.com/emberjs/data/pull/3345#discussion-diff-38619627R600

wecc:
Hm, I'm a bit confused. `store.push()` doesn't use serializers so the change in JSONSerializer shouldn't be affecting the issue you're describing.

wecc:
Can you provide a minimal JSBin or Embed Twiddle demonstrating the issue you're experiencing?

sebweaver:
As a matter of fact the problem has nothing to do with the `push` but with the `normalize` (when you fetch data with undefined relationships).

I edited my comment https://github.com/emberjs/data/pull/3345#diff-for-comment-38619627 and added the 2 following JSBins:

* ED 1.13: http://emberjs.jsbin.com/puguhi/edit?html,js,console,output
* ED 2.0: http://emberjs.jsbin.com/towofa/edit?html,js,console,output

wecc:
Yeah, this definitely looks like a bug. Would you be able to provide a PR fixing it?

sebweaver:
Done.

pangratz:
This issue can be closed since https://github.com/emberjs/data/pull/3747 got merged ...

Turbo87:
I'm getting very confused by this change... Is there any documentation on the expected behavior when a relationship key is `null`? I would assume that it does not change anything, but it seems that in v1.13 it actually empties the array.

see https://github.com/rust-lang/crates.io/pull/208#issuecomment-155714465 for a more detailed description of the bug

PR #3747, opened 2015-09-10
[BUGFIX beta] JSONSerializer.extractRelationships() issue #3736

sebweaver:
`JSONSerializer.extractRelationships()` sets the undefined hasMany relationships in the
same way it was for the belongsTo relationships, i.e. to `null`.

I also enriched the corresponding test accordingly in order to cover the case.


wecc:
Would you be able to copy the test you modified so we're still testing with a non-existent relationship key?

sebweaver:
Both cases were indeed in the same test. For better granularity I have now splitted it into two separate tests.

sebweaver:
Appveyor build failed due to the following error:

    npm ERR! network read ECONNRESET

Do I have to push an empty commit to force rebuild? Or is it unnecessary?

wecc:
Looks good, can you squash the commits and prefix with `[BUGFIX beta]`?

bmac:
Thanks @sebweaver.

PR #3762, opened 2015-09-13
[BUGFIX release] Improve links vs. local data for relationships

### LINKED ISSUE #3380: JSON-API correct handling of included entities
The implementation currently always loads `related` entities from the provided url. If there are ids inside a `data` object in a relation that match the ids of objects inside the `included` object, those already loaded objects must be taken instead of requesting them from the `related` url, because that is not necessary. 

basic example

```json
{
    "data": [
        {
            "type": "posts",
            "id": "1",
            "attributes": {
                "title": "JSON API paints my bikeshed!"
            },
            "relationships": {
                "author": {
                    "links": {
                        "self": "http://example.com/posts/1/relationships/author",
                        "related": "http://example.com/posts/1/author"
                    },
                    "data": {
                        "type": "people",
                        "id": "9"
                    }
                }
            },
            "links": {
                "self": "http://example.com/posts/1"
            }
        }
    ],
    "included": [
        {
            "type": "people",
            "id": "9",
            "attributes": {
                "first-name": "Dan",
                "last-name": "Gebhardt",
                "twitter": "dgeb"
            },
            "links": {
                "self": "http://example.com/people/9"
            }
        }
    ]
}
```

here, the author got already provided in the `included` object and does not need to be fetched.

wecc:
This improves the way we fetch links depending on if we have local data or not.

Fixes #3380

### DIFF
--- a/packages/ember-data/lib/system/relationships/state/relationship.js
+++ b/packages/ember-data/lib/system/relationships/state/relationship.js
@@ -16,6 +16,7 @@ export default function Relationship(store, record, inverseKey, relationshipMeta
   this.linkPromise = null;
   this.meta = null;
   this.hasData = false;
+  this.hasLoaded = false;
 }
 
 Relationship.prototype = {
@@ -190,6 +191,7 @@ Relationship.prototype = {
     if (link !== this.link) {
       this.link = link;
       this.linkPromise = null;
+      this.setHasLoaded(false);
       this.record.notifyPropertyChange(this.key);
     }
   },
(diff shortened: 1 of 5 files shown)


wecc:
If possible this should be cherry-picked to 1.13

packages/ember-data/lib/system/relationships/state/relationship.js
```diff
@@ -216,5 +219,9 @@ Relationship.prototype = {
 
   setHasData: function(value) {
     this.hasData = value;
+  },
+
+  setHasLoaded: function(value) {
+    this.hasLoaded = value;
```
bmac: What is the difference between `hasLoaded` and `hasData`?

wecc: `hasData` is "we consider the state of this relationship known"
`hasLoaded` is "the adapter has given us the content of this relationship"

They look very similar but have different purposes.

```
let post = store.createRecord('post', { id: 'dummy' });

// all relationships for post are now known, hasData = true,
// since the record was created locally. but hasLoaded = false

store.pushPayload('post', { id: 'dummy', links: { author: 'author/1' }});

// now let's get the author
post.get('author');

// since we do have a link, and hasLoaded is false,
// we know to fetch it even if hasData is true
```

Another example is when a relationship is known (hasData = true) and a link is updated (updateLink sets hasLoaded = false). This way ED know to fetch the link again, even though we have a known state.

bmac: Great. Do you mind adding a comment to the source to explain the difference for future readers?

wecc: I'll fix that!

wecc:
Added comments explaining `hasData` and `hasLoaded`.

john-griffin:
Awesome, thanks :+1:

jonkoops:
Thanks @wecc! I think a cherry-pick for 1.13 would help prevent some confusion for folks looking to upgrade from the betas.

wecc:
@jonkoops yeah, definitely. @bmac has cherry-picked and released this in 1.13.13 :heart:

gkaran:
This seems to break my application. After I moved from version 2.0.1 from 2.0.0 some of my links stopped working. I'm using json-serializer. Will try and identify the issue and provide more information. (Perhaps it is only me)

wecc:
@gkaran a JSBin demonstrating the problem would be great! feel free to open a new issue.

PR #3764, opened 2015-09-13
[BUGFIX beta] Support JSON API links object

### LINKED ISSUE #3588: Support links with href and meta (JSONAPI)
According to the spec http://jsonapi.org/format/#document-links

This kind of links should be handle by ember-data:

```
"links": {
  "related": {
    "href": "http://example.com/articles/1/comments",
    "meta": {
      "count": 10
    }
  }
}
```

At the moment, Ember-data complains with:

`You have pushed a record of type '...' with '...' as a link, but the value of that link is not a string.`

wecc:
This PR introduces a private helper called `_normalizeLinks` that accepts a string or an object and returns an normalized "links object", see http://jsonapi.org/format/#document-links

Currently this helper is only used in the store's `setupRelationships` but we should probably have links objects reach further down in ED in the future (store => relationship => adapter) to allow access to meta and additional information on the links object.

Fixes #3588

### DIFF
--- a/packages/ember-data/lib/system/normalize-link.js
+++ b/packages/ember-data/lib/system/normalize-link.js
@@ -1,0 +1,19 @@
+/**
+  This method normalizes a link to an "links object". If the passed link is
+  already an object it's returned without any modifications.
+
+  See http://jsonapi.org/format/#document-links for more information.
+
+  @method _normalizeLink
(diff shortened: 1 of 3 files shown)


bmac:
Looks good. :+1:

PR #3767, opened 2015-09-14
[BUGFIX release] Do not export jsdoc information for store helper fun…

bmac:
…ctions

Currently these methods are showing up on the DS.Adapter class. However, there is no way for user code to access these methods. http://emberjs.com/api/data/classes/DS.Adapter.html

screen shot 2015-09-14 at 5 06 46 pm


wecc:
:+1:

PR #3770, opened 2015-09-15
update coalesceFindRequests doc for JSONAPIAdapter

drewclauson:
I noticed that the JSONAPIAdapter sends different query params then the RESTAPIAdapter, so took a stab at updating the documentation.  I felt it was better to put the info in the json-api-adapter.js rather than putting a "if you're using JSONAPIAdapter" statement into the rest-adapter.js.  I did override the coalesceFindRequests bool in json-api-adapter.js, but let me know if this is not the desired way to do things at this point.

### DIFF
--- a/packages/ember-data/lib/adapters/json-api-adapter.js
+++ b/packages/ember-data/lib/adapters/json-api-adapter.js
@@ -62,6 +62,55 @@ export default RESTAdapter.extend({
     return Ember.String.pluralize(dasherized);
   },
 
+/**
+    By default the RESTAdapter will send each find request coming from a `store.find`
+    or from accessing a relationship separately to the server. If your server supports passing
+    ids as a query string, you can set coalesceFindRequests to true to coalesce all find requests
+    within a single runloop.
+
+    For example, if you have an initial payload of:
+
+    ```javascript
+    {
+      post: {
+        id: 1,
+        comments: [1, 2]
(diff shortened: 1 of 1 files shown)


packages/ember-data/lib/adapters/json-api-adapter.js
```diff
@@ -62,6 +62,44 @@ export default RESTAdapter.extend({
...
+    within a single runloop.
+    For example, if you have an initial payload of:
+    ```javascript
+    {
+      post: {
+        id: 1,
+        comments: [1, 2]
+      }
+    }
+    ```
```
bmac: Do you mind adding a new line after all these code block sections? The markdown parser for the api docs doesn't format these sections correctly without the new line to separate content after the code block.

drewclauson: Yep, sorry about that, I lost the newlines in copying and pasting.

bmac:
Looks like travis is now complaining about trailing white space in the new docs. Do you mind fixing that and then squashing everything into 1 commit?

drewclauson:
Still working on squashing the commits into one, made a mistake (still a little new to git)

drewclauson:
Closing because I messed up the squashing so badly I don't know how to make it happen. :-)

PR #3781, opened 2015-09-17
[BUGFIX beta] Fix id property in DS.Model once and for all

acburdine:


packages/ember-data/lib/system/model/model.js
```diff
@@ -781,7 +782,11 @@ var Model = Ember.Object.extend(Ember.Evented, {
 
   hasMany: function() {
     Ember.assert("The `hasMany` method is not available on DS.Model, a DS.Snapshot was probably expected. Are you passing a DS.Model instead of a DS.Snapshot to your serializer?", false);
-  }
+  },
+
+  setId: Ember.observer('id', function () {
```
stefanpenner: in the tests bellow, we only update the ID via:

https://github.com/emberjs/data/blob/0f06c86b881ffb22e3e3fd8b2ef61726b9c17b22/packages/ember-data/lib/system/store.js#L1394-L1403

which already calls [internalModel.setId](https://github.com/emberjs/data/blob/0f06c86b881ffb22e3e3fd8b2ef61726b9c17b22/packages/ember-data/lib/system/store.js#L1402)

stefanpenner:
@bmac on id update, do we not update old IDs in the identify map? https://github.com/emberjs/data/blob/0f06c86b881ffb22e3e3fd8b2ef61726b9c17b22/packages/ember-data/lib/system/store.js#L1400

packages/ember-data/lib/system/model/model.js
acburdine: But what if someone needs to edit the id before saving a new model? That worked before 1.0.0-beta.19

stefanpenner: then a getter/setter pair does the trick.

acburdine: if I'm understanding you correctly, that was the first attempt with the whole object.defineproperty thing. That didn't work because of a bug in ember itself.

acburdine: See: https://embercommunity.slack.com/archives/dev-ember-data/p1441825521000097

bmac:
I don't think a record is slotted into the identity map by id until the id has been acknowledged by the server. This scenario is specifically for updating a record in the `new` state before it has been slotted into the identity map by id.

packages/ember-data/lib/system/model/model.js
stefanpenner: ug the internal flow here seems very strange.

It really feels like it should be:

```js

model ... {
  get id()   { return this.internalModel.id; }
  set id(id) { this.internalModel.id = id; }
}
```
```
internalModel .. {
  set id(id) {
    if (this._id === id) { return; }

    this.store.moveToId(this, id);
    if (this.model) {
      this.model.notifyPropertyChange('id');
    }
  },

   get id() {
     return this._id;
   }
}
```

stefanpenner:
strange that unloading a record, transitions it to `saved.deleted`

pangratz:
> strange that unloading a record, transitions it to saved.deleted

just a quick interception: maybe related #2870

stefanpenner:
> just a quick interception: maybe related #2870

ya looks related, it seems sloppy that we are reusing the same state.

packages/ember-data/lib/system/model/model.js
acburdine: That does make sense, but atm it would break when the id property is watched.

acburdine:
so.....what can I do to finish this?

packages/ember-data/lib/system/model/model.js
bmac: @stefanpenner Trying out the code you provided I'm seeing `model.id` return the correct value however, `model.get('id')` appears to return a cached `undefined` value when the id property is watched by the template. I thought `notifyPropertyChange` would invalidate the cache...

acburdine: In the internalModel code, shouldn't `this.model` actually be `this.record`?

bmac:
@stefanpenner I made another attempt at using `id` as a ES5 getter/setter. Here are the 2 blockers I've hit so far.

First I tried using the code you provided and I found that `this.record.notifyPropertyChange('id');` does not correctly update the `_values` object on Ember's meta object when `id` is a watched property. This means doing `record.id` returns the correct value but `record.get('id')` returns the old cached value.

Next I tried updating the code so `store.updateId` calls `record.set('id', newId)` and allowing the computed setter on id to update the internal model. This correctly updates the cached value in the meta object and both `record.id` and `record.get('id')` return the correct value when watched. However the id on internalModel is never updated because Ember ends up [defining its own computed setter here](https://github.com/emberjs/ember.js/blob/a3fb7a1dd8a449bd8a157821789a8641ca079434/packages/ember-metal/lib/property_set.js#L83) that supplants the computed setter defined by ember data.

PR #3787, opened 2015-09-18
Add support for HATEOAS link parsing in JSONSerializer

gkaran:
When using a HATEOAS approach (e.g. with spring HATEOAS) JSONSerializer cannot parse the links of a resource.

This is due to the fact that extractRelationships expects links to be an object where the keys are the model's attributes. This is not the case though when using HATEOAS.

Suppose we have a spring HATEOAS application where users have posts. The response that comes back from the server has the following format:

```javascript
{
  "firstName": "John",
  "lastName": "Doe",
  "username": "john.doe",
  "links": [
    {
      "rel": "posts",
      "href": "user/john.doe/posts"
    }
  ]
}
````

The above links format is not supported from JSONSerializer since it expects an object as:

```javascript
{
  "firstName": "John",
  "lastName": "Doe",
  "username": "john.doe",
  "links": {
    "posts": "user/john.doe/posts"
  }
}
````

This pull request resolves that issue and adds support for HATEOAS links without breaking the existing behavior. Two tests cases have also been added. One for the existing link parsing and one for HATEOAS.

### DIFF
--- a/packages/ember-data/lib/serializers/json-serializer.js
+++ b/packages/ember-data/lib/serializers/json-serializer.js
@@ -605,10 +605,21 @@ var JSONSerializer = Serializer.extend({
       }
 
       let linkKey = this.keyForLink(key, relationshipMeta.kind);
-      if (resourceHash.links && resourceHash.links.hasOwnProperty(linkKey)) {
-        let related = resourceHash.links[linkKey];
-        relationship = relationship || {};
-        relationship.links = { related };
+      if (resourceHash.links) {
+        var related;
+        if (Ember.isArray(resourceHash.links)) {
+          let links = resourceHash.links.filter(link => link.rel === linkKey);
+          if (links && links[0]) {
+            related = links[0].href;
+          }
+        } else if (resourceHash.links.hasOwnProperty(linkKey)) {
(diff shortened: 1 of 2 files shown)


wecc:
I think this would be a great case for an add-on. The `JSONSerializer` is intentionally kept simple as we're focusing on JSON API support in the `JSONAPIAdapter`/`Serializer`. The existing hooks should make it trivial to implement separately.

@bmac @igorT what do you think?

bmac:
Hi @gkaran. Thanks for making this pr. I think it would make a great Ember Data serializer addon. 

Unfortunately, I am going to close this pr. `JSONSerializer` was suppose to be an extremely basic serialize with no special knowledge of any specific backend. It was intended that other Serializer could use `JSONSerializer` as a starting point to extend and implement their backend specific rules and functionality.

The fact that it supports `links` today is somewhat of a historic accident. `JSONSerializer` originally didn't touch links at all and instead the `store.push` format knew how to correctly interpret `links`. When Ember Data transitioned the store to use JSON API Ember Data needed the JSONSerializer to transform the old `links` format to a JSON API format to preserved backwards compatibility.

Issue #3708, opened 2015-08-26
Extending DS.attr getter/setter


rmachielse:
Would it be possible to be able to change the default get/set method for DS.attr?

I would like to be able to modify an attribute before it is being set on the model (and events are fired).
This is comparable to DS.Transform, however transforms are only used during (de)serialization.

I can do this with an observer as well, however that will fire the set event twice since I have to set it again with the modified version.
Also, I could use a second property with get and set methods that use the original attribute, but I think there should be a cleaner approach, like the following:

```
Product = DS.Model.extend({
  price: DS.attr('number', {
    set: function(value) {
      if (typeof(value) === 'string) {
        return value.replace(',', '.');
      } else {
        return value;
      }
    })
  }
});
```

Another maybe even cleaner option would be to extend DS.Transform with get and set methods so I could do:

```
Product = DS.Model.extend({
  price: DS.attr('price')
});

PriceTransform = DS.NumberTransform.extend({
  set: function(value){
    if (typeof(value) === 'string) {
      return value.replace(',', '.');
    } else {
      return value;
    }
  }
});
```

Anyone else interested in this idea?

leimig:
It does look interesting. I'm definitely interested.

leimig:
It does look interesting. I'll try to modify the DS.attr to support that this weekend. I'll update this post once I have news about it.

rmachielse:
Thanks!

leimig:
Hi @rmachielse, sorry about the delay. These last couple weeks were crazy :P About this thread, I just submitted the Pull Request https://github.com/emberjs/data/pull/3789. Hope they will accept :)

bmac:
The Ember Data team does not think it is the right approach for solving the problem of transforming data before assigning it to an attribute because it requires user code to be aware of all the type that may be set on an attribute and requires the user code to handle the correct transformation without knowing where the value is originating from.

A better solution may be to: 
- Transform the data closer to the input source where there is more context. One example of this approach for data bound values is @rwjblue's [one-way-input twiddle](http://ember-twiddle.com/2d7246875098d0dbb4a4) which sanitizes the user input in the component. 

- Or use a computed property to transform the attribute's value and data bind to the computed property instead of the original attribute.

PR #3789, opened 2015-09-20
Add support to custom getters and setters at DS.attr

leimig:
Add support to custom getters and setters to Model attributes as suggested by @rmachielse in issue https://github.com/emberjs/data/issues/3708

The user now should be able to customize them following the example
```javascript
  age: DS.attr('number', {
    // value: the stored value of the property age
    get: function(value) {
      var ret;
      // do whatever you want and
      // return the value the user will get
      return ret;
    },
    // value: the new value for property age
    set: function(value) {
      var ret;
      // do whatever you want and
      // return the value that will be stored
      return ret;
    }
  })
});
```

### DIFF
--- a/packages/ember-data/lib/system/model/attributes.js
+++ b/packages/ember-data/lib/system/model/attributes.js
@@ -314,30 +314,56 @@ export default function attr(type, options) {
   return Ember.computed({
     get: function(key) {
       var internalModel = this._internalModel;
+      var ret;
+
       if (hasValue(internalModel, key)) {
-        return getValue(internalModel, key);
+        ret = getValue(internalModel, key);
       } else {
-        return getDefaultValue(this, options, key);
+        ret = getDefaultValue(this, options, key);
       }
+
+      if (options.get) {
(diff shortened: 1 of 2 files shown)


bmac:
Hi @leimig. Thanks for the work on this pr. The Ember Data team talked about this pr today in our meeting today.

We do not think it is the right approach for solving the problem of transforming data before assigning it to an attribute because it requires user code to be aware of all the type that may be set on an attribute and requires the user code to handle the correct transformation without knowing where the value is originating from.

A better solution may be to: 
- Transform the data closer to the input source where there is more context. One example of this approach for data bound values is @rwjblue's [one-way-input twiddle](http://ember-twiddle.com/2d7246875098d0dbb4a4) which sanitizes the user input in the component. 

- Or use a computed property to transform the attribute's value and data bind to the computed property instead of the original attribute.

If you would like to continue pursuing this path a good idea may be to create an RFC enumerating the problem and how this would address them. Alternately, it may be interesting to see this functionality implemented as an add-on.

leimig:
I see, thanks for the feedback. I'll close the PR :)

PR #3792, opened 2015-09-22
Check if primary model has type attribute using _normalizePolymorphicRecord

splattne:
Add the primaryHasTypeAttribute to the _normalizePolymorphicRecord call in rest-serializer's _normalizeResponse. Fixes bug if the primary model has a type property.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "ember-data",
-  "version": "1.13.12",
+  "version": "1.13.13",
   "namespace": "DS",
   "repository": "git://github.com/emberjs/data.git",
   "license": "MIT",


wecc:
I think we could DRY this up a bit, it feels redundant that we're passing `primaryModelClass`, `primarySerializer` and `primaryHasTypeAttribute` to `_normalizePolymorphicRecord` every time. `primaryModelClass` should be sufficient? It might be out of scope for this PR but feel free to do it if you want.

Can we also add a test?

bmac:
Hi @splattne. 

I believe this issue is fixed by https://github.com/emberjs/data/commit/5573760fb9675143d9d04b099c6b909d1e18b4ad which should be `1.13.14`.

bmac:
Closing this issue since I believe it has fixed.

PR #3794, opened 2015-09-22
[DOC release] Update documentation for store

### LINKED ISSUE #3687: store.push API documentation outdated
Still references passing a type, and nothing about the JSON API.

Looking at the JSON API the examples show plural types, but for ED it seems like singular types are needed.

Kuzirashi:
This attempts to resolve #3687 and update other parts of store docs to mention JSONAPI move of ED. It's currently a massive WIP, but I'm open to all feedback.

### DIFF
--- a/packages/ember-data/lib/system/store.js
+++ b/packages/ember-data/lib/system/store.js
@@ -1522,15 +1522,64 @@ Store = Service.extend({
   /**
     Push some data for a given type into the store.
 
-    This method expects normalized data:
+    This method expects normalized [JSON API](http://jsonapi.org/) document. This means you have to follow [JSON API specification](http://jsonapi.org/format/) with few minor adjustments:
+    - record's `type` should always be in singular, dasherized form
+    - members (properties) should be camelCased
 
-    * The ID is a key named `id` (an ID is mandatory)
-    * The names of attributes are the ones you used in
-      your model's `DS.attr`s.
-    * Your relationships must be:
-      * represented as IDs or Arrays of IDs
-      * represented as model instances
-      * represented as URLs, under the `links` key
+    [Your primary data should be wrapped inside `data` property](http://jsonapi.org/format/#document-top-level):
+
+    ```js
+    store.push({
+      data: {
+        // primary data for single record of type `Person`
+        id: '1',
+        type: 'person',
+        attributes: {
+          firstName: 'Daniel',
+          lastName: 'Kmak'
+        }
+      }
+    });
+    ```
+
+    [Demo.](http://ember-twiddle.com/fb99f18cd3b4d3e2a4c7)
+
+    `data` property can also hold an array (of records):
+
+    ```js
+    store.push({
+      data: [
+        // an array of records
+        {
+          id: '1',
+          type: 'person',
+          attributes: {
+            firstName: 'Daniel',
+            lastName: 'Kmak'
+          }
+        },
+        {
+          id: '2',
+          type: 'person',
+          attributes: {
+            firstName: 'Tom',
+            lastName: 'Dale'
+          }
+        }
+      ]
+    });
+    ```
+
+    [Demo.](http://ember-twiddle.com/69cdbeaa3702159dc355)
+
+    There are some typical properties for `JSONAPI` payload:
+    * `id` - mandatory, unique record's key
+    * `type` - mandatory string which matches `model`'s dasherized name in singular form
+    * `attributes` - object which holds data for record attributes - `DS.attr`'s declared in model
+    * `relationships` - object which must contain any of the following properties under each relationships' respective key (example path is `relationships.achievements.data`):
+      - [`links`](http://jsonapi.org/format/#document-links)
+      - [`data`](http://jsonapi.org/format/#document-resource-object-linkage) - place for primary data
+      - [`meta`](http://jsonapi.org/format/#document-meta) - object which contains meta-information about relationship
 
     For this model:
 
@@ -1538,8 +1587,8 @@ Store = Service.extend({
     import DS from 'ember-data';
 
     export default DS.Model.extend({
-      firstName: DS.attr(),
-      lastName: DS.attr(),
+      firstName: DS.attr('string'),
+      lastName: DS.attr('string'),
 
       children: DS.hasMany('person')
     });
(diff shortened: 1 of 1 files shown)


packages/ember-data/lib/system/store.js
```diff
@@ -1522,15 +1522,53 @@ Store = Service.extend({
...
+        }
+      ]
+    })
+    ```
+
+    There are some typical properties for `JSONAPI` payload:
+    * `id` - mandatory, unique record's key
+    * `type` - mandatory string which matches `model` name (can be singular or plural form)
+    * `attributes` - object which holds data for record attributes - `DS.attr`'s declared in model
+    * `relationships` - object which must contain any of the following:
```
wecc: Could you mention that `links`, `data` and `meta` are supposed to be under each relationships' respective key, to remove confusion.

packages/ember-data/lib/system/store.js
```diff
@@ -1522,15 +1522,53 @@ Store = Service.extend({
...
+      - [`links`](http://jsonapi.org/format/#document-links)
+      - [`data`](http://jsonapi.org/format/#document-resource-object-linkage) - place for primary data for relationships, property `data` isn't direct child of `relationships`, but rather child of object named after related model, eg. path `relationships.achievements.data`
```
wecc: This is not only valid for `data`, see comment above.

packages/ember-data/lib/system/store.js
```diff
@@ -1561,8 +1619,10 @@ Store = Service.extend({
     ```js
     {
       id: 1,
-      firstName: "Tom",
-      lastName: "Dale",
+      attributes: {
+        firstName: "Tom",
+        lastName: "Dale"
+      }
       links: {
```
wecc: This should be `relationships.children.links.related`

packages/ember-data/lib/system/store.js
```diff
@@ -1522,15 +1522,53 @@ Store = Service.extend({
   /**
     Push some data for a given type into the store.
 
-    This method expects normalized data:
+    This method expects [JSON API](http://jsonapi.org/) compound document with [your primary data wrapped inside `data` property](http://jsonapi.org/format/#document-top-level):
```
wecc: Can we add that this method expects a "normalized JSON API Document" (as in, camelCased members and singular dasherized types).

packages/ember-data/lib/system/store.js
```diff
@@ -1522,15 +1522,53 @@ Store = Service.extend({
...
+            firstName: 'Tom'
+          }
+        }
+      ]
+    })
+    ```
+
+    There are some typical properties for `JSONAPI` payload:
+    * `id` - mandatory, unique record's key
+    * `type` - mandatory string which matches `model` name (can be singular or plural form)
```
wecc: _should_ be singular dasherized even though plural currently works.

wecc:
Nice PR @Kuzirashi, thanks! Left a couple of comments for minor tweaks.

Kuzirashi:
What do you think now? @wecc

packages/ember-data/lib/serializers/json-api-serializer.js
```diff
@@ -58,12 +58,12 @@ var dasherize = Ember.String.dasherize;
               "data": [
                 {
                   "id": "3",
-                  "type": "players"
+                  "type": "player"
```
bmac: I believe the `JSONAPISerializer` expects types to be plural.

packages/ember-data/lib/system/store.js
```diff
@@ -1522,15 +1522,64 @@ Store = Service.extend({
...
-    * The ID is a key named `id` (an ID is mandatory)
-    * The names of attributes are the ones you used in
-      your model's `DS.attr`s.
-    * Your relationships must be:
-      * represented as IDs or Arrays of IDs
-      * represented as model instances
-      * represented as URLs, under the `links` key
+    [Your primary data should be wrapped inside `data` property](http://jsonapi.org/format/#document-top-level):
+
+    ```
```
bmac: Lets add `js` after the back ticks so the markdown hi-lighter knows to use the JavaScript formatter instead of the generic one.

packages/ember-data/lib/system/store.js
```diff
@@ -1522,15 +1522,64 @@ Store = Service.extend({
...
+        }
+      }
+    });
+    ```
+
+    [Demo.](http://ember-twiddle.com/fb99f18cd3b4d3e2a4c7)
+
+    `data` property can also hold an array (of records):
+
+    ```
```
bmac: js

packages/ember-data/lib/serializers/json-api-serializer.js
Kuzirashi: Store.push expects singular, JSON API serializer expects plural? I thought it is unified. Should I revert changes to json-api-serializer?

wecc: @bmac is correct. The internal Ember Data JSON API format differs a bit from the one used by `JSONAPISerializer`.

bmac:
This looks great @Kuzirashi. Excited to see this merged.

packages/ember-data/lib/serializers/json-api-serializer.js
Kuzirashi: This is confusing, at least for me. Shouldn't this be mention somewhere, somehow?

Kuzirashi:
@bmac, @wecc are we able to embed ember-twiddles via iframe in docs?

packages/ember-data/lib/serializers/json-api-serializer.js
wecc: ED uses camelCased attributes and singular types internally, regardless of what adapter/serializer you're using.

When you're using the JSON API adapter/serializer we want users to be able to use the examples available on jsonapi.org and have it just work. Most users never have to care about the internal format since the serializer handles the work for them.

This is documented in the guides, http://guides.emberjs.com/v2.0.0/models/pushing-records-into-the-store/

Kuzirashi: Thanks for explanation.

Kuzirashi:
Changes applied.

bmac:
Its possible to embed an ember-twiddles via iframe in the docs by just inlining the html. However, they are discouraged because they are slow to load, are not very PR friendly and in general lead to some extra overhead in managing them which makes it harder to notice when iframe content is out of date.

packages/ember-data/lib/serializers/json-api-serializer.js
bmac: I think we should document this difference in some other places. Not everyone reads the guides.

bmac:
I think this pr looks good. @Kuzirashi do you mind squashing your commits into 1 and prepending the commit message with `[DOC beta]`?

Kuzirashi:
Done. @bmac

wecc:
Great work @Kuzirashi!

bmac:
Thanks @Kuzirashi.

PR #3795, opened 2015-09-23
Trigger an assertion when calling `findRecord` with falsy (except 0) id

### LINKED ISSUE #2348: store.find('model', ''); shouldn't query 'api/models/'
basically if you do find with an empty string then it does a search for api/models/ with no id which leads to it downloading all of the models, [coerceID](https://github.com/emberjs/data/blob/2997e47034564aa803c1cacb0cf182ca36922fdc/packages/ember-data/lib/system/store.js#L57-L59) seems like it should be coercing empty strings to null.

Kuzirashi:
Attempt to close #2348. Do you think it is sufficient or should `Ember.Error` be thrown as @stefanpenner suggested [here](https://github.com/emberjs/data/pull/518#issuecomment-17819789)?

### DIFF
--- a/packages/ember-data/tests/integration/store-test.js
+++ b/packages/ember-data/tests/integration/store-test.js
@@ -211,8 +211,7 @@ function ajaxResponse(value) {
 
 
 
-module("integration/store - findRecord", {
-});
+module("integration/store - findRecord");
 
 test("store#findRecord fetches record from server when cached record is not present", function() {
   expect(2);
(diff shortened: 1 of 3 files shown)


packages/ember-data/tests/integration/store-test.js
```diff
@@ -1,3 +1,5 @@
+import { badIdFormatAssertion } from 'ember-data/system/store';
```
Kuzirashi: This causes an error:

~~~
 706 tests complete (9256 ms)
  706) integration/snapshot - DS.Snapshot: global failure
     Error: Could not find module ember-data/system/store
~~~

Is there any other way I can import exported module variables? I wonder why I can't use normal import like in HTMLBars. I also haven't found any example of using `import` in unit tests. I doubt they are real unit tests if they have access to all global namespaces, no imports. Is this because of old Ember CLI version?

PS: Tests on my local machine pass, appveyor build runs fine, only travis errors.

@wecc @bmac

bmac: I think this is a limitation of our testing infrastructure. Modules can be access in dev mode but the production build (which I suspect travis is using) wraps Ember Data's module loader in a closure so tests (or user code) can't get at internal paths like 'ember-data/system/store'.

Kuzirashi: Ok, I'll try to work around this. Do you plan to fix this in future?

Kuzirashi:
Tests pass.

bmac:
Looks good @Kuzirashi. Do you mind squashing this pr into 1 commit?

Kuzirashi:
Commits squashed. @bmac

Kuzirashi:
Ping @bmac :S hope you notice and see if there's anything for this pr to get merged or you just didn't notice it. :P

bmac:
Thanks @Kuzirashi.

PR #3796, opened 2015-09-24
[DOC release] Update documentation for EmbeddedRecordsMixin.

### LINKED ISSUE #2794: "EmbeddedRecordsMixin belongsTo serialize" used "records" instead of "record"
When a comment belongsTo a user, we should be able to embed user inside comment and configure CommentSerializer like the following:
```
App.CommentSerializer = DS.RESTSerializer.extend( DS.EmbeddedRecordsMixin, {
  attrs: {
    user: {
      serialize: 'record',   // this is NOT working. However 'records' works.
      deserialize: 'record'
    }
  }
});
```
However 'record' does not work and 'records' works which is confusing.

I will probably create a PR later.

Kuzirashi:
Attempts to close #2794 by removing falsy mention of `record` method for deserializing (was it even ever supported?). Some minor changes also, like missed dot and adding backticks when mentioning classes.

### DIFF
--- a/packages/ember-data/lib/serializers/embedded-records-mixin.js
+++ b/packages/ember-data/lib/serializers/embedded-records-mixin.js
@@ -7,10 +7,10 @@ var camelize = Ember.String.camelize;
 
   `DS.EmbeddedRecordsMixin` supports serializing embedded records.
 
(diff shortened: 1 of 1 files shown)


wecc:
Thanks @Kuzirashi!

PR #3811, opened 2015-10-02
Format lists in JSONSerializer docs for rendering

jyunderwood:
Currently, the [JSONSerializer override process docs](http://emberjs.com/api/data/classes/DS.JSONSerializer.html) are hard to follow because of the markdown formatting. This patch fixes the formatting for rendering.


bmac:
Thanks @jyunderwood.

PR #3812, opened 2015-10-02
Don't preload the belongsTo to be available synchronously

elucid:
Having this preload is inconsistent with the way that the `hasMany` relationship proxy works. In particular, it has different semantics when rendering a property on proxied relationship.

For example, let's say there is a template containing `{{post.author.someComputedProperty}}` where author `belongsTo` post. If `someComputedProperty` relies on a dependent key being present in order to compute correctly, `{{post.author.someComputedProperty}}` will throw.

Without the preload, the lookup chain stops at the null `content` property of the `PromiseObject` and `someComputedProperty` is only evaluated once the promise resolves and a fully materialized author is set as the `content`. This is how things used to work up until 1.0.0.beta.10

The preload was added in 12301b9405c4ffde31d9dd027c8b266c1231d0f8 and released in 1.0.0.beta.11. I don't fully understand why it was needed at that time, but without it the tests did not pass.

This is no longer the case. After removing the preload the suite is green and the above rendering issue is resolved.

/cc @ghedamat who tracked this down with me and @mmun who pointed us in the right direction.

### DIFF
--- a/packages/ember-data/lib/system/relationships/state/belongs-to.js
+++ b/packages/ember-data/lib/system/relationships/state/belongs-to.js
@@ -130,8 +130,7 @@ BelongsToRelationship.prototype.getRecord = function() {
     }
 
     return PromiseObject.create({
-      promise: promise,
-      content: this.inverseRecord ? this.inverseRecord.getRecord() : null
+      promise: promise
     });
   } else {
     if (this.inverseRecord === null) {


elucid:
Hmm...not sure what is going on here. `npm test` is green locally. I will try nom bom and rerunning :/

elucid:
I get the same failures locally now. How unfortunate.

Is this worth pursuing and trying to fix the tests?

bmac:
Hi @elucid. I think it would be valuable to continue pursuing this pd and fix the tests.

bmac:
Hi @elucid. Do you have time to put together a failing test of someComputedProperty throwing when the reloaded `content` is null? I think that would be helpful in solving this issue. Thanks.

bmac:
Hi @elucid. Thanks for this pr. I spent some time playing with it but I couldn't figure out a way to remove preloaded `content` property without breaking some existing functionality.  The long term plan for Ember Data is to migrate away from ObjectProxies which I think should help avoid this issue altogether. I'm going to close this pr but feel free to open a new one if you can figure out a way to solve this problem.

Issue #3784, opened 2015-09-18
BelongsTo relation called 'type' in combination with a link returns error


jevanlingen:
Today I was updating Ember Data from 1.13.8 to 1.13.13. As #3725, the `type` attribute can be used again. My case is a little different, as I have a belongs-to relation. This still shouldn't be a problem as @tschoartschi in ticket #3769 also has a `type` belongs-to relation and the fix worked for him.

I got following file model:
```javascript
export default DS.Model.extend({
    name: attr(),
    size: attr(),
    lastModified: attr(),
    created: attr(),
    
    directory: DS.belongsTo('directory', { async: true }),
    type: DS.belongsTo('file-type', { async: true })
}
```

The type property belongs to a `file-type` model, which is loaded async with an own given link. So I made a file serializer:
```javascript
export default DS.RESTSerializer.extend({
    isNewSerializerAPI: true,

    normalize(typeClass, hash, prop) {
        hash.links = {
            type: `link-to-another-system?type_id=${hash.type}`
        };
             
        return this._super(...arguments);
    }
});
```

--

Yet, I still seem to face a problem regarding the `type` property. Pull #3725 just fixes the problem for attributes (by adding the `primaryHasTypeAttribute` variable).

Maybe my own serializer is the problem, but 
```javascript
var primaryHasTypeAttribute = get(modelClass, 'attributes').get('type');
```
returns `undefined` for me. I find that logicly, as a belong-to relation is not a attribute (to be honest I don't understand how it's possible pull #3725 fixes #3769).

Whaterever the case. if a set `primaryHasTypeAttribute` to `true` by hand, the problem is solved! Maybe the check should also take relation properties into account?

jevanlingen:
Nobody?

pangratz:
@jevanlingen can you confirmed that #3814 solves your issue, so this one can be closed?

jevanlingen:
@pangratz Yeah, I can confirm #3814 does fix this issue. Added the `get(modelClass, 'relationshipsByName').get('type');`line to my local code, problem was fixed!

pangratz:
won der ful! Good to hear! :sparkles:

jevanlingen:
I still face one problem with the `type` atrribute. When I tested this back in october, I just checked if no errors occured.

But today I noticed the related link does not work. The serializer should return something like this for an async relation according to the JSON API spec:

```javascript
data: {
  id: "something",
  ... // other props
  relationships: {
    type: {
      links: {
[...]

PR #3814, opened 2015-10-02
[BUGFIX] Do not deserialize when a relationship named `type` exists

andycrum:
69dca0c fixed this for attributes named `type`, but the RESTSerializer
was still trying to deserialize relationships with that name as
polymorphic, which was causing errors for anyone having a domain
model called "Type".

cc @bttf @bmac


andycrum:
I believe this fixes #3784

bmac:
Looks great. Thanks @andycrum.

jevanlingen:
:+1:

jevanlingen:
Btw, can this fix be applied to the 1.13.x branche too?

PR #3815, opened 2015-10-04
convert undefined json values into null in rest adapter. fixes #3807

### LINKED ISSUE #3807: JSONAPI Adapter not sending undefined attributes.
Hey JSONAPI adapter is removing attributes from the payload when they're undefined. In the application I'm working on right now I need to unset an attribute in our database. 

````javascript
  this.get('group').set('disabledAt', undefined);
  this.get('group').save();
````

In the request payload the disabledAt is not sent to the server. The stange thing is that this.get('group') serialize includes the disabledAt attribute.

Is there any way to prevent the removal of the disabledAt attribute from the payload? Thanks

I'm on Ember 1.13.6 and Ember-Data 1.13.13

### PR DESCRIPTION
**Pull request by webPapaya**
When using Ember Data with the rest adapter it's not possible to unset a model attribute on the server because of: 
```javascript
JSON.stringify({key: undefined}) // => "{}"
JSON.stringify({key: null}) // => "{"key":null}"
```
This pull requests converts all undefined keys to null so you're able to unset values on the server.

### DIFF
--- a/packages/ember-data/lib/adapters/rest-adapter.js
+++ b/packages/ember-data/lib/adapters/rest-adapter.js
@@ -875,7 +875,10 @@ var RESTAdapter =  Adapter.extend(BuildURLMixin, {
 
     if (hash.data && type !== 'GET') {
       hash.contentType = 'application/json; charset=utf-8';
-      hash.data = JSON.stringify(hash.data);
+      hash.data = JSON.stringify(hash.data, (key, val) => {
+        if (val === undefined) { return null; }
+        return val;
+      });
     }
 
     var headers = get(this, 'headers');
--- a/packages/ember-data/tests/unit/adapters/rest-adapter/ajax-test.js
+++ b/packages/ember-data/tests/unit/adapters/rest-adapter/ajax-test.js
@@ -102,3 +102,18 @@ test("ajaxOptions() empty data", function() {
     url: 'example.com'
   });
 });
+
+test("ajaxOptions() converts undefined data values to null", function() {
+  var url = 'example.com';
+  var type = 'POST';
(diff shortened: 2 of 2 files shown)


matthooks:
Hmm -- can't you just manually set the value to null to unset? It seems like in some instances you would want to not send the attribute, and setting it to undefined would allow that...

webPapaya:
So the issue is that the serializer automatically converts null values to undefined. So I can't easily unset a value without making a manual request or overwriting the serializer. 

```javascript
this.get('group').set('disabledAt', null);
this.get('group').serialize(); // {..., attributes: {disabledAt: undefined}}
```

As I wrote in issue #3807 for now I could solve this problem by overwriting the serialize method in our application serializer. I think ember-data is using undefined internally for exclusively for consistency and I don't want to break this consistency.

I mean if you really don't want to send a certain attribute to the server you should probably create a custom serializer which filters this particular attribute from serialization or do this request manually. I think it's much more confusing if you set a model attribute to undefined or null, save it to the server and the attribute gets set to the server value.

wecc:
`null` values being serialized to `undefined` sounds like a bug to me.

Forcing `undefined` to `null` in the adapter (should be a serializer concern) is most likely not the correct solution.

Let's explore this further in #3807.

bmac:
@webPapaya I made a jsbin to demonstrate the existing attribute serialization behavior. http://emberjs.jsbin.com/cirifa/edit?html,css,js,output

Currently the only case where an `undefined` attribute does not get serialized is when there is no transform defined for the attribute. I think this pr would be a breaking change to that behavior. Instead I believe defining a transform for an attribute that converts `undefined` to `null` is the correct solution for this problem. 

I'm going to close this pr. Feel free to re-open if I missed something.

PR #3819, opened 2015-10-05
[CLEANUP] Remove duplicate `default export` in snapshot.js

pangratz:
`Snapshot` is default export-ed already at the end of the file.

### DIFF
--- a/packages/ember-data/lib/system/snapshot.js
+++ b/packages/ember-data/lib/system/snapshot.js
@@ -12,7 +12,7 @@ var get = Ember.get;
   @constructor
   @param {DS.Model} internalModel The model to create a snapshot from
 */
-export default function Snapshot(internalModel) {
+function Snapshot(internalModel) {
   this._attributes = new EmptyObject();
   this._belongsToRelationships = new EmptyObject();
   this._belongsToIds = new EmptyObject();


packages/ember-data/lib/system/snapshot.js
```diff
@@ -12,7 +12,7 @@ var get = Ember.get;
   @constructor
   @param {DS.Model} internalModel The model to create a snapshot from
 */
-export default function Snapshot(internalModel) {
```
rwjblue: I'd personally prefer to remove the one at the end, but not a big deal either way....

pangratz: Hmm, [Model](https://github.com/emberjs/data/blob/4745dcb6fa00664ddb19bab9f4913695656f7e1c/packages/ember-data/lib/system/model/model.js#L848) and [Store](https://github.com/emberjs/data/blob/4745dcb6fa00664ddb19bab9f4913695656f7e1c/packages/ember-data/lib/system/store.js#L2155) are `export default`ing at the end. But I am happy to change if this is desired ...

rwjblue: Nah, if that is the style in the codebase already it is better to be consistent (or update them all).

bmac:
Thanks @pangratz

PR #3821, opened 2015-10-06
[CLEANUP] Remove unecessary wrap of rejection in run-loop

pangratz:
Rejection is handled correctly without having to wrap it inside an call
on the run loop.


bmac:
:rocket:

PR #3825, opened 2015-10-07
[BUGFIX] Restore IE8 compatibility by using Ember.create

### LINKED ISSUE #3791: Ember Data should use Ember.create in the 1.13 branch
Pr https://github.com/emberjs/data/pull/3689 ported the EmptyObject work into the 1.13 branch used Object.create. Unfortunately, Object.create is not supported by IE8.  We should replace this with Ember.create.

https://github.com/emberjs/data/blob/release-1-13/packages/ember-data/lib/system/empty-object.js#L6

pangratz:
The back ported performance optimization of using EmptyObject uses
`Object.create`, which is not available in IE8. Hence `Ember.create` is
used, which shims this behavior for IE8 and uses `Object.create` when it
is available.

This closes #3791 

cc @bmac @stefanpenner


bmac:
:hammer: :bug:

PR #3306, opened 2015-06-11
Rename adapter and serializer methods to match the new store methods

bmac:
`Adapter#find` -> `Adapter#findRecord` and `Adapter#findQuery` -> `Adapter#query`
`BuildURLMixin#urlForFind` -> `BuildURLMixin#urlForFindRecord` and `BuildURLMixin#urlForFindQuery` -> `BuildURLMixin#urlForQuery`
`Serializer#extractFind` -> `Serializer#extractFindRecord` and `Serializer#findQuery` -> `Serializer#extractQuery`

PR #3829, opened 2015-10-07
Remove references to findQuery

HeroicEric:
`findQuery` was deprecated back in 1.0 and removed for 2.0 (https://github.com/emberjs/data/pull/3235)

A test file was also removed that was referencing `findQuery` as well
as `RestAdapter#find`, which was also deprecated and removed.(https://github.com/emberjs/data/pull/3306)

In addition to referencing code that was no longer present, this test
file was missing the `-test` in its name and has not been running as
part of the build. It did not pass before or after these changes.


bmac:
Thanks

PR #3830, opened 2015-10-08
Fix tests for non-dasherized lookups

HeroicEric:
The test file for non-dasherized lookups was missing the `-test` in its
filename, which means it hasn't been running.

This commit fixes the filename and gets the tests passing again.


bmac:
Thanks @HeroicEric

PR #3835, opened 2015-10-09
[BUGFIX] extract polymorphic belongsTo in RESTSerializer

### LINKED ISSUE #3806: Polymorphic belongs-to relation model is not serialized in Ember-data v1.13
Hi team, we moved our app to ember-data v1.13.8 (We are using rest adapter) everything is working as expected but polymorphic belongs-to relation model is not serialized. I followed this https://github.com/emberjs/data/blob/master/TRANSITION.md#polymorphic-relationships. But when i changed data from  
{
  "message": 12,
  "messageType": "post"
}
to 
{
  "message": {id: 12, type: 'post'}
}
it will work. This may be an issue or is there any work around for this?

pangratz:
This change adds the correct extraction of a polymorphic belongsTo
specified in the payload in the following form:

``` js
{
  id: 123,
  // ...
  message: 1,
  messageType: 'post'
}
```

where the model is specified as:

``` js
DS.Model.extend({
  message: DS.belongsTo('message', { polymorphic: true })
})
```

---

If this gets accepted, this should likely be ported back to `1.13` as well.

This closes #3806

### DIFF
--- a/packages/ember-data/lib/serializers/rest-serializer.js
+++ b/packages/ember-data/lib/serializers/rest-serializer.js
@@ -56,6 +56,37 @@ var camelize = Ember.String.camelize;
 var RESTSerializer = JSONSerializer.extend({
 
   /**
+   `keyForPolymorphicType` can be used to define a custom key when
+   serializing and deserializing a polymorphic type. By default, the
+   returned key is `${key}Type`.
+
+   Example
+
+    ```app/serializers/post.js
+    import DS from 'ember-data';
+
+    export default DS.RESTSerializer.extend({
+      keyForPolymorphicType: function(key, relationship) {
+        var relationshipKey = this.keyForRelationship(key);
+
+        return 'type-' + relationshipKey;
+      }
+    });
+    ```
+
+   @method keyForPolymorphicType
+   @param {String} key
+   @param {String} typeClass
+   @param {String} method
+   @return {String} normalized key
+  */
+  keyForPolymorphicType: function(key, typeClass, method) {
+    var relationshipKey = this.keyForRelationship(key);
+
+    return `${relationshipKey}Type`;
+  },
+
+  /**
     Normalizes a part of the JSON payload returned by
     the server. You should override this method, munge the hash
     and call super if you have generic normalization to do.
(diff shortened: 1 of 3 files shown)


packages/ember-data/lib/serializers/rest-serializer.js
```diff
@@ -689,6 +689,49 @@ var RESTSerializer = JSONSerializer.extend({
...
+    //   {
+    //     message: 1,
+    //     messageType: 'post'
+    //   }
+    //
+    // The next code checks if the latter case is present and returns the
+    // corresponding JSON-API representation. The former case is handled within
+    // the super class (JSONSerializer).
+    var isPolymorphic = relationshipMeta.options.polymorphic;
+    var typeProperty = `${relationshipKey}Type`;
```
wecc: I think this should probably go through `keyForAttribute`. I think we had a bug related to this previously. It would also make the ActiveModelSerializer work out of the box... I think.

pangratz: `relationshipKey` is the result of `this.keyForRelationship(key, relationshipMeta.kind, 'deserialize');` from [this line](http://git.io/vCsx9).

In `serializePolymorphicType` above in line 686 `keyForAttribute` is used, so using `keyForAttribute` here would make it consistent I guess. But judging by the name, wouldn't be `keyForRelationship` the correct one, for both `serializePolymorphicType` and the new `serializePolymorphicType`?

It looks like AMS [uses `keyForRelationship](http://git.io/vCGTi) ...

wecc: Agree, `keyForRelationship` makes more sense.

wecc: Just to make sure, my intention here was to do ~~`keyForAttribute`~~ `keyForRelationship` on the `typeProperty` string.

Even though `relationshipKey` is cased correctly that doesn't mean we can just append `Type` to the end of it.

Say you have a `bestFriend: belongsTo()` and you've overriden `keyForRelationship` to returned dasherized. `bestFriend` is now `best-friend`. Odds are your payload contains `best-friend-type`, not `best-friendType`.

pangratz: That's a good point @wecc.

I will add a `keyForPolymorphicType` hook, which will also be used [here](https://github.com/emberjs/data/blob/921e8c9960fab34a7f6de4cdc618893b5394f7d4/packages/ember-data/lib/serializers/rest-serializer.js#L688-L690)...

bmac:
@pangratz could you update the commit message to start with `[BUGFIX beta]`?

packages/ember-data/lib/serializers/rest-serializer.js
```diff
@@ -683,12 +714,72 @@ var RESTSerializer = JSONSerializer.extend({
...
+    // check if the old way differs with the new, correct one, and if so, log a
+    // deprecation warning and restore the old behavior
+    if (key !== typeKey) {
+      Ember.deprecate("The typeKey to serialize a polymorphic type uses keyForAttribute which has been deprecated in favor of keyForPolymorphicType.", {
+        id: 'ds.rest-serializer.deprecated-key-for-polymorphic-type',
+        until: '3.0.0'
+      });
+
+      typeKey = key;
+    }
```
pangratz: The current way to get the `myKeyType` string for the `myKey` relationship is using `keyForAttribute` which is not correct IMHO. This is changed in the course of this PR by adding a `keyForPolymorphicType` hook.

This could break current behavior, that's why a deprecation message is added. I will add a test for the deprecation if the deprecation is found to be necessary.

pangratz: Actually, I think this would be save without the deprecation, since there wouldn't have been another way to change the way the polymorphic record is serialized except for overriding `serializePolymorphicType` as a whole...

bmac: @pangratz it looks like a user can't use `keyForPolymorphicType` if it differs from `${key}Type`; since `key` always wins. Is that correct? Is there anyway we could rewrite this logic so uses who define a custom `keyForPolymorphicType` get the new happy path without a deprecation warning?

pangratz: Uff, you're absolutely right. Nice catch! Lol, logic, how do you even work. I will change the code so the deprecation warning is only logged when `keyForPolymorphicType` is not overwritten...

pangratz:
@bmac I've updated the logic and added a check which ignores the fallback when the `keyForPolymorphicType` has been overwritten.

bmac:
Thanks @pangratz.

Issue #3726, opened 2015-09-03
Embedded records in old rest API have type attribute values lowercased


iammer:
I am using Ember-data 2.0.0.

I have a model sent from the server using the old REST API with embedded records. It looks like this using the old rest API.  
```js
{
  parent: {
    children: [{
      type: 'SOMEVALUE'
    }]
  }
}
```
Type is an attribute of the child model.

When the child record is added to the store type is changed from 'SOMEVALUE' to 'somevalue'.

I believe this is caused by the extractRelationship method in JSONSerializer.  The relationshipHash.type is sent through Ember.string.dasherize.  Which works for the JSONAPI and for polymorphic types, but isn't the correct thing to do for embedded records in the old REST API.

pangratz:
Hey @iammer, thanks for reporting! This is reproduced in http://ember-twiddle.com/7f0cde397742775ab97c and a fix is proposed in #3837.

pangratz:
I think this issue can be closed since #3837 has been merged.

PR #3837, opened 2015-10-10
[BUGFIX] Attribute/relationship named "type" of embedded record is considered before normalization

pangratz:
In #3726 a bug is reported, where an attribute or relationship named `type` of an embedded record is mistakenly normalized.

If this is accepted, it should likely be ported back to `1.13` as well I guess.


bmac:
Looks good @pangratz. Do you mind squashing these commits and prefixing with `[BUGFIX beta]`? so we can remember to pull this into the 2.2 branch.

Issue #3605, opened 2015-07-29
Add a method for normalizing a payload instead of a single resource


jonkoops:
I am building a Service that takes care of authenticating a user in my application. The Service does a request to the API, normalizes the data, pushes and gets the record from the store and resolves with said record of an authenticated user.

The issue I am encountering is that there is no way to normalize the payload without manually stripping the `data` hash from the root of the server response. This is because `store.normalize()` accepts only a single resource and not a payload like JSON API where the type and primary data is known. The Service now needs to know what the format of the payload is because the API response needs to be modified, which means that if the API response changes this code will break.

After some discussion with @wecc and @igorT a suggestion popped up that it might be an idea to add a `normalizePayload` method which should take an optional `modelName` parameter  to determine which serializer will be used and a `inputPayload` argument to pass the payload (much the same as [store.pushPayload](http://emberjs.com/api/data/classes/DS.Store.html#method_pushPayload)).

*Current situation:*
```js
var payload = store.normalize('user', response.data);
var record = store.push(payload);
```

*Proposed solution:*
```js
var payload = store.normalizePayload(response); // Or optionally store.normalizePayload('user', response);
var record = store.push(payload);
```

fotinakis:
I think I just hit the same thing you described—it looks like `store.push` only normalizes the JSON API primary data, not included resources, so relationships included with my user object die, for example: `"No model was found for 'subscriptions'` (because `subscriptions` in the JSON API response does not get normalized to `subscription` in ember-data internals).

I was sort of able to get this to work as a workaround, though with a bunch of deprecation warnings:
```javascript
var userId = payload.data.id;  // Grab now because pushPayload modifies the hash.
store.pushPayload(payload);
var userRecord = store.peekRecord('user', userId);
```

To fix this, shouldn't `store.push(payload)` normalize compound documents by default?

wecc:
> I think I just hit the same thing you described—it looks like store.push only normalizes the JSON API primary data

Just for the record, `store.push` expects fully normalized data, it does no normalization on it's own.

fotinakis:
Ah makes sense—let me reframe then, shouldn't `store.normalize` normalize full compound documents by default?

I'd expect this to work: `store.push(store.normalize('user', payload))` but that also fails with the above `No model was found for 'subscriptions'` error because the compound resources are not normalized.

wecc:
[...]

Issue #3576, opened 2015-07-23
Have this.store.pushPayload(type, payload) return records


bcardarella:
It would be a nice convenience if `pushPayload` returned the resulting records. Otherwise we have to go through the manual process of normalizing the payload and using `this.store.push`

bcardarella:
Here is a bit more about my use case:

We are using `torii` for authentication. In the torri adapter we make a `$.get` request to our backend's `sessions` endpoint. The response comes back in the json-api format. The torii adapter needs to return the account model as the currentUser to indicate the authenticated state. We get the raw payload as the response from the server so we need a way to deserialize/normalize that data, get it into the store, and retrieve the record for telling torii that this is the currentUser.

`this.store.push` will return the record as we need but we have to go through the overhead of something like:

```js
let modelClass = this.store.modelFor('account');
let serializer = this.store.serializerFor('account');
let normalizedPayload = serializer.normalizeSingleResponse(this.store, modelClass, payload, null, true);
let account = this.store.push(normalizedPayload);
```

`this.store.pushPayload` will encapsulate all of that overhead but does not return any record:

```js
let account = this.store.pushPayload('account', payload);
account === undefined
```

If it is not desirable to break the current API (as this request would) then could we get some other convenience function to handle this use case?

wecc:
This and #3576 seem very related.

workmanw:
Bump. It seems odd to me that `store.push` returns "{DS.Model|Array} the record(s) that was created or updated." while `store.pushPayload` doesn't.

Looking at the code, it seems easy enough to update the store and serializers. Would a PR be accepted to fix this?

karellm:
We ran into the same problem couple times in our app. It would be a really appreciated addition!

sly7-7:
Closing since the PR was merged

pangratz:
Cross-Linking to [RFC#161](https://github.com/emberjs/rfcs/pull/161), which addresses the problem raised in this issue.

bbansalWolfPack:
any updates on this? facing the same issue because pushPayload does not return anything back. So how can I use normalized response?

runspired:
@bbansalWolfPack we decided not to move forward with pushPayload return. We intend on deprecating `pushPayload`, and will eventually introduce an improved push API.

jagthedrummer:
[...]

PR #3838, opened 2015-10-11
[WIP] Add `normalizePayload(modelName, payload)` to DS.Store

pangratz:
A payload can be normalized by `store.normalizePayload()` into the JSON
API format Ember Data expects for `store.push`. The method optionally
takes a primary model name as first argument. If a primary model is
specified, the normalized payloads' primary data contains data for this
model from the payload and all other data is included.

---

This addresses #3605.

Opened this PR to get some feedback if I understood #3605 correctly and if this is heading the right direction...

### DIFF
--- a/packages/ember-data/lib/serializers/json-api-serializer.js
+++ b/packages/ember-data/lib/serializers/json-api-serializer.js
@@ -161,6 +161,10 @@ const JSONAPISerializer = JSONSerializer.extend({
     store.push(normalizedPayload);
   },
 
+  normalizePayload: function(primaryModelName, payload) {
+    return this._normalizeDocumentHelper(payload);
+  },
+
   /**
     @method _normalizeResponse
     @param {DS.Store} store
--- a/packages/ember-data/lib/serializers/rest-serializer.js
+++ b/packages/ember-data/lib/serializers/rest-serializer.js
@@ -344,6 +344,16 @@ var RESTSerializer = JSONSerializer.extend({
     @param {Object} payload
   */
   pushPayload: function(store, payload) {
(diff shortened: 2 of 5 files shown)


bmac:
I like this change. @pangratz do you mind wrapping this code in feature flags?

workmanw:
Just a small piece of feedback on the naming. As someone who maintains a custom adapter and serializer, this is kind of makes my head spin. There was a `normalizePayload` on the `JSONSerializer` previously, which was deprecated and renamed to `normalizeResponse` in Ember-Data 1.13. The old `normalizePayload` would take a server response and normalize it via the serialize. And now I'm still not totally sure what the difference between the current `normalizeResponse` and the new `normalizePayload` will be.

bmac:
Thanks @workmanw. That is excellent feedback. I've flagged this feature for discussion at the Ember Data weekly meeting to hopefully clarify its role, naming and necessity in the system.

bmac:
We chatted about this method in the Ember Data weekly meeting and were a little skeptical about the benefits of adding `normalizePayload` without a clear overall understanding of the problem space. I think the next steps here are to look enumerate the problems that `store#pushPayload`, `store#normalize` and `store#normalizePayload` solve and try to figure out if there are any gaps or places where APIs can be simplified.

Making `normalizePayload` return a value as @bcardarella suggested in https://github.com/emberjs/data/issues/3576 may also be an acceptable solution to this problem.

bmac:
@pangratz I'm going to close this because I think it would be better see `normalizePayload` return a value to solve this issue.

pangratz:
:+1:

PR #3841, opened 2015-10-12
Log a deprecation warning when when Ember Data is loaded with Ember 1.13

bmac:


packages/ember-data/lib/main.js
```diff
@@ -4,10 +4,17 @@
...
+}
+
+if (Ember.VERSION.match(/^1\.13\./)) {
+  Ember.deprecate(`Use of Ember Data 2+ with Ember 1.13 is deprecated. Please upgrade your version of Ember to 2.0 or higher.`, false, {
+    id: 'ds.version.ember-1-13',
+    until: '3.0.0'
+  });
 }
 
 if (Ember.VERSION.match(/^1\.12\.0/)) {
```
wecc: I think this could be removed since we're erroring out < 1.13 anyway?

bmac: good call

packages/ember-data/lib/main.js
```diff
@@ -4,14 +4,17 @@
...
-                        ". Please upgrade your version of Ember, then upgrade Ember Data");
+                        ". Please upgrade your version of Ember, then upgrade Ember Data.");
 }
 
-if (Ember.VERSION.match(/^1\.12\.0/)) {
-  throw new Ember.Error("Ember Data does not work with Ember 1.12.0. Please upgrade to Ember 1.12.1 or higher.");
+if (Ember.VERSION.match(/^1\.13\./)) {
+  Ember.deprecate(`Use of Ember Data 2+ with Ember 1.13 is deprecated. Please upgrade your version of Ember to 2.0 or higher.`, false, {
+    id: 'ds.version.ember-1-13',
+    until: '3.0.0'
```
fivetanley: I think this means we are telling people that we are supporting 1.13 ember until ember data 3.0. I don't know how achievable this goal really is.

bmac: Thanks for pointing this out @fivetanley. I'm not sure what to put for this value. Any thoughts @igorT? @wycats?

bmac: We chatted about this in the weekly meeting and decided we should add a warning for now and will likely choose the first LTS release as the version where we drop 1.13 support for good.

PR #3842, opened 2015-10-13
Remove ISO 8601 date.parse shim code

bmac:


packages/ember-data/lib/ext/date.js
```diff
@@ -16,46 +16,13 @@
...
-  } else {
-    timestamp = origParse ? origParse(date) : NaN;
-  }
-
-  return timestamp;
-};
+Ember.Date.parse = origParse;
 
 if (Ember.EXTEND_PROTOTYPES === true || Ember.EXTEND_PROTOTYPES.Date) {
   Date.parse = Ember.Date.parse;
```
pangratz: I guess this line doesn't make sense then anymore, right?

bmac: Good call.

pangratz: Save all the bytes! :hammer:

bmac:
We have decided to keep this code in the date transform and warn the user if there is no timezone or the timezone is not specified.

PR #3844, opened 2015-10-13
Remove conditional statement that is always true

### LINKED ISSUE #3629: likely broken code due to refactoring
this `if` will always be true, i suspect some meaning existed in the past but now something seems sideways.

https://github.com/emberjs/data/blob/master/packages/ember-data/lib/adapters/rest-adapter.js#L818

bmac:
Closes #3629

### DIFF
--- a/packages/ember-data/lib/adapters/rest-adapter.js
+++ b/packages/ember-data/lib/adapters/rest-adapter.js
@@ -815,15 +815,12 @@ export default Adapter.extend(BuildURLMixin, {
       var hash = adapter.ajaxOptions(url, type, options);
 
       hash.success = function(payload, textStatus, jqXHR) {
-        let response;
 
-        if (!(response instanceof AdapterError)) {
-          response = adapter.handleResponse(
-            jqXHR.status,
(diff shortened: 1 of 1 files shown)


packages/ember-data/lib/adapters/rest-adapter.js
```diff
@@ -815,15 +815,12 @@ export default Adapter.extend(BuildURLMixin, {
...
-            jqXHR.status,
-            parseResponseHeaders(jqXHR.getAllResponseHeaders()),
-            response || payload
-          );
-        }
+
+        let response = adapter.handleResponse(
+          jqXHR.status,
+          parseResponseHeaders(jqXHR.getAllResponseHeaders()),
+          response || payload
```
tchak: `response || payload` -> `payload`

tchak:
@bmac LGTM except for the little adjustment I made in the comments

packages/ember-data/lib/adapters/rest-adapter.js
stefanpenner: :+1:

PR #3848, opened 2015-10-13
[BUGFIX] EmbeddedRecordMixin should include the type serializing hasMany as ids

cibernox:
It is better explained with an example.

Lets say that that you have an error that contains many attachable items 
(`attachables: DS.hasMany(‘attachable’, { polymorphic: true })`) and the serializer looks like this:

```
export default ActiveModelSerializer.extend(DS.EmbeddedRecordsMixin, {
  attrs: {
    attachables: {
      serialize: 'ids',
      deserialize: 'ids'
    }
  }
});
```

At the moment the serialized json contains `attachables: [1,2,3, ...]` but that is useless for polymorphic relationships, the serialized array should contain `{id, type}` tuples (`attachables: [{type: 'pdf', id: 1}, {type: 'xls', id: 1}]`)

### DIFF
--- a/FEATURES.md
+++ b/FEATURES.md
@@ -30,3 +30,27 @@ entry in `config/features.json`.
 
   Enables `pushPayload` to return the model(s) that are created or
   updated via the internal `store.push`. [PR 4110](https://github.com/emberjs/data/pull/4110)
+
+- `ds-serialize-ids-and-types`
+
+  Enables a new `ids-and-type` strategy (in addition to the already existing `ids` and `records`) for
+  serializing has many relationships using the `DS.EmbeddedRecordsMixin` that  will include both
+  `id` and `type` of each model as an object.
+
+  For instance, if a use has many pets, which is a polymorphic relationship, the generated payload would be:
+
+  ```js
+  {
+    "user": {
+      "id": "1"
+      "name": "Bertin Osborne",
+      "pets": [
+        { "id": "1", "type": "Cat" },
+        { "id": "2", "type": "Parrot"}
+      ]
+    }
+  }
+  ```
+
+  This is particularly useful for polymorphic relationships not backed by STI when just including the id
+  of the records is not enough.
--- a/addon/serializers/embedded-records-mixin.js
+++ b/addon/serializers/embedded-records-mixin.js
@@ -1,5 +1,6 @@
 import Ember from 'ember';
 import { warn } from "ember-data/-private/debug";
+import isEnabled from 'ember-data/-private/features';
 
 var get = Ember.get;
 var set = Ember.set;
@@ -235,7 +236,7 @@ export default Ember.Mixin.create({
   },
 
   /**
-    Serialize `hasMany` relationship when it is configured as embedded objects.
+    Serializes `hasMany` relationships when it is configured as embedded objects.
 
     This example of a post model has many comments:
 
@@ -285,7 +286,7 @@ export default Ember.Mixin.create({
     ```
 
     The attrs options object can use more specific instruction for extracting and
-    serializing. When serializing, an option to embed `ids` or `records` can be set.
+    serializing. When serializing, an option to embed `ids`, `ids-and-types` or `records` can be set.
     When extracting the only option is `records`.
 
     So `{ embedded: 'always' }` is shorthand for:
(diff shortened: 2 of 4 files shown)


bmac:
@cibernox I don't think we can merge this as is without potentially breaking backwards compatibility. It would be better if there were a flag to opt into this behavior, possibly changing `serialize: 'ids',` to something like `serialize: 'ids-and-types',`.

cibernox:
I completely forgot about this PR :sweat: 

Then, would you the accept then a 3rd serializing strategy called `ids-and-types`? I don't remember the details but I think that this wouldn't be too hard to implement.

bmac:
I think we would accept a 3rd serializing strategy called `ids-and-types` (feel free to suggest a better name) in the EmbeddedRecordMixin.

cibernox:
Nahh, that name is as explicit at is can possibly be. I'll try to allocate some time over the weekend for this.

igorT:
@cibernox ping?

cibernox:
Santa Claus left this PR under the tree :+1:

cibernox:
@igort any comments on this PR? I will have some free time this weekend in case something needs to be changed

bmac:
@cibernox do you mind wrapping the changes in this pr in a [feature flag](https://github.com/emberjs/data/pull/3979)?

cibernox:
Flagged under `ds-serialize-ids-and-types`. 
However, what happens with the doc comments I've updated? If the API docs is generated from those it might mislead users. Should I remove them?

FEATURES.md
```diff
@@ -20,3 +20,27 @@ entry in `config/features.json`.
...
+      "pets": [
+        { "id": "1", "type": "Cat" },
+        { "id": "2", "type": "Parrot"}
+      ]
+    }
+  }
+  ```
+
+  This is particularly useful for polymorphic relationships not backed by STI when just including the id
+  of the records is not enought.
```
pangratz: `s/enought/enough/`

addon/-private/serializers/embedded-records-mixin.js
```diff
@@ -321,16 +372,38 @@ export default Ember.Mixin.create({
...
-    var includeRecords = this.hasSerializeRecordsOption(attr);
-    if (includeIds) {
+
+    if (this.hasSerializeIdsOption(attr)) {
       let serializedKey = this.keyForRelationship(attr, relationship.kind, 'serialize');
       json[serializedKey] = snapshot.hasMany(attr, { ids: true });
-    } else if (includeRecords) {
+    } else if (this.hasSerializeRecordsOption(attr)) {
       this._serializeEmbeddedHasMany(snapshot, json, relationship);
+    } else if (isEnabled("ds-serialize-ids-and-types") && this.hasSerializeIdsAndTypesOption(attr)) {
```
pangratz: If I am not mistaken this is not yet possible with feature flags (aka the code will not be stripped correctly when the feature flag is disabled).

Unfortunately you need to split this up:

``` js
} else {
  if (isEnabled("ds-serialize-ids-and-types")) {
    if (this.hasSerializeIdsAndTypesOption(attr)) {
      this._serializeHasManyAsIdsAndTypes(snapshot, json, relationship);
    }
  }
}
```

addon/-private/serializers/embedded-records-mixin.js
```diff
@@ -281,7 +282,7 @@ export default Ember.Mixin.create({
     ```
 
     The attrs options object can use more specific instruction for extracting and
-    serializing. When serializing, an option to embed `ids` or `records` can be set.
+    serializing. When serializing, an option to embed `ids`, `ids-and-types` or `records`can be set.
```
pangratz: Tiny nitpick but there is a space missing after `` `records` ``

pangratz:
> However, what happens with the doc comments I've updated? If the API docs is generated from those it might mislead users. Should I remove them?

Maybe a solution would be to mention in the comment that the functionality is only available when `ds-serialize-ids-and-types` feature is enabled? Not sure if this is less confusing though...

addon/-private/serializers/embedded-records-mixin.js
pangratz: @mmun would you accept a PR for [`babel-plugin-feature-flags`](https://github.com/ember-cli/babel-plugin-feature-flags) which replaces calls to `isEnabled("xxx")` with either `true` or `false` if a specific option is passed to the babel plugin?

cibernox:
Applied feedback, and also tagged the commit as `[FEATURE ds-serialize-ids-and-types]`

cibernox:
ping?

cibernox:
Rebased and green.

bmac:
cc @igorT we should look at this tomorrow.

cibernox:
Rebased.

cibernox:
:+1:

PR #3850, opened 2015-10-13
[DOC] Document snapshotRecordArray param passed to adapter.findAll

HeroicEric:
The [docs for DS.Adapter#findAll](https://github.com/emberjs/data/blob/4e79b27d48b71f8ad737471db706149164c4a27a/packages/ember-data/lib/system/adapter.js#L152) already include this and it is already being passed
[when DS.Store#findAll](https://github.com/emberjs/data/blob/8f83fbf27acdd642cb824e3286832acebdac0da1/packages/ember-data/lib/system/store/finders.js#L121) is used.


bmac:
Thanks @HeroicEric

Issue #3851, opened 2015-10-14
Records are remaining in root.loaded.created.inFlight after server returns validation error.


arenoir:
I noticed issue #1764 is back and have a test that proves it.  It has to do with a `InvalidError` object being returned.

Issue #3677, opened 2015-08-19
rollbackAttributes() not working after more than one 422 error


akappen:
I have an API server that performs server side validation of records and responds with the expected 422 status and a body of validation errors. The model correctly parses the validation errors in the body and makes them available on the errors object for display in the template. At this point if the user cancels the edit operation, I rollback the attribute changes to revert the model to the last saved version. This all works as expected.

If the model is saved two or more times with validation errors, each request responds with the same 422 status and validation errors body. If I try to rollback the attribute values at this point the original attributes are not restored!

I attempted to capture this in a test, but I couldn't figure out how to test this 422 error case. If anyone can point me to tests in ember data that exercise the 422 response condition I might be able to use those as a guide.

Instead of having a nice reproducible test, I can walk through the problem in the browser's javascript console. The first example shows proper rollback after a single 422 failure. The second shows a broken rollback after two 422 failures. This behavior happens on all of my models, not just the task model used in the example.

```
DEBUG: -------------------------------
DEBUG: Ember                    : 1.13.7
DEBUG: Ember Data               : 1.13.8
DEBUG: jQuery                   : 1.11.3
DEBUG: Ember Simple Auth        : 0.8.0-beta.3
DEBUG: Ember Simple Auth Devise : 0.8.0-beta.3
DEBUG: -------------------------------
```

With one 422 response the rollback works:
```
> var task
undefined
> $E.store.findRecord('task', 1).then(function(t) {task = t;})
Promise {_id: 181, _label: undefined, _state: undefined, _result: undefined, _subscribers: Array[0]}
> task
Class {id: "1", store: Class, container: Container, _internalModel: InternalModel, currentState: Object…}
> task.get("name")
"Make a card!"
> task.set("name", null)
Class {id: "1", store: Class, container: Container, _internalModel: InternalModel, currentState: Object…}
> task.get("name")
null
> task.save()
Class {__ember1440016252384: null, __nextSuper: undefined, __ember_meta__: Object}
jquery.js:9664 PUT http://localhost:4200/api/v1/tasks/1 422 (Unprocessable Entity)
. . . cut stack trace . . .
Error: The adapter rejected the commit because it was invalid
. . . cut stack trace . . .
> task.get("name")
null
> task.get("errors.name")
[Object, Object]
> task.get("errors.name")[0].message
"Name can't be blank"
> task.get("errors.name")[1].message
"Name is too short (minimum is 10 characters)"
> task.rollbackAttributes()
undefined
[...]

Issue #1764, opened 2014-02-24
Record remains inFlight after consecutive 422 responses 


sheldonbaker:
Assuming the server in both cases returns 422 with the usual error JSON as a payload:

```
record.save().then(null, function(record) {
  record.get('currentState.stateName') // 'root.loaded.updated.invalid'

  record.save().then(null, function(record) {
    record.get('currentState.stateName') // 'root.loaded.updated.inFlight'
  });
});
```

Perhaps `becameInvalid` should always be triggered [here](https://github.com/emberjs/data/blob/master/packages/ember-data/lib/system/model/errors.js#L132) regardless of whether or not the record `wasEmpty`?

PR #3852, opened 2015-10-14
test demonstrating issue #3851 and #1764 possibly related to #3677

arenoir:
Records are remaining in `root.loaded.created.inFlight` after server returns validation error.


arenoir:
I think all saving tests that return a rejected promise should return a `InvalidError` object.

Issue #3707, opened 2015-08-26
Adding an error to an already invalid record throws unhandled `becameInvalid` event


offirgolan:
Ember Data > 1.13.6
JSBin: http://emberjs.jsbin.com/tapeva/edit?html,js,output

1. Change ember-data to 1.13.7 or above
2. Type in text field
3. Focus out
4. Type in text field again
5. Focus out
6. Error will be shown in console

`Error: Attempted to handle event 'becameInvalid' on <App.Book:ember418:null> while in state root.loaded.created.invalid.`

[Original StackOverflow Post] (https://stackoverflow.com/questions/32070450/error-attempted-to-handle-event-becameinvalid-while-in-state-root-loaded-crea)

thomasdurin:
Same problem for me. It seems that removing the model's error do not send the event `becameValid`.

bmac:
@tchak I'm not sure what the expected behavior is here. Do you mind taking a look at this issue?

offirgolan:
@bmac i believe the error was introduced through PR #3555 but don't quote me on that.

tchak:
will investigate tonight or tomorrow

stefanpenner:
I tried to replicate in test form, but was unable to. @offirgolan maybe you can use my PR as a stepping stone?

offirgolan:
Okay so after a long time of debugging I figured out what's going on. @stefanpenner couldn't replicate the issue since it was actually fixed but I dont know when or how. Here is my findings:

This is how the flow should be:

1. start in `root.loaded.created.uncommitted` state
2. Add an error which will put the record in `root.loaded.created.invalid` state
3. Change value of `model.firstName` which will trigger `didSetProperty` from the invalid state. This will call `errors.remove` which will put you back in `root.loaded.created.uncommitted` state

This is what was actually happening

1. start in `root.loaded.created.uncommitted` state
2. Add an error which will put the record in `root.loaded.created.invalid` state
3.  Change value of `model.firstName` which will trigger `didSetProperty` from the invalid state. This will call `errors.remove`

[...]

PR #3853, opened 2015-10-14
fix transitioning into invalid state

arenoir:
Fixes issue #3851.  

Removes events from error class that was causing side effects.   I don't think the error class should be responsible for updating the model state.

### DIFF
--- a/addon/system/model/errors.js
+++ b/addon/system/model/errors.js
@@ -107,11 +107,30 @@ export default Ember.ArrayProxy.extend(Ember.Evented, {
     @param {Object} target
     @param {Function} becameInvalid
     @param {Function} becameValid
+    @deprecated
   */
   registerHandlers: function(target, becameInvalid, becameValid) {
+    Ember.deprecate(
+      `Record errors will no longer be evented.`, false, {
+        id: 'ds.errors.registerHandlers',
+        until: '3.0.0'
+      });
+
+    this._registerHandlers(target, becameInvalid, becameValid);
+  },
+
+
+  /**
+    Register with target handler
+
+    @method _registerHandlers
+    @private
+  */
+  _registerHandlers: function(target, becameInvalid, becameValid) {
     this.on('becameInvalid', target, becameInvalid);
     this.on('becameValid', target, becameValid);
   },
+
 
   /**
     @property errorsByAttributeName
@@ -214,19 +233,35 @@ export default Ember.ArrayProxy.extend(Ember.Evented, {
     @method add
     @param {String} attribute
     @param {(Array|String)} messages
+    @deprecated
   */
   add: function(attribute, messages) {
+    Ember.warn(`Interacting with a record errors object will no longer change the record state.`, false, {
+      id: 'ds.errors.add'
+    });
+
     var wasEmpty = get(this, 'isEmpty');
 
+    this._add(attribute, messages);
+
+    if (wasEmpty && !get(this, 'isEmpty')) {
+      this.trigger('becameInvalid');
+    }
+  },
+
+
+  /**
+    Adds error messages to a given attribute without sending event.
+
+    @method _add
+    @private
+  */
+  _add: function(attribute, messages) {
     messages = this._findOrCreateMessages(attribute, messages);
     this.addObjects(messages);
     get(this, 'errorsByAttributeName').get(attribute).addObjects(messages);
 
     this.notifyPropertyChange(attribute);
-
-    if (wasEmpty && !get(this, 'isEmpty')) {
-      this.trigger('becameInvalid');
-    }
   },
 
   /**
@@ -277,8 +312,29 @@ export default Ember.ArrayProxy.extend(Ember.Evented, {
 
     @method remove
     @param {String} attribute
+    @deprecated
   */
   remove: function(attribute) {
+    Ember.warn(`Interacting with a record errors object will no longer change the record state.`, false, {
+      id: 'ds.errors.remove'
+    });
+
+    if (get(this, 'isEmpty')) { return; }
+
+    this._remove(attribute);
+
+    if (get(this, 'isEmpty')) {
+      this.trigger('becameValid');
+    }
+  },
+
+  /**
+    Removes all error messages from the given attribute without sending event.
+
+    @method _remove
+    @private
+  */
+  _remove: function(attribute) {
     if (get(this, 'isEmpty')) { return; }
 
     let content = this.rejectBy('attribute', attribute);
@@ -286,10 +342,6 @@ export default Ember.ArrayProxy.extend(Ember.Evented, {
     get(this, 'errorsByAttributeName').delete(attribute);
 
     this.notifyPropertyChange(attribute);
-
-    if (get(this, 'isEmpty')) {
-      this.trigger('becameValid');
-    }
   },
 
   /**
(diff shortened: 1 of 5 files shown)


packages/ember-data/tests/integration/adapter/store-adapter-test.js
```diff
@@ -470,15 +470,13 @@ test("if a created record is marked as invalid by the server, it enters an error
...
       ok(get(yehuda, 'errors.name'), "The errors.name property exists");
-
       set(yehuda, 'updatedAt', true);
       equal(get(yehuda, 'isValid'), false, "the record is still invalid");
-
       set(yehuda, 'name', "Brohuda Brokatz");
-
-      equal(get(yehuda, 'isValid'), true, "the record is no longer invalid after changing");
+      //shouldn't we use rollbackAttributes. or perhaps a rollbackAttribute method is needed.
+      //keeping track of changes this is fragile.
```
bmac: @arenoir I'd like to get a better understanding of what you are suggesting here. Are you saying rollbackAttributes should be responsible for transitioning a model out of the invalid state?

pangratz:
To be semver compatible, this change entails an increase of the major version since it changes the current behavior. Maybe it should be possible to opt out of this behavior via an environment flag and add deprecation warning that the current behavior is about to change in `3.0`?

---

Apart from that, I in general agree with

> I don't think the error class should be responsible for updating the model state.

I think ember data would need a `didUpdateAttribute` hook so the current situation could be implemented nicely once the errors for an attribute are not cleared automatically when it changes:

``` javascript
DS.Model.extend({
  didUpdateAttribute: function(name, oldValue, newValue) {
    var errors = this.get('errors');
    errors.clearErrorsFor(attributeName);

    if (errors.get('isEmpty')) {
      // currently not public ...
      this.send('becameValid');
    }
   }
});
```

packages/ember-data/tests/integration/adapter/store-adapter-test.js
arenoir: I think changing the state should be more explicit. The rollback method to me says put this record back to the state it was in when instantiated.

arenoir:
@pangratz I would argue that this behavior is a bug and not documented?   However I know you are correct; it is a breaking change.  

I like the idea of a didUpdateAttribute hook and that would restore the behavior thus allowing us to remove the side effect causing event bindings.  I will try to implement it.  In the mean time I am using this pull request in my application as it fixes a bigger breaking change.

arenoir:
@pangratz @bmac,  I have restored the functionality to become valid when errors are removed.  I believe this is a better solution than sending events from the error class.

bmac:
Thanks for the update @arenoir. This code looks good. 

cc @igorT @tchak for review.

packages/ember-data/lib/system/model/errors.js
```diff
@@ -91,19 +91,8 @@ var MapWithDefault = Ember.MapWithDefault;
...
 
-    @method registerHandlers
-    @param {Object} target
-    @param {Function} becameInvalid
-    @param {Function} becameValid
-  */
-  registerHandlers: function(target, becameInvalid, becameValid) {
-    this.on('becameInvalid', target, becameInvalid);
-    this.on('becameValid', target, becameValid);
-  },
```
pangratz: I don't know how many apps/addons rely on those events being triggered when registered via `model.get('errors').on('becameValid')` or `model.get('errors').registerHandlers()`. Since `Ember.Evented` is removed in this PR, code which relies on the `on` method being available on the `DS.Errors` throws an error.

The `registerHandler` method isn't advertised but it also hasn't been explicitly marked as `@private` so I technically this would break public API, I guess. If I am misinterpreting semver, please ignore me :pensive: .

Again, just to be clear: I am :+1: :+1: on removing those events and eventually make clearing the errors for an attribute an opt in solution, but I just want to point out possible implications for semver.

A possible, compatible solution could be to overwrite the `on` method and throw a deprecation...

pangratz:
> I believe this is a better solution than sending events from the error class.

:100: agree

packages/ember-data/lib/system/model/errors.js
bmac: I think @pangratz is correct. @arenoir do you mind adding `registerHandlers` back with a deprecation warning and continuing to use the `Ember.Evented` mixin even though the behavior is no longer needed?

tchak: The behivor is not dicumented but it is definitly used by some applications. Basicly the use case is to be able to implement client side validation in terms of `errors` object. Pushing an error in to `errors` object transitions the record in to invalid state.

I am fine with deprecating it and removing in next major version. But I am not sure where the deprecation should be. For example in my app, I don't care about events at all. What I do rely on, is the fact that if I am pushing an error in to `errors` object, the record became invalid.

arenoir:
@bmac okay I have added the Evented functionality back and added my attempt to deprecate it.   The problem is we don't want to deprecate `add`, `remove`, `clear` but only warn that they will no longer be responsible for modifying a records state.

tchak:
@arenoir I am not sure how I feel about this... It feels like manual `errors` management have valid use cases... And if we decide that we want to remove this functionality we should simplify the errors object.

In any case, this probably should go behind a flag to begin with.

arenoir:
@tchak,  I agree with the use case of manually adding errors but I don't think doing so should update the state of the record. 

If the server has put the record into the invalid state clearing the errors shouldn't change it to 'uncommitted' state.  As it stands now calling willCommit from the 'invalid' state first clears the errors inconsequential changing the state to 'uncommitted' and then transitions the state to 'inflight'.

tchak:
@arenoir I am fine with this reasoning.
A thew observations:
- if we doing so, I think we need a clearly documented api to manually mark a record as invalid.
- we do agree that `_add` and co methods are temporary, until we can remove deprecations?
- Speaking of deprecations, shouldn't they be warnings?

arenoir:
@tchak I agree there should be a public api to change a records state. The `_add` , `_remove` and `_clear` private methods are temporary.  Once we remove the `Evented` functionality from the error class the temporary methods can be removed.  As far a warnings vs deprecations I am in favor of a warning as the methods are not going away their behavior is just changing a bit.

I think this will fix some other inconsistent bugs like #3707.

tchak:
@arenoir :+1: for warnings. For the rest, I am :+1: on merging it.

arenoir:
@tchak I switched to warnings and rebased.

tchak:
I am happy with this PR.
@bmac @igorT can we merge it as is, or dose it have to go behind a feature flag or something? It's not really a feature and it do not actually change current behaviour. Just adding some warnings.

tchak:
LGTM. @bmac is it ok with you to merge this ?

bmac:
Sorry @arenoir. I missed @tchak ping. Do you mind rebasing this pr so it can be merged?

arenoir:
@bmac okay I rebased with master.  Sorry It took so long I was on holiday.

bmac:
Thanks @arenoir

addon/system/model/errors.js
```diff
@@ -214,19 +233,35 @@ export default Ember.ArrayProxy.extend(Ember.Evented, {
     @method add
     @param {String} attribute
     @param {(Array|String)} messages
+    @deprecated
```
forgetboy: From the conversation, it seems, these methods were not meant to be deprecated. Should this be undone?

forgetboy:
Going through the conversation, it seems, that add, remove and clear methods shouldn't have been marked deprecated. Only the warnings need to exist. @abuiles, could you verify this? Thanks.

PR #3854, opened 2015-10-14
Merge `normalizeRelationships` and `setupRelationships` methods in store

trentmwillis:
While working on some performance related tasks, I noticed that `normalizeRelationships` and `setupRelationships` both loop over the relationships for a record. The only time these functions are used is in sequence, so it looks like these could be merged to cut down on some time complexity.

Will be trying this out in our application tomorrow to see if it has any side effects, but it doesn't appear to so far.

### DIFF
--- a/packages/ember-data/lib/system/store.js
+++ b/packages/ember-data/lib/system/store.js
@@ -1346,7 +1346,7 @@ Store = Service.extend({
     }
     if (data) {
       // normalize relationship IDs into records
-      this._backburner.schedule('normalizeRelationships', this, '_setupRelationships', internalModel, internalModel.type, data);
+      this._backburner.schedule('normalizeRelationships', this, '_setupRelationships', internalModel, data);
       this.updateId(internalModel, data);
     }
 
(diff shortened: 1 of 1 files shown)


tchak:
LGTM :+1:

trentmwillis:
Any chance this could get merged? Just updated to remove the `type` param (not sure it was needed to begin with).

wecc:
@igorT r?

packages/ember-data/lib/system/store.js
```diff
@@ -2135,13 +2108,16 @@ function setupRelationships(store, record, data) {
       relationship = record._relationships.get(key);
       relationship.updateMeta(data.relationships[key].meta);
     }
+
     var value = data.relationships[key].data;
```
igorT: can we normalize here, and put the comment from 1727 back in?

igorT:
:+1: after the small refactor. Thanks for the PR

packages/ember-data/lib/system/store.js
trentmwillis: I can, but not sure I see the point in doing that. If we normalize here, then we will have to check that `value` exists and the `kind` of relationship in two places (here and then again below).

igorT: Mostly for code readability/place to put the deleted comment/ease of
understanding

trentmwillis:
@igorT updated. Let me know if there's anything else.

bmac:
Thanks @trentmwillis. I have rebased this pr and merged is as https://github.com/emberjs/data/commit/828f8eaef753320bf39793b6b3f54a64c970f5a1.

PR #3855, opened 2015-10-14
Deprecate complex objects as an attributes `defaultValue`.

### LINKED ISSUE #3766: [Request] deprecate `defaultValue: primitiveValue`
https://github.com/emberjs/data/blob/c208f2c299ecd5566ad9d0eee0a89f9913e924c6/packages/ember-data/lib/system/model/attributes.js#L231

It is quite common to see: `defaultValue: []` but this results in a shared array between all instances, obviously nearly never something a user actual intends.

An alternative, and likely the better option, is to only allow primitive values, and require functions for complex objects. If someone wants the old behavior, that can close over a shared complex object.

bmac:
They often do not work how a user expects.
Closes #3766

### DIFF
--- a/packages/ember-data/lib/system/model/attributes.js
+++ b/packages/ember-data/lib/system/model/attributes.js
@@ -226,7 +226,13 @@ function getDefaultValue(record, options, key) {
   if (typeof options.defaultValue === "function") {
     return options.defaultValue.apply(null, arguments);
   } else {
-    return options.defaultValue;
+    let defaultValue = options.defaultValue;
+    Ember.deprecate(`Non primitive defaultValues are deprecated because they are shared between all instances. If you would like to use a complex object as a default value please provide a function that returns the complex object.`,
+      typeof defaultValue !== 'object' || defaultValue === null, {
+        id: 'ds.defaultValue.complex-object',
+        until: '3.0.0'
+      });
+    return defaultValue;
   }
 }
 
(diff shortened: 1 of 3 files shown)


wecc:
:clap:

packages/ember-data/lib/system/model/attributes.js
```diff
@@ -226,7 +226,13 @@ function getDefaultValue(record, options, key) {
   if (typeof options.defaultValue === "function") {
     return options.defaultValue.apply(null, arguments);
   } else {
-    return options.defaultValue;
+    let defaultValue = options.defaultValue;
+    Ember.deprecate(`Non primative defaultValues are deprecated because they are shared between all instances. If you would like to use a complex object as a default value please provide a function that returns the complex object.`,
```
sly7-7: small typo 'primitive'

packages/ember-data/tests/unit/model-test.js
```diff
@@ -611,6 +611,41 @@ test("a defaultValue function gets the record, options, and key", function() {
...
+  var store = createStore({
+    tag: Tag
+  });
+
+  run(function() {
+    tag = store.createRecord('tag');
+  });
+  expectDeprecation(function() {
+    get(tag, 'tagInfo');
+  }, /Non primative defaultValues are deprecated/);
```
sly7-7: same here ;)

igorT:
:heart:

PR #3858, opened 2015-10-14
[DOC] remove documentation for `isError` in states

pangratz:
This flag moved out of the state machine in d5de4e1.


bmac:
Thanks @pangratz

PR #3859, opened 2015-10-15
[BUGFIX beta] Correctly handle invalid errors without payload or pointer

### LINKED ISSUE #3851: Records are remaining in root.loaded.created.inFlight after server returns validation error.

tchak:
Closes #3851, #3852

### DIFF
--- a/packages/ember-data/lib/system/model/internal-model.js
+++ b/packages/ember-data/lib/system/model/internal-model.js
@@ -644,6 +644,8 @@ InternalModel.prototype = {
       }
     }
 
+    this.send('becameInvalid');
+
     this._saveWasRejected();
   },
 
--- a/packages/ember-data/lib/system/model/states.js
+++ b/packages/ember-data/lib/system/model/states.js
@@ -326,8 +326,8 @@ var DirtyState = {
       didSetProperty(internalModel, context);
     },
 
+    becameInvalid: Ember.K,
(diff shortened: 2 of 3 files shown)


packages/ember-data/lib/system/model/internal-model.js
```diff
@@ -644,6 +646,10 @@ InternalModel.prototype = {
       }
     }
 
+    if (isEmpty(Object.keys(errors))) {
+      this.send('becameInvalid');
+    }
```
pangratz: Why not always send the `becameInvalid` event?

bmac:
@tchak could you amend this commit to start with `[BUGFIX beta]`

packages/ember-data/lib/system/model/internal-model.js
tchak: because it will boule send it (because of events on errors) and it breaks lots of tests

bmac: What happens if this code runs to put the record in the invalid state then an error is added to the errors object?

tchak:
@bmac so after some thought it probably make sens to be able to call `becameInavlid` multiple times. Fixed it.

bmac:
Looks good. :+1:

PR #3862, opened 2015-10-16
Provide some context about the http request when the RESTAdapter rejects

bmac:
### Before 
screen shot 2015-10-16 at 7 57 54 am

### After
screen shot 2015-10-16 at 7 58 28 am

### DIFF
--- a/packages/ember-data/lib/adapters/rest-adapter.js
+++ b/packages/ember-data/lib/adapters/rest-adapter.js
@@ -743,9 +743,10 @@ export default Adapter.extend(BuildURLMixin, {
     @param  {Number} status
     @param  {Object} headers
     @param  {Object} payload
+    @param  {String} errorMessage
     @return {Object | DS.AdapterError} response
   */
-  handleResponse: function(status, headers, payload) {
+  handleResponse: function(status, headers, payload, errorMessage) {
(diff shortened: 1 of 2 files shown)


fivetanley:
@bmac LGTM. can you resolve conflicts?

packages/ember-data/lib/adapters/rest-adapter.js
```diff
@@ -743,9 +743,10 @@ export default Adapter.extend(BuildURLMixin, {
     @param  {Number} status
     @param  {Object} headers
     @param  {Object} payload
+    @param  {String} errorMessage
     @return {Object | DS.AdapterError} response
   */
-  handleResponse: function(status, headers, payload) {
+  handleResponse: function(status, headers, payload, errorMessage) {
```
tchak: I feel like we should not call the param `errorMessage`. It can be `ok`, or `created` in case of success. I would be ok with `statusText`, even if in case of obvious errors we "augment" the proper `statusText`.

btecu:
I was looking forward this. @tchak are you planning on doing something better? ...

bmac:
@btecu I closed this pr because I think https://github.com/emberjs/data/pull/3930 is going to be a better solution for this problem.

PR #3864, opened 2015-10-16
Throw a more helpful error message if the response from queryRecord i…

bmac:
…s empty

### DIFF
--- a/addon/system/store/finders.js
+++ b/addon/system/store/finders.js
@@ -190,6 +190,7 @@ export function _queryRecord(adapter, store, typeClass, query) {
     var record;
     store._adapterRun(function() {
       var payload = normalizeResponseHelper(serializer, store, typeClass, adapterPayload, null, 'queryRecord');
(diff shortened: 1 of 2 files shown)


packages/ember-data/lib/system/store/finders.js
```diff
@@ -175,6 +175,7 @@ export function _queryRecord(adapter, store, typeClass, query) {
     var record;
     store._adapterRun(function() {
       var payload = normalizeResponseHelper(serializer, store, typeClass, adapterPayload, null, 'queryRecord');
+      Ember.assert('`store.queryRecord` expected the adapter to return one record but the response from the adapter was empty.', payload.data);
```
wecc: Shouldn't this assert that `payload.data` is an object, not just truthy?

fivetanley: confirm.

bmac:
Igor and I looked at this again and we released its not clear where in the stack to reject a missing `findQuery` response when there is no information form the protocol (like a 404) thus requiring you to inspect the payload to determine if there response is empty.

Issue #3833, opened 2015-10-08
Replace hasMany relationships fails


IsaiahJTurner:
Doing something like: 
`ama.set("questions", [Question, Question, Question]);`
does not work.

pangratz:
Can you give a little bit more information? What version are you using? What does the model definition for `ama` look like? Best would be a reproduced example in a JSBin. Here is a starting point: http://emberjs.jsbin.com/lavacavige/edit?html,js,output

bmac:
Relationships are more complex object then an array so its not possible to set the relationship property directly. Instead can use the [setObjects](http://emberjs.com/api/classes/Ember.MutableArray.html#method_setObjects) method on the relationship. 
`ama.get("questions").setObjects([Question, Question, Question]);`

bmac:
My bad @IsaiahJTurner. 

@pangratz reminded me that there is special code to handle the case of setting an array: https://github.com/emberjs/data/blob/fcf54523b4b9f87e403baa5a1c73af0486914936/packages/ember-data/lib/system/relationships/has-many.js#L146-L152

What is the error that you are seeing?

IsaiahJTurner:
I made a JSBin: http://emberjs.jsbin.com/xemunezoki/edit?html,js,console,output

pangratz:
OK, so the problem is that `store.findRecord` returns a `PromiseObject`. This can't be added directly to a hasMany relationship. The `findRecord` resolves with the specific record, which in turn can be added to the relationship.

One way to do this is via `RSVP.all`, which takes an array as argument and resolves when all items in the array are resolved. So the code within your action [would become something like](http://emberjs.jsbin.com/ciyefi/1/edit?html,js,output):

``` js
update: function() {
  var newBooks = [
    this.store.find("book", 4)
  ];

  var model = this.get('model');
  Ember.RSVP.all(newBooks).then(function(newBooks) {
    model.get('books').setObjects(newBooks);

    // or you can use model.set('books', newBooks);
  });
}
```

---

The thrown error message `Uncaught TypeError: Cannot read property 'type' of undefined` when setting a `PromiseObject` is not very helpful. We could add an assertion below [this line](https://github.com/emberjs/data/blob/fcf54523b4b9f87e403baa5a1c73af0486914936/packages/ember-data/lib/system/relationships/has-many.js#L149) and check if the contents of the array are actual instances of `DS.Model`.

@IsaiahJTurner do you want to submit a PR?

pangratz:
[...]

PR #3865, opened 2015-10-16
Assert that passed array to set a hasMany consists of records

pangratz:
This addresses #3833

### DIFF
--- a/packages/ember-data/lib/system/relationships/has-many.js
+++ b/packages/ember-data/lib/system/relationships/has-many.js
@@ -144,9 +144,13 @@ export default function hasMany(type, options) {
       return relationship.getRecords();
     },
     set: function(key, records) {
+      Ember.assert("You must pass an array of records to set a hasMany relationship", isArrayLike(records));
+      Ember.assert(`All elements of a hasMany relationship must be instances of DS.Model, you passed ${Ember.inspect(records)}`, (function() {
+        return Ember.A(records).every((record) => Model.detectInstance(record) );
+      })());
+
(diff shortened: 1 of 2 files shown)


packages/ember-data/lib/system/relationships/has-many.js
```diff
@@ -144,9 +144,14 @@ export default function hasMany(type, options) {
       return relationship.getRecords();
     },
     set: function(key, records) {
+      Ember.assert("You must pass an array of records to set a hasMany relationship (elements must be instances of DS.Model)", (function() {
+        if (!isArrayLike(records)) { return false; }
+
+        return Ember.A(records).every((record) => Model.detectInstance(record) );
+      })());
```
pangratz: I am not sure how helpful this assertion message is (considering the use case in #3833). Should I add an explicit check if a `DS.PromiseObject` is present in the passed array, and if so, warn that the passed array contains the result of a `store.findRecord`/`store.queryRecord` which is not supported and how it should be resolved?

bmac: I would keep the existing assertion about an array. Then add an new assertion that says something like `all elements of a hasMany relationship must be instances of DS.Model but Ember Data found ${Ember.inspect(badRecord)}`. 

It may be enough to just test if the first object (if it exists) in the array is a DS.Model and assume that all elements in the array are likely to be of the same type.

igorT: :+1: but no reason not to test all of them that i can think of

bmac:
Thanks @pangratz

PR #3866, opened 2015-10-16
Allow store.push to accept { data: null }

### LINKED ISSUE #3790: store.push throws with JSONAPI document { data: null }
The [JSONAPI spec](http://jsonapi.org/format/#fetching-resources-responses-200) indicates that the following payload is a valid response when asking for a has-one relationship from a server:

```json
{"data": null}
``` 

However, when that normalized response makes it to `store.push`, ED throws the following error:

```
Assertion Failed: You must include an 'id' for undefined in an object passed to 'push'"
```

It should be simple to add a failing test case for this in `store-test.js` but I'm getting timeout errors when trying to install the ED dependencies at the moment.

bmac:
Closes #3790

cc @wecc could review this pr?

### DIFF
--- a/packages/ember-data/lib/system/store.js
+++ b/packages/ember-data/lib/system/store.js
@@ -1683,7 +1683,13 @@ Store = Service.extend({
       return internalModels.map((internalModel) => internalModel.getRecord());
     }
 
-    var internalModel = this._pushInternalModel(data.data || data);
+    if (data.data === null) {
+      return null;
+    }
+
+    Ember.assert(`Expected an object in the 'data' property in a call to 'push' for ${data.type}, but was ${Ember.typeOf(data.data)}`, Ember.typeOf(data.data) === 'object');
+
+    var internalModel = this._pushInternalModel(data.data);
 
     return internalModel.getRecord();
   },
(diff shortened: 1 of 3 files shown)


packages/ember-data/lib/system/store.js
```diff
@@ -1683,6 +1683,10 @@ Store = Service.extend({
       return internalModels.map((internalModel) => internalModel.getRecord());
     }
 
+    if (data.data === null) {
+      return null;
+    }
+
     var internalModel = this._pushInternalModel(data.data || data);
```
wecc: I think this special case (I've never really liked it) should be considered if we return early if `data.data === null` above.

bmac: I believe this works because the the `return null;` never gets hit if `data.data` is undefined. I'd like to remove this `|| data` from this line if possible. It seems like we could do that if we pull one of the assertions out of `_pushInternalModel` and put it into this method.

igorT:
@wecc are we good here?

wecc:
LGTM!

@bmac squash?

bmac:
@wecc squashed.

igorT:
Maybe a warning for people upgrading that we switched to json api?

jherdman:
Would it be possible to back-port this to 2.2.0?

bmac:
@jherdman yes I can release a 2.2.1 tomorrow with this fix.

bmac:
Sorry for the delay @jherdman. I just published 2.2.1 with this commit.

jherdman:
No worries! I greatly appreciate the release.

PR #3867, opened 2015-10-17
Allow serializers to normalize response, remove old internal serializers code

lostinpatterns:
Continuing https://github.com/emberjs/data/commit/c5e2859e72b230f831d5cfe869cee504bfc3245b, we should use the normalized response directly from our serializers whenever we receive a response from a save, just like we do with responses from finders. I've removed the old serializer code that's now unnecessary and am instead passing the payload data directly to didSaveRecord. This allows the same serializers to be used for normalizing the response from find and save calls rather than internal code.


wecc:
LGTM! :+1: 

Can you prefix the commit with `[CLEANUP beta]`?

lostinpatterns:
No problem, done. Thanks!

bmac:
Looks great. Thanks @lostinpatterns.

Issue #3265, opened 2015-06-08
changed attributes should be reset when matching data is pushed


bmac:
Changed attributes should be reset to if data is pushed into the store that matches the current "changed" value.

Example:

```js
var record = store.push({ id:1, name: 'Igor' }) -> attributes.name === undefined
record.set('name', 'Tomster') -> attributes.name === 'Tomster'
store.push({ id:1, name: 'Tomster' }) -> attributes.name === undefined
```

pangratz:
This issue can be closed since #3868 has been merged.

PR #3868, opened 2015-10-17
Reset changed attributes when matching data is pushed

pangratz:
If an attribute is changed locally and data is pushed to the store which
matches the changed attribute, then this changed attribute is considered
as confirmed and is no more considered as dirty.

---

This addresses #3265


igorT:
Can you please squash. Not super happy about one more method, but don't have any great ideas right now :(

pangratz:
Done. Also, is this considered a breaking change?

bmac:
I believe this is a bugfix.

bmac:
Thanks @pangratz

Issue #3871, opened 2015-10-19
Not able to override serializePolymorphicType for Embedded records


maheshiv:
Here is my models,

```js
APP.User = DS.Model.extend({
  name: DS.attr('string'),
  itemLock: DS.belongsTo('item-lock')
});
```

```js
APP.ItemLock = DS.Model.extend({
  item: DS.belongsTo('item', {polymorphic: true}),
  UserId: DS.belongsTo('user')
});
```

Here is the serializers,

```js
APP.UserSerializer = DS.RESTSerializer.extend({
  attrs: {
    itemLock: {embedded: 'always'}
  }
});
```
```js
APP.ItemLockSerializer = DS.RESTSerializer.extend({
  serializePolymorphicType: function(snapshot, json, relationship) {
    console.log('snapshot:-', snapshot);
  }
});
```

Here is sample response
```js
{
 user: {
  id: 1,
  name: 'John',
  itemLock: {
   id: 1,
   item: 2,
   itemType: 'assignment'
  }
 }
}
```

I am using Ember-data V1.13.8. I am not able to override serializePolymorphicType. Any thoughts on this?

pangratz:
I think this issue can be closed since #3875 and #3876 have been merged...

PR #3875, opened 2015-10-20
[BUGFIX beta] serialize type for embedded, polymorphic belongsTo

pangratz:
This fixes the issue where the `serializePolymorphicType` hook on the
serializer hasn't been invoked when an embedded, polymorphic belongsTo
relationship is serialized.

---

This issue has been reported in #3871 (though for ember-data `v1.13.8`).

### DIFF
--- a/packages/ember-data/tests/integration/serializers/embedded-records-mixin-test.js
+++ b/packages/ember-data/tests/integration/serializers/embedded-records-mixin-test.js
@@ -1157,6 +1157,57 @@ test("serialize with embedded object (belongsTo relationship)", function() {
   });
 });
 
+test("serialize with embedded object (polymorphic belongsTo relationship)", function() {
+  env.registry.register('serializer:super-villain', DS.RESTSerializer.extend(DS.EmbeddedRecordsMixin, {
+    attrs: {
+      secretLab: { embedded: 'always' }
+    }
+  }));
+
+  SuperVillain.reopen({
+    secretLab: DS.belongsTo('secret-lab', { polymorphic: true })
+  });
+
+  var json, tom;
+  run(function() {
+    tom = env.store.createRecord(
+      'super-villain',
+      {
+        id: "1",
+        firstName: "Tom",
+        lastName: "Dale",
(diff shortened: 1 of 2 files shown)


packages/ember-data/tests/integration/serializers/embedded-records-mixin-test.js
```diff
@@ -1157,6 +1157,46 @@ test("serialize with embedded object (belongsTo relationship)", function() {
...
+    lastName: get(tom, "lastName"),
+    homePlanet: get(tom, "homePlanet").get("id"),
+    secretLabType: 'batCave',
+    secretLab: {
+      id: get(tom, "secretLab").get("id"),
+      minionCapacity: get(tom, "secretLab").get("minionCapacity"),
+      vicinity: get(tom, "secretLab").get("vicinity"),
+      infiltrated: true
+    }
+  });
```
pangratz: I am not sure if this is the way this should work, or if the `type` of the embedded, polymorphic record should be a property within the `secretLab` object ...

bmac: I'm also not sure about the correct behavior here. cc @igorT @cyril-sf

packages/ember-data/tests/integration/serializers/embedded-records-mixin-test.js
```diff
@@ -1157,6 +1157,46 @@ test("serialize with embedded object (belongsTo relationship)", function() {
...
+
+  SuperVillain.reopen({
+    secretLab: DS.belongsTo('secret-lab', { polymorphic: true })
+  });
+
+  var json, tom;
+  run(function() {
+    tom = env.store.createRecord(
+      'super-villain',
+      { firstName: "Tom", lastName: "Dale", id: "1",
```
fivetanley: can we clean up the formatting here a bit? it'd be preferable for this to span multiple lines:

```js
{
  firstName: "tom",
  lastName: "Dale",
  id: "1",
}
```

PR #3885, opened 2015-10-24
fix model relationship name generation

gabrielgrant:
It is not currently possible to generate a model with a relationship named differently from the model's name ([despite the help indicating it should be](https://github.com/emberjs/data/blob/addonified/blueprints/model/HELP.md))

This patch brings the code's behavior in line with the docs.

I've also filed ember-cli/ember-cli#4986 as a hold-over until ember-cli/ember-cli#4909 lands

(not sure how closely you monitor ED PRs, so cc: @stefanpenner )

### DIFF
--- a/blueprints/model/index.js
+++ b/blueprints/model/index.js
@@ -19,25 +19,40 @@ module.exports = {
 
     for (var name in entityOptions) {
       var type = entityOptions[name] || '';
+      var foreignModel = name;
+      if (type.indexOf(':') > -1) {
+        foreignModel = type.split(':')[1];
+        type = type.split(':')[0];
+      }
       var dasherizedName = stringUtils.dasherize(name);
       var dasherizedNameSingular = inflection.singularize(dasherizedName);
       var camelizedName = stringUtils.camelize(name);
       var dasherizedType = stringUtils.dasherize(type);
+      var dasherizedForeignModel = stringUtils.dasherize(foreignModel);
+      var dasherizedForeignModelSingular = inflection.singularize(dasherizedForeignModel);
 
       if (/has-many/.test(dasherizedType)) {
         var camelizedNamePlural = inflection.pluralize(camelizedName);
-        attrs.push(camelizedNamePlural + ': ' + dsAttr(dasherizedName, dasherizedType));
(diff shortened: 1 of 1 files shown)


igorT:
Not sure what you mean, can you please explain and/or write a test?

gabrielgrant:
To clarify, this is a bug in the new ember-cli blueprints that @stefanpenner is bringing into ED from the ember-cli repo (hence CCing him and opening the PR against [this branch](https://github.com/emberjs/data/tree/addonified))

[The docs for the model blueprint](https://github.com/emberjs/data/blob/addonified/blueprints/model/HELP.md) state `ember generate model taco filling:belongs-to:protein toppings:has-many:toppings name:string price:number misc` should result in the following model:
 
```
import DS from 'ember-data';
  filling: DS.belongsTo('protein'),
  toppings: DS.hasMany('topping'),
  name: DS.attr('string'),
  price: DS.attr('number'),
  misc: DS.attr()
});
```

But instead, this is what that command actually generates:

```
import DS from 'ember-data';

  filling: DS.belongsTo('filling'),  // <------ should be 'protein'
  toppings: DS.hasMany('topping'),
  name: DS.attr('string'),
  price: DS.attr('number'),
  misc: DS.attr()
});
```

That is, the name of the model gets ignored, and instead just pulled from the name of the field. This patch fixes the issue in the blueprint, but is dependent on ember-cli/ember-cli#4986 to fix ember-cli's argument parsing code ([it currently just throws out everything after the second `:`](https://github.com/ember-cli/ember-cli/pull/4986/files#diff-8690f2d0f74746d76ea9fc51ecfc0ab6L8))

Sorry for lack of tests; I don't believe ember-cli has any testing infrastructure for it's blueprint generation code, and putting that in place should probably be a separate PR.

stefanpenner:
> (not sure how closely you monitor ED PRs, so cc: @stefanpenner )

thanks for the CC, reviewing now.

stefanpenner:
looks good, we should add tests on the ember-cli side: https://github.com/ember-cli/ember-cli/pull/4986#issuecomment-151158330

We are still working to extract common tests for this PR (@trabus has been doing some good job here, maybe he has some insight)

Issue #3881, opened 2015-10-23
Inflector rules


fsmanuel:
@corrspt reported https://github.com/stefanpenner/ember-inflector/issues/91

I did some digging and found an inconsistency in the way we use `pluralize`.
Sometimes we use `Ember.String.pluralize` ([here](https://github.com/emberjs/data/blob/master/packages/ember-data/lib/adapters/build-url-mixin.js#L282) and [here](https://github.com/emberjs/data/blob/master/packages/ember-data/lib/adapters/json-api-adapter.js#L111)) and in the serializer we import from `ember-inflector` ([here](https://github.com/emberjs/data/blob/master/packages/ember-data/lib/serializers/json-api-serializer.js#L7)).

It seems that they use different instances. `Ember.String.pluralize` doesn't use the new inflection rules.

related #3802, #3730

fsmanuel:
as i posted in slack:

```js
// inside an app:
console.log(Ember.Inflector.inflector);
// => ember$inflector$lib$lib$system$inflector$$Inflector

// inside ember data
console.log(Ember.Inflector.inflector);
// => Inflector
```

fivetanley:
fixed by https://github.com/emberjs/data/pull/3888

PR #3887, opened 2015-10-26
[BUGFIX release] update ember-inflector to resolve default Inflector …

fivetanley:
…issues

This should fix issues with handlebars helpers, see
https://github.com/stefanpenner/ember-inflector/pull/90

### DIFF
--- a/package.json
+++ b/package.json
@@ -56,7 +56,7 @@
     "ejs": "^1.0.0",
     "ember-cli": "^1.13.8",
     "ember-cli-dependency-checker": "^1.0.1",
-    "ember-inflector": "^1.9.2",
+    "ember-inflector": "^1.9.3",
     "ember-publisher": "0.0.7",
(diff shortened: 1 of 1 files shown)


fivetanley:
I think this should fix #3881

fsmanuel:
@fivetanley :+1: will test it asap

fsmanuel:
@fivetanley i looked what you did in the PR https://github.com/stefanpenner/ember-inflector/pull/90/files and I think ember-data should do the [same](https://github.com/emberjs/data/blob/master/packages/ember-data/lib/serializers/json-api-serializer.js#L7).

we should change:
```js
import { pluralize, singularize } from 'ember-inflector/lib/system/string';
// to
import { pluralize, singularize } from 'ember-inflector';
```

fivetanley:
confirm. would you mind making a PR for that?

fsmanuel:
jup

PR #3888, opened 2015-10-26
[BUGFIX release] import from ember-inflector to use the default instance

fsmanuel:
Instead of importing from the path ember data should import from `ember-inflector` to use the default instance.

related issues: #3887, #3881

### DIFF
--- a/packages/ember-data/lib/serializers/json-api-serializer.js
+++ b/packages/ember-data/lib/serializers/json-api-serializer.js
@@ -4,7 +4,7 @@
 
 import JSONSerializer from 'ember-data/serializers/json-serializer';
 import normalizeModelName from 'ember-data/system/normalize-model-name';
-import { pluralize, singularize } from 'ember-inflector/lib/system/string';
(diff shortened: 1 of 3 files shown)


fivetanley:
Can you amend the commit and PR title to be prefixed with `[BUGFIX release]`? Otherwise looks good to me.

fsmanuel:
@fivetanley done

fivetanley:
:tada:

joostdevries:
We're actually still having problems with this. I've locally applied this patch to 1.13.14 but I still have two inflector instances. This is probably caused by the fact that ember-cli-mirage also requires ember-inflector. Is there a reason ember-data bundles the inflector?

fivetanley:
Hey @joostdevries sorry for the late reply. Ember Data no longer bundles Ember inflector since 2.3 if you are using it as an ember cli addon. This change won't be backported to 1.x, sadly, as it will take a lot of effort. I backported the changes in this issue/PR to 1.13.16, let me know if that helps. https://github.com/emberjs/data/pull/4143

PR #3889, opened 2015-10-26
 [perf] minor adjustments to `store#push` flow

trentmwillis:
Some simple suggestions of refactoring to squeeze out a little extra performance when calling `store#push`. In our application this shows ~50ms improvement for slower devices on medium-sized payloads. Changes are as follows:

- Don't call `coerceId` in `_load` since it gets called in `_internalModelForId`
- Use `Ember.isArray` instead of `Ember.typeOf`
- Use `for-loop`s instead of `Array.prototype` methods

### DIFF
--- a/packages/ember-data/lib/system/store.js
+++ b/packages/ember-data/lib/system/store.js
@@ -53,6 +53,7 @@ export let badIdFormatAssertion = '`id` has to be non-empty string or number';
 
 var Backburner = Ember._Backburner || Ember.Backburner || Ember.__loader.require('backburner')['default'] || Ember.__loader.require('backburner')['Backburner'];
 var Map = Ember.Map;
+var isArray = Array.isArray || Ember.isArray;
 
 //Shim Backburner.join
 if (!Backburner.prototype.join) {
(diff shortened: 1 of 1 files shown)


stefanpenner:
We should prefer Array.isArray over Ember.isArray as the later is merely an alias for the former and exists for compar

packages/ember-data/lib/system/store.js
```diff
@@ -1672,13 +1672,21 @@ Store = Service.extend({
...
+      for (i = 0, l = included.length; i < l; i++) {
+        this._pushInternalModel(included[i]);
+      }
     }
 
-    if (Ember.typeOf(data.data) === 'array') {
-      var internalModels = data.data.map((recordData) => this._pushInternalModel(recordData));
-      return internalModels.map((internalModel) => internalModel.getRecord());
+    if (isArray(data.data)) {
+      var internalModels = [];
```
stefanpenner: new Array(length)

stefanpenner:
Two comments otherwise this Lgtm

trentmwillis:
@stefanpenner updated, looks like AppVeyor had issues with installing node modules though

stefanpenner:
@trentmwillis silly appveyor...

anyways thanks this looks good to me.  Its really sad that `forEach` & `map` have such a silly high cost..

PR #3891, opened 2015-10-27
Don't use Ember.create by default

knownasilya:
Fall back to Ember.create if Object.create is missing.

Fixes the deprecation coming back..

### DIFF
--- a/packages/ember-data/lib/system/empty-object.js
+++ b/packages/ember-data/lib/system/empty-object.js
@@ -3,7 +3,8 @@
 // when you're treating the object instances as arbitrary dictionaries
 // and don't want your keys colliding with build-in methods on the
 // default object prototype.
-var proto = Ember.create(null, {
+var create = Object.create || Ember.create;
+var proto = create(null, {
   // without this, we will always still end up with (new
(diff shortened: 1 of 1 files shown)


bmac:
Thanks @knownasilya

knownasilya:
Glad I could help.

pangratz:
Hmm, I thought `Ember.create` only shims the `create` functionality if it doesn't exist and uses `Object.create` if it's available? See https://github.com/emberjs/ember.js/blob/release-1-13/packages/ember-metal/lib/platform/create.js

With which version of Ember does which deprecation warning occur?

rwjblue:
@pangratz  -   You are correct that the implementation of `Ember.create` was essentially a polyfil, but using `Ember.create` in Ember 1.13 will result in a deprecation (and it was removed in Ember 2.0).

bmac:
For future reference the deprecation warning gets added here: https://github.com/emberjs/ember.js/blob/release-1-13/packages/ember-metal/lib/main.js#L423

pangratz:
Ah, thanks for the clarification!

PR #3907, opened 2015-11-05
Addonize

fivetanley:
todo:

- [x] get booting as an addon
- [x] fix tests and their import paths
- [ ] rewrite the build for the globals file to use Babel and the defeautureify replacement by @mmun
- ???
- profit


fsmanuel:
@fivetanley do you need any help or feedback on that? :+1:

fivetanley:
@fsmanuel yeah that's what peerDependency does

stefanpenner:
@fivetanley it doesn't in v3

fivetanley:
as they should...?

stefanpenner:
v2 auto-installed, v3 does not.  Just making sure this was known

Issue #3602, opened 2015-07-28
Object-level errors are ignored


dustinfarris:
There should be support for object-level validation errors if pointer is `/data`.

In ember-django-adapter we want to support a response from our backend which may look like:

```json
{"non_field_errors": ["Your first name and last name cannot begin with the same letter"]}
```

This is for errors that involve multiple fields, or for non-field-level errors.

[The adapter translates this](https://github.com/dustinfarris/ember-django-adapter/blob/master/addon/adapters/drf.js#L120) into JSON API compliant objects for DS.InvalidError, for example:

```js
{
  source: { pointer: "/data" },
  detail: "Your first name and last name cannot begin with the same letter",
  title: "Validation Error"
}
```

Our use of `/data` comes from the JSON API [error object spec](http://jsonapi.org/format/#error-objects) example that reads:

> pointer: a JSON Pointer [RFC6901] to the associated entity in the request document [e.g. __"/data" for a primary data object__, or ...

Current ED implementation does not seem to do anything with this.

In #3524, @tchak mentions the possibility of a baseErrorKey.  Would that work for /data pointers as well?  Would this be the usage?:

```hbs
{{#each post.errors.base as |error|}}
{{error.message}}
{{/each}}
```

adam-knights:
@bmac @tchak Any update on this? Is https://github.com/emberjs/data/pull/3859 expected to resolve this or would this need to be a separate fix? If I return anything from my api with the pointer '/data' it is lost by the time I interact with `record.errors`.

PR #3909, opened 2015-11-05
[BUGFIX beta] Correctly handle object level errors in json api

adam-knights:
Should fix https://github.com/emberjs/data/issues/3602. As per the JSON API spec object level errors should be placed in an object with pointer `/data`.

I have chosen to put object level data into errors.base, but this could be anything. If anyone is already using base in the form `pointer: '/data/attributes/base` then the object level ones are combined with this.

This is my first ever open source pull request so I fully expect to have missed other things I need to do or improve the code's style.

### DIFF
--- a/packages/ember-data/lib/adapters/errors.js
+++ b/packages/ember-data/lib/adapters/errors.js
@@ -1,6 +1,8 @@
 const EmberError = Ember.Error;
 
 const SOURCE_POINTER_REGEXP = /^\/?data\/(attributes|relationships)\/(.*)/;
+const SOURCE_POINTER_PRIMARY_REGEXP = /^\/?data/;
+const PRIMARY_ATTRIBUTE_KEY = 'base';
 
 /**
   @class AdapterError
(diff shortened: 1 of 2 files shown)


adam-knights:
@bmac any chance of a review?

bmac:
Looks good to me. @tchak do you have time to take a look at this pr?

tchak:
@bmac works for me. I know @igorT wanted the base key to be something like `_base` or `__base`... I would like it to be configurable. I am not sure @igorT agree.

packages/ember-data/lib/adapters/errors.js
```diff
@@ -115,11 +117,17 @@ export function errorsHashToArray(errors) {
     Object.keys(errors).forEach((key) => {
       let messages = Ember.makeArray(errors[key]);
       for (let i = 0; i < messages.length; i++) {
+        let title = 'Invalid Attribute';
+        let pointer = `/data/attributes/${key}`;
+        if (key === PRIMARY_ATTRIBUTE_KEY) {
+          title = 'Invalid Object';
```
tchak: "Invalid Document" seems more inline with JSON-API naming.

tchak:
@adam-knights Thank's a lot for you PR. It is very welcome and needed!

adam-knights:
Thanks @tchak I changed the title to Invalid Document as per your comment. _base would suggest its private when this feels like something people should be happy accessing. Configurable sounds good. I could do that in this PR (or a second one after this is merged, leaving base as default) if you could point me toward something else in a file in the codebase that is configurable so I could follow a similar pattern that would get me going.

tchak:
@bmac I am :+1: on this

bmac:
Thanks @adam-knights

PR #3912, opened 2015-11-10
Use public API for container/registry when possible.

rwjblue:
Ember 2.3 will include `Ember.getOwner` and `Ember.setOwner` as public API's to access various registry and container features. As of Ember 2.3, using `this.container` in any instance looked up from the container will issue a deprecation (instructing the user to use `Ember.getOwner(this)` instead).

This PR fixes all deprecations against ember#canary channel by adding a simplified polyfill so that the internal code can always operate with the `Ember.getOwner` API.


bmac:
Looks good to me. :+1:

PR #3913, opened 2015-11-10
Make `RESTAdapter#ajax` public

danmcclain:
It is beneficial to use `this.ajax` within functions defined on your
adapter, since it provides a promise wrapped ajax call. I commonly
define various ajax request within adapters, to keep ajax
requests all within the Adapter layer, serving as a way to define the
interactions with the backend in a common spot

### DIFF
--- a/packages/ember-data/lib/adapters/rest-adapter.js
+++ b/packages/ember-data/lib/adapters/rest-adapter.js
@@ -803,7 +803,7 @@ export default Adapter.extend(BuildURLMixin, {
     * Registers success and failure handlers.
 
     @method ajax
-    @private
+    @public
     @param {String} url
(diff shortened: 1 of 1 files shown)


bmac:
Hi @danmcclain. I'll bring this up in the Ember Data team meeting. Historically I think the Ember Data team has been reluctant to make this public because we do not want to couple public methods to the jQuery api.

bmac:
I brought up this issue in the weekly meeting and the consensus was to hold off on making this public for now while we wait to see how Ember's plan for a network service evolves https://github.com/emberjs/rfcs/pull/94.

fivetanley:
Closing until the rfc evolves, sorry. I don't think we want the current implementation public.

PR #3915, opened 2015-11-10
Add null implementation of findMany in DS.Adapter

jjbohn:
Since the other primary methods for override (findRecord, createRecord,
findAll, etc) in the abstract class have null implementations, it makes
sense to me to have this for findMany as well.


bmac:
Thanks @jjbohn

jjbohn:
Anytime

PR #3916, opened 2015-11-10
`Store._find` asserts `adapterPayload` not empty

seanpdoyle:
Currently, [Store._find][find] only asserts that the `adapterPayload` is
non-null (i.e. `{}` is a valid payload).

This commit forces the `adapterPayload` to be non-empty.

When TDDing with Mirage, `adapterPayload` is often `{}`, which sneaks
through the asserts and ends up as a non-descriptive:

```
Assertion Failed: You must include an 'id' for undefined in an object
passed to 'push'
```

Which is triggered in [Store._pushInternalModel][_pushInternalModel],
which is a few method invocations away from the source of the issue.

[_pushInternalModel]: https://github.com/emberjs/data/blob/f1ccad8ab9356dd51f44d63585c27d8bb4ec9c3a/packages/ember-data/lib/system/store.js#L1709
[find]: https://github.com/emberjs/data/blob/f1ccad8ab9356dd51f44d63585c27d8bb4ec9c3a/packages/ember-data/lib/system/store/finders.js#L27

### DIFF
--- a/addon/system/store/finders.js
+++ b/addon/system/store/finders.js
@@ -14,6 +14,14 @@ import {
 
 var Promise = Ember.RSVP.Promise;
 
+function payloadIsNotBlank(adapterPayload) {
+  if (Ember.isArray(adapterPayload)) {
+    return true;
+  } else {
+    return Object.keys(adapterPayload || {}).length;
+  }
+}
+
 export function _find(adapter, store, typeClass, id, internalModel, options) {
   var snapshot = internalModel.createSnapshot(options);
   var promise = adapter.findRecord(store, typeClass, id, snapshot);
@@ -24,7 +32,7 @@ export function _find(adapter, store, typeClass, id, internalModel, options) {
   promise = _guard(promise, _bind(_objectIsAlive, store));
 
   return promise.then(function(adapterPayload) {
-    Ember.assert("You made a request for a " + typeClass.typeClassKey + " with id " + id + ", but the adapter's response did not have any data", adapterPayload);
+    Ember.assert("You made a `find` request for a " + typeClass.typeClassKey + " with id " + id + ", but the adapter's response did not have any data", payloadIsNotBlank(adapterPayload));
     return store._adapterRun(function() {
       var payload = normalizeResponseHelper(serializer, store, typeClass, adapterPayload, id, 'findRecord');
       //TODO Optimize
@@ -56,6 +64,7 @@ export function _findMany(adapter, store, typeClass, ids, internalModels) {
   promise = _guard(promise, _bind(_objectIsAlive, store));
 
   return promise.then(function(adapterPayload) {
+    Ember.assert("You made a `findMany` request for a " + typeClass.typeClassKey + " with ids " + ids + ", but the adapter's response did not have any data", payloadIsNotBlank(adapterPayload));
     return store._adapterRun(function() {
       var payload = normalizeResponseHelper(serializer, store, typeClass, adapterPayload, null, 'findMany');
       //TODO Optimize, no need to materialize here
@@ -77,6 +86,7 @@ export function _findHasMany(adapter, store, internalModel, link, relationship) 
   promise = _guard(promise, _bind(_objectIsAlive, internalModel));
 
   return promise.then(function(adapterPayload) {
+    Ember.assert("You made a `findHasMany` request for a " + internalModel.modelName + "'s `" + relationship.key + "` relationship, using link " + link + ", but the adapter's response did not have any data", payloadIsNotBlank(adapterPayload));
     return store._adapterRun(function() {
       var payload = normalizeResponseHelper(serializer, store, typeClass, adapterPayload, null, 'findHasMany');
       //TODO Use a non record creating push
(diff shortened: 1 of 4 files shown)


packages/ember-data/lib/system/store/finders.js
```diff
@@ -77,6 +82,7 @@ export function _findHasMany(adapter, store, internalModel, link, relationship)
   promise = _guard(promise, _bind(_objectIsAlive, internalModel));
 
   return promise.then(function(adapterPayload) {
+    Ember.assert("You made a `findHasMany` request for a " + internalModel.modelName + "'s `" + relationship.key + "` relationship, using link " + link + ", but the adapter's response did not have any data", payloadIsNotBlank(adapterPayload));
```
seanpdoyle: > that could use a better assertion. What adapter are you using?
> --@bmac

What would you suggest?

bmac:
These assertions look great @seanpdoyle. Do you mind adding tests to make sure the assertions are called when the payload is `{}`? Let me know if you need help doing figuring out where the test goes.

seanpdoyle:
@bmac I think I'll need some guidance.

I'm surprised that these assertions caused test failures to begin with.

addon/system/store/finders.js
```diff
@@ -56,6 +60,7 @@ export function _findMany(adapter, store, typeClass, ids, internalModels) {
   promise = _guard(promise, _bind(_objectIsAlive, store));
 
   return promise.then(function(adapterPayload) {
+    Ember.assert("You made a `findMany` request for a " + typeClass.typeClassKey + " with ids " + ids + ", but the adapter's response did not have any data", payloadIsNotBlank(adapterPayload));
```
seanpdoyle: @bmac how can I exercise this assertion?

addon/system/store/finders.js
```diff
@@ -77,6 +82,7 @@ export function _findHasMany(adapter, store, internalModel, link, relationship)
   promise = _guard(promise, _bind(_objectIsAlive, internalModel));
 
   return promise.then(function(adapterPayload) {
+    Ember.assert("You made a `findHasMany` request for a " + internalModel.modelName + "'s `" + relationship.key + "` relationship, using link " + link + ", but the adapter's response did not have any data", payloadIsNotBlank(adapterPayload));
```
seanpdoyle: @bmac I expected this payload would look like:

```js
{ comments: [] }
```

which would make 

```js
{}
```

invalid (and what we're asserting against.

However, [this test](https://github.com/emberjs/data/blob/1d5dfa142523a4aefdde182262923a42ccb56286/tests/integration/relationships/has-many-test.js#L347) mocks the response as `[]`.

Why isn't it `{ comments: [] }`? What should this assertion be checking for?

addon/system/store/finders.js
bmac: @seanpdoyle I would follow the pattern used in this test of stubbing the adapter method to return a value that triggers the assertion then using the `expectAssertion` helper to test that it is called. https://github.com/emberjs/data/blob/1d5dfa142523a4aefdde182262923a42ccb56286/tests/integration/adapter/queries-test.js#L47-L63

seanpdoyle: @bmac unfortunately, there isn't a `store` operation that maps directly to this one like the others.

For instance, `adapter.findRecord` maps to `store.find`, `adapter.query` maps to `store.query` (and maybe `store.queryRecord`?).

In this case, isn't `findMany` used to coalesce find requests?

Would I be able to exercise `adapter.findMany` with a `store.query({ ids: [1,2,3] })`?

seanpdoyle: @bmac I've tried the following:

```js
adapter.findMany = () => Ember.RSVP.resolve({});

assert.expectAssertion(() => {
  run(() => store.query('person', { ids: [1, 2] }));
}, /the adapter's response did not have any data/);
```

But it didn't raise an error.

seanpdoyle: @bmac which `store` operation maps to a `findMany`?

bmac: I believe you can trigger it by setting `coalesceFindRequests: true` on the adapter then making 2 `findRecord` requests in the same run loop.

```js
assert.expectAssertion(() => {
  run(() => {
    store.findRecord('person', 1);
    store.findRecord('person', 2);
  });
}, /the adapter's response did not have any data/);
```

tests/integration/adapter/find-test.js
```diff
@@ -137,3 +137,27 @@ test("When a single record is requested, and the promise is rejected, the record
...
+  env.registry.register('adapter:person', DS.Adapter.extend({
+    coalesceFindRequests: true,
+    findMany: () => Ember.RSVP.resolve({})
+  }));
+
+  assert.expectAssertion(() => {
+    run(() => {
+      store.findRecord('person', '1');
+      store.findRecord('person', '2');
+    });
```
seanpdoyle: @bmac :+1: that did the trick.

addon/system/store/finders.js
```diff
@@ -77,6 +86,7 @@ export function _findHasMany(adapter, store, internalModel, link, relationship)
...
```
seanpdoyle: @bmac how can I exercise this assertion?

seanpdoyle:
@bmac what's stopping this from being merged?

Is there anything else I can do?

bmac:
Sorry, @seanpdoyle I've been very busy recently and hadn't had a chance to review this pr. Looking at it now.

bmac:
Thanks @seanpdoyle. Sorry for the delay.

PR #3927, opened 2015-11-13
restore globals publishing to canary build

fivetanley:
The globals only build in production mode, so we change travis to build
that after success.

fixes components/ember-data#31


rwjblue:
:heart:

PR #3946, opened 2015-11-20
Report better error when `type` is missing from a JSONApi response

### LINKED ISSUE #3945: Better error for when the `type` property is missing from a JSON API response.
Currently it will just throw `Uncaught TypeError: Cannot read property 'replace' of undefined`, because somewhere along the parsing of the response replace is used on the value of `type`.

The error is triggered at: https://github.com/emberjs/data/blob/master/addon/serializers/json-api-serializer.js#L139

martndemus:
Fixes #3945

### DIFF
--- a/addon/serializers/json-api-serializer.js
+++ b/addon/serializers/json-api-serializer.js
@@ -136,6 +136,10 @@ const JSONAPISerializer = JSONSerializer.extend({
     @private
   */
   _normalizeResourceHelper: function(resourceHash) {
+    Ember.assert(this.warnMessageForUndefinedType(), !Ember.isNone(resourceHash.type), {
+      id: 'ds.serializer.type-is-undefined'
+    });
+
     let modelName = this.modelNameFromPayloadKey(resourceHash.type);
 
     if (!this.store._hasModelFor(modelName)) {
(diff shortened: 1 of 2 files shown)


addon/serializers/json-api-serializer.js
```diff
@@ -465,6 +472,9 @@ const JSONAPISerializer = JSONSerializer.extend({
 
 Ember.runInDebug(function() {
   JSONAPISerializer.reopen({
+    warnMessageForUndefinedType: function() {
+      return 'Encountered a resource object with an undefined type';
```
wecc: Can we expand this error message and refer to the current serializer? Using `this.constructor.toString()` like in `warnMessageNoModelForType`

martndemus: Good point! I have added it to the patch.

addon/serializers/json-api-serializer.js
```diff
@@ -136,6 +136,13 @@ const JSONAPISerializer = JSONSerializer.extend({
     @private
   */
   _normalizeResourceHelper: function(resourceHash) {
+    if (Ember.isNone(resourceHash.type)) {
+      Ember.warn(this.warnMessageForUndefinedType(), false, {
```
bmac: I wonder if Ember Data should be asserting here instead of a warning. @martndemus do you have a use case where you expect your type to be undefined?

martndemus: Nope, I don't, I agree with you. I'll update the code!

addon/serializers/json-api-serializer.js
```diff
@@ -136,6 +136,12 @@ const JSONAPISerializer = JSONSerializer.extend({
     @private
   */
   _normalizeResourceHelper: function(resourceHash) {
+    if (Ember.isNone(resourceHash.type)) {
+      Ember.assert(this.warnMessageForUndefinedType(), false, {
```
bmac: Do you mind changing the `false` here to `!Ember.isNone(resourceHash.type)` and remove the if statement that way when the production build strips the `assert` there won't be an empty if statement.

bmac:
Thanks @martndemus

martndemus:
:+1:

PR #3958, opened 2015-11-28
[DOCS fix] fix incorrect argument description...

acburdine:
...in `store.queryRecord`

Docs incorrectly say that passed in modelName can be a subclass of DS.Model. That behavior has been removed.


bmac:
Thanks @acburdine

PR #3965, opened 2015-12-04
[Doc] Specify the status code expected to populate the error object

brianstorti:
For what I see, the `error` object is populated just when the status code is specifically `422`, and not for every code in the 400 family. I think it's worth having this information in the docs (as I spent some time trying to figure out why this was not working the way I was expecting).


bmac:
Thanks @brianstorti

PR #3973, opened 2015-12-07
[CLEANUP] use debug helpers from ember-data/debug

pangratz:
By this, the calls to `assert`, `warn` etcetera are corectly stripped
when a new production build is created.

cc @fivetanley

### DIFF
--- a/addon/index.js
+++ b/addon/index.js
@@ -1,4 +1,5 @@
 import Ember from "ember";
+import { warn } from "ember-data/debug";
 /**
   Ember Data
   @module ember-data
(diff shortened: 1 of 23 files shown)


fivetanley:
LGTM, although I don't think this actually removes them quite yet because we don't have any broccoli modules for that I think.

fivetanley:
Appveyor is failing due to some syntax errors.

pangratz:
> although I don't think this actually removes them quite yet because we don't have any broccoli modules for that I think.

Hmm, I am not sure. Doing a `ember build --environment=production` and searching for `assert` shows that `dist/globals/ember-data.js` contains the calls, whereas `dist/globals/ember-data.prod.js` doesn't...

fivetanley:
I barely remember doing this, but here we are :p https://github.com/emberjs/data/blob/master/lib/javascripts.js#L60

pangratz:
Hooray :tada:

PR #3976, opened 2015-12-08
Allow `include` query parameter with store.findRecord & store.findAll

HeroicEric:
Related RFC: https://github.com/emberjs/rfcs/pull/99

### Description

These changes allow an `include` parameter to be specified when using
`store.findRecord` and `store.findAll`:

```javascript
store.findRecord('post', 123, { include: 'comments' });
store.findAll('post', { include: 'comments' });
```

The value for `include` is copied onto the `DS.Snapshot` or
`DS.SnapshotRecordArray` that is passed to the corresponding adapter
function. This value is then pulled from the snapshot and used as a
query parameter (`include`) when making an AJAX request via
`adapter.ajax`.

### Questions

1. ~~Is it OK that the `include` option is copied into `adapterOptions` or should it be copied into `adapterOptions.query`. I originally copied it into `adapterOptions.query` but the code was a little uglier so I went with just copying it into `adapterOptions`, but I'm still not sure.~~
1. If the value for `include` is an array, should we automatically join the values into a comma separated list? The [JSON API spec for "Inclusion of Related Resources"](http://jsonapi.org/format/#fetching-includes) requires that the value be a comma separated list but I didn't think that was necessarily what everyone would want. We could just do that in the JSON API adapter but I figured I'd ask for some feedback before proceeding with that.

### Work to be done in additional PRs

It was also [mentioned in the RFC](https://github.com/emberjs/rfcs/pull/99#issuecomment-153833942) that we could provide some additional functionality:

```javascript
post.save({ include: 'comments' });
post.destroyRecord({ include: 'comments' });
```

I'm not totally clear on what the behavior would be there yet so those were left unimplemented for now.

### DIFF
--- a/addon/-private/system/record-arrays/record-array.js
+++ b/addon/-private/system/record-arrays/record-array.js
@@ -202,8 +202,7 @@ export default Ember.ArrayProxy.extend(Ember.Evented, {
   },
 
   createSnapshot(options) {
-    var adapterOptions = options && options.adapterOptions;
-    var meta = this.get('meta');
-    return new SnapshotRecordArray(this, meta, adapterOptions);
+    const meta = this.get('meta');
+    return new SnapshotRecordArray(this, meta, options);
   }
 });
--- a/tests/integration/adapter/rest-adapter-test.js
+++ b/tests/integration/adapter/rest-adapter-test.js
@@ -4,6 +4,7 @@ import Ember from 'ember';
 import {module, test} from 'qunit';
 
 import DS from 'ember-data';
+import isEnabled from 'ember-data/-private/features';
 
 var env, store, adapter, Post, Comment, SuperUser;
 var passedUrl, passedVerb, passedHash;
@@ -56,7 +57,7 @@ test("findRecord - basic payload", function(assert) {
   run(store, 'findRecord', 'post', 1).then(assert.wait(function(post) {
     assert.equal(passedUrl, "/posts/1");
     assert.equal(passedVerb, "GET");
-    assert.equal(passedHash, undefined);
+    assert.deepEqual(passedHash.data, {});
 
     assert.equal(post.get('id'), "1");
     assert.equal(post.get('name'), "Rails is omakase");
@@ -83,7 +84,7 @@ test("find - basic payload (with legacy singular name)", function(assert) {
   run(store, 'findRecord', 'post', 1).then(assert.wait(function(post) {
     assert.equal(passedUrl, "/posts/1");
     assert.equal(passedVerb, "GET");
-    assert.equal(passedHash, undefined);
+    assert.deepEqual(passedHash.data, {});
 
     assert.equal(post.get('id'), "1");
     assert.equal(post.get('name'), "Rails is omakase");
@@ -101,7 +102,7 @@ test("findRecord - payload with sideloaded records of the same type", function(a
   run(store, 'findRecord', 'post', 1).then(assert.wait(function(post) {
     assert.equal(passedUrl, "/posts/1");
     assert.equal(passedVerb, "GET");
-    assert.equal(passedHash, undefined);
+    assert.deepEqual(passedHash.data, {});
 
     assert.equal(post.get('id'), "1");
     assert.equal(post.get('name'), "Rails is omakase");
@@ -121,7 +122,7 @@ test("findRecord - payload with sideloaded records of a different type", functio
   run(store, 'findRecord', 'post', 1).then(assert.wait(function(post) {
     assert.equal(passedUrl, "/posts/1");
     assert.equal(passedVerb, "GET");
-    assert.equal(passedHash, undefined);
+    assert.deepEqual(passedHash.data, {});
 
     assert.equal(post.get('id'), "1");
     assert.equal(post.get('name'), "Rails is omakase");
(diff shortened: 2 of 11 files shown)


bmac:
> 1. Is it OK that the include option is copied into adapterOptions or should it be copied into adapterOptions.query. I originally copied it into adapterOptions.query but the code was a little uglier so I went with just copying it into adapterOptions, but I'm still not sure.

I think copying into `adapterOptions` is the best thing to do here. 

> 2. If the value for include is an array, should we automatically join the values into a comma separated list? The JSON API spec for "Inclusion of Related Resources" requires that the value be a comma separated list but I didn't think that was necessarily what everyone would want. We could just do that in the JSON API adapter but I figured I'd ask for some feedback before proceeding with that.

I was originally against this because I think it would be more consistent if the API only accepted 1 type. However, @igorT convinced me that allowing the value to be an array could allow Ember Data to be smarter and warn if an included relationship was not defined on the model.

wecc:
I agree with @bmac

addon/adapters/rest-adapter.js
```diff
@@ -389,13 +403,19 @@ export default Adapter.extend(BuildURLMixin, {
     @return {Promise} promise
   */
   findAll(store, type, sinceToken, snapshotRecordArray) {
-    var query, url;
+    const url = this.buildURL(type.modelName, null, null, 'findAll');
+    const adapterOptions = snapshotRecordArray.adapterOptions || {};
+    const { include } = adapterOptions;
+
+    let query;
```
bmac: Do you mind refactoring some of the duplicated code in `findRecord` and `findAll` into a helper method?

addon/system/store.js
```diff
@@ -499,12 +500,16 @@ Store = Service.extend({
...
     assert('Passing classes to store methods has been removed. Please pass a dasherized string instead of '+ Ember.inspect(modelName), typeof modelName === 'string');
     assert(badIdFormatAssertion, (typeof id === 'string' && id.length > 0) || (typeof id === 'number' && !isNaN(id)));
 
-    var internalModel = this._internalModelForId(modelName, id);
-    options = options || {};
+    const internalModel = this._internalModelForId(modelName, id);
+    const { include } = options;
+
+    if (include) {
+      options.adapterOptions = merge(options.adapterOptions || {}, { include });
```
bmac: @igorT points out that we may not want to use `adapterOptions` since that is a namespace for the users to put whatever they'd like. Instead we could put possibly put `include` on the snapshot. **This is still to be decided...**

addon/adapters/rest-adapter.js
HeroicEric: Sure

tests/unit/adapters/rest-adapter/ajax-test.js
```diff
@@ -35,8 +35,8 @@ test("When an id is searched, the correct url should be generated", function(ass
     return Ember.RSVP.resolve();
   };
   run(function() {
-    adapter.findRecord(store, Person, 1);
-    adapter.findRecord(store, Place, 1);
+    adapter.findRecord(store, Person, 1, {});
+    adapter.findRecord(store, Place, 1, {});
```
bmac: Is the 3rd argument needed here?

bmac: Nevermind, I see now that this is the adapter method.

addon/system/store.js
bmac: We talked about this in today's meeting and agreed to move `include` to snapshot instead of the `adapterOptions`. That way `adapterOptions` is reserved for users to do anything they'd like in it without Ember Data polluting it with its own options.

HeroicEric: cool :) I should be able to work on this later tonight

bmac:
@HeroicEric Do you mind wrapping all your changes in [feature flags](https://github.com/emberjs/data/pull/3979#issue-121089686)?

HeroicEric:
@bmac I tried wrapping the changes in a feature flag but I'm not sure I did all the steps correctly. The feature flagged tests don't seem to run with or without the `--environment=test-optional-features` flag or with the "Enable Opt Features" boxed checked.

tests/integration/adapter/rest-adapter-test.js
```diff
@@ -56,7 +57,7 @@ test("find - basic payload", function(assert) {
   run(store, 'find', 'post', 1).then(assert.wait(function(post) {
     assert.equal(passedUrl, "/posts/1");
     assert.equal(passedVerb, "GET");
-    assert.equal(passedHash, undefined);
+    assert.deepEqual(passedHash.data, {});
```
HeroicEric: Would it be preferred to continue returning `undefined` rather than `{}` in these cases? `adapter.ajax` seems to act the same with `{ data: {} }` and `{}`.

addon/-private/system/record-arrays/record-array.js
```diff
@@ -201,9 +202,16 @@ export default Ember.ArrayProxy.extend(Ember.Evented, {
...
 
-  createSnapshot(options) {
-    var adapterOptions = options && options.adapterOptions;
-    var meta = this.get('meta');
-    return new SnapshotRecordArray(this, meta, adapterOptions);
+  createSnapshot(options = {}) {
+    const adapterOptions = options.adapterOptions;
+    const meta = this.get('meta');
+
+    const snapshot = new SnapshotRecordArray(this, meta, adapterOptions);
```
HeroicEric: it might be nice if the two snapshot constructors had the same support for providing `adapterOptions` and possibly other options. It seems a little weird that we set properties on it directly after we initialize it rather than using the constructor. Thoughts?

bmac: Go ahead a change it to improve the consistency. The Snapshot constructors are private.

tests/integration/adapter/rest-adapter-test.js
bmac: returning `{}` seems ok to me.

HeroicEric:
Rebased, updated the description, added the feature flag to `FEATURES.md`. Anything else I can do here? Other than start writing some docs :smile:

bmac:
:tada:

Issue #3922, opened 2015-11-12
Restore feature flags / code stripping in both addon/globals build


fivetanley:

HeroicEric:
Was this solved via https://github.com/emberjs/data/pull/3979?

pangratz:
Not yet as it seems that the canary builds have all the features stripped. I am investigating...

pangratz:
Hmm, looks like I had a browser caching issue. I think this is good to be closed!

PR #3979, opened 2015-12-08
Re-enable feature flags

pangratz:
This re-enables testing feature flags and stripping disabled feature flags from production build (using [ember-cli/babel-plugin-feature-flags](https://github.com/ember-cli/babel-plugin-feature-flags)).

Feature flags are now possible within ember-data by using the following setup:

```js
import isEnabled from 'ember-data/features';

if (isEnabled("ds-my-feature")) {
  // ...
}
```

The features need to be defined within `config/features.json`. Disabled feature flags are also stripped from production builds.

---

Running the tests with all optional feature flags enabled is possible via `ember test --environment=test-optional-features`. When the tests are run via `ember serve`, the optional feature flags can be enabled via a checkbox.

---

This addresses #3922.

### DIFF
--- a/lib/enable-optional-features-via-url/index.js
+++ b/lib/enable-optional-features-via-url/index.js
@@ -1,0 +1,27 @@
+/*jshint node:true*/
+module.exports = {
+  name: 'enable-optional-features-via-url',
+
+  /**
+    So the ENABLE_OPTIONAL_FEATURES flag is considered correctly within the
+    index.html, it needs to be set before Ember.js is loaded. Since there is
+    currently no way to access the `window` object within config/environment.js
+    (and hereby check if there is a query parameter present for the checkbox),
+    a script is injected, before Ember.js is loaded. The script checks if there
+    is no value yet for the ENABLE_OPTIONAL_FEATURES flag, and if so, it sets
+    the flag to true when there is a `enableoptionalfeatures` query parameter.
+   */
+  contentFor: function(name) {
+    if (name === "vendor-prefix") {
+      var array = [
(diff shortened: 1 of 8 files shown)


rwjblue:
Nice work here @pangratz!

lib/enable-optional-features-via-url/index.js
```diff
@@ -0,0 +1,27 @@
...
+        "<!-- injected by lib/enable-optional-features-via-url -->",
+        "window.EmberENV = window.EmberENV || {};",
+        "if (typeof window.EmberENV.ENABLE_OPTIONAL_FEATURES === 'undefined') {",
+        "  window.EmberENV.ENABLE_OPTIONAL_FEATURES = window.location.search.indexOf('enableoptionalfeatures') !== -1;",
+        "}"
+      ];
+
+      return array.join('\n');
+    }
+  }
```
pangratz: @rwjblue do you, as an ember-cli resident, have a nicer solution than this? Basically `EmberENV` needs to be modified after `QUnit` is loaded (so we have access to `QUnit.urlconfig` and before `ember` is loaded (so the updated, latest version of `EmberENV` is considered).

bmac:
:+1: You are my hero @pangratz \o/

lib/enable-optional-features-via-url/index.js
fivetanley: If this is javascript we should probably use JS comments instead of HTML

pangratz: :+1: absolutely

PR #3980, opened 2015-12-08
Add license field to bower.json

nickiaconis:
From components/ember-data#33.

bower.json is missing the license field. If I'm not mistaken, Data uses the same license as Ember (MIT). https://github.com/components/ember/blob/master/bower.json#L3

Our automated external library requests service is blocking the import of Ember Data canary for this reason.

I've also positioned the license field in component.json to mirror that of Ember.


fivetanley:
:pray: thank you!

PR #3983, opened 2015-12-09
Fix documentation to use "serializedHasManyName"

monofone:
The variable was declared but never used in the documentation


wecc:
Well spotted, thanks! Could you prefix the commit with `[DOC beta]`?

monofone:
@wecc ahm, well there seems to be some problem with the AppVeyor configuration. As I have no idea if I can fix there something I will discard this from my mind and leave it to you to clear things out.

Happy contributing.

wecc:
Yeah, AppVeyor does that from time to time :)

PR #3985, opened 2015-12-10
Remove extraneous conditional in ajax()

davewasmer:
_I might be missing something obvious, so feel free to close if this code was intended and I just missed the effect._

The conditional was checking if `error` was not an instance of `Error`, but it was initialized on the previous line, ensuring the conditional would always pass.


fivetanley:
thanks!

davewasmer:
:beers:

PR #3986, opened 2015-12-11
Move private modules into the `-private` directory

bmac:
to discourage importing from paths that may change in future versions

cc @fivetanley @stefanpenner @rwjblue 

This pr moves everything but `index.js` into the `-private` directory. We can make individual prs to move stable modules out of private directory when we decide they are stable.


rwjblue:
:+1:

addon/-private/adapters/rest-adapter.js
```diff
@@ -3,18 +3,18 @@
 */
 
 import Ember from 'ember';
-import Adapter from "ember-data/system/adapter";
+import Adapter from "ember-data/-private/system/adapter";
```
fivetanley: The RESTAdapter is private API though right?

bmac: Right now in Ember Data 2.2.0 the public api to access any Ember Data module is: 

```js
import DS from 'ember-data';

// then do something with `DS.RESTAdapter`
```

Instead of trying to make a judgement call about what we should allow users to import and what we should keep hidden I decided to move everything into the private directory. After this is merged we can open new prs to move things like the `RESTAdapter` to a public path. That way we can be more deliberate and discerning about what gets exposed publicly.

rwjblue: Another way of putting this: to date Ember Data's public API has been strictly the globals output (which seems unchanged by this PR). This PR helps ensure that we are not inadvertently making many more of our internals "public" because the modules are importable and not marked as private.

PR #3988, opened 2015-12-11
[PERF] Don't use array methods

chadhietala:
We shouldn't use array methods as they allocate and close over items that need to be GC'd. Memory preasure can be an issue and we should do everything to reduce that preasure.

### DIFF
--- a/addon/-private/serializers/embedded-records-mixin.js
+++ b/addon/-private/serializers/embedded-records-mixin.js
@@ -348,11 +348,17 @@ export default Ember.Mixin.create({
   */
   _generateSerializedHasMany(snapshot, relationship) {
     let hasMany = snapshot.hasMany(relationship.key);
-    return Ember.A(hasMany).map((embeddedSnapshot) => {
-      var embeddedJson = embeddedSnapshot.record.serialize({ includeId: true });
+    let manyArray = Ember.A(hasMany);
+    let ret = new Array(manyArray.length);
+
+    for (let i = 0, l = manyArray.length; i < l; i++) {
+      let embeddedSnapshot = manyArray[i];
+      let embeddedJson = embeddedSnapshot.record.serialize({ includeId: true });
       this.removeEmbeddedForeignKey(snapshot, embeddedSnapshot, relationship, embeddedJson);
-      return embeddedJson;
-    });
+      ret[i] = embeddedJson;
+    }
+
+    return ret;
   },
 
   /**
@@ -450,11 +456,15 @@ export default Ember.Mixin.create({
   */
   _extractEmbeddedHasMany(store, key, hash, relationshipMeta) {
     let relationshipHash = get(hash, `data.relationships.${key}.data`);
+
     if (!relationshipHash) {
       return;
     }
 
-    let hasMany = relationshipHash.map(item => {
+    let hasMany = new Array(relationshipHash.length);
+
+    for (let i = 0, l = relationshipHash.length; i < l; i++) {
+      let item = relationshipHash[i];
       let { data, included } = this._normalizeEmbeddedRelationship(store, relationshipMeta, item);
       hash.included = hash.included || [];
       hash.included.push(data);
@@ -462,8 +472,8 @@ export default Ember.Mixin.create({
         hash.included.push(...included);
       }
 
-      return { id: data.id, type: data.type };
-    });
+      hasMany[i] = { id: data.id, type: data.type };
+    }
 
     let relationship = { data: hasMany };
     set(hash, `data.relationships.${key}`, relationship);
--- a/addon/-private/serializers/json-api-serializer.js
+++ b/addon/-private/serializers/json-api-serializer.js
@@ -108,12 +108,26 @@ const JSONAPISerializer = JSONSerializer.extend({
 
     if (Ember.typeOf(documentHash.data) === 'object') {
       documentHash.data = this._normalizeResourceHelper(documentHash.data);
-    } else if (Ember.typeOf(documentHash.data) === 'array') {
-      documentHash.data = documentHash.data.map(this._normalizeResourceHelper, this);
+    } else if (Array.isArray(documentHash.data)) {
+      let ret = new Array(documentHash.data.length);
+
+      for (let i = 0, l = documentHash.data.length; i < l; i++) {
+        let data = documentHash.data[i];
+        ret[i] = this._normalizeResourceHelper(data);
+      }
+
+      documentHash.data = ret;
     }
 
-    if (Ember.typeOf(documentHash.included) === 'array') {
-      documentHash.included = documentHash.included.map(this._normalizeResourceHelper, this);
+    if (Array.isArray(documentHash.included)) {
+      let ret = new Array(documentHash.included.length);
+
+      for (let i = 0, l = documentHash.included.length; i < l; i++) {
+        let included = documentHash.included[i];
+        ret[i] = this._normalizeResourceHelper(included);
+      }
+
+      documentHash.included = ret;
     }
 
     return documentHash;
@@ -215,8 +229,15 @@ const JSONAPISerializer = JSONSerializer.extend({
       relationshipHash.data = this._normalizeRelationshipDataHelper(relationshipHash.data);
     }
 
-    if (Ember.typeOf(relationshipHash.data) === 'array') {
-      relationshipHash.data = relationshipHash.data.map(this._normalizeRelationshipDataHelper, this);
+    if (Array.isArray(relationshipHash.data)) {
+      let ret = new Array(relationshipHash.data.length);
+
+      for (let i = 0, l = relationshipHash.data.length; i < l; i++) {
+        let data = relationshipHash.data[i];
+        ret[i] = this._normalizeRelationshipDataHelper(data);
+      }
+
+      relationshipHash.data = ret;
     }
 
     return relationshipHash;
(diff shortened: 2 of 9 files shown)


addon/-private/serializers/embedded-records-mixin.js
```diff
@@ -348,11 +348,17 @@ export default Ember.Mixin.create({
   */
   _generateSerializedHasMany(snapshot, relationship) {
     let hasMany = snapshot.hasMany(relationship.key);
-    return Ember.A(hasMany).map((embeddedSnapshot) => {
-      var embeddedJson = embeddedSnapshot.record.serialize({ includeId: true });
+    let manyArray = Ember.A(hasMany);
+    let ret = [];
```
stefanpenner: Let's pre alloc the array, looks like we know the size

addon/-private/serializers/embedded-records-mixin.js
```diff
@@ -450,20 +456,22 @@ export default Ember.Mixin.create({
...
       hash.included = hash.included || [];
       hash.included.push(data);
       if (included) {
         hash.included.push(...included);
       }
-
-      return { id: data.id, type: data.type };
-    });
+      
+      hasMany.push({ id: data.id, type: data.type });
```
stefanpenner: ^^

addon/-private/serializers/json-api-serializer.js
```diff
@@ -109,11 +109,25 @@ const JSONAPISerializer = JSONSerializer.extend({
     if (Ember.typeOf(documentHash.data) === 'object') {
       documentHash.data = this._normalizeResourceHelper(documentHash.data);
     } else if (Ember.typeOf(documentHash.data) === 'array') {
-      documentHash.data = documentHash.data.map(this._normalizeResourceHelper, this);
+      let ret = [];
```
stefanpenner: ^

addon/-private/serializers/json-api-serializer.js
```diff
@@ -109,11 +109,25 @@ const JSONAPISerializer = JSONSerializer.extend({
...
+        let data = documentHash.data[i];
+        ret.push(this._normalizeResourceHelper(data));
+      }
+
+      documentHash.data = ret;
     }
 
     if (Ember.typeOf(documentHash.included) === 'array') {
-      documentHash.included = documentHash.included.map(this._normalizeResourceHelper, this);
+      let ret = [];
```
stefanpenner: Etc...

addon/-private/serializers/embedded-records-mixin.js
chadhietala: Fixed.

addon/-private/serializers/embedded-records-mixin.js
```diff
@@ -450,11 +456,13 @@ export default Ember.Mixin.create({
   */
   _extractEmbeddedHasMany(store, key, hash, relationshipMeta) {
     let relationshipHash = get(hash, `data.relationships.${key}.data`);
+    let hasMany = new Array(relationshipHash.length);
```
stefanpenner: maybe do this after the conditional on 460

addon/-private/serializers/json-api-serializer.js
```diff
@@ -109,11 +109,25 @@ const JSONAPISerializer = JSONSerializer.extend({
...
+
+      for (let i = 0, l = documentHash.data.length; i < l; i++) {
+        let data = documentHash.data[i];
+        ret[i] = this._normalizeResourceHelper(data);
+      }
+
+      documentHash.data = ret;
     }
 
     if (Ember.typeOf(documentHash.included) === 'array') {
```
stefanpenner: i wonder if `Array.isArray(documentHash.data)` is sufficient (As it is much faster)

chadhietala:
Windows failed because its windows. I didn't realized that some of the array methods were actually Maps being based around that conforms to the iterable interface.

addon/-private/system/store.js
```diff
@@ -578,11 +578,13 @@ Store = Service.extend({
   */
   findByIds(modelName, ids) {
     assert('Passing classes to store methods has been removed. Please pass a dasherized string instead of '+ Ember.inspect(modelName), typeof modelName === 'string');
-    var store = this;
+    let promises = [];
```
stefanpenner: pre alloc

addon/-private/system/store.js
```diff
@@ -841,7 +852,13 @@ Store = Service.extend({
     @return {Promise} promise
   */
   findMany(internalModels) {
-    return Promise.all(internalModels.map((internalModel) => this._findByInternalModel(internalModel)));
+    let finds = [];
```
stefanpenner: pre alloc

topaxi:
Just curious, how much does this affect perf and gc pressure? :)

chadhietala:
TL;DR Don't put things into new generation space to allow less GC events which must stop the world to collect.

@topaxi It depends on several factors but the largest is the granularity of your models and relationships between them. On mobile devices where you are much more constrained you need to be very observant of allocations. Contrary to popular belief engines like V8 create a "context" object when you create a closure, regardless of "functional is future" soup of the day :wink:. From my understanding*, these are new and short lived they are placed into new generation space which has a size limit. Once that limit is reached a minor GC scavenger runs to release the unreachable. This must stop the world to collect, so if you can perform the same amount of work with less allocations you should.

* This is my interpretation of this and it may be incorrect. For more information you might want to read this. https://strongloop.com/strongblog/node-js-performance-garbage-collection/

topaxi:
@chadhietala thanks a lot for the detailed explanation!

fivetanley:
Does this have benefits on runtimes that aren't V8?

fivetanley:
It would also be great to see before/after numbers from this commit.

krisselden:
Why would it not? All of them allocate for closures and not being careful with them is problematic.

krisselden:
The is an excessive amount of looping and nested looping for relationship setup especially if you have not specified the inverse. The forEach and map closures likely could be made to avoid repeatedly creating them but in hot paths we've already demonstrated this is a hot path, profile an app that uses ember data, you will see relationshipsByName.

No, one single spot of waste is going to make a huge difference but we have to get rid of a lot of them to have an big impact.  When a low end android phone is about 10x slower it becomes a bigger deal.

addon/-private/system/model/errors.js
```diff
@@ -272,14 +272,24 @@ export default Ember.ArrayProxy.extend(Ember.Evented, {
     @private
   */
   _findOrCreateMessages(attribute, messages) {
-    var errors = this.errorsFor(attribute);
+    let errors = this.errorsFor(attribute);
+    let _messages = [];
```
stefanpenner: alloc

fivetanley:
Often these kind of commits feel like insider baseball. It'd be great if we could measure these like we do in the [Ember Performance Suite](https://github.com/eviltrout/ember-performance/tree/master/app/templates). Chad's explanation was great and needed, and having documentation in the form of Pull Requests or commit messages gives us resources to come back to for those of us who aren't familiar with performance tuning.

addon/-private/system/model/errors.js
fivetanley: Are errors generated often enough for this to matter?

addon/-private/serializers/embedded-records-mixin.js
```diff
@@ -348,11 +348,17 @@ export default Ember.Mixin.create({
   */
   _generateSerializedHasMany(snapshot, relationship) {
     let hasMany = snapshot.hasMany(relationship.key);
-    return Ember.A(hasMany).map((embeddedSnapshot) => {
-      var embeddedJson = embeddedSnapshot.record.serialize({ includeId: true });
+    let manyArray = Ember.A(hasMany);
+    let ret = new Array(manyArray.length);
+
+    for (let i = 0, l = manyArray.length; i < l; i++) {
```
stefanpenner: Caching the lengths isn't something need to do, lets not.

stefanpenner:
@krisselden i suspect ^ will have measurable improvements, and even if there are none (unlikely) displaying the due diligence is important.

Doing so will also help others build a better mental model for this sort of optimization.

runspired:
I'd love to see before/after data, cause to me this looks potentially like we're trading a small amount of execution time and more verbose code for a smaller chance of hitting a blocking GC.

stefanpenner:
@runspired it's a known improvement. I'm merely asking for confirmation of the impact. If we don't see an improvement it is likely the result of some other issue. Which would be worth exploring further as it may uncover something we are not yet sure about.

The explicit code, is brain dead simple for the jit to handle, the callback variant is significantly more complex and as far as we know no runtime can yet reduce the existing code into anything nearly as performant and  as the patched variant should be. Maybe someday we will be blessed with more zero cost abstraction run times...


These loops are hit so often, and are themselves in other loops it's worth ensuring we aren't just trolling ourselves with some slip up

chadhietala:
Will try to grab numbers tomorrow.

chadhietala:
Probably some conflation going on here but appears to confirm that the hypothesis.

__Before:__
screen shot 2015-12-14 at 11 44 10 am

__After:__

screen shot 2015-12-14 at 11 41 47 am

stefanpenner:
@fivetanley my above comment has not been addreseed https://github.com/emberjs/data/pull/3988#discussion_r47439430

fivetanley:
We decided to merge this in the core team meeting. The Ember and Ember Data codebases both use this pattern of caching length, it doesn't seem enough to block merging it.

stefanpenner:
@fivetanley can you address it then please, it would be unfortunate to continue to propogate this cargo culting

PR #3995, opened 2015-12-14
[BUGFIX beta] Implement public module API.

rwjblue:
* This replaces the public module API from ember-cli-shims. These shims have been removed from ember-cli-shims in https://github.com/ember-cli/ember-cli-shims/pull/38.
* Adds shims when `define && define.petal` (which is set by loader.js) is present.
* Remove old addon + blueprint.
* If `bower.json` includes `ember-data` a warning will be issued, and that
version will be used.


bmac:
LGTM :+1:

rwjblue:
Updated again, added the following:

* Remove old addon + blueprint.
* Use `bower` version of ember-data if present.

In order to keep backwards compat, we must honor the version of ember-data that is present in `bower.json` (even though it is not needed). This is due to the fact that many people update the `package.json` and `bower.json` versions of Ember Data completely independently.

If `bower.json` includes `ember-data` a warning will be issued, and that version will be used.

rwjblue:
Added two more changes:

* Do not ENABLE_OPTIONAL_FEATURES in consuming applications. -- We were previously setting `EmberENV.ENABLE_OPTIONAL_FEATURES = true` in consuming applications, which is generally not what we want.  This updates to ensure that only the internal addon development is tweaked/changed.
* Provide helpful warning for using older ember-cli-shims. -- As reported with https://github.com/jcope2013/active-model-adapter-ember-cli-2-beta-issue, ember-cli-shims@0.1.0 does not play well with Ember Data 2.3, this warning helps folks know to update.

bmac:
Thanks @rwjblue.

PR #3997, opened 2015-12-14
require missing `fs` module

bmac:


lib/javascripts.js
```diff
@@ -11,6 +11,7 @@ var path          = require('path');
 var Funnel        = require('broccoli-funnel');
 var versionReplace = require('./version-replace');
 var fileCreator   = require('broccoli-file-creator');
+var fs            = require('fs');
```
stefanpenner: its already on line 9

bmac: So it is... I hit this issue on the `beta` branch and assumed it was also a problem with master.

PR #4000, opened 2015-12-16
Re-enable pushing builds to S3

pangratz:
If I am not mistaken, it looks like the builds are not published on S3...

The `bin/publish-builds` script is inspired by [the one in the Ember.js repo](https://github.com/emberjs/ember.js/blob/fbab176eb41727dc6a4718ee3dfdc6490706b3b2/bin/publish_builds).


bmac:
Thanks @pangratz

PR #4002, opened 2015-12-16
Fix paths to files which should be uploaded to S3

pangratz:
The generated files via `ember build --env=production` are located
inside the `dist/globals` directory.

---

This is yet another follow up for #4000, but when I am reading [the corresponding code in `ember-publisher`](https://github.com/rondale-sc/ember-publisher/blob/8bff5417cf94fced70fb50705f88a9563c7befff/index.js#L75-L81) correctly, this should finally work so the files - which are located inside `dist/globals/` and not just `dist/` - are uploaded to S3.


fivetanley:
:+1:

PR #4006, opened 2015-12-17
[CLEANUP] Separate buildURL tests from pathForType tests

HeroicEric:
It was not obvious that I would find the tests for `adapter.buildURL`
inside the file at `build-url-mixin/path-for-type-test.js`.

I figured it might be helpful to others to move these tests to their own
location.


stefanpenner:
:+1:

bmac:
Thanks @HeroicEric

PR #4007, opened 2015-12-17
[CLEANUP] Use `isEnabled` instead of `Ember.FEATURES` directly

HeroicEric:
To stay consistent with the usage elsewhere


pangratz:
I missed that - thank you kindly!

bmac:
Thanks @HeroicEric

Issue #3936, opened 2015-11-17
Avoid private API


krisselden:
Should not use  `relatedTypesDescriptor._cacheable = false;` in testing as this is now `_volatile = true`

wecc:
It was suggested in #4008 that this code should be removed completely:

> It seems like if the change to _cacheable/_volatile isn't causing our tests to fail then we should be able to remove the test code that modifies that property since it doesn't appear to be adding much value.

PR #4008, opened 2015-12-17
[CLEANUP] avoid private API for computed properties in tests

pangratz:
This addresses #3936

cc @krisselden

### DIFF
--- a/addon/-private/system/relationships/ext.js
+++ b/addon/-private/system/relationships/ext.js
@@ -12,8 +12,8 @@ var Map = Ember.Map;
 var MapWithDefault = Ember.MapWithDefault;
 
 var relationshipsDescriptor = Ember.computed(function() {
-  if (Ember.testing === true && relationshipsDescriptor._cacheable === true) {
-    relationshipsDescriptor._cacheable = false;
+  if (Ember.testing === true && relationshipsDescriptor._volatile === false) {
+    relationshipsDescriptor._volatile = true;
   }
 
   var map = new MapWithDefault({
@@ -39,8 +39,8 @@ var relationshipsDescriptor = Ember.computed(function() {
 }).readOnly();
 
 var relatedTypesDescriptor = Ember.computed(function() {
-  if (Ember.testing === true && relatedTypesDescriptor._cacheable === true) {
-    relatedTypesDescriptor._cacheable = false;
+  if (Ember.testing === true && relatedTypesDescriptor._volatile === false) {
+    relatedTypesDescriptor._volatile = true;
   }
 
   var modelName;
@@ -67,8 +67,8 @@ var relatedTypesDescriptor = Ember.computed(function() {
 }).readOnly();
 
 var relationshipsByNameDescriptor = Ember.computed(function() {
-  if (Ember.testing === true && relationshipsByNameDescriptor._cacheable === true) {
-    relationshipsByNameDescriptor._cacheable = false;
+  if (Ember.testing === true && relationshipsByNameDescriptor._volatile === false) {
+    relationshipsByNameDescriptor._volatile = true;
   }
 
   var map = Map.create();
--- a/tests/integration/relationships/belongs-to-test.js
+++ b/tests/integration/relationships/belongs-to-test.js
@@ -560,54 +560,54 @@ test("relationshipsByName does not cache a factory", function(assert) {
 test("relationshipsByName is cached in production", function(assert) {
   var model = store.modelFor('user');
   var oldTesting = Ember.testing;
-  //We set the cacheable to true because that is the default state for any CP and then assert that it
+  //We set volatile to false because that is the default state for any CP and then assert that it
   //did not get dynamically changed when accessed
   var relationshipsByName = getComputedPropertyDesc(model, 'relationshipsByName');
-  var oldCacheable = relationshipsByName._cacheable;
-  relationshipsByName._cacheable = true;
+  var oldVolatile = relationshipsByName._volatile;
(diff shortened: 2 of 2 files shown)


addon/-private/system/relationships/ext.js
```diff
@@ -12,8 +12,8 @@ var Map = Ember.Map;
 var MapWithDefault = Ember.MapWithDefault;
 
 var relationshipsDescriptor = Ember.computed(function() {
-  if (Ember.testing === true && relationshipsDescriptor._cacheable === true) {
-    relationshipsDescriptor._cacheable = false;
+  if (Ember.testing === true && relationshipsDescriptor._volatile === true) {
```
sly7-7: should'nt the condition be inversed ? Per @krisselden comment: "Should not use relatedTypesDescriptor._cacheable = false; in testing as this is now _volatile = true"

pangratz: Ouch, wonderful catch! Thanks @sly7-7!!

sly7-7: Hum... seems not to be so wonderfull, the tests are now failing... I'm not sure I'm understand, or something else should be changed too ?

pangratz: It looks like updating the tests as well helps with a happy Travis :wink: Thanks again for your :eyes:!

addon/-private/system/relationships/ext.js
```diff
@@ -39,8 +39,8 @@ var relationshipsDescriptor = Ember.computed(function() {
 }).readOnly();
 
 var relatedTypesDescriptor = Ember.computed(function() {
-  if (Ember.testing === true && relatedTypesDescriptor._cacheable === true) {
-    relatedTypesDescriptor._cacheable = false;
+  if (Ember.testing === true && relatedTypesDescriptor._volatile === true) {
```
sly7-7: so I think this one has been missed :eyeglasses: ?

pangratz: :weary:, next :beer: is on me!

tests/integration/relationships/belongs-to-test.js
```diff
@@ -563,15 +563,15 @@ test("relationshipsByName is cached in production", function(assert) {
   //We set the cacheable to true because that is the default state for any CP and then assert that it
```
sly7-7: maybe the comment could be switched too ?

tests/integration/relationships/belongs-to-test.js
```diff
@@ -581,15 +581,15 @@ test("relatedTypes is cached in production", function(assert) {
...
```
sly7-7: ^^

tests/integration/relationships/belongs-to-test.js
```diff
@@ -599,15 +599,15 @@ test("relationships is cached in production", function(assert) {
...
```
sly7-7: ^^

tests/integration/relationships/belongs-to-test.js
pangratz: :weary:, next :100: :beers: are on me!

sly7-7: Now I'm drunk :astonished:

fivetanley:
I'm not sure what the original code is trying to accomplish. While this seems fine from a cleanup perspective, I wonder if we could remove it entirely. It seems to be only set in testing which could possibly lead to confusing bugs that don't show up in our test suite but might show up in an app's test suite.

fivetanley:
@igorT do you know what's going on in the original code?

krisselden:
@fivetanley I agree with you on this, I have no idea what this is doing, or why it would be passing now.   

Also cleanup implies you are refactoring the need for private API, here you are just renaming the field.

Please remove this code instead as @fivetanley suggests, then it actually will be cleanup.

stefanpenner:
> I'm not sure what the original code is trying to accomplish.

Might be worth investigating this before prescribing solutions.

If I recall, this was a cross ember version related performance fix. And the above method at the time was the appropriate solution (while still being backwards compat and all that).

I am unsure if this is still needed, maybe @igorT knows.

fivetanley:
> Might be worth investigating this before prescribing solutions.

"I wonder" is less of a prescription and more of a suggestion for investigation. :stuck_out_tongue:

igorT:
There are two parts to this:
1. In ember data tests, we mess with volatile to assert that we do cache it in production.
2. We cache relationship links in production code, as model relationship options are assumed immutable once accessed, however this may not be the case when running your tests, as you could be reopening your classes very often. 

@stefanpenner I don't quite recall this being a backwards compat only thing?

krisselden:
I'm still not clear, can you not just use `Ember.cacheFor(obj, key)` public API instead to assert it has been cached?

The descriptor property here was changed by me because I thought it was safe to change being private and I wanted it to align with the setter which is `desc.volatile()`.  This change also did not seem to break any tests, so I'm very unclear what the PR is about but this should not be touching a private field.

stefanpenner:
just checked, the code i was thinking of no longer exists. So my above comment is likely just a distraction.

bmac:
It seems like if the change to `_cacheable`/`_volatile` isn't causing our tests to fail then we should be able to remove the test code that modifies that property since it doesn't appear to be adding much value.

bmac:
@pangratz can this pr be closed?

pangratz:
@bmac yeah, let's close and investigate if the tests can be removed as you suggested here in your comment above.

PR #4009, opened 2015-12-17
[FEATURE ds-references] More conistency for RecordReference

pangratz:
Since those properties are functions on a BelongsToReference and a
HasManyReference, they are changed for the RecordReference as well.
This makes it consistent.


bmac:
Thanks @pangratz

Issue #3640, opened 2015-08-12
BelongsTo seems not update canonicalRecord when it is set to null


SelokYeung:
I was trying to rollback a relationship, and I found that canonicalRecord didn't update after save if it's value is set to null.  I don't know whether it is issue, but it is weird for canonicalRecord still keep the invalid record after save.

[data/packages/ember-data/lib/system/relationships/state/belongs-to.js](https://github.com/emberjs/data/blob/v1.13.7/packages/ember-data/lib/system/relationships/state/belongs-to.js#L34)

I guess it should be modify as follow
```javascript
BelongsToRelationship.prototype.setCanonicalRecord = function (newRecord) {
	if (newRecord) {
		this.addCanonicalRecord(newRecord);
	} else {
		this.removeCanonicalRecord(this.canonicalState);
		if (this.inverseRecord) {
			this.removeCanonicalRecord(this.inverseRecord);
		}
	}
	this.setHasData(true);
};
```

gilpinheiro:
I'm seeing the same thing, this has bitten me. ember-data: 1.13.4

SelokYeung:
I do it in dirty ways, I hooked didUpdate event on every model. 
So everytime those model saved. It will check for all belongsto relationship. if it is empty then cleanup canonical record.

```javascript
onDataChanged: function () {
	var self = this;

	this.eachRelationship(function (name, relationship) {
		if (relationship.kind === 'belongsTo') {
			if (Em.isEmpty(self.get(name))) {
				var internalRelation = self.get('_internalModel')._relationships.get(name);
				internalRelation.removeCanonicalRecord(internalRelation.canonicalState);
			}
		}
	});
}.on('didUpdate'),
```

Seems work fine in ember-data: 1.13 and 2.0

bmac:
Resolved via 8400b9ddbce36f7d93c5b519eda974e46f612fd8

PR #4012, opened 2015-12-17
setCanonicalRecord should check for this.canonicalState, not this.inverseRecord

jsedlacek:
when calling setCanonicalRecord(null) it should check for presence
of canonicalState instead of inverseRecord.

Otherwise setCanonicalRecord(null) does nothing when inverseRecord is also null.

### DIFF
--- a/packages/ember-data/lib/system/relationships/state/belongs-to.js
+++ b/packages/ember-data/lib/system/relationships/state/belongs-to.js
@@ -35,8 +35,8 @@ BelongsToRelationship.prototype.setRecord = function(newRecord) {
 BelongsToRelationship.prototype.setCanonicalRecord = function(newRecord) {
   if (newRecord) {
     this.addCanonicalRecord(newRecord);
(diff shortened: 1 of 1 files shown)


bmac:
@jsedlacek Do you mind adding a test case for this?

@igorT do you mind reviewing this pr?

jsedlacek:
@bmac Sure. Can you point me to a place to put the test in? I did not find a place where is this class tested.

pangratz:
Seem related: #3640

fivetanley:
How does this get hit from the public API?

jsedlacek:
@fivetanley I am using `canonicalState` to rollback relationship from my own mixin. I realized that when you set relationship to `null` and `save`, then the `canonicalState` is incorrect.

I am not sure if this touches ember-data public interface.

bmac:
Resolved via https://github.com/emberjs/data/commit/7561cb7900d1d6c8875fa0a84d70cf51575accf8

PR #4014, opened 2015-12-18
Register the version before the DEBUG version are printed when loadin…

bmac:
…g Ember Data via ember cli

## Before
screen shot 2015-12-18 at 11 26 38 am

## After
screen shot 2015-12-18 at 11 22 54 am


stefanpenner:
:+1:

PR #4021, opened 2015-12-21
[BUGFIX beta] Move ember-inflector to an explicit dependency for the …

bmac:
…Ember CLI build

### DIFF
--- a/package.json
+++ b/package.json
@@ -30,6 +30,7 @@
     "ember-cli-string-utils": "^1.0.0",
     "ember-cli-test-info": "^1.0.0",
     "ember-cli-version-checker": "^1.1.4",
+    "ember-inflector": "^1.9.4",
     "inflection": "^1.8.0",
     "silent-error": "^1.0.0"
   },
@@ -62,7 +63,6 @@
     "ember-disable-prototype-extensions": "^1.0.0",
     "ember-disable-proxy-controllers": "^1.0.0",
     "ember-export-application-global": "^1.0.3",
-    "ember-inflector": "^1.9.4",
     "ember-publisher": "0.0.7",
     "ember-try": "0.0.6",
     "ember-watson": "^0.7.0",


fivetanley:
This will cause it not to load. Ember CLI won't look at nested dependencies.

fivetanley:
Err, it's already in peerDependencies, do we need it as a dependency?

rwjblue:
> This will cause it not to load. Ember CLI won't look at nested dependencies.

Not sure what you mean. ember-cli definitely supports nested dependencies. There are some issues with nested deps that use `included`, but that isn't the case here ([based in this](https://github.com/stefanpenner/ember-inflector/blob/master/index.js)).

rwjblue:
`peerDependencies` allows the application to bring its own also, but if ember-data does actually depend on ember-inflector why would we not want it as a dep?

bmac:
@fivetanley this may not be the correct fix. 
I just created [this repo](https://github.com/bmac/test-2-3-0-beta) in an attempt to debug http://discuss.emberjs.com/t/jsonapiadapter-not-recognized-findall-is-not-a-function/9307 and I was seeing this error. 

screen shot 2015-12-21 at 10 28 55 am


Previously I had been using [this repo](https://github.com/bmac/active-model-adapter-ember-cli-2-beta-issue) to test the 2.3 beta but I didn't notice until now that it explicitly included the [ember-inflector dependency](https://github.com/bmac/active-model-adapter-ember-cli-2-beta-issue/blob/master/package.json#L42).

fivetanley:
> ember-cli definitely supports nested dependencies

when I was starting to look at addonifying in the early part of the year, it sounded like this was not the case. I'm glad it does account for this now. Using it as a `dependency` seems fine then.

@bmac are you using npm 2 or 3? 2 warns, 3 fails install if the peerDep isn't met

bmac:
Looks like I'm currently using npm `2.14.2`.

fivetanley:
Doing it as a dependency, since it's possible (my bad, i thought it wasn't), seems better so we won't get tickets like this. Not sure what the advantage of using a peerDependency would be if nested dependencies work.

rwjblue:
> when I was starting to look at addonifying in the early part of the year, it sounded like this was not the case.

Ya, you are totally correct.  It only landed a bit before 1.13.0 of ember-cli, so it hasn't been around too long.

Issue #3785, opened 2015-09-18
BooleanTransform should allow null values


jaaksarv:
Currently BooleanTransform converts null values to false. This should not be default behavior. It is common to have nullable boolean values in database that represent situation where value is not yet defined. StringTransform, NumberTransform and DateTransform all support null values, so should BooleanTransform to make things consistent.

bmac:
Hi @jaaksarv. I think you raise a very good point that `BooleanTransform` is the only transform that doesn't support null. Unfortunately, because of semver I don't think this is a change Ember Data could make until 3.0. For now I think the best work around is to define a custom transform that supports nulls. 

```js
// app/transforms/nullable-boolean.js
export default DS.BooleanTransform.extend({
  deserialize: function(serialized) {
    if (serialized === null) {
      return null;
    }
    return this._super(serialized);
  },

  serialize: function(deserialized) {
    if (deserialized === null) {
      return null;
    }
    return this._super(deserialized);
  }
});
```
```js
// app/models/user.js
export default DS.Model.extend({
  isAdmin: attr('boolean'),
  isStaff: attr('nullable-boolean'),
  email: attr('string')
});
```

bmac:
Hi @jaaksarv. I'm going to close this issue because I don't think we can change the behavior of the `BooleanTransform` at this time.

Feel free to reopen or create a new pr if you would like to add a new Transform that supports null.

jaaksarv:
bmac Why closing? It should be fixed in 3.0 as you described above.

bmac:
Ok I will leave it open.

fivetanley:
We should add support for this feature when emberjs/rfcs#1, assuming:

1) Feature flag the deprecations
2) Decide later whether this should be the default
3) Opt in with `{allowNull: true}`

bmac:
This should be solved by https://github.com/emberjs/data/pull/4022 which will be enabled in the next beta.

Issue #3111, opened 2015-05-26
Transforms have no idea about attributes


louy:
I'd like to discuss an issue that I have faced recently using Ember data.

As I see it, the current functionality of Transforms is very limited, which shouldn't be the case. Specifically, there is no way to know anything about the Model or the attribute in Transform which makes a few use cases hard to implement. For example, [`ember-data.model-fragments`#96](https://github.com/lytics/ember-data.model-fragments/issues/95).

What I suggest here is that there could be a way for the transform class to know which model and/or which attribute it's serializing/deserializing, so that it can have far more functionality. Then it will also be able to read attribute's meta and apply some filters according to that. I think there are many use cases for this (like `String.trim()`.)

I'm not sure if this fits the concept of transform, but I believe it would be a good thing to have. What do you think?

igorT:
I know @wycats was against this before for simplicity sake, but I had hacked together in couple apps, so seems beneficial. I'll tag it for review

fivetanley:
We have chosen to not support this API. @tomdale says this introduces too much complexity. We have decided to implement emberjs/rfcs#1 instead, which should allow you to do this kind of thing, even if it is less dynamic in nature.

PR #4022, opened 2015-12-21
[FEATURE ds-boolean-transform-allow-null] allow null for boolean

pangratz:
*NOTE:* this description is not completely accurate anymore, since the implementation changed so the `allowNull` flag needs to be passed to the options for the attribute. See discussion below for further context.

---

This feature allows `null` / `undefined` values for `DS.attr('boolean')`
attributes. Currently `DS.BooleanTransform` converts those values to
`false`; which means that a `DS.attr('boolean')` cannot be set to `null`
from the server. Other transforms (`string`, `number`) do allow `null`
values, so this is an inconsistency between the transforms.

The new feature `ds-boolean-transform-allow-null` changes this behavior
so `null` / `undefined` are converted to `null`. Since **this is a semver
breaking change**, a corresponding deprecation warning should be added
before this is accepted in ember data.

Current behavior of `DS.attr('boolean')`:

| incoming    | DS      | serialized |
|-------------|---------|------------|
| `null`      | `false` | `false`    |
| `undefined` | `false` | `false`    |

Behavior of `DS.attr('boolean')`, having
`ds-boolean-transform-allow-null` feature enabled:

| incoming    | DS      | serialized |
|-------------|---------|------------|
| `null`      | `null`  | `null`     |
| `undefined` | `null`  | `null`     |

---

Note that this feature only works if `ds-transform-pass-options` is
enabled as well, since passing the options from `DS.attr` to the
transform is added with that feature.

---

This addresses #3785.

---

### TODO

- [x] wait for implementation of https://github.com/emberjs/rfcs/pull/1 so this PR can be updated with `allowNull` flag

### DIFF
--- a/addon/-private/transforms/boolean.js
+++ b/addon/-private/transforms/boolean.js
@@ -1,4 +1,8 @@
+import Ember from 'ember';
 import Transform from "ember-data/transform";
+import isEnabled from 'ember-data/-private/features';
+
+const { isNone } = Ember;
 
 /**
   The `DS.BooleanTransform` class is used to serialize and deserialize
@@ -23,8 +27,16 @@ import Transform from "ember-data/transform";
   @namespace DS
  */
 export default Transform.extend({
-  deserialize(serialized) {
+  deserialize(serialized, options) {
     var type = typeof serialized;
+
+    if (isEnabled('ds-transform-pass-options')) {
+      if (isEnabled('ds-boolean-transform-allow-null')) {
+        if (isNone(serialized) && options.allowNull === true) {
+          return null;
+        }
+      }
+    }
 
     if (type === "boolean") {
       return serialized;
@@ -37,7 +49,15 @@ export default Transform.extend({
     }
   },
 
-  serialize(deserialized) {
+  serialize(deserialized, options) {
+    if (isEnabled('ds-transform-pass-options')) {
+      if (isEnabled('ds-boolean-transform-allow-null')) {
+        if (isNone(deserialized) && options.allowNull === true) {
+          return null;
+        }
+      }
+    }
+
     return Boolean(deserialized);
   }
 });
--- a/FEATURES.md
+++ b/FEATURES.md
@@ -11,6 +11,12 @@ entry in `config/features.json`.
 
 ## Feature Flags
 
+- `ds-boolean-transform-allow-null`
+
+  Allow `null`/`undefined` values for `boolean` attributes via `DS.attr('boolean', { allowNull: true })`
+
+  Note that this feature only works when `ds-transform-pass-options` is enabled too.
+
 - `ds-finder-include`
 
   Allows an `include` query parameter to be specified with using
--- a/config/features.json
+++ b/config/features.json
@@ -1,4 +1,5 @@
 {
+  "ds-boolean-transform-allow-null": null,
   "ds-finder-include": null,
   "ds-references": null,
   "ds-transform-pass-options": null,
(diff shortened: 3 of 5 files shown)


fivetanley:
@pangratz During the team discussion someone suggested passing an options hash to `DS.attr({allowNull: true})` so we could maintain backwards compatibility. What do you think about this change?

pangratz:
Seems good to me. But for this to work https://github.com/emberjs/rfcs/pull/1 (also see discussion in #3111) needs to be go'ed and implemented, since currently the hash is not passed to transformations.

Would the eventual default in `3.x` be that `null` are allowed for `DS.attr('boolean')`, so it is consistent with `number`, `string` and `date` transforms?

pangratz:
Alright, https://github.com/emberjs/rfcs/pull/1 has been accepted. Once that RFC is implemented I will update this PR so the `allowNull` flag is considered.

pangratz:
Alrighty, I rebased onto latest `master` and now null is returned for `null`/`undefined` if the `allowNull` flag is set to `true`:

```js
export default Model.extend({
  leBool: attr('boolean', { allowNull: true })
});
```

addon/-private/transforms/boolean.js
```diff
@@ -38,6 +50,14 @@ export default Transform.extend({
   },
 
   serialize(deserialized) {
+    if (isEnabled('ds-transform-pass-options')) {
+      if (isEnabled('ds-boolean-transform-allow-null')) {
```
bmac: Should this be checking the 2nd argument for `{ allowNull: true }`?

pangratz:
@bmac I somehow forgot to commit the changes considering the `allowNull` flag :facepalm: sorry for the jump start.

I - once again - updated the PR; now it should be ready for review :wink:

btecu:
Wouldn't `defaultValue` work for this?
I think it should work at least for `undefined`.

pangratz:
I think this wouldn't work since `defaultValue` is only used when the value of an attribute [is retrieved from a `DS.Model`](https://github.com/emberjs/data/blob/v2.3.3/addon/attr.js#L104-L111), but not when an attribute is serialized...

addon/-private/transforms/boolean.js
```diff
@@ -23,9 +27,17 @@ import Transform from "ember-data/transform";
   @namespace DS
  */
 export default Transform.extend({
-  deserialize(serialized) {
+  deserialize(serialized, options) {
     var type = typeof serialized;
 
+    if (isEnabled('ds-transform-pass-options')) {
+      if (isEnabled('ds-boolean-transform-allow-null')) {
+        if (isNone(serialized) && options.allowNull === true) {
```
bmac: Should we use `Ember.get(options, 'allowNull')` in case `options` is undefined?

pangratz: Theoretically `options` should never be `undefined` because of [this](https://github.com/emberjs/data/blob/c78d7f0ea9b12c077dd6573b6d9e01e9ee1d2d70/addon/serializers/json.js#L201) and ultimately because of [this](https://github.com/emberjs/data/blob/48a0094f719b39c3dfbe48884cf0b7cbb930e5b2/addon/attr.js#L94).

It would only be `undefined` if the transform is used outside of regular `ember-data` usage...

bmac: Sounds good.

sandstrom:
This is a great adjustment! 🎆 

@fivetanley for 3.0, would you mind flipping the default of the option (or removing the option altogether)? In our app I'd say 99% of boolean values across all our models would use the allowNull option. Also, it's inconsistent with string, number etc. where no option is required.

PR #4029, opened 2015-12-23
Move public modules out of the `-private` folder

### LINKED ISSUE #3998: New public API does not compile in ember-cli 1.13.8
`addon/relationships.js` doesn't get compiled in my output (using ember-cli 1.13.8) and therefore results in a syntax error.

`vendor.js`:
```javascript
// snip...
define('ember-data/relationships', function () {

    'use strict';

    export * from 'ember-data/-private/system/relationships';

});
// ...snip
```

bmac:
Closes #3998 

cc @rwjblue @fivetanley


fivetanley:
It's odd that most of these aren't showing up as moves. Did you use `git mv`?

fivetanley:
Otherwise it looks fine to me but it would be nice to preserve history if we could. (even if it requires `git log --follow`)

bmac:
I did use git move originally. I think changing some of the imports is confusing github's display of the moves. 

git status shows several of these changes as a rename where github is making it look like a file was deleted.
screen shot 2015-12-30 at 12 45 17 pm

fivetanley:
Ah nice. Seems like it needs `[bugfix BETA]` and then it's good to go?

PR #4032, opened 2015-12-28
[CLEANUP] remove obsolete code within Store#willDestroy

pangratz:
`this._containerCache` has was added in 31583db and removed from the store in ee97760.


bmac:
Thanks @pangratz

PR #4033, opened 2015-12-29
[BUGFIX release] extractMeta impossible to use with JSONSerializer

raido:
When using JSONSerializer it is impossible to extract metadata from query response because when payload is modified in normalizeQueryResponse (or any other hook) the metadata will go missing by the time extractMetadata is called.

I dug up earlier implementation before v1.13 and back then extractMeta was called before extractQueryResponse etc. methods - https://github.com/emberjs/data/blob/v1.0.0-beta.19.2/packages/ember-data/lib/serializers/json-serializer.js#L763

Which means that extractMeta always received unmodified payload. After 1.13 (making it a breaking change in a minor version) this changed, so extractMeta was called in _normalizeResponse method (https://github.com/emberjs/data/blob/v1.13.0/packages/ember-data/lib/serializers/json-serializer.js#L446, https://github.com/emberjs/data/blob/v2.2.1/packages/ember-data/lib/serializers/json-serializer.js#L437), which is the end path for all the normalize(Query/Find) etc. methods. This is the main issue here, because for example if one passes extracted array of records from the payload  to normalizeQueryResponse _super call, then there will no metadata left to extract. 

This restores previous "extract" method implementation logic, where first "extractMeta" is called and then "normalize*" methods.

### DIFF
--- a/addon/-private/serializers/json-serializer.js
+++ b/addon/-private/serializers/json-serializer.js
@@ -230,6 +230,29 @@ export default Serializer.extend({
     @return {Object} JSON-API Document
   */
   normalizeResponse(store, primaryModelClass, payload, id, requestType) {
+    let meta = this.extractMeta(store, primaryModelClass, payload);
+
+    let documentHash = this._normalizeResponseForType(...arguments);
+
+    if (meta) {
+      assert('The `meta` returned from `extractMeta` has to be an object, not "' + Ember.typeOf(meta) + '".', Ember.typeOf(meta) === 'object');
+      documentHash.meta = meta;
+    }
+    return documentHash;
+  },
+
+  /**
+    @method _normalizeResponseForType
+    @param {DS.Store} store
+    @param {DS.Model} primaryModelClass
+    @param {Object} payload
+    @param {String|Number} id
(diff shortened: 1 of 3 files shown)


bmac:
This seems ok to me. @wecc do you mind reviewing this pr?

bmac:
@raido do you mind sharing an example of your `extractQueryResponse` with a meta property you'd like to extract?

raido:
@bmac it is something like in the unit test i included here. But if that is not enough i can make an example on Twiddle.

raido:
Twiddle - https://ember-twiddle.com/efc6b04753e705b40bbc?openFiles=application.serializer.js%2Cuser.adapter.js

raido:
There's a clever trick how to have metadata with JSONSerializer - https://ember-twiddle.com/efc6b04753e705b40bbc?openFiles=application.serializer.js%2Cuser.adapter.js

raido:
I turned this into an addon - https://www.npmjs.com/package/ember-json-serializer-meta - which provides a mixin for extracting metadata with JSONSerializer and queries.

PR #4035, opened 2015-12-30
Simplify the urlPrefix method to make it easier to understand

bmac:


fivetanley:
*much better* thank you!

PR #4039, opened 2016-01-04
Replace calls to store.find with store.findRecord

pangratz:
`store.find` has been deprecated in `1.13` and been officially removed in the `2.x` series. This updates all references in the public documentation to use `store.findRecord`. Also some tests were still using `store.find` (which is still available since it is needed in the `model()` hook of Ember, but it is private now).

### DIFF
--- a/addon/-private/system/coerce-id.js
+++ b/addon/-private/system/coerce-id.js
@@ -1,5 +1,5 @@
 // Used by the store to normalize IDs entering the store.  Despite the fact
-// that developers may provide IDs as numbers (e.g., `store.find(Person, 1)`),
+// that developers may provide IDs as numbers (e.g., `store.findRecord('person', 1)`),
 // it is important that internally we use strings, since IDs may be serialized
 // and lose type information.  For example, Ember's router may put a record's
 // ID into the URL, and if we later try to deserialize that URL and find the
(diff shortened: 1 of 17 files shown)


addon/-private/system/model/internal-model.js
```diff
@@ -543,7 +543,7 @@ InternalModel.prototype = {
     came back from the server, except the user obtained them out of band and is informing
     the store of their existence. The most common use case is for supporting client side
     nested URLs, such as `/posts/1/comments/2` so the user can do
-    `store.find('comment', 2, {post:1})` without having to fetch the post.
+    `store.findRecord('comment', 2, {post:1})` without having to fetch the post.
```
wecc: I think this should be `store.findRecord('comment', 2, { preload: { post: 1 } })`

addon/-private/system/coerce-id.js
```diff
@@ -1,5 +1,5 @@
 // Used by the store to normalize IDs entering the store.  Despite the fact
-// that developers may provide IDs as numbers (e.g., `store.find(Person, 1)`),
+// that developers may provide IDs as numbers (e.g., `store.findRecord(Person, 1)`),
```
wecc: `store.find('person', 1)`?

wecc:
Nice @pangratz! Other than the comments above this LGTM

addon/-private/system/coerce-id.js
pangratz: :+1:

addon/-private/system/model/internal-model.js
pangratz: Good catch! :eyeglasses:

Issue #2456, opened 2014-11-08
Use shared test helpers from ember-dev.


rwjblue:
Pulled into ember in https://github.com/emberjs/ember.js/pull/9516.

Helpers from https://github.com/emberjs/ember-dev.  Can be used as a plain bower repo, or an ember-cli addon (though I have not tested as an addon just yet).

Basically in Ember we did the following:

```javascript
var EmberDevTestHelperAssert = window.Ember.__loader.require('ember-dev/test-helper/index')['default'];
var setupQUnit = window.Ember.__loader.require('ember-dev/test-helper/setup-qunit')['default'];

var testHelpers = new EmberDevTestHelperAssert(window.Ember, EmberDev.runningProdBuild);
setupQUnit(testHelpers);
```

This adds the following helpers (not a complete list, just what I could think of):

* `expectDeprecation`
* `expectNoDeprecation` 
* `ignoreDeprecation`
* `expectAssertion`
* `ignoreAssertion`

Also, confirms that the following is true at the end of each test (to ensure tests do not bleed through to each other):

* No pending run loops.
* No views in `Ember.View.views`
* No templates in `Ember.TEMPLATES`

igorT:
Ping

bmac:
I'll try to get to this issue this week.

fivetanley:
closing thanks to @pangratz in #4040

PR #4040, opened 2016-01-04
Use test helpers from ember dev

pangratz:
This PR replaces the currently used test helpers to check for correct assertions, deprecations etcetera with the ones from [emberjs/ember-dev](https://github.com/emberjs/ember-dev).

This addresses #2456

---

### TODO

- [x] investigate why tests are not failing although `ember-dev` helper [raises error](https://travis-ci.org/pangratz/data/builds/100128108#L1023-L1032)
- [x] fix failing test by wrapping [this](https://git.io/vzZw9) line in `Ember.run`
- [x] decide if we want to use patched `QUnit.module` or our own `module`


pangratz:
Hmm, looks like although the `ember-dev` helpers throw an error that one test has a pending run loop at the end, the build does not fail: https://travis-ci.org/pangratz/data/builds/100128108#L1023-L1032.

I am investigating...

stefanpenner:
awesome

pangratz:
Alright, so `ember-dev` is now used correctly. The failing test is expected since a run-loop is triggered [here](https://git.io/vzZw9), but it is not wrapped within `Ember.run`.

The problem with the previous approach was that the `ember-dev` helpers were instantiated within `QUnit.testStart()` and tested in `QUnit.testDone()`. If an failing assertion is added within those callbacks, the corresponding test is **not**  marked as failed.

---

The current, working approach patches the `QUnit.module` so the `ember-dev` helpers are setup and used in the `beforeEach` and `afterEach` hooks.

The question is now if we want to use the approach of a patched `QUnit.module` - like it is also done in `ember-dev`'s `setupQUnit` - or if we want to use another approach, where we create a custom `module` and use this in all the tests:

```js
// ember-data/tests/helpers/qunit.js
import { module } from "ember-qunit";

export default function(description, options = {}) {
  var originalBeforeEach = options.beforeEach || function() {};
  var originalAfterEach = options.afterEach || function() {};
  
  // ...

  return module(...arguments);
}
```

And use this module in all the tests (instead of `import { module } from "ember-qunit";`):

```js
// ember-data/tests/integration/le-test.js
import module from "../helpers/qunit";
import { test } from "ember-qunit";

module("le module");

test("le test", function() {
  // ...
});
```

pangratz:
I think it is ok to use the patched QUnit module version for now, as this is also the way it is done in `ember-dev`. Once this lands in ember-data, I will backport stuff to `ember-dev` so it is usable for other addons as well...

bmac:
Thanks @pangratz

PR #4042, opened 2016-01-04
[BUGFIX beta] prevent calls to store.query leaking

### LINKED ISSUE #4041: Potential memory leak in store.query()
When calling store.query(), the heap grows.  I've setup a sample project that demonstrates the issue here:  https://github.com/trumb1mj/ember-data-growing-recordarrays

fivetanley:
Although the result from `store.query` is intended to never be re-used
as we delegate to the server for the truth of a query, the user still
needs an escape hatch to destroy these AdapterPopulatedRecordArrays
created by query in order to prevent excessive memory.

Before, calling `destroy` on an AdapterPopulatedRecordArray (the type
returned from store.query), the logic in
RecordArrayManager#unregisterRecordArray did not account for adapter
populated record arrays resulting in a leak.

This commit changes the logic to search for the array in the
RecordArrayManager's AdapterPopulatedRecordArray collection when
unregistering a record array.

Fixes #4041

### DIFF
--- a/addon/-private/system/record-array-manager.js
+++ b/addon/-private/system/record-array-manager.js
@@ -279,17 +279,22 @@ export default Ember.Object.extend({
     var typeClass = array.type;
 
     // unregister filtered record array
-    var recordArrays = this.filteredRecordArrays.get(typeClass);
-    var index = recordArrays.indexOf(array);
-    if (index !== -1) {
-      recordArrays.splice(index, 1);
+    const recordArrays = this.filteredRecordArrays.get(typeClass);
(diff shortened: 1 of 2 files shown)


GeekSMN:
In the changelog I only see this fix for 2.3.0, was it merged to 1.13.x anywhere?

bmac:
I do not believe this pr was merged into 1.13 @GeekSMN.

GeekSMN:
Would getting this in 1.13 be possible or are there blocking differences? We're seeing a gradual leak that could be this, but can't migrate to 2.0 just yet.

PR #4043, opened 2016-01-04
chore(build): strip debug statements from production builds

eriktrom:
@fivetanley - from our discussion the other day - this does not fix the entire problem, which goes much deeper, but this does fix the bower production build (which has been broken for some time b/c debug statements(or callers of them) we're not stripped.)

To make this fully work, meaning for it to work without using the legacy bower build we'll need to address this error, which is the next error that pops up when using the non-bower build.

```
Could not find module `ember-data/-private/setup-container` imported from `my-app/initializers/ember-data`
```

Sorry I wish I had more context, perhaps you or someone else does. I've hacked around it by copying the output when running `ember build --prod`, e.g., dist/global/ember-data.js, ember-data.prod.js and ember-data.min.js, (with this commit merged) into my own bower repo(which i check in). There and in my package.json I've set the version number back to ^2.2.1 - like I said, super hacky - but after trying to get rid of error above, including updating ember-cli and some other deps, I found lots of odd things - even double module prefixes such as `ember-data/ember-data/blah/blah` in one case, and this error **always**. 

Anyway, let me know if you have any questions or drop me your thoughts/wisdom and I'll try to move forward with this PR or another which fixes the problem entirely.

### DIFF
--- a/lib/javascripts.js
+++ b/lib/javascripts.js
@@ -1,3 +1,5 @@
+/* jshint node:true */
+
 var filterImports = require('babel-plugin-filter-imports');
 var featureFlags  = require('babel-plugin-feature-flags');
 var babel         = require('broccoli-babel-transpiler');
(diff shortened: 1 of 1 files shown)


fivetanley:
Merged as 9615003ad5cb9c7f503f198ff5617a4cd9fba547. Thanks!

Issue #3886, opened 2015-10-24
ember 2.X this.get('model').addObjects(records); throw excption "Uncaught TypeError: internalModel.getRecord is not a function"


tomarsamar:
Afrer updating to ember 2.X, this.get('model').addObjects(records); throw excption "Uncaught TypeError: internalModel.getRecord is not a function"

```
 load_more: function(){
 var self = this;
  this.get("store").query("actor",{pg: 1}}).then(function(records) {

    self.get('model').addObjects(records); // this throw excepton 
   
  }
}
```
this was working perfectlly but now self.get('model').addObjects(records); throw excption "Uncaught TypeError: internalModel.getRecord is not a function" in record-array.js at line 86 "return internalModel && internalModel.getRecord()". I tried using self.get('model').pushObjects(records) but it also give same error

jcope2013:
https://github.com/emberjs/data/issues/3530

fivetanley:
Closing because as @tchak said in #3530, query adapter populated record arrays are intended to be immutable. I opened #4045 to improve the error messaging. You can use `toArray()` if you want to copy it and then mutate that.

ckl1989:
@fivetanley Hi, should the error message be updated for this.findAll() as well? I am getting the same error for a recordArray returned from findAll().

PR #4045, opened 2016-01-05
[feature ds-better-adapter-populated-record-array-error-messages]

fivetanley:
The result returned from `store.query` is an
AdapterPopulatedRecordArray. This adds better error messages when you
try to mutate the array, as it is considered immutable and the mutation
methods should not be used.

Currently, there are some issues like #3886 where this does not fail in
a helpful way. This feature improves the developer ergonomics to prevent
the mutation from happening and explaining why.

### DIFF
--- a/addon/-private/system/record-arrays/immutable-array-mixin.js
+++ b/addon/-private/system/record-arrays/immutable-array-mixin.js
@@ -1,0 +1,34 @@
+import Ember from 'ember';
+import isEnabled from 'ember-data/-private/features';
+
+const { get } = Ember;
+
+const MUTATION_METHODS = [
+  'addObject',
+  'addObjects',
+  'removeObject',
+  'removeObjects',
+  'unshiftObject',
+  'unshiftObjects',
+  'pushObject',
+  'pushObjects',
+  'reverseObjects',
+  'setObjects',
+  'shiftObject'
+];
+
+function useToArray() {
+  if (isEnabled('ds-better-adapter-populated-record-array-error-messages')) {
+    let type = get(this, 'type').toString();
+    throw new Error(`The result of a server query (on ${type}) is immutable. Use .toArray() to copy the array instead.`);
+  } else {
+    return this._super(...arguments);
+  }
+}
+
+const ImmutableArrayMixin = MUTATION_METHODS.reduce((mixin, method) => {
+  mixin[method] = useToArray;
+  return mixin;
+}, {});
+
+export default Ember.Mixin.create(ImmutableArrayMixin);
--- a/addon/-private/system/record-arrays/immutable-array-mixin.js
+++ b/addon/-private/system/record-arrays/immutable-array-mixin.js
@@ -25,17 +25,16 @@ const MUTATION_METHODS = [
 ];
 
 function useToArray() {
-  if (isEnabled('ds-better-adapter-populated-record-array-error-messages')) {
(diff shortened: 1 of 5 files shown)


bmac:
I'm excited for this. :dancer:

addon/-private/system/record-arrays/immutable-array-mixin.js
```diff
@@ -0,0 +1,34 @@
...
+  'addObjects',
+  'removeObject',
+  'removeObjects',
+  'unshiftObject',
+  'unshiftObjects',
+  'pushObject',
+  'pushObjects',
+  'reverseObjects',
+  'setObjects',
+  'shiftObject'
```
pangratz: Should the following methods be added as well, since they are public and mutate the array too?

- [`clear`](http://emberjs.com/api/classes/Ember.MutableArray.html#method_clear)
- [`popObject`](http://emberjs.com/api/classes/Ember.MutableArray.html#method_popObject)
- [`removeAt`](http://emberjs.com/api/classes/Ember.MutableArray.html#method_removeAt)
- [`removeObject`](http://emberjs.com/api/classes/Ember.MutableArray.html#method_removeObject)
- [`removeObjects`](http://emberjs.com/api/classes/Ember.MutableArray.html#method_removeObjects)
- [`insertAt`](http://emberjs.com/api/classes/Ember.MutableArray.html#method_insertAt)

addon/-private/system/record-arrays/immutable-array-mixin.js
```diff
@@ -0,0 +1,34 @@
...
+];
+
+function useToArray() {
+  if (isEnabled('ds-better-adapter-populated-record-array-error-messages')) {
+    let type = get(this, 'type').toString();
+    throw new Error(`The result of a server query (on ${type}) is immutable. Use .toArray() to copy the array instead.`);
```
pangratz: Should this be `Ember.Error`, since I think it provides a better stacktrace?

bmac:
@fivetanley Would you mind rebasing this pr and responding to @pangratz's comments?

bmac:
@fivetanley Do you have time to rebase this pr and responding to @pangratz's comments?

bmac:
ping @fivetanley. I'd like to see this pr get merged. Do you have any free bandwidth to finish up this pr?

buschtoens:
:bell: 

Users appear to have been running into this in https://github.com/offirgolan/ember-light-table/issues/223#issuecomment-253341733 and https://github.com/offirgolan/ember-light-table/issues/224#issuecomment-253607988.

I'd love to see this get merged. :)

pangratz:
I've updated this PR to add more mutation methods and throw an `Ember.Error`.

@bmac @fivetanley can you take a look and squash and merge if this is ok?

bmac:
@pangratz looks like there are a couple of failing tests in this pr.

I also think we need to check to make sure this change doesn't break the emberfire addon.

https://github.com/firebase/emberfire/blob/master/addon/adapters/firebase.js#L264
https://github.com/firebase/emberfire/blob/master/addon/adapters/firebase.js#L275

pangratz:
@bmac rebased.

Hmm, good point about the firebase addon. Not sure what the way forward is here though, as I don't see a way to not break the addon with the current code...

runspired:
With the refactor to native arrays and proxies we also implemented nicer error messaging. This will no longer be needed.

Issue #4047, opened 2016-01-05
Strip assertions when using the Ember CLI Addon with a prod build


bmac:
Stanley Stuart [10:55 AM] 
@bmac based off the existing code https://gist.github.com/fivetanley/1708106ba3372b41dc94

​[10:55] 
we would have to return something like this in `treeForAddon`

​[10:55] 
so basically do an ES6 to ES6 compilation so ember-cli can do the rest as it normally would

​[10:55] 
although this gist needs to be updated for the `-private`-ification

​[10:56] 
similar to https://github.com/emberjs/data/pull/4043

pangratz:
This issue can be closed since #4048 has been merged.

PR #4048, opened 2016-01-05
Strip stuff from addon before it is added to app

pangratz:
Currently the code of the `ember-data` addon is added to the app. This is problematic since the code contains `assert`, `debug`, etcetera calls and also no feature flag code is stripped.

This PR is my stab at fixing this, so only the stripped code is added via the `treeForAddon` hook in `index.js`.

This addresses #4047 and takes inspiration from [@fivetanley's gist](https://gist.github.com/fivetanley/1708106ba3372b41dc94).

---

#### TODO

- [x] Check if this works as expected
- [x] Only strip debug statements imported from `ember-data/-private/debug` if `environment === "production"`? Currently the statements are always stripped.
- [x] ~~Correctly strip features which have a value other than `true` in `config/features.json`.~~ will be addressed in an upcoming PR

### DIFF
--- a/index.js
+++ b/index.js
@@ -49,10 +49,25 @@ module.exports = {
       return { inputTree: dir, rebuild: function() { return []; } };
     }
 
-    var version      = require('./lib/version');
-    var merge        = require('broccoli-merge-trees');
+    var version   = require('./lib/version');
+    var merge     = require('broccoli-merge-trees');
+    var addonTree = merge([version(), dir]);
 
-    return this._super.treeForAddon.call(this, merge([version(), dir]));
+    if (process.env.EMBER_ENV === 'production') {
+      var strippedBuild = require('./lib/stripped-build');
+
+      // blacklist es6.modules so the modules are not compiled but simply the
+      // debug statements / features are stripped; this is taken from
+      // ember-cli-babel:
+      // https://github.com/babel/ember-cli-babel/blob/master/index.js#L71
+      var strippedAddon = strippedBuild('ember-data', addonTree, {
+        blacklist: ['es6.modules', 'useStrict']
+      });
+
+      return this._super.treeForAddon.call(this, strippedAddon);
+    }
+
+    return this._super.treeForAddon.call(this, addonTree);
   },
 
   included: function(app) {
--- a/lib/babel-build.js
+++ b/lib/babel-build.js
@@ -1,0 +1,48 @@
+var babel = require('broccoli-babel-transpiler');
+var path  = require('path');
+
+function babelOptions(libraryName, _options) {
+  _options = _options || {};
+
+  var options = {
+    whitelist: [
+      'es6.templateLiterals',
+      'es6.parameters',
+      'es6.arrowFunctions',
+      'es6.destructuring',
+      'es6.spread',
+      'es6.properties.computed',
+      'es6.properties.shorthand',
+      'es6.blockScoping',
+      'es6.constants',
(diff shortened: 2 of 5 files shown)


pangratz:
Ok, so I tested this locally:

```sh
# checkout ember-data from PR and link to npm
git clone https://github.com/emberjs/data.git
cd data
hub checkout https://github.com/emberjs/data/pull/4048
npm link

# create new ember app using linked ember-data
cd ..
ember new my-app --skip-bower --skip-npm
cd my-app
bower install
bower uninstall ember-data --save
bower install ember-cli/ember-cli-shims#0.1.0 --save
npm link ember-data

# create build and check if dist/assets/vendor.js contains correctly stripped ember-data
ember build
```

All the debug statements from `ember-data/-private/debug` are correctly stripped (and the feature-flagged stuff is still present behind `if(isEnabled("ds-xxx")) {}`, which is expeceted).

There is a problem though: all the ember-data modules in `dist/assets/vendor.js` are inside additional `define('ember-data/xxx', ...)` calls:

```js
define('ember-data/-private/adapters/build-url-mixin', function () {

  'use strict';

  define('ember-data/-private/adapters/build-url-mixin', ['exports', 'ember'], function (exports, _ember) {

    var get = _ember['default'].get;
    
    // ...
  });

});
```

index.js
```diff
@@ -49,10 +49,20 @@ module.exports = {
       return { inputTree: dir, rebuild: function() { return []; } };
     }
 
-    var version      = require('./lib/version');
-    var merge        = require('broccoli-merge-trees');
+    // TODO only return stripped code if we are in production environment?
```
rwjblue:
```
if (process.env.EMBER_ENV !== 'production') {
  var strippedAddonCode = strippedBuild('ember-data', dir, {
    blacklist: ['es6.modules', 'useStrict']
  });

  return this._super.treeForAddon.call(this, strippedAddonCode);
} else {
  return this._super.treeForAddon.call(this, merge([version(), dir]));
}
```

pangratz: :pray:

rwjblue:
This looks good to me, and I'd like to see this land even without the features.json work done.  The debug statement stripping in production will affect ember-data@2.3.0 when we release it, but since 2.3.x doesn't include features yet the `features.json` portion does not have an impact on 2.3.0.

Thoughts on landing this into beta, and iterating on it for 2.4.0-beta.x series for the feature flag stripping?

bmac:
@pangratz do you mind squashing these commits and prefix the commit message with `[BUGFIX beta]`

eriktrom:
@pangratz - this doesn't work unless the following are moved to dependencies (instead of devDependencies) - at least that's what I found (im using ember-cli-deploy though)

Here's a compare: https://github.com/pangratz/data/compare/fix-ember-addon-production...eriktrom:fix-ember-addon-production

Otherwise, looks good and works for me

Thanks for tackling this

pangratz:
Awesome catch @eriktrom! Thanks for testing!

I've updated this PR to move the dependencies which are needed for a `ember build --prod` into `dependencies` and now it works:

```js
ember new my-app --skip-bower --skip-npm
cd my-app
bower install
bower uninstall ember-data --save
bower install ember-cli/ember-cli-shims#0.1.0 --save
npm install pangratz/data#fix-ember-addon-production
npm install

ember build --prod
```

bmac:
Thanks @pangratz and @eriktrom

PR #4051, opened 2016-01-07
Make setup-container public to allow consuming apps more flexibility with initializer

lukemelia:
Per slack conversation with @fivetanley


eriktrom:
Thanks @lukemelia - i'll give this a shot and see if it works in both of my setups(basic and fastboot)

Just a quick q, if there's time - why does moving this out of `-private` solve the problem? Are modules nested under `-private` not auto re-exported from the addon directory to the app directory - aka - in most addons initializers are re-exported from the app directory w/ `export { default } from 'addon/initializers/some-initializers';` and thus merged into the app directory from the addon directory within your app. Why do we not do that in this addon and how did moving initializer out of `-private` make this work?

thanks again - no need to answer unless you have time

eriktrom:
just an update - this fixes 75% of the problem - the error that both luke and I we're seeing is gone, aka, `cannot find module ember-data/-private/setup-container`

there is still an issue where the debug statements are not being stripped from the production builds when *not* using the bower legacy version of the production build - but we'll fix that in due time

ship it

rwjblue:
There is an open PR to fix the debug statements in production issue (when using addon without an entry in bower.json).

PR #4052, opened 2016-01-07
[DOC beta] Hide private classes from YUIDoc

pangratz:
It looks like marking classes as `@private` is not enough to hide them
from showing up in the API doc at http://emberjs.com/api/data/.

This commit changes the used comments to start with `/*`, which are
definitely ignored by JSDoc.

### DIFF
--- a/addon/-private/system/container-proxy.js
+++ b/addon/-private/system/container-proxy.js
@@ -1,6 +1,6 @@
 import { deprecate } from "ember-data/-private/debug";
 
-/**
+/*
   This is used internally to enable deprecation of container paths and provide
   a decent message to the user indicating how to fix the issue.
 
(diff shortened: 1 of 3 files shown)


rwjblue:
:+1:

rwjblue:
(we use YUIDoc, but JSDoc and YUIDoc agree here)

fivetanley:
Seems like the website should filter out private classes? They are still marked as private on the website.

rwjblue:
@fivetanley - yes, https://github.com/emberjs/website/blob/master/lib/api_docs.rb should be better about not including private classes in the sidebar

bmac:
@pangratz do you mind also removing the 2nd `*` from the method definitions (`createSnapshot ` ect) in `internal-model.js`? iirc if yui doc doesn't see a class for these it will attribute these methods to the wrong class.

pangratz:
@bmac done

PR #4054, opened 2016-01-07
Support for `Collection`

olivierlesnicki:
#### Rationale

I've been using ember and ember-data for the past 12 months now. `Models` are a great to location to store business logic. 

It's very often that we need to derive business logic from an array of models. I was inspired by `Backbone.Collections` and leveraged `RecordArrays` to create a solution that fits ember-data ecosystem.

#### Example

Given the following model

```js
// models/car
export default DS.Model({
  make: DS.attr('string'),
  color: DS.attr('string')
})
```

And route

```js
// routes/index
model() {
  return this.store.findAll('car');
}
```

```hbs
// templates/index
{{model.numberOfRedCars}}
```

Creating the following `collection` object would allow all modules retrieving cars via the store to access the numberOfRedCars property:

```js
//collections/car
export default DS.RecordArray({
  numberOfRedCars: Ember.computed('content', function() {
    return this.get('content').filterBy('color', 'red').get('length');
  })
})
```


bmac:
Hi @olivierlesnicki. Thanks for this pr. we discussed this in the Ember Data meeting today and though this sounded like a good idea for an addon instead of core Ember Data functionality. @igorT is especially interested in solving any pain points you may have integrating an addon with this functionality with Ember Data.

olivierlesnicki:
Will do thanks for the feedback :)

PR #4055, opened 2016-01-09
[BUGFIX beta] pin jquery to 1.11.3 to fix broken build

bmac:
See: http://benlimmer.com/2016/01/08/ember-jquery-dependancies/


pangratz:
:+1:

PR #4057, opened 2016-01-11
[BUGFIX beta] fail builds in ember-cli when ember-cli-shims isn't met

### LINKED ISSUE #4044: If ember-cli-shims is not >= 0.1.0, this should cause the build to fail.
I see people are already hitting this a lot by trying out the betas and missing it in any various log outputs. This should probably exit the build with an error code if it doesn't produce a working build.

https://github.com/emberjs/data/blob/master/index.js#L38

fivetanley:
fixes #4044

I disagree with @rwjblue that this should be a silent error. I've found this same kind of error to be nice to see which plugin is throwing an error where, in case I need to debug it or a file an issue about it.

### DIFF
--- a/index.js
+++ b/index.js
@@ -2,6 +2,7 @@
 'use strict';
 
 var path = require('path');
+var SilentError = require('silent-error');
 
 module.exports = {
   name: 'ember-data',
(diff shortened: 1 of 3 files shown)


fivetanley:
These fail in node because:

```
Using a version of ember-cli-shims prior to 0.1.0 will cause errors while loading Ember Data 2.3+. Please update ember-cli-shims from undefined to 0.1.0.
Error: Using a version of ember-cli-shims prior to 0.1.0 will cause errors while loading Ember Data 2.3+. Please update ember-cli-shims from undefined to 0.1.0.
    at Class.module.exports.init (/Users/stanley/code/data/index.js:47:13)
    at new Class (/Users/stanley/code/data/node_modules/core-object/core-object.js:18:12)
    at /Users/stanley/code/data/node_modules/ember-cli/lib/models/addons-factory.js:48:19
    at visit (/Users/stanley/code/data/node_modules/ember-cli/lib/utilities/DAG.js:23:3)
    at DAG.topsort (/Users/stanley/code/data/node_modules/ember-cli/lib/utilities/DAG.js:82:7)
    at AddonsFactory.initializeAddons (/Users/stanley/code/data/node_modules/ember-cli/lib/models/addons-factory.js:44:9)
    at Project.initializeAddons (/Users/stanley/code/data/node_modules/ember-cli/lib/models/project.js:375:36)
    at Project.eachAddonCommand (/Users/stanley/code/data/node_modules/ember-cli/lib/models/project.js:426:10)
    at module.exports (/Users/stanley/code/data/node_modules/ember-cli/lib/cli/lookup-command.js:34:13)
    at CLI.<anonymous> (/Users/stanley/code/data/node_modules/ember-cli/lib/cli/cli.js:36:26)
```

This error happens when `init` is ran the second time, but not the first.

fivetanley:
The node tests are generating a new app which is running `npm install`, wonder if we can override the bower dependency somehow. Maybe something in this file can help us? https://github.com/ember-cli/ember-cli-blueprint-test-helpers/blob/master/lib/helpers/setup.js

rwjblue:
\<rant>
Consumers of ember-data do not care about Ember Data's internal stack trace for this error.  You are specifying in the message exactly who is throwing the error.  The stack is just needlessly noisey and will drown out the actual message.
\</rant>

rwjblue:
It seems to me that we need more tailored messages.  I see at least three scenarios that all seem to deserve their own messaging:

* when `_forceBowerUsage` is true AND the bower version of ember-data is `^2.3.0-beta.5` then you must have `ember-cli-shims` at `>= 0.1.0`. We should suggest to folks that ember-data@2.3.x should be consumed from `npm` instead of `bower.json` and that they should remove `ember-data` from `bower.json` (this is a warning, not an error condition).
* when `_forceBowerUsage` is true AND the bower version of ember-data is `< 2.3.0-beta.3` (or whenever the shims were introduced to the ember-data bower build) then you must have `ember-cli-shims` at `< 0.1.0` (otherwise the various `ember-data` modules are not found).  
* when `_forceBowerUsage` is false AND ember-cli-shims is `< 0.1.0`. We should throw a `SilentError` with a message telling them that they must use `ember-cli-shims@0.1.0`.

fivetanley:
> Consumers of ember-data do not care about Ember Data's internal stack trace for this error. You are specifying in the message exactly who is throwing the error. The stack is just needlessly noisey and will drown out the actual message.

In my testing, this dies very fast and the error message and stack are clear to see. It would be useful to have the stack later when people will inevitably open issues for this further down the line. If people choose to debug this themselves, having the stack is very useful. I needed it for Ember Pouch to try to get it to load for this version of Ember Data. I wouldn't have pursued it otherwise.

> I see at least three scenarios that all seem to deserve their own messaging

I agree, seems great.

rwjblue:
@fivetanley - Thanks for working on this.  I'm sorry that I didn't think all of this through when I added the  warning...

fivetanley:
I may skip the assertion when we're running node tests for now. There's not a way to get in the middle of the test helpers and install the correct version of ember-cli-shims before the test helpers run as far as I can tell.

fivetanley:
Once travis comes back green we are good to go. I tested each of the scenarios Rob outlined in a new ember app.

PR #4063, opened 2016-01-12
[BUGFIX release] don't load "app" code when loading globals files

fivetanley:
The code that gets compiled for the globals build is in a closure with
its own instance of `loader.js`, so the entries for ember-data don't end
up in ember-cli's instance of `loader.js`. The ember-data code does
include shims that end up in the app's instance of loader.js, so we
exclude the app code from running in this case.

### DIFF
--- a/index.js
+++ b/index.js
@@ -64,6 +64,17 @@ module.exports = {
     return path.join(__dirname, 'blueprints');
   },
 
+  treeForApp: function(dir) {
+    if (this._forceBowerUsage) {
+      // Fake an empty broccoli tree
+      return { inputTree: dir, rebuild: function() { return []; } };
+    }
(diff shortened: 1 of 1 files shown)


index.js
```diff
@@ -64,6 +64,15 @@ module.exports = {
...
   },
 
+  treeForApp: function(dir) {
+    if (this._forceBowerUsage) {
+      // Fake an empty broccoli tree
+      return { inputTree: dir, rebuild: function() { return []; } };
+    }
+
+    return this._super.treeForApp.apply(this, arguments);
+  }
```
rwjblue: missing comma here

bmac: :bird: :eyes:

PR #4064, opened 2016-01-12
[BUGFIX release] Use correct version number when used as an addon.

rwjblue:


fivetanley:
git-repo-version is still being used in other files, probably needs to be replaced with calculateVersion instead.

fivetanley:
This code (not the fault of this PR) is also appending `v` to versions, as in `v2.3.0`, making it show up as `v2.3.0` when it should show up as `2.3.0` in `Ember.libraries` and `DS.VERSION`. Might also be nice to stick that in `calculateVersion`

rwjblue:
Closed in favor of #4066

PR #4023, opened 2015-12-21
Enhance blueprints

pangratz:
This is a followup PR, made due to [a comment](https://github.com/emberjs/data/pull/4019#issuecomment-166396159) by @igorT: currently a serializer generate via `ember generate ember-data:serializer foo` extends `DS.JSONAPIAdapter`, where it should extend the `ApplicationAdapter`. This behavior is already the case when an adapter is generated, so this PR aligns the behavior of the serializer blueprint to the one of the adapter blueprint.

---

### TODO

- [x] Investigate timeout in test for `ember g adapter foo` when application adapter is present
- [x] Investigate how to check if a file exists within a blueprints' `index.js`
- [x] Check why test for serializer fail but for the adapter doesn't (though it's the same)

PR #4070, opened 2016-01-13
Update blueprints to import modules directly

HeroicEric:
## TL;DR;

This does two things:

- Adds more tests to the adapter & model blueprints
- Updates blueprints to import modules directly rather than access them through `DS`

I originally added the extra tests to cover the existing behavior before I changed it. If you'd rather have those split out into a separate PR LMK.

### Details

- Imports `JSONAPIAdapter` module directly in blueprint
- Imports `JSONAPISerializer` module directly in blueprint
- Imports `Transform` module directly in blueprint
- Imports `attr` module directly in model blueprints when needed
- Imports `belongsTo` and `hasMany` modules directly in blueprints when needed

### DIFF
--- a/blueprints/model/index.js
+++ b/blueprints/model/index.js
@@ -16,6 +16,10 @@ module.exports = {
     var attrs = [];
     var needs = [];
     var entityOptions = options.entity.options;
+    var importStatements = ['import Model from \'ember-data/model\';'];
+    var shouldImportAttr = false;
+    var shouldImportBelongsTo = false;
+    var shouldImportHasMany = false;
 
     for (var name in entityOptions) {
       var type = entityOptions[name] || '';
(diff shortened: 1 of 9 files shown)


pangratz:
FYI, I have an somewhat accompanying PR #4023 open, which would make your adapter test `adapter when is named "application"` obsolete.

The failing tests should be fixed once a new release for `ember-cli-blueprint-test-helpers` is cut, as pointed out by @trabus here https://github.com/emberjs/data/pull/4023#issuecomment-169152370.

This looks very good to me :+1: ! Nice change importing modules directly!

HeroicEric:
@pangratz turns out, the tests were failing because of a typo `importStatment !== importStatement` :smile: 

Would you like me to remove the `adapter when is named "application"` test?

pangratz:
No, you can leave it in I guess. This PR will likely be merged soon, so I will update the test in #4023 once this is the case.

blueprints/model/index.js
```diff
@@ -35,26 +39,44 @@ module.exports = {
...
     var needsDeduplicated = needs.filter(function(need, i) {
       return needs.indexOf(need) === i;
     });
 
+    if (shouldImportAttr) {
+      importStatements.push('import attr from \'ember-data/attr\';');
+    }
+
+    if (shouldImportBelongsTo && shouldImportHasMany) {
+      importStatements.push('import { belongsTo, hasMany } from \'ember-data/relationships\';');
```
bmac: Out of scope for this pr but I've been thinking it may be useful to add a path that allows users to import `belongsTo`, `hasMany` and `attr` all on the same line.

HeroicEric: :+1:

HeroicEric:
This should wait until https://github.com/emberjs/data/issues/4075 is resolved

HeroicEric:
^^ was fixed by https://github.com/emberjs/data/pull/4091

bmac:
Thanks @HeroicEric

PR #4074, opened 2016-01-13
Update README, now that ember-data is a proper addon

pangratz:


fivetanley:
We need to have instructions for both for the interim between now and 3.0.

pangratz:
Ok, will update...

pangratz:
@fivetanley better?

PR #4076, opened 2016-01-13
Remove duplicate exports in addon/-private/system/promise-proxies.js

jcope2013:
The export list was duped

### DIFF
--- a/addon/-private/system/promise-proxies.js
+++ b/addon/-private/system/promise-proxies.js
@@ -135,8 +135,5 @@ var promiseManyArray = function(promise, label) {
 export {
   PromiseArray,
   PromiseObject,
-  PromiseManyArray,
-  promiseArray,
-  promiseObject,
-  promiseManyArray
+  PromiseManyArray
 };


wecc:
`PromiseArray` and `promiseArray` are different exports, see [here](https://github.com/jcope2013/data/blob/fix-duplicate-exports-promise-proxies/addon/-private/system/promise-proxies.js#L36) and [here](https://github.com/jcope2013/data/blob/fix-duplicate-exports-promise-proxies/addon/-private/system/promise-proxies.js#L75) (same goes for `*Object` and `*ManyArray`).

The lower camel case ones are helper methods to create the upper camel cased ones. A couple of examples where they are used: [here](https://github.com/emberjs/data/blob/c8f03a5c1c7b91160e6c3ad51b80638efd1b0413/addon/-private/system/store.js#L588) and [here](https://github.com/emberjs/data/blob/c8f03a5c1c7b91160e6c3ad51b80638efd1b0413/addon/-private/system/store.js#L113).

Issue #2905, opened 2015-03-19
[Meta Issue] JSON-API support


wecc:
Let's use this issue to keep track of status for the JSON-API adapter.
- [x] Basic working adapter - #2904

**Improve metadata handling**
- [ ] single records
- [x] hasMany (#3221)
- [ ] findAll
- [X] query

**Features**
- [ ] Pagination (collections, hasMany relationships)
- [ ] Filtering
- [x] Includes

**Links**
- [ ] Support `self` links for relationships (reloading, saving, deleting)
- [ ] Support `self` links for records (to override `urlFor`-methods)
## 

Ping @kurko

kurko:
:clap: :smile:

btecu:
I thought the RESTAdapter was going to follow JSON AP, which currently does for the most part.
So we're going to have a new different adapter?

Panman8201:
Where is this at? I'm working on my backend API and wondering if I should follow the JSON-API spec or just follow the RESTAdapter for now. Thanks!

IanVS:
+1, What is the status of JSON-API in Ember Data?  According to jsonapi.org:

> JSON API is at a third release candidate state. This means that it is not yet stable, however, libraries intended to work with 1.0 should implement this version of the specification. We may make very small tweaks, but everything is basically in place.

seawatts:
May 21st 2015 is the release for json-api 1.0
https://twitter.com/dgeb/status/588179514817130497

IanVS:
@seawatts That's great, but it's only about two weeks from now.  Will Ember Data be ready to use with a json-api compliant API at that time?  What's left to be done?  This meta-issue doesn't seem to have been updated in quite some time.  Does that mean no progress has been made, or the progress has just been silent?

bmac:
There is work in progress on making Ember Data json-api compliant in this branch: https://github.com/emberjs/data/pull/2904

kottenator:
Hi! 

I'm new to Ember.js and I've found out in official documentation that `ember-data` supports JSON API. 

I've created some mock data using JSON API v1.0 but `RESTAdapter` doesn't parse top-level "data" property into an array - it throws an error instead.

Could you help me? Am I doing something wrong?

WooDzu:
RESTAdapter is here to deal with simpler resources like:
http://guides.emberjs.com/v1.12.0/models/the-rest-adapter/

[...]

PR #4077, opened 2016-01-14
[CLEANUP beta] Change the way metadata in response is stored on record array

danmcclain:
The previous implementation of inserting metadata from an ajax response
onto a record array used the store as temporary storage for the metadata
via `store._metadataFor` and `store._setMetadataFor`. Although rare, this could
store the wrong value on the record array due to a race condition, as
the metadata was stored based on the query's model's name. This moves
the storing of metadata to `AdapterPopulatedRecordArray.loadRecords`
since, at the time that that method is called, the payload is present,
so we don't have to use `store.{_setMetadataFor,_metadataFor}`.
`store._metadataFor` was only used in
`AdapterPopulatedRecordArray.loadRecords`, so it is no longer needed
with this new implementation.

### DIFF
--- a/addon/-private/system/record-arrays/adapter-populated-record-array.js
+++ b/addon/-private/system/record-arrays/adapter-populated-record-array.js
@@ -29,20 +29,16 @@ export default RecordArray.extend({
   /**
     @method loadRecords
     @param {Array} records
+    @param {Object} payload normalized payload
     @private
   */
-  loadRecords(records) {
-    var store = get(this, 'store');
-    var type = get(this, 'type');
-    var modelName = type.modelName;
-    var meta = store._metadataFor(modelName);
-
+  loadRecords(records, payload) {
     //TODO Optimize
     var internalModels = Ember.A(records).mapBy('_internalModel');
     this.setProperties({
       content: Ember.A(internalModels),
       isLoaded: true,
-      meta: cloneNull(meta)
+      meta: cloneNull(payload.meta)
     });
 
     internalModels.forEach((record) => {
--- a/addon/-private/system/store.js
+++ b/addon/-private/system/store.js
@@ -1253,30 +1253,6 @@ Store = Service.extend({
     return this.hasRecordForId(modelName, id);
   },
 
-  /**
-    @method _metadataFor
-    @param {String} modelName
-    @return {object}
-    @private
-  */
-  _metadataFor(modelName) {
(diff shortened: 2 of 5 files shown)


bmac:
Looks good to me. @wecc can your also review this pr?

danmcclain:
I am planning on opening an additional PR to add `links` to the record array like `meta`, do you want me to bring it into this PR? Or keep it separate?

bmac:
I'd say do it in a separate pr.

danmcclain:
Ok, I'll wait for this PR to be accepted first, as it will tie into the same `loadRecords` function :+1:

wecc:
Thanks for the PR @danmcclain!

It's nice to see `_metadataFor` and `_setMetadataFor` finally go :) I'm trying to remember why we didn't do this for the 1.13 release/New Serializer API refactor but I'm struggling. I think it was along the lines that we didn't want _both_ records the payload to "leak" to the array. I'm inclined to suggest to just put meta on the array like we do [here](https://github.com/emberjs/data/blob/v2.3.1/addon/-private/system/store/finders.js#L87) but I can see how that might make continued JSON API compliance more verbose.

I'm not gonna block this due to :arrow_up:, the code is private and easily changed if needed. It's very nice to get rid of `_metadataFor` and `_setMetadataFor` acting like a proxy, I'm :+1: 

Can you prefix the commit with `[CLEANUP beta]`? (not really a `BUGFIX` imo)

danmcclain:
@wecc Updated!

I was thinking the same (just passing `payload.meta`), but I am actually adding `links` support on one of our projects and grabbed `links` off of payload as well. When I open the `links` PR, I'll discuss the approach I've taken for our project

Thanks

danmcclain:
I take that back, I never pushed it until now. Updated for real @wecc

amk221:
I was using @vanadar's [temporary solution](https://github.com/emberjs/data/issues/2905#issuecomment-179813131) to get meta for a single record. Since this PR, is there a similar solution?

duizendnegen:
Similar issue, the workaround in https://github.com/emberjs/data/issues/4115 now no longer works.

amk221:
@duizendnegen hack:

base-model.js
```
_meta: computed(function() {
  return this._internalModel.type.__meta__;
}).volatile()
```
application-serializer.js
```
extractMeta(store, typeClass) {
  let meta = this._super(...arguments);
  typeClass.__meta__ = meta;
  return meta;
},
```

PR #4078, opened 2016-01-14
Use `Ember.merge` instead of ember-data's shim

tricknotes:
This shim was introduced to be compatible with old Ember
that couldn't merge object that has no `hasOwnProperty` method.
https://github.com/emberjs/data/pull/2506

However current Ember Data supports Ember >= 2 (it can merge objects
safely since this change. 751f76dac )
https://github.com/emberjs/ember.js/pull/5645

So now this shim is unnecessary.


wecc:
LGTM :+1:

@igorT, @fivetanley ?

stefanpenner:
LGTM

Issue #4075, opened 2016-01-13
import Model from 'ember-data/model' doesn't work


bmac:
Unless you also import `ember-data`.

rwjblue [2:20 PM] 
basically there are some side effects that happen upon importing `ember-data`

​[2:21] 
yep

​[2:21] 
https://github.com/emberjs/data/blob/a0408cc1f1ec4004368e6c553f219b8360ec6bb3/addon/-private/system/relationships/ext.js#L101 doesn't get ran if you `import Model from 'ember-data/model';`

GitHub
emberjs/data
A data persistence library for Ember.js.
(edited)

​[2:21] 
@bmac: thoughts on ^^?

​[2:21] 
@atimberlake: add `import "ember-data";` to `app/app.js` for now to fix your app code

atimberlake [2:22 PM] 
@rwjblue: Thanks, will do

rwjblue [2:22 PM] 
I wonder if there is a specific reason we `.reopen` there instead of provide a mixin that `Model` can include

atimberlake [2:23 PM] 
Thanks, all working again

bmac [3:23 PM] 
@rwjblue: sounds like model should be importing `relationships/ext`

rwjblue [3:24 PM] 
@bmac: it should import a mixin from there(edited)

​[3:24] 
but if model imports it for side effects, then you have a circular dep and `Model` will be undefined in relationships/ext

bmac [3:31 PM] 
makes sense

bmac:
Closed by https://github.com/emberjs/data/pull/4091

MattNguyen:
After upgrading to 2.3.3, we still need to add `import ember-data` in `app/app.js`. #4091 was supposed to resolve this, correct?

bmac:
@MattNguyen That is correct what error are you seeing?

PR #4081, opened 2016-01-15
Move the Model.reopen from relationships/ext.js into model/model to m…

bmac:
…ake it easier to import model

@igorT is there a reason the relationship stuff lives in `relationships/ext` vs where the model class is defined.


HeroicEric:
Just adding that this is for https://github.com/emberjs/data/issues/4075

pangratz:
@bmac I think the `reopen`ings in [`addon/attr.js`](https://github.com/emberjs/data/blob/9fefef6917530c7df5e148950eec8425684e5162/addon/attr.js) should be part of this as well, right?

bmac:
Closing as https://github.com/emberjs/data/pull/4091 is a better solution for this issue.

PR #4082, opened 2016-01-15
Move the date import to where it is used intead of the root module

bmac:


pangratz:
I wonder if `Ember.Date` should be removed as a whole an only the `parse` part should be imported? `Ember.Date` shows up in the API doc, but it only has the `parse` function, and doesn't extend JS `Date`. 

I am not sure what the correct way would be though, semver-wise. Maybe deprecate it for `3.0`?

rwjblue:
We can provide a public module for date parsing, I know some users rely on `Ember.Date.parse` existing...

bmac:
I opened an issue to track the Date.parse comments outside of this pr.

pangratz:
> I opened an issue to track the Date.parse comments outside of this pr.

:+1:

PR #4084, opened 2016-01-15
[cleanup] Remove unnecessary inline JSHint config

HeroicEric:
- `Ember` is imported below so we don’t need the global :smile:
- The `eqnull:true` rule was originally added for
https://github.com/emberjs/data/commit/beb2f26 but was later added to the `.jshintrc` in
https://github.com/emberjs/data/commit/caa2753

### DIFF
--- a/addon/-private/system/store.js
+++ b/addon/-private/system/store.js
@@ -1,6 +1,3 @@
-/*globals Ember*/
-/*jshint eqnull:true*/
-
 /**
   @module ember-data
 */


fivetanley:
:pizza:

fivetanley:
can you prefix the commit message with `[cleanup]`, please?

HeroicEric:
This should also fix the test failures in https://github.com/emberjs/data/pull/4082 & https://github.com/emberjs/data/pull/4081

bmac:
Thanks @HeroicEric

PR #4086, opened 2016-01-18
[FEATURE ds-transform-pass-options] pass options to DS.Transform

pangratz:
This feature passes the options hash defined for an attribute to the
`serialize` and `deserialize` methods of `DS.Transform`:

```js
// app/models/post.js
export default Model.extend({
  text: DS.attr('text'),

  otherText: DS.attr('text', {
    lowercaseInput: true,
    uppercaseOutput: true
  })
});

// app/transforms/text.js
export default DS.Tranform.extend({
  serialize(value, options) {
    if (options.uppercaseOutput) {
      return value.toUpperCase();
    }

    return value;
  },

  deserialize(value, options) {
    if (options.lowercaseInput) {
      return value.toLowerCase();
    }

    return value;
  }
});
```

---

This implements emberjs/rfcs#1.

### DIFF
--- a/FEATURES.md
+++ b/FEATURES.md
@@ -20,3 +20,8 @@ entry in `config/features.json`.
 - `ds-references`
 
   Adds references as described in [RFC 57](https://github.com/emberjs/rfcs/pull/57)
+
+- `ds-transform-pass-options`
+
+  Pass options specified for a `DS.attr` to the `DS.Tranform`'s `serialize` and
+  `deserialize` methods (described in [RFC 1](https://github.com/emberjs/rfcs/pull/1))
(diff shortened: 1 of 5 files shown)


knownasilya:
@pangratz you beat me to it :+1:

josemarluedke:
Hey @pangratz! I need to use this feature with json api instead of just the json adapter. Does this feature was supposed to be in json api as well? Or it was specifically done just for json?

knownasilya:
This feature should be adapter agnostic.

pangratz:
@josemarluedke this basically should be an adapter agnostic change, but it looks like you found a bug. Thanks for fixing in #4425 🚀  !

PR #4088, opened 2016-01-18
[BUFGIX release] Fix regression with missing initializers

### LINKED ISSUE #4087: Ember Data 2.3.1 is missing initializers when loaded with Ember CLI
See https://github.com/Vestorly/torii/issues/261

https://github.com/emberjs/data/blob/master/tests/ember-data-initializers.js#L18-L40

wecc:
This PR adds missing initializers to keep backwards compatibility.

Fixes #4087

### DIFF
--- a/app/initializers/data-adapter.js
+++ b/app/initializers/data-adapter.js
@@ -1,0 +1,14 @@
+import Ember from 'ember';
+
+/*
+  This initializer is here to keep backwards compatibility with code depending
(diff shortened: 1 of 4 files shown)


bmac:
Thanks @wecc

rwjblue:
I wonder if there is a way to deprecate depending on these.  In theory, we could hook into the DAG map that is used for initializers and see if anyone depends on it?

PR #4092, opened 2016-01-18
Delete TRANSITION.md

rwjblue:
This is pretty out of date, and isn't a good resource any longer.  If folks need the document for their version, they can view it via browsing the tag itself.


bmac:
Thanks @rwjblue

PR #4095, opened 2016-01-19
[BUGFIX release] Remove 'v' prefix from `DS.VERSION`

tricknotes:
The change of version number format may break version checking from other libs.

Before:
``` js
DS.VERSION //=> "v2.3.2"
```

After:
``` js
DS.VERSION //=> "2.3.2"
```

### DIFF
--- a/lib/calculate-version.js
+++ b/lib/calculate-version.js
@@ -9,7 +9,7 @@ module.exports = function() {
   if (existsSync(gitPath)) {
     var info = gitRepoInfo(gitPath);
     if (info.tag) {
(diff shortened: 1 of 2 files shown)


bmac:
cc @fivetanley

rwjblue:
Seems good to me.

bmac:
@tricknotes I've published [v2.3.3](https://github.com/emberjs/data/releases/tag/v2.3.3) with this fix.

tricknotes:
Thanks! It works fine for me.

PR #4097, opened 2016-01-19
Update `.npmignore`

tricknotes:
* Ruby files and testing yaml is unnecessary as a npm module.
* Now `Brocfile.js` is not exist anywhere.


bmac:
Thanks @tricknotes

PR #4100, opened 2016-01-20
[DOC] Replace reference to DS.Store.find

HeroicEric:
Replaces a reference to `DS.Store.find` in the docs for
`DS.Store.filter`


HeroicEric:
Should `DS.Store.filter` be marked private since it was moved to an addon?

bmac:
Thanks @HeroicEric. Making `filter` private would be great.

PR #4101, opened 2016-01-20
[cleanup] Remove unneeded `getComputedPropertyDesc` test helper

HeroicEric:
As the comment in the code says, this is no longer needed since
https://github.com/emberjs/ember.js/pull/10323 was merged in.


bmac:
Thanks @HeroicEric.

PR #4111, opened 2016-01-22
[BUGFIX beta] Add ducktyping of `AdapterError`

danmcclain:
There are `instanceof` checks for `AdapterError`; this removes it in favor of checking for a property on the error. `isAdapterError` appears on all errors that use `AdapterError` as the base.

This PR is looking to avoid the need to import private errors in fastboot to keep the same code flow in the ajax method used in node land, related PR: https://github.com/tildeio/ember-cli-fastboot/pull/95

### DIFF
--- a/addon/adapters/rest.js
+++ b/addon/adapters/rest.js
@@ -853,7 +853,7 @@ export default Adapter.extend(BuildURLMixin, {
           requestData
         );
 
-        if (response instanceof AdapterError) {
+        if (response && response.isAdapterError) {
           Ember.run.join(null, reject, response);
         } else {
           Ember.run.join(null, resolve, response);
--- a/addon/-private/adapters/errors.js
+++ b/addon/-private/adapters/errors.js
@@ -12,6 +12,7 @@ const PRIMARY_ATTRIBUTE_KEY = 'base';
   @namespace DS
 */
 export function AdapterError(errors, message = 'Adapter operation failed') {
+  this.isAdapterError = true;
   EmberError.call(this, message);
 
   this.errors = errors || [
--- a/tests/unit/adapter-errors-test.js
+++ b/tests/unit/adapter-errors-test.js
@@ -10,24 +10,28 @@ test("DS.AdapterError", function(assert) {
   var error = new DS.AdapterError();
   assert.ok(error instanceof Error);
   assert.ok(error instanceof Ember.Error);
+  assert.ok(error.isAdapterError);
 });
 
 test("DS.InvalidError", function(assert) {
   var error = new DS.InvalidError();
   assert.ok(error instanceof Error);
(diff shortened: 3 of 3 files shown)


danmcclain:
Tests were failing locally even when reverting so I figured I'd try pushing anyway. Will dig back in tomorrow morning. Would like feedback on direction as well

/cc @rwjblue

rwjblue:
Using flags seems better than using instance of to me...

danmcclain:
Slightly confusing, there are 2 different definitions for `InvalidError`, changed this so it only handles `AdapterError` which is the one that fastboot is checking for ultimately

bmac:
> there are 2 different definitions for `InvalidError`

@tchak is this an oversite or do we need 2 definitions for `InvalidError`?

bmac:
@danmcclain Is this something that should be considered a `[BUGFIX beta]` so it can be included in the 2.4 release?

danmcclain:
@bmac Here are the 2 different definitions:
Adapter InvalidError: https://github.com/emberjs/data/blob/209326f22324d8e29eee0207908d02725f7cb83e/addon/-private/adapters/errors.js#L85-L90

Model InvalidError:
https://github.com/emberjs/data/blob/209326f22324d8e29eee0207908d02725f7cb83e/addon/-private/system/model/errors/invalid.js

Also, I can label as such, I didn't give it a label, as I was unsure what you would want it to be

danmcclain:
@bmac I think `[BUGFIX beta]` would apply. Labeling as such now

tchak:
I am not a big fan of this. I started some work to make it easy to inherit from `AdapterError` which seems to me like a better practice. But I do understand that it makes life easier in some cases, so I guess it's :+1:

The two definitions of `InvalidError` is definitely a bug/oversight.

rwjblue:
@tchak Using `instanceof` has generally been problematic for us in Ember, so we have been switching to using flags like this.

addon/adapters/rest.js
```diff
@@ -853,7 +853,7 @@ export default Adapter.extend(BuildURLMixin, {
           requestData
         );
 
-        if (response instanceof AdapterError) {
+        if (response.isAdapterError) {
```
rwjblue: Should likely be `response && response.isAdapterError` in case `response` is undefined / null.

bmac:
@danmcclain @rwjblue added one final comment. Once that is fixed this pr should be ready to be merged.

danmcclain:
@bmac will update shortly 👍🏼

danmcclain:
@bmac updated per @rwjblue's comment

bmac:
Thanks @danmcclain

tchak:
@rwjblue ok. I see quite well why `instanceof` can be problematic. It just makes me very sad... But such is life :)

PR #4116, opened 2016-01-26
Add assertions for store#query()

pangratz:
Currently invoking `store.query('person')` will fail with an error
within `RESTAdapter#sortQueryParams`, which is not helpful.

This adds the following assertions to store#query():

- a type is passed to the query method
- a query hash is passed to the query method

---

Those two assertions are available on `store#queryRecord` so this PR aligns `store#query` so it has the same assertions.


bmac:
Want to tag this `[BUGFIX beta]`?

PR #4119, opened 2016-01-26
Add ember.js and ember-data tags automatically to SO question

pangratz:
This prefills the tags with `ember.js` and `ember-data`.


bmac:
Thanks @pangratz

PR #4120, opened 2016-01-26
Return the date as it is when it's not a date object

GCorbel:


fivetanley:
What is the bug this is fixing?

GCorbel:
I have this model : 

      birthDate: DS.attr('date')
    });

and `$E.set('birthDate', '2015-01-01'); $E.toJSON().birthDate` gives `null`. I think it's more useful to have the date have the string.

tests/unit/transform/date-test.js
```diff
@@ -17,6 +17,8 @@ test("#serialize", function(assert) {
   assert.equal(transform.serialize(undefined), null);
```
pangratz: Hmm, I would have expected that this assertion fails with the change in this PR (it doesn't because `assert.equal` uses non strict comparison).

I am wondering if `assert.deepEqual` is better suited here?

GCorbel: It works, I think another PR have to be created for this modification.

bmac:
When using `attr('date')` its usually assumed the value of the property is always a JavaScript date object or null. Its seems like this change will implicitly allow a `attr('date')` property to also be a string which may make it harder to reason about the state of `attr('date')` and add additional complexity onto consumers of the property as they may now need to account for sting values.

quaertym:
@GCorbel Why don't you use `attr('string')`?

GCorbel:
@bmac and @quaertym it seems your right. `date` is the more obvious format for thinks related to date and I need to change it with input tags. I think I can use `string` but to return the date is it is correct but not mandatory. As you which, you can close this PR if you want.

bmac:
Hi @GCorbel we discussed this in the Ember Data meeting today and decided the date transform should not include this behavior. 

@igorT suggested that the date transform should warn when returning null because the `date` is not an instance of a Date object.

PR #4125, opened 2016-01-28
[BUGFIX beta] Export more public API's via modules

pangratz:
This change allows to import more public API's via modules:

- `import Serializer from "ember-data/serializer"`
- `import EmbeddedRecords from "ember-data/serializers/embedded-records-mixin"`

---

Should this be `[BUGFIX release]`?

cc @fivetanley @bmac


rwjblue:
We started with these alias type of modules for the early 2.3.0-betas, but ended up having issues and changing things around to move the actual module to the new public location (instead of shim modules).

IMHO, we should be consistent in our public modules.  So either they existing ones should be made into shims, or these should be moving the underlying modules and avoid shims...

fivetanley:
I am all for avoiding shims.

bmac:
Are `BooleanTransform` and friends typically used in app code in a way where they would need to be imported?

bmac:
I don't think this needs to be a `[BUGFIX release]` unless these paths were previously supported by https://github.com/ember-cli/ember-cli-shims and their omission is breaking existing applications.

pangratz:
> Are BooleanTransform and friends typically used in app code in a way where they would need to be imported?

You are right, they might not be needed as public API. I removed the public API modules for the transforms.

addon/serializer.js
```diff
@@ -0,0 +1,3 @@
+import Serializer from "ember-data/-private/system/serializer";
+
+export default Serializer;
```
bmac: @pangratz could you move the `Serializer` and `EmbeddedRecordsMixin` out of the `-private` folder and into their new public locations.

addon/mixins/embedded-records.js
```diff
@@ -0,0 +1,3 @@
+import EmbeddedRecordsMixin from "ember-data/-private/serializers/embedded-records-mixin";
+
+export default EmbeddedRecordsMixin;
```
bmac: The existing path is `serializers/embedded-records-mixin`. Any one have any preference if we should keep it in the `serializers` folder when we expose it publicly since this mixin is coupled to the existing serializers?

pangratz:
I moved the files from the private location to the public ones.

Also, the embedded records mixin is now located at `ember-data/serializers/embedded-records-mixin`. As pointed out by @bmac, I think this path might be better since it clarifies that the mixin should be used in combination with a serializer...

tests/unit/modules-test.js
```diff
@@ -0,0 +1,74 @@
...
+  assert.ok(JSONAPISerializer);
+});
+
+test("ember-data/serializers/json", function(assert) {
+  assert.ok(JSONSerializer);
+});
+
+test("ember-data/serializers/rest", function(assert) {
+  assert.ok(RESTSerializer);
+});
```
bmac: Yay Tests!

bmac:
:+1: looks good to me. Just needs a rebase.

pangratz:
@bmac done

PR #4128, opened 2016-01-29
Fix calls to keyForRelationship in embedded records

aldhsu:
keyForRelationship arguments are(key, typeClass, method)


bmac:
Thanks @aldhsu

PR #4130, opened 2016-02-02
[BUGFIX beta] Add sha suffix to `DS.VERSION` for unreleased version

tricknotes:
When repository URL is specified for ember-data in `package.json`, `DS.VERSION` should have git sha as suffix.
```
# package.json
"ember-data": "git@github.com:emberjs/data.git"
```

Before:

After:


bmac:
Thanks @tricknotes

PR #4135, opened 2016-02-05
import require

### LINKED ISSUE #4134: relationships/has-many.js doesn't compile with Babel 6
https://github.com/emberjs/data/blob/master/addon/-private/system/relationships/has-many.js

The issue appears to be because of the require in the middle of the file..

Suggested solution: /* globals require */

Will test this and possibly open a PR over the coming days

jmurphyau:
The global require in this file causes an error when using babel 6 with the transform-es2015-modules-amd plugin

fixes #4134

### DIFF
--- a/addon/-private/system/relationships/has-many.js
+++ b/addon/-private/system/relationships/has-many.js
@@ -2,6 +2,7 @@
   @module ember-data
 */
 
+
 import Ember from 'ember';
 import { assert } from "ember-data/-private/debug";
 import normalizeModelName from "ember-data/-private/system/normalize-model-name";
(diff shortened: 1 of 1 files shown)


stefanpenner:
i believe you should be able to actually import require
```
import require from 'require'
```

jmurphyau:
I tried that originally and travis tests failed so switched to the way it is now

fivetanley:
Is there a reason we can't just `import model from 'ember-data/model';` instead?

bmac:
Originally when this code was added `import Model from 'ember-data/model';` resulted in a circular reference which was breaking the tests.

pangratz:
Just an idea since the `Model` is only needed for the assertion, what about a duck typing approach of checking if `record.hasOwnProperty('_internalModel') === true` or something?

bmac:
Thanks @jmurphyau.

sly7-7:
I don't know if this matters here, but it seems to me we should avoid using hasOwnProperty() because it was slow. Maybe here it's not a hot path, so I guess it's ok.

pangratz:
Since this will be stripped in production builds I think this is ok...

sly7-7:
You're completely right

Issue #4122, opened 2016-01-28
parseResponseHeaders does not follow HTTP spec of optional spaces


varoot:
According to [HTTP spec](https://tools.ietf.org/html/rfc7230#section-3.2), a header field has **optional** leading whitespace. Currently, [parseResponseHeaders](https://github.com/emberjs/data/blob/master/addon/adapters/rest.js#L1009) requires a space character. It does not allow a TAB character, or no whitespace at all.

Also, we should trim the value since, according to the spec, the header values cannot have whitespace on either end.

PR #4137, opened 2016-02-06
[BUGFIX beta] Allow optional spaces when parsing response headers

HeroicEric:
Fixes https://github.com/emberjs/data/issues/4122

- Allows header field-value without preceding whitespace
- Removes leading/trailing whitespace from header field-names
- Removes leading/trailing whitespace from header field-values
- Moves `parseResponseHeaders` function out of `DS.RestAdapter` to its own module

The main reason for moving to a module was to make it easier to unit test in order to document current behavior before making the fix. Let me know if this is undesirable.

### DIFF
--- a/addon/-private/utils/parse-response-headers.js
+++ b/addon/-private/utils/parse-response-headers.js
@@ -1,0 +1,26 @@
+import EmptyObject from 'ember-data/-private/system/empty-object';
+
+const CLRF = '\u000d\u000a';
+
+export default function parseResponseHeaders(headersString) {
+  let headers = new EmptyObject();
+
+  if (!headersString) {
+    return headers;
+  }
+
+  let headerPairs = headersString.split(CLRF);
+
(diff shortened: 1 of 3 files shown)


addon/-private/utils/parse-response-headers.js
```diff
@@ -0,0 +1,26 @@
...
+  let headers = new EmptyObject();
+
+  if (!headersString) {
+    return headers;
+  }
+
+  let headerPairs = headersString.split(CLRF);
+
+  headerPairs.forEach((header) => {
+    let [field, ...value] = header.split(':');
```
pangratz: :+1: ES6 :heart_eyes:

pangratz:
This looks very good to me :+1: . Much more readable now! Thanks @HeroicEric!

bmac:
Thanks @HeroicEric. You rock!

PR #4139, opened 2016-02-08
[BUGFIX beta] strip assertPolymorphicType in production builds

pangratz:
This assert statement checks if a record is of the expected polymporhic
type. Currently, this function is located in -private/utils and it is
not stripped from production builds. Since this function only contains
an `assert` (which gets stripped correctly), having this function in a
production build is not a severe issue.

Stripping it in production makes sense though since hereby we can get
rid of superfluous calls and it minifies the build size :hammer:


bmac:
:hammer:

PR #4141, opened 2016-02-09
[DOC] Document behavior of object level errors with JSON API

adam-knights:
This documents behavior added in https://github.com/emberjs/data/pull/3909.

### DIFF
--- a/addon/-private/system/model/errors.js
+++ b/addon/-private/system/model/errors.js
@@ -96,6 +96,33 @@ var MapWithDefault = Ember.MapWithDefault;
   {{/each}}
   ```
 
+  The JSON API spec also allows for object level errors to be placed
+  in an object with pointer `data`.
+
+  ```javascript
+  {
+    "errors": [
+      {
+        "detail": "Some generic non property error message",
+        "source": {
+          "pointer": "data"
+        }
+      }
+    ]
+  }
+  ```
+
+  You can access these errors by using the `base` property on the errors
+  object.
+
+  ```handlebars
+  {{#each model.errors.base as |error|}}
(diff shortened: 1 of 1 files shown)


addon/-private/system/model/errors.js
```diff
@@ -95,7 +95,35 @@ var MapWithDefault = Ember.MapWithDefault;
     </div>
   {{/each}}
   ```
-
+  
+  The JSON API spec also allows for object level errors to be placed
+  in an object with pointer `data`
+
```
bmac: Looks like travis didn't like this pr because of extra trailing whitespace on some of these blank lines. @adam-knights do you mind amending this commit and removing the whitespace?

addon/-private/system/model/errors.js
```diff
@@ -96,6 +96,34 @@ var MapWithDefault = Ember.MapWithDefault;
...
+      ...
+	  {
+        "detail": "Some generic non property error message",
+	      "source": {
+		    "pointer": "data"
+	      }
+	  }
+	]
+  }
+  ```
```
pangratz: I am not sure if this is necessary or if it shows up correctly nevertheless in the API docs, but can you reformat this so the indentation is correct for this code sample? Also, I think the `...` might interfere with the API docs since it's not a valid JS syntax, so maybe they should be removed? What do you think?

pangratz:
Besides my tiny comments this looks good to me. Thank you very much @adam-knights!

adam-knights:
np, thanks @pangratz @bmac, those changes are now made.

pangratz:
Thanks @adam-knights!

PR #4147, opened 2016-02-10
Fix key remapping for embedded belongsTo

### LINKED ISSUE #4068: EmbeddedRecordsMixin does not seem to respect `key` values in `attrs` hash
While using `DS.EmbeddedRecordsMixin`, I discovered this while trying to embed records and also map a key for said records:
```
attrs: {
  steps: { key: 'process_steps', embedded: 'always'}
}
```
I had assumed that i could still use the `attrs` hash to mapp the json key to model attribute.. however it seems to be ignored.

When the json payload is deserialized into a model, `steps` seems to be resolved correctly.. but when attempting, but when doing a POST/PUT.. serializing the model back to json.. they key remains `steps` and not the intended `process_steps`

I have attempted workarounds by overriding `serializeHasMany` and also `keyForRelationship .. which results in the correct payload key.. however at that point none of the actual embedded records get serialized at all?

Edit: Solution for now is to just not map "embedded record keys" and just use a `computed.alias`

I am wondering if anyone else has experienced or overcome this limitation?

ef4:
Closes #4068. This implements the same semantics for `key` that are used in JSONSerializer.

### DIFF
--- a/addon/-private/serializers/embedded-records-mixin.js
+++ b/addon/-private/serializers/embedded-records-mixin.js
@@ -217,7 +217,11 @@ export default Ember.Mixin.create({
 
   _serializeEmbeddedBelongsTo(snapshot, json, relationship) {
     let embeddedSnapshot = snapshot.belongsTo(relationship.key);
-    let serializedKey = this.keyForRelationship(relationship.key, relationship.kind, 'serialize');
+    var serializedKey = this._getMappedKey(relationship.key, snapshot.type);
+    if (serializedKey === relationship.key && this.keyForRelationship) {
+      serializedKey = this.keyForRelationship(relationship.key, relationship.kind, "serialize");
+    }
+
     if (!embeddedSnapshot) {
       json[serializedKey] = null;
     } else {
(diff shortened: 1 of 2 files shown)


bmac:
Does the same change need to happen in `_serializeEmbeddedHasMany` as well?

bmac:
Also @ef4 do you mind amending the commit message to start with `[BUGFIX beta]`?

grapho:
@ef4 @bmac: I would think yes.. it needs to be included in `_serializeEmbeddedHasMany` as well... i primarily use embedded hasMany and that is the area in which i am having the issue.

ef4:
I fixed hasMany too, and amended the commit message.

bmac:
Thanks @ef4

grapho:
@bmac @ef4  thanks for the effort to fix this!

Quick question:  since this thing is now `merged`.. that meant the fix will be available in the next release? or can i just reinstall my ember data and the fix will be there?

fivetanley:
@grapho This will be in the next release. Otherwise, you can pin to the commit in `package.json` if you are using ember-cli post 2.3.0, or `#canary` in bower.json

grapho:
thanks @fivetanley!

PR #4153, opened 2016-02-11
calling reload multiple times on a has many triggers only one request

### LINKED ISSUE #2327: Calling reload on hasMany with a link the second time, sends a new request even if the first one hasn't returned

sly7-7:
Closes #2327

cc @igorT @wecc

### DIFF
--- a/addon/-private/system/relationships/state/has-many.js
+++ b/addon/-private/system/relationships/state/has-many.js
@@ -1,5 +1,5 @@
 import { assert } from "ember-data/-private/debug";
-import { PromiseManyArray } from "ember-data/-private/system/promise-proxies";
+import { PromiseManyArray, promiseManyArray } from "ember-data/-private/system/promise-proxies";
 import Relationship from "ember-data/-private/system/relationships/state/relationship";
 import OrderedSet from "ember-data/-private/system/ordered-set";
 import ManyArray from "ember-data/-private/system/many-array";
@@ -99,7 +99,6 @@ ManyRelationship.prototype.notifyRecordRelationshipAdded = function(record, idx)
 };
 
 ManyRelationship.prototype.reload = function() {
-  var self = this;
   var manyArrayLoadedState = this.manyArray.get('isLoaded');
 
   if (this._loadingPromise) {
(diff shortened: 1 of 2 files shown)


tests/integration/relationships/has-many-test.js
```diff
@@ -627,6 +627,43 @@ test("A hasMany relationship can be directly reloaded if it was fetched via link
...
+  env.adapter.findHasMany = function(store, record, link, relationship) {
+    count++;
+    return new Ember.RSVP.Promise(function(resolve, reject) {
+      setTimeout(function() {
+        var value = [
+          { id: 1, body: "First" },
+          { id: 2, body: "Second" }
+        ];
+        resolve(value);
+      }, 100);
```
bmac: Lets drop the setTimeout and instead return: 

```js
return Ember.RSVP.resolve([
           { id: 1, body: "First" },
           { id: 2, body: "Second" }
       ]);
```

addon/-private/system/relationships/state/has-many.js
```diff
@@ -112,13 +111,13 @@ ManyRelationship.prototype.reload = function() {
...
   } else {
-    return this.store.scheduleFetchMany(this.manyArray.toArray()).then(function() {
-      //Goes away after the manyArray refactor
-      self.manyArray.set('isLoaded', true);
-      return self.manyArray;
-    });
+    this._loadingPromise = promiseManyArray(this.store.scheduleFetchMany(this.manyArray.toArray()).then(() => {
+      return this.manyArray;
+    }), 'Reload with ids');
+    return this._loadingPromise;
```
bmac: @pangratz would you mind adding a test to ensure this else case is working correctly as well?

pangratz: Kindly delegating to @sly7-7 :wink:

bmac: Oops. Sorry about that.

sly7-7: :smile: No problem, I update this today.

tests/integration/relationships/has-many-test.js
sly7-7: I thought there must be a setTimeout in order to be "sure" the multiple reloads where not resolved. But for sure it's better if I can avoid that. Will do this :)

sly7-7:
@bmac Done. So I have to wrap the promises inside a promiseManyArray in order to be able to check the promise state. This seems to work, but I wonder if it's the right way even if I don't think we have an other way now to check the state of Promises

pangratz:
LGTM

bmac:
LGTM @wecc or @igorT do either of you have the bandwidth for a review?

wecc:
LGTM :+1:

bmac:
Thanks @sly7-7

PR #4154, opened 2016-02-11
Make sure new record are not pushed twice when parent is saved before.

olivierchatry:


steffenbrem:
@olivierchatry Does it also update the record again when new data is returned by the server (such as the id for example)?

olivierchatry:
yes, it should.

steffenbrem:
Nice, would be cool to see this in core :+1:

addon/-private/system/relationships/state/relationship.js
```diff
@@ -173,7 +173,9 @@ Relationship.prototype = {
     //TODO(Igor) make this less abysmally slow
     this.members = this.canonicalMembers.copy();
     for (i=0; i<newRecords.length; i++) {
-      this.members.add(newRecords[i]);
+      if (!this.members.has(newRecords[i])) {
+        this.members.add(newRecords[i]);
+      }
```
bmac: @olivierchatry did you run into an issue that required this if check? Under the hood `members` does a [similar check](https://github.com/emberjs/ember.js/blob/master/packages/ember-metal/lib/map.js#L113) which should be preventing the record from being added twice.

tests/integration/relationships/has-many-test.js
```diff
@@ -414,6 +414,64 @@ test("A hasMany updated link should not remove new children when the parent reco
...
+                data: [
+                  {
+                    id: 5,
+                    type: "comment"
+                  }
+                ]
+              }
+            }
+          }
+        });
```
bmac: @olivierchatry is the `store.push` here important to this test or can we move the info about the `comments` relationship into the return form the stubbed `createRecord` method?

olivierchatry: No it is important it showcase the bug.

addon/-private/system/relationships/state/relationship.js
olivierchatry: Yes, using firebase, record get duplicate, see test.

bmac: When I remove this `if` check all of the tests still pass.

tests/integration/relationships/has-many-test.js
bmac: Seems good. :+1:

olivierchatry: The fixes is in two seperate files, but maybe we only need to do it in one ?

olivierchatry:
So, what do I need to do to make it OK ?

bmac:
3 things:
1. I'd like to get @igorT to review this pr.
2. I think the change in `addon/-private/system/relationships/state/relationship.js` can be reverted.
3. Would you mind squashing this pr down to 1 commit and prefixing the commit message with `[BUGFIX beta]` so it can be pulled into the 2.4 beta branch.

olivierchatry:
I will revert the relationship one as it is redundant. How do you want me
to squash down the PR ? create a new one ?

bmac:
@olivierchatry you can rebase your branch locally and then force push it to github and it will update this pr.

Here is a good short tutorial on how to perform a rebase locally: https://egghead.io/lessons/javascript-how-to-squash-multiple-git-commits

olivierchatry:
Am turning a bit crazy, wondering if it would not be easier to create a new branch and a new pull request from it.

steffenbrem:
@olivierchatry Do the following commands to squash your stuff:

```bash
git rebase -i HEAD~6
```

It will open vim or something equalvant. Then replace `pick` with `s` on the last `5` commits and save + exit the file.

Your history is now been altered to one single commit, now run `git push -f` to force push your changes to the remote.

olivierchatry:
stupid me, was missing the -f on the push ... thanks a lot !

sirvine:
I notice this isn't in 2.5.0-beta1. Should we expect this to appear in a release soon?

bmac:
I'll make sure this gets included in the `2.5.0-beta.2` release.

PR #4155, opened 2016-02-11
remove container related deprecation warnings

### LINKED ISSUE #4151: Deprecation ember-application.injected-container triggered inside ember-data code
I've got a customized set of store, adapter, and serializer classes. For testing I used a setup of those classes similar to https://github.com/emberjs/data/blob/v2.3.3/tests/helpers/store.js. When running tests in this setup (that worked fine before 2.3.*) I get

> DEPRECATION: Using the injected `container` is deprecated. Please use the `getOwner` helper instead to access the owner of this object. [deprecation id: ember-application.injected-container] See http://emberjs.com/deprecations/v2.x#toc_injected-container-access for more details.
>         at HANDLERS.(anonymous function) (http://localhost:4200/assets/vendor.js:16695:7)
>         at raiseOnDeprecation (http://localhost:4200/assets/vendor.js:16605:12)
>         at HANDLERS.(anonymous function) (http://localhost:4200/assets/vendor.js:16695:7)
>         at invoke (http://localhost:4200/assets/vendor.js:16711:7)
>         at deprecate (http://localhost:4200/assets/vendor.js:16664:32)
>         at Object.deprecate (http://localhost:4200/assets/vendor.js:26672:37)
>         at Object.defineProperty.get (http://localhost:4200/assets/vendor.js:12125:26)
>         at getOwner (http://localhost:4200/assets/vendor.js:79144:22)
>         at modelFactoryFor (http://localhost:4200/assets/vendor.js:78280:55) 

This basically complains about https://github.com/emberjs/data/blob/v2.3.3/addon/-private/utils.js#L71 using the container (in this case caused by https://github.com/emberjs/data/blob/v2.3.3/addon/-private/system/store.js#L1530, there are different causes, but all of the referring to getOwner in utils.js).

https://github.com/emberjs/data/blob/v2.3.3/addon/-private/utils.js#L60 promises that this polyfill won't trigger a deprecation, so I would consider that a bug. 

Having said that: Is this approach to test a set of custom store, adapter and serializer as intended (and: future proof), or could you recommend a better approach that might not trigger this behaviour by e.g. returning an owner in https://github.com/emberjs/data/blob/v2.3.3/addon/-private/utils.js#L67?

sly7-7:
Closes #4151

It seems like some code was forgotten during the addonization.
Thanks to @rwjblue pointing it to me.

### DIFF
--- a/tests/helpers/owner.js
+++ b/tests/helpers/owner.js
@@ -1,0 +1,11 @@
+import Ember from 'ember';
+
+let Owner;
+
+if (Ember._RegistryProxyMixin && Ember._ContainerProxyMixin) {
+  Owner = Ember.Object.extend(Ember._RegistryProxyMixin, Ember._ContainerProxyMixin);
+} else {
+  Owner = Ember.Object.extend();
+}
+
+export default Owner;
--- a/tests/helpers/store.js
+++ b/tests/helpers/store.js
@@ -1,14 +1,21 @@
 import Ember from 'ember';
 import DS from 'ember-data';
+import Owner from './owner';
 
 export default function setupStore(options) {
-  var container, registry;
+  var container, registry, owner;
   var env = {};
   options = options || {};
 
   if (Ember.Registry) {
     registry = env.registry = new Ember.Registry();
-    container = env.container = registry.container();
+    owner = Owner.create({
+      __registry__: registry
+    });
+    container = env.container = registry.container({
+      owner: owner
+    });
+    owner.__container__ = container;
   } else {
     container = env.container = new Ember.Container();
     registry = env.registry = container;
@@ -63,4 +70,3 @@ export {setupStore};
 export function createStore(options) {
   return setupStore(options).store;
 }
-
--- a/tests/integration/references/has-many-test.js
+++ b/tests/integration/references/has-many-test.js
@@ -171,7 +171,7 @@ if (isEnabled("ds-references")) {
   test("push(array) works with polymorphic type", function(assert) {
     var done = assert.async();
 
-    env.container.register('model:mafia-boss', Person.extend());
+    env.registry.register('model:mafia-boss', Person.extend());
 
     var family;
     run(function() {


pangratz:
#4158 fixes the failing test in this PR. After that one has merged and this PR is rebased, I think this is good to go :rocket:

sly7-7:
@pangratz I was working on the failing test too, I've integrated the change in this PR. I've also removed the env.container.destroy thing (is this necessary though ?)

pangratz:
> I've also removed the env.container.destroy thing (is this necessary though ?)

We are not consistent with properly destroying the `env` created via `setupSore`/`createStore` in the tests. Sometimes only the [`store` is destroyed](https://github.com/emberjs/data/blob/45b36b563691f3f4f84b089998c7cd413da23dcf/tests/integration/store/query-record-test.js#L27), sometimes the [`container`](https://github.com/emberjs/data/blob/7bef33dc8701e0bf2a31eed403075bf545efcb14/tests/integration/adapter/serialize-test.js#L24), sometimes [both](https://github.com/emberjs/data/blob/7bef33dc8701e0bf2a31eed403075bf545efcb14/tests/unit/store/model-for-test.js#L29-L30) :laughing: Eventually the test setup should be consistent.

> I've also removed the env.container.destroy thing (is this necessary though ?)

I think it should be kept for now. It doesn't hurt I guess...

sly7-7:
@pangratz I will bring back the env.container.destroy. It was just that the tests were passing without this (I like to remove LOC when the tests pass :smile: )
Anyway, I don't know if it would be possible, but it would be great if we could add a unique afterEach hook doing the dirty work for all the tests. So we don't have to care about destroying (or not) env in each tests.

rwjblue:
>  I don't know if it would be possible, but it would be great if we could add a unique afterEach hook doing the dirty work for all the tests. So we don't have to care about destroying (or not) env in each tests.

We can add a `QUnit.testDone` ([see here](https://github.com/aptible/dashboard.aptible.com/blob/master/tests/test-helper.js#L29-L36)).

rwjblue:
But I suppose that can be a separate PR since this passes fine without...

rwjblue:
This looks good to me...

bmac:
Thanks @sly7-7

PR #4158, opened 2016-02-12
[BUGFIX beta] Use registry instead of container in tests

pangratz:
This is the preferred method way to register additional stuff.


pangratz:
@bmac this is a fix for a test which is behind a feature flag. I'm never sure: to `[BUGFIX beta]` or not to `[BUGFIX beta]`, that is the question...

sly7-7:
In the meantime I've integrated the same change in my PR, do you think we should merge this one, then the other, or just close this and merge the other ?

pangratz:
I created this PR since fixing the `env.container` is IMHO not directly related to the changes in your PR. But it's ok too to fix this in the course of #4155.

PR #4159, opened 2016-02-12
[DOC] Mark store.filter as deprecated in JSDoc.

chriskrycho:
The method is marked as deprecated internally, but not in the docs.


bmac:
Thanks @chriskrycho

PR #4160, opened 2016-02-12
Fixes broken link

jbailey4:
Previously the link to the emberjs guides on customizing adapters was broken, this fix points to the correct link for the current stable version of emberjs and ember data.

Looks like the previous link (https://guides.emberjs.com/v2.0.0/models/connecting-to-an-http-server/) was resolving to the version 2.0 guides on Models -> Connecting to an HTTP Server, but that does not exist anymore. It is now just Models -> Customizing Adapters.

### DIFF
--- a/README.md
+++ b/README.md
@@ -139,8 +139,7 @@ set of RESTful JSON conventions.
 
 To learn more about adapters, including what conventions the
 `RESTAdapter` follows and how to build your own, see the Ember.js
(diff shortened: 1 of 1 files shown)


pangratz:
Thank you kindly!

I just leave this here as a note for future readers: the link should be updated to point to the latest version number, once emberjs/guides#604 is sorted out.

PR #4164, opened 2016-02-15
[CLEANUP] remove `toString` definitions for models in tests

pangratz:
My string theory: the definitions for `toString` are a relict from the
pre-container era where this might have been needed. :barber:

This makes the recently failing builds for `master` green again. :green_apple: 

---

Note: the tests were failing since `2.3.1` ember seals `Mixin` so adding stuff
after `create` will error out: emberjs/ember.js#12612.


bmac:
Thanks @pangratz :traffic_light:

PR #4171, opened 2016-02-19
[CLEANUP] Remove shim for Backburner.join

HeroicEric:
This shim was added a long while back when we could not rely on the version of Backburner that Ember was using to have the `join()` function.

Any currently supported version of Ember (2.x) should have this now.

Similarly, any currently supported (2.x) version of Ember also has `Ember._Backburner` available.

Shim was added in:
https://github.com/emberjs/data/pull/2608

`Backburner.join()` was added in:
https://github.com/ebryn/backburner.js/pull/119

### DIFF
--- a/addon/-private/system/store.js
+++ b/addon/-private/system/store.js
@@ -57,50 +57,9 @@ import isEnabled from 'ember-data/-private/features';
 
 export let badIdFormatAssertion = '`id` has to be non-empty string or number';
 
-var Backburner = Ember._Backburner || Ember.Backburner || Ember.__loader.require('backburner')['default'] || Ember.__loader.require('backburner')['Backburner'];
+const Backburner = Ember._Backburner;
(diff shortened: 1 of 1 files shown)


bmac:
Thanks @HeroicEric

sly7-7:
not sure if it's related, but this makes a test fail on travis: 
not ok 92 PhantomJS 2.1 - integration/adapter/build-url-mixin - BuildURLMixin with RESTAdapter: buildURL - with host and namespace
    ---
        actual: >
            null
        message: >
            Test timed out
        Log: |
    ...

bmac:
I just restarted travis to see if its related to this change. It looks like travis said the pr was green before I merged it...

sly7-7:
@bmac thanks, it passes now. Maybe an unexpected latency on the server, or something like that

PR #4172, opened 2016-02-19
[CLEANUP] Remove shim for Ember.Service

HeroicEric:
Removes a shim for `Ember.Service`.

This isn't needed anymore since Ember 2.x.x all have `Ember.Service` available.


rwjblue:
:+1: - But we do support 1.13 also (still fine, I'm just pointing it out...)

PR #4177, opened 2016-02-22
Elaborate on running tests in browser

courajs:
Visting `http://localhost:4200` displays the dummy app's "Welcome to Ember" screen. The tests reside at `/tests`

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -66,7 +66,7 @@ We love pull requests. Here's a quick guide:
 
 2. Run the tests. We only take pull requests with passing tests, and it's great
 to know that you have a clean slate, see notes on how to run unit tests [here](https://github.com/emberjs/data#how-to-run-unit-tests). (To see tests in the browser,
-run `npm start` and open `http://localhost:4200`.)
+run `npm start` and open `http://localhost:4200/tests`.)
 
 3. Add a test for your change. Only refactoring and documentation changes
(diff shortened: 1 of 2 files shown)


sly7-7:
Thank you for the contribution. Thinking a bit more about that, since ember-data is a true ember-cli addon, I wonder if we could just use the basic blueprint generated by `ember addon` command.
```
## Installation

* `git clone` this repository
* `npm install`
* `bower install`

## Running

* `ember server`
* Visit your app at http://localhost:4200.

## Running Tests

* `npm test` (Runs `ember try:testall` to test your addon against multiple Ember versions)
* `ember test`
* `ember test --server`

## Building

* `ember build`

```

@bmac @fivetanley Any thoughts ?

courajs:
That's a good point. There are some advantages to the `npm start` + `/tests` method though.
`ember test --server` does pop up a new chrome instance, which I find a little annoying - it's difficult to cmd+tab to the correct one first try.
Since the filtering/single test rerun support isn't there for the phantom runner, I usually stick to the browser anyway.
Not really sure we want to mention both, since that will be confusing to new contributors. "Why are there two options? Which is better?"

bmac:
@sly7-7 I'm going to merge this pr as is because it is consistent with how things worked pre ember addon. I'm not opposed to change it to say `ember test --server` if you or someone else would like to open another pr.

PR #4178, opened 2016-02-22
DS.Store type presence checks

### LINKED ISSUE #4170: Misleading assert - passing classes to store methods has been removed
Specific message: "Passing classes to store.serializerFor has been removed. Please pass a dasherized string instead of undefined"
See [JSBin](http://emberjs.jsbin.com/bowugudede/edit?js,output)
I bumped into this when using the `EmbeddedRecordMixin`, and was failing to include `type` on one of the embedded records. This message was not very helpful

courajs:
Fixes #4170.
Are you alright with generated tests like this?
I wasn't quite sure where to put these. `serializerFor`, `createRecord`, and a few others have their own test files, but many of these didn't really seem to have a home. It also seems a small shame to split these tests up when they're almost exactly the same.

### DIFF
--- a/addon/-private/system/store.js
+++ b/addon/-private/system/store.js
@@ -73,6 +73,7 @@ var get = Ember.get;
 var set = Ember.set;
 var once = Ember.run.once;
 var isNone = Ember.isNone;
+var isPresent = Ember.isPresent;
 var Promise = Ember.RSVP.Promise;
 var copy = Ember.copy;
 var Store;
@@ -281,6 +282,7 @@ Store = Service.extend({
     @return {DS.Model} record
   */
   createRecord(modelName, inputProperties) {
+    assert("You need to pass a model name to the store's createRecord method", isPresent(modelName));
     assert(`Passing classes to store methods has been removed. Please pass a dasherized string instead of ${Ember.inspect(modelName)}`, typeof modelName === 'string');
     var typeClass = this.modelFor(modelName);
     var properties = copy(inputProperties) || new EmptyObject();
@@ -457,6 +459,7 @@ Store = Service.extend({
     @return {Promise} promise
   */
   findRecord(modelName, id, options) {
+    assert("You need to pass a model name to the store's findRecord method", isPresent(modelName));
     assert('Passing classes to store methods has been removed. Please pass a dasherized string instead of '+ Ember.inspect(modelName), typeof modelName === 'string');
     assert(badIdFormatAssertion, (typeof id === 'string' && id.length > 0) || (typeof id === 'number' && !isNaN(id)));
 
@@ -531,6 +534,7 @@ Store = Service.extend({
     @return {Promise} promise
   */
   findByIds(modelName, ids) {
+    assert("You need to pass a model name to the store's findByIds method", isPresent(modelName));
     assert('Passing classes to store methods has been removed. Please pass a dasherized string instead of '+ Ember.inspect(modelName), typeof modelName === 'string');
     let promises = new Array(ids.length);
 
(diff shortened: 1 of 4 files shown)


wecc:
Thanks @courajs!

We have a couple of other tests where we loop inside of one `test(...)` instead of generating multiple tests. This gives us the ability to set expected asserts (something like `assert.expect(MODEL_NAME_METHODS .length)` in this case).

See [here](https://github.com/emberjs/data/blob/f4d3b9ef34145d2e75ebb8e70ac757a8c613bf73/tests/integration/store-test.js#L319-L332) for an example.

courajs:
Alrighty, I've updated to use only one test with many asserts

addon/-private/system/store.js
```diff
@@ -281,6 +282,7 @@ Store = Service.extend({
     @return {DS.Model} record
   */
   createRecord(modelName, inputProperties) {
+    assert("You need to pass a type to the store's createRecord method", isPresent(modelName));
```
bmac: Ember Data uses `modelName` for the type argument in most of these functions. I also know that `type` has also been used to refer to the same information. I think we should pick one term and be consistent. Right now I have a preference for `modelName` but I could easily be convinced to use `type` instead.

pangratz: It looks like it's not very consistent at the moment (see [this](https://git.io/v2NHM) and [this](https://git.io/v2NHH)), but I am in favor of `modelName` too, since it looks like the parameter for most method documentations is named `modelName` as well.

tests/unit/store/asserts-test.js
```diff
@@ -0,0 +1,36 @@
...
+  'hasRecordForId',
+  'recordForId',
+  'query',
+  'queryRecord',
+  'findAll',
+  'peekAll',
+  'filter',
+  'recordIsLoaded',
+  'modelFor',
+  'modelFactoryFor',
```
pangratz: `recordIsLoaded` is marked as private in the API docs; `modelFactoryFor` currently has no public documentation (though the name suggests it is public).

Tiny nitpick: I wonder if those methods should be excluded from the assertion tests here, since they are not public API?

courajs: I looked at a few existing asserts and saw `type` so I matched that. I'll change to `modelName` and take a look for existing asserts that would be good to change as well.

I think asserts in private API can still be useful occasionally when bad values (erroneously) sneak through into internal code. If they're usually avoided in ED though I can remove them.
https://github.com/emberjs/data/pull/4178/files#r55129914

bmac: I think its ok to leave the asserts on private APIs. If these methods were ever marked public in the future its unlikely they someone would remember to come back and update this test.

pangratz: > If these methods were ever marked public in the future its unlikely they someone would remember to come back and update this test.

:+1:

courajs:
I've changed "type" to "model name" in the assertion messages

tests/unit/store/asserts-test.js
```diff
@@ -0,0 +1,36 @@
...
+  'normalize',
+  'adapterFor',
+  'serializerFor'
+];
+
+test("Calling Store methods with no type asserts", function(assert) {
```
pangratz: Since #4223 the tests are also run in the `production` environment, where the assertions are stripped. That's why this test is failing.

You need to use `testInDebug` here instead of `test`:

```js
import testInDebug from 'dummy/tests/helpers/test-in-debug';
```

Sorry you fell over this. This is my fault, I need to update `CONTRIBUTING.md` ASAP.

pangratz:
Can you squash into 1 commit and prefix with `[BUGFIX beta]`?

I think then this is good to go :rocket:

bmac:
Thanks @courajs.

PR #4187, opened 2016-02-25
Add acceptance tests for the mocha blueprints

Turbo87:


Turbo87:
It looks like the blueprints on Windows are actually being generated with `\r\n` instead of `\n` on AppVeyor due to https://github.com/emberjs/data/blob/master/appveyor.yml#L3-L5 which is causing trouble with the tests, but the same is also true for the existing QUnit blueprints.

Should I adjust the tests to not test the line breaks for now or change AppVeyor to always use the original line breaks?

/cc @stefanpenner @kellyselden

Turbo87:
rebased on top of #4189 to solve the line ending issues

bmac:
Do you mind rebasing this pr @Turbo87

Turbo87:
@bmac done

pangratz:
@Turbo87 this looks good. Can you please squash into 1 commit?

Turbo87:
@pangratz done

pangratz:
:tada:

Turbo87:
@pangratz out of curiosity, what is the reason for squashing the commits all the time?

pangratz:
I can only speak for myself but I can think of the following reasons:

- easier possibility to back port stuff into other release channels and older releases via a single `cherry-pick`
- cleaner to revert if stuff breaks
- one PR handles one specific issue in one commit
- cleaner `git log` as every entry corresponds to one single, cohesive change
- consistency, since almost all PR's in this repo are handled this way

Turbo87:
> easier possibility to back port stuff into other release channels and older releases via a single cherry-pick

:+1: although it would manageable :wink: 

> cleaner to revert if stuff breaks

GitHub actually allows you to just revert the merge commit

> one PR handles one specific issue in one commit

I personally try to keep the commits as small as possible to make them easier to review and then limit the PRs to single, specific features

> cleaner git log as every entry corresponds to one single, cohesive change

granted, although since we're still using non-fast-forward merge commits the history graph isn't entirely clean anyway

> consistency, since almost all PR's in this repo are handled this way

sure, that's essentially the reason why I'm not complaining, just asking :wink:

PR #4194, opened 2016-02-28
Use modelNameFromPayloadKey when type is given

backspace:
The code checked for a model using the possibly-overridden `modelNameFromPayloadKey` but didn’t use the override when getting the serialiser and model.

The test is really only verifying the `store.modelFor(this.modelNameFromPayloadKey(hash.type));` change, not the serialiser one, since I didn’t see any tests for custom serialisers in this file. Let me know if I should make a test for that regardless.

### DIFF
--- a/addon/serializers/rest.js
+++ b/addon/serializers/rest.js
@@ -200,8 +200,8 @@ var RESTSerializer = JSONSerializer.extend({
     const primaryHasTypeAttribute = modelHasAttributeOrRelationshipNamedType(primaryModelClass);
     // Support polymorphic records in async relationships
     if (!primaryHasTypeAttribute && hash.type && store._hasModelFor(this.modelNameFromPayloadKey(hash.type))) {
-      serializer = store.serializerFor(hash.type);
-      modelClass = store.modelFor(hash.type);
+      serializer = store.serializerFor(this.modelNameFromPayloadKey(hash.type));
+      modelClass = store.modelFor(this.modelNameFromPayloadKey(hash.type));
     } else {
       serializer = primarySerializer;
       modelClass = primaryModelClass;
--- a/tests/integration/serializers/rest-serializer-test.js
+++ b/tests/integration/serializers/rest-serializer-test.js
@@ -144,6 +144,45 @@ test("normalizeResponse with custom modelNameFromPayloadKey", function(assert) {
   });
 });
 
+test("normalizeResponse with type and custom modelNameFromPayloadKey", function(assert) {
+  assert.expect(2);
+
+  var homePlanetNormalizeCount = 0;
+
+  env.restSerializer.modelNameFromPayloadKey = function(root) {
+    return "home-planet";
+  };
+
+  env.registry.register('serializer:home-planet', DS.RESTSerializer.extend({
+    normalize() {
+      homePlanetNormalizeCount++;
+      return this._super.apply(this, arguments);
+    }
+  }));
+
+  var jsonHash = {
+    "my-custom-type": [{ id: "1", name: "Umber", type: "my-custom-type" }]
+  };
+  var array;
+
+  run(function() {
+    array = env.restSerializer.normalizeResponse(env.store, HomePlanet, jsonHash, '1', 'findAll');
+  });
+
+  assert.deepEqual(array, {
+    data: [{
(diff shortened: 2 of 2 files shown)


pangratz:
I think it would be good to add a test for the correct serializer too. [This is an example](https://github.com/backspace/data/blob/aca709666a60b8fbbd97177b7f5b303b78abebc5/tests/integration/serializers/rest-serializer-test.js#L288-L293) of using a custom serializer in the tests.

tests/integration/serializers/rest-serializer-test.js
```diff
@@ -143,6 +143,35 @@ test("normalizeResponse with custom modelNameFromPayloadKey", function(assert) {
...
 
+test("normalizeResponse with type and custom modelNameFromPayloadKey", function(assert) {
+  assert.expect(1);
+
+  env.restSerializer.modelNameFromPayloadKey = function(root) {
+    return Ember.String.singularize(root);
+  };
+
+  var jsonHash = {
+    "home-planets": [{ id: "1", name: "Umber", type: "home-planets" }]
```
pangratz: What do you think of changing `type: "home-planets"` to `type: "my-custom-type"` to make it more clear that `modelNameFromPayloadKey` is used to resolve the model and serializer?

backspace:
Thanks for the pointer to an example serialiser override. I’ve overwritten the commit with a similar assertion for that and added the `[BUGFIX beta]` prefix to the commit message. Let me know if anything else is necessary!

addon/serializers/rest.js
```diff
@@ -200,8 +200,8 @@ var RESTSerializer = JSONSerializer.extend({
     const primaryHasTypeAttribute = modelHasAttributeOrRelationshipNamedType(primaryModelClass);
     // Support polymorphic records in async relationships
     if (!primaryHasTypeAttribute && hash.type && store._hasModelFor(this.modelNameFromPayloadKey(hash.type))) {
-      serializer = store.serializerFor(hash.type);
-      modelClass = store.modelFor(hash.type);
+      serializer = store.serializerFor(this.modelNameFromPayloadKey(hash.type));
+      modelClass = store.modelFor(this.modelNameFromPayloadKey(hash.type));
```
pangratz: I just realized this now, but at a first glance it looks weird the the modelName is normalized by a function called `modelNameFromPayloadKey`; since the passed argument is technically not a key from the payload.

That function originally has been added for the case where the payload key is different. So the `blog/post` entry in the following payload is used for the `post` model (see [this documentation](https://github.com/emberjs/data/blob/2322aacc3f3e4fb4fe9b30c7d790c45126fa5c73/addon/serializers/rest.js#L414-L471)):

```js
{
  "blog/post": {
    "id": 1
  }
}
```

`modelNameFromPayloadKey` is however already used within `json` serializer to normalize the modelName from `hash.type` (see [here](https://github.com/emberjs/data/blob/2322aacc3f3e4fb4fe9b30c7d790c45126fa5c73/addon/serializers/json.js#L603)); so this would not be the first time here that this function is used differently than the name would imply.

---

To be clear, I think it is definitely a bugfix to use the normalized modelName within the `if` branch here, so this PR looks good to me. I am just not sure if there should be a different function (like `modelNameFromPayloadType` or something), which would be more appropriate...

backspace: hmm yes, I see the confusion. The API that led me to find this bug is a weird pseudo-JSON API that was settled during the RC phase, it uses `type`. I was confused while debugging by the role that `JSONSerializer` inheritance played, but you probably know better than me whether it’s feasible to change the method name.

backspace:
I see that this has conflicts now. Is it worth rebasing to keep updated? Might it eventually be merged?

pangratz:
So this has been discussed in the team meeting. Eventually the inconsistency should be addressed via new additional `modelNameFromPayloadType` and `payloadTypeFromModelName` hooks, but they are out of scope for this PR.

@backspace can you rebase this PR and fix the merge conflicts? After that, this is good to go :rocket:

backspace:
Thanks for the update on the inconsistency.

Rebased and ready! :grinning:

pangratz:
:tada: :dancers: thanks @backspace!

PR #4196, opened 2016-02-29
Restructure and elaborate on test which fixes a tricky bug

pangratz:
This slightly restructures and adds more comments to a test, which fixed
a tricky bug which occurred while updating the canonical state of a
hasMany relationship in combination with locally created records.


bmac:
:+1:

PR #4200, opened 2016-03-02
Avoid errors when ember-cli-shims is not included.

### LINKED ISSUE #4192: ember-cli-shim check errors if no shim is provided
this prevents us from bundling shims with the future ember module.

I suspect something like `this.project.findAddonByName('ember-source')` will be a sufficient check

rwjblue:
Closes #4192.

### DIFF
--- a/index.js
+++ b/index.js
@@ -25,7 +25,8 @@ module.exports = {
     var VersionChecker = require('ember-cli-version-checker');
 
     var checker = new VersionChecker(this);
-    var shims = checker.for('ember-cli-shims', 'bower');
+    // prevent errors when ember-cli-shims is no longer required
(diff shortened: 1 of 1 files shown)


bmac:
Restarted travis. Looked like the original issue was unrelated to this code change.

rwjblue:
Ya, but soon we will not be using ember-cli-shims at all (once ember addon follows ember-data's lead and ships its own public modules) so it makes sense to ensure that we don't error when it isn't present...

bmac:
Agreed. I see now my comment was ambiguous. I intended to suggest the reason that travis failed appeared unrelated to this code change.

rwjblue:
Ahh! I see, sorry about the confusion/noise :)

PR #4204, opened 2016-03-02
Fix RESTAdapter.findRecord without a snapshot

drogus:
From the commit message:

```
Before introducing a feature ds-finder-include it was possible to run
findRecord on an adapter without passing a snapshot. After introducing
the feature it's no longer possible even if the feature is disabled,
because some of the code was outsief of the flag check.

This commit fixes the code to work like before introducing the
ds-finer-include change.
```

I'm not 100% sure if this is public behaviour of `findRecord`, but it will cause regressions for people that use it. In case it's not an issue, I can also submit a PR to have a more meaningful error (like "You need to pass a snapshot to `findRecord`").

### DIFF
--- a/tests/unit/adapters/rest-adapter/build-query-test.js
+++ b/tests/unit/adapters/rest-adapter/build-query-test.js
@@ -13,6 +13,13 @@ test("buildQuery() returns an empty query when snapshot has no query params", fu
   assert.deepEqual(query, {}, 'query is empty');
 });
 
+test("buildQuery - doesn't fail without a snapshot", function(assert) {
+  const adapter = DS.RESTAdapter.create();
+  const query = adapter.buildQuery();
+
+  assert.deepEqual(query, {}, 'returns an empty query');
+});
+
 if (isEnabled('ds-finder-include')) {
   test("buildQuery() returns query with `include` from snapshot", function(assert) {
     const adapter = DS.RESTAdapter.create();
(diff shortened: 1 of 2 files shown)


tests/unit/adapters/rest-adapter/build-query-test.js
```diff
@@ -14,6 +14,13 @@ test("buildQuery() returns an empty query when snapshot has no query params", fu
 });
 
 if (isEnabled('ds-finder-include')) {
+  test("buildQuery - doesn't fail without a snapshot", function(assert) {
+    const adapter = DS.RESTAdapter.create();
+    const query = adapter.buildQuery();
+
+    assert.deepEqual(query, {}, 'returns an empty query');
+  });
```
bmac: I think we want to move this test outside of the `isEnabled('ds-finder-include')` check.

bmac:
@drogus I think this should be a `[BUGFIX beta]` or potentially a `[BUGFIX release]`.

drogus:
Thanks for the comments @bmac, I'll look into fixing this. I think [BUGFIX release] fits, because it's a bug in the released version.

Is there any guide that I missed that explains contributing? In CONTRIBUTING.md there're mostly rules on how to submit issues.

bmac:
Thanks @drogus. There is a guide that briefly explains the commit tagging here  https://guides.emberjs.com/v2.4.0/contributing/adding-new-features/. However, I think Ember Data's CONTRIBUTING.md could use some love. It hasn't been updated in a long while.

PR #4206, opened 2016-03-02
unify prod/dev add-on build

stefanpenner:
@rwjblue / @fivetanley r?

### DIFF
--- a/lib/babel-build.js
+++ b/lib/babel-build.js
@@ -25,13 +25,7 @@ function babelOptions(libraryName, _options) {
     getModuleId: function (name) {
       return name.replace(/\/index$/g, '');
     },
-    resolveModuleSource: function(name, filename) {
-      if (name.indexOf('.') === 0) {
-        return libraryName + '/' + path.join(path.dirname(filename), name);
-      } else {
-        return name;
-      }
-    }
+    resolveModuleSource: require('amd-name-resolver').moduleResolve
   };
 
   Object.keys(_options).forEach(function(opt) {
--- a/package.json
+++ b/package.json
@@ -22,6 +22,7 @@
   "author": "",
   "license": "MIT",
   "dependencies": {
+    "amd-name-resolver": "0.0.5",
     "babel-plugin-feature-flags": "^0.2.0",
     "babel-plugin-filter-imports": "^0.2.0",
     "broccoli-babel-transpiler": "^5.5.0",


index.js
```diff
@@ -22,7 +27,16 @@ module.exports = {
 
   init: function() {
     var bowerDeps = this.project.bowerDependencies();
+
     var VersionChecker = require('ember-cli-version-checker');
+    var options = this.options = this.options || {};
```
pangratz: Is `this.options` intentionally assigned to `this.options || {}` or is this a "typo"?

stefanpenner: intentional, it aims to default it to {} if it is not already set

pangratz: But it's never used again in this file, that's why I am asking. Is `init` called multiple times?

stefanpenner: Well, an ancestor could set it, and a downstream could use it.

pangratz: Ah, alright. Thanks for the clarification :gift:

stefanpenner:
the failure appears to be:

```js
 1) Acceptance: generate and destroy adapter blueprints adapter throws when --base-class is same as name:
     Error: the object {
  "errorLog": [
    {
      "isSilentError": true
      "message": "Adapters cannot extend from themself. To resolve this, remove the `--base-class` option or change to a different base-class."
      "name": "SilentError"
      "suppressedStacktrace": true
    }
  ]
  "exitCode": 1
  "inputStream": []
  "outputStream": [
    "version: 2.4.1\n"
    "\u001b[33mThe option '--disable-analytics' is not registered with the generate command. Run `ember generate --help` for a list of supported options.\u001b[39m\n"
    "\u001b[33mThe option '--watcher' is not registered with the generate command. Run `ember generate --help` for a list of supported options.\u001b[39m\n"
    "\u001b[33mThe option '--skipGit' is not registered with the generate command. Run `ember generate --help` for a list of supported options.\u001b[39m\n"
    "installing adapter\n"
  ]
  "statusCode": 1
} was thrown, throw an Error :)
```

stefanpenner:
ah, it might be ember-cli bump

stefanpenner:
Alright, this should be good for review.

stefanpenner:
r?

index.js
rwjblue: Also, it is used just below (via `options.babel` which is also the same object as `this.options.babel`).

rwjblue:
This looks good to me, but I need pull it down and test:

* linked app in production
* linked app in development
* app with globals output 

Will try to do it tonight/tomorrow morning...

bmac:
Code looks good. I'll test out the 3 scenarios @rwjblue mentioned later today.

index.js
```diff
@@ -22,7 +27,16 @@ module.exports = {
...
+
+    options.babel = options.babel || {};
+
+    if (process.env.EMBER_ENV === 'production') {
+      add(options.babel, 'backlist', ['es6.modules', 'useStrict']);
```
bmac: `backlist` -> `blacklist`

index.js
```diff
@@ -22,7 +27,16 @@ module.exports = {
 
   init: function() {
     var bowerDeps = this.project.bowerDependencies();
+
     var VersionChecker = require('ember-cli-version-checker');
+    var options = this.options = this.options || {};
+
+    options.babel = options.babel || {};
+
+    if (process.env.EMBER_ENV === 'production') {
```
bmac: Does this get set automatically by an `ember build --environment=production`? When testing locally I had to set it manually.

rwjblue: Yes, it does get set automatically, but not before `init` of addons.  It is set here: https://github.com/ember-cli/ember-cli/blob/3073a5c6a67debc6f4a1c37dc404ccab126d7309/lib/models/builder.js#L45 during the starting of the build.

PR #4214, opened 2016-03-07
Fix AMD dependencies

chadhietala:
This fixes the module name resolution. Currently production builds do
not resolve correctly as the dependencies of modules are being resolved
like the following.

```
define('ember-data/-private/system/references',
['exports',
'-private/system/references/record',
'-private/system/references/belongs-to',
'-private/system/references/has-many'],
function (exports, _privateSystemReferencesRecord, _privateSystemReferencesBelongsTo, _privateSystemReferencesHasMany) {...});
```

The module names should get resolved now. Such babel config.

### DIFF
--- a/lib/babel-build.js
+++ b/lib/babel-build.js
@@ -1,5 +1,6 @@
 var babel = require('broccoli-babel-transpiler');
 var path  = require('path');
+var moduleResolve = require('amd-name-resolver').moduleResolve;
 
 function babelOptions(libraryName, _options) {
   _options = _options || {};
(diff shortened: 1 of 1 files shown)


lib/babel-build.js
```diff
@@ -20,12 +21,12 @@ function babelOptions(libraryName, _options) {
     sourceMaps: false,
     modules: 'amdStrict',
     moduleRoot: libraryName,
-    moduleId: true,
```
stefanpenner: such plural

stefanpenner:
was this test with a prod build?

PR #4215, opened 2016-03-07
unify prod/dev add-on build

stefanpenner:
force-push mistake, resurrection of https://github.com/emberjs/data/pull/4206


index.js
```diff
@@ -84,28 +91,21 @@ module.exports = {
...
+      version(),
+      dir
+    ]));
   },
 
   included: function(app) {
     this._super.included.apply(this, arguments);
 
+    if (process.env.EMBER_ENV === 'production') {
+      add(this.options.babel, 'backlist', ['es6.modules', 'useStrict']);
```
bmac: `backlist` -> `blacklist`

stefanpenner:
Did I really fuck up that spelling again...

fivetanley:
it was your heart on the line

rwjblue:
I am so confused by what is going on. Wasn't this merged already?

stefanpenner:
@rwjblue nope not yet

index.js
```diff
@@ -84,28 +91,21 @@ module.exports = {
...
+      version(),
+      dir
+    ]));
   },
 
   included: function(app) {
     this._super.included.apply(this, arguments);
 
+    if (process.env.EMBER_ENV === 'production') {
+      add(this.options.babel, 'blacklist', ['es6.modules', 'useStrict']);
```
bmac: When I run `npm run production` the add function fails because `this.options.babel` is undefined.

index.js
```diff
@@ -22,7 +27,9 @@ module.exports = {
 
   init: function() {
     var bowerDeps = this.project.bowerDependencies();
+
     var VersionChecker = require('ember-cli-version-checker');
+    var options = this.options = this.options || {};
```
bmac: Locally I change this line to use `{ babel: {} }` as the default option and tested out the resulting build. The built assets looked good.

index.js
stefanpenner: hmm, good catch. I admittedly have been lazy the last few updates here, sorry for not catching.

index.js
stefanpenner: ya that works in some cases, but actually it should be done at 105, submiting PR

stefanpenner:
recent changes are tested in small app + ember data builds with both --prod and not

bmac:
Looks good thanks @stefanpenner

PR #4222, opened 2016-03-08
Strip code for DS_WARN_ON_UNKNOWN_KEYS warning in production

pangratz:
- wrap code inside runInDebug so it is stripped from production
- update checks so they work correctly with passed payload, which is
  actually a JSON-API document

### DIFF
--- a/addon/-private/system/store.js
+++ b/addon/-private/system/store.js
@@ -4,7 +4,7 @@
 
 import Ember from 'ember';
 import Model from 'ember-data/model';
-import { assert, warn } from "ember-data/-private/debug";
+import { assert, warn, runInDebug } from "ember-data/-private/debug";
 import _normalizeLink from "ember-data/-private/system/normalize-link";
 import normalizeModelName from "ember-data/-private/system/normalize-model-name";
 import {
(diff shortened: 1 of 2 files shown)


addon/-private/system/store.js
```diff
@@ -1674,21 +1674,28 @@ Store = Service.extend({
...
-        })) + ". Make sure they've been defined in your model.",
-        Object.keys(data).filter((key) => {
-          return !(key === 'id' || key === 'links' || get(type, 'fields').has(key) || key.match(/Type$/));
-        }).length === 0,
-        { id: 'ds.store.unknown-keys-in-payload' }
-      );
-    }
+    runInDebug(() => {
+      // If Ember.ENV.DS_WARN_ON_UNKNOWN_KEYS is set to true and the payload
+      // contains unknown attributes or relationships, log a warning.
```
bmac: Fun fact: All the code inside `runInDebug` get striped but this comment is preserved.

bmac:
Thanks @pangratz

addon/-private/system/store.js
pangratz: :stuck_out_tongue_winking_eye:

Issue #4203, opened 2016-03-02
prod builds are not currently tested


stefanpenner:

pangratz:
Closing since this has been implemented in #4223

PR #4223, opened 2016-03-09
Run tests in production

pangratz:
This adds a new test helper `testInDebug` which executes a test via `QUnit.test` if we are not in the `production` environment. If the tests are executed in `production` environment though, then the test is skipped via `QUnit.skip`.

The tests are now also run on TravisCI via `ember test --environment=production`.

This addresses #4203.

#### TODO

- [x] wait for relative paths to work, then remove relevant commit from this PR :arrow_right: looks like this has been resolved with #4215

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,6 +15,7 @@ script:
   - ./bin/lint-features
   - npm run-script test
   - npm run-script test:optional-features
+  - npm run-script test:production
   - npm run-script node-tests
 after_success:
   - npm run-script production
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -27,6 +27,7 @@ test_script:
   - npm version
   - cmd: npm run test
   - cmd: npm run test:optional-features
+  - cmd: npm run test:production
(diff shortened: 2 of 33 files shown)


bmac:
@pangratz do you mind rebasing this pr?

addon/-private/system/references.js
```diff
@@ -1,5 +1,5 @@
-import RecordReference from './references/record';
-import BelongsToReference from './references/belongs-to';
-import HasManyReference from './references/has-many';
+import RecordReference from 'ember-data/-private/system/references/record';
```
stefanpenner: relative should work correctly, if it does not that is a bug in the build we need to fix.

pangratz: Sure thing. Let's wait until that is resolved, then I will rebase and remove this commit.

pangratz: It looks like this has been resolved with #4215 :tada:

bmac:
Thanks @pangratz

Issue #4186, opened 2016-02-25
[Docs] DS.RESTSerializer needs update


James1x0:
normalizeHash is deprecated, Parse is going extinct.

bmac:
For some context: `normalizeHash` and `Parse` are referenced on this page
http://emberjs.com/api/data/classes/DS.RESTSerializer.html which is auto generated from the API docs in this file https://github.com/emberjs/data/blob/master/addon/serializers/rest.js.

James1x0:
Sorry @bmac, had to make a quick issue regarding this so I wouldn't forget. I may have time to tackle this soon. May not tho.

jsangilve:
@James1x0 I just realized that `normalizeHash` is deprecated. Apart from docs, it seems like no deprecation warning is thrown when you use it. I'm using it with Ember Data 2.3

James1x0:
@jsangilve It threw in ED 1.x, I think it's supposed to be completely gone from the api in 2.x

jsangilve:
@James1x0 it's gone from the API, but it seems like the `normalize` method of `DS.RESTSerializer` still uses a `normalizeHash` property:
```javascript
normalize(modelClass, resourceHash, prop) {
    if (this.normalizeHash && this.normalizeHash[prop]) {
      this.normalizeHash[prop](resourceHash);
    }
    return this._super(modelClass, resourceHash, prop);
  },
```
If you create a normalizeHash property while extending DS.RESTSerializer, it will work. I just created a PR that only solves the docs issue.

pangratz:
> comment by @jsangilve 
> > I've just realized that normalizeHash is deprecated. Apart from docs, it seems like no deprecation warning is thrown when you use it. I'm using it with Ember Data 2.3

There is [a deprecation warning in `1.13`](https://github.com/emberjs/data/blob/v1.13.16/packages/ember-data/lib/serializers/rest-serializer.js#L199-L202), but it looks like there has been something overlooked when [the deprecation has been removed in `2.0`](https://github.com/emberjs/data/blob/v2.0.0/packages/ember-data/lib/serializers/rest-serializer.js#L124-L126).

---

Since it has been deprecated in `1.13`, the `normalizeHash` stuff should be removed from within `normalize()` of the `rest` serializer; this eventually makes the `normalize()` obsolete in that serializer since it would only call `_super()`.

[...]

PR #4228, opened 2016-03-13
Solves #4186 and fixes `this._super` call within `normalize` method.

jsangilve:
See #4186.

Also, the `normalize` method was calling super with three parameters while it only receives two: http://emberjs.com/api/data/classes/DS.JSONSerializer.html#method_normalize


bmac:
Thanks @jsangilve

PR #4231, opened 2016-03-14
fixed date parsing in IE

IgorKvasn:
currently Internet Explorer cannot parse dates like: 
2003-12-02T23:00:00.000+**0000**

it accepts only:
2003-12-02T23:00:00.000+**00**

I encountered this bug when I was using date transformer.


bmac:
@IgorKvasn what version of Internet Explorer?

IgorKvasn:
IE 10

bmac:
This was merged into Ember Data as https://github.com/emberjs/data/pull/4396/commits/81615417961e393f732b6489534a9bbf92fd8c34

PR #3559, opened 2015-07-18
[BUGFIX release] Guard against isDestroyed in ManyArray.flushCanonical

jgwhite:
Addresses #3084

This is the simplest-possible-patch I could find to get around a real pain point for unloading records. It looks like you guys are doing more fundamental work to address this but I wonder if we could put a fix like this in place as a stopgap.

If it’s needless and the real fix is about to land, I totally understand. If you think it is worthwhile I’ll flesh it out and add tests.

PR #4234, opened 2016-03-16
Don't Call `set()` on Destroyed ManyArray

spinlock99:
This is a simple "fix" for a problem I encountered in my app. Essentially, MannyArray was attempting to set a property on itself after it had been destroyed. A simple guard clause stopped the error from crashing my app.

I'd love some feedback on this issue which is why I'm submitting it before it has tests, etc... Any comments would be greatly appreciated.


jcope2013:
this seems similar 

https://github.com/emberjs/data/pull/3559

spinlock99:
Yes, this looks like it's addressing the same bug. I've confirmed that the bug has been fixed on canary.

Issue #4238, opened 2016-03-16
Improve CONTRIBUTING.md


pangratz:
- [x] add section about prefixing commits, similar to [the one in ember.js](https://github.com/emberjs/ember.js/blob/master/CONTRIBUTING.md#commit-tagging) (added in #4244 by @courajs :sparkles: )
- [x] add ember-twiddles for easier reporting of a bug (setup using pretender) (added in #4716 by @pete-the-pete)
- [ ] add note about RFC for requesting a feature instead of opening issue on emberjs/data
  - [x] implement new feature using feature flags (see https://github.com/emberjs/data/issues/4238#issuecomment-198482078, added in #4327 by @sbatson5)
- [x] add section about tests for `assert` / `warn` to use [`testInDebug`](https://github.com/emberjs/data/pull/4223) so they are skipped when testing in production environment (added in #4240 by @courajs :rocket: )

courajs:
Can we just copy in the section on commit prefixing (and credit at the end)? Are there any differences?

pangratz:
> Can we just copy in the section on commit prefixing (and credit at the end)? Are there any differences?

They are basically the same. I think it is ok to copy them (cc @rwjblue @bmac).

courajs:
- [ ] Add section about wrapping new feature work in a feature flag (?)

pete-the-pete:
can someone describe the work involved with?:
- [ ] add ember-twiddles for easier reporting of a bug (setup using pretender)

courajs:
@pete-the-pete I believe this involves updating the jsbins/jsfiddle linked in [Reporting a Bug](https://github.com/emberjs/data/blob/master/CONTRIBUTING.md#reporting-a-bug) to
- Use [ember twiddle](https://ember-twiddle.com) instead of jsbin/fiddle
- Use [pretender](https://github.com/pretenderjs/pretender) instead of [mockjax](https://github.com/jakerella/jquery-mockjax)

Although I'd ask @pangratz to weigh in on why pretender is a better choice than mockjax - it's not obvious to me.

Dan-Ste:
@pangratz There is still need help? I'm a beginner in OSS and want to make my first PR :)

wecc:
From #4362 re Ember Twiddle:

[...]

PR #4240, opened 2016-03-18
Add note on `testInDebug`

courajs:
Addresses point 4 in Issue #4238

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -90,7 +90,7 @@ taken straight from the Ruby on Rails guide:
 * Update the documentation, the surrounding one, examples elsewhere, guides,
   whatever is affected by your contribution
 
-Syntax:
+## Syntax:
 
 * Two spaces, no tabs.
 * No trailing whitespace. Blank lines should not have any space.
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -99,4 +99,9 @@ Syntax:
 
 And in case we didn't emphasize it enough: we love tests!
 
+## Writing Tests
+
+* We do write tests for our warns and assertion messages. You may use the `assert.expectAssertion()` and `assert.expectWarning()` helpers for this. Note: Our Travis tests that will run on your PR run in the `production` envrionment, where warnings and assertions are stripped. If you're writing a test for these, you should use `testInDebug` helper rather than `test` to define your test.
+
```
pangratz: Typo: `envrionment` -> `environment`

I think it would be good to add a small sample which clarifies where to import that `testInDebug` helper from and how it is used instead of `test`...

courajs: I'll link to an existing test using it

pangratz: That's good also! Just in case you haven't thought already about that: I would suggest a link using the specific commit, and not `master`, so the link stays resilient to changes. You can achieve that by pressing `y` to get the canonical form: https://help.github.com/articles/getting-permanent-links-to-files/.

bmac:
Thanks @courajs

PR #4244, opened 2016-03-18
[DOC] Add section on commit tagging to CONTRIBUTING.md

courajs:
2 questions:
- Does ember data use feature flags in the same way as ember.js?
- I left "submit your PR against `emberjs:master`", since people's remote in their `data` repo may be named `emberjs`. I could also change it to `data:master`

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -74,7 +74,18 @@ a test!
 
 4. Make the test pass.
 
-5. Push to your fork and submit a pull request. Please provide us with some
+5. Commit your changes. Please use an appropriate commit prefix.
+If your pull request fixes an issue specify it in the commit message. Some examples:
+
+  ```
+  [DOC beta] Update CONTRIBUTING.md for commit prefixes
+  [FEATURE ds-pushpayload-return] Change `pushPayload` to return a value. #4110
+  [BUGFIX beta] Allow optional spaces when parsing response headers
+  ```
+
+  For more information about commit prefixes see [Commit Tagging](#commit-tagging).
+
+6. Push to your fork and submit a pull request. Please provide us with some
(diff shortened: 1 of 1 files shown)


courajs:
For #4238

CONTRIBUTING.md
```diff
@@ -99,10 +110,48 @@ taken straight from the Ruby on Rails guide:
...
+Documentation commits are tagged as `[DOC channel]` where channel is `canary`,
+`beta`, or `release`. If no release is provided `canary` is assumed. The channel should be the most stable release that this documentation change applies to.
+
+### Security
+
+Security commits will be tagged as `[SECURITY cve]`. Please do not submit security related PRs without coordinating with the security team. See the [Security Policy](http://emberjs.com/security/) for more information.
+
+### Other
+
+In general almost all commits should fall into one of these categories. In the cases where they don't please submit your PR untagged. An Ember contributor will let you know if tagging is required.
```
pangratz: Maybe we should use `An ember-data contributor will let ...` here

pangratz:
> Does ember data use feature flags in the same way as ember.js?

Basically yes, the only difference is that `isEnabled` is imported from `ember-data/-private/features`;


> I left "submit your PR against emberjs:master", since people's remote in their data repo may be named emberjs. I could also change it to data:master

What do you thik of changing this to "submit your PR against the `master` branch of https://github.com/emberjs/data"? I think this should be clear enough where the create PR to...

---

Thank you so much @courajs for helping out here. Much appreciated!!!

@bmac @fivetanley since you are the masters of release ceremony, can you take a look here?

courajs:
I opted to just say "against the `master` branch.", unless you think the url is useful.
They are reading this on the proper repo, after all

bmac:
Looks great. Thanks @courajs

PR #4245, opened 2016-03-18
Tweaks to docs on error responses

pablobm:
### Wut?

Recently I was looking at how to get errors to work, and found the docs slightly confusing. I decided to give it a go and see if I could improve them in some way. This is what I got so far.

Does this make sense? I probably can add some other details and make corrections, but I wanted to fly this past you before spending more time on it. I'm happy to take any feedback you may have.

### Rationale

I removed the example provided at `addon/-private/system/model/errors.js` because I feel it can lead to confusion making people think `DS.Errors` actually expects responses in a specific format.

I think the docs at `addon/adapters/rest.js` need to be more explicit at to what exactly it expects.

I think the JSON example belongs in `addon/serializers/json.js`, with details of what keys are actually used. The JSON-API spec lists a few keys that "MAY" appear here, and users need information as to which ones ember-data actually uses.

### Incidentally...

I have no idea of how to generate the docs from this code. I figured out how to do it for ember, using yuidoc, but I don't know what the procedure is for ember-data. Hence, I haven't been able to read the results, and my Markdown may have errors I haven't been able to spot.

### DIFF
--- a/addon/adapters/rest.js
+++ b/addon/adapters/rest.js
@@ -27,6 +27,24 @@ const {
   This adapter is designed around the idea that the JSON exchanged with
   the server should be conventional.
 
+  ## Success and failure
+
+  The REST adapter will consider a success any response with a status code
+  of the 2xx family ("Success"), as well as 304 ("Not Modified"). Any other
+  status code will be considered a failure.
+
+  On success, the request promise will be resolved with the full response
+  payload.
+
+  Failed responses with status code 422 ("Unprocessable Entity") will be
+  considered "invalid". The response will be discarded, except for the
+  `errors` key. The request promise will be rejected with a `DS.InvalidError`.
+  This error object will encapsulate the saved `errors` value.
+
+  Any other status codes will be treated as an "adapter error". The request
+  promise will be rejected, similarly to the "invalid" case, but with
+  an instance of `DS.AdapterError` instead.
+
   ## JSON Structure
 
   The REST adapter expects the JSON returned from your server to follow
@@ -103,6 +121,24 @@ const {
     }
   }
   ```
+
+  ### Errors
+
+  If a response is considered a failure, the JSON payload is expected to include
+  a top-level key `errors`, detailing any specific issues. For example:
+
+  ```js
+  {
+    "errors": {
+      "msg": "Something went wrong"
+    }
+  }
+  ```
+
+  This adapter does not make any assumptions as to the format of the `errors`
+  object. It will simply be passed along as is, wrapped in an instance
+  of `DS.InvalidError` or `DS.AdapterError`. The serializer can interpret it
+  afterwards.
 
   ## Customization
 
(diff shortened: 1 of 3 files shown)


bmac:
Looks great @pablobm. :+1:

bmac:
To generate the API docs there is a `rake generate_ember_data_docs` command in the [website repo](https://github.com/emberjs/website/blob/master/Rakefile#L89) that assumes the Ember Data repo is an a folder called `ember-data` that lives adjacent to the folder where the `website` is checked out.

addon/adapters/rest.js
```diff
@@ -104,6 +124,24 @@ const {
...
+  ```js
+  {
+    "errors": {
+      "msg": "Something went wrong"
+    }
+  }
+  ```
+
+  This adapter does not make any assumptions as to the format of the `errors`
+  object. It will simply be passed along as a black box, wrapped in an instance
```
pangratz: Hmm, IMHO "be passed along as a black box, wrapped ..." is a little confusing here. What do you think of "be passed along as is, wrapped ..."?

addon/adapters/rest.js
```diff
@@ -27,6 +27,26 @@ const {
...
+  Failed responses with status code 422 ("Unprocessable Entity") will be
+  considered "invalid". The response will be discarded, except for the
+  `errors` key. The request promise will be rejected with a `DS.InvalidError`.
+  This error object will encapsulate the saved `errors` value.
+
+  Any other status codes will be treated as an "adapter error". The request
+  promise will be rejected, similarly to the "invalid" case, but with
+  an instance of `DS.AdapterError` instead. The rejection will receive
+  a second argument: an object with additional details about the
+  circumstances of the error.
```
pangratz: I think `The rejection will receive a second argument` is a little confusing here, as it sounds to me that the rejection handler takes a second argument: `.then(null, function(error, additonalDetails) {})`.

I think you meant that the `DS.AdapterError` is instantiated with a `message` as a second argument here, right? Or am I misunderstanding something?. I am looking [at this](https://github.com/emberjs/data/blob/e88dfed1e26b554c41029a218ea6dc41b7a786a6/addon/-private/adapters/errors.js#L14)...

addon/adapters/rest.js
pablobm: I like

addon/adapters/rest.js
pablobm: You are right. At some point I got confused and ended up describing the wrong thing.

pablobm:
OK. I'll have another pass during the weekend. Thank you for the feedback.

pablobm:
I'm happy with this now. Any other comments?

pangratz:
LGTM, can you squash all commits into 1, prefixed with `[DOC]`? Thanks!

pablobm:
@pangratz: done

bmac:
Thanks @pablobm.

Issue #3705, opened 2015-08-26
Add ability to rollback specific attributes


Panman8201:
TL;DR `model.rollbackAttributes(['attrA','attrC'])`

It would be nice if the `.rollbackAttributes()` method accepted an array of attribute names that would be rolled back, leaving the others as-is. Ex: `model.rollbackAttributes(['attrA','attrC'])`

So in the example above, attributes 'attrA' and 'attrC' would have their values reverted to the previous, if changed. Other attributes, such as 'attrB', would retain their new value, if changed.

Related state information should properly reflect the rollback; `hasDirtyAttributes` and `changedAttributes()`. If there are other attributes _not_ rolled back ('attrB') that have changed values, then `hasDirtyAttributes` would remain true. And calls to `changedAttributes()` would not return the rolled back attributes, but _would_ return others that were changed.

Passing in an empty array should trigger the default behavior, as though no array was passed in at all, rolling back all attributes.

Panman8201:
Another situation to be aware of, if `model.get('isNew') === true` and `model.rollbackAttributes()`, the model is put in a "deleted" state. So, if `model.rollbackAttributes(['attrA','attrC'])` but "attrB" is also changed, the model should *not* be rolled back into a "deleted" state.

bmac:
Interesting idea @Panman8201. Would you be interested in working on an implementation?

Panman8201:
@bmac I took a look at the [`internal-model.js` `rollbackAttributes()`](https://github.com/emberjs/data/blob/v2.0.0/packages/ember-data/lib/system/model/internal-model.js#L326) but I'm having a tough time following along. The whole internal model thing confuses me. Perhaps with direction I could figure it out.

courajs:
I'd be interested in taking a stab at this. I propose the following:
- We add `model.rollbackAttribute('attr')`.
- This should be equivalent to `model.set('attr', originalValue)`. That is, it should update `isDirty` appropriately, but would not roll a newly created record back into a deleted state. It would be very confusing to change the state of the record based on whether other attributes have been set or not.
- We don't add `model.rollbackAttributes(['attrA', 'attrB'])`. This is debatable, but given the point above, I wouldn't want to muddle the `rollbackAttributes` API. It would act _very_ differently in the array-arg and no-arg cases, and this would be confusing.

Panman8201:
@courajs I'd still love to have this feature. While I'm not part of the ember-data team at all, I wouldn't mind if you took a stab at it!

[...]

PR #4246, opened 2016-03-18
[FEATURE ds-rollback-attribute] Add rolling back of a single model attribute

courajs:
First pass at #3705 
First feature PR!
Is there any subtlety around in-flight records that I'm missing?

### DIFF
--- a/addon/-private/system/model/internal-model.js
+++ b/addon/-private/system/model/internal-model.js
@@ -4,6 +4,7 @@ import RootState from "ember-data/-private/system/model/states";
 import Relationships from "ember-data/-private/system/relationships/state/create";
 import Snapshot from "ember-data/-private/system/snapshot";
 import EmptyObject from "ember-data/-private/system/empty-object";
+import isEnabled from 'ember-data/-private/features';
 
 import {
   getOwner
@@ -434,6 +435,7 @@ InternalModel.prototype = {
     this.record._notifyProperties(dirtyKeys);
 
   },
+
   /*
     @method transitionTo
     @private
@@ -849,4 +851,21 @@ InternalModel.prototype = {
 
     return reference;
   }
-};
+}
+
+if (isEnabled('ds-reset-attribute')) {
+  /*
+     Returns the latest truth for an attribute - the canonical value, or the
+     in-flight value.
+
+     @method lastAcknowledgedValue
+     @private
+  */
+  InternalModel.prototype.lastAcknowledgedValue = function lastAcknowledgedValue(key) {
+    if (key in this._inFlightAttributes) {
+      return this._inFlightAttributes[key];
+    } else {
+      return this._data[key];
+    }
+  };
+}
--- a/tests/unit/model-test.js
+++ b/tests/unit/model-test.js
@@ -428,6 +428,184 @@ test("changedAttributes() works while the record is being updated", function(ass
   });
 });
 
+if (isEnabled('ds-reset-attribute')) {
+  test("resetAttribute() reverts a single attribute to its canonical value", function(assert) {
+    assert.expect(5);
+
+    run(function() {
+      store.push({
+        data: {
+          type: 'person',
+          id: '1',
+          attributes: {
+            name: 'Peter',
+            isDrugAddict: true
+          }
+        }
+      });
+
+      let person = store.peekRecord('person', 1);
+
+      assert.equal(person.get('hasDirtyAttributes'), false, "precond - person record should not be dirty");
+      person.setProperties({
+        name: 'Piper',
+        isDrugAddict: false
+      });
+      assert.equal(person.get('hasDirtyAttributes'), true, "record becomes dirty after setting property to a new value");
+      person.resetAttribute('isDrugAddict');
+      assert.equal(person.get('isDrugAddict'), true, "The specified attribute is rolled back");
+      assert.equal(person.get('name'), 'Piper', "Unspecified attributes are not rolled back");
+      assert.equal(person.get('hasDirtyAttributes'), true, "record with changed attributes is still dirty");
+    });
+  });
+
+  test("calling resetAttribute() on an unmodified property has no effect", function(assert) {
+    assert.expect(5);
+
+    run(function() {
+      store.push({
+        data: {
+          type: 'person',
+          id: '1',
+          attributes: {
+            name: 'Peter',
+            isDrugAddict: true
+          }
+        }
+      });
+
+      let person = store.peekRecord('person', 1);
+
+      assert.equal(person.get('hasDirtyAttributes'), false, "precond - person record should not be dirty");
+      person.set('name', 'Piper');
+      assert.equal(person.get('hasDirtyAttributes'), true, "record becomes dirty after setting property to a new value");
+      person.resetAttribute('isDrugAddict');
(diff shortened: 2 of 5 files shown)


addon/-private/system/model/internal-model.js
```diff
@@ -435,6 +435,7 @@ InternalModel.prototype = {
     this.record._notifyProperties(dirtyKeys);
 
   },
+
```
bcardarella: Probably don't want to introduce additional whitespace

courajs: Actually, all other methods in this file have whitespace separators, so I left it in as a small style fix.

pangratz:
Basically :+1: to the feature! My gut feeling says that this should be handled on the state machine though. Take for example the case where a record in in the invalid state and a [property is reset](https://github.com/emberjs/data/blob/e88dfed1e26b554c41029a218ea6dc41b7a786a6/addon/-private/system/model/states.js#L702-L708): in this case the corresponding error is removed and the record [could become valid](https://github.com/emberjs/data/blob/e88dfed1e26b554c41029a218ea6dc41b7a786a6/addon/-private/system/model/states.js#L706-L708). Not sure if there are other cases, but inFlight might also be a noteworthy scenario, as you @courajs already said.

I hope that I have time this weekend to take a deeper look. Will keep you posted!

pangratz:
Hey @courajs,

finally I have time to get back to you here. I am very sorry for the delay :pensive:.

This has been discussed in the team meeting and we agreed that this functionality would be useful. We think that naming the method `rollbackAttribute` might be confusing though, so we suggest to name it `resetAttribute`. This method should basically set the attribute to the latest inFlight value or the canonical value from the adapter.

Can you update this PR and:

- rename the method
- add test for when an attribute is reset when the record is in flight

Thanks for your work so far! Looking forward to getting this merged!

addon/-private/system/model/internal-model.js
```diff
@@ -855,3 +856,16 @@ if (isEnabled('ds-references')) {
...
+if (isEnabled('ds-rollback-attribute')) {
+  InternalModel.prototype.rollbackAttribute = function rollbackAttribute(attributeName) {
+    var attrs = this._attributes;
+    var dirtyKeys = Object.keys(attrs);
+    if (dirtyKeys.indexOf(attributeName) !== -1) {
+      delete this._attributes[attributeName];
+      this.send('propertyWasReset', attributeName);
+      this.record.notifyPropertyChange(attributeName);
+    }
+  }
```
pangratz: To leave all the logic to the state machine I would suggest adding a new method to the `InternalModel` which returns the latest truth for an attribute: either the inFlight value or the canonical value. Then, in `DS.Model#resetAttribute` we simply get the latest value from the internal model and `set` it. By this everything should basically behave the same as if the attribute would be reset manually to the latest value...

What do you think?

courajs: Good suggestion!

courajs:
@pangratz Sounds good to me

pangratz:
@courajs kindly asking if you had any chance of getting this up to date? I would love to see this feature in `master` 😉  . Ping me if you need further assistance, happy to help!

courajs:
@pangratz Thanks for the ping! This is mostly updated now, but I bumped into an existing bug with in-flight records we need to fix first. I'll have a PR for that soon and I'll link it here

courajs:
I believe this build only failed because of #4379, and when it's merged these tests will pass

courajs:
Confirmed locally, a rebase on top of #4379 and the tests pass.

courajs:
@pangratz I think we're pretty much good here! Any final feedback?

pangratz:
I meant to review but got caught up in work. I will try to review not later than saturday. Thanks for the work so far!

courajs:
sounds good :)

addon/-private/system/model/internal-model.js
```diff
@@ -849,4 +851,21 @@ InternalModel.prototype = {
...
+
+if (isEnabled('ds-reset-attribute')) {
+  /*
+     Returns the latest truth for an attribute - the canonical value, or the
+     in-flight value.
+
+     @method latestTrueValue
+     @private
+  */
+  InternalModel.prototype.latestTrueValue = function latestTrueValue(key) {
```
pangratz: Though this is an internal method, I am not not too happy about the name here to be honest. Coming up with a better one is difficult though 😔  What do you think of `lastPreassignedAttributeValue()` or even `lastCanonicalOrInFlightAttributeValue()` or something? I know this is a tiny nitpick, but I was hoping we find a better name here...

tests/unit/model-test.js
```diff
@@ -427,6 +427,107 @@ test("changedAttributes() works while the record is being updated", function(ass
...
+          type: 'person',
+          id: '1',
+          attributes: {
+            name: 'Peter',
+            isDrugAddict: true
+          }
+        }
+      });
+
+      store.findRecord('person', 1).then(function(person) {
```
pangratz: Let's use `peekRecord` here: by this we can get rid of the `.then` and overwriting the `shouldBackgroundReloadRecord()`, which makes the test more readable.

tests/unit/model-test.js
```diff
@@ -427,6 +427,107 @@ test("changedAttributes() works while the record is being updated", function(ass
...
```
pangratz: Same as above.

addon/-private/system/model/internal-model.js
courajs: I could settle for `lastCanonicalOrInFlightAttributeValue`. What about `latestSavedValue` or `latestComittedValue`? Since it's not necessarily canonical yet, but it's trying to be.

pangratz: I tend to favor the - admittedly verbose but expressive - `lastCanonicalOrInFlightAttributeValue`. We can change that always later once we find a better name...

pangratz:
Overall, this looks very good 🚀 

Some few additions and this is ready to :shipit: 

- please add a test which asserts that nothing fancy happens if we reset an attribute, which hasn't changed and a test for reseting an attribute which doesn't exist (just want to make sure a future regression are caught)
- add another test reseting an in flight attribute for a record which has been saved twice: the value of the attribute should be the one of the latest `save()`
- add an entry to `FEATURES.md` which outlines the basic usage

Thanks!

courajs:
👍

courajs:
@pangratz updated!

addon/-private/system/model/internal-model.js
pangratz: Let's use `lastAcknowledgedValue` here. This should be clear enough.

pangratz:
This just needs a rename for the `latestTrueValue` method and a rebase onto latest master, then this PR is good to go 🚀

courajs:
Whoops! I thought I did that rename 😅
On it

leifdejong:
@courajs how is this feature going?

courajs:
Just needs the rename mentioned above and a rebase when I get a chance

courajs:
@pangratz updated!

courajs:
@pangratz I think we're finally ready for a merge here. Look good?

pangratz:
Awesome! Thanks for the hard work on this @courajs 🚀  🎉

PR #4260, opened 2016-03-22
Modified the setup-ember-dev test helper to use `ember-metal/debug`s override hooks

workmanw:
I've been helping to "addonify"® [ember-data-model-fragments](https://github.com/lytics/ember-data-model-fragments) which also uses the `ember-dev` test suite. I noticed that `expectNoDeprecations` was no longer working as expected. After investigating I discovered that because Ember no longer uses `Ember.deprecate` internally, overriding that has little effect. I checked to see how ember-data's test suite handled this and discovered you have the same problem. See: [setup-ember-dev.js#L7-L13](https://github.com/emberjs/data/blob/master/tests/helpers/setup-ember-dev.js#L7-L13).

This PR fixes that issue by using `ember-metal/debug` proper hooks. This PR is meant to keep backwards compatibility and work with older versions of ember which did not offer these hooks. If the feeling is that ember and ember-data usage is truly lockstep, this PR could be simplified by removing the `getDebugFunction` and `setDebugFunction ` all together. Example: [ember-data-model-fragments/.../setup-ember-dev.js](https://github.com/lytics/ember-data-model-fragments/blob/addonify/tests/helpers/setup-ember-dev.js#L21-L25).

~~Lastly, I ran the tests locally and there were quite a few failures due to deprecations spewing out. This PR was intended to start a discussion about that.~~ Temporary insanity :stuck_out_tongue_winking_eye: .

### DIFF
--- a/tests/helpers/setup-ember-dev.js
+++ b/tests/helpers/setup-ember-dev.js
@@ -4,12 +4,26 @@ import EmberTestHelpers from "ember-dev/test-helper/index";
 
 const AVAILABLE_ASSERTIONS = ['expectAssertion', 'expectDeprecation', 'expectNoDeprecation', 'expectWarning', 'expectNoWarning'];
 
+// Maintain backwards compatiblity with older versions of ember.
+var emberDebugModule;
+if (Ember.__loader && Ember.__loader.registry && Ember.__loader.registry["ember-metal/debug"]) {
+  emberDebugModule = Ember.__loader.require('ember-metal/debug');
+}
+
 function getDebugFunction(name) {
-  return Ember[name];
+  if (emberDebugModule && emberDebugModule.getDebugFunction) {
+    return emberDebugModule.getDebugFunction(name);
+  } else {
+    return Ember[name];
+  }
 }
 
 function setDebugFunction(name, func) {
-  Ember[name] = func;
+  if (emberDebugModule && emberDebugModule.setDebugFunction) {
+    emberDebugModule.setDebugFunction(name, func);
+  } else {
+    Ember[name] = func;
+  }
 }
 
 var originalModule = QUnit.module;


bmac:
This is great @workmanw! What are the next steps? Do you want some help fixing the failures due to deprecations?

tests/helpers/setup-ember-dev.js
```diff
@@ -2,14 +2,24 @@
...
 import EmberTestHelpers from "ember-dev/test-helper/index";
 
+var debugModule = Ember.__loader && Ember.__loader.require('ember-metal/debug');
+
 const AVAILABLE_ASSERTIONS = ['expectAssertion', 'expectDeprecation', 'expectNoDeprecation', 'expectWarning', 'expectNoWarning'];
 
 function getDebugFunction(name) {
-  return Ember[name];
+  if (debugModule && debugModule.getDebugFunction) {
+    return debugModule.getDebugFunction();
```
pangratz: Is this missing the `name` being passed to `getDebugFunction`?

rwjblue: Ya, I think so...

workmanw: Ooops! Good catch. I'll fix that.

workmanw:
@bmac I guess I'm not completely sure about what the next steps should be. Maybe we should catalog the deprecations and see how easy they are to solve.

pangratz:
Basically I am :100: :+1: on this. My initial implementation was never the way to go anyway and using `ember-meta/debug` is definitely the way to go. Thanks for this nicer solution!!

> Lastly, I ran the tests locally and there were quite a few failures due to deprecations spewing out. This PR was intended to start a discussion about that.

I am not sure I follow completely: if I run the code in this PR locally via `ember try:testall`, all scenarios are working and no uncaught deprecations are causing test failures (well, to be exact, I had to update [this line](https://git.io/vabW4) and add a check that the `ember-metal/debug` module is available before requiring it; it looks like this is needed because this module is not available in the ember 1.13 scenario).

workmanw:
@pangratz Yea ... you're absolutely right. I swear there were test failures last night, but they're all gone today. ¯\\\_(ツ)\_/¯.

I'll push up a change to get ember 1.13 working.

workmanw:
@bmac @pangratz Sorry for the confusion on my part here. I think this commit is actually good to go. No rush or anything. I just wanted to be clear that there actually aren't any deprecations. Just my craziness.

bmac:
Thanks @workmanw

PR #4264, opened 2016-03-25
Moved the adapter errors into the public API space. 

workmanw:
These errors are documented as public and required for people implementing a customer adapter. Example: http://emberjs.com/api/data/classes/DS.InvalidError.html#stq=&stp=0

We have to use them in our app so I feel like I should be able to import them from a public import path.

I wasn't sure if I should have moved the `errors.js` file from the private space or just created a new one in the public space that imported from the private space and re-exported. I decided to do the former, but I'd be happy to do the later.


bmac:
Looks good to me. @pangratz and @tchak do you mind taking a look at this pr as well?

pangratz:
LGTM :+1: 

@workmanw can you add corresponding tests to [tests/unit/modules-test.js](https://github.com/emberjs/data/blob/master/tests/unit/modules-test.js)?

workmanw:
@pangratz Done! 

Side note, I'm in the emberconf contributors workshop now. @mixonic showed a slide of the ember-data repo's file structure. If we hurry, we can invalidate that slide before the workshop is over. :trollface:

bmac:
Thanks @workmanw

workmanw:
Excellent. Thanks @bmac!

PR #4265, opened 2016-03-27
[BUGFIX release] set babel options on addon instance

fivetanley:
When being built by ember-cli-fastboot, or anything that does multiple
builds, storing the babel options on `this` caused Babel to complain
that plugins were being registered twice (whenever the app was rebuilt).
By moving that to `this.addon`, we can avoid this error and allow for
multiple instances.

### DIFF
--- a/index.js
+++ b/index.js
@@ -26,6 +26,12 @@ module.exports = {
   },
 
   init: function() {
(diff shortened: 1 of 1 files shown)


fivetanley:
this is not the right approach.

PR #4271, opened 2016-03-28
[BUGFIX beta] Ensure `null` is returned for Reference#value()

pangratz:
The documentation states that `null` is returned for `Reference#value()`,
when it is not yet loaded. This commit fixes the bug where `undefined`
has been returned for `BelongsToReference#value()`.


bmac:
:+1:

PR #4272, opened 2016-03-28
[BUGFIX beta] Don't unnecessarily materialize records

pangratz:
This adds a `store#_push` method which returns DS.InternalModel's
instead of materialized DS.Model instances. This allows us to defer the
materialization process of an InternalModel until the corresponding
record is actually needed.

### DIFF
--- a/addon/-private/system/references/belongs-to.js
+++ b/addon/-private/system/references/belongs-to.js
@@ -66,8 +66,8 @@ BelongsToReference.prototype.push = function(objectOrPromise) {
 BelongsToReference.prototype.value = function() {
   var inverseRecord = this.belongsToRelationship.inverseRecord;
 
-  if (inverseRecord && inverseRecord.record) {
-    return inverseRecord.record;
+  if (inverseRecord && inverseRecord.isLoaded()) {
+    return inverseRecord.getRecord();
   }
 
   return null;
--- a/addon/-private/system/store.js
+++ b/addon/-private/system/store.js
@@ -1295,7 +1295,15 @@ Store = Service.extend({
     assert("You tried to make a query but you have no adapter (for " + typeClass + ")", adapter);
     assert("You tried to make a query but your adapter does not implement `queryRecord`", typeof adapter.queryRecord === 'function');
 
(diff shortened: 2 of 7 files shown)


addon/-private/system/references/belongs-to.js
```diff
@@ -58,7 +58,12 @@ BelongsToReference.prototype.push = function(objectOrPromise) {
 
 BelongsToReference.prototype.value = function() {
   var inverseRecord = this.belongsToRelationship.inverseRecord;
-  return inverseRecord && inverseRecord.record;
+
+  if (inverseRecord && inverseRecord.isLoaded()) {
+    return inverseRecord.getRecord();
+  }
```
pangratz: This or #4271 needs to be updated, depending on which is merged first.

bmac:
Looks good @pangratz. 👍

pangratz:
@igorT can you take a look at this PR and check if this is what you had in mind?

addon/-private/system/store.js
```diff
@@ -1909,6 +1917,32 @@ Store = Service.extend({
...
+
+  /**
+    Push some data into the store, without creating materialized records.
+
+    @method _push
+    @private
+    @param {Object} data
+    @return {DS.InternalModel|Array} pushed InternalModel(s)
+   */
+  _push(data) {
```
igorT: Can we name this to refer to internalModel somehow

pangratz: Struggling to find a better name here, any suggestions? `_shallowPush` or `_lightweightPush`? Because `_pushInternalModel` doesn't sound right to me...

pangratz: What do you think of making this a public method `pushRef` or something, which returns `RecordReference`(s)? I think @fivetanley you were mentioning something like that somewhere?

bmac: ping @igorT. I believe @pangratz is looking for some feedback here.

pangratz: I've created [RFC#161](https://github.com/emberjs/rfcs/pull/161) to get a discussion on the `store.pushRef` going...

pangratz:
:dancing_men:

PR #4275, opened 2016-03-28
Deprecates Ember.Date.parse

### LINKED ISSUE #4089: Look into deprecating Ember.Date.parse
Ember Data defines its own `Date.parse` function. https://github.com/emberjs/data/blob/master/addon/-private/ext/date.js#L28-L60. This may not be needed and we should be able to rely on the browser's `Date.parse` function.

See comments in this pr: https://github.com/emberjs/data/pull/4082

zigahertz:
Resolves #4089 

Also-by: Dale Ma, <dale.ma@locusenergy.com>

### DIFF
--- a/addon/-private/ext/date.js
+++ b/addon/-private/ext/date.js
@@ -3,62 +3,25 @@
 */
 
 import Ember from 'ember';
+import { deprecate } from 'ember-data/-private/debug';
 
-/**
-  Date.parse with progressive enhancement for ISO 8601 <https://github.com/csnover/js-iso8601>
-
-  © 2011 Colin Snover <http://zetafleet.com>
-
-  Released under MIT license.
-
-  @class Date
-  @namespace Ember
-  @static
-*/
 Ember.Date = Ember.Date || {};
-
-var origParse = Date.parse;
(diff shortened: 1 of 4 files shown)


bmac:
Looks good to me @zigahertz. I'm going to flag this for discussion at the next Ember Data team meeting to make sure I have buy-in on this deprecation before I merge it.

zigahertz:
@bmac sounds good!

bmac:
Hi @zigahertz. We talked about this pr on the last meeting Ember Data meeting. Overall everyone present was in favor of this change. However there is one edge case that `Ember.Date.parse` does not support that browser's `Date.parse` does. This happens when no timezone information is supplied by the date string. IE `2015-01-01` instead of `2015-01-01T00:00:00.000Z`.

Before merging we would like to add a warning to the `date` transform to warn the user if the the date string that does not contain timezone information is passed into the `date` transform's `deserialize` method.

bmac:
This was merged into Ember Data as https://github.com/emberjs/data/pull/4396/commits/dcafebfc5a00ac6b4e1ea805a70e9ee37628e606

Issue #4217, opened 2016-03-07
Error message on empty response "request for a undefined with id" typeClassKey?


erkie:
The following asserts resolve to weird error messages because the property `typeClassKey` is undefined on the `typeClass` variables. I searched the entire `ember-data` repo and the `ember` core repo for the string `typeClassKey` but couldn't find anything. So I'm not really sure what it should be...

https://github.com/emberjs/data/blob/master/addon/-private/system/store/finders.js#L37
https://github.com/emberjs/data/blob/master/addon/-private/system/store/finders.js#L148
https://github.com/emberjs/data/blob/master/addon/-private/system/store/finders.js#L195
https://github.com/emberjs/data/blob/master/addon/-private/system/store/finders.js#L70

pangratz:
Nice find @erkie! It looks like this hasn't been caught by existing tests for this. Also, this is a good chance to clean up that code a bit. Would you be interested in helping out?

As a first step we should update the existing tests so the expected assertion messages are checked. I would suggest to update the regex in [this line](https://git.io/vaIQ1) to something like `/You made a findRecord request for person with id the-id but the adapter's response did not have any data/`. This test should be failing now. As next the assertion message needs to be updated accordingly [here](https://git.io/vaI7r), where the type of the model can be get via `typeClass.modelName` (this has been changed sometimes in the past and not been updated here). The tests should be passing again.

I haven't looked for the other assertions you pointed out, but I hope that there are also tests for this. So they should be updated accordingly.

Let me know of you are interested in helping out and if you need help somewhere. I think this is an ideal chance to update [`CONTRIBUTING.md`](https://github.com/emberjs/data/blob/master/CONTRIBUTING.md) if there are any obscurities.

Thanks!

bmac:
Fixed via https://github.com/emberjs/data/pull/4276

PR #4276, opened 2016-03-28
[BUGFIX canary] Improve finders assertion messages

Arkham:
As explained [here](https://github.com/emberjs/data/issues/4217) the current assertion messages for finders are outdated because they are still accessing `typeClass.typeClassKey` which has been removed. This PR updates those messages to use `typeClass.modelName` instead, and updates the tests to be more specific.


bmac:
Thanks @Arkham.

Issue #4126, opened 2016-01-28
unloadRecord doesn't disconnect observers on the model


jerel:
Here is a twiddle demonstrating the problem (on canary): https://ember-twiddle.com/1523fd2fae4a21c261d7?numColumns=1&openFiles=models.event.js%2C

I have a scenario where a list of models is shown and each model is responsible for removing itself after it becomes old. However when `unloadRecord` is called the model is removed but the observers stay connected.

alexspeller:
it seems like `unloadRecord` doesn't actually destroy the record - no idea if this is intended or not, the workaround is just:

```
model.unloadRecord();
model.destroy();
```

mitchlloyd:
Perhaps this could be resolved with some updates to the current API documentation which is pretty sparse.

http://emberjs.com/api/data/classes/DS.Store.html#method_unloadAll

wecc:
This should be fixed by #4593

PR #4277, opened 2016-03-28
unloadRecord does not remove observers

ryanpatrickcook:
**Issue:** https://github.com/emberjs/data/issues/4126

This is a work in progress to get conversation going on what the expectation should be. The test shows the issue where when you unload a record from the store, the observers on that object are not torn down. 

* Should `unloadRecord` set the object as destroyed?
* Should `unloadRecord` clear observers?

### DIFF
--- a/addon/-private/system/model/model.js
+++ b/addon/-private/system/model/model.js
@@ -122,8 +122,8 @@ var Model = Ember.Object.extend(Ember.Evented, {
     @type {Boolean}
     @readOnly
   */
-  hasDirtyAttributes: Ember.computed('currentState.isDirty', function() {
-    return this.get('currentState.isDirty');
+  hasDirtyAttributes: Ember.computed('currentState', function() {
+    return this.get('_internalModel.currentState.isDirty');
   }),
   /**
     If this property is `true` the record is in the `saving` state. A
--- a/tests/unit/store/unload-test.js
+++ b/tests/unit/store/unload-test.js
@@ -66,7 +66,7 @@ testInDebug("unload a dirty record asserts", function(assert) {
 });
 
 test("unload a record", function(assert) {
-  assert.expect(5);
+  assert.expect(6);
 
   run(function() {
     store.push({
@@ -87,11 +87,48 @@ test("unload a record", function(assert) {
       });
 
       assert.equal(get(record, 'hasDirtyAttributes'), false, "record is not dirty");
+      assert.equal(get(record, 'isDestroyed'), true, 'record is destroyed');
       assert.equal(get(record, 'isDeleted'), true, "record is deleted");
 
       tryToFind = false;
       return store.findRecord('record', 1).then(function() {
         assert.equal(tryToFind, true, "not found record with id 1");
+      });
+    });
+  });
+});
+
+test("unload a record - observers should be removed", function(assert) {
+  assert.expect(4);
+
+  run(function() {
+    store.push({
+      data: {
+        type: 'record',
+        id: '1',
+        attributes: {
+          title: 'toto'
+        }
+      }
+    });
+
+    store.findRecord('record', 1).then(function(record) {
+      Ember.addObserver(record, 'title', record, function() {
+        assert.ok(false, 'observer for record.title');
+      });
+
+      let observers = Ember.observersFor(record, 'title');
+      assert.equal(observers.length, 1, 'record should have 1 observer');
+
+      run(function() {
+        store.unloadRecord(record);
+      });
+
+      run(function() {
+        let observers = Ember.observersFor(record, 'title');
+        assert.equal(get(record, 'isDeleted'), true, 'record is deleted');
+        assert.equal(get(record, 'isDestroyed'), true, 'record is destroyed');
+        assert.equal(observers.length, 0, 'record should not have observers');
       });
     });
   });
--- a/addon/-private/system/store.js
+++ b/addon/-private/system/store.js
@@ -376,6 +376,7 @@ Store = Service.extend({
   */
   unloadRecord(record) {
     record.unloadRecord();
+    record.destroy();
   },
 
   // ................
@@ -1476,8 +1477,7 @@ Store = Service.extend({
 
       for (let i = 0; i < records.length; i++) {
         record = records[i];
-        record.unloadRecord();
-        record.destroy(); // maybe within unloadRecord
+        this.unloadRecord(record);
       }
 
       typeMap.metadata = new EmptyObject();


bmac:
@ryanpatrickcook I believe unloadRecord should destroy the object and clear observers.

ryanpatrickcook:
That is what I would think as well, but it doesn't appear to be clearing observers. There is a twiddle in the issue linked that shows an example. In the test written here, it also appears that `record.isDestroyed` is false after unloading a record. Maybe the test is not accurate.

pangratz:
Hey @ryanpatrickcook, this has been discussed in this weeks' team meeting and we decided that the record should indeed be destroyed when it's unloaded.

Sorry for the delay, but can you update this PR to do the following:

- Destroy the record within `store.unloadRecord`
- Change [this](https://git.io/vVy5U) code so it re-uses the stores' `unloadRecord` so the record is destroyed too
- Update the test so a) the setup (push to store, add observer), b) the destroying and c) the assertions are in a separate run-loop

After those changes the test should be green and this should be good to go! :rocket:

pangratz:
@ryanpatrickcook do you think you have time to update this PR? Happy to help or take over if you won't have time in the near future...

ryanpatrickcook:
@pangratz I've updated with your input. Let me know what you think. Thanks!

tests/unit/store/unload-test.js
```diff
@@ -86,7 +86,7 @@ test("unload a record", function(assert) {
         store.unloadRecord(record);
       });
 
-      assert.equal(get(record, 'hasDirtyAttributes'), false, "record is not dirty");
```
ryanpatrickcook: Using Ember 1.13.13, after a record was unloaded/destroyed, `get(record, 'hasDirtyAttributes')` returned `undefined`. Since the record was now destroyed, I felt as though checking `hasDirtyAttributes` was no longer valid. Replaced with an `isDestroyed` check. Would it be preferred that `hasDirtyAttributes` responds with `false`?

pangratz: Hmm, not sure what the expected behavior of `hasDirtyAttributes` is here. I will bring this up in the next meeting. Good point @ryanpatrickcook 👍

tests/unit/store/unload-test.js
```diff
@@ -97,6 +97,39 @@ test("unload a record", function(assert) {
...
+        attributes: {
+          title: 'toto'
+        }
+      }
+    });
+
+    store.findRecord('record', 1).then(function(record) {
+      Ember.addObserver(record, 'title', record, function() {
+        assert.ok(false, 'observer for record.title');
+      });
```
pangratz: Can you add an assertion for `obersvers.length` being `1` here? Just to be sure that the assertion below does what we expect it to do...

tests/unit/store/unload-test.js
pangratz: I just checked and it seems that `hasDirtyAttributes` is `false` after a record is destroyed. See [this ember-twiddle](https://ember-twiddle.com/255f0c49b1f4d33bc5b9221317bdaf9c?fileTreeShown=false&numColumns=2&openFiles=routes.application.js%2Ctemplates.application.hbs). Am I missing something?

I think in this case it would be ok to add back the assertion for `hasDirtyAttributes`...

tests/unit/store/unload-test.js
ryanpatrickcook: I like that idea. I'll add that in.

tests/unit/store/unload-test.js
ryanpatrickcook: In `1.13`, it looks like if the assertion for `hasDirtyAttributes` is `undefined` when outside of the run-loop with `unloadRecord`. When inside the run-loop with `unloadRecord`, it returns `false`.

tests/unit/store/unload-test.js
```diff
@@ -84,9 +84,10 @@ test("unload a record", function(assert) {
 
       run(function() {
         store.unloadRecord(record);
+        assert.equal(get(record, 'hasDirtyAttributes'), false, "record is not dirty");
       });
 
-      assert.equal(get(record, 'hasDirtyAttributes'), false, "record is not dirty");
```
pangratz: The reason this assertion is failing is due to the implementation of `hasDirtyAttributes` in [`DS.Model`](https://git.io/vrGlF). It looks like a change to the `currentState.isDirty` dependency of the computed property is triggered when the record is unloaded.

@emberjs/ember-data-contributors are there any objections in changing the implementation so it re-uses the [`retrieveFromCurrentState`](https://git.io/vrGlp) function, like we already do for the other flags like `isDirty`?

pangratz: @ryanpatrickcook so, we decided to change the implementation of the `hasDirtyAttributes` CP on the model, by aligning it to the implementation of the other flags. Can you change the dependent key [here](https://git.io/vrVBG) to be just `currentState` and add back the assertion. After that, this PR is good to go!

addon/-private/system/model/model.js
```diff
@@ -120,8 +120,8 @@ var Model = Ember.Object.extend(Ember.Evented, {
     @type {Boolean}
     @readOnly
   */
-  hasDirtyAttributes: Ember.computed('currentState.isDirty', function() {
-    return this.get('currentState.isDirty');
+  hasDirtyAttributes: Ember.computed('currentState', function() {
+    return this.get('_internalModel.currentState.isDirty');
```
ryanpatrickcook: @pangratz I updated the dependent key in the CP here to `currentState`, but the `hasDirtyAttributes` assertion still failed in `1.13` since `this.get('currentState')` is `undefined`. I updated the returned value to `_internalModel.currentState.isDirty` similar to what `retrieveFromCurrentState` is doing. Will this work?

pangratz: How did you test this? Because we have an `ember-try` scenario for ember.js 1.13 and it looks like the test passes on Travis and AppVeyor.

tchak:
There is a PR https://github.com/emberjs/data/pull/3301 to fix this in a more general way. Is it this old PR still relevant? cc @igorT

addon/-private/system/model/model.js
ryanpatrickcook: @pangratz I switched the return back to `this.get('currentState.isDirty')`. Note the current failure on the branch. I think this should be returning `this.get('_internalModel.currentState.isDirty')`. In my previous question I was just wanting to verify with you that it was appropriate to refer to `_internalModel.currentState.isDirty`. That may have been confusing.

pangratz: Ah, now I see. Sorry for being slow here. Re-reading your comment makes it clear what you meant. Yes, please change it back to `this.get('_internalModel.currentState.isDirty')`. Thanks!

ryanpatrickcook: Great! Thanks!

fivetanley:
I think this is still relevant. @ryanpatrickcook can you rebase onto the latest master, and @pangratz can you take another look?

pangratz:
Sorry this took so long @ryanpatrickcook. Thanks for your patience and this fix!

Issue #2494, opened 2014-11-20
Provide an extension point in JSONSerializer to determine which hasMany relationships should be serialized


rsutphin:
A frequently brought up issue (#2068, #1678, 7f752ad15e, to name a few) is the fact that `JSONSerializer#serializeHasMany` does not serialize `manyToOne` (i.e., bidirectional) hasMany relationships. It is clear from the discussion elsewhere that this is a considered behavior and won't be changed.

The recommended alternative is to use `EmbeddedRecordsMixin` in a custom serializer for each model to indicate which hasManys should be serialized. If you only want some bidrectional hasManys to be treated this way this seems like a fine solution.

However, if you want to serialize _all_ hasMany relationships[1], it adds complexity and cognitive load — every time you add a new hasMany relationship, you have to remember add it to a serializer somewhere else also. So practically, the solution that is used (e.g., in [ember-localstorage-adapter][LS], [ember-indexeddb-adapter][IDB], in my own app using ember-pouch) is to copy `serializeHasMany` from `DS.JSONSerializer` into the application's serializer and add `manyToOne` to the list of relationships to be serialized.

[LS]: https://github.com/kurko/ember-localstorage-adapter/blob/c32477626ce0e7c8447e1ee79eb78e64a8357d70/localstorage_adapter.js#L6
[IDB]: https://github.com/kurko/ember-indexeddb-adapter/blob/0b1c73a1134606ccc9dfc1aadb86af5b4dd9fe4d/packages/indexeddb-adapter/lib/indexeddb_serializer.js#L2

This is less than ideal for all the reasons that copying and pasting code is always less than ideal. It would be great if JSONSerializer could provide an extension point that a customized serializer could use to indicate which hasMany associations should be serialized by default, without requiring that the rest of the implementation be copied also.

I'd be happy to write a PR for this. Question: are the values returned by `determineRelationshipType` considered part of the public API for ember-data and therefore unlikely to change?

[1]: One reason you'd want to do this that I'm aware of is if your read and write formats are the same. This happens for sure when you are writing to a local document store, like PouchDB or IndexedDB.

igorT:
cc @sly7-7 I think you are handling this

sly7-7:
If we consider adding  `{serialize: true}` to the attrs hash to be a solution, then yes the PR I'm working on should handle this.

broerse:
@sly7-7 Looking forward to it.

sly7-7:
[...]

PR #4279, opened 2016-03-28
make shouldSerializeHasMany public #2494

jondayton:
From https://github.com/emberjs/data/issues/2494

### DIFF
--- a/addon/serializers/json-api.js
+++ b/addon/serializers/json-api.js
@@ -571,8 +571,12 @@ const JSONAPISerializer = JSONSerializer.extend({
   */
   serializeHasMany(snapshot, json, relationship) {
     var key = relationship.key;
+    var shouldSerializeHasMany = '_shouldSerializeHasMany';
+    if (isEnabled("ds-check-should-serialize-relationships")) {
+      shouldSerializeHasMany = 'shouldSerializeHasMany';
+    }
 
-    if (this._shouldSerializeHasMany(snapshot, key, relationship)) {
+    if (this[shouldSerializeHasMany](snapshot, key, relationship)) {
       var hasMany = snapshot.hasMany(key);
       if (hasMany !== undefined) {
 
--- a/addon/serializers/json.js
+++ b/addon/serializers/json.js
@@ -843,6 +843,27 @@ var JSONSerializer = Serializer.extend({
   /**
     Check if the given hasMany relationship should be serialized
 
+    @method shouldSerializeHasMany
+    @param {DS.Snapshot} snapshot
+    @param {String} key
+    @param {String} relationshipType
+    @return {boolean} true if the hasMany relationship should be serialized
+  */
+
+  shouldSerializeHasMany(snapshot, key, relationship) {
+    if ((this._shouldSerializeHasMany !== JSONSerializer.prototype._shouldSerializeHasMany)) {
+      deprecate('The private method _shouldSerializeHasMany has been promoted to the public API. Please remove the underscore to use the public shouldSerializeHasMany method.', false, {
+        id: 'ds.serializer.private-should-serialize-has-many',
+        until: '3.0.0'
+      });
+    }
+
+    return this._shouldSerializeHasMany(snapshot, key, relationship);
+  },
+
+  /**
+    Check if the given hasMany relationship should be serialized
+
     @method _shouldSerializeHasMany
     @private
     @param {DS.Snapshot} snapshot
@@ -1196,8 +1217,12 @@ var JSONSerializer = Serializer.extend({
   */
   serializeHasMany(snapshot, json, relationship) {
     var key = relationship.key;
+    var shouldSerializeHasMany = '_shouldSerializeHasMany';
+    if (isEnabled("ds-check-should-serialize-relationships")) {
+      shouldSerializeHasMany = 'shouldSerializeHasMany';
+    }
 
-    if (this._shouldSerializeHasMany(snapshot, key, relationship)) {
+    if (this[shouldSerializeHasMany](snapshot, key, relationship)) {
       var hasMany = snapshot.hasMany(key, { ids: true });
       if (hasMany !== undefined) {
         // if provided, use the mapping provided by `attrs` in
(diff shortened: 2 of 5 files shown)


bmac:
Thanks @jondayton. Since this change is going to add a new method to Ember Data's public api would you mind wrapping it in a [feature flag](https://github.com/emberjs/data/pull/3979)?

jondayton:
@bmac no problem. In that case, since it's removing the private api and adding the public, it may be better to leave the private method and add a public reference under the feature flag

bmac:
Good plan @jondayton.

jondayton:
@bmac added a deprecation warning + test. I've been looking at it again and making this a feature flag will require either warnings in several places or a third private method `__shouldSerializeHasMany` which would sit in the middle to handle the deprecation warning and flag. This seems like a bit of overkill for changing the name of a method, as functionality hasn't really changed, it's just been renamed.

addon/serializers/json.js
```diff
@@ -829,15 +829,21 @@ export default Serializer.extend({
...
     @param {DS.Snapshot} snapshot
     @param {String} key
     @param {String} relationshipType
     @return {boolean} true if the hasMany relationship should be serialized
   */
-  _shouldSerializeHasMany(snapshot, key, relationship) {
-    var relationshipType = snapshot.type.determineRelationshipType(relationship, this.store);
+  shouldSerializeHasMany(snapshot, key, relationship) {
+    if (this._shouldSerializeHasMany) {
+      deprecate('The private method `_shouldSerializeHasMany` has been deprecated. Please use public method `shouldSerializeHasMany`.', false, {
```
fivetanley: I think "The private method `_shouldSerializeHasMany` has been promoted to the public API. Please remove the underscore to use the public `shouldSerializeHasMany` method instead."

igorT:
Needs a feature flag, we should also make a shouldSerializeBelongsTo as well

jondayton:
sounds good. I'll take a pass at it tonight.

bmac:
ping @jondayton do you have time to update this pr?

jondayton:
@bmac updated the pr with a feature flag.

jondayton:
logic goes something like this

`if feature flag enabled ? use public method : use private method`
`if private override exists throw a deprecation warning`

technically we could break people who are overriding the private method without a deprecation warning, but I do think it's better to throw the warning. 

I used `shouldSerializeHasMany` for the public name, `_shouldSerializeHasMany` for the private name that is currently being overridden, and `__shouldSerializeHasMany` for the new private method to use if the feature flag is not enabled

bmac:
Thanks @jondayton. This looks good. Do you mind rebasing this one last time and I will merge it?

addon/serializers/json-api.js
```diff
@@ -571,8 +571,9 @@ const JSONAPISerializer = JSONSerializer.extend({
   */
   serializeHasMany(snapshot, json, relationship) {
     var key = relationship.key;
+    var shouldSerializeHasMany = isEnabled('ds-check-should-serialize-relationships') ? 'shouldSerializeHasMany' : '__shouldSerializeHasMany';
```
pangratz: Unfortunately this doesn't work so the feature is correctly stripped in production. You need to change this to something like:

```js
var shouldSerializeHasMany = '__shouldSerializeHasMany';
if (isEnabled("ds-check-should-serialize-relationships")) {
  shouldSerializeHasMany = 'shouldSerializeHasMany';
}
```

addon/serializers/json.js
```diff
@@ -843,14 +843,35 @@ var JSONSerializer = Serializer.extend({
...
     @param {String} relationshipType
     @return {boolean} true if the hasMany relationship should be serialized
   */
-  _shouldSerializeHasMany(snapshot, key, relationship) {
+  __shouldSerializeHasMany(snapshot, key, relationship) {
+    if (this._shouldSerializeHasMany) {
+      deprecate('The private method _shouldSerializeHasMany has been promoted to the public API. Please enable feature flag `ds-check-should-serialize-relationships` and remove the underscore to use the public shouldSerializeHasMany method.', false, {
+        id: 'ds.serializer.private-should-serialize-has-many',
+        until: '3.0.0'
+      });
```
pangratz: The `deprecate` also needs to be behind a `if (!isEnabled("ds-check-should-serialize-relationships")) {` so we don't log the deprecation if the feature is not enabled..

addon/serializers/json.js
```diff
@@ -843,14 +843,35 @@ var JSONSerializer = Serializer.extend({
...
+
+    @method __shouldSerializeHasMany
     @private
     @param {DS.Snapshot} snapshot
     @param {String} key
     @param {String} relationshipType
     @return {boolean} true if the hasMany relationship should be serialized
   */
-  _shouldSerializeHasMany(snapshot, key, relationship) {
+  __shouldSerializeHasMany(snapshot, key, relationship) {
```
bmac: Lets leave `_shouldSerializeHasMany` unchanged in case anyone is calling `this._super()` in `_shouldSerializeHasMany`.

addon/serializers/json.js
```diff
@@ -843,14 +843,35 @@ var JSONSerializer = Serializer.extend({
...
 
-    @method _shouldSerializeHasMany
+    @method shouldSerializeHasMany
+    @param {DS.Snapshot} snapshot
+    @param {String} key
+    @param {String} relationshipType
+    @return {boolean} true if the hasMany relationship should be serialized
+  */
+  shouldSerializeHasMany(snapshot, key, relationship) {
+    return this.__shouldSerializeHasMany(snapshot, key, relationship);
```
bmac: In here we can check if `this._shouldSerializeHasMany === JSONSerializer.prototype._shouldSerializeHasMany` and only issue a deprecation warning and call `_shouldSerializeHasMany` if they are not equal. See https://github.com/emberjs/data/blob/master/addon/serializers/rest.js#L755-L773 for an example.

bmac:
Sorry @jondayton. @pangratz reminded me that we've been using another pattern for detecting deprecated functions. I added a few comments.

jondayton:
@bmac was out of town for the long weekend but should have some time tonight.

jondayton:
@bmac took another pass at it.

bmac:
Thanks @jondayton

PR #4287, opened 2016-03-29
[FEATURE ds-extended-errors] add DS.ServerError

tchak:
This is to align with `ember-ajax`


pangratz:
:tada:  thanks @tchak

Issue #4207, opened 2016-03-04
Getting error when API returning a null response


jpoiri:
I'm getting this error when saving the model and the API returns a code 201 with an empty response. If the API returns a blank object problems goes away.

I am using ember-data@2.3.0

SyntaxError: Unexpected end of input
    at Object.parse (native)
    at jQuery.parseJSON (http://localhost:4200/assets/vendor.js:7877:14)
    at ajaxConvert (http://localhost:4200/assets/vendor.js:8154:19)
    at done (http://localhost:4200/assets/vendor.js:8570:15)
    at XMLHttpRequest.<anonymous> (http://localhost:4200/assets/vendor.js:8960:9)

bmac:
I've hit this error on several projects. The error is originating from jQuery which attempts to `JSON.parse` the response text if the `Content-Type` header in the response suggests the payload is json.

pangratz:
Looks like this is the intended behavior by jQuery: https://jquery.com/upgrade-guide/1.9/#jquery-ajax-returning-a-json-result-of-an-empty-string.

bmac:
Closed via https://github.com/emberjs/data/pull/4290

enspandi:
Just had the same issue, but with HTTP 205 Reset Content.

In my opinion **we always tell jQuery to expect a json payload using the `dataType` option**:
http://api.jquery.com/jquery.ajax/
`dataType: The type of data that you're expecting back from the server.`

RESTAdapter sets `dataType: 'json'`:
https://github.com/emberjs/data/blob/v2.9.0/addon/adapters/rest.js#L1048

Overriding `ajaxOptions` with the 'auto' dataType works for me:
`dataType: '*'`

Interestingly jQuery seems to have custom implementation for 204 No Content but not for 205 Reset Content which also must not have a payload per spec :(

Is there a reason for this? It seems to be there since first commit in 2012...
https://github.com/emberjs/data/blob/v2.9.0/addon/adapters/rest.js#L1048

PR #4290, opened 2016-03-30
[BUGFIX beta] add custom warning when invalid JSON is returned

pangratz:
When the server returns an empty string for a response which doesn't
indicate that the content is empty, jQuery tries to parse the JSON which
results in an error, which is not very helpful at first:

    JSON.parse("") === "SyntaxError: Unexpected end of input"

This change adds a custom warning when such a response is returned, so
the feedback is more valuable.

---

This addresses #4207


bmac:
Awesome. I'm excited for this change.

Issue #3688, opened 2015-08-21
destroyRecord fails for a new unsaved record if it is not valid


sergetab:
After creating a new record with store.createRecord I might need to destroy it if user prefers not to save. If the new record is valid then destroyRecord deletes it (but the record still remains in the store???). But if the record is NOT valid then store is trying to make a DELETE request to the server. The url in this case is wrong, e.g. for a 'user' it makes a call to api/users/ without any id (new record doesn't have an id).
After some debugging  I think the problem is [here](https://github.com/emberjs/data/blob/master/packages/ember-data/lib/system/store.js#L1318). 'currentState' in case of invalid record would be "root.deleted.inFlight", not 'root.deleted.saved'

typeoneerror:
Just saw this behavior in 2.4.3. Call `save()` on a record. Server responds with 422 error. Then call `destroyRecord` and a DELETE call is made to `api/users`, not like `api/users/1`.

pangratz:
@typeoneerror the bugfix from #4293 is available since `2.6`. Can you upgrade ember-data to that version and check if the error persists? Thanks!

PR #4293, opened 2016-03-31
Transition an invalid created record to the deleted saved state when …

### LINKED ISSUE #4289: Strange behaviour of `deleteRecord` after validation error
I'm seeing a strange behaviour that might be a bug. `Model#deleteRecord` removes the record from the store, unless there has been an **unsuccessful** attempt to save it.

To reproduce in an app, first create a record on the store. From there, the behaviour of `deleteRecord` varies depending on whether `save` is called on the record or not, and what the result of this was. Presently:

  * No attempt to save: removes the record form the store
  * After a successful save: removes the record from the store
  * After a failed save: apparently nothing

Here's a twiddle that reproduces the behaviour: https://ember-twiddle.com/e820604c62b94a5828d6890d2e37efe2?fileTreeShown=false&numColumns=2&openFiles=templates.application.hbs%2Ccontrollers.application.js

From the Twiddle, this is the important part (the model name is `line`):

```js
import Ember from 'ember';

  init() {
    this.store.createRecord('line', {name: "yksi"});
    this.store.createRecord('line', {name: "kaksi"});
    this.store.createRecord('line', {name: "kolme"});
  },
  
  lines: Ember.computed(function() {
    return this.store.findAll('line');
  }),
  
  actions: {
    invalidate(line) {
      // Recklessly using private APIs in order
      // to reproduce the problem with minimum code
      this.store.recordWasInvalid(
        line._internalModel, {
          name: "is not up to scratch",
        }
      );
    },
    
    delete(line) {
      line.deleteRecord();
    },
  },
});
```

The action `delete` removes the record from the store. However, this doesn't work if we previously call `invalidate`. I would try to mimic the case of the successful save too, but can't work out the private APIs for that.

bmac:
…deleted.

Closes #4289

### DIFF
--- a/tests/integration/records/delete-record-test.js
+++ b/tests/integration/records/delete-record-test.js
@@ -8,6 +8,7 @@ import DS from 'ember-data';
 var attr = DS.attr;
 var Person, env;
 var run = Ember.run;
+var get = Ember.get;
 
 module("integration/deletedRecord - Deleting Records", {
   beforeEach() {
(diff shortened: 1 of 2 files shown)


tests/unit/model-test.js
```diff
@@ -487,6 +487,27 @@ test("currentState is accessible when the record is created", function(assert) {
...
+  assert.equal(get(store.peekAll('person'), 'length'), 1, 'The new person should be in the store');
+
+  run(function() {
+    record.deleteRecord();
+  });
+  
+  assert.equal(get(record, 'currentState.stateName'), 'root.deleted.saved');
+  assert.equal(get(store.peekAll('person'), 'length'), 0, 'The new person should be removed from the store');
+});
+
```
pangratz: Just curious: any specific reason why this is not in [tests/integration/records/delete-record-test.js](https://github.com/emberjs/data/blob/master/tests/integration/records/delete-record-test.js)?

bmac: Nope. I will move it.

tests/integration/records/delete-record-test.js
```diff
@@ -149,3 +150,25 @@ test("when deleted records are rolled back, they are still in their previous rec
...
+                 'records should start in the created.invalid state');
+  assert.equal(get(store.peekAll('person'), 'length'), 1, 'The new person should be in the store');
+
+  run(function() {
+    record.deleteRecord();
+  });
+
+  assert.equal(get(record, 'currentState.stateName'), 'root.deleted.saved');
+  assert.equal(get(store.peekAll('person'), 'length'), 0, 'The new person should be removed from the store');
+});
```
pangratz: Can you add a test case for `destroyRecord` too? I think this fix should also work for #3688.

bmac: Added a test case for `destroyRecord`.

pangratz:
Looks good :+1:

Tiny nitpick: I would prefer using the adapter instead of `store.recordWasInvalid` so the whole stack is run through, which might make this tests more robust for eventual future changes. But again, this is just my personal preference. Current approach is fine too :smiley: .

bmac:
@pangratz I've updated the test for this pr to not use the private `store.recordWasInvalid` method.

pangratz:
Nice @bmac, feel free to merge once :green_apple:

PR #4295, opened 2016-04-01
[DOC release] Mistaken back tick.

wizvishak:
Hi! I was going through the docs and found an apostrophe(') mistaken for a back tick(`). Set that right.

You guys are doing some amazing work :clap:


bmac:
Thanks @wizvishak

PR #4299, opened 2016-04-05
Fix example code on normalize method

ahmadsoe:
The push method only accepts one parameter.


pangratz:
Good :eyes: , thanks @ahmadsoe :rocket:

Issue #4255, opened 2016-03-21
queryRecord returns array, expect single record


MarkMT:
According to the description of `queryRecord` in the [guides](https://guides.emberjs.com/v2.4.0/models/finding-records/)  (emphasis added) -

> If you know your query will return only one result Ember Data provides a convenience method that will return a promise that resolves with **a single record**. Calling `store.queryRecord()` will make a `GET` request with the passed object serialized as query params.

However what I'm seeing in practice is that the return value actually resolves to an _array_ with as many elements as fit the query (and it appears to be a regular javascript array not an Ember.Array). So if I want to use the method in a model I have to do something like this -
```
export default Ember.Route.extend({
  model() {
    return this.store.queryRecord('person', {filter: {name: 'foo'}}).then(function(result) {
      return result[0].get('data');
    });
  }
});
```
It works, but it doesn't seem consistent with what's advertised. The only difference I'm seeing from `query` is that the latter returns an Ember.ArrayProxy rather than a javascript array.

MarkMT:
Seems similar to #4227 but I don't think the discussion there resolves my concern.

MarkMT:
Hmm, looks like I'm on ember 2.3. Maybe that's what the difference is.

MarkMT:
Nope. Just upgraded, same behavior.

pangratz:
What does your response look like?

MarkMT:
If I add this to my model hook:
```
    this.store.queryRecord('person', {filter: {name: 'foo'}}).then(function(result) {
      console.log(result);
      console.log(result[0].get('data'));
    });
```
I get this in the console:

```
[Class]
0: Class
length: 1
__proto__: Array[0]
etc...

EmptyObject {name: "foo"}
```

pangratz:
I mean what does your server return when `this.store.queryRecord('person', { ... })` is invoked?

`store.queryRecord` expects the server to return a single resource. I looks like the response in your case is an array. If your server returns an array, you should use `store.query` instead:

``` jas
this.store.queryRecord('person', {filter: {name: 'foo'}}).then(function(result) {
  return result[0];
});
```

MarkMT:
Sorry... the server returns an array even if there is only one element (which it is). It's a vanilla rails 5 api server.

MarkMT:
If an array is the intended result when the server returns a one element array (a) the method doesn't seem quite as useful as I'd hoped and (b) the guides are a little misleading and I'll file an issue there.

[...]

Issue #4227, opened 2016-03-13
queryRecord doesn't seem to return a fullly hydrated model


samselikoff:
Working on a brand new 2.4 Ember app and I have the following route:

```js
import Ember from "ember";

export default Ember.Route.extend({

  model(params) {
    return this.store.queryRecord('video', {
      slug: params.slug
    });
  },

  afterModel(model) {
    debugger;
  }

});
```

`model.toString()` suggests I indeed have the model, but `model.get('title')` in `afterModel` (as well as `{{model.title}}` in my template) is blank. However, if I look in the Ember Inspector I see the data is there:

![image](https://cloud.githubusercontent.com/assets/2922250/13729126/f51b1cae-e903-11e5-8126-5f19a4718e22.png)

If I change the `model` hook to

```js
model(params) {
  return this.store.query('video', {
    slug: params.slug
  }).then(models => models.get('firstObject'));
}
```

then everything works.

The bug is there on `master` as well.

Here's a JSBIN: http://emberjs.jsbin.com/dajisaqufe/edit?html,js,console,output

sly7-7:
I'm not sure, but I think `queryRecord`expects a single record to be returned from the server. 
Maybe I'm missing something, but http://emberjs.jsbin.com/zagohifevu/edit?html,js,console,output seems to work as expected

pangratz:
If I recall correctly there was once the implementation where the first item is used, if an array is returned (also see last comment by @bcardarella [here](https://github.com/emberjs/data/issues/3977#issuecomment-163163862)). It looks like this is [no more the case](https://github.com/emberjs/data/blob/v2.4.0/addon/serializers/json.js#L450-L454)...

samselikoff:
Ah I see, ok closing for now. Thanks!

pangratz:
@samselikoff what adapter are you using? I assume `JSON-API`?

samselikoff:
Yep

pangratz:
For clarification of my comment above:

> If I recall correctly there was once the implementation where the first item is used, if an array is returned 

This is currently the case for the `rest` adapter, which returns the first object when `queryRecord` returns an array.

---

#4300 addressed some issues with `queryRecord` and the outcome is that `store.queryRecord` should only be used, when a single record is returned and the id is not known beforehand. For all other cases `store.query` in combination with returning the first object in the array is encouraged. Assertions added in #4300 should now inform you when an array is returned for `queryRecord`.

[...]

Issue #3977, opened 2015-12-08
queryRecord is not guaranteed to return a single record


bcardarella:
https://github.com/emberjs/data/blob/master/addon/system/store/finders.js#L193

push returns a single model for objects and an array of records for arrays, so if the response is a collection of records instead of a single record the result of `queryRecord` will be a collection. The documentation implies that this function will always return a single record.

If the position of the ember-data team is to say "the endpoint should return a single record" I'd like to challenge this.

The url being built by `queryRecord` is identical to the one being built by `query`. From the backend's perspective there may be little control over how to shape the response, even if we assume one of the query criteria to force a unique response. I would advocate for `queryRecord` to force a collection from `push` to return `firstObject`.

fivetanley:
I agree.

bcardarella:
@fivetanley I can put together a PR for this. Should I target different versions or just master?

fivetanley:
Just master and I'll let you know if it doesn't cleanly apply to beta.

bcardarella:
We must have something else going on, it appears that `normalizeSingleResponse` will force all records not firstObject into the `included` key which satisfies this issue.

PR #4300, opened 2016-04-05
[BUGFIX beta] Overhaul queryRecord

pangratz:
Currently the use case for `store.queryRecord` is not communicated very
well and there has been some confusion on what the expected server
response should look like (array, or an object).

There are several issues with the current code base:

- in general, if the serializer returns an array for the primary data
  returned by `normalizeQueryRecordResponse`, then `store.queryRecord`
  resolves with an array

- if rest serializer is used and an array is returned by the adapter,
  then the first entry of the array is used as primary data and
  `store.queryRecord` resolves with the first record

- if json-api serializer is used and the primary data is an array, then
  `store.queryRecord` resolves with an array of the primary records

- the API documentation for `queryRecord` is similar to `query` and
  doesn't indicate when this method should be used in contrast to
  `store.query` and getting the first record of the array

- an assertion for the payload returned by the adapter for `queryRecord`
  not being an empty object has been added in 2.4, which is a regression
  to the behavior of 2.3

-----------------------------------------------------------------------

This commit addresses the above issues and makes the following changes:

- add assertion that `store.queryRecord` never resolves with an array

- add deprecation warning for the rest-serializers' `queryRecord`, if an
  array is returned instead of a single record

- removes the assertion that the returned payload from the adapter for
  `queryRecord` is not an empty object

- add assertion within json-api serializer that the primary data of the
  normalized response is not an array

---

I am not so sure about the deprecation for the `rest-serializer`, since it looks like currently also `findRecord` works when an array is returned (see [this test](https://github.com/emberjs/data/blob/v2.4.3/tests/integration/adapter/rest-adapter-test.js#L94-L114)). But since `json-api-adapter` doesn't allow `data: [...]` responses for `queryRecord`, this seems like an inconsistency which is a little surprising IMHO `¯\_(ツ)_/¯` .

This addresses issues raised in #3977, #4227 and #4255.

### DIFF
--- a/addon/-private/system/store.js
+++ b/addon/-private/system/store.js
@@ -1043,20 +1043,96 @@ Store = Service.extend({
   },
 
   /**
-    This method delegates a query to the adapter. This is the one place where
-    adapter-level semantics are exposed to the application.
+    This method makes a request for one record, where the `id` is not known
+    beforehand (if the `id` is known, use `findRecord` instead).
 
-    Exposing queries this way seems preferable to creating an abstract query
-    language for all server-side queries, and then require all adapters to
-    implement them.
+    This method can be used when it is certain that the server will return a
+    single object for the primary data.
 
-    This method returns a promise, which is resolved with a `RecordObject`
-    once the server returns.
+    Let's assume our API provides an endpoint for the currently logged in user
+    via:
+
+    ```
+    // GET /api/current_user
+    {
+      user: {
+        id: 1234,
+        username: 'admin'
+      }
+    }
+    ```
+
+    Since the specific `id` of the `user` is not known beforehand, we can use
+    `queryRecord` to get the user:
+
+    ```javascript
+    store.queryRecord('user', {}).then(function(user) {
+      let username = user.get('username');
+      console.log(`Currently logged in as ${username}`);
+    });
+    ```
+
+    The request is made through the adapters' `queryRecord`:
+
+    ```javascript
+    // app/adapters/user.js
+    import Adapter from "ember-data/adapter";
+
+    export default Adapter.extend({
+      queryRecord(modelName, query) {
+        return Ember.$.getJSON("/api/current_user");
+      }
+    });
+    ```
+
+    Note: the primary use case for `store.queryRecord` is when a single record
+    is queried and the `id` is not kown beforehand. In all other cases
+    `store.query` and using the first item of the array is likely the preferred
+    way:
+
+    ```
+    // GET /users?username=unique
+    {
+      data: [{
+        id: 1234,
+        type: 'user',
+        attributes: {
+          username: "unique"
+        }
+      }]
+    }
+    ```
+
+    ```javascript
+    store.query('user', { username: 'unique' }).then(function(users) {
+      return users.get('firstObject');
+    }).then(function(user) {
+      let id = user.get('id');
+    });
+    ```
+
+    This method returns a promise, which resolves with the found record.
+
+    If the adapter returns no data for the primary data of the payload, then
+    `queryRecord` resolves with `null`:
+
+    ```
+    // GET /users?username=unique
+    {
+      data: null
+    }
+    ```
+
+    ```javascript
+    store.queryRecord('user', { username: 'unique' }).then(function(user) {
+      console.log(user); // null
+    });
+    ```
 
     @method queryRecord
     @param {String} modelName
     @param {any} query an opaque query to be used by the adapter
-    @return {Promise} promise
+    @return {Promise} promise which resolves with the found record or `null`
   */
   queryRecord(modelName, query) {
     assert("You need to pass a model name to the store's queryRecord method", isPresent(modelName));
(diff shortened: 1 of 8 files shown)


addon/-private/system/store.js
```diff
@@ -935,15 +935,75 @@ Store = Service.extend({
...
+    store.queryRecord('user', {}).then(function(user) {
+      let username = user.get('username');
+      console.log(`Currently logged in as ${username}`);
+    });
+    ```
+
+    The request is made through the adapters' `queryRecord`:
+
+    ```javascript
+    // app/adapters/application.js
```
sly7-7: maybe pointing to app/adapters/user.js would be better here ?

addon/-private/system/store.js
```diff
@@ -935,15 +935,75 @@ Store = Service.extend({
...
+
+    ```javascript
+    store.query('user', { username: 'unique' }).then(function(users) {
+      return users.get('firstObject');
+    }).then(function(user) {
+      let id = user.get('id');
+    });
+    ```
+
+    This method returns a promise, which is resolves with the found record.
```
sly7-7: `which is resolved` / or `which resolves`

addon/-private/system/store/finders.js
```diff
@@ -196,6 +196,11 @@ export function _queryRecord(adapter, store, typeClass, query) {
     var record;
     store._adapterRun(function() {
       var payload = normalizeResponseHelper(serializer, store, typeClass, adapterPayload, null, 'queryRecord');
+
+      assert("Expected the primary data returned by the serializer for a `queryRecord` response to be a single object but instead it was an array.", !Array.isArray(payload.data), {
+        id: 'ds.store.queryRecord-array-response'
+      });
+
```
sly7-7: At first sight it seemed that was not necessary here, but I guess it handles the case of customs adapters/serializers, right ?

addon/-private/system/store.js
pangratz: :+1: You're absolutely right!

addon/-private/system/store.js
pangratz: lol, english, how do you even work.

addon/-private/system/store/finders.js
pangratz: Yep :+1:

pangratz:
Thanks for the feedback @sly7-7!

sly7-7:
I think this is a good answer to related issues, disambiguate, and making the things consistent, I'm all for it

pangratz:
This has been discussed in the team meeting and overall the taken approach looks good.

I will take a look at #3916 to see if it can be addressed with this PR as well.

addon/serializers/json-api.js
```diff
@@ -199,6 +199,13 @@ const JSONAPISerializer = JSONSerializer.extend({
   */
   _normalizeResponse(store, primaryModelClass, payload, id, requestType, isSingle) {
     let normalizedPayload = this._normalizeDocumentHelper(payload);
+
+    if (requestType === 'queryRecord' && Array.isArray(normalizedPayload.data)) {
+      assert('Expected the primary data returned by the serializer for a `queryRecord` response to be a single object but instead it was an array.', false, {
+        id: 'ds.serializer.json-api.queryRecord-array-response'
+      });
+    }
```
bmac: What are your thoughts on doing this in `normalizeQueryRecordResponse`?

pangratz:
I've updated this PR to address the following issues:

- put assertion for `queryRecord` being an array in the `json-api` serializer from `_normalizeResponse` to `normalizeQueryRecordResponse`, which is a better place as pointed out by @bmac 👍 

- remove assertion for payload not being empty in the [`_queryRecord` method of `finder.js`](https://git.io/vw3WF). This assertion (introduced in #3916) caused a regression which has been reported in #4310. Removing the assertion allows to return an empty object again, when the `rest` adapter is used. Tests for `store.queryRecord` using the `rest` and `json-api` adapter have been added, so the expected behavior for those methods are tested and future regressions are caught.

- updated the documentation for `store.queryRecord` to point out the behavior: the returned promise resolves with either the found record or `null` when the adapter returned nothing for the primary data

taras:
Just to clarify the clarification, does this PR ensure that `queryRecord` always returns a single object and not an array?

pangratz:
> Just to clarify the clarification, does this PR ensure that queryRecord always returns a single object and not an array?

@taras yes, `store.queryRecord` [will always resolve with a single record](https://github.com/pangratz/data/blob/overhaul-queryRecord/addon/-private/system/store/finders.js#L199-L201). Is that what you meant?

taras:
Cool, thank you :)

addon/serializers/json-api.js
```diff
@@ -202,6 +202,18 @@ const JSONAPISerializer = JSONSerializer.extend({
     return normalizedPayload;
   },
 
+  normalizeQueryRecordResponse() {
+    let normalized = this._super(...arguments);
+
+    if (Array.isArray(normalized.data)) {
+      assert('Expected the primary data returned by the serializer for a `queryRecord` response to be a single object but instead it was an array.', false, {
+        id: 'ds.serializer.json-api.queryRecord-array-response'
+      });
```
bmac: Instead of wrapping this in an if block should we replace `false` with `!Array.isArray(normalized.data)` so it gets stripped correctly is a prod build?

pangratz: Absolutely! Don't know why I chose the if here 😕

bmac:
Thanks @pangratz

Issue #4302, opened 2016-04-06
Options not passed through to adapter


trevorrjohn:
I have this interesting use case where I have two routes that return the same model. For example: `/posts` and `/archived`, both return an array of `Post`. I was hoping to implement the logic in my adapter such that I would switch the url based on the options I passed into the store, something like `DS.Store.findAll('post', { archived: true })` then in my adapter I would have something like 

```javascript
DS.RESTAdapter.extend({
  findAll(store, type, sinceToken, snapshotRecordArray, options) {
    if(options.archived) { /* do something different */ }
  }
})
````

Looking at https://github.com/emberjs/data/blob/master/addon/-private/system/store/finders.js#L140 I see that options are present, but not passed to the adapter. Is this a conscious design choice? If so, what is the reasoning?

Are there other options right now, besides creating a new model that inherits from Post?

Thoughts?

**Comment by sly7-7** [comment]
It looks like the snapshotRecordArray should have the options, as `adapterOptions`property: see https://github.com/emberjs/data/blob/master/addon/-private/system/snapshot-record-array.js#L53

So you should be able to do: 
```
DS.RESTAdapter.extend({
  findAll(store, type, sinceToken, snapshotRecordArray) {
    var options = snapshotRecordArray.adapterOptions;
    if(options.archived) { /* do something different */ }
  }
})
```

Please close this issue if this is working for you :)

**Comment by trevorrjohn** [comment]
That is exactly what I was looking for. Thanks!

**Comment by pangratz** [comment]
@trevorrjohn be aware that you need to pass the options to the `findAll` like this:

```js
this.store.findAll('user', {
  adapterOptions: {
    archived: true
  }
});
```

See [this twiddle for example](https://ember-twiddle.com/3c418e00c6b209416e7bd9ea3fb83bb7?fileTreeShown=false&numColumns=2&openFiles=application.route.js%2Cadapters.application.js).

---

This definitely needs to be documented better so thanks for opening this issue @trevorrjohn. I will open an issue for that.

trevorrjohn:
Thanks, I did notice that. The only issue I am seeing now, is how `Adapter.buildURL` is defined, since the `adapterOptions` aren't passed through into that method, I have to override the full `findAll` method when really only the `buildURL` method needs to change. This is also making me think that this is not the best way to handle this scenario, thoughts?

pangratz:
[...]

PR #4304, opened 2016-04-06
pass DS.SnapshotRecordArray to build-url-mixin buildURL

trevorrjohn:
See discussion: https://github.com/emberjs/data/issues/4302

### DIFF
--- a/addon/-private/adapters/build-url-mixin.js
+++ b/addon/-private/adapters/build-url-mixin.js
@@ -56,7 +56,7 @@ export default Ember.Mixin.create({
       case 'findRecord':
         return this.urlForFindRecord(id, modelName, snapshot);
       case 'findAll':
-        return this.urlForFindAll(modelName);
+        return this.urlForFindAll(modelName, snapshot);
       case 'query':
         return this.urlForQuery(query, modelName);
       case 'queryRecord':
@@ -121,9 +121,10 @@ export default Ember.Mixin.create({
   /**
    * @method urlForFindAll
    * @param {String} modelName
+   * @param {DS.SnapshotRecordArray} snapshot
    * @return {String} url
    */
-  urlForFindAll(modelName) {
+  urlForFindAll(modelName, snapshot) {
     return this._buildURL(modelName);
   },
 
--- a/addon/adapters/rest.js
+++ b/addon/adapters/rest.js
@@ -458,7 +458,7 @@ var RESTAdapter = Adapter.extend(BuildURLMixin, {
 
       return this._makeRequest(request);
     } else {
-      const url = this.buildURL(type.modelName, null, null, 'findAll');
+      const url = this.buildURL(type.modelName, null, snapshotRecordArray, 'findAll');
 
       if (sinceToken) {
         query.since = sinceToken;
--- a/addon/adapters/rest.js
+++ b/addon/adapters/rest.js
@@ -1268,7 +1268,7 @@ if (isEnabled('ds-improved-ajax')) {
 
       switch (requestType) {
         case 'findAll':
-          return this.buildURL(type.modelName, null, null, requestType);
+          return this.buildURL(type.modelName, null, snapshots, requestType);
 
         case 'query':
         case 'queryRecord':
--- a/tests/integration/adapter/rest-adapter-test.js
+++ b/tests/integration/adapter/rest-adapter-test.js
@@ -1033,8 +1033,11 @@ test("findAll - returning an array populates the array", function(assert) {
 });
 
 
-test("findAll - passes buildURL the requestType", function(assert) {
+test("findAll - passes buildURL the requestType and snapshot", function(assert) {
+  assert.expect(2);
+  let adapterOptionsStub = { stub: true };
   adapter.buildURL = function(type, id, snapshot, requestType) {
+    assert.equal(snapshot.adapterOptions, adapterOptionsStub);
     return "/" + requestType + "/posts";
   };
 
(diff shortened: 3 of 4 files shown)


addon/-private/adapters/build-url-mixin.js
```diff
@@ -47,16 +47,17 @@ export default Ember.Mixin.create({
     @param {String} modelName
     @param {(String|Array|Object)} id single id or array of ids or query
     @param {(DS.Snapshot|Array)} snapshot single snapshot or array of snapshots
+    @param {(DS.SnapshotRecordArray)} snapshotRecord single snapshotRecordArray
```
sly7-7: I think you don't have to add this parameter here. It is passed as the `snapshot` argument

addon/-private/adapters/build-url-mixin.js
```diff
@@ -47,16 +47,17 @@ export default Ember.Mixin.create({
...
     @param {String} requestType
     @param {Object} query object of query parameters to send for query requests.
     @return {String} url
   */
-  buildURL(modelName, id, snapshot, requestType, query) {
+  buildURL(modelName, id, snapshot, snapshotRecord, requestType, query) {
```
pangratz: You don't need to pass the `snapshotRecord` here, since `snapshot` will either be the single `snapshot` or the `snapshotRecordArray`.

Sorry, I should have made that clear in the other isse :pensive:

addon/-private/adapters/build-url-mixin.js
trevorrjohn: OK, I was confused with this as I wasn't sure if it was an array of Snapshot or the DS.SnapshotRecordArray. Will update...

trevorrjohn:
@pangratz I think this is more along the lines of what you were looking for.

pangratz:
:heart_eyes: awesome, that was quick! Thanks you very much @trevorrjohn!

One tiny thing left: since this is a bugfix, can you prefix the commit with `[BUGFIX beta]` to make sure this gets into the next `beta` release?

trevorrjohn:
no problem done.

sly7-7:
Hm, I may be wrong but I think adapterOptions should still be passed in buildUrl, like here: https://github.com/emberjs/data/blob/8959c4ed43f48e0c473384fa9b6fca58f7e965f8/addon/adapters/rest.js#L461

trevorrjohn:
you're right, I missed that when I was removing that other code.

pangratz:
Awesome catch @sly7-7. Completely missed that. Looks like this has been removed accidentally when our comments were addressed. Since no tests fail here I think that's a good sign that there is one missing, which tests this scenario.

@trevorrjohn can you pass the `snapshotRecordArray` where @sly7-7 pointed it out and additionally ad a test in `tests/integration/adapter/build-url-mixin-test.js`, which asserts that a snapshot array is passed to `urlForFindAll`?

pangratz:
@trevorrjohn nvm, just saw the test you added. That one looks good too!

sly7-7:
Unfortunately, I don't have the time right now, but I think there are other method where the snapshot is not passed. (here for example https://github.com/emberjs/data/blob/8959c4ed43f48e0c473384fa9b6fca58f7e965f8/addon/adapters/rest.js#L637) I'm not sure if we should pass the snapshots where it's available, but in that case, maybe we could add some integration test to have the complete flow tested (from the store.findXXX methods).

trevorrjohn:
Cool! Yeah I had it in the first commit, but accidentally ripped it out
when I was removing the extra param.

trevorrjohn:
I can look into adding it to the other methods as well.

pangratz:
> I can look into adding it to the other methods as well.

Let's do that in a separate PR

trevorrjohn:
OK do you want one for each method or can I do the rest in  one?

pangratz:
Alright, so this gets tricky now, please bear with me: the tests are failing now since you [added an assertion here](https://github.com/trevorrjohn/data/blob/tj/snapshot-record-array-on-find-all/tests/integration/adapter/rest-adapter-test.js#L1040). This assertion will fail since the snapshot array is not passed to `buildURL` [here](https://git.io/vVrSM), where this code path is taken when the `ds-improved-ajax` feature is enabled.

The feature is not yet available in the `beta` branch, that's why simply passing the snapshot record array [here](https://git.io/vVrSM) would indeed make this PR green, but it would make cherry-pick the bugfix into beta complicated (we only want the code path in the `beta` branch, which is not behind the `ds-improved-ajax` feature).

So, here's what you need to do (sorry in advance :pensive:): pass the snapshot array [here](https://git.io/vVrSM) and commit this change with the message `[FEATURE ds-improved-ajax] pass snapshotRecordArray to urlForFindAll`.

I will add a team discussion flag here to check that I am not going insane. @trevorrjohn I am very sorry for this mess, your first contribution should NOT be that complicated. If you have any further questions ping me on slack. Thanks for sticking through!

Note: technically, if every commit in this PR should have a green CI, you would need to rebase the commits and put the feature commit before the bugfix..

pangratz:
I discussed this with @bmac and the approach of 2 commits is the way to go here. Thanks @trevorrjohn once again for tackling this and thanks @sly7-7 for your :eyes: here! Much appreciated.

Regarding "passing snapshot/snapshots" to `urlForFindHasMany´ / `urlForFindBelongsTo`: @trevorrjohn the approach is the same as in this PR; so first you need to create the commits for the feature and afterwards the actual bugfix.

PR #4308, opened 2016-04-07
pass snapshot through to urlForFindHasMany and urlForFindBelongsTo

trevorrjohn:
related to: https://github.com/emberjs/data/pull/4304

### DIFF
--- a/addon/-private/adapters/build-url-mixin.js
+++ b/addon/-private/adapters/build-url-mixin.js
@@ -64,9 +64,9 @@ export default Ember.Mixin.create({
       case 'findMany':
         return this.urlForFindMany(id, modelName, snapshot);
       case 'findHasMany':
-        return this.urlForFindHasMany(id, modelName);
+        return this.urlForFindHasMany(id, modelName, snapshot);
       case 'findBelongsTo':
-        return this.urlForFindBelongsTo(id, modelName);
+        return this.urlForFindBelongsTo(id, modelName, snapshot);
       case 'createRecord':
         return this.urlForCreateRecord(modelName, snapshot);
       case 'updateRecord':
@@ -163,9 +163,10 @@ export default Ember.Mixin.create({
    * @method urlForFindHasMany
    * @param {String} id
    * @param {String} modelName
+   * @param {DS.Snapshot} snapshot
    * @return {String} url
    */
-  urlForFindHasMany(id, modelName) {
+  urlForFindHasMany(id, modelName, snapshot) {
     return this._buildURL(modelName, id);
   },
 
@@ -173,9 +174,10 @@ export default Ember.Mixin.create({
    * @method urlForFindBelongsTo
    * @param {String} id
    * @param {String} modelName
+   * @param {DS.Snapshot} snapshot
    * @return {String} url
    */
-  urlForFindBelongsTo(id, modelName) {
+  urlForFindBelongsTo(id, modelName, snapshot) {
     return this._buildURL(modelName, id);
   },
 
--- a/addon/adapters/rest.js
+++ b/addon/adapters/rest.js
@@ -1279,7 +1279,7 @@ if (isEnabled('ds-improved-ajax')) {
 
         case 'findHasMany':
         case 'findBelongsTo':
-          let url = this.buildURL(type.modelName, id, null, requestType);
+          let url = this.buildURL(type.modelName, id, snapshot, requestType);
           return this.urlPrefix(params.url, url);
       }
 
--- a/addon/adapters/rest.js
+++ b/addon/adapters/rest.js
@@ -634,7 +634,7 @@ var RESTAdapter = Adapter.extend(BuildURLMixin, {
       var id   = snapshot.id;
       var type = snapshot.modelName;
 
-      url = this.urlPrefix(url, this.buildURL(type, id, null, 'findHasMany'));
+      url = this.urlPrefix(url, this.buildURL(type, id, snapshot, 'findHasMany'));
 
       return this.ajax(url, 'GET');
     }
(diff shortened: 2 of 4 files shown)


trevorrjohn:
Not sure what is going on with appveyor here. I took a look at the logs, but wasn't able to discern the problem.

addon/adapters/rest.js
```diff
@@ -1279,7 +1279,7 @@ if (isEnabled('ds-improved-ajax')) {
 
         case 'findHasMany':
         case 'findBelongsTo':
-          let url = this.buildURL(type.modelName, id, null, requestType);
+          let url = this.buildURL(type.modelName, id, snapshots, requestType);
```
pangratz: We want to pass the `snapshot` here.

addon/-private/adapters/build-url-mixin.js
```diff
@@ -165,7 +165,7 @@ export default Ember.Mixin.create({
    * @param {String} modelName
```
pangratz: Can you add an entry here for the snapshot?

addon/-private/adapters/build-url-mixin.js
```diff
@@ -175,7 +175,7 @@ export default Ember.Mixin.create({
...
```
pangratz: Also here...

pangratz:
You also need to pass the `snapshot` as third argument [here](https://github.com/emberjs/data/blob/6cb550929caf1ab48eeff0947b1032b8fbe2a298/addon/adapters/rest.js#L690) and [here](https://github.com/emberjs/data/blob/6cb550929caf1ab48eeff0947b1032b8fbe2a298/addon/adapters/rest.js#L637). Since this hasn't been caught a test, can you modify tests for the [`findBelongsTo`](https://github.com/emberjs/data/blob/6cb550929caf1ab48eeff0947b1032b8fbe2a298/tests/integration/adapter/rest-adapter-test.js#L1848-L1881) and [`findHasMany`](https://github.com/emberjs/data/blob/6cb550929caf1ab48eeff0947b1032b8fbe2a298/tests/integration/adapter/rest-adapter-test.js#L1702-L1742). I guess an assertion that the passed `snapshot` is present or it is an instance of `DS.Snapshot` would be good.

Apart from my comments, this looks great!

addon/adapters/rest.js
trevorrjohn: thanks, I am still confused about when snapshot vs snapshots is passed in here. 

Is `snapshot` always a `DS.Snapshot` and this `snapshots` always `DS.SnapshotRecordArray`?

pangratz: Yes. Basically the `snapshot` here represents the "owner" of the relationship.

trevorrjohn:
So I updated the docs and added the snapshot where you pointed to above. I think I am still a little lost on what is going on between the call to `findRecord` and the call to `findHasMany`. I did some digging, but wasn't really able to follow the code path of where `findRecord` starts loading the associations.

addon/adapters/rest.js
```diff
@@ -1154,6 +1154,10 @@ var RESTAdapter = Adapter.extend(BuildURLMixin, {
   buildQuery(snapshot) {
     let query = {};
 
+    if (snapshot) {
+      query.adapterOptions = snapshot.adapterOptions
+    }
```
pangratz: This isn't needed, the main thing this PR tries to solve is just passing the `snapshot` to the `urlForFindBelongsTo` and `urlForFindHasMany` hook. Sorry if my comments confused you here...

tests/integration/adapter/rest-adapter-test.js
```diff
@@ -1700,8 +1700,11 @@ test("findHasMany - returning an array populates the array", function(assert) {
 });
 
 test("findHasMany - passes buildURL the requestType", function(assert) {
+  assert.expect(2);
   adapter.shouldBackgroundReloadRecord = () => false;
+  let optionsStub = { options: true };
   adapter.buildURL = function(type, id, snapshot, requestType) {
+    assert.equal(snapshot.adapterOptions, optionsStub);
```
pangratz: An assertion that the `snapshot` is instance of `DS.Snapshot` is sufficient here. There is actually no way to pass `adapterOptions` here, so that's not needed.

tests/integration/adapter/rest-adapter-test.js
```diff
@@ -1726,7 +1729,7 @@ test("findHasMany - passes buildURL the requestType", function(assert) {
     });
   });
 
-  run(store, 'findRecord', 'post', '1').then(assert.wait(function(post) {
+  run(store, 'findRecord', 'post', '1', { adapterOptions: optionsStub }).then(assert.wait(function(post) {
```
pangratz: This can also be removed.

pangratz:
> I think I am still a little lost on what is going on between the call to findRecord and the call to findHasMany. I did some digging, but wasn't really able to follow the code path of where findRecord starts loading the associations.

The `urlForFindBelongsTo` and `urlForFindHasMany` hooks are not called as a result of a `store.findXXX` or something, but as the result of a linked relationship: it is possible that a relationship in `ember-data` is defined not via an `id` but via a `link` (for example defined [here](https://github.com/emberjs/data/blob/6cb550929caf1ab48eeff0947b1032b8fbe2a298/tests/integration/adapter/rest-adapter-test.js#L1867) in the test). The `urlForFindBelongsTo` hook is then called as a result of accessing the linked relationship [here](https://github.com/emberjs/data/blob/6cb550929caf1ab48eeff0947b1032b8fbe2a298/tests/integration/adapter/rest-adapter-test.js#L1877)...

trevorrjohn:
ok updated, thanks for walking me through that. I was lost there for a bit.

pangratz:
Yeah, this can be quite confusing. But it looks good now!

Thanks @trevorrjohn!

Issue #2289, opened 2014-09-18
applyTransforms hasOwnProp is mega slow


stefanpenner:
https://github.com/emberjs/data/blob/e1ca37a9f52b1f6f94840a242b76ca4a577a10f0/packages/ember-data/lib/serializers/json_serializer.js#L112

we should try and merely have sensible behaviour for the if the key is undefined and avoid this costly step until browsers make it fast.

bmac:
@stefanpenner would checking to see `if (data[key] === undefined) {` result in a performance improvement over `!data.hasOwnProperty(key)`?

stefanpenner:
@bmac yes

balinterdi:
I took a stab at fixing this with #4281.

The suggested fix to use `data[key] === undefined` instead of `!data.hasOwnProperty(key)` did not work because there are a [couple of](https://github.com/emberjs/data/blob/master/tests%2Funit%2Fmodel-test.js#L914) [tests](https://github.com/emberjs/data/blob/master/tests%2Funit%2Fmodel-test.js#L915) that assert that if the key is present but has a value of `undefined` or `null`, it gets serialized to `null` in both cases.

So I ended up using `if (!(key in data))` which seems to work. The caveat is that I'm not sure whether `key in data` is any faster than `data.hasOwnProperty(key)`. If somebody knows this for sure, it'd be great if they could confirm/refute this, otherwise I'll try to profile it with some app code.

balinterdi:
I discussed this with @stefanpenner and @bmac at EmberConf and Stef suggested that, since the current fix might actually not improve performance, I look at the tests that fail when I applied the simple property lookup solution (suggested above by Brendan).

More precisely, we don't need to test if `undefined` is ever deserialized (transformed) to anything since `undefined` is not a valid value in a server response (in JSON). So summing up, the correct solution is probably to remove those tests and just write `if (data[key] === undefined) { return }`

balinterdi:
@stefanpenner @bmac How should I go about this? Should this issue be reopened and then another PR submitted with the solution suggested in my previous comment?

bmac:
@balinterdi feel free to open another PR with the solution from your previous comment.

balinterdi:
[...]

PR #4311, opened 2016-04-07
Use property lookup in applyTransforms

balinterdi:
A property lookup is used to speed up the hot path of applying transforms when normalizing data received from the server.

See discussion on #2289


bmac:
Thanks @balinterdi.

balinterdi:
The pleasure is mine!

PR #4313, opened 2016-04-08
During normalization, use property lookup instead of hasOwnProp checks

sly7-7:
To follow #4311 and according to https://github.com/emberjs/data/issues/2289#issuecomment-206990820, the server returning an undefined value is just invalid, I think we can now replace all those slowly hasOwnProp checks during the normalization process.
Seems like this would improve perfs :)

This will need #2289 to be merged before (the same tests where failing).
I will rebase after that

kindly ping to @stefanpenner @igorT @wecc @bmac

### DIFF
--- a/addon/serializers/json-api.js
+++ b/addon/serializers/json-api.js
@@ -214,7 +214,7 @@ const JSONAPISerializer = JSONSerializer.extend({
     if (resourceHash.attributes) {
       modelClass.eachAttribute((key) => {
         let attributeKey = this.keyForAttribute(key, 'deserialize');
-        if (resourceHash.attributes.hasOwnProperty(attributeKey)) {
+        if (resourceHash.attributes[attributeKey] !== undefined) {
           attributes[key] = resourceHash.attributes[attributeKey];
         }
       });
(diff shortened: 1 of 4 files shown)


stefanpenner:
This most likely will, at the time I made the recommendation I had an active benchmark (in ember-data not just a micro one) but I have since mis-placed it.

I can try finding it, or hopefully i'll have some time soon to revisit it, so we can double check this improvement.

bmac:
I tried to get this (old) benchmark working to see if it could be used to test this pr https://github.com/emberjs/data/tree/igor-benchmarks. After spending about an hour looking at it I was still unable to get it to run successfully.

stefanpenner:
@chadhietala @asakusuma testing this in one your ED heavy apps might reveal something, specifically the call times of the methods in question.

asakusuma:
@stefanpenner you wanting to confirm a perf boost, or wanting to make sure this doesn't break something?

@msranade is doing some perf testing right now.

stefanpenner:
> @stefanpenner you wanting to confirm a perf boost, or wanting to make sure this doesn't break something?

both, and especially the selfTime of the call-sites that have been updated. such as `extractAttributes` etc.

bmac:
@minasmart or @asakusuma did you get a chance to pref test this pr?

stefanpenner:
@bmac I'm confident in saying that this will be positive for performance. Latest v8 starts to optimize hOP, but this form is still faster. Most importantly all those older android devices will also benefit.

bmac:
Thanks @stefanpenner. @sly7-7 do you have time to rebase this pr?

bmac:
This pr has been committed as https://github.com/emberjs/data/pull/4393/commits/03b34e20521ca285a679b340746d3450e481bcb5

sly7-7:
@bmac I'm slowly recovering from a big burn out. I'm still not able to develop anything :disappointed:

PR #4315, opened 2016-04-08
Prevent exception in generatedDetailedMessage when requestData is missing

TheAxeMan301:
I kept getting exceptions here on unauthorized requests. For some reason the requestData is missing. With a little extra care this function can be robust against the omission. Added a test case.

### DIFF
--- a/addon/adapters/rest.js
+++ b/addon/adapters/rest.js
@@ -1143,10 +1143,13 @@ var RESTAdapter = Adapter.extend(BuildURLMixin, {
       shortenedPayload = payload;
     }
 
-    var requestDescription = requestData.method + ' ' + requestData.url;
+    var requestDescription = ' '
+    if (requestData && typeof requestData === 'object') {
+        requestDescription = ' ' + requestData.method + ' ' + requestData.url + ' ';
+    }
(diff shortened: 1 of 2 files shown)


pangratz:
Hey @TheAxeMan301, I am interested how it is possible that `requestData` is not available, since it contains the method and URL, which should always be available when a request is made. Can you provide a code sample which demonstrates the issue you're having?

Maybe you can [use this ember-twiddle](https://ember-twiddle.com/0e1a24aabb8fa7c1fdd8/copy?fileTreeShown=false&numColumns=2&openFiles=routes.application.js%2Ctemplates.application.hbs) as a starting point.

TheAxeMan301:
It is called asynchronously so it wasn't obvious. An addon is involved, either ember-simple-auth or mirage.

TheAxeMan301:
Actually no, it was my own override of handleResponse. Serves me right for not using ...arguments with _super(). It's possible I copied an old demo or prototype and was missing the requestData argument.

pangratz:
Thanks for the response @TheAxeMan301. I'm glad you figured this out! :+1:

PR #4316, opened 2016-04-09
Fix `isUpdating` for DS.AdapterPopulatedRecordArray#update()

pangratz:
The `isUpdating` flag is not set to `true` when the `update()` method on
a `DS.AdapterPopulatedRecordArray` is invoked. As with the flag on
`DS.RecordArray`, this should be `true` until the update is finished and
the array contains the most reset result for the query, fetched from the
adapter.

---

The `[FEATURE]` commit DRY's up some code in the `DS.AdapterPopulatedRecordArray` and by this makes it easier to cherry-pick this bugfix into `beta` (where this feature is not yet available).

### DIFF
--- a/tests/integration/adapter/queries-test.js
+++ b/tests/integration/adapter/queries-test.js
@@ -57,6 +57,37 @@ test("When a query is made, the adapter should receive a record array it can pop
   }));
 });
 
+test("a query can be updated via `update()`", function(assert) {
+  adapter.query = function() {
+    return Ember.RSVP.resolve([{ id: 'first' }]);
+  };
+
+  run(function() {
+    store.query('person', {}).then(function(query) {
+      assert.equal(query.get('length'), 1);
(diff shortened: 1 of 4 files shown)


tests/integration/adapter/queries-test.js
```diff
@@ -57,6 +57,42 @@ test("When a query is made, the adapter should receive a record array it can pop
   }));
 });
 
+test("a query can be updated via `update()`", function(assert) {
+  let done = assert.async();
```
bmac: Is this done needed? I thought the run loop would remove the need for this?

bmac:
Overall looks good.

tests/integration/adapter/queries-test.js
pangratz: You are absolutely right :+1: Removed the `done` and the amount of expected assertions, since this is handled by the run-loop...

PR #4317, opened 2016-04-09
[CLEANUP] remove reference to no more used isNewSerializerAPI flag

pangratz:
This flag has been used when transitioning from 1.13 to 2.0 and is no
more needed in 2.x

:checkered_flag: :wave:


bmac:
Thanks @pangratz

Issue #3801, opened 2015-09-27
[Question] JSONApi disallowes SLASH in namespaced model names


artych:
Now JsonApiSerializer preserves slashes: `'docs/ticket'` model is serialized as `{ type: "docs/ticket" }`. Related PRs: #3026, #3392, emberjs/ember.js#10980.

Unfortunately, this is not allowed by JSON API, though there are no recommendations which symbol to use, see json-api/json-api#850. 

For example `rails_api/active_model_serialzer (0.10.rc3)` uses `underscore`, same as for words splitting, so `'Docs::Ticket'` and `'DocsTicket'` both will be serialized as `{type: 'docs_ticket'}`. IMHO not comfortable, but bearable.

Question: should SLASH be changed by underscore (or another symbol) in JsonApiSerializer?

bmac:
Adding a link to this json-api issue where they are discussing this same problem. https://github.com/json-api/json-api/issues/850

mattgrannary:
Within that thread @ethanresnick mentioned the possibility of defining the namespace in the meta object. I started thinking about this and it really seems like a good place for it. In this situation `_extractType` in the `JSONAPISerializer` could look at the meta object and if a namespace is found then just prepend it separated by a SOLIDUS "/".

mattgrannary:
Actually, taking back this recommendation because it feels like a bit much adding this additional layer of logic just to support namespacing when there should be no reason to include a solidus in the type definition.

ekampp:
If this is made a change, I believe it should be opt-in, since some projects are in production, and have solved the dash vs. underscore problem otherwise.

ivanjolic95:
I'm working on an app where I use ember + rails and [JSONAPI Resources](https://github.com/cerebris/jsonapi-resources) gem to handle communication between backend and frontend.

I've put some models in rails in namespaces, like `Itinerary::Item` and so on. For each model I've specified a resource like `Api::V1::Itinerary::ItemResource`. I did the same in ember, I've made models like `itinerary/item` and everything should be working fine. But it doesn't.

What I get as a response when I try to save/get data is that the resource cannot be found. The type param in JSONAPI object should be `api/v1/itinerary/item` to make it work but it's `itinerary/item`.

I suppose one solution is to add namespace api/v1 to my ember models but I could get ugly with using the models in code and pluralization in the JSONAPI adapter (at least I think). 

[...]

PR #4318, opened 2016-04-10
[FEATURE ds-payload-hooks] Add hooks to map type in payload to modelName

pangratz:
The `modelNameFromPayloadKey` and `payloadKeyFromModelName` hooks on the
serializer allow to customize the mapping between the key of a JSON and
the corresponding name of the model. Consider the following payload:

```json
{
  "blog/post": {
    "id": 1
  }
}
```

To map the "blog/post" key to the `post` model, the
`modelNameFromPayloadKey` hook is used:

```js
serializer.modelNameFromPayloadKey("blog/post")
```

---

There are some issues with the current code base, as the
`modelNameFromPayloadKey` and `payloadKeyFromModelName` hooks are also
used to map the "value" of a type from / to the payload. Consider the
following payload of a JSON-API document:

```json
{
  "data": {
    "id": 1,
    "type": "API:V1::User"
  }
}
```

To map the namespaced type to the model name ember-data is using,
currently the `modelNameFromPayloadKey` is used:

```js
serializer.modelNameFromPayloadKey("API::V1::User")
```

Now this gets complicated if your API responds with the following
payload (using custom key and custom types for polymorphic records for
example):

```json
{
  "blog/post": {
    "id": 1,
    "user": 2,
    "userType": "API:V1::Administrator"
  }
}
```

Now the `modelNameFromPayloadKey` is invoked for both:

```js
serializer.modelNameFromPayloadKey("blog/post")
serializer.modelNameFromPayloadKey("API::V1::Administrator")
```

This means that the logic within the hook would get complicated.

---

As the name suggests, the method should only be used to map the key, not
the value. This commit adds `modelNameFromPayloadType` and
`payloadTypeFromModelName` hooks and uses them during normalization and
serialization.

This commit maintains the old behavior using, if it is used and logs a
deprecation warning to move to the new hooks.

---

This PR should help tackling the issues raised in #3801 and #4208.

### DIFF
--- a/addon/serializers/json-api.js
+++ b/addon/serializers/json-api.js
@@ -3,7 +3,7 @@
 */
 
 import Ember from 'ember';
-import { assert, runInDebug, warn } from 'ember-data/-private/debug';
+import { assert, deprecate, runInDebug, warn } from 'ember-data/-private/debug';
 import JSONSerializer from 'ember-data/serializers/json';
 import normalizeModelName from 'ember-data/-private/system/normalize-model-name';
 import { pluralize, singularize } from 'ember-inflector';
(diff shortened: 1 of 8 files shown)


bmac:
Overall looks good. Do you mind wrapping this in a feature flag?

addon/serializers/json-api.js
```diff
@@ -141,8 +141,20 @@ const JSONAPISerializer = JSONSerializer.extend({
     @private
   */
   _normalizeRelationshipDataHelper(relationshipDataHash) {
-    let type = this.modelNameFromPayloadKey(relationshipDataHash.type);
-    relationshipDataHash.type = type;
+    let modelName = this.modelNameFromPayloadType(relationshipDataHash.type);
+    let deprecatedModelNameLookup = this.modelNameFromPayloadKey(relationshipDataHash.type);
+
+    if (modelName !== deprecatedModelNameLookup && this.modelNameFromPayloadType === JSONAPISerializer.prototype.modelNameFromPayloadType) {
```
bmac: It seems like if a user intentionally wanted `modelNameFromPayloadType` and `modelNameFromPayloadKey` to return different values and they needed to define a custom `modelNameFromPayloadType` we would always hit the deprecated case.

What about if we always called `modelNameFromPayloadType` but had its default implementation just call `modelNameFromPayloadKey`?

fivetanley:
We should also make an ember-watson transform after this gets merged.

pangratz:
@bmac I've updated the code and put everything behind the `ds-payload-type-hooks` feature flag.

You're comment made me re-think the approach and now the following "rules" apply:

- `modelNameFromPayloadKey` and `payloadKeyFromModelName` are deprecated on `json` and `json-api` serializer, since the their payload doesn't have the type in any key; there is a deprecation warning to use the new `modelNameFromPayloadType` and `payloadTypeFromModelName` hooks instead
- `rest` serializer supports all 4 hooks (`payloadKey` and `payloadType`) since they make sense for that payload
  - if the value for the new hook differs with the old one, then a deprecation is logged iff the new hook is not implemented but the old one is
  - if they are the same then everything is ok

PR #4320, opened 2016-04-12
Fixup feature flagging infrastructure.

rwjblue:
Prior to this:

* When used as an ember-cli addon, the feature flags would only be processed for `production` builds.
* When built for globals builds, the features flags would only be processed for `ember-data.prod.js` builds.

This is incorrect, they should be processed for all builds (for every environment) so that any features explicitly enabled or disabled in the version of Ember data being used are stripped properly.

### DIFF
--- a/lib/stripped-build-plugins.js
+++ b/lib/stripped-build-plugins.js
@@ -3,7 +3,7 @@ var path          = require('path');
 var filterImports = require('babel-plugin-filter-imports');
 var featureFlags  = require('babel-plugin-feature-flags');
 
-module.exports = function() {
+module.exports = function(environment) {
   var featuresJsonPath = __dirname + '/../config/features.json';
   var featuresJson = fs.readFileSync(featuresJsonPath, { encoding: 'utf8' });
   var features = JSON.parse(featuresJson);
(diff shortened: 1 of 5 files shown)


lib/stripped-build-plugins.js
```diff
@@ -16,24 +16,29 @@ module.exports = function() {
   //     features[feature] = false;
   //   }
   // }
```
pangratz: I guess we can finally remove this comment too, as it is not relevant anymore...

rwjblue:
Sample assets:

* sample ember-cli app
  * [development vendor.js](http://assets.rwjblue.com/ember-data/feature-flag-test-assets/ember-cli-app-dist/development/assets/vendor.js)
  * [non-minified production vendor.js](http://assets.rwjblue.com/ember-data/feature-flag-test-assets/ember-cli-app-dist/production/assets/vendor-c8e8a48e41fe74ab0c1d5b17574d19a5.js)
* bower builds
  * [ember-data.js](http://assets.rwjblue.com/ember-data/feature-flag-test-assets/bower-build/ember-data.js)
  * [ember-data.prod.js](http://assets.rwjblue.com/ember-data/feature-flag-test-assets/bower-build/ember-data.prod.js)

bmac:
Thanks @rwjblue.

PR #4331, opened 2016-04-14
[BUGFIX beta] Overhaul attr/relationships imports in model blueprint

pangratz:
When a model is generated via `ember generate model` without attributes
and relationships, then it is not immediately clear from the generated
blueprint, on how to add attributes and relationships via importing
stuff from the modules.

This change adds the imports as comments, so uncommenting allows to
quickly add the imports without writing the import statement. So the
command `ember genrate model foo` now creates this file:

```js
import Model from "ember-data/model";
// import attr from "ember-data/attr";
// import { belongsTo, hasMany } from "ember-data/relationships";

export default Model.extend({
});
```

---

This has been raised in the `#-ember-data` slack channel, where it was confusing on how to add attributes and relationships, since there is no `import DS from "ember-data"` in the generated file anymore, and it is not immediately clear where `attr`, `belongsTo` and `hasMany` are imported from.

### DIFF
--- a/blueprints/model/HELP.md
+++ b/blueprints/model/HELP.md
@@ -14,12 +14,16 @@ For instance: <green>\`ember generate model taco filling:belongs-to:protein topp
 would result in the following model:
 
 ```js
-import DS from 'ember-data';
-export default DS.Model.extend({
-  filling: DS.belongsTo('protein'),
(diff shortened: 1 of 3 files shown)


bmac:
Looks good. 👍

PR #4333, opened 2016-04-14
[BUGFIX release] use es5 syntax for addon's index.js file

fivetanley:
Older versions of node can't use the concise method syntax.


rwjblue:
Oh darn it. Sorry :(

PR #4343, opened 2016-04-20
[BUGFIX beta] Avoid using `Ember.lookup` to setup global.

rwjblue:
`Ember.lookup` is an internal mechanism used by older versions of Ember
to "lookup" things on the global scope. In Ember 2.7 usage of
`Ember.lookup` will be deprecated, so this removes our need for it.

This also adds a deprecation when using the global `DS` (suggesting
imports instead).

---

Currently, `import DS from 'ember-data';` when ran as an ember-cli addon, still adds a `window.DS` global. This PR maintains that status quo (but prevents issues on Ember 2.7), if we want to only set `window.DS` up for globals builds (which seems better to me) I can modify this PR.

### DIFF
--- a/addon/index.js
+++ b/addon/index.js
@@ -1,6 +1,8 @@
 import Ember from "ember";
-import { warn } from "ember-data/-private/debug";
+import { warn, deprecate } from "ember-data/-private/debug";
 import isEnabled from "ember-data/-private/features";
+import global from "ember-data/-private/global";
+
 /**
   Ember Data
   @module ember-data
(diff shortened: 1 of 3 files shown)


rwjblue:
Updated to add a deprecation message when accessing `window.DS`.

Turbo87:
this is actually breaking some addons currently that are tested via ember-try against ember#canary

rwjblue:
Yep, the real regression is in Ember (and and issue is reported over there).

addon/index.js
```diff
@@ -164,6 +166,17 @@ Object.defineProperty(DS, 'normalizeModelName', {
   value: normalizeModelName
 });
 
-Ember.lookup.DS = DS;
+Object.defineProperty(global, 'DS', {
```
bmac: Should we allow `DS` in the bower build?

rwjblue: Yes, I think we have to.  I'll update soon.

bmac: ping @rwjblue. Do you have time to work on this? If not I can try to find someone else to get this pr over the finish line.

bmac:
This pr has been merged as https://github.com/emberjs/data/commit/fe076470b77e2d6abd614bc8710c37fb31b64e3b

PR #4345, opened 2016-04-20
[DOC] fix example syntax

ghedamat:
(will fix highlighting in docs)


pangratz:
Thanks @ghedamat! Sharp 👀 !

Issue #4344, opened 2016-04-20
ds-finder-include doesn't work for create/update operations


cibernox:
The `ds-finder-include` feature (enabled by default in ED 2.5) works for finder calls `store.find/findAll` but it doesn't for save operations.

This is useful when creating/updating a record must request also related resources.

Suggested syntax:

``` js
blogPost.save({ include: 'comments' }) // update
```

During slack conversation with @bmac this looks like a bug in the feature that could be implemented without adding any extra feature flag.

mitchlloyd:
I was just playing around with this feature today and tried something like this:

```javascript 
  model: function() {
    let project = this.modelFor('projects.project');
    return project.get('evidenceRequests', { include: 'assignee' });
  },
```

I was hoping that would add the same `include` query param to the URL but it did not. Should this feature be considered for these type of requests?

cibernox:
@mitchlloyd That seems tricky to me, because unlike find/save operations that are fired ad-hoc, relationships are more a declaration of the relations between resources and fetching them from different points of the apps in different ways looks odd to me.

Although what you suggest I think that can be done with the `ds-references` feature enabled by default in 2.5 (http://emberjs.com/blog/2016/03/13/ember-data-2-4-released.html#toc_code-ds-references-code). I haven't played with it myself.

In any case, it worth discussing, so I summon @mitchlloyd to see if it makes sense to add this to has Many relationships or it makes more sense to use `ds-references` magic this way: `post.hasMany('comments').load({ include: 'comments' });`

mitchlloyd:
@cibernox Adding parameters to `get` does seem dubious and I'm beginning to think I want to avoid using using `get` to fetch async relationships in general.

The API you proposed looks great to me. I updated ED to the latest master commit to see if it would work today, but it seems the query param doesn't make it into the URL yet. Seems useful!

pangratz:
@mitchlloyd querying a relationship with additional parameters is not (yet) supported in ember data. But there is an addon which might interest you: https://github.com/mdehoog/ember-data-has-many-query

sukima:
For anyone who comes here and realizes that this is not supported. No need to despair; here is how to do this currently:

```js
// app/adapters/<resource>.js
import ApplicationAdapter from './application';

export default ApplicationAdapter.extend({
  urlForCreateRecord() {
    let url = this._super(...arguments);
    let include = '<comma separated resource types>';
    return `${url}?include=${escapeURIComponent(include)}`;
  },
[...]

PR #4352, opened 2016-04-24
Remove ds-finder-include conditionals since it is enabled by default now

cibernox:
I'm investigating this https://github.com/emberjs/data/issues/4344 and I did some cleanup in the process


pangratz:
Thanks @cibernox , but this has already been addressed in #4348.

PR #4353, opened 2016-04-24
Send `include` query with save/update

cibernox:
WIP.

Not sure if this is the right fix.


stefanpenner:
this seems to have gone stale, if there is still interest please let us know (and rebase), we can then gladly reopen.

PR #4354, opened 2016-04-25
Add `include` param support for create/update/delete operations

HeroicEric:
Fixes https://github.com/emberjs/data/issues/4344

Sorry, @cibernox. I worked on this on a plane and did not realize you had already started working on it. I will post this here so maybe we can together come up with the best solution 😄 

For reference, https://github.com/emberjs/data/pull/4353 is the PR from @cibernox 

I do have the same question as you mentioned in the #dev-ember-data about whether or not this param belongs in the URL or request body. I decided to keep the implementation similar to the `findRecord` and `findAll` and delegate the decision to `$.ajax`.

You went one way and I went the other 😺 .

This allows an `include` parameter to be specified when creating,
updating, or deleting a model.

This feature was originally discussed in
https://github.com/emberjs/rfcs/pull/99 but was not part of the initial
implementation.

Examples:

``` javascript
// Create
let post = this.store.createRecord('post', { title: 'Hurray' });

post.save({ include: 'author' });

// POST /posts?include=author
```

``` javascript
// Update
let post = this.store.peekRecord('post', 123);

set(post, 'title', 'Hurray');

post.save({ include: 'comments' });

// PUT /posts/123?include=comments
```

``` javascript
// Delete
let post = this.store.peekRecord('post', 123);

post.deleteRecord();

post.save({ include: 'comments' });

// DEL /posts/123?include=comments
```


cibernox:
@HeroicEric Let's see if @dgeb can clarify what the spec prefers. I went for the includes in the URL por homogeneity with GET requests, and because using other formats other than json-api (I'm using AMS but with some borrower json-api conventions like the `include` and things like filtering/sorting) is slightly easier to implement for me.

bmac:
My understanding is the `include` parameter should always be a query parameter and it should *not* be included in the payload body for a POST, PUT or DELETE. Can you confirm @dgeb or @tkellen?

dgeb:
> My understanding is the include parameter should always be a query parameter and it should not be included in the payload body

That's correct. Adding `include` to the request document does not conform to the spec.

cibernox:
@HeroicEric If you really want to implement this, if you have a clear idea about how to implement this (check my PR see if I was walking the good way) feel free to do it or reuse any code from my PR. You will surely implement this way faster than me.

If you are short of time I could continue.

@dgeb Maybe it worth to edit the spec to clarify this even more. This is a prove that least two people found it open to interpretation.

HeroicEric:
Updated this to always add the query params to the URL rather than in the request body

tests/integration/adapter/rest-adapter-test.js
```diff
@@ -590,6 +590,19 @@ test("createRecord - relationships are not duplicated", function(assert) {
   }));
 });
 
+if (isEnabled('ds-finder-include')) {
```
cibernox: This flag is already enabled in 2.5. I don't think you need it

HeroicEric: Nice catch!

cibernox:
I'm craving for this fix. Is it waiting for some sort of team discussion?

fivetanley:
Hey so we discussed this just now in the team meeting.  We think that this is a good use case to support, but we are stuck over whether to name it `include` or not.

@igorT brought up a good point which is that `include` may be slightly confusing in the `find` case, as one could think it means "send these relationships in the request to the server" when it actually means "give me these relationships back when the save is successful."

We are summoning the Tomhuda (@tomdale @wycats) to help us figure out the correct thing going forward.

Right now with finder methods , we have the following signature: 

```javascript
store.findAll(modelName, options);
```

Where `options` generally looks like this:

```javascript
{
  // Ember Data reserved stuff. Usually stuff that's reserved in JSONAPI
  include: relationshipName,
  
  // Put any custom stuff for your adapter in `adapterOptions` so we don't clobber it later
  adapterOptions: {

  }
}
```

Right now, this contract for `options` is not established for side effect methods like create, update, delete.

Ideally, we want to make it the same as find, but the `include` naming is a bit ambiguous right now.

at a minimum we should expose a `queryParams` option, but I'm hoping we can work together through this soon. I'll continue to ping tom and yehuda, feel free to ask me for updates as well.

HeroicEric:
@fivetanley makes sense. I definitely thought that `.save({ include: 'something' })` was for saying to include the relationships in the request when I first read the issue (https://github.com/emberjs/data/issues/4344).

Would you want me to work on an RFC for the `queryParams` option you mentioned?

cibernox:
I see. For read operations seems evident what it means, but not so much for saving things. I never saw it that way. 

Actually in my use-cases where I want to the response to include data in the payload and those cases where I'm also sending that data in the request (with `DS.EmbeddedRecordsMixin`). 

The typical "Send an snapshot of how I want this tree of data to look like and the backend will figure out what to add, remove or update to make it happen" scenario.

However It seems desirable to me to have the same set of options work for both `store.find` and `model.save`, but I can understand the confusion. Maybe is just a matter of documenting it properly?

cibernox:
I've just seen this feature mentioned in a deck of the Munich Ember meetup. Any consensus on this? I might be able to put time on this if it needs a champion.

runspired:
I'm in favor of moving this forward despite the confusing ambiguity that arises for `create` and `update` requests.  I think it is less confusing for us to reflect API parity across our various request methods (we already use `include` elsewhere and it is `jsonapi` spec) than to try to shoehorn a different name into place. This seems especially true when there is a spec to guide us.

We also already have offered a delineation between `request` params and `adapterOptions`. I suspect we should message that options for serialization (such as `embedded`) should be placed in `adapterOptions` for now.

In the near term, I would like to RFC deprecating giving snapshots to request methods eagerly, in favor of an object more like the following which would clear up this confusion:

```ts
{
   params: <Object>;
   options: <Object>;
   snapshots: <Array>;
}
```

That said, if end users decide they want to use the include flag on both sides of the equation (what to serialize and what to request back), that's something they can do, but which we won't offer out of the box.

hjdivad:
The team discussed this today and the consensus was we're happy to see this move forward and merged after rebasing.

runspired:
At this point this branch is pretty stale and the value here is now a bit questionable. Internally, we try to be opaque, and as such `includes` is largely not a concept anymore except in the legacy codepaths. In modern stuff query params are assembled at the point the request is issued and so whatever the app would like it to mean it will mean.

Issue #3772, opened 2015-09-15
Add { backgroundReload: false } to store.findRecord() and store.findAll()


HeroicEric:
This feature is mentioned in http://emberjs.com/blog/2015/06/18/ember-data-1-13-released.html#toc_better-caching-defaults-for-code-findall-code-and-code-findrecord-code but is not yet implemented.

I was going to work on adding it but I could not find a full description of the intended behavior and it was not clear to me when I began working on an implementation.

What is the desired behavior when the record is **not already present in the store**? Should it behave like `{ reload: true }` and wait for data to be fetched from the server or should nothing be returned?

pangratz:
Hey @HeroicEric, sorry for the late response :pensive: 

> What is the desired behavior when the record is not already present in the store? Should it behave like { reload: true } and wait for data to be fetched from the server or should nothing be returned?

I think the case your'e describing is currently handled [here](https://github.com/emberjs/data/blob/v2.5.2/addon/-private/system/store.js#L469-L471): if there is no record, then it is loaded from the adapter and no check for the passed `reload` and `backgroundReload` options are made. This is the current state in the code as of `v2.5.2`. You opened this issue last year, so this functionality might have been added in the meantime.

I will bring this up in the next team meeting to check the path which should be taken here. I will get back to you, so you can start implementing this (if you are still interested, which would be highly appreciated :wink: )

Does that sound good to you?

HeroicEric:
@pangratz yes :smile:

pangratz:
Hey @HeroicEric, we discussed this and it would be lovely if the `backgroundReload` option is considered as well in `findAll` and `findRecord` 😉

PR #4356, opened 2016-04-26
Add `backgroundReload` option to `store.findRecord` & `store.findAll`

HeroicEric:
Fixes https://github.com/emberjs/data/issues/3772

This adds support for controlling when records are reloaded in the
background to `store.findRecord()` and `store.findAll()`.

### store.findRecord()

If you would like to force or prevent background reloading, you can set a
boolean value for `backgroundReload` in the options object for
`findRecord`.

```javascript
import Ember from 'ember';

  model: function(params) {
    return this.store.findRecord('post', params.post_id, { backgroundReload: false });
  }
});
```

### store.findAll()

If you would like to force or prevent background reloading, you can set a
boolean value for `backgroundReload` in the options object for
`findAll`.

```javascript
import Ember from 'ember';

  model: function() {
    return this.store.findAll('post', { backgroundReload: false });
  }
});
```

### DIFF
--- a/addon/-private/system/store.js
+++ b/addon/-private/system/store.js
@@ -440,6 +440,8 @@ Store = Service.extend({
     store, it depends on the reload behavior _when_ the returned promise
     resolves.
 
+    ### Reloading
+
     The reload behavior is configured either via the passed `options` hash or
     the result of the adapter's `shouldReloadRecord`.
 
@@ -471,6 +473,8 @@ Store = Service.extend({
 
     If no reload is indicated via the abovementioned ways, then the promise
     immediately resolves with the cached version in the store.
+
+    ### Background Reloading
 
     Optionally, if `adapter.shouldBackgroundReloadRecord` evaluates to `true`,
     then a background reload is started, which updates the records' data, once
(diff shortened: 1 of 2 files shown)


HeroicEric:
@pangratz ask and ye shall receive

pangratz:
😍  awesome @HeroicEric, I will review on monday!

addon/-private/system/store.js
```diff
@@ -547,16 +565,21 @@ Store = Service.extend({
       return this.scheduleFetch(internalModel, options);
     }
 
-    // Refetch the record if the adapter thinks the record is stale
+    if (options.backgroundReload === false) {
+      return Promise.resolve(internalModel);
+    }
```
pangratz: I tend to think that this needs to be moved below the code which checks if a reload should be done (either via the options or via the adapters' `shouldReloadRecord`). In my understanding a reload should take precedence over a background reload.

IMHO the logic should be:

- if we should do a reload (either configured via the passed option or via the adapters' `shouldReloadRecord`), then return a promise which resolves once the record has been reloaded
- if we shouldn't do a reload but we should do a background reload (either configured via the passed option or via , we return a promise which resolves immediately but the record is reloaded in the background.

What do you think?

HeroicEric: Agreed.

pangratz:
Thanks @HeroicEric 🚀

PR #4360, opened 2016-04-28
Replace JSHint/JSCS with ESLint

Turbo87:
This PR is replacing all JSHint/JSCS rules with equivalent ESLint rules. Ember CLI recently switched to ESLint and it is only a matter of time before we start transitioning the user projects to ESLint too.

/cc @rwjblue


bmac:
jshint use to fail the build when there was an issue (through test failures). It looks like the eslint plugin ins't integrating with the test tests the same way so the travis build still passes when there is a eslint error. @Turbo87 do you have some thoughts on how eslint should be integrated with travis and the dev workflow?

Turbo87:
@bmac ESLint does exactly the same thing actually, but you are right that currently the `addon` folder doesn't seem to be linted. I've tried it on master and it's not linted there either, so I guess that is an unrelated issue.

rwjblue:
ember-cli@2.4 was not linting addon/ tree

Turbo87:
@pangratz @bmac I've rebased this PR on top of #4391 and ESLint is now also running against the `addon` folder

Turbo87:
looks like AppVeyor is failing due to https://github.com/emberjs/data/pull/4392

Turbo87:
rebase on the current master, CI issues are fixes now. this should be ready for merging.

/cc @pangratz @bmac

bmac:
Thanks @Turbo87.

PR #4363, opened 2016-04-29
[DOC] Fix hard-coded "Customizing Adapters" link

menewman:
The link to the Customizing Adapters guide was hard-coded to point to the v2.3.0 docs. I updated the link so that it will redirect to the latest version's guide page instead.


bmac:
Thanks @menewman

Issue #4367, opened 2016-05-04
Coalesced requests repeat ID's


orf:
Hey,
I've noticed that in some cases coalesced requests repeat the ID's being fetched, i.e issuing two requests for a model with the ID 1 will result in `/api/models?ids[]=1&ids[]=1` being fetched.

This normally wouldn't be an issue, but if you are fetching something with a long ID that is duplicated many times then the request URI can get very long which can cause some web servers to reject the request, and as far as I can tell duplicating the ID's is not needed. The ID should be unique so it should only send a value once per coalesced request. Is this possible to only send _unique_ coalesced ID's per request?

Edit: I'm using the [ember-django-adapter](https://github.com/dustinfarris/ember-django-adapter/) but I believe this is handled by ember-data?

nadavshatz:
👍 
Took me a really long while to realize what was happening.

pangratz:
> This normally wouldn't be an issue, but if you are fetching something with a long ID that is duplicated many times then the request URI can get very long which can cause some web servers to reject the request

This should be handled by the logic within `adapter.groupRecordsForHasMany()`, which checks if the resulting URL would be too long and if so, creates a separate request: https://git.io/vrYz5. Just for clarification: did you just mention this as a theoretical consequence of having duplicate, long ids or did you run into this issue. I am asking because this should be handled and if not, then there is a bug there too.

orf:
I actually ran into this issue. Everything was working fine until an
important user went to view a specific page, which issued a request 4 bytes
too long for our UWSGI server.
On 13 May 2016 08:41, "Clemens Müller" <notifications@github.com> wrote:

This normally wouldn't be an issue, but if you are fetching something with
a long ID that is duplicated many times then the request URI can get very
long which can cause some web servers to reject the request

This should be handled by the logic within adapter.groupRecordsForHasMany(),
which checks if the resulting URL would be too long and if so, creates a
separate request: https://git.io/vrYz5. Just for clarification: did you
just mention this as a theoretical consequence of having duplicate, long
ids or did you run into this issue. I am asking because this should be
handled and if not, then there is a bug there too.

—
You are receiving this because you authored the thread.
Reply to this email directly or view it on GitHub
<https://github.com/emberjs/data/issues/4367#issuecomment-218973175>

pangratz:
Hmm, ok 😔  . You can set the [`maxURLLength` property](https://git.io/vrYao) on the adapter which should help you there...

fivetanley:
[...]

Issue #4190, opened 2016-02-26
Adapter not coalescing requests when ids contains ":"


engwan:
We're on ember data 2.3. And our ids are in the format: `USER:1234556789`

When I have `users: [1, 2, 3]` in my model, loading the async relationship correctly coalesces them into one request with the `ids` param.

But when I have `users: ['USER:1234567890', 'USER:686234234']`, the models are requested individually.

Is this a bug? Probably caused by a URL encoded character?

We can't really easily change the id format because this is for a legacy system.

pangratz:
Okay, this is tricky: the problem is that `lastSegment` needs to be `decodeURIComponent`ified [here](https://github.com/emberjs/data/blob/v2.3.3/addon/adapters/rest.js#L666), before it is compared with the `id` of the snapshot. The reason is that `lastSegment` represents the the `encodedURIComponent`ified version of the `id` (this happens [here](https://github.com/emberjs/data/blob/v2.3.3/addon/-private/adapters/build-url-mixin.js#L97)).

So the comparison is `"USER%3A1" === "USER:31"` which lets the condition evaluate to `false` and hereby the URL for coalescing is not correctly "calculated".

I couldn't come up easily with tests which test this and the other case where the URL is something like `/model?id=xxx`. I hope that I will have time soon to fix this.

In the meantime I can't think of another way than to monkey patch your adapters' `_stripIDFromURL` until this is fixed upstream in ember-data:

```js
// adapters/application.js
import RESTAdapter from "ember-data/adapters/rest";

export default RESTAdapter.extend({

  _stripIDFromURL(store, snapshot) {
    var url = this.buildURL(snapshot.modelName, snapshot.id, snapshot);

    var expandedURL = url.split('/');
    //Case when the url is of the format ...something/:id
    var lastSegment = expandedURL[expandedURL.length - 1];
    var id = snapshot.id;

    // monkey patch warning
    if (decodeURIComponent(lastSegment) === id) {
      expandedURL[expandedURL.length - 1] = "";
    } else if (endsWith(lastSegment, '?id=' + id)) {
      //Case when the url is of the format ...something?id=:id
      expandedURL[expandedURL.length - 1] = lastSegment.substring(0, lastSegment.length - id.length - 1);
    }

    return expandedURL.join('/');
  }

};
```

Issue #1849, opened 2014-04-03
Bulk operations support


andrewfan:
From `TRANSITION.md` i understood that ember-data will handle bulk operations in future. In my project it is really critical to submit data to server in 1 http\s request. I take a look inside `store` code and looks like we can implement bulk operations right now. 

`store.flushPendingSave` (https://github.com/emberjs/data/blob/master/packages/ember-data/lib/system/store.js#L949) methods can separate records by record.type and then by operation type and call `_commit` (https://github.com/emberjs/data/blob/master/packages/ember-data/lib/system/store.js#L1761) function with new operation name f.e `createRecords`. 

after this the only thing we have to do is to add new methods to serializer that will call existing `extractArray` method and new methods to adapter.

should i start working on it, or my guesswork are incorrect?

btecu:
@igorT any news on this?
I would really like batch PATCH or at least PUT operations.

tim-evans:
I have a bug revolving around records needing to be committed in order. I would be delighted to use bulk commit when creating these records. (The use case is adding many items to an ordered list.)

leejt489:
This is pretty important for apps with network latency.  Would like very much to see this integrated!

SeyZ:
+1

sparkrevolutions:
+1

workmanw:
We actually do this now by implementing a custom adapter which models a [fork-join queue](https://en.wikipedia.org/wiki/Fork%E2%80%93join_queue). 

`<braindump>`

  The TL;DR; is that when `updateRecord` is called on the adapter we create a "work-item" and push it into a work queue. The "work-item" has the serialized record, the record type, the id and an instance of `Ember.RSVP.defer()`. As for the `updateRecord` we return the `defer.promise` (Defer is used because the API is more convenient). 

  Pushing things into a work queue starts a timer (we are using a 100ms duration). We also have a maximum batch size. Once the queue's timer expires or the max batch size is reached, the AJAX is fired. Once the AJAX returns it's results we loop back over the work item, matching the corresponding commit status with the proper workitem. We use that status to do `defer.resolve` or `defer.reject`.

  From an application perspective, we are just calling `record.save()` on a bunch of entities and the adapter will coalesces them properly.

  This approach is very naive in that is only batches entities of a common type and it doesn't allow the application to build it's own batch groups which could be beneficial if you wanted to have parallel transactions. 

`</braindump\>`

kurko:
@workmanw any chance you can opensource that?

[...]

PR #4368, opened 2016-05-04
Make _findMany only fetch unique IDs

orf:
This is a MR to address #4367. I think it is as simple as this, I will amend this and write some tests if the issue isn't rejected.

### DIFF
--- a/addon/-private/system/store/finders.js
+++ b/addon/-private/system/store/finders.js
@@ -55,7 +55,8 @@ export function _find(adapter, store, typeClass, id, internalModel, options) {
 
 export function _findMany(adapter, store, typeClass, ids, internalModels) {
   let snapshots = Ember.A(internalModels).invoke('createSnapshot');
-  let promise = adapter.findMany(store, typeClass, ids, snapshots);
+  let unique_ids = Ember.A(ids).uniq();
+  let promise = adapter.findMany(store, typeClass, unique_ids, snapshots);
   let serializer = serializerForAdapter(store, adapter, typeClass.modelName);
   let label = "DS: Handle Adapter#findMany of " + typeClass;
 
@@ -67,7 +68,7 @@ export function _findMany(adapter, store, typeClass, ids, internalModels) {
   promise = _guard(promise, _bind(_objectIsAlive, store));
 
   return promise.then(function(adapterPayload) {
-    assert("You made a `findMany` request for " + typeClass.modelName + " records with ids " + ids + ", but the adapter's response did not have any data", payloadIsNotBlank(adapterPayload));
+    assert("You made a `findMany` request for " + typeClass.modelName + " records with ids " + unique_ids + ", but the adapter's response did not have any data", payloadIsNotBlank(adapterPayload));
     return store._adapterRun(function() {
       let payload = normalizeResponseHelper(serializer, store, typeClass, adapterPayload, null, 'findMany');
       //TODO Optimize, no need to materialize here
--- a/tests/unit/store/adapter-interop-test.js
+++ b/tests/unit/store/adapter-interop-test.js
@@ -101,6 +101,32 @@ test("Calling Store#findRecord multiple times coalesces the calls into a adapter
   });
 });
 
+test("Only unique ID's are passed to coalesced find requests", function(assert) {
+  assert.expect(1);
+  let done = assert.async();
+
+  var adapter = TestAdapter.extend({
+    findMany(store, type, ids, snapshots) {
+      assert.deepEqual(ids, ["1", "2"], 'Correct unique ids were passed in to findMany');
+      return Ember.RSVP.resolve([{ id: 1 }]);
+    },
+    coalesceFindRequests: true
+  });
+
+  var currentType = DS.Model.extend();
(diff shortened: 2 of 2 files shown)


bmac:
Looks like a good change @orf. Once you add a test I think we can merge this pr.

orf:
Done @bmac, is that test sufficient?

orf:
Done :+1:

pangratz:
It looks like the test doesn't fail when the changes in `finders.js` are reverted. I am curious how it is possible that multiple find requests are triggered in the first place. Can you re-create a use case which demonstrates the issue you have when you reported #4367?

Don't get me wrong, in general I agree that this change is ok, I just want to find out what the underlying issue is here...

orf:
I'm not entirely sure what causes it, but it definitely happens occasionally. I just set a breakpoint in my app to break when this occurs, and it just did with this:

`[04/May/2016 16:16:46] "GET /api/employees/?ids%5B%5D=EMPLOYEE_NAME&ids%5B%5D=EMPLOYEE_NAME HTTP/1.1" 200 194`

fivetanley:
> It looks like the test doesn't fail when the changes in finders.js are reverted

Maybe this is because we are using `.finally` in the tests? I think we just want to return the promise instead. That will fail if the promise rejects, and pass if it succeeds.

nadavshatz:
I've personally encountered it when a model was being loaded by a route and then an ember-concurrency task that is set to run in an inner component was trying to reload said model.
it would create the same issue.

and because the reload would return a json with root "plural" it would not work and the reload would fail.

pangratz:
@fivetanley I think using `finally` is fine. What I mean is that the `assert.deepEqual(ids, ["1", "2"], 'Correct unique ids were passed in to findMany');` doesn't fail when the change is reverted, so it looks like the current test doesn't cover the bug which is solved here.

@orf are you using custom `primaryKey`s in your serializers? If you can reliably trigger this bug in a specific view you have, are you interested in pairing so we can dig into this?

@nadavshatz can you share some code?

nadavshatz:
@pangratz http://emberjs.jsbin.com/pedijepepu/edit?html,js,output
Finally got a version working!

Take a look at the console at the mockjax log and see that the `findRecord` from the route is combined with the `reload` from the component into one call with the same id showing twice.

pangratz:
Thanks @nadavshatz, this helped a ton tracking down the issue. @orf also thanks for your test case, you we're very close: the bug can be reproduced if a record is already pushed into the store and it is fetched multiple times within the same run-loop (either via a reload or a background reload).

@orf can you confirm that the record(s) is/are already pushed to the store when you run into the issue reported in #4367? I currently don't see a way how this bug occurs for multiple `findRecord`'s within the same run-loop, without the record(s) being already in the store beforehand.

---

The root of the bug basically lies within [`store.scheduleFetch`](https://git.io/vrGnq), where it is checked if there is already a `_loadingPromise` for the `InternalModel` [here](https://git.io/vrGnm), and if there is none, then another fetch item is created, without checking if there might already be a fetch scheduled for the given `InternalModel`.

To fix this bug there are several different approaches:

- looks like the state machine currently doesn't always set the `_loadingPromise` for the `InternalModel` when the [`loadingData`](https://git.io/vrGnz) event is triggered (compare https://git.io/vrGni to https://git.io/vrGnw). As pointed out in the comment above that event handler, I am not sure what implications there are when we adapt the state machine here.
- another fix could be to check if `this._pendingFetch` already contains [an item for the `internalModel`](https://git.io/vrGny) and if so, return its' resolvers promise.

One of those approaches would be suitable as a quick fix here IMHO.

---

On the long run and given that there are already some bugs around the find coalescing stuff (#4190, #4367) and there is need for bulk saving #1849, I would suggest to resurrect @mike-north's #2891 and refactor the whole coalescing stuff a whole. But before that happens this should definitely be fixed in the meantime.

@emberjs/ember-data-contributors thoughts?

addon/-private/system/store/finders.js
```diff
@@ -55,7 +55,8 @@ export function _find(adapter, store, typeClass, id, internalModel, options) {
 
 export function _findMany(adapter, store, typeClass, ids, internalModels) {
   let snapshots = Ember.A(internalModels).invoke('createSnapshot');
-  let promise = adapter.findMany(store, typeClass, ids, snapshots);
+  let unique_ids = Ember.A(ids).uniq();
```
sandstrom: Should this be `uniqueIds`? Variable names are typically camelCased in Javascript.

broerse:
It seems unloadRecord should run for the first id. Just as in this test:  https://github.com/jlami/ember-bug-empty-data

broerse:
See also https://github.com/emberjs/data/issues/4463

stefanpenner:
Just reviewing the code which invokes `_findMany` im not sure it works correctly with duplicate `ids` as `seeking` within `_flushPendingFetchForType` will only store the last deferred and resolve that, not an array of all reserves.

I believe more work is required to make this feature complete, or someone should demonstrate ^ is already handled.

runspired:
Closing in favor of #4596 test-pr and eventual fix for fetch scheduling

Issue #4175, opened 2016-02-20
Using ember-ajax with Ember Data


taras:
The purpose of this issue is to flash out and later keep track of the progress of integrating [`ember-ajax`](http://github.com/ember-cli/ember-ajax) and Ember Data. 

@alexlafroscia and I spent some time thinking through how the adoption process might work. Here are some of our thoughts. We're open to suggestions, both strategic and tactical. 

The migration path will be gradual and will eventually result in ajax code being removed from Ember Data's RESTAdapter in Ember Data 3.0. 

To get there we propose the following path:
- [ ] Refactor RESTAdapter to allow users to opt-in to use Ember Ajax Service to make ajax requests
- [ ] Encourage users to use error detection helpers from Ember Ajax Service to ease transition
- [ ] Introduce deprecation warnings to inform users who have dead code in their application's adapters
- [ ] Write Ember Data blog post that introduces Ember Ajax and provides information about migration path
- [ ] Remove ajax code from Ember Data RESTAdapter before 3.0
## Opt-in to use Ember Ajax Service

To opt-in the user will inject Ember Ajax Service into the adapter on `ajaxService` property.

``` javascript
export default DS.JSONAPIAdapter.extend({
  ajaxService: Ember.inject.service('ajax')
});
```
## Error detection helpers

Ember Ajax comes with helpers that can be used to distinguish errors. Ember Ajax provides the following helpers: `isUnauthorizedError`, `isForbiddenError`, `isInvalidError`, `isBadRequestError`, `isNotFoundError` and `isServerError`. As part of the migration path, we'll make it possible for Ember Ajax's helpers to detect Ember Data ajax errors: `AdapterError`, `TimeoutError` and `AbortError`. 

Using Ember Ajax's helpers to detect errors will allow us to deprecate Ember Data's `AdapterError`, `TimeoutError` and `AbortError`. (this part might need to be flashed out a bit further)
## Deprecation warnings

When the user opts-in to using Ember Ajax Service, the default ajax request/response methods on RESTAdapter becomes dead code that's never called. Detecting this dead code is difficult because it's never called. We'll add deprecation warnings that'll inform the user at built time about dead code in their adapters.
## Remove ajax code

We'll be able to remove the following methods from the RESTAdapter
    - handleResponse
    - isSuccess
    - isInvalid
    - ajaxOptions
    - parseErrorResponse
    - normalizeErrorResponse
    - generatedDetailedMessage

WIP PR for this issue is #4176

[...]

PR #4373, opened 2016-05-08
[FEATURE ds-customizable-error-key] Allow custom error keys in RESTAdapter

andrei-zhidkov:
Hey,
AFAIK, there is currently no way to customize the RESTAdapter's behavior on errors.
It will always look for `errors` top-level key, as described in [documentation](http://emberjs.com/api/data/classes/DS.RESTAdapter.html#toc_errors).
I think it would be nice to have a way to at least specify the key, which will be used to access the error data.
This PR adds an `errorKey` property to RESTAdapter, which allows to specify an expected key for failed request's payload to have.

### DIFF
--- a/FEATURES.md
+++ b/FEATURES.md
@@ -65,3 +65,7 @@ entry in `config/features.json`.
   * [404] `DS.NotFoundError`
   * [409] `DS.ConflictError`
   * [500] `DS.ServerError`
+
+- `ds-customizable-error-key`
+
+  Allows to specify a custom key by which the errors will be extracted by the `DS.RESTAdapter`.
--- a/addon/adapters/rest.js
+++ b/addon/adapters/rest.js
@@ -130,8 +130,9 @@ const {
 
   ### Errors
 
-  If a response is considered a failure, the JSON payload is expected to include
-  a top-level key `errors`, detailing any specific issues. For example:
+  If a response is considered a failure, the JSON payload by default is
+  expected to include a top-level key `errors`, detailing any specific issues.
+  For example:
 
   ```js
   {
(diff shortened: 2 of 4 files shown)


andrei-zhidkov:
Not really sure about handling the invalid responses (those having 422 status code) though.
On one hand, the error in such responses is expected to be a JSONAPI errors array, and having it under a non-JSONAPI-compliant key doesn't sound right.
On the other, it is strange to worry about JSONAPI in RESTAdapter, so maybe the concept of "invalid request" might change in the future.
Currently, both 422 and all other responses are handled the same way for consistency.

pangratz:
> AFAIK, there is currently no way to customize the RESTAdapter's behavior on errors.

Currently you would need to overwrite `Adapter#handleResponse` if your error payload doesn't contain an `errors` root level key:

```js
import RESTAdapter from "ember-data/adapters/rest";

export default RESTAdapter.extend({

  handleResponse(status, headers, payload) {
    if (!this.isSuccess(status, headers, payload)) {
      payload.errors = payload._my_errors_key;
      delete payload._my_errors_key;
    }

  }

});
```

I can see how this might not be ideal at the moment, but ember data will move to use [`ember-ajax`](https://github.com/ember-cli/ember-ajax) for making requests (see #4175). Once this is done, I think customizations for an error payload will be handled in the ajax service rather than the adapter.

> On one hand, the error in such responses is expected to be a JSONAPI errors array, and having it under a non-JSONAPI-compliant key doesn't sound right.
On the other, it is strange to worry about JSONAPI in RESTAdapter, so maybe the concept of "invalid request" might change in the future.

Currently the flow for handling errors is that the adapter returns an `InvalidError` from `handleResponse` and the `InvalidError` is initialized with the `errors` object from the payload; then `extractErrors` on the serializer is called to normalize the errors into the `JSON-API` format: so the adapter itself needs to take care that the errors are in the payload under the `errors` key, the normalization of the `errors` object is handled by the serializer.

bmac:
ping @tchak. Do you mind taking a look at this issue?

tchak:
@bmac yep this is on my TODO list for the week :)

andrei-zhidkov:
Closing this in favor of #4409.

PR #4374, opened 2016-05-08
Fix #4366: EmbeddedRecordsMixin embeds the record, not the snapshot

### LINKED ISSUE #4366: EmbeddedRecordsMixin embeds the record, not the snapshot
On lines:

https://github.com/emberjs/data/blob/v2.5.2/addon/serializers/embedded-records-mixin.js#L364

and

https://github.com/emberjs/data/blob/v2.5.2/addon/serializers/embedded-records-mixin.js#L228

The serialization of the embedded record is done like this:
```javascript
embeddedSnapshot.record.serialize({ includeId: true });
```

This means that instead of serializing the snapshot, the current record is serialized, which may or may not be the correct state.

dnachev:
The change replaces references to `snapshot.record.serialize()` with `snapshot.serialize()` in the `EmbeddedRecordsMixin` to ensure that the snapshot state is serialized rather than the current state of the record.

### DIFF
--- a/addon/serializers/embedded-records-mixin.js
+++ b/addon/serializers/embedded-records-mixin.js
@@ -226,7 +226,7 @@ export default Ember.Mixin.create({
     if (!embeddedSnapshot) {
       json[serializedKey] = null;
     } else {
-      json[serializedKey] = embeddedSnapshot.record.serialize({ includeId: true });
+      json[serializedKey] = embeddedSnapshot.serialize({ includeId: true });
       this.removeEmbeddedForeignKey(snapshot, embeddedSnapshot, relationship, json[serializedKey]);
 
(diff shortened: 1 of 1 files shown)


pangratz:
LGTM, thanks @dnachev.

@wecc can you do a double check if this looks good to you as well?

wecc:
👍

pangratz:
thanks @wecc 

@dnachev I would consider this a bugfix, can you update the commit message to something like `[BUGFIX beta] Re-use snapshot when serializing embedded relationships`... ?

dnachev:
@pangratz I've changed the commit message. Let me know if you need something else.

pangratz:
Thanks @dnachev

PR #4379, opened 2016-05-10
[BUGFIX beta] Fix resetting of properties to in-flight values

courajs:
Changing the value of an in-flight property, then changing back, had the
side effect that the record would be marked dirty once the save
completed.

`changedAttributes` would report that `name` had been changed from
'Thomas' to 'Thomas'

This is because the attr computed was only checking to see if the value
was being reset to the canonical value, not if it was being reset to the
in-flight value

### DIFF
--- a/tests/unit/model-test.js
+++ b/tests/unit/model-test.js
@@ -97,6 +97,51 @@ test("resetting a property on a record cause it to become clean again", function
   });
 });
 
+test("resetting a property to the current in-flight value causes it to become clean when the save completes", function(assert) {
+  assert.expect(4);
+
+  var person, finishSaving;
+
+  env.adapter.updateRecord = function(store, type, snapshot) {
+    // Make sure the save is async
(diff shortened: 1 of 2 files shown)


courajs:
Dug up while working on #4246

tests/unit/model-test.js
```diff
@@ -97,6 +97,52 @@ test("resetting a property on a record cause it to become clean again", function
...
+
+    saving = person.save();
+
+    assert.equal(person.get('name'), "Thomas");
+
+    person.set('name', 'Tomathy');
+    assert.equal(person.get('name'), "Tomathy");
+
+    person.set('name', 'Thomas');
+    assert.equal(person.get('name'), "Thomas");
```
courajs: Without these 4 lines 129-133, this test passed without code changes. Setting `name` to 'Tomathy', then back to 'Thomas' put the record in a bad state

bmac:
Thanks @courajs.

Issue #4306, opened 2016-04-07
Model blueprint has some trailing spaces that aren't "Suave"


sheriffderek:
Model blueprint has some trailing spaces that aren't "Suave" https://github.com/DockYard/ember-suave

`disallowTrailingWhitespace: Illegal trailing whitespace at models/xxx.js`

I don't know if it would be any different, but I generated a resource and a model and they were the same

Removing the spaces or the line would do the trick.

Does Ember Data conform to this?

RE: https://github.com/ember-cli/ember-cli/issues/5736

sbatson5:
I was looking at the blueprint and it looks like the newline is there because you can optionally add `attrs` when you generate your model:
`ember g model --attrs name`

Would it make sense to optionally newline it if attrs are present?  Not sure of an elegant solution.

fivetanley:
i think you may be able to remove the trailing whitespace by wrapping that in an `if`

```javascript
<%= importStatements %>

export default Model.extend({
  <% if (attrs) { %>
    <%= attrs %>
  <% } %>
});
```

The file is located here: https://github.com/emberjs/data/blob/master/blueprints/model/files/__root__/__path__/__name__.js

sbatson5:
I seem to always get an error when using `-%>`.  Tried it with just `<%= attrs -%>` and received an `Unexpected token` error.  Tried using it on just the `if` and received the same error.

The problem with wrapping it in an `if` is that if we do have attrs, we actually get 2 extra new lines.  Thought about doing a ternary:
`<%= typeof attrs !== 'undefined' ? attrs : '' %>`
But we would have to remove the leading spaces, otherwise we still get the empty whitespace error with Ember Suave.

My first thought was just prepending a space:

```
export default Model.extend({
<%= typeof attrs !== 'undefined' ? ' ' + attrs : '' %>
});
```
This ensures we indent the attrs but don't enforce the whitespace if there are none. However, leading spaces are added to all attrs after the first here:
https://github.com/emberjs/data/blob/master/blueprints/model/index.js#L81

Could remove it from the `index.js` file and rely on the blueprint to add the leading spaces. Doesn't seem like a good solution. I think I am missing something obvious 😛

pangratz:
Fixed by #4380

PR #4380, opened 2016-05-11
[BUGFIX beta] remove trailing whitespace from model blueprint

sbatson5:
Addresses: https://github.com/emberjs/data/issues/4306

This ensures that if attrs are present when generating a model, we indent them properly.  If no attrs are present, there should be no trailing whitespace.

### DIFF
--- a/blueprints/model/files/__root__/__path__/__name__.js
+++ b/blueprints/model/files/__root__/__path__/__name__.js
@@ -1,5 +1,5 @@
 <%= importStatements %>
 
 export default Model.extend({
-  <%= attrs %>
+<%= attrs.length ? '  ' + attrs : '' %>
 });


blueprints/model/files/__root__/__path__/__name__.js
```diff
@@ -1,5 +1,5 @@
 <%= importStatements %>
 
 export default Model.extend({
-  <%= attrs %>
+<%= attrs.length ? '  ' + attrs : '' %>
```
pangratz: I personally think that using an `if` here, might make this more readable:

``` js
export default Model.extend({
<% if (attrs) { %>
  <%= attrs %>
<% } %>
});
```

What do you think?

---

Another possibility would be to prepend the empty string within `blueprints/model/index.js`. We are already doing that for the [`needs` statement](https://github.com/emberjs/data/blob/afeb5e052b97b3113d993965d205bb725a742a63/blueprints/model/index.js#L82), so I think this would be fine for `attrs` too...

pangratz:
I left a tiny nitpick comment, apart from that: thank you very much for fixing this @sbatson5 🚀  !

Can you also prefix with `[BUGFIX beta]` so it is ensured that this is fixed in the next beta release?

blueprints/model/files/__root__/__path__/__name__.js
sbatson5: @pangratz Thanks for the feedback!

The problem I run into with the `if` condition is that if I do pass in attrs 
`ember g model foo --attrs name, title`
I get empty lines surrounding them:

```
export default Model.extend({

  name,: attr(),
  title: attr()

});
```

The ternary doesn't have that issue.  I tried the suggestion found here: https://github.com/emberjs/data/issues/4306#issuecomment-218454549 using `-%>` but received an `Unexpected token` error no matter where I used it (both for the closing `if` and `<%= attrs -%>`).

I saw that ternary used here https://github.com/emberjs/data/blob/master/blueprints/model-test/qunit-files/tests/unit/__path__/__test__.js#L5 so went that route. But I agree that it isn't that readable.

pangratz: Ah, ok. Thanks for the heads up @sbatson5. Let's leave it then the way it is, we can always change this later on.

pangratz:
Thank you kindly!

PR #4384, opened 2016-05-15
switch to ESLint from JSHint

fivetanley:
updated version of https://github.com/emberjs/data/pull/4360.

We will likely need Ember CLI 2.5.1 for these to pass as the `version` file is not getting loaded correctly due to a bug with `treeForAddon`.


bmac:
Superseded by https://github.com/emberjs/data/pull/4360

PR #4385, opened 2016-05-16
[cleanup] refactor relationship lookup/cache code

fivetanley:
This helps us remove state from the model classes themselves. Instead of placing data directly on the class definition, we create an object the class can use at runtime to help it look up its relationship info. This should not break the public API unless you are using classes directly without looking them up through the store, which might be broken in other ways anyway. This helps
align us towards ember's vision of how they want classes to be
injectable in the future.

TODO:
- [ ] [relationshipNames](https://github.com/emberjs/data/blob/master/addon/-private/system/relationships/ext.js#L400)
- [ ] [fields](https://github.com/emberjs/data/blob/master/addon/-private/system/relationships/ext.js#L530)

### DIFF
--- a/addon/-private/system/model/model-information-registry.js
+++ b/addon/-private/system/model/model-information-registry.js
@@ -1,0 +1,226 @@
+import Ember from 'ember';
+import { assert, warn } from "ember-data/-private/debug";
+import EmptyObject from "ember-data/-private/system/empty-object";
+import {
+  typeForRelationshipMeta,
+  relationshipFromMeta
+} from "ember-data/-private/system/relationship-meta";
+
+const {Map, MapWithDefault, get} = Ember;
+
+/**
+ * Contains relationship information that used to live directly on
+ * DS.Model subclasses. The ModelInformationRegistry is used to figure
+ * out inverse relationships for classes and cache that information.
+ *
+ * @private
+*/
+export default Ember.Object.extend({
+  init() {
+    this._super(...arguments);
+    this._inverseMaps = new EmptyObject();
+    this._relationshipsByName = new EmptyObject();
+    this._relationships = new EmptyObject();
+    this._relatedTypes = new EmptyObject();
+  },
+
+  relationshipsByNameFor(modelName) {
+    let map = this._relationshipsByName[modelName];
+
+    if (!map) {
+      map = this._relationshipsByName[modelName] = Map.create();
+      let store = this.get('store');
+      let modelType = store.modelFor(modelName);
+
+      modelType.eachComputedProperty((name, meta) => {
+        if (meta.isRelationship) {
+          meta.key = name;
+          let relationship = relationshipFromMeta(meta);
+          relationship.type = typeForRelationshipMeta(meta);
+          map.set(name, relationship);
+        }
+      });
+    }
+    return map;
+  },
+
+  relationshipsFor(modelName) {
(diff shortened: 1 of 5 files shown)


addon/-private/system/model/model-information-registry.js
```diff
@@ -0,0 +1,226 @@
...
+      // in relationships
+      modelType.eachComputedProperty((name, meta) => {
+        if (meta.isRelationship) {
+          meta.key = name;
+          let modelName = typeForRelationshipMeta(meta);
+
+          assert("You specified a hasMany (" + meta.type + ") on " + meta.parentType + " but " + meta.type + " was not found.", modelName);
+
+          if (!types.contains(modelName)) {
+            assert("Trying to sideload " + name + " on " + this.toString() + " but the type doesn't exist.", !!modelName);
```
fivetanley: should be `modelType.toString()` instead of `this`

addon/-private/system/model/model-information-registry.js
```diff
@@ -0,0 +1,226 @@
...
+import { assert, warn } from "ember-data/-private/debug";
+import EmptyObject from "ember-data/-private/system/empty-object";
+import {
+  typeForRelationshipMeta,
+  relationshipFromMeta
+} from "ember-data/-private/system/relationship-meta";
+
+const {Map, MapWithDefault, get} = Ember;
+
+/**
```
pangratz: Currently the API docs list all such docs, despite the class being marked as `@private` (see emberjs/website#2452). Can you change this to `/*` so it doesn't show up in the API docs?

addon/-private/system/model/model-information-registry.js
```diff
@@ -0,0 +1,226 @@
...
+      assert("You defined the '" + name + "' relationship on " + modelType + ", but you defined the inverse relationships of type " +
+        inverseType.toString() + " multiple times. Look at http://emberjs.com/guides/models/defining-models/#toc_explicit-inverses for how to explicitly specify inverses",
+        filteredRelationships.length < 2);
+
+      if (filteredRelationships.length === 1 ) {
+        possibleRelationships = filteredRelationships;
+      }
+
+      assert("You defined the '" + name + "' relationship on " + modelType + ", but multiple possible inverse relationships of type " +
+        this + " were found on " + inverseType + ". Look at http://emberjs.com/guides/models/defining-models/#toc_explicit-inverses for how to explicitly specify inverses",
```
fivetanley: should be `modelType` instead of `this`

addon/-private/system/model/model-information-registry.js
```diff
@@ -0,0 +1,226 @@
...
+const {Map, MapWithDefault, get} = Ember;
+
+/**
+ * Contains relationship information that used to live directly on
+ * DS.Model subclasses. The ModelInformationRegistry is used to figure
+ * out inverse relationships for classes and cache that information.
+ *
+ * @private
+*/
+export default Ember.Object.extend({
```
stefanpenner: it seems like this doesn't need to be an ember object, rather just a plain `function ModelInformation`

stefanpenner:
As an extraction this seems ok although I'm not entirely sure as to the direct value especially if its just another ember object, this doesn't actually address the issue. It seems like the callers of these API need to be fixed. With having a plan in-place to fix those, the value is this seems unknown.

Issue #4376, opened 2016-05-09
Update `isReloading` when a record is reloaded in the background


felixkiss:
As discussed with @bmac on Slack:

Currently, neither `record.isLoading` nor `record.isReloading` is set to true while the adapter performs a background reload of the specific record.

At least `isReloading` should be `true` during the time the adapter waits for a server response.

pete-the-pete:
@runspired after a quick and random look through the code, it seems that reload has some pretty specific meaning...the app is explicitly calling `reload` on the model or relationship. Does it make sense to have a new property to signify backgroundReloading specifically?

Also, the comments at [data/addon/-private/system/model/states.js](https://github.com/emberjs/data/blob/4bcf1f98e51708bb6c9eab82e8c29bfd5c6813da/addon/-private/system/model/states.js#L11) are pretty helpful, sad I missed the before

runspired:
We believe the request state RFC provides infrastructure necessary for an add-on or app to expose a flag for whether a record has any pending requests (not just a primary reload or a background reload).

efx:
This is the RFC: https://emberjs.github.io/rfcs/0466-request-state-service.html?highlight=request#request-state-service

PR #4386, opened 2016-05-16
[BUGFIX release] `isUpdating` flag is set correctly for `store.findAll`

### LINKED ISSUE #4339: isUpdating is broken since 2.4.3
Prior to version 2.5.0, we used `isUpdating` to determine the loading state of an ArrayProxy resulting from `findAll`. The reason for using `isUpdating` is that other states, such as `isFulfilled`, were not triggered during background loads.

To be clear, since version 2.5.0, `isUpdating` returns false when it used to return true. It used to return true when: (1) the `findAll` promise was resolving, where `reload: true` was set, and (2) the records were being (re)loaded in the background for ordinary `findAll` operations.

Did we miss an API change, or perhaps is there a better way of tracking the loading state of an ArrayProxy resulting from a find operation?

pangratz:
A regression for the `isUpdating` flag on the RecordArray returned for
`store.peekAll` has been introduced in #4316: the flag isn't set to true
anymore when records are reloaded or a background reloaded.

---

This closes #4339.

### DIFF
--- a/addon/-private/system/store.js
+++ b/addon/-private/system/store.js
@@ -1313,6 +1313,8 @@ Store = Service.extend({
     assert("You tried to load all records but you have no adapter (for " + typeClass + ")", adapter);
     assert("You tried to load all records but your adapter does not implement `findAll`", typeof adapter.findAll === 'function');
 
+    set(array, 'isUpdating', true);
+
     if (options.reload) {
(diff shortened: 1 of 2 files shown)


bmac:
Thanks @pangratz.

felixkiss:
Does this fix only `peekAll` or also `findAll`? I'm confused, because you mention the one and then the other one.

Does this fix #4376 as well?

PR #4387, opened 2016-05-17
Fix `ds.store.missing-records-from-adapter` with mixed adapters

cesarizu:
When using mixed adapters and referencing a model in one adapter from a model in a different adapter and coalescing finds, the adapter presents an error similar to:

```
Ember Data expected to find records with the following ids in the adapter response but they were missing: [...]
```

When loading the records directly, for any of the two models, there's no problem.

As an example one might have:
- A `project` model that is loaded from a JSON API backend, with coalescing finds by id
- A `token` model that is loaded from a `ember-local-storage` backend that references the project (with `inverse: null`)

Then trying to load all projects from the list of tokens like this:

``` javascript
Ember.RSVP.all(this.store.findAll(`token`).then((tokens) => tokens.mapBy('project')))
```

Executes a coalesced find many by id that gives the error above.

The root of this problem is that the records passed to `_flushPendingFetchForType` are not the same as found by the adapter, and they are rejected. This PR makes the comparison of those records by `id` instead of comparing directly.

There might be a more basic problem here but I don't seem to find the real culprit of it.

P.S. I need some guidance to write some tests for this.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,16 @@
 
 ### Master
 
+### Release 2.9.0-beta.1 (September 10, 2016)
+- [#4490](https://github.com/emberjs/data/pull/4490) [DOCS] No need to use Ember.run in succes and failure of wrapped getJSON
+- [#4493](https://github.com/emberjs/data/pull/4493) Addresses #4492
+- [#4503](https://github.com/emberjs/data/pull/4503) Update ember-try config to test against alpha.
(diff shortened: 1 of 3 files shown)


runspired:
I suspect that if there is a bug here that the underlying issue is us accidentally changing from internal model to record or from record to snapshot or some such, thus changing the reference.  However, this PR has odd merge commits in it, and we should be looking to address the underlying issue vs a fix that likely just masks it further. For this reason I am closing it.

A PR with a failing test would be welcome! :)

PR #4392, opened 2016-05-23
Fix lint errors caused by merging an older pr

bmac:


pangratz:
Since this only affects a commit for a feature, should this be prefixed with `[FEATURE ds-payload-hooks]`?

Otherwise looks good to me, thanks @bmac!

courajs:
Ember core uses homu to prevent this sort of thing, right? Have y'all considered pulling her in?

PR #4397, opened 2016-05-24
Rebase 4343

bmac:


addon/index.js
```diff
@@ -163,6 +166,18 @@ Object.defineProperty(DS, 'normalizeModelName', {
   value: normalizeModelName
 });
 
-Ember.lookup.DS = DS;
+Object.defineProperty(global, 'DS', {
+  configurable: true,
+  get() {
+    deprecate(
+      'Using the global version of DS is deprecated. Please either import ' +
+        'the specific modules needed or `import DS from \'ember-data\';`.',
```
pangratz: Tiny nitpick: do we want to advertise the `import DS from 'ember.data'`? I think when a pre-addonified-version of ember-data is used, with the shims all the necessary imports should be available via modules, so there is no need for importing `DS` namespace?

pangratz:
LGTM 👍

addon/index.js
bmac: I think suggesting users import `DS` is the easiest fix for them if they are using the `DS` global.

pangratz: You're right!

PR #4398, opened 2016-05-24
[FEATURE ds-overhaul-references] Fix inconsistencies with Reference#push

pangratz:
`push()` for a belongs-to and has-many reference accepts weird formats,
where it basically should only support a JSON-API document. If this
feature is enabled, it adds deprecation messages if something different
than a JSON-API document is pushed.


bmac:
👍

PR #4400, opened 2016-05-25
[BUGFIX beta] Use lower-case header fields

### LINKED ISSUE #4342: headers are case insensitive yet ember-data looks for case sensitive headers
[ RFC 2616](http://www.w3.org/Protocols/rfc2616/rfc2616.html) states that 

> Each header field consists of a name followed by a colon (":") and the field value. Field names are case-insensitive.

Yet ember data looks for headers using case sensitive lookups. 

https://github.com/emberjs/data/blob/master/addon/adapters/rest.js#L1138

```
generatedDetailedMessage: function(status, headers, payload, requestData) {
    var shortenedPayload;
    var payloadContentType = headers["Content-Type"] || "Empty Content-Type";
```

pangratz:
According to the spec header fields are case-insensitive.

https://tools.ietf.org/html/rfc7230#section-3.2
https://www.w3.org/Protocols/rfc2616/rfc2616-sec4.html#sec4.2

The value is also available via the original case of the header field;
this is kept for backwards compatibility.

---

This closes #4342

### DIFF
--- a/addon/-private/utils/parse-response-headers.js
+++ b/addon/-private/utils/parse-response-headers.js
@@ -18,6 +18,16 @@ export default function parseResponseHeaders(headersString) {
     value = value.join(':').trim();
 
     if (value) {
+      // according to the spec header fields are case insensitive, that's why
+      // we are using the lowercase version here
+      //
+      // https://tools.ietf.org/html/rfc7230#section-3.2
+      let lowerCaseField = field.toLowerCase();
+      headers[lowerCaseField] = value;
+
+      // deprecated way since headers are case insensitive; this is kept here
+      // for backwards compatibility and should be remove in the next major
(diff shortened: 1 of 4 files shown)


pangratz:
I've updated the implementation so the header field is available in the original case as well. This is needed for backwards compatibility, since folks might rely on the headers having a specific case.

I am not sure how we can deprecate this properly, except adding this as a breaking change in the `3.0` release...

tomdale:
Given that the header object returned here is passed into public API, can we instead return an implementation of https://developer.mozilla.org/en-US/docs/Web/API/Headers? For backwards compatibility, we can also set the (non-case-changed) headers on the object that is returned for backwards compatibility.

pangratz:
Closing in favor of #4456

Issue #4146, opened 2016-02-09
hasMany relationship loses its records if they fail to load


dnachev:
Twiddle: https://ember-twiddle.com/8be9daa561c3b5fd9b6c

My research so far indicates the following:
- When a record has failed to load, it will be unloaded from the store. This causes it to be removed from the relationships, which have a reference to it.
- Although the records are still available in the relationship canonical state, this doesn't help much. Both `parent.save()` and `parent.get('children').reload()` considers the relationships to be empty.

The above behavior is making it harder to handle errors loading hasMany collections, because only the first request to `parent.get('children')` will return a `PromiseManyArray`, which will be rejected (supposedly). Subsequent requests will just return successfully resolved empty `PromiseManyArray`. This means that it is essential that the very first instance is saved somewhere and propagated to the rest of the app.

An expected behavior would be to have the `parent.get('children')` return the same PromiseManyArray until `reload()` is called. This would allow anyone to get the current state of the affairs. This is how it is supposed to work, because of `Ember.computed` cached value, but it is invalidated when the problematic record is unloaded.

runspired:
The `PromiseManyArray` here is stable, I think the underlying issue was that the state of the backing array changes. I doubt this is still an issue today for an async relationship as the unloaded record would still be in the membership for the reload with how we do things, but we should add a test before closing this ticket.

runspired:
I recently spent time on this set of load-failure behaviors when making some of our unsupported emergent behaviors more recoverable (see #8846 and associated tickets). In doing so I've realized the original ask is not something we could ever support.

If an API fails to load a record, regardless of the reason, it needs to be removed from relationships until there is a successful payload containing that information.

What this means is that relationships that are expected to potentially fail need to be based on links instead of identity, which is a mode that is extremely well supported here.

PR #4401, opened 2016-05-25
[BUGFIX beta] References load() enhancements

tchak:
- ensure reference.load() is always returning a simple RSVP.Promise not an ObjectProxy

I don't expect any breaking changes here.

### DIFF
--- a/addon/-private/system/references/belongs-to.js
+++ b/addon/-private/system/references/belongs-to.js
@@ -5,6 +5,9 @@ import Reference from './reference';
 import isEnabled from '../../features';
 import { assertPolymorphicType, deprecate } from "ember-data/-private/debug";
 
+const {
+  RSVP: { resolve }
+} = Ember;
 
 /**
    A BelongsToReference is a low level API that allows users and
@@ -240,7 +243,7 @@ BelongsToReference.prototype.meta = function() {
    @return {Promise<record>} A promise that resolves with the new value in this belongs-to relationship.
 */
 BelongsToReference.prototype.push = function(objectOrPromise) {
-  return Ember.RSVP.resolve(objectOrPromise).then((data) => {
+  return resolve(objectOrPromise).then((data) => {
     let record;
 
     if (data instanceof Model) {
@@ -358,13 +361,11 @@ BelongsToReference.prototype.value = function() {
 */
 BelongsToReference.prototype.load = function() {
   if (this.remoteType() === "id") {
-    return this.belongsToRelationship.getRecord();
+    return resolve(this.belongsToRelationship.getRecord());
   }
 
   if (this.remoteType() === "link") {
-    return this.belongsToRelationship.findLink().then((internalModel) => {
-      return this.value();
-    });
+    return this.belongsToRelationship.findLink().then(() => this.value());
   }
 };
 
@@ -403,9 +404,7 @@ BelongsToReference.prototype.load = function() {
    @return {Promise} a promise that resolves with the record in this belongs-to relationship after the reload has completed.
 */
 BelongsToReference.prototype.reload = function() {
-  return this.belongsToRelationship.reload().then((internalModel) => {
-    return this.value();
-  });
+  return this.belongsToRelationship.reload().then(() => this.value());
 };
 
 export default BelongsToReference;
--- a/addon/-private/system/references/has-many.js
+++ b/addon/-private/system/references/has-many.js
@@ -152,11 +152,9 @@ HasManyReference.prototype.link = function() {
    @return {Array} The ids in this has-many relationship
 */
 HasManyReference.prototype.ids = function() {
-  let members = this.hasManyRelationship.members.toArray();
-
-  return members.map(function(internalModel) {
-    return internalModel.id;
-  });
+  return this.hasManyRelationship.members
+    .toArray()
+    .map(internalModel => internalModel.id);
 };
 
 /**
@@ -286,8 +284,7 @@ HasManyReference.prototype.push = function(objectOrPromise) {
         return record._internalModel;
       });
     } else {
-      let records = this.store.push(payload);
-      internalModels = Ember.A(records).mapBy('_internalModel');
+      internalModels = this.store.push(payload).map(record => record._internalModel);
 
       runInDebug(() => {
         internalModels.forEach((internalModel) => {
@@ -401,7 +398,7 @@ HasManyReference.prototype.value = function() {
 */
 HasManyReference.prototype.load = function() {
   if (!this._isLoaded()) {
-    return this.hasManyRelationship.getRecords();
+    return resolve(this.hasManyRelationship.getRecords());
   }
 
   return resolve(this.hasManyRelationship.manyArray);
@@ -442,7 +439,7 @@ HasManyReference.prototype.load = function() {
    @return {Promise} a promise that resolves with the ManyArray in this has-many relationship.
 */
 HasManyReference.prototype.reload = function() {
-  return this.hasManyRelationship.reload();
+  return this.hasManyRelationship.reload().then(() => this.value());
 };
 
 export default HasManyReference;
--- a/addon/-private/system/references/record.js
+++ b/addon/-private/system/references/record.js
@@ -1,6 +1,10 @@
 import Ember from 'ember';
 import Reference from './reference';
 
+const {
+  RSVP: { resolve }
+} = Ember;
+
 /**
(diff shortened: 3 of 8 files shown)


tchak:
@igorT do you see any problems with exposing `isLoaded()` on the references?

addon/-private/system/references/belongs-to.js
```diff
@@ -63,32 +63,33 @@ BelongsToReference.prototype.push = function(objectOrPromise) {
   });
 };
 
-BelongsToReference.prototype.value = function() {
-  var inverseRecord = this.belongsToRelationship.inverseRecord;
+BelongsToReference.prototype.isLoaded = function() {
+  let inverseRecord = this.belongsToRelationship.inverseRecord;
+
+  return inverseRecord && inverseRecord.isLoaded();
```
igorT: This seems to miss the case where it is null. If we know we are null, are we loaded?

addon/-private/system/references/belongs-to.js
```diff
@@ -63,32 +63,33 @@ BelongsToReference.prototype.push = function(objectOrPromise) {
...
   }
 
   return null;
 };
 
 BelongsToReference.prototype.load = function() {
-  if (this.remoteType() === "id") {
-    return this.belongsToRelationship.getRecord();
-  }
+  let record = this.value();
```
igorT: Does this cache get wiped correctly?

addon/-private/system/references/has-many.js
```diff
@@ -121,24 +121,26 @@ HasManyReference.prototype._isLoaded = function() {
...
   }
 
-  var manyArray = this.hasManyRelationship.manyArray;
   return Ember.RSVP.resolve(manyArray);
 };
 
 HasManyReference.prototype.reload = function() {
-  return this.hasManyRelationship.reload();
+  return this.hasManyRelationship.reload()
+    .then(() => this.value());
```
igorT: This seems to swallow the error case?

fivetanley: should still reject with the error.

addon/-private/system/references/record.js
```diff
@@ -26,18 +26,25 @@ RecordReference.prototype.push = function(objectOrPromise) {
...
+  if (this.isLoaded()) {
+    return this.internalModel.getRecord();
+  }
+
+  return null;
 };
 
 RecordReference.prototype.load = function() {
-  return this.store.findRecord(this.type, this._id);
+  return Ember.RSVP.resolve(this.store.findRecord(this.type, this._id));
```
igorT: What if the load fails?

tests/integration/references/belongs-to-test.js
```diff
@@ -230,7 +230,7 @@ test("push(promise)", function(assert) {
     push = familyReference.push(deferred.promise);
   });
 
-  assert.ok(push.then, 'BelongsToReference.push returns a promise');
+  assert.ok(push instanceof Ember.RSVP.Promise, 'BelongsToReference.push() returns a promise');
```
igorT: I think the official promise spec is to check for a .then.

igorT:
@tchak the semantics are a bit tricky with relationships that are not links. If you keep foreign keys on the belongs to side, how can you know if the has many has been loaded. We have an internal `hasData` for a lot of these purposes which @wecc wrote, might be useful to look at it

tests/integration/references/belongs-to-test.js
tchak: yes but what I am trying to test is that we are not returning `PromiseProxyObject`. If you have a better idea on how to test this...

fivetanley: maybe `! instanceof PromiseProxyObject` ? might be the most clear, as well as updating the description.

addon/-private/system/references/belongs-to.js
tchak: What cache are you talking about?

tests/integration/references/belongs-to-test.js
tchak: I changed, but I feel much less confident in the tests this way...

addon/-private/system/references/record.js
tchak: `Ember.RSVP.resolve(Ember.RSVP.reject("yolo")).catch(console.log.bind(console)) // => "yolo"`

But I added some tests to be sure :)

addon/-private/system/references/belongs-to.js
igorT: Actually this seems weird. So first you do `this.value()` which is  `this.belongsToRelationship.inverseRecord.getRecord()`  but then if it's null you do `this.belongsToRelationship.getRecord();` Do we just have two different things named `getRecord`? One on intneralModel and the other on the relationship? Maybe we should rename?

addon/-private/system/references/has-many.js
tchak: And it will reject with errors. I am only overriding resolved result. I added some tests to prouve me right :)

addon/-private/system/references/belongs-to.js
tchak: yes, we do. `inverseRecord` is actually an instance of `internalModel`. `getRecord` is the correct way of getting, well a record, from an `internalModel`. `getRecord` on the `BelongsToRelationship` is method to fetch a record (it deals internally with links and stuff).

tchak:
@igorT yes we already check `hasData` in the `has-many-reference#isLoaded`. I added the same check to `belongs-to-reference#isLoaded`.

addon/-private/system/store/finders.js
```diff
@@ -44,9 +44,6 @@ export function _find(adapter, store, typeClass, id, internalModel, options) {
     });
   }, function(error) {
     internalModel.notFound();
-    if (internalModel.isEmpty()) {
```
tchak: @igorT do you remember why we need this?

addon/-private/system/store/finders.js
```diff
@@ -80,7 +77,13 @@ export function _findMany(adapter, store, typeClass, ids, internalModels) {
 
       return internalModels;
     });
-  }, null, "DS: Extract payload of " + typeClass);
+  }, function(error) {
+    for (let i = 0; i < internalModels.length; i++) {
```
tchak: This basically mimics what we already doing in the `_find` method.

tchak:
@igorT @fivetanley I added a bunch of tests for failure/retry cases. Bunch of stuff were broken (before my refactoring). I have fixes for all of the broken stuff but it involves some changes to what looks to be non tested edge cases.

tchak:
I think this PR also fix #4146

addon/-private/system/references/belongs-to.js
```diff
@@ -5,6 +5,9 @@ import Reference from './reference';
 import isEnabled from '../../features';
 import { assertPolymorphicType, deprecate } from "ember-data/-private/debug";
 
+const {
+  RSVP: { resolve }
```
igorT: I think we don't do this

igorT: The destructuring i mean

hjdivad: What's the reason not to?  It's the array destructuring that's dangerous.

tests/integration/references/belongs-to-test.js
```diff
@@ -405,13 +406,62 @@ test("value() returns the referenced record when loaded", function(assert) {
   assert.equal(familyReference.value(), family);
 });
 
+test("value() is null when reference is loading", function(assert) {
+  var done = assert.async();
+  var deferred = Ember.RSVP.defer();
```
hjdivad: `var`?

tests/integration/references/belongs-to-test.js
```diff
@@ -405,13 +406,62 @@ test("value() returns the referenced record when loaded", function(assert) {
...
+
+  assert.equal(familyReference.value(), null);
+
+  deferred.resolve({
+    id: 1, name: "Coreleone"
+  });
+
+  load.then(family => {
+    assert.equal(familyReference.value(), family);
+    done();
```
hjdivad: we can just `return` the promise rather than using `assert.async`

runspired:
@tchak I really like this, can we reboot it?

runspired:
I think we should reboot this, but I'm closing this attempt in favor of a new attempt that takes into consideration `RecordData`.

PR #4403, opened 2016-05-26
[BUGFIX beta] Correctly coalesce URI encoded ids

### LINKED ISSUE #4190: Adapter not coalescing requests when ids contains ":"

pangratz:
When checking if the URL of a record is of the format
`...something/:id`, we need to decode the last part of the path before
comparing, since the part in the URL is encoded when the URL is
constructed within buildURL.

---

This fixes #4190.

### DIFF
--- a/addon/adapters/rest.js
+++ b/addon/adapters/rest.js
@@ -794,10 +794,14 @@ var RESTAdapter = Adapter.extend(BuildURLMixin, {
     var url = this.buildURL(snapshot.modelName, snapshot.id, snapshot);
 
     var expandedURL = url.split('/');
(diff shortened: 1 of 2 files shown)


bmac:
👍

Issue #4404, opened 2016-05-26
Expose more access to jqXHR object in handling adapter responses.


eshtadc:
I would like to add some custom functionality when an AJAX request returns and need access to the jqXHR object.  The use case is to validate a server signature from an authentication proxy server that requires knowing the responseText from the jqXHR request.  I can't get to this from handleResponse.  I tried looking at using ajaxOptions to push a success callback but I see that the rest adapter would overwrite this with it's own success callback.
https://github.com/emberjs/data/blob/fec260a38c3f7227ffe17a3af09973ce2718acca/addon/adapters/rest.js#L998

jQuery 1.5 and up allow an array of success callbacks which would give me one way of accomplishing this.

Another way would just be to allow handleResponse to get access to the underlying jqXHR rather than just the little bit of requestData.  It seems that ember-data is trying to insulate a bit from the type of AJAX call used, so I can see how this might be a bit distasteful.... 

I'm happy to work on a PR - but feel like there are a few directions this could go and want to work through the approaches before coding it up.

Of course - if there is some super obvious way of getting back the raw response info from the request that I'm missing - that would be awesome too :)

fivetanley:
I am disappointed that end users have lost access to this. I think it's definitely something we should restore. Maybe we should stick it on `requestData.jqXHR` ?

lkhaas:
Yeah - ideally it would be something I could access within handleResponse like that - so that I could validate and then return an error when the server signature is invalid.  Feels the most "ember-y".  I can take a stab at the code here and have something up shortly.

lkhaas:
OK - so this is my first ever ember-data PR.  I welcome any feedback on what I did right/wrong, etc.  Thanks!

stefanpenner:
closing in-favor of and RFC

PR #4406, opened 2016-05-26
[FEATURE ds-adapter-jqxhr] Enable adapter to have access to full jqXHR in response #4404

eshtadc:
I couldn't find an existing test that utilized the "ajax" method to ensure that the response is properly handled there as well.

### DIFF
--- a/FEATURES.md
+++ b/FEATURES.md
@@ -133,3 +133,9 @@ entry in `config/features.json`.
       }
       ```
 
+- `ds-adapter-jqxhr` [#4404](https://github.com/emberjs/data/pull/4404)
+
+  This provides access to the original jqXHR request in the data returned
+  to the `adapter.handleResponse` method.  It allows using any part of the request
+  to determine if the response is valid.  The `adapter.handleResponse`
+  requestData parameter will now include keys for `url`, `method`, and `jqXHR`.
--- a/addon/adapters/rest.js
+++ b/addon/adapters/rest.js
@@ -996,6 +996,9 @@ var RESTAdapter = Adapter.extend(BuildURLMixin, {
       var hash = adapter.ajaxOptions(url, type, options);
 
       hash.success = function(payload, textStatus, jqXHR) {
+        if (isEnabled('ds-adapter-jqxhr')) {
+          requestData.jqXHR = jqXHR;
+        }
 
         let response = adapter.handleResponse(
           jqXHR.status,
@@ -1012,6 +1015,10 @@ var RESTAdapter = Adapter.extend(BuildURLMixin, {
       };
 
       hash.error = function(jqXHR, textStatus, errorThrown) {
+        if (isEnabled('ds-adapter-jqxhr')) {
+          requestData.jqXHR = jqXHR;
+        }
+
         runInDebug(function() {
           let message = `The server returned an empty string for ${type} ${url}, which cannot be parsed into a valid JSON. Return either null or {}.`;
           let validJSONString = !(textStatus === "parsererror" && jqXHR.responseText === "");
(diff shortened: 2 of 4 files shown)


tchak:
May I ask what is the use case to have access to `xhr` object? Adapter hooks are designed explicitly to NOT expose `xhr` object. `xhr` is an implementation detail and in this case it's not even real `xhr`, but the `jQuery` wrapper that you trying to expose. It means `jQuery` becoming our public API. This is something we want to avoid because sometime in the feature we going to have to switch to `fetch` API and there will be no `xhr` object.

We tried to make sure we exposed all the useful information from `xhr` object. What's missing?

lkhaas:
Yes - I had commented in the issue #4404 that it did appear that jqXHR wasn't being exposed intentionally.  Our specific use case is to be able to validate a header that our server adds as part of our authorization handshake.  The header added is generated from some information from the request as well as the responseText (response body).  We store a few properties on the jqXHR itself, but I could get access to them during the beforeSend and then store them for validation later in the handleResponse function.  This still leaves me without the response body, however.  

I could also rewrite this PR to only add the response body (`responseText`) property in order to prevent the jQuery dependency.  Although the jQXHR is in itself just an abstraction as you point out.  It does expose all functionality of the underlying XMLHttpRequest object.  

What if ember-data still exposed the jqXHR but with the notion that it is the XMLHttpRequest object.  Then if one was to use something other than jQuery, they could still provide access to the XMLHttpRequest object.  The expectation could be that the base XMLHttpRequest is the only portion part of the ember-data api and things that jQuery adds (like promises, etc.) are just bonus?  Otherwise I feel like the ember-data will just keep adding individual values from the XMLHttpRequest object in a piecemeal fashion as needed by developers.

fivetanley:
> It means jQuery becoming our public API. 

Not necessarily. We have had plans to roll jQuery based stuff into its own adapter once ember-ajax nears completion. Since we're just sticking it on an options hash, we could document it as "jquery specific." This does not mean we have to support it on adapters that don't use jQuery.

I think it is very important for developers to have escape hatches, whether that's the raw XHR object or a jqXHR object. For instance, if your app needs to consume an API like Github's where the metadata is *very* header based, you can't do it currently without hacks or custom `$.ajax` rolling. You used to be able to access them and I remember working with a colleague at Instructure to expose jqXHRs pre-1.13.

I agree that most of these types of decisions should be available in some sort of raw form (e.g. we should be exposing headers), but it's important for users to have the escape hatch. Putting it on the options hash allows us to either deprecate or remove support in one isolated instance. It does not mean that jQuery is the new public facing API.

bmac:
Sorry for the slow response @eshtadc. I've been busy with work work.

We talked about this pr again in the last meeting and @igorT's concern with exposing a "jquery specific" escape hatch is because of Fastboot the same app may be expected to run in a jQuery and a non jQuery environment.

At this point I think the best option would be to expose some non-jquery specific wrapper for the XMLHttpRequest  object, or the raw XMLHttpRequest itself.

@tomdale Does fastboot expose request/response objects to the Ember app?

tchak:
@bmac `fastboot` is relying on `najax` which is an implementation `of `$.ajax` in node. It will expose a sort of `jqXHR` object.

eshtadc:
@bmac Yes - work work... it is the bane of my involvement in all sorts of things ;-)

I don't have any real understanding of fastboot yet.  I can definitely update this pr to be 1) re-rolled; 2) more generic to xhr object, but I'm at a loss of how to take fastboot into account.

gnanavelum:
is there a conclusion this?

I have a same use case to access the jqXhr on both success and false callbacks.

topaxi:
@gnanavelum there seems to be a migration away from jQuery, I think there needs to be a new solution to this as exposing `jqXHR` would block the migration.

runspired:
@eshtadc I am sad this languished so long :( We do want to improve the request layer, and are working on a concept that should hopefully resolve a lot of these issues around not enough information being exposed.

However, this approach isn't the direction we want to take both because we want to ensure Fastboot support and because we want to migrate away from jQuery and thus need to avoid leaking it.

PR #4408, opened 2016-05-27
Log an assertion if the response from createRecord does not have an i…

### LINKED ISSUE #4346: null model is stored in cache
So I am encountering this issue when using `DS.RESTAdapter` and getting an empty JSON object response when calling `DS.Model#save()`. The API returns an empty object and as expected nothing is stored in the store. However this happens: 

```javascript
DS.Store#findAll('user').get('length') // => 0 (good)
DS.Store#peekAll('user').get('length') // => 1 (bad)
```
The desired outcome would be for both to evaluate to `0`.

When I inspect on the object stored in the cache, every field is `null`. So even calling `unclearRecord` does not work. Calling `unclearAll('user')` does actually remove the cached object, but I don't think this should ever get into this state. I think `peekAll` should return no objects or there should be an error thrown when the adapter tries to parse a response with no `id`.

I created an example project that shows the behavior: https://github.com/trevorrjohn/bad-cached-data

Thanks!

bmac:
…d and the id was not already provided by the client

Closes #4346

### DIFF
--- a/addon/-private/system/store.js
+++ b/addon/-private/system/store.js
@@ -1634,7 +1634,7 @@ Store = Service.extend({
       this._backburner.schedule('normalizeRelationships', this, '_setupRelationships', internalModel, data);
       this.updateId(internalModel, data);
     }
-
+    assert(`Your ${internalModel.type.modelName} record was saved but it does not have an id. Please make the server provides an id in the createRecord response or you are setting the on the client side before saving the record.`, internalModel.id !== null);
     //We first make sure the primary data has been updated
     //TODO try to move notification to the user to the end of the runloop
     internalModel.adapterDidCommit(data);
--- a/tests/integration/adapter/rest-adapter-test.js
+++ b/tests/integration/adapter/rest-adapter-test.js
@@ -328,14 +328,14 @@ test("createRecord - a serializer's primary key and attributes are consulted whe
 test("createRecord - a serializer's attributes are consulted when building the payload if no id is pre-defined", function(assert) {
   var post;
   env.registry.register('serializer:post', DS.RESTSerializer.extend({
-    primarykey: '_id_',
-
     attrs: {
       name: '_name_'
     }
   }));
 
-  ajaxResponse();
+  ajaxResponse({
+    post: { '_name_': "The Parley Letter", id: '1' }
+  });
 
   run(function() {
     post = store.createRecord('post', { name: "The Parley Letter" });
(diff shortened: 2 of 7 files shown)


bmac:
ping @emberjs/ember-data-contributors this is ready for review.

pangratz:
👍  this looks good to me! Feel free to merge - I haven't because I was wondering if this should be prefixed so it lands in beta?

pangratz:
🎉

OmShiv:
With this update and assertion logging, my success callback is not executed. The control flow goes instead in the error call back.
```
record.save().then(
   () => {
       // before ember-data 2.8.0
   }, 
   () => {
       // after ember-data 2.8.0, with EmberError log for missing id.
   }
);
```

fivetanley:
That behavior is intended @OmShiv. Ember Data doesn't support records from the server which don't have IDs.

fivetanley:
Local records without IDs created using `createRecord` are, however allowed

OmShiv:
@fivetanley 
My APIs require a `PUT` for creating a record for the first time. And `POST` for update then onwards.
When I do:
```
record = store.createRecord(modelType, data);
record.save().then( ... )
``` 
It fails client side, since I don't jump in success callback. Though the n/w call succeeds.

Now if I set a random id, say `record.set('id', 1)` before the save, it works on client side without errors. But then, the updates don't work.
`record.save()` doesn't work thereafter, even if I manually set the `id` to 1 every save attempt.

fivetanley:
Does your server return an `id` for the `PUT` request to create the record?

OmShiv:
none of the PUT requests in our APIs return the id. Is there a way I can set id in a serializer?

fivetanley:
You could set a fake ID in the serializer, but I'm a little confused on to how you would then update it later without an ID the server knows about. From the server's perspective that could be a different record.

runspired:
@fivetanley @bmac this is actually breaking. I suspect that at a minimum the assertion needs moved within the conditional, but also not all payloads return IDs, previously a null ID for a record that already knew it's ID would work, now it will not.  So I also suspect this should warn with deprecation, not fail.  Finally, I'm seeing situations in which although an ID is provided client side, this assertion is still triggered: it appears to be when createRecord is used to generate a record that never actually receives an ID (yes, we have those, TL;DR think a message)

Working on a fix.

runspired:
Fix is being worked on here: https://github.com/emberjs/data/pull/4544

PR #4409, opened 2016-05-29
[FEATURE ds-normalize-invalid-errors] add normalizeInvalidErrorResponse hook to the adapter

tchak:
This is another take on the problem reported in #4373. The hook solution IMHO aligns better with current adapter API.


tchak:
closing this, pending proper RFC

PR #4411, opened 2016-05-29
Outdated documentation on DS.Model.eachRelationship and DS.Model.relationshipsByName return values

### LINKED ISSUE #4388: Outdated documentation on DS.Model.eachRelationship and DS.Model.relationshipsByName return values
The documentation seems to indicate that the returned `type` should be a `DS.Model` Object, and I believe this is the way it has worked previously, but now when testing with Ember.Data 2.5.2 `type` is a string with the model name. Is this a bug or is the documentation incorrect? `parentType` seems to be a `DS.Model` just like the documentation states.

http://emberjs.com/api/data/classes/DS.Model.html#method_eachRelationship
http://emberjs.com/api/data/classes/DS.Model.html#property_relationshipsByName

```
Ember.get(this.get('store').modelFor('customer'), 'relationshipsByName')
```
```
Map {_keys: OrderedSet, _values: EmptyObject, size: 1}
_keys: OrderedSet
_values: EmptyObject
  st322: Object
  isRelationship: true
  key: "orders"
  kind: "hasMany"
  options: Object
  parentType: (unknown mixin)
  type: "order"
```
I was able to track down the change to this commit: https://github.com/emberjs/data/commit/405f9c503c704ecb5dae569f54529b46e5e53282

Seems like it returned the type in some cases and the string in others previously. But I guess now it should always return a string?
```
-  if (typeof modelName === 'string') {
 -    if (meta.kind === 'hasMany') {
 -      modelName = singularize(modelName);
 -    }
 -    typeClass = store.modelFor(modelName);
 -  } else {
 -    typeClass = meta.type;
 -  }
 -
 -  return typeClass;
 +  return singularize(normalizeModelName(modelName));
```
Are they intended to always return an string? In that case I guess the documentation should be updated.

mriska:
Corrected documentation regarding return values from the following methods/properties:
- DS.Model eachRelationship
- DS.Model relationshipsByName

Fixes #4388

### DIFF
--- a/addon/-private/system/relationships/ext.js
+++ b/addon/-private/system/relationships/ext.js
@@ -473,9 +473,9 @@ export const RelationshipsClassMethodsMixin = Ember.Mixin.create({
 
     var relationshipsByName = Ember.get(Blog, 'relationshipsByName');
     relationshipsByName.get('users');
-    //=> { key: 'users', kind: 'hasMany', type: App.User }
+    //=> { key: 'users', kind: 'hasMany', type: 'user', options: Object, isRelationship: true }
     relationshipsByName.get('owner');
-    //=> { key: 'owner', kind: 'belongsTo', type: App.User }
+    //=> { key: 'owner', kind: 'belongsTo', type: 'user', options: Object, isRelationship: true }
     ```
 
     @property relationshipsByName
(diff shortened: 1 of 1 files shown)


pangratz:
Thank you kindly! 🌟

PR #4413, opened 2016-05-31
Small improvement for the badIdFormatAssertion

grapho:
Just a small helpful improvement.

Could potentially save users tens of minutes finding the cause. :tada:

My original assumption was that a payload was not including an `id`.. then i discovered it was my `findRecord()` call


bmac:
Thanks @grapho.

PR #4418, opened 2016-06-02
remove DEPRECATION: Using the global version of DS is deprecated when running tests

frunjik:
remove DEPRECATION: Using the global version of DS is deprecated.
Imported DS explicitly in tests/ember-data-initializers.js.

### DIFF
--- a/tests/ember-data-initializers.js
+++ b/tests/ember-data-initializers.js
@@ -1,11 +1,11 @@
 /* eslint no-extra-semi: "off" */
 
-import DS from 'ember-data';
-
 ;(function() {
-  /* globals Ember */
+  /* globals Ember, require */
   var K = Ember.K;
   Ember.onLoad('Ember.Application', function(Application) {
+
+    var DS = require('ember-data').default;
 
     Application.initializer({
       name:       "ember-data",
--- a/tests/ember-data-initializers.js
+++ b/tests/ember-data-initializers.js
@@ -1,8 +1,9 @@
 /* eslint no-extra-semi: "off" */
 
+import DS from 'ember-data';
+
 ;(function() {
   /* globals Ember */
-  /* globals DS */
   var K = Ember.K;
   Ember.onLoad('Ember.Application', function(Application) {
 


frunjik:
I have no idea why this change would break the build ... ?

frunjik:
it looks like:
- tests green when: "ember test --server"
- tests break when: "ember test --environment=production" (some broccoli error)

Help?

frunjik:
I tried several different permutations (of the import statements) to get rid of errors.

But it looks like as soon as i put an import statement (either importing Ember or DS), at the top of the file, i get a brocolli error when running:
"ember test --environment=production" 

the tests build and run when:
"ember test" 

Is this an (unexpected) side effect of switching to ESlint / and or using the iffy ?

Any hints appreciated

@bmac

bmac:
The build is failing with the production flag because confusingly the `tests/ember-data-initializers.js` get added to the Ember Data build for bower when built in production mode. See here: https://github.com/emberjs/data/blob/master/lib/javascripts.js#L49.

I'm not sure what the best solution is at the moment although I can think of a few hacks to just make it work. Let me get back to you.

frunjik:
Thx for that,

I also see a typo there: emberDataInitialierPath

- J

2016-06-10 22:53 GMT+02:00 Brendan McLoughlin <notifications@github.com>:

tests/ember-data-initializers.js
```diff
@@ -1,8 +1,9 @@
 /* eslint no-extra-semi: "off" */
 
+import DS from 'ember-data';
```
bmac: Part of the reason why the build fails is this file is not transformed before it gets added to the bower build output. One thing we could do is remove this line and add  `var DS = require('ember-data').default;` inside the [iffe](https://en.wikipedia.org/wiki/Immediately-invoked_function_expression).

tests/ember-data-initializers.js
```diff
@@ -1,8 +1,9 @@
...
+
 ;(function() {
   /* globals Ember */
-  /* globals DS */
```
bmac: If we add `var DS = require('ember-data').default;` we can remove this globals declaration but we will need to add `/* globals require */`.

frunjik:
Thx,

Your 'hacks' make the build work ...
But they don't solve the issue of tests/ember-data-initializers.js wrongly? being included ...

@bmac

rwjblue:
I am unsure why exactly we need to duplicate these initializers. Ideally we can fix that issue.

One possible solution might be to do what "normal" ember-cli apps do, and use ember-load-initializers...

bmac:
@rwjblue I have merged this pr as is and opened a new issue for cleaning up the duplicated initializers https://github.com/emberjs/data/issues/4437. I'd like to sync up with you later on how to use `ember-load-initializers`.

rwjblue:
sounds good

PR #4176, opened 2016-02-20
WIP Allow users to opt-in to use ember-ajax

taras:
For context, please refer to #4175

PR #4430, opened 2016-06-15
WIP Allow users to opt-in to use ember-ajax

yads:
This is a feature flag wrapping of #4176


runspired:
Closing this because

- @tchak is refactoring towards an ajax story that supports fetch and fastboot
- We've decided to revisit the request layer in more detail (see #5384 for instance) and will be proposing a new direction very soon.

PR #4438, opened 2016-06-21
add `relationship` property to findHasMany RESTAdapter docs

fivetanley:


pangratz:
If I am not mistaken the `relationship` is the object which describes the relationship – something like `{key: "comments", kind: "hasMany", type: "comment", options: {}, isRelationship: true}`...

PR #4440, opened 2016-06-22
Add examples to the Serializer API docs

bmac:


addon/serializer.js
```diff
@@ -27,10 +27,34 @@ import Ember from 'ember';
 export default Ember.Object.extend({
 
   /**
-    The `store` property is the application's `store` that contains all records.
-    It's injected as a service.
-    It can be used to push records from a non flat data structure server
-    response.
+    The `store` property is the application's `store` that contains
```
fivetanley: so it's a bit confusing here because you have the store passed into the function's parameter list, and also `this.store`, which gets set when the serializer is instantiated.

I think this documentation is for `this.store` because of `@property`, but isn't used in the example below.

bmac: Good call.

addon/serializer.js
```diff
@@ -63,6 +93,32 @@ export default Ember.Object.extend({
...
+          }
+        });
+
+        return json;
+      },
+    });
+    ```
+
     @method serialize
     @param {DS.Model} record
```
pangratz: Can you update this to be `DS.Snapshot`?

pangratz:
Nice writeup! 💯  👍

addon/serializer.js
```diff
@@ -27,10 +27,21 @@ import Ember from 'ember';
 export default Ember.Object.extend({
 
   /**
-    The `store` property is the application's `store` that contains all records.
-    It's injected as a service.
-    It can be used to push records from a non flat data structure server
-    response.
+    The `store` property is the application's `store` that contains
+    all records.  It can be used to look up serializer for other model
```
wecc: Extreme Nitpicking™️ – double space here :)

Also `look up serializer for other model type` => `look up serializers for other model types` ?

addon/serializer.js
```diff
@@ -63,8 +93,34 @@ export default Ember.Object.extend({
...
+        });
+
+        return json;
+      },
+    });
+    ```
+
     @method serialize
-    @param {DS.Model} record
+    @param {DS.Snapshot} record
```
wecc: Should this parameter be named `snapshot` instead of `record`?

pangratz:
Sorry @wecc, GitHub didn't show up your comment when I was merging. Wanna open up a follow up PR?

wecc:
Planck time trolled us! 🕐 

#4614

Issue #173, opened 2012-03-23
RESTAdapter issue


joeyjojo-zz:
The following simple example fails with:

Result of expression 'type.proto' [undefined] is not a function

``` javascript
window.App = Ember.Application.create();

App.store = DS.Store.create({
    revision: 4,
    adapter: DS.RESTAdapter.create()
});

App.Contact = DS.Model.extend({
    firstName: DS.attr('string')
});

App.contactsController = Ember.ArrayProxy.create({
    content: App.store.findAll(App.Contact)
});
```

Whereas when its a fixture adapter it seems to work.

I can fix it by mimicking the code in the fixture adapter in the RESTAdapter in findAll like so

``` javascript
  findAll: function(store, type) {
    var root = this.rootForType(type), plural = this.pluralize(root);

    this.ajax(this.buildURL(root), "GET", {
      success: function(json) {
        var ids = json[plural].map(function(item, index, self){ return item.id; });
        store.loadMany(type, ids, json[plural]);
        //store.loadMany(type, json[plural]);
        this.sideload(store, type, json, plural);
      }
    });
  },
```

But the error appears to be in the store.loadMany when you only pass in data.

I am open to it being my implementation mistake!

PR #4441, opened 2016-06-22
[BUGFIX beta] prevent record duplication due to EmbeddedRecordsMixin

tarzan:
This is a proposed fix to #1829

When the EmbeddedRecordsMixin is used to persist embedded records to the backend, the parent record is handled as expected. However, nested children are neglected in the run loop and upon pushing a returned payload from the server into the store we end up with duplication. This commits adds a procedure for cleaning up these flagged 'uncommitted' records from the store before pushing the created ones.

### DIFF
--- a/addon/-private/system/store.js
+++ b/addon/-private/system/store.js
@@ -2437,7 +2437,7 @@ function _commit(adapter, store, operation, snapshot) {
           types.forEach(function (type) {
             var existingRecords = store.peekAll(type);
             var uncommitted = existingRecords.filter(function (record) {
-              return record.get('isNew') && !record.get('isSaving');
+              return record.get('id') === 'embedded';
             });
             // and remove them
             uncommitted.invoke('unloadRecord');
--- a/addon/-private/system/store.js
+++ b/addon/-private/system/store.js
@@ -2432,6 +2432,16 @@ function _commit(adapter, store, operation, snapshot) {
       if (adapterPayload) {
         payload = normalizeResponseHelper(serializer, store, typeClass, adapterPayload, snapshot.id, operation);
         if (payload.included) {
+          // first find possibly uncommitted records from this inclusion
+          var types = payload.included.map(function (record) { return record.type; });
+          types.forEach(function (type) {
+            var existingRecords = store.peekAll(type);
+            var uncommitted = existingRecords.filter(function (record) {
+              return record.get('isNew') && !record.get('isSaving');
+            });
+            // and remove them
+            uncommitted.invoke('unloadRecord');
+          });
           store.push({ data: payload.included });
         }
         data = payload.data;
--- a/tests/integration/adapter/rest-adapter-test.js
+++ b/tests/integration/adapter/rest-adapter-test.js
@@ -2615,6 +2615,49 @@ test("createRecord - with EmbeddedRecordsMixin does not duplicate sideloaded rec
   });
 });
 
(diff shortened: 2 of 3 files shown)


tarzan:
This is my first try at fixing this issue that has been haunting our software at times. Please let me know what you guys think. :)

addon/-private/system/store.js
```diff
@@ -2432,6 +2432,16 @@ function _commit(adapter, store, operation, snapshot) {
       if (adapterPayload) {
         payload = normalizeResponseHelper(serializer, store, typeClass, adapterPayload, snapshot.id, operation);
         if (payload.included) {
+          // first find possibly uncommitted records from this inclusion
+          var types = payload.included.map(function (record) { return record.type; });
+          types.forEach(function (type) {
+            var existingRecords = store.peekAll(type);
+            var uncommitted = existingRecords.filter(function (record) {
+              return record.get('isNew') && !record.get('isSaving');
```
bmac: Looks like this will unload all record of a type the `isNew` state even if they were not the records sent in the original request?

tarzan: correct. It does. There currently is no test that fails showcasing that scenario, but it's easy to write one. 

Problem is, I am stumbling with a way to let the store know which records are being persisted to the backend. Should the EmbeddedRecordMixin flag the records by changing their currentState to 'inFlight'? How should I handle this in the Ember data runloop?

tarzan:
Ok. So I added a test that fails due to my current implementation. Can anyone offer some advice on how to improve it?

bmac:
@igorT I think you had some ideas on how handle this?

tests/integration/adapter/rest-adapter-test.js
```diff
@@ -2573,6 +2573,91 @@ test("createRecord - sideloaded records are pushed to the store", function(asser
...
+        id: 2,
+        name: 'First comment'
+      }]
+    }
+  });
+  var post;
+
+  run(function() {
+    post = store.createRecord('post', { name: 'The Parley Letter' });
+    post2 = store.createRecord('post', { name: 'Another Parley Letter' });
```
sbatson5: I doesn't look like `post2` is defined above on line 2643 like `post`.

tarzan: yeah, the linter fails 'cause of that as well. It has no bearing on the failing of this test, however.

tarzan:
@jcbvm correct. That's why there still is a failing spec. I am still looking for suggestions on how to handle this. Problem is that the `EmbeddedRecordsMixin` is mixed into the serializer and at this point we have to flag the records that are being embedded and persisted this way. Then when the store commits the action we have to unload the flagged records or merge them with the newly received ones from the response-payload.

tarzan:
Ok. I changed it around a bit. Now the embedding of a (hasmany) record that has not yet persisted, will get a dummy 'embedded' ID. As a result, the store will be able to find these records when processing the payload.

To me, this implementation seems a bit hacky, but I'm having a hard time coming up with a different approach. Any feedback will be greatly appreciated! 

@igorT how does it differ from the ideas you have?

tarzan:
Any news on this?

jcbvm:
@tarzan see #173

runspired:
The duplication issue can be resolved via use of `lid` and identifiers when the IDENTIFIERS ff is turned on #6366 . See https://github.com/emberjs/rfcs/pull/403

PR #4443, opened 2016-06-23
Remove deprecation on using normalizeHash when running tests

frunjik:
This removes (2) deprecation messages on using deprecated method normalizeHash, by using normalize instead.


frunjik:
Again - i have no idea why one of the automated builds is failing (and the other succeeds) ...
Locally all tests pass ...

@bmac If u are bored (thx for help sofar :))

bmac:
I wouldn't worry about the appveyor test failing it looks like a network error unrelated to the build. 

Since these tests are intended to test `normalizeHash` I think we need to keep these tests as is incase to catch any changes that may break `normalizeHash` before the next major release (when we release Ember Data 3.0 and no longer have to support `normalizeHash` I think we can just drop them all together). 

Poking around the `ember-dev` repo which Ember Data includes in its tests it looks like there is a global `ignoreDeprecation` function that you could use to wrap the body of these 2 tests. https://github.com/emberjs/ember-dev/blob/master/addon/test-helper/deprecation.js#L67

pangratz:
First of all, thank you very much @frunjik for tackling these issues ❤️  🚀 

---

Another possibility would be to add an assertion for the expected deprecation message, like [used here](https://github.com/emberjs/data/blob/0e15f2f605e7616a6148a5b5feddd2e150adbeca/tests/integration/serializers/rest-serializer-test.js#L1081). You would also need to change from `test` [to `testInDebug`](https://github.com/emberjs/data/blob/0e15f2f605e7616a6148a5b5feddd2e150adbeca/tests/integration/serializers/rest-serializer-test.js#L1068) - this is needed so the tests success when they are run with the prod build.

PR #4445, opened 2016-06-24
[BUGFIX beta] Fixes issue with GET requests appending ?{} to url

JeroenvdV:
This fixes a typo in the REST adapter comparing `request.type`, which doesn't exist. Either `request.method` or `hash.type` was meant. The symptom is that `?{}` is appended to the url of any GET requests by jQuery.

### DIFF
--- a/addon/adapters/rest.js
+++ b/addon/adapters/rest.js
@@ -1341,7 +1341,7 @@ if (isEnabled('ds-improved-ajax')) {
       hash.context = this;
 
       if (request.data) {
-        if (request.type !== 'GET') {
+        if (hash.type !== 'GET') {
           hash.contentType = 'application/json; charset=utf-8';
           hash.data = JSON.stringify(request.data);
         } else {
--- a/addon/adapters/rest.js
+++ b/addon/adapters/rest.js
@@ -1341,7 +1341,7 @@ if (isEnabled('ds-improved-ajax')) {
       hash.context = this;
 
       if (request.data) {
-        if (hash.type !== 'GET') {
+        if (request.method !== 'GET') {
           hash.contentType = 'application/json; charset=utf-8';
           hash.data = JSON.stringify(request.data);
         } else {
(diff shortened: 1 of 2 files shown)


addon/adapters/rest.js
```diff
@@ -1341,7 +1341,7 @@ if (isEnabled('ds-improved-ajax')) {
       hash.context = this;
 
       if (request.data) {
-        if (request.type !== 'GET') {
+        if (hash.type !== 'GET') {
```
pangratz: Can you change this to be `request.method` instead?

pangratz:
Thanks for this @JeroenvdV. I think it would be good to have a test to catch future regressions for this.

Can you add a test to the `rest-adapter-test.js` which looks something like this?

```js

if (isEnabled('ds-improved-ajax')) {

  test("_requestToJQueryAjaxHash works correctly for GET requests - GH-4445", function(assert) {
    let done = assert.async();
    let server = new Pretender();

    server.get('/posts/1', function(request) {
      assert.equal(request.url, "/posts/1", "no query param is added to the GET request");

      return [201, { "Content-Type": "application/json" }, JSON.stringify({ post: { id: 1 } })];
    });

    run(function() {
      let post = store.findRecord('post', 1);

      post.then(function() {
        server.shutdown();
        done();
      });
    });
  });

}
```

After this, this is good to go 🚀

JeroenvdV:
I added your test verbatim as I'm not in a position to test it fully at the moment.

pangratz:
Just one more tiny change: if you add a newline at the end of the test then ESLint is happy and this is good to be merged!

pangratz:
@JeroenvdV, I should have been more clear: I meant adding a newline at the end of the test file. I am very sorry for the back and forth in this PR, I can see how that is not very motivating and I try to communicate more clearly in the future 😔  ...

If you don't have time for this today, I can add the newline later manually...

pangratz:
Closing in favor of rebased version #4466

pangratz:
@JeroenvdV I merged this via the rebased #4466. Thank you very much! 🚀

PR #4446, opened 2016-06-24
`modelHasAttributeOrRelationshipNamedType` should not show up in the …

bmac:
…API docs


bmac:
It is currently displaying on the [AdapterError](http://emberjs.com/api/data/classes/DS.AdapterError.html) page even though it is not a method on AdapterError.

pangratz:
👍  thank you kindly!

PR #4447, opened 2016-06-24
Remove initializer methods from the API docs

bmac:
(they are showing up on the DS.Model class)

Add missing API docs for Ember.Date

Add minor doc update to the DS namespace API docs.


pangratz:
👍  thanks!

PR #4448, opened 2016-06-24
[BUGFIX beta] The adapter should call `ajax` instead of the new metho…

bmac:
…ds if it has been customized.

cc @pangratz @rwjblue

### DIFF
--- a/addon/adapters/json-api.js
+++ b/addon/adapters/json-api.js
@@ -5,6 +5,7 @@
 import Ember from 'ember';
 import RESTAdapter from "ember-data/adapters/rest";
 import isEnabled from 'ember-data/-private/features';
+import { deprecate } from 'ember-data/-private/debug';
 
 /**
   @class JSONAPIAdapter
@@ -99,7 +100,7 @@ var JSONAPIAdapter = RESTAdapter.extend({
     @return {Promise} promise
   */
   findMany(store, type, ids, snapshots) {
-    if (isEnabled('ds-improved-ajax')) {
+    if (isEnabled('ds-improved-ajax') && !this._hasCustomizedAjax()) {
       return this._super(...arguments);
     } else {
       var url = this.buildURL(type.modelName, ids, snapshots, 'findMany');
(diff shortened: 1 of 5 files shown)


addon/adapters/json-api.js
```diff
@@ -99,7 +100,7 @@ var JSONAPIAdapter = RESTAdapter.extend({
     @return {Promise} promise
   */
   findMany(store, type, ids, snapshots) {
-    if (isEnabled('ds-improved-ajax')) {
+    if (isEnabled('ds-improved-ajax') && !this._hasCustomizedAjax()) {
```
pangratz: This only works if we support https://github.com/ember-cli/babel-plugin-feature-flags/pull/4. Are the necessary dependencies in `ember-data` updated so this is satisfied? I'm just asking; haven't had time to follow recent developments in `ember-data` too closely...

rwjblue: Should be allowed, as that was released as 0.2.1 and we are depending on `^0.2.0` ([see here](https://github.com/emberjs/data/blob/master/package.json#L27)).

tests/integration/adapter/rest-adapter-test.js
```diff
@@ -2593,3 +2593,44 @@ testInDebug("warns when an empty response is returned, though a valid stringifie
...
+    assert.ok(true, 'The ajaxOptions method should be called when it is overridden');
+    return oldAjaxOptions.apply(this, arguments);
+  };
+
+  assert.expectDeprecation(function() {
+    run(function() {
+      store.findRecord('post', 1);
+    });
+  }, /RESTAdapter#ajaxOptions has been deprecated/)
+});
```
pangratz: Altough the feature flag is go'ed, I think for consistency this tests should be behind a `if (isEnabled("ds-improved-ajax"))` check, no? Just in case we quickly need to de-go the feature. Or am I missing something here?

addon/adapters/json-api.js
pangratz: Ah, awesome. Sorry for making you check this, I could have easily done this myself 😔

config/features.json
```diff
@@ -1,10 +1,10 @@
 {
-  "ds-boolean-transform-allow-null": null,
-  "ds-improved-ajax": null,
+  "ds-boolean-transform-allow-null": true,
+  "ds-improved-ajax": true,
```
pangratz: Are the updates in this file intended? What this basically does is aligning `master` with what we already have in `beta`, right?

pangratz:
Apart from a few comments, this looks great! Thanks @bmac for tackling this!

config/features.json
bmac: Correct.

PR #4450, opened 2016-06-26
[BUGFIX beta] flag ds-serialize-ids-and-types should not be enabled by default

lkhaas:
The feature is enabled by default in the repo and therefore throwing deprecation warnings.

### DIFF
--- a/config/features.json
+++ b/config/features.json
@@ -2,7 +2,7 @@
   "ds-boolean-transform-allow-null": null,
   "ds-improved-ajax": null,
   "ds-pushpayload-return": null,
-  "ds-serialize-ids-and-types": true,
+  "ds-serialize-ids-and-types": null,
   "ds-extended-errors": null,
   "ds-links-in-record-array": null,
   "ds-overhaul-references": null,


bmac:
Hi @lkhaas. I think the `ds-serialize-ids-and-types` feature was [released](http://emberjs.com/blog/2016/06/09/ember-data-2-6-released.html#toc_changes-in-ember-data-2-6) with Ember Data 2.6 so I think we want to keep it enabled by default. 

I'd like to know which deprecation warning you are seeing from this feature.

Also since this feature has been released we can now remove this feature flag. If you'd like to make a PR similar to this one https://github.com/emberjs/data/pull/4348/files that removes all of the references to `ds-serialize-ids-and-types` that would be great. Thanks.

lkhaas:
I'm seeing deprecation warnings when running the tests, but completely missed its release in 2.6.  I'll go ahead and make a PR to remove the flag.  Thanks for the example PR.

lkhaas:
Closing in favor of removing the feature flag in PR https://github.com/emberjs/data/pull/4451

Issue #4416, opened 2016-06-01
Running tests shows Warnings and Deprecation's


frunjik:
When i run the tests i get warnings and deprecation's.

Before i try to see if i can get rid of them i want to understand if this is expected behavior, or am i doing something wrong ?
- WARNING: FEATURE["ds-serialize-ids-and-types"] is set as enabled, but FEATURE flags are only available in canary builds.
- DEPRECATION: Using the global version of DS is deprecated. Please either import the specific modules needed or `import DS from 'ember-data';`. [deprecation id: ember-data.global-ds]
      at HANDLERS.(anonymous function) (http://localhost:4200/assets/vendor.js:16464:7)
      at raiseOnDeprecation (http://localhost:4200/assets/vendor.js:16372:12)
      at HANDLERS.(anonymous function) (http://localhost:4200/assets/vendor.js:16464:7)
      at invoke (http://localhost:4200/assets/vendor.js:16480:7)
      at deprecate (http://localhost:4200/assets/vendor.js:16433:32)
      at Object.deprecate (http://localhost:4200/assets/vendor.js:26472:37)
      at deprecate (http://localhost:4200/assets/vendor.js:69315:40)
      at get (http://localhost:4200/assets/vendor.js:82215:44)
      at http://localhost:4200/assets/tests.js:21:21
- DEPRECATION: The adapter returned an array for the primary data of a `queryRecord` response. This is deprecated as `queryRecord` should return a single record. [deprecation id: ds.serializer.rest.queryRecord-array-response]
      at HANDLERS.(anonymous function) (http://localhost:4200/assets/vendor.js:16464:7)
      at raiseOnDeprecation (http://localhost:4200/assets/vendor.js:16372:12)
      at HANDLERS.(anonymous function) (http://localhost:4200/assets/vendor.js:16464:7)
      at invoke (http://localhost:4200/assets/vendor.js:16480:7)
      at deprecate (http://localhost:4200/assets/vendor.js:16433:32)
      at Object.deprecate (http://localhost:4200/assets/vendor.js:26472:37)
      at deprecate (http://localhost:4200/assets/vendor.js:69315:40)
      at http://localhost:4200/assets/vendor.js:85334:48
      at runInDebug (http://localhost:4200/assets/vendor.js:16645:5)
- WARNING: Interacting with a record errors object will no longer change the record state.
- WARNING: There is no attribute or relationship with the name `notInMapping` on `post`. Check your serializers attrs hash.
[...]

PR #4451, opened 2016-06-27
[CLEANUP beta] Remove feature flag for ds-serialize-ids-and-types (shipped in 2.6) #4416

lkhaas:
This was originally found by looking into ember-data test deprecation warnings.  Then it was pointed out by @bmac that the feature actually shipped in 2.6, so now submitting a PR to remove the feature flag.


bmac:
Thanks @lkhaas.

PR #4456, opened 2016-06-28
Provide a Headers polyfill for case insensitive access to response he…

bmac:
…aders

### DIFF
--- a/addon/-private/utils/headers.js
+++ b/addon/-private/utils/headers.js
@@ -1,0 +1,82 @@
+/* global Symbol */
+// Partially implements Headers from the Fetch API
+// https://developer.mozilla.org/en-US/docs/Web/API/Headers
+function HeadersPolyfill(headers) {
+  headers = headers || {};
+  this.headers = {};
+
+  for (var header in headers) {
+    this.headers[header] = headers[header].split(', ');
+  }
+}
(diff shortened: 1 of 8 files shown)


addon/-private/utils/headers.js
```diff
@@ -0,0 +1,82 @@
...
+  var entries = [];
+
+  for (var key in this.headers) {
+    var values = this.headers[key];
+    for (var index = 0; index < values.length; ++index ) {
+      entries.push(values[index]);
+    }
+  }
+
+  return entries[Symbol.iterator]();
```
rwjblue: I'm not sure we can use Symbol here. IIRC, it is unsupported by IE10 and IE11.

addon/-private/utils/headers.js
```diff
@@ -0,0 +1,82 @@
...
+    var values = this.headers[key];
+    for (var index = 0; index < values.length; ++index ) {
+      entries.push([key, values[index]]);
+    }
+  }
+
+  return entries[Symbol.iterator]();
+};
+
+HeadersPolyfill.prototype.get = function(header) {
```
bmac: Since there is a `get` method this may cause a user to assume this is an Ember Object.

PR #4461, opened 2016-07-04
Ember Data is no Dummy 😎

rwjblue:
I kept getting confused between Ember Data's tests and an addon I had been testing.


bmac:
Thanks @rwjblue.

Issue #4469, opened 2016-07-08
Belongs to not updating when response contains a change with the previous value


chrisgame:
If the response to a save of a model that includes a belongs to relationship is successful but that response contains data that sets the belongs to back to it's original value the change is not applied to the store.

pangratz:
Fixed by #4470

PR #4470, opened 2016-07-08
Belongs to not updating when response contains a change with the previous value

chrisgame:
If the response to a save of a model that includes a belongs to relationship is successful but that response contains data that sets the belongs to back to it's original value the change is not applied to the store. #4469

### DIFF
--- a/addon/-private/system/relationships/state/belongs-to.js
+++ b/addon/-private/system/relationships/state/belongs-to.js
@@ -36,6 +36,7 @@ BelongsToRelationship.prototype.setCanonicalRecord = function(newRecord) {
   } else if (this.canonicalState) {
     this.removeCanonicalRecord(this.canonicalState);
   }
+  this.flushCanonicalLater();
   this.setHasData(true);
   this.setHasLoaded(true);
 };
--- a/tests/unit/model/relationships/belongs-to-test.js
+++ b/tests/unit/model/relationships/belongs-to-test.js
@@ -167,6 +167,50 @@ test("async belongsTo relationships work when the data hash has already been loa
   });
 });
 
+test("when response to saving a belongsTo is a success but includes changes that reset the users change", function(assert) {
+  var Tag = DS.Model.extend();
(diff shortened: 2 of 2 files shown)


addon/-private/system/relationships/state/belongs-to.js
```diff
@@ -31,11 +31,14 @@ BelongsToRelationship.prototype.setRecord = function(newRecord) {
 };
 
 BelongsToRelationship.prototype.setCanonicalRecord = function(newRecord) {
+  if (this.canonicalState) {
+    this.removeCanonicalRecord(this.canonicalState);
+  }
```
bmac: I think the real fix is the fact that `removeCanonicalRecord` ends up calling `this.flushCanonicalLater();` somewhere in its call stack. Instead of moving this check up I'd prefer to see it left where it is an add an explicit call to `this.flushCanonicalLater();` before `this.setHasData(true);`.

jgwhite:
@bmac would you mind giving this another look over?

bmac:
Looks good. 👍 Thanks @chrisgame.

chrisgame:
No problem, thank you @bmac

chrisccerami:
I'm not sure if this is a breaking change or if our app was relying on a load bearing bug, but this is causing us some issues. We have a workaround, but I just wanted to check that this is indeed the intended behavior.

bmac:
@chrisccerami I believe this is a bug fix as now relationship have the same behavior as attributes when the server returns a new value for the property.  I am interested in hearing more about how this change is impacting your app. Do you mind sharing more details about your desired behavior and work around?

chrisccerami:
Oh, it's not about desired behavior, and if this is the correct behavior then I'm sure it means someone did something silly in the way they built the particular part that broke. I believe it was something like we were updating a record and its hasMany association, saving the record, and then with the response we were saving the hasMany. This was working as of 2.6.2, but not 2.7.0. In order to work around it, we're now updating the record and saving it, and then updating the hasMany and saving that. I was only surprised to have something break in a minor release.

PR #4473, opened 2016-07-11
Fix key remapping for id/ids belongsTo with embedded-records-mixin

mdentremont:
Closes 4472. This implements the same semantics for `key` that are used
in JSONSerializer, for both belongsTo and hasMany records when
serialized as id/ids.

### DIFF
--- a/addon/serializers/embedded-records-mixin.js
+++ b/addon/serializers/embedded-records-mixin.js
@@ -198,13 +198,16 @@ export default Ember.Mixin.create({
     var includeIds = this.hasSerializeIdsOption(attr);
     var includeRecords = this.hasSerializeRecordsOption(attr);
     var embeddedSnapshot = snapshot.belongsTo(attr);
-    var key;
     if (includeIds) {
-      key = this.keyForRelationship(attr, relationship.kind, 'serialize');
+      var serializedKey = this._getMappedKey(relationship.key, snapshot.type);
+      if (serializedKey === relationship.key && this.keyForRelationship) {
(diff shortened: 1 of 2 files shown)


mdentremont:
Would anyone care to take a look? It's a pretty straightforward change which follows the footsteps of another previously merged PR.

pangratz:
@mdentremont sorry for the delay here. I'm pretty busy at the moment, but I will try to find some time on the weekend to take a look!

addon/serializers/embedded-records-mixin.js
```diff
@@ -198,13 +198,16 @@ export default Ember.Mixin.create({
     var includeIds = this.hasSerializeIdsOption(attr);
     var includeRecords = this.hasSerializeRecordsOption(attr);
     var embeddedSnapshot = snapshot.belongsTo(attr);
-    var key;
     if (includeIds) {
```
pangratz: Hmm, the logic inside `if (includeIds)` basically is the same compared to the one in [`super.serializeBelongsTo`](https://github.com/emberjs/data/blob/db49f961653246318f67f86d0427c929810fae44/addon/serializers/json.js#L1184-L1202), so it looks like this should definitely be DRY'ed up...

pangratz:
The PR seems fine to me, thanks @mdentremont for taking the time to open this PR!

It seems that the logic for the embedded records and the serializers contain stuff which can be shared, so this should definitely be DRY'ed up.

I'm gonna merge this one as refactoring definitely is another issue.

Thanks again @mdentremont 🚀

mdentremont:
@pangratz Thanks for the merge!

Issue #4488, opened 2016-07-29
Brittle Promise Usage


stefanpenner:
The following snippets make an invalid assumption, that is all code they execute cannot possibly throw. Although it may seems reasonable to assume this, it unfortunately isn't, especially for framework provided code, that calls into other extension points.

- [ ] https://github.com/emberjs/data/blob/master/addon/adapters/rest.js#L1024-L1049
- [ ] https://github.com/emberjs/data/blob/master/addon/adapters/rest.js#L1008-L1020
- [ ] https://github.com/emberjs/data/blob/master/addon/adapters/rest.js#L1406-L1417
- [ ]  https://github.com/emberjs/data/blob/master/addon/adapters/rest.js#L1421-L1446

For example: `adapter.handleResponse` is a public API extension point, which has a pretty high likelihood of failing. https://github.com/emberjs/data/blob/master/addon/adapters/rest.js#L1041

Anyways, when wrapping contents of callbacks invoked by foreign async we must be careful, or unexpected errors will result breaking the completion guarantee of the given function, and can result in unrecoverable blank screens for our users. One such use-case is described here: https://github.com/emberjs/ember.js/issues/13919

The solution is to wrap the entire foreign invoked callback in a try/catch + reject https://github.com/emberjs/data/blob/master/addon/adapters/rest.js#L1024-L1049

```js
options.error = function() {
  try {
    // other code.
  } catch(e) {
    reject(e);
  }
}
```

For those that will argue that this is overly defensive and we cannot possibly guard against all possible failures, I would agree in the general case, but the above falls is merely our reasonable due diligence.

pangratz:
I believe this is fixed by #4489. Feel free to reopen if I am mistaken.

PR #4489, opened 2016-08-01
[BUGFIX beta] Fix Brittle Promise Usage

bmac:
This pr attempts to fix the issues identified in https://github.com/emberjs/data/issues/4488 by wrapping the ajax success/error code in a try/catch.

cc @stefanpenner @pangratz for review.

### DIFF
--- a/addon/adapters/rest.js
+++ b/addon/adapters/rest.js
@@ -25,6 +25,8 @@ const {
   get
 } = Ember;
 
+const Promise = Ember.RSVP.Promise;
+
 /**
   The REST adapter allows your store to communicate with an HTTP server by
   transmitting JSON via XHR. Most Ember.js apps that consume a JSON API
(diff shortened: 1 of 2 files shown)


addon/adapters/rest.js
```diff
@@ -1454,6 +1405,50 @@ if (isEnabled('ds-improved-ajax')) {
 
 }
 
+function ajaxSuccess(adapter, jqXHR, payload, requestData, resolve, reject) {
```
stefanpenner: lets not pass reject/resolve here, instead simply `return` or `throw` (or `return Promise.reject()`)

bmac:
@stefanpenner I've updated the pr do you mind giving it another review?

addon/adapters/rest.js
```diff
@@ -1001,52 +1003,30 @@ var RESTAdapter = Adapter.extend(BuildURLMixin, {
...
-          parseResponseHeaders(jqXHR.getAllResponseHeaders()),
-          payload,
-          requestData
-        );
-
-        if (response && response.isAdapterError) {
-          Ember.run.join(null, reject, response);
-        } else {
-          Ember.run.join(null, resolve, response);
+        var successValue;
```
pangratz: This avriable can be moved inside the `try catch` block.

addon/adapters/rest.js
pangratz: These unused params can be removed...

pangratz:
LGTM 👍

addon/adapters/rest.js
```diff
@@ -25,6 +25,8 @@ const {
   get
 } = Ember;
 
+const Promise = Ember.RSVP.Promise;
```
locks: could you import RSVP directly here, instead of using the Ember namespaced one?

fivetanley: I don't think we have support for that for the globals build

PR #4490, opened 2016-08-01
[DOCS] No need to use Ember.run in succes and failure of wrapped getJSON

chadhietala:
/cc @stefanpenner


addon/adapter.js
```diff
@@ -102,10 +102,9 @@ export default Ember.Object.extend({
 
         return new Ember.RSVP.Promise(function(resolve, reject) {
           Ember.$.getJSON(`/${type.modelName}/${id}`).then(function(data) {
```
stefanpenner: `resolve(Ember.$.getJSON(`/${type.modelName}/${id}`))` would also be the same.

bmac:
Thanks @chadhietala.

Issue #4459, opened 2016-07-01
`createRecord()` obscures the `modelName` property


nwhittaker:
Twiddle: https://ember-twiddle.com/7fdbb7303a5e95ba4fee7e3268dac4e4

The `modelName` property on a model generated via `createRecord()` appears to be undefined. By comparison, the same property on a model generated via `modelFor()` is populated as expected.

bmac:
Looks like `modelName` isn't documented property http://emberjs.com/api/data/classes/DS.Model.html#property_modelName. `modelName` should be listed as a static property that lives on the Model Class. 

`modelFor()` returns the Model class definition for a type.

While `createRecord()` returns the instance of the class.

You can access `modelName` from an instance by using the `constructor` property to access the Model class and then accessing `modelName` from that.

```js
var record = this.get('store').createRecord('my-model');
record.constructor.modelName;
```

nwhittaker:
Thanks for clarifying, @bmac. Another point of confusion for me was that [`modelFor`](http://emberjs.com/api/data/classes/DS.Store.html#method_modelFor) is documented as returning a `DS.Model` object (same as `createRecord()`.

pangratz:
@nwhittaker I've opened #4498 to clarify the documentation. Can you take a look and check if that helps folks in the future might stumbling over this? Thanks!

nwhittaker:
👍 for [adding `@static`](https://github.com/emberjs/data/pull/4498/files#diff-9e3eb8996b822cf8719c23d9ab4dc16eR1006) -- I believe this helps a lot in clarifying the `modelName` property.

In regards to the description change for `modelFor`, I think the crux of my confusion was that `@returns` doesn't distinguish between whether the `DS.Model` object is a class object or an instance. I see the description points this out, but I'm not sure enough emphasis exists in the general documentation to make "model class" jump out as an important descriptor. Putting "model class" in italics might help quickly communicate this. Adding or moving the description of the return type to the `@returns` line might also be something to consider.

Thanks for soliciting the feedback, @pangratz.

wecc:
Closing this since #4498 has been merged.

PR #4498, opened 2016-08-15
[DOC] clarify docs on Model#modelName and Store#modelFor

pangratz:
Addresses #4459


fivetanley:
@pangratz fwiw i think we should also alias this on instances. it's very handy to have and prevents people from looking stuff up on `constructor` if they don't need to.

bmac:
@fivetanley do you mind opening a new issue for that change?

@pangratz I'm 👍 on this pr.

PR #4499, opened 2016-08-15
[BUGFIX beta] add assertions for reference methods on DS.Model

pangratz:
Assertions are thrown when the specified relationship doesn't exist or
the passed type mismatches the one of the actual relationship.

---

This issue has been raised in the slack channel so adding assertions should help preventing this in the future.


fivetanley:
is there a way to check for existence of a relationship without getting an exception?

pangratz:
Yep:

```js
let Person = this.store.modelFor("person");
let rels = Ember.get(Person, "relationshipsByName"); 
let personRel = rels.get("father");
let { kind } = personRel;
assert.ok(kind, "belongsTo");
```

bmac:
👍

Issue #4426, opened 2016-06-08
IDs on relationships for new records are serialized as null


cristinawithout:
When you create a new record (Record B) and set its relationship to another record (on Record A), saving the existing record (Record A) without first saving the new record (Record B) serializes the relationship to the new record with ID null.

`{
  "data": {
    //...
    "relationships": {
      "relB": {
        "data" : {
          "type": "modelB",
          "id": null
        }
      }
   }
  }
}`

Demo:
https://ember-twiddle.com/a27a8e9da784895c266ddd3e0b39bb1f?openFiles=templates.application.hbs%2C

My recommendation is that an object in a relationship with ID null should be excluded during serialization. Otherwise, you're leaving it up to your server (for which you may not control the code) to figure out how to handle these null IDs. If the server doesn't know and saves the null relationship, next time you request data for the record from the server, you have an invalid relationship with a null ID.

The issue of null IDs was also reported in https://github.com/emberjs/data/issues/3953, but attached to the request to serialize attributes instead. That issue was closed because the JSON API spec does not support relationship attributes, but the underlying issue of null IDs went unaddressed.

mixonic:
@cristinawithout When you remove a relationship, the server must be passed that removal somehow. I expect the two systems (sending an empty initial relationship and sending an empty relationship that had content) are related.

`null` as an id just represents "there is nothing at this relationship", yes?

cristinawithout:
No. It's not when removing a relationship that a null ID is being serialized. It's when you add a newly created recordA to a relationship on recordB. Then when saving recordB before saving recordA, you get id null for recordA.

Also a removal is not done via a null id. Per the JSON API spec, it's done by removing the object, which in a belongsTo relationship serializes to null (not an object with ID null) and on a hasMany serializes to an array with only the remaining items in the relationship or an empty array (if there are no other remaining items).

Null IDs in a relationship are not covered in the spec, which leaves whoever is implementing the server to guess at what Ember Data's current implementation intends when it sends them.

A shorter example than the twiddle:

```
// modelA
export default Model.extend({
   belongsToRel: DS.belongsTo('modelB')
})
```

```
// modelB
export default Model.extend({
   inverseBelongsToRel: DS.belongsTo('modelA')
})
```

```
var modelA = store.findRecord('modelA', '1');
var modelB = store.createRecord('modelB', { ... });
modelA.set('belongsToRel', modelB);
modelA.save();
```

modelA is serialized to:

[...]

PR #4500, opened 2016-08-17
WIP don't serialize new belongsTo relationship

pangratz:
This addresses #4426

### DIFF
--- a/tests/integration/serializers/json-api-serializer-test.js
+++ b/tests/integration/serializers/json-api-serializer-test.js
@@ -286,6 +286,35 @@ test('options are passed to transform for serialization', function(assert) {
   env.store.serializerFor('user').serialize(user._createSnapshot());
 });
 
+test('a belongsTo relationship with a new record is serialized as { data: null }', function(assert) {
+  run(function() {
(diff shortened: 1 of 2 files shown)


tests/integration/serializers/json-api-serializer-test.js
```diff
@@ -286,6 +286,35 @@ test('options are passed to transform for serialization', function(assert) {
...
+    let user = store.createRecord('user');
+    handle.set('user', user);
+
+    let serialized = handle.serialize({ includeId: true });
+    assert.deepEqual(serialized, {
+      data: {
+        type: 'handles',
+        id: '1',
+        relationships: {
+          user: {
```
wecc: Shouldn't the whole `user` relationship be absent instead of `data: null`?

samselikoff:
I agree with @wecc, `data: null` is a positive statement that "this relationship is now null", whereas in most of these situations, the relationship is never null, it's just going from one state to some future state (once the id is known).

pangratz:
Added in #5317, thanks @ryanto for pushing this forward ❤️  👍 .

PR #4503, opened 2016-08-19
Update ember-try config to test against alpha.

rwjblue:


rwjblue:
The failures right now are due to an auto-run assertion that is triggered by calling `  new Ember.ArrayProxy({ content: []});` (in the `RecordArrayManager`).

With https://github.com/emberjs/ember.js/pull/14091, this branch passes but we need clarification from folks more run-loop aware that that is the right solution.

rwjblue:
emberjs/ember.js#14091 was closed in favor of emberjs/ember.js#14093, but this is green now...

PR #4510, opened 2016-08-24
Heimdall instrumentation

runspired:
This is a WIP branch that allows us to easily measure where our performance costs are in different record loading scenarios.

Example:

screen shot 2016-08-23 at 9 45 30 pm


runspired:
The .json files generated by the instrumentation runner should be added to git-ignore, that's the bulk of the line changes.  Will prune those out in a moment

stefanpenner:
Awesome. I suspect we may (once this matures) shake out pieces into their own repo's and modules, but keeping it together during experimentation seems legit.

fivetanley:
beat me to it. nice!

fivetanley:
it would be great to allow instrumentation from the outside. on example would be to add some instrumentation for "how many models/requests were loaded during this route transition", but that's probably a separate thing i could implement

(sorry thinking aloud in github comments :P)

runspired:
I made some progress on reducing the cost of the instrumentation itself, so we may be able to leave some of it in for people if they want and use babel transforms to strip the rest of it out for production builds.

runspired:
Results of a recent round of measurements are below, I'm currently working on enabling ED to strip out heimdall instrumentation for production builds.

screen shot 2016-08-26 at 8 08 29 pm

hjdivad:
@fivetanley 

> it would be great to allow instrumentation from the outside. on example would be to add some instrumentation for "how many models/requests were loaded during this route transition", but that's probably a separate thing i could implement

This is an excellent suggestion and, unless I've misunderstood you, also something heimdalljs explicitly supports.  This is how we use it to track fs usage within ember-cli for instance.

see:
  - https://github.com/heimdalljs/heimdall-fs-monitor
  - https://github.com/ember-cli/ember-cli/blob/master/lib/models/builder.js#L13-L15

runspired:
Getting happier, I've squashed and rebased. Got more instrumentation tweaks to make, need to push out an alpha release for heimdall-query and the heimdal babel transform, and figure out what is breaking tests.

runspired:
@igorT not sure why this test is failing, could use some help tracking it down.

lib/babel-build.js
```diff
@@ -18,6 +18,7 @@ function babelOptions(libraryName, _options) {
       'es6.constants',
       'es6.modules'
     ],
+    externalHelpers: true,
```
runspired: this is from implementing a plugin to strip classCallCheck while investigating alternative adapters, it should be removed.

package.json
```diff
@@ -47,6 +48,7 @@
   "devDependencies": {
     "bower": "^1.6.5",
     "broccoli-asset-rev": "^2.1.2",
+    "broccoli-babel-transpiler": "^5.6.1",
```
runspired: this creeping in here was accidental, will be removed.

package.json
```diff
@@ -77,12 +79,18 @@
...
     "ember-try": "^0.2.5",
     "ember-watson": "^0.7.0",
+    "express": "^4.14.0",
     "github": "^0.2.4",
     "glob": "5.0.13",
+    "heimdalljs": "~0.3.0-alpha2",
+    "heimdall-query": "^0.0.4",
+    "json-api-mock-server": "0.0.2",
     "loader.js": "^4.0.0",
+    "lodash": "^4.15.0",
```
runspired: I'm not sure lodash is directly needed anymore, I'll do a quick review to make sure. I believe this was only here for what is now heimdall-query.

tests/dummy/app/helpers/reopen-instrumentation.js
```diff
@@ -0,0 +1,51 @@
+/* global heimdall */
```
runspired: despite setting this globally many places ESLint was yelling at me if I didn't set it at the top of files :/ This should be worked out.

tests/dummy/app/routes/query/route.js
```diff
@@ -0,0 +1,55 @@
...
+    let store = this.get('store');
+    heimdall.stop(storeToken);
+
+    // console.log('ember-data-start', stats.startTime);
+    return store.query(modelName, params)
+      .then((records) => {
+        heimdall.stop(token);
+        // console.log('ember-data-end', stats.endTime);
+        // console.profileEnd(label);
+        // console.info('findAll#item -> fetched ' + get(records, 'length') + ' records');
```
runspired: all these console calls should be removed.

addon/-private/system/model/internal-model.js
```diff
@@ -44,6 +44,24 @@ function retrieveFromCurrentState(key) {
 }
 
 var guid = 0;
+
+const [
+  __a, __b, __c, __d, __e, __f, __g, __h, __i,
```
stefanpenner: we should use more descriptive names?

fivetanley:
i think a small facade layer between heimdall and ember data would be nice. this allows us to document it as part of the Ember Data API and also change the implementation more easily if we needed to in the future.

runspired:
@fivetanley Describe what you are thinking. I had no intentions of the instrumentation being public beyond a user being able to leave it in their build and provide their own heimdall instance if desired.

stefanpenner:
@fivetanley I believe the public API you seek is heimdall itself.

rwjblue:
>  I believe the public API you seek is heimdall itself.

Does it have a public API? I didn't think it was stable yet (don't see a 1.x or anything)...

runspired:
@rwjblue we're narrowing in on it, 0.3.x is close to stable

runspired:
@rwjblue and fwiw, the public API in heimdall 0.3.x-alpha3 is at this point highly unlikely to change (the last thing likely to change was the switch over from returning an array to returning an object when calling registerMonitor which I just made this morning), we're mostly playing around with internals at this point.

fivetanley:
Is this PR for benchmarking, or is it desired to also be useful to end users?

stefanpenner:
> Is this PR for benchmarking, or is it desired to also be useful to end users?

first benchmarking, later end users.

runspired:
@fivetanley this PR is part of a concerted effort to have consistent instrumentation across ember-cli, ember, glimmer2, and ember-data to help us more quickly identify performance problems and regressions.

Benchmarking is secondary, but nice.  One of the awesome things about this approach is that our instrumentation cost is low enough that it can be left in for end users when they desire, but easily stripped out for the primary case.  Soon an end user wanting in depth measurements can select to not strip heimdall and provide their own heimdall instance for use.

chadhietala:
@runspired I'm going to kill some CI jobs that you have running that are clogging up the concurrency.

runspired:
@chadhietala np, I finally found a way to make tests to work locally so I'm finishing up what I gots there.

runspired:
Remaining test failures are inconsistent and I do not believe they are related to the PR.  Running tests locally (usually) all scenarios will pass.

addon/-private/system/store.js
```diff
@@ -1093,7 +1137,10 @@ Store = Service.extend({
     assert("You tried to load a query but you have no adapter (for " + typeClass + ")", adapter);
     assert("You tried to load a query but your adapter does not implement `query`", typeof adapter.query === 'function');
 
-    return promiseArray(_query(adapter, this, typeClass, query, array));
+    let pA = promiseArray(_query(adapter, this, typeClass, query, array));
+
+    pA.finally(() => { heimdall.stop(token); });
+    return pA;
```
runspired: This should be wrapped in `instrument` so that we don't have an extra `finally` dangling around.

addon/-private/system/store.js
```diff
@@ -1355,9 +1402,14 @@ Store = Service.extend({
   findAll(modelName, options) {
     assert("You need to pass a model name to the store's findAll method", isPresent(modelName));
     assert('Passing classes to store methods has been removed. Please pass a dasherized string instead of '+ Ember.inspect(modelName), typeof modelName === 'string');
+    let token = heimdall.start('store.findAll');
+
     var typeClass = this.modelFor(modelName);
 
-    return this._fetchAll(typeClass, this.peekAll(modelName), options);
+    let fetch = this._fetchAll(typeClass, this.peekAll(modelName), options);
+    fetch.finally(() => { heimdall.stop(token); });
```
runspired: same as above

addon/-private/system/store.js
```diff
@@ -2042,17 +2103,20 @@ Store = Service.extend({
...
+      heimdall.stop(token);
       return null;
     }
 
     assert(`Expected an object in the 'data' property in a call to 'push' for ${data.type}, but was ${Ember.typeOf(data.data)}`, Ember.typeOf(data.data) === 'object');
 
     var internalModel = this._pushInternalModel(data.data);
 
+    heimdall.stop(token);
     return internalModel.getRecord();
```
runspired: wrap in instrument and call stop after `getRecord()`

addon/-private/system/store/common.js
```diff
@@ -2,7 +2,12 @@ import Ember from 'ember';
 
 var get = Ember.get;
 
+const {
+  __bind, __guard, __objectIsAlive
+} = heimdall.registerMonitor('system.store.common', '_bind', '_guard', '_objectIsAlive');
+
```
runspired: should have a consistent format, so expand this

addon/-private/system/store/container-instance-cache.js
```diff
@@ -25,10 +26,13 @@ export default function ContainerInstanceCache(owner) {
   this._cache = new EmptyObject();
 }
 
+const  { __get, instanceFor } = heimdall.registerMonitor('system.store.container-instance-cache', 'get', 'instanceFor');
```
runspired: consistent formatting

tests/dummy/app/routes/index/template.hbs
```diff
@@ -0,0 +1,5 @@
+<ul>
+    <li>
+      {{#link-to 'query'}}Query{{/link-to}}
+    </li>
+</ul>
```
runspired: indent 2 spaces

tests/dummy/app/models/simple.js
```diff
@@ -0,0 +1,6 @@
+import DS from 'ember-data';
+
+export default DS.Model.extend({
+  title: DS.attr('string'),
```
runspired: - don't use 'string' explicitly, or at least be consistent
- destructure DS

runspired:
@igorT code review cleanup is now complete.

bmac:
Overall this pr looks good to me @runspired. 👍

lib/stripped-build-plugins.js
```diff
@@ -36,7 +37,26 @@ module.exports = function(environment) {
...
+      }),
+      // comment out when running non-baseline production benchmarks
+      // WARNING do not ever commit the commented out version!
+      stripHeimdall
+    );
+  }
+
+  if (environment === 'development') {
+    // uncomment when running baseline development benchmarks
+    // plugins.push(stripHeimdall);
```
fivetanley: would prefer an environment variable here. inevitably this comment to not commit it will be skipped or forgotten or not seen in a code review.

lib/stripped-build-plugins.js
```diff
@@ -36,7 +37,26 @@ module.exports = function(environment) {
           'warn',
           'debugSeal'
         ]
-      })
+      }),
+      // comment out when running non-baseline production benchmarks
+      // WARNING do not ever commit the commented out version!
+      stripHeimdall
```
fivetanley: would prefer an environment variable here. inevitably this comment to not commit it will be skipped or not seen in a code review.

runspired: @fivetanley I've found it very hard to replicate a production build without using the production env, if you have suggestions I'm open. Was going to look into a process flag to use as a toggle to keep us safe.

fivetanley: production builds are issued on CI requests and tags auto publish to bower. I wouldn't want heimdall to get in there accidentally; historically build related changes have been difficult to track down and reproduce. something like `process.env.ENABLE_HEIMDALL` would prevent this change from accidentally being checked into source control.

runspired: > Was going to look into a process flag to use as a toggle to keep us safe.

Agreed ;)

runspired:
This should be good now, Travis tests will pass I think appveryor was just a fluke.

cc @fivetanley I added a bit of documentation to CONTRIBUTING.md, and `--instrument` as a flag must be used to turn instrumentation on, it's always off and stripped otherwise.

cc @HeroicEric the counters are now alphabetized :)

fivetanley:
we'll need to double check with @tricknotes and other globals-mode users to ensure this works in gems/bower. it looks great from a build perspective but just trying to be careful since we tend to have build-related regressions whenever we change this stuff.

runspired:
@igor passing :)

bmac:
@runspired do you mind rebasing this pr?

runspired:
@bmac I don't but did you know you can rebase and squash from the merge button now? 💃

runspired:
@bmac rebased :)

fivetanley:
@tricknotes do you have some cycles so we can make sure this still works in the globals version of ember data?

bmac:
@fivetanley I'd like to merge this pr into the master branch. I can cut an alpha gem release to make testing easer and we will have plenty of time to fix any regressions because these changes will have to go through a full beta cycle. Does that sound good to you?

fivetanley:
Sounds good to me. I'll open a tracking issue.

tricknotes:
@fivetanley Sorry to my late :<
Now I confirmed global builds, and it seems to work fine for me 🚀

PR #4514, opened 2016-08-26
Document the options argument in the DS.Transform class

bmac:


locks:
Are the options arbitrary?

bmac:
Yes.

PR #4515, opened 2016-08-26
Document the allowNull property on the boolean transform

bmac:


locks:
When was this option introduced?

bmac:
http://emberjs.com/blog/2016/06/09/ember-data-2-6-released.html
Ember Data 2.6.

addon/-private/transforms/boolean.js
```diff
@@ -22,6 +22,20 @@ const { isNone } = Ember;
   });
   ```
 
+  By default the boolean transform only allows for values of `true` or
+  `false`. You can opt into allowing `null`/`undefined` values for
```
pangratz: Technically `null` is returned when the serialized value `isNone`.

pangratz:
Thank you kindly

PR #4517, opened 2016-08-26
Remove ContainerInstanceCache from the API docs since it is not expos…

bmac:
…ed to consumers

screen shot 2016-08-26 at 2 02 11 pm


pangratz:
Thanks!

PR #4519, opened 2016-08-26
[BUGFIX beta] Prefer includes over contains

bmac:
This pr adds the
[ember-runtime-enumerable-includes-polyfill](https://github.com/rwjblue/ember-runtime-enumerable-includes-polyfill)
to add includes if the user's version of Ember doesn't support it.

cc @rwjblue

### DIFF
--- a/package.json
+++ b/package.json
@@ -36,6 +36,7 @@
     "ember-cli-test-info": "^1.0.0",
     "ember-cli-version-checker": "^1.1.4",
     "ember-inflector": "^1.9.4",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -36,6 +36,7 @@
     "ember-cli-test-info": "^1.0.0",
     "ember-cli-version-checker": "^1.1.4",
     "ember-inflector": "^1.9.4",
+    "ember-runtime-enumerable-includes-polyfill": "1.0.0",
```
rwjblue: We should use `^1.0.0` so we can ship fixes if any bugs are found...

Issue #4513, opened 2016-08-26
Immutability of RecordArray from this.store.findAll() should be characterized with a better error message


ckl1989:
Currently, mutations to a RecordArray returned by a call to the store's findAll() is throwing the following error:

TypeError: internalModel.getRecord coming from record-array.js

This message is really ambiguous to the root cause and took me days of searching to sort it out. I feel a message about the RecordArray's immutability should be shown instead of a TypeError. 

This is also the case with AdapterPopulatedRecordArrays coming from query() calls, mutations should not be possible on the live array. I believe this was fixed for the query case, but not for the findAll() case.
This issue was raised in  #4045

The issue can be seen in the following ember fiddle with a reverseObjects() call:

https://ember-twiddle.com/d64b21293b6d7a2a2886f3b8368b1081?openFiles=controllers.application.js%2C

this is related to #3530 when trying to push objects. 

I hope my understanding of this issue is correct. Please correct me if I am wrong! I really haven't found too much information available online relating to this.

ckl1989:
@bmac Sorry to call directly call out to you. I know you've probably got lots of other things to deal with but when you get a chance, can you provide some feedback on this PR? Thank you!

wecc:
Closing this as #4520 has been merged!

PR #4520, opened 2016-08-27
Add a real error message for attempts to manipulate a recordArray and…

ckl1989:
… prevent it from trying to perform manipulation #4513 

When a findAll() is called on the store, a recordArray is returned. If a user were to try and manipulate this recordArray with something like a pushObject() or reverseObjects(), a TypeError will be thrown for one of the internal models. See #4513.

This error, while warranted, is a little misleading to the actual problem. There isn't much documentation to describe that RecordArrays are not mutable, leading to hours of frustration for beginners to figure out why such a simple task on an apparently ember Array type is not "working".   

AdapterPopulatedRecordArray and FilteredRecordArray all have prevent attempts at manipulation and have descriptive error messages. I think RecordArray should also have it,


bmac:
Overall I'm 👍 for this change as it provides a much better error message then exists today. 

@fivetanley has a similar change in pr https://github.com/emberjs/data/pull/4045, but I don't think that prevents us from landing this pr now.

@pangratz, @fivetanley what are your thoughts?

pangratz:
Basically looks good to me, thanks @ckl1989 🚀 

One minor tiny nitpick: what do you think of moving the test to the existing [`tests/unit/record-array-test.js`](https://github.com/emberjs/data/blob/master/tests/unit/record-array-test.js) and using a "real" `type` (`Person` is already available in the tests), so we can make sure that the thrown error message contains the correct type information (in the current test the type is mocked). I would also tend to try to modify the result of `store.peekAll()` and not use the `DS.RecordArray` directly. What do you think?

ckl1989:
@pablobm @bmac Thank you for your feedback! I've added your change suggestions. They make a lot of sense. Currently I've modeled my test's expected error message after the one in the adapter-populated-record-array. Did you want it more specific? (instead of just showing "subclass of DS.Model", actually show "person").

I'm a junior ember developer and it really means a lot to have this validation after all the uncertain research I had been doing about this issue. Thank you again. Let me know if there's are any other change requests.

pangratz:
Thanks for the update @ckl1989.

Hmm, it seems that the `toString` functionality is somehow not setup as I though it would be, when using the `setupStore` utility. But `<subclass of DS.Model>` showing up instead of the actual `modelName` is also kind of expected [another test](https://github.com/emberjs/data/blob/v2.7.0/tests/unit/model-test.js#L204).

Resolving this (if desired at all), is out of scope for this PR, so I would say we go ahead and merge. @bmac @fivetanley thoughts?

ckl1989:
bump

fivetanley:
@ckl1989 Can you update the message to tell them to use `toArray` instead?

ckl1989:
@bmac @pangratz @fivetanley 

I've updated the error message and locally built and ran the tests on my machine. However AppVeyor is failing on the canary tests but I'm not really understanding the logs. Can you advise? Thanks!

bmac:
Thanks @ckl1989.

PR #4524, opened 2016-08-30
Update docs for shouldReload* methods

blimmer:
It was very unclear to me why the server was being hit when returning `false` from `shouldReloadRecord`. Upon digging through the tests, I finally realized that background reload was what was actually hitting the server.

### DIFF
--- a/addon/adapter.js
+++ b/addon/adapter.js
@@ -508,6 +508,10 @@ export default Ember.Object.extend({
 
     By default this hook returns `false`, as most UIs should not block user
     interactions while waiting on data update.
+    
+    Note that, with default settings, `shouldBackgroundReloadRecord` will always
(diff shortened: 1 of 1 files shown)


iezer:
👍 Thanks for clarifying. I ran into this issue recently as well. There's another issue the example. The format for diffing dates using `moment` has changed. 

```js
// example in doc throws error
moment().diff(ticketSnapshot.attr('lastAccessedAt')).minutes();

// new format
moment().diff(ticketSnapshot.attr('lastAccessedAt'), 'minutes');
```

Do you think you can update that as well?

blimmer:
@pangratz I added that and fixed the whitespace line (I was cheating and editing on GitHub).

@iezer I also updated the moment bit

pangratz:
Awesome, thanks @blimmer!

PR #4528, opened 2016-09-08
try busting the cache by keeping phantomjs info in npm

fivetanley:


fivetanley:
real fix was the node version update

Issue #4527, opened 2016-09-02
isUpdating remains true if backgroundReloadAll is false


wongpeiyi:
See http://emberjs.jsbin.com/xicosasazo/1/edit?html,js,output – click "Click here", then "Go back" –> `isUpdating` is true even though `backgroundReloadAll` returns false and no fetch was triggered

Should https://github.com/emberjs/data/blob/03d098d5cdb2fba2249f6c750268240c6400518d/addon/-private/system/store.js#L1378 reset `isUpdating` to `false` before returning?

bmac:
Closed by https://github.com/emberjs/data/pull/4529

PR #4529, opened 2016-09-09
isUpdating should be true only if a reload happens

wongpeiyi:
See https://github.com/emberjs/data/issues/4527 for jsbin

I think this was fixed incorrectly in https://github.com/emberjs/data/pull/4386 – `isUpdating` would be true on the returned cached array even if `backgroundReload` or `shouldBackgroundReloadAll` is false


bmac:
Thanks @wongpeiyi.

PR #4532, opened 2016-09-13
Infer missing a type field get the type from the relationship which is in scope

craigmcnamara:
I keep getting the error

```
Assertion Failed: You need to pass a model name to the store's modelFor method
```

In my debugger I was seeing a record without a type field which is what the API I'm calling returns, and there is a relationship which has the appropriate type information, but it's not being used to determine what model to load. This seems wrong to look to the record and not the application defined association, but this patch preserves existing behavior yet fixes the issue I'm hitting using an API that doesn't assert a type.

### DIFF
--- a/addon/-private/system/store.js
+++ b/addon/-private/system/store.js
@@ -2395,7 +2395,7 @@ function deserializeRecordId(store, key, relationship, id) {
   assert(`A ${relationship.parentType} record was pushed into the store with the value of ${key} being ${Ember.inspect(id)}, but ${key} is a belongsTo relationship so the value must not be an array. You should probably check your data payload or serializer.`, !Array.isArray(id));
 
   //TODO:Better asserts
-  return store._internalModelForId(id.type, id.id);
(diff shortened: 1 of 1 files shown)


bmac:
@craigmcnamara What serializer are you using? I suspect we'd like for the serializer to fill in the missing type information from the relationship before the JSONAPI document gets to this part of the internalModel.

fivetanley:
A reproduction case would also be super helpful in reviewing this PR. What payload triggers this change?

pangratz:
@craigmcnamara any change you can provide a sample which demonstrates the issue this PR is solving?

As said above this should probably be handled by the serializer so such an information would be super helpful to track this down to the root cause. Thanks!

runspired:
We shouldn't be inferring, but we could do a much better job validating what we are given here.

runspired:
Closing in favor of https://github.com/emberjs/data/pull/5445

PR #4535, opened 2016-09-15
bump ember-cli

fivetanley:

PR #4536, opened 2016-09-16
bump to latest ember-cli, run ember init, cleanup post init

runspired:
This also seems to fix all our tests.  Would supplant #4535 and #4534


fivetanley:
awesome, thanks @runspired!

bmac:
Thanks so much for doing this @runspired!

PR #4537, opened 2016-09-19
[BUGFIX beta] Ember Data should not swallow exceptions from the run loop

### LINKED ISSUE #4533: RESTAdapter silently swallows exceptions
Let's say we using `findAll` to retrieve a set of models in our route

```
   model() {
     return this.store.findAll('my-model');
   }
});
```

We then have a template like:

```
{{#each model as |instance|}}
  {{#each instance.prop key="id" as |p|}}
     <span>{{p.text}}</span>
   {{/each}}
{{/each}}
```

If we get a model instance like:

```
{
  prop: [{id: undefined, text: 'text'}]
}
```

The each helper will throw the error 'You must provide a string key when calling `yieldItem`; you provided undefined'. However, because of 

https://github.com/emberjs/data/blob/v2.8.0/addon/adapters/rest.js#L1014
and then subsequently 
https://github.com/tildeio/rsvp.js/blob/v3.3.1/lib/rsvp/-internal.js#L243

We have the error fail silently. The rest adapter catches the error, tries to reject it, but because the promise is already resolved, it swallows the exception. The rendering might completely fail, but the console will not show any errors, making it quite difficult to trace the error down.

bmac:
Closes #4533

### DIFF
--- a/addon/adapters/rest.js
+++ b/addon/adapters/rest.js
@@ -1007,25 +1007,17 @@ var RESTAdapter = Adapter.extend(BuildURLMixin, {
       var hash = adapter.ajaxOptions(url, type, options);
 
       hash.success = function(payload, textStatus, jqXHR) {
-        try {
-          var response = ajaxSuccess(adapter, jqXHR, payload, requestData);
-          Ember.run.join(null, resolve, response);
-        } catch (error) {
-          Ember.run.join(null, reject, error);
-        }
+        var response = ajaxSuccess(adapter, jqXHR, payload, requestData);
+        Ember.run.join(null, resolve, response);
       };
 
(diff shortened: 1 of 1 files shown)


bmac:
ping @stefanpenner. @igorT requests your eyes on this pr.

bmac:
I'm going to make this a `[BUGFIX release]`.

fivetanley:
@bmac was this introduced in the last release cycle?

bmac:
Yes it was introduced by https://github.com/emberjs/data/pull/4489.

fivetanley:
ah cool, just wanted to verify. happy to merge once we move it to `[bugfix release]`

PR #4545, opened 2016-09-22
[BUGFIX beta] Make `Model#data` a plain getter.

rwjblue:
In Ember 2.9, using an `Ember.computed.alias` or `Ember.computed.readOnly` automatically adds mandatory setter's getter/setter to the properties being watched, and defers reading the current value to the objects own meta.

This avoids creating an watcher for `_internalModel` and therefore prevents an error that ocurrs when we attempt to read from `meta` after the object has been destroyed.


bmac:
Thanks @rwjblue!

PR #4547, opened 2016-09-22
[BUGFIX beta] Correctly serialize invalid dates

TylerHansen:
Invalid Date objects - such as `new Date("foo")` - are Date objects that
throw a `RangeError` when calling `date.toISOString()`. These are now
serialized like all other garbage-in inputs.


bmac:
Looks good to me. 👍

bmac:
Thank you @TylerHansen.

PR #4553, opened 2016-09-26
[DOCS] Fix store.unloadAll modelName param type

knownasilya:
Remove extra character


knownasilya:
YUI Docs print warnings, maybe those could be made to return a non-zero response in travis.

bmac:
Thanks @knownasilya.

PR #4558, opened 2016-09-28
[BUGFIX beta] Inverse null relationships should throw if model doesn't exist

canufeel:
`2.6.0` [Introduced](https://github.com/asakusuma/data/blob/ced181ff68f87eecd149c6d15b1f8b6887f17142/addon/-private/system/relationships/state/create.js#L16) a bug such that for `{ inverse: null }` relationships the relationship target is not checked at model creation time. However without `{ inverse: null }` we would get an error when trying to create a model with relationship to another model that doesn't exist. This PR tries to bring in consistent behaviour for such cases - perform a check for model type during model creation and throw an error if the model is not present in the store.


fivetanley:
I believe the laziness of the relationship lookup was done for perf. @asakusuma @krisselden can you review + provide some context?

canufeel:
The simplest example of inconsistency would be unit test for model:

my-model.js
```
export default Model.extend({
  rel: belongsTo('model-that-doesnt-exist')
});
```

my-model-test.js
```
moduleForModel('my-model', 'Unit | Model | my model', {
  unit: true
});

test('it exists', function(assert) {
  var model = this.subject();
  assert.ok(!!model);
});
```

The test above would fail for 2 reasons:
1. We haven't defined `model-that-doesnt-exist`
2. We haven't specified it in a `needs` for the unit test.

However changing the model code to:
```
export default Model.extend({
  rel: belongsTo('model-that-doesnt-exist', {inverse: null})
});
```
would result in a passing test for versions above `2.6.0`. The test would still fail in `ember-data` versions less then `2.6.0`.

In my opinion this should be either fixed or documented properly from testing point of view.

runspired:
I feel as though this code path should just be wrapped in a strippable dev mode only block and always run even when it should be lazy.

addon/-private/system/relationships/state/create.js
```diff
@@ -15,6 +15,8 @@ function createRelationshipFor(record, relationshipMeta, store) {
   let inverse = null;
   if (shouldFindInverse(relationshipMeta)) {
     inverse = record.type.inverseFor(relationshipMeta.key, store);
+  } else {
+    record.type.typeForRelationship(relationshipMeta.key, store);
```
runspired: I think the correct code here to ensure we throw when the inverse is an undefined model is:

```
import { runInDebug } from 'ember-data/-private/system/debug';

...

runInDebug(() => {
  // for perf reasons, this should be lazy, but for explicit `inverse: null;` situations
  // we need to run the lookup to throw the assert if it were improperly defined (e.g. the related model is not defined).
  record.type.typeForRelationship(relationshipMeta.key, store);
});
```

runspired: alternatively, a strippable check should be added at this location: https://github.com/emberjs/data/blob/master/addon/-private/system/relationships/ext.js#L242

runspired:
The laziness of the lookup was indeed done for perf, however I'm fairly certain we are still defeating that. I have a few commits coming that include test coverage to ensure the laziness is not defeated, and will fix it if that turns out to be the case.  Would love to work on resolving this PR along with that.

PR #4560, opened 2016-09-29
[BUGFIX beta] Ensure record array length is reset during willDestroy.

rwjblue:
In Ember 2.10+ objects that are in the middle of being destroyed no longer notify property changes on itself (or other destroying objects) to avoid wasted work.  This means that `set(this, 'content', undefined)` isn't enough to ensure that `Enumerable` mixin will `notifyPropertyChange(this, 'length')` (since its observers are not fired).

This change also sets `length` during cleanup so it properly resets to `0`.

This fixes the build against latest ember#canary.


igorT:
Seems scary, but hard to observe the problem, not sure how that test came about

PR #4562, opened 2016-09-29
[PERF] make internal model lazier

runspired:
Branches off of heimdall-instrumentation, so that should be merged fist ;)

### DIFF
--- a/addon/-private/system/model/internal-model.js
+++ b/addon/-private/system/model/internal-model.js
@@ -16,11 +16,33 @@ import {
   HasManyReference
 } from "ember-data/-private/system/references";
 
-var Promise = Ember.RSVP.Promise;
-var get = Ember.get;
-var set = Ember.set;
-var copy = Ember.copy;
-var assign = Ember.assign || Ember.merge;
+const {
+  get,
+  set,
+  copy,
+  Error: EmberError,
+  inspect,
+  isEmpty,
+  isEqual,
+  run: emberRun,
+  setOwner,
+  RSVP,
+  RSVP: { Promise }
+} = Ember;
+
+const assign = Ember.assign || Ember.merge;
+
+/*
+  The TransitionChainMap caches the `state.enters`, `state.setups`, and final state reached
+  when transitioning from one state to another, so that future transitions can replay the
+  transition without needing to walk the state tree, collect these hook calls and determine
+   the state to transition into.
+
+   A future optimization would be to build a single chained method out of the collected enters
+   and setups. It may also be faster to do a two level cache (from: { to }) instead of caching based
+   on a key that adds the two together.
+ */
+const TransitionChainMap = new EmptyObject();
 
 var _extractPivotNameCache = new EmptyObject();
 var _splitOnDotCache = new EmptyObject();
@@ -35,12 +57,6 @@ function extractPivotName(name) {
   return _extractPivotNameCache[name] || (
     _extractPivotNameCache[name] = splitOnDot(name)[0]
   );
-}
-
-function retrieveFromCurrentState(key) {
-  return function() {
-    return get(this.currentState, key);
-  };
 }
 
 // this (and all heimdall instrumentation) will be stripped by a babel transform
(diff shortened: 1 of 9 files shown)


fivetanley:
so excited :pray:

we'll have to make sure we don't break @bookingsync's and others' addons.

runspired:
@fivetanley this provides identical private API's to what was present before, they are just lazy, with one notable exception. I've removed the use of `get` and `set` for `currentState`. It was being used inconsistently, half the places used dot notation, half used get/set. Looking into the code paths it appears there isn't a strong case for it needing to be `get/set` based, and obvs with it being inconsistent before if it did "need to be" it would have been broken.

fivetanley:
yeah i am not really concerned about the api changing here, just a last step before merging to run it against addons that use `internalModel` heavily.

runspired:
Also note, the tests on this pass locally but I need to rebase off master for them to pass here, If we could merge heimdall today I've got a follow up PR after this one that's also fun :)

runspired:
@igor rebase complete

addon/-private/system/model/internal-model.js
```diff
@@ -851,26 +946,26 @@ InternalModel.prototype = {
...
       }
     }
 
     return changedKeys;
-  },
+  }
 
   toString() {
     if (this.record) {
       return this.record.toString();
```
stefanpenner: this branch seems un-needed

addon/-private/system/model/states.js
```diff
@@ -584,6 +584,7 @@ var RootState = {
       },
 
       didCommit(internalModel) {
+        // TODO @runspired lastDirtyType doesn't seem to be a real thing
         internalModel.send('invokeLifecycleCallbacks', get(internalModel, 'lastDirtyType'));
```
stefanpenner: ^ todo + get, if its not used then this line may be a candidate for removal

addon/-private/system/model/internal-model.js
```diff
@@ -574,37 +570,49 @@ export default class InternalModel {
...
       state = state.parentState;
-    } while (!state.hasOwnProperty(pivotName));
+    } while (!state[pivotName]);
+
+    let setups;
+    let enters;
+    let i;
+    let l;
+
+    if (TransitionChainMap[transitionMapId]) {
```
stefanpenner: maybe `let map = TransitionChainMap[transitionMapId]` then reference map everywhere

addon/-private/system/model/states.js
```diff
@@ -241,7 +244,7 @@ var DirtyState = {
 
     //TODO(Igor) reloading now triggers a
     //loadingData event, though it seems fine?
-    loadingData: Ember.K,
+    loadingData: K,
```
stefanpenner: lets just do `loadingData() { }` rather then the single shared function of `K`

addon/-private/system/model/states.js
```diff
@@ -4,7 +4,10 @@
 import Ember from 'ember';
 import { assert } from "ember-data/-private/debug";
 
-var get = Ember.get;
+const {
+  K
```
stefanpenner: lets not use `K` regular concise functions are likely fine.

runspired:
Eyeballing the production build benchmarks (no filtering of outliers but I am looking at a box plot), we moved from roughly 35ms spent in `store.push` for 238 records of a single type to a bit less than 28ms.

tchak:
LGTM 👍

addon/-private/system/model/internal-model.js
```diff
@@ -16,11 +16,22 @@ import {
...
+  isEmpty,
+  isEqual,
+  run: emberRun,
+  RSVP,
+  RSVP: { Promise }
+} = Ember;
+
+const assign = Ember.assign || Ember.merge;
+const setOwner = Ember.setOwner;
+const TransitionChainMap = new EmptyObject();
```
igorT: Can you put a comment here or in the method as to how the transitionchain map works

addon/-private/system/model/model.js
```diff
@@ -305,6 +308,7 @@ var Model = Ember.Object.extend(Ember.Evented, {
     @private
     @type {Object}
   */
+  currentState: null,
```
stefanpenner: should we define this on the prototype

addon/-private/system/model/states.js
```diff
@@ -430,7 +433,7 @@ createdState.uncommitted.pushedData = function(internalModel) {
   internalModel.triggerLater('didLoad');
 };
 
-createdState.uncommitted.propertyWasReset = Ember.K;
+createdState.uncommitted.propertyWasReset = K;
```
stefanpenner: `= function() { } `

runspired:
@igorT Travis is passing after making the changes from @stefanpenner's last review.

Issue #4561, opened 2016-09-29
RestAdapter is not detecting XHR interrupted  by refresh as an abort


bryanhickerson:
For several months, we have been hitting `Ember Data 0 Payload Empty Content-Type` errors on our production server (Related: http://stackoverflow.com/questions/38983556/ember-data-0-payload-empty-content-type). I can reproduce the error by using Chrome Dev Tools Network throttling and hitting refresh in the middle of an Ember Data related XHR. 

Since this is an abort, I was curious why we were getting this error instead of the abort error seen here: https://github.com/emberjs/data/blob/16cf35db2aa0ddbe28c043570ef6d725139b0c47/addon/adapters/rest.js#L1433. So I logged the `textStatus` while triggering the error and found out that the textStatus was `error` instead of `abort`. 

```
if (responseData.textStatus === 'abort' || jqXHR.status === 0) {
   error = new AbortError();
}
```

Seemingly, if the adapter did the above, this would flag the error correctly. I'm open to writing the PR if this seems like the right approach.

bmac:
I believe this is fixed by your pr in https://github.com/emberjs/data/pull/4567. I've just published the v2.9.0-beta.4 release which includes that fix. Feel free to reopen if its still an issue.

PR #4567, opened 2016-10-03
[bugfix beta] Treat status code 0 as an abort

bryanhickerson:
This PR resolves issue #4561. An abort by a user's browser has `textStatus === error`, which results in  a confusing Ember Data Payload error. This PR checks for status code 0, and returns an AbortError if that is the case.


fivetanley:
@bryanhickerson this looks great to me!

Can you run:

`git commit --amend -m '[bugfix beta] Treat status code 0 as an abort'` and re-push, so we can pull it in for the next release?

bryanhickerson:
Done. Thanks for the quick response!

PR #4574, opened 2016-10-12
Move heimdalljs to dependencies

hjdivad:
heimdalljs is actually a dev dependency, but `npm install` will fail to install
the nested (non-dev) dependencies, eg for broccoli-funnel, due to
https://github.com/npm/npm/issues/13974


hjdivad:
cc @bmac @stefanpenner

hjdivad:
npm bug https://github.com/npm/npm/issues/13974

stefanpenner:
thanks dude!

cyril-sf:
@hjdivad which approach did you take to figure out the package you needed to update?

hjdivad:
@cyril-sf `npm install` from clean, then `npm ls` and npm will report that some dependencies are not met.  `npm install` again and now those dependencies are met.

Those dependencies were the ones causing problems for us.

PR #4580, opened 2016-10-17
Fixing the import for inflector so that it imports the right function

msranade:
We came across this scenario when `define.amd` is true, `singularize` and `pluralize` functions are not available and it breaks the app because of this. The check-in fixes that issue. 

Here is how compiled code looks before the fix:

```
define("ember-inflector/index", ["exports", "ember", "ember-inflector/lib/system", "ember-inflector/lib/ext/string"], function (exports, _ember, _emberInflectorLibSystem, _emberInflectorLibExtString) {
  /* global define, module */

  "use strict";

  _emberInflectorLibSystem.Inflector.defaultRules = _emberInflectorLibSystem.defaultRules;
  _ember["default"].Inflector = _emberInflectorLibSystem.Inflector;

  _ember["default"].String.pluralize = _emberInflectorLibSystem.pluralize;
  _ember["default"].String.singularize = _emberInflectorLibSystem.singularize;exports["default"] = _emberInflectorLibSystem.Inflector;
  exports.pluralize = _emberInflectorLibSystem.pluralize;
  exports.singularize = _emberInflectorLibSystem.singularize;
  exports.defaultRules = _emberInflectorLibSystem.defaultRules;

  if (typeof define !== 'undefined' && define.amd) {
    define('ember-inflector', ['exports'], function (__exports__) {
      __exports__['default'] = _emberInflectorLibSystem.Inflector;
      return _emberInflectorLibSystem.Inflector;
    });
  } else if (typeof module !== 'undefined' && module['exports']) {
    module['exports'] = _emberInflectorLibSystem.Inflector;
  }
});
```

### DIFF
--- a/addon/-private/system/relationship-meta.js
+++ b/addon/-private/system/relationship-meta.js
@@ -1,4 +1,4 @@
-import {singularize} from 'ember-inflector';
+import {singularize} from 'ember-inflector/index';
 import normalizeModelName from 'ember-data/-private/system/normalize-model-name';
 
 export function typeForRelationshipMeta(meta) {


msranade:
Does this pull request need more information for it to be considered?  Just wanted to get a status on this.

bmac:
Hi @msranade. A little more info on how to reproduce the scenario where the `singularize` and `pluralize` functions are not available would be great.

msranade:
Hi @bmac, Thanks for the response. We have a rendering mode for Ember applications which works similar to [Facebook's Bigpipe](https://www.facebook.com/notes/facebook-engineering/bigpipe-pipelining-web-pages-for-high-performance/389414033919/) mode. In this mode, we load the Ember app in Node.js environment and make API calls (server to server) while the app is booting up in the browser. As we get responses from the API, we stream them back to the browser so that response is readily available as soon as the app is booted. In this scenario, in Node.js environment, we have a value of `define.amd` as true. In this scenario, app breaks because it complains that `singularize` and `pluralize` functions are not available. @kratiahuja and I faced this issue together. @kratiahuja please add more details if I am missing something.

rwjblue:
Just like in node, `foo-bar/index` is a fallback for `foo-bar`.  You can see that [here in loader.js](https://github.com/ember-cli/loader.js/blob/master/lib/loader/loader.js#L224).

I do not think we should change this to explicitly use `/index` (as that breaks the abstraction that `ember-inflector` can move where these come from).

rwjblue:
@msranade - Are you using a different loader? Does that loader support the node style fallback behavior?

kratiahuja:
@rwjblue we do not use a different loader but loader.js in the sandbox, however we have `define.amd = true` in our current sandbox (which is there for historical reasons that simple-dom used to publish a version with this check). Due to this being there and inflector generating a different output when this flag is set to true. I think inflector itself will break when any other loader is used or when loader.js starts supporting `define.amd=true`. 

This is what inflector is doing: https://github.com/stefanpenner/ember-inflector/blob/master/addon/index.js#L27 when define.amd is set to true. So you don't have singularize/pluralize etc.

rwjblue:
Awesome! That is the bug we should be fixing! 😸

rwjblue:
https://github.com/stefanpenner/ember-inflector/pull/112

rwjblue:
The upstream PR has landed, just needs a release now...

PR #4581, opened 2016-10-17
[DOC] Update descriptions of findRecord() and findAll()

MarkMT:
This commit adds documentation on the use of the `include` query parameter in `DS.Store` methods `findRecord()` and `findAll()`. This feature allows records of models related to those requested to be retrieved from JSON API compliant server in a single request. This capability was added in https://github.com/emberjs/data/pull/3976 but is not currently documented.

### DIFF
--- a/addon/-private/system/store.js
+++ b/addon/-private/system/store.js
@@ -619,8 +619,7 @@ Store = Service.extend({
     model, when we retrieve a specific post we can have the server also return that post's
     comments in the same request:
 
-    ```
-    // app/routes/post.js
+    ```app/routes/post.js
     import Ember from 'ember';
 
     export default Ember.Route.extend({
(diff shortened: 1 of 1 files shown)


addon/-private/system/store.js
```diff
@@ -605,6 +605,51 @@ Store = Service.extend({
...
+    ['include' query parameter](http://jsonapi.org/format/#fetching-includes),
+    you can use `findRecord()` to automatically retrieve additional records related to
+    the one you request by supplying an `include` parameter in the `options` object.
+
+    For example, given a `post` model that has a `hasMany` relationship with a `comment`
+    model, when we retrieve a specific post we can have the server also return that post's
+    comments in the same request:
+
+    ```
+    // app/routes/post.js
```
bmac: Instead of writing the filename in a comment you can write it after the 3 back ticks and it will get rendered as a header for the code block.

bmac:
Thanks @MarkMT. Overall this is looking good 👍.

addon/-private/system/store.js
MarkMT: Thanks, I thought that was the case, but initially I was using https://github.com/emberjs/api to render the edits locally and it seems for some reason that that doesn't actually support that behavior, so I put the filenames in the comment instead. I've switched to https://github.com/emberjs/website for local rendering and it's now working as advertised.

bmac:
Thanks @MarkMT.

PR #4582, opened 2016-10-17
tidy up integration/filter-test

stefanpenner:
* remove unneeded reopens
* remove assert.wait, instead use the promise chaining
* consistent quotes
* remove unneeded promises
* some ES6


bmac:
👍

PR #4583, opened 2016-10-17
[PERF] optimise notifications when has-many array is changed

BryanCrotaz:
diff the old and new array and optimise the notifications accordingly.

Particularly important when polling server for changes with a repeating `model.reload()` or `relationship.update()`


BryanCrotaz:
tested in our app where we refresh a large relationship every second and have 10 or so filters chained off this. CPU usage has gone from 100% of one core to a trickle.

stefanpenner:
Cool, i was hoping someone would work on this!

BryanCrotaz:
Whaddayathink?

.gitignore
```diff
@@ -24,3 +24,4 @@ node_modules/
 bower_components/
 .metadata_never_index
 npm-debug.log
+.vscodeignore
```
stefanpenner: unsure if this should be added here

tests/integration/records/relationship-changes-test.js
```diff
@@ -0,0 +1,358 @@
+import setupStore from 'dummy/tests/helpers/store';
+import Ember from 'ember';
+
+import {module, test} from 'qunit';
+
+import DS from 'ember-data';
+
+var env, store;
+const attr = DS.attr;
```
stefanpenner: `const { attr, hasMany}  = DS`

tests/integration/records/relationship-changes-test.js
```diff
@@ -0,0 +1,358 @@
+import setupStore from 'dummy/tests/helpers/store';
+import Ember from 'ember';
+
+import {module, test} from 'qunit';
+
+import DS from 'ember-data';
+
+var env, store;
```
stefanpenner: let

tests/integration/records/relationship-changes-test.js
```diff
@@ -0,0 +1,358 @@
...
+const attr = DS.attr;
+const hasMany = DS.hasMany;
+const run = Ember.run;
```
stefanpenner: `const { run } = Ember`

addon/-private/system/many-array.js
```diff
@@ -87,15 +87,39 @@ export default Ember.Object.extend(Ember.MutableArray, Ember.Evented, {
...
+      // we found a change
+      var added = newLength - firstChangeIndex;
+      var removed = oldLength - firstChangeIndex;
+      this.arrayContentWillChange(firstChangeIndex, added, removed);
+      // It’s possible the parent side of the relationship may have been unloaded by this point
+      if (_objectIsAlive(this)) {
+        this.set('length', toSet.length);
+      }
+      this.currentState = toSet;
+      this.arrayContentDidChange(firstChangeIndex, added, removed);
```
stefanpenner: as we are notifiying array change events, tests should likely test that these are accurate.

addon/-private/system/many-array.js
```diff
@@ -87,15 +87,39 @@ export default Ember.Object.extend(Ember.MutableArray, Ember.Evented, {
...
-    if (_objectIsAlive(this)) {
-      this.set('length', toSet.length);
+    var newLength = toSet.length;
+
+    var shortestLength = Math.min(oldLength, newLength);
+
+    var firstChangeIndex = -1; // -1 signifies no changes
+    // find the first change
+    var currentArray = this.currentState;
+    for (var i=0; i<shortestLength; i++) {
```
stefanpenner: we should likely skip this if within `init`, but that isn't needed i think until https://github.com/emberjs/data/pull/4600 is merged, so either this one or that one need to take this into account.

addon/-private/system/many-array.js
```diff
@@ -87,15 +87,39 @@ export default Ember.Object.extend(Ember.MutableArray, Ember.Evented, {
     );
     toSet = toSet.concat(newRecords);
     var oldLength = this.length;
-    this.arrayContentWillChange(0, this.length, toSet.length);
-    // It’s possible the parent side of the relationship may have been unloaded by this point
-    if (_objectIsAlive(this)) {
```
stefanpenner: if the object is dead, we should exit at the top of the function

addon/-private/system/many-array.js
```diff
@@ -87,15 +87,39 @@ export default Ember.Object.extend(Ember.MutableArray, Ember.Evented, {
...
+    }
+    if (firstChangeIndex === -1) {
+      // no change found in the matching part of the arrays
+      if (newLength !== oldLength) {
+        firstChangeIndex = shortestLength;
+      }
+    }
+    if (firstChangeIndex !== -1) {
+      // we found a change
+      var added = newLength - firstChangeIndex;
```
stefanpenner: what happens if we have non-contiguous changes, e.g. the collection was replaced and some members of the previous set and the current set are the same.

```js
[model#1, model#2, model#3, model#4, model#10]
```

turns into:

```js
[model#1, model#5, model#10]
```

addon/-private/system/many-array.js
```diff
@@ -87,15 +87,39 @@ export default Ember.Object.extend(Ember.MutableArray, Ember.Evented, {
...
+      var removed = oldLength - firstChangeIndex;
+      this.arrayContentWillChange(firstChangeIndex, added, removed);
+      // It’s possible the parent side of the relationship may have been unloaded by this point
+      if (_objectIsAlive(this)) {
```
stefanpenner: we should have already exited above if the object is dead.

stefanpenner:
cc @hjdivad your thoughts on diffing/batching would be handy here.

addon/-private/system/many-array.js
BryanCrotaz: Then we'd put out 
{
 firstChange: 1
  removed: 4
  added: 2
}

stefanpenner: the following example may be clearer as to what the problem may be.

```
[model#1, model#2, model#3, model#4, model#10]
```

becomes:

```
[model#1, model#10, model#3, model#11, model#10]
```

BryanCrotaz:
@stefanpenner Implemented a better change finder, but now the internals of `arrayContentWillChange` are failing and I can't see how - seems like the target of the call is the array, not the observer. Have a feeling I'll cause more damage than I fix if I try and work on this one.

addon/-private/system/many-array.js
BryanCrotaz: Done

addon/-private/system/many-array.js
BryanCrotaz: not sure what you mean

addon/-private/system/many-array.js
BryanCrotaz: optimised in latest commit

addon/-private/system/many-array.js
BryanCrotaz: done

addon/-private/system/many-array.js
BryanCrotaz: done

BryanCrotaz: but found a bug in array change event mechanism

runspired:
@BryanCrotaz I'd love to land this, think you'd have a little time to pair / cleanup either this Sunday or the week after the Thanksgiving break?

BryanCrotaz:
@runspired as I said in my comment above I think I need some help - the latest problem is unrelated to what I've been doing and I don't have any knowledge

BryanCrotaz:
@runspired sorry just reread your comment and realised what you're suggesting yes Sunday would be fine. I'm on GMT so morning your time?

runspired:
@BryanCrotaz morning my time is good :)

BryanCrotaz:
Invite me to a hangout when you're ready

runspired:
I ended up working on this problem last night by accident :D

BryanCrotaz:
@runspired :) Did you get anywhere? Still happy to pair...

BryanCrotaz:
Odd - many-array `flushCanonical` is not being called any more during `store.push` so none of my optimisations are kicking in.

What's the new pattern?

addon/-private/system/many-array.js
```diff
@@ -141,31 +148,64 @@ export default Ember.Object.extend(Ember.MutableArray, Ember.Evented, {
...
+      for (let i=1; i<shortestLength; i++) {
+        // compare each item in the array
+        if (currentArray[oldLength-i] !== toSet[newLength-i]) {
+          unchangedEndBlockLength = i-1;
+          break;
+        }
+      }
+      const added = newLength - unchangedEndBlockLength - firstChangeIndex;
+      const removed = oldLength - unchangedEndBlockLength - firstChangeIndex;
+      this.arrayContentWillChange(firstChangeIndex, removed, added);
```
stefanpenner: if nothing changed, should be bail out here entirely? Seems like we don't want to emit any events in that case.

addon/-private/system/many-array.js
```diff
@@ -141,31 +148,64 @@ export default Ember.Object.extend(Ember.MutableArray, Ember.Evented, {
...
       // been 'acknowleged' to be in the relationship via a store.push)
       (internalModel) => internalModel.isNew() && toSet.indexOf(internalModel) === -1
     );
     toSet = toSet.concat(newRecords);
-    let oldLength = this.length;
-    this.arrayContentWillChange(0, this.length, toSet.length);
-    // It’s possible the parent side of the relationship may have been unloaded by this point
-    if (_objectIsAlive(this)) {
-      this.set('length', toSet.length);
+    const oldLength = this.length;
```
stefanpenner: what follows is relatively complicated, could it be extracted to its own  "pure"  function so it can be thoroughly unit tested?

stefanpenner:
- [ ] address feedback
- [ ] tests green
- [ ] rebase

BryanCrotaz:
Oh lovely. When I don't fire the events (if no changes), 180 tests in the adapters fail!

BryanCrotaz:
@stefanpenner rest of tests are failing because of a change on master.

relationship.push(data) used to call manyArray.flushCanonical but doesn't any more. I'm not sure why this would be the case. What's the correct behaviour here?

addon/-private/system/many-array.js
BryanCrotaz: done

addon/-private/system/many-array.js
BryanCrotaz: done

BryanCrotaz:
Very strange.

I have a test that pushes a `siblings` hasMany into a `person` using `store.push`.

This observer fires ok
```    
    person.hasMany('siblings').hasManyRelationship.getManyArray().addObserver('[]', function() {
      observerCount++;
    });
```
This observer does not
```    
    person.addObserver('siblings.[]', function() {
      observerCount++;
    });
```

sly7-7:
@BryanCrotaz Maybe because `siblings.[]` is not really the same as the manyarray ?

BryanCrotaz:
@sly7-7 This test passed back in October. If I'm observing `siblings.[]` on a `person` I'd expect that observer to trigger when the relationship is changed via a `store.push`.

sly7-7:
@BryanCrotaz I fully agree with that. Are the tests failing because of your work ? Or do they not pass with the actual code ?

BryanCrotaz:
I think they're not passing because of changes on master

stefanpenner:
@BryanCrotaz I played with this approach in your example app, it very much improved performance of the pathological cases. I believe it is very much worth exploring further.

I am spending some time with @igorT tomorrow, we will be talking about this and how to move it forward.

I have left some feedback, that being addressed will help move this forward.

jsconfig.json
```diff
@@ -0,0 +1 @@
+{"compilerOptions":{"target":"es6","experimentalDecorators":true},"exclude":["node_modules","bower_components","tmp","vendor",".git","dist"]}
```
BryanCrotaz: ok

tests/unit/diff-array-test.js
```diff
@@ -0,0 +1,486 @@
...
+const g = "ggg";
+const h = "hhh";
+const w = "www";
+const x = "xxx";
+const y = "yyy";
+const z = "zzz";
+
+test('diff array returns no change given two empty arrays', function(assert) {
+  const result = diffArray([], []);
+  assert.equal(result.firstChangeIndex, null);
```
BryanCrotaz: I did that in my first version but thought null was more obviously "none". -1 dates from not having nullable int, thought we could be more modern.

tests/unit/diff-array-test.js
```diff
@@ -0,0 +1,486 @@
+import {module, test} from 'qunit';
+
+import diffArray from 'ember-data/-private/system/diff-array';
+
+module('unit/diff-array Diff Array tests', {
```
BryanCrotaz: can you see any cases I've missed?

addon/-private/system/many-array.js
```diff
@@ -141,31 +150,36 @@ export default Ember.Object.extend(Ember.MutableArray, Ember.Evented, {
...
+
+    if (diff.firstChangeIndex !== null) { // it's null if no change found
+      // we found a change
+      this.arrayContentWillChange(diff.firstChangeIndex || 0, diff.removedCount, diff.addedCount);
+      set(this, 'length', toSet.length);
+      this.currentState = toSet;
+      this.arrayContentDidChange(diff.firstChangeIndex || 0, diff.removedCount, diff.addedCount);
       this.relationship.notifyHasManyChanged();
     }
+    this.record.updateRecordArrays();
```
BryanCrotaz: I thought that's where the canonical was updated. I'll go and check again

addon/-private/system/many-array.js
```diff
@@ -141,31 +150,36 @@ export default Ember.Object.extend(Ember.MutableArray, Ember.Evented, {
...
-    this.arrayContentDidChange(0, oldLength, this.length);
 
-    if (isInitialized) {
-      //TODO Figure out to notify only on additions and maybe only if unloaded
+    // diff to find changes
+    const diff = diffArray(this.currentState, toSet);
+
+    if (diff.firstChangeIndex !== null) { // it's null if no change found
+      // we found a change
+      this.arrayContentWillChange(diff.firstChangeIndex || 0, diff.removedCount, diff.addedCount);
```
BryanCrotaz: Ah - that's left over from some testing

addon/-private/system/many-array.js
```diff
@@ -52,9 +54,16 @@ const { get, set } = Ember;
...
+
+  canonicalState: null,
+  currentState: null,
+
+  length: 0,
+
   init() {
     this._super(...arguments);
-
+    this.currentState = Ember.A([]);
```
BryanCrotaz: because we need array observer and app might not have [] mapped to Ember.A?

BryanCrotaz:
We have lists with 100 entries where we're filtering on a model attribute. We now only check new array entries and with these changes we've see 50-fold changes in cpu usage.

stefanpenner:
> We have lists with 100 entries where we're filtering on a model attribute. We now only check new array entries and with these changes we've see 50-fold changes in cpu usage.

Yup, the downstream side-affects of overchatty arrays can be massive.

addon/-private/system/many-array.js
BryanCrotaz: this is a change on master btw - I've merged it into my branch because my branch started so long ago

BryanCrotaz:
@stefanpenner  I think all your points are dealt with

stefanpenner:
@BryanCrotaz very cool, will review thoroughly in the AM \w @igorT

BryanCrotaz:
Unfortunately... it's so efficient that CPs hung off a `peekAll` don't work any more!

BryanCrotaz:
@stefanpenner I need some help on this one.

I've added a test for the following code not working.

```
test('Calling push with relationship triggers observer of array if the relationship was empty and is added to', function(assert) {
// ...
    person.addObserver('siblings.[]', function() {
      observerCount++;
    });
// ...
```

The observer doesn't trigger. However the array observer on the many array in other tests does trigger.

How does an observer of 'array.[]' get triggered?

EDIT: Odd - this test doesn't pass on master either. Can't see anything wrong with it though. Thoughts?

stefanpenner:
@BryanCrotaz let me take a look

stefanpenner:
@BryanCrotaz it appears you CP is never accessed, so the observer has nothing to watch.

The following addresses the issue, but the test fails because relationships do silly things still and we mutate the array twice not once (we will have to address in the future)

```diff
commit 7b8ad1a47bd2ae8e999b95e26b15a4cecffa378f
Author: Stefan Penner <stefan.penner@gmail.com>
Date:   Wed Mar 8 20:12:51 2017 -0800

    ensure the CP is accessed at-least once, so that the observer has something to observe

diff --git a/tests/integration/records/relationship-changes-test.js b/tests/integration/records/relationship-changes-test.js
index b2595315..ed99e5a5 100644
--- a/tests/integration/records/relationship-changes-test.js
+++ b/tests/integration/records/relationship-changes-test.js
@@ -276,7 +276,6 @@ test('Calling push with relationship recalculates computed alias property to fir
 
 test('Calling push with relationship triggers observer of array if the relationship was empty and is added to', function(assert) {
   assert.expect(1);
-  let person = null;
   let observerCount = 0;
 
   run(() => {
@@ -297,12 +296,15 @@ test('Calling push with relationship triggers observer of array if the relations
     });
   });
 
-  person = store.peekRecord('person', 'wat');
+  let person = store.peekRecord('person', 'wat');
 
   run(() => {
-    person.addObserver('siblings.[]', function() {
-      observerCount++;
-    });
+    // materialize the CP, so that the later siblings observer enages (observers
+    // dont have side affects, so they cause a cp, in this case a relationship, to
+    // be accessed themsleves).
+    person.get('siblings');
+
+    person.addObserver('siblings.[]', () => observerCount++);
   });
 
   run(() => {
@@ -324,9 +326,7 @@ test('Calling push with relationship triggers observer of array if the relations
     });
   });
 
-  run(() => {
-    assert.equal(observerCount, 1, 'siblings observer should be triggered once');
-  });
+  assert.equal(observerCount, 1, 'siblings observer should be triggered once');
 });
 
 test('Calling push with relationship triggers observers once if the relationship was not empty and was added to', function(assert) {
```

BryanCrotaz:
@stefanpenner  doh! that old chestnut.

I've found another inefficiency that I don't know whether to fix. manyArray fires property change events during destruction. Does it need to? Is that going to break something else?

BryanCrotaz:
I think this PR is complete now.

BryanCrotaz:
The diff is a bit of a mess so I'll resubmit as a new PR with it all flattened.

BryanCrotaz:
replaced by PR #4850

PR #4584, opened 2016-10-18
[WIP] [Feat] implements Schema to separate store logic and model description from the model layer.

runspired:
TODO
- [ ] Refactor areas where internal model is being used as schema
- [ ] Move as much logic off of DS.Model methods as we can, since it's public API we'll need to continue exposing it on that class, just not use it ourselves.
- [ ] Refactor InternalModel down to just a data/state container
- [ ] ? Refactor Model methods to use schema for descriptor iteration (tricky because Map => EmptyObject ) ?
- [ ] ? consider refactoring out relationshipMap, as it's almost certain to be greedy and allocates a bunch of arrays. While this may be better if we often iterate over relationships by related className, I suspect such iteration is rare ?
- [ ] ? InternalModel._loadingPromise ? 
- [ ] ? debugSeal on InternalModel so we can find anywhere we're being bad currently ?

### DIFF
--- a/addon/-private/system/model/internal-model.js
+++ b/addon/-private/system/model/internal-model.js
@@ -1,5 +1,5 @@
 import Ember from 'ember';
-import { assert, runInDebug } from "ember-data/-private/debug";
+import { assert, deprecate, runInDebug } from "ember-data/-private/debug";
 import RootState from "ember-data/-private/system/model/states";
 import Relationships from "ember-data/-private/system/relationships/state/create";
 import Snapshot from "ember-data/-private/system/snapshot";
@@ -110,6 +110,7 @@ export default class InternalModel {
     this.id = id;
     this.store = store;
     this._data = data || new EmptyObject();
+    this._schema = null;
     this.modelName = modelClass.modelName;
     this.dataHasInitialized = false;
     this._loadingPromise = null;
@@ -131,7 +132,19 @@ export default class InternalModel {
     this.__implicitRelationships = null;
   }
 
+  get schema() {
+    if (this._schema === null) {
+      this._schema = this.store._schemaForModelClass(this.modelClass);
+    }
+
+    return this._schema;
+  }
+
   get type() {
+    deprecate(`InternalModel.type has been deprecated in favor of InternalModel.modelClass`, {
+      id: 'ember-data.private.type-to-modelClass'
+    });
+
     return this.modelClass;
   }
 
@@ -284,7 +297,6 @@ export default class InternalModel {
       }
 
       this._record = this.modelClass._create(createOptions);
-
       this._triggerDeferredTriggers();
       heimdall.stop(token);
     }
@@ -345,16 +357,16 @@ export default class InternalModel {
     this.send('unloadRecord');
   }
 
-  eachRelationship(callback, binding) {
-    return this.modelClass.eachRelationship(callback, binding);
+  eachRelationship(callback, context) {
+    this.schema.eachRelationship(callback, context);
   }
 
-  eachAttribute(callback, binding) {
-    return this.modelClass.eachAttribute(callback, binding);
+  eachAttribute(callback, context) {
+    this.schema.eachAttribute(callback, context);
   }
 
   inverseFor(key) {
-    return this.modelClass.inverseFor(key);
+    return this.schema.inverseFor(key);
   }
 
   setupData(data) {
@@ -721,8 +733,9 @@ export default class InternalModel {
     //TODO(Igor) consider the polymorphic case
     Object.keys(preload).forEach((key) => {
       let preloadValue = get(preload, key);
-      let relationshipMeta = this.modelClass.metaForProperty(key);
-      if (relationshipMeta.isRelationship) {
+      let relationshipMeta = this.schema.relationships && this.schema.relationships[key];
+
+      if (relationshipMeta) {
         this._preloadRelationship(key, preloadValue);
       } else {
         this._data[key] = preloadValue;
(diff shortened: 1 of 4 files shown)


addon/-private/system/store.js
```diff
@@ -993,15 +999,14 @@ Store = Service.extend({
...
-  _internalModelForId(typeName, inputId) {
+  _internalModelForId(modelName, inputId) {
     heimdall.increment(_internalModelForId);
-    var typeClass = this.modelFor(typeName);
-    var id = coerceId(inputId);
-    var idToRecord = this.typeMapFor(typeClass).idToRecord;
-    var record = idToRecord[id];
+    let id = coerceId(inputId);
+    let schema = this.schemaFor(modelName);
+    let record = schema.recordMap.idToRecord[id];
```
runspired: this should be internalModel, I was carrying over the old names and should have noticed the old one was bad :trollface:

addon/-private/system/schema.js
```diff
@@ -0,0 +1,295 @@
...
+    return this.properties[propertyName];
+  }
+
+  get recordMap() {
+    if (this._recordMap === null) {
+      this._recordMap = {
+        idToRecord: new EmptyObject(),
+        records: [],
+        metadata: new EmptyObject(),
+        type: this.model
```
runspired: It's nothing, it should be `this.modelClass`

addon/-private/system/model/internal-model.js
```diff
@@ -857,44 +982,40 @@ InternalModel.prototype = {
...
-      return this.record.toString();
-    } else {
-      return `<${this.modelName}:${this.id}>`;
-    }
-  },
+    return `<${this.modelName}:${this.id}>`;
+  }
 
-  referenceFor(type, name) {
+  referenceFor(modelClass, name) {
```
tchak: here `type` is referring to relationship `kind` : `hasMany` or `belongsTo`

addon/-private/system/schema.js
tchak: Are you keeping `type` here and bellow because it would be too much work for now to refactor everything?

runspired: @tchak more or less yes, I actually went down the road of refactoring everything in an earlier spike, but the problem is that a few of these places leak to public code and I'm no sure if we consider this one to yet or not.

addon/-private/system/model/internal-model.js
runspired: whoops, will change that back. I think maybe we should consistently refer to this as `kind` internally? I noticed this confusion in another spot too.

runspired:
Early results for this branch have it at roughly (4ms / 238 records of 1 type over the 28ms of the internal-model branch it branches off of (this branch does not include the cache upgrades currently).  This is a happy surprise, because most of the perf improvements this unlocks have not been worked on yet.

fivetanley:
will give a second review tomorrow, but i am crying tears of joy right now.

addon/-private/system/model/internal-model.js
```diff
@@ -87,72 +92,170 @@ const {
...
+    if (this._schema === null) {
+      this._schema = this.store._schemaForModelClass(this.modelClass);
+    }
+
+    return this._schema;
+  }
+
+  get type() {
+    deprecate(`InternalModel.type has been deprecated in favor of InternalModel.modelClass`, {
+      id: 'ember-data.private.type-to-modelClass'
```
pangratz: needs an `until: "3.0"`

addon/-private/system/model/internal-model.js
pangratz: 👍  for using `kind` instead of `type`

addon/-private/system/schema.js
```diff
@@ -0,0 +1,295 @@
...
+    return this._adapter;
+  }
+
+  get serializer() {
+    if (this._serializer === null) {
+      this._serializer = this.store.serializerFor(this.modelName);
+    }
+    if (!this._serializer) {
+      throw new Error('Cant find serializer!');
+    }
```
pangratz: Is this a leftover from refactoring? Do we need this check for an serializer, since there should always be a default serializer, no?

addon/-private/system/schema.js
```diff
@@ -0,0 +1,295 @@
...
+    let inverseKind;
+    let inverse;
+
+    // If inverse is specified manually, return the inverse
+    if (options.inverse) {
+      inverseName = options.inverse;
+      inverse = inverseSchema.relationships[inverseName];
+
+      assert("We found no inverse relationships by the name of '" + inverseName + "' on the '" + inverseSchema.modelName +
+        "' model. This is most likely due to a missing attribute on your model definition.", !Ember.isNone(inverse));
```
pangratz: `s/attribute/relationship/`

addon/-private/system/store.js
```diff
@@ -1905,38 +1895,55 @@ Store = Service.extend({
...
-    typeMaps[guid] = typeMap;
+    return this._schemas[modelName];
+  },
+
+  _schemaForModelClass(modelClass) {
+    let modelName = modelClass.modelName;
+
+    if (this._schemas === null) {
+      this._schemas = new EmptyObject();
+    }
```
pangratz: I think it makes sense to initialize `this._schemas` in the `init`.

addon/-private/system/store.js
```diff
@@ -1905,38 +1895,55 @@ Store = Service.extend({
...
 
     @method typeMapFor
     @private
-    @param {DS.Model} typeClass
-    @return {Object} typeMap
+    @param {DS.Model} modelClass
+    @return {Object} recordMap
   */
-  typeMapFor(typeClass) {
+  typeMapFor(modelClass) {
```
pangratz: Should this be renamed to `recordMapFor` to better reflect what is returned?

runspired: @pangratz yes, but I suspect we leaked this publicly and should add a deprecation.

addon/-private/system/schema.js
runspired: I believe it's just a left over check for something that was failing for me early in the process and can be removed

runspired:
I rebased this, but I I want to shrink it's scope a bit and reconsider things. I should be able to land this by end of day tomorrow by branching off of #4664 and moving parts of the schema into `ClassMeta` instead and discarding much of the rest.

I think this implementation at this point in time is too broad, because we will likely want to expose a primitive (not necessarily exactly this one) to adapters in place of modelClass, but this one is not ready for that.  I have a plan, but it's happily more incremental, and will turn into multiple simpler PRs instead of one larger one.

runspired:
As luck would have it, once I started trying to port some of the iterator / model parse optimizations into DS.Model, turns out we still need this! :trollface: So nvm on the new direction, cleaning up this PR and shifting it's focus a wee bit to be an object tied more closely to DS.Model for now.

runspired:
Closing in favor of #4865

PR #4589, opened 2016-10-18
Feat/improved cache

runspired:
This PR branches off of the internal-model work in #4562 
Perf numbers for both this and that PR are in the works. I don't expect them to be overly substantial, as in the initial benchmarking these two areas were good chunks but not where the bulk of our costs were, but we should still see nice gains for both.

### DIFF
--- a/addon/-private/system/store.js
+++ b/addon/-private/system/store.js
@@ -4,7 +4,7 @@
 
 import Ember from 'ember';
 import Model from 'ember-data/model';
-import { instrument, assert, warn, runInDebug } from "ember-data/-private/debug";
+import { instrument, assert, deprecate, warn, runInDebug } from "ember-data/-private/debug";
 import _normalizeLink from "ember-data/-private/system/normalize-link";
 import normalizeModelName from "ember-data/-private/system/normalize-model-name";
 import { InvalidError } from 'ember-data/adapters/errors';
@@ -97,7 +97,6 @@ const {
   normalize,
   peekAll,
   peekRecord,
-  retrieveManagedInstance,
   serializerFor,
   typeMapFor,
   typeMapFor_allocate
@@ -116,7 +115,6 @@ const {
   'normalize',
   'peekAll',
   'peekRecord',
-  'retrieveManagedInstance',
   'serializerFor',
   'typeMapFor',
   'typeMapFor_allocate'
@@ -210,7 +208,8 @@ Store = Service.extend({
       store: this
     });
     this._pendingSave = [];
-    this._instanceCache = new ContainerInstanceCache(getOwner(this));
+    this._instanceCache = new ContainerInstanceCache(getOwner(this), this);
+
     //Used to keep track of all the find requests that need to be coalesced
     this._pendingFetch = Map.create();
   },
@@ -267,13 +266,11 @@ Store = Service.extend({
     @return DS.Adapter
   */
   defaultAdapter: Ember.computed('adapter', function() {
-    var adapter = get(this, 'adapter');
+    let adapter = get(this, 'adapter');
 
     assert('You tried to set `adapter` property to an instance of `DS.Adapter`, where it should be a name', typeof adapter === 'string');
 
-    adapter = this.retrieveManagedInstance('adapter', adapter);
-
-    return adapter;
+    return this.adapterFor(adapter);
   }),
 
   // .....................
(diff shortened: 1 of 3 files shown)


addon/-private/system/model/internal-model.js
```diff
@@ -483,54 +570,66 @@ InternalModel.prototype = {
...
+      setups = TransitionChainMap[transitionMapId].setups;
+      enters = TransitionChainMap[transitionMapId].enters;
+      state = TransitionChainMap[transitionMapId].state;
+    } else {
+      setups = [];
+      enters = [];
 
-    for (i=0, l=path.length; i<l; i++) {
-      state = state[path[i]];
+      let path = splitOnDot(name);
```
stefanpenner: this is also likely something we can cache/precompute

addon/-private/system/model/states.js
```diff
@@ -241,7 +244,7 @@ var DirtyState = {
 
     //TODO(Igor) reloading now triggers a
     //loadingData event, though it seems fine?
-    loadingData: Ember.K,
+    loadingData: K,
```
stefanpenner: can likely just use concise functions

addon/-private/system/store.js
```diff
@@ -2408,60 +2414,34 @@ Store = Service.extend({
...
-    set(instance, 'store', this);
-    return instance;
+    return this._instanceCache.get('serializer', normalizedModelName);
   },
 
   lookupAdapter(name) {
-    return this.retrieveManagedInstance('adapter', name, this.get('_adapterFallbacks'));
+    deprecate(`Use of lookupAdapter is deprecated, use adapterFor instead.`, {
+      id: 'ember-data:lookupAdapter',
+      since: '2.10'
```
pangratz: This should be `since: '3.0'`...

bmac: `until: '3.0'`

runspired: to clarify, both `since` and `until` should be set, c?

bmac: I believe `until` is the only option http://emberjs.com/api/#method_deprecate. Setting `since` should have no impact on the deprecation message.

runspired:
A sidenote on this caching, In our tests we hit the cache 5400 times for 41 total transition paths (instead of allocating and building the transition map each time).

runspired:
Eyeballing the production build benchmarks (no filtering of outliers but I am looking at a box plot), we moved from roughly 28ms spent in store.push for 238 records of a single type to a bit less than 27ms. However, the overall mass left shifted significantly and if I spend the time to filter the outliers it looks likely this was closer to a 2-2.5ms win over 481 adapter and serializer lookups. This is a bit less than I was hoping for.

fivetanley:
Why is the number 238 records in the numbers?

runspired:
@fivetanley common multiple of the various record distribution scenarios.

addon/-private/system/store/container-instance-cache.js
```diff
@@ -1,7 +1,15 @@
...
+const  {
+  __get,
+  _instanceFor
+} = heimdall.registerMonitor('system.store.container-instance-cache',
+  '__get',
+  '_instanceFor'
+);
 
 /*
  * The `ContainerInstanceCache` serves as a lazy cache for looking up
```
igorT: Please update the comments

runspired: updated

runspired:
@igorT has brought up that this approach entangles the cache and the store via closing over `getFallbacks`.  While I don't think this closure is a very big issue as store:cache is 1:1 and they are created and destroyed together, there may be a cleaner pattern of doing this by moving the _getFallbacks knowledge into the cache itself. I can justify this as the cache is already namespace aware.

I will make this change later today before we should merge this.

runspired:
@igor fallbacks are now in the cache

igorT:
Any perf #?

runspired:
@igort https://github.com/emberjs/data/pull/4589#issuecomment-254653187

> Eyeballing the production build benchmarks (no filtering of outliers but I am looking at a box plot), we moved from roughly 28ms spent in store.push for 238 records of a single type to a bit less than 27ms. However, the overall mass left shifted significantly and if I spend the time to filter the outliers it looks likely this was closer to a 2-2.5ms win over 481 adapter and serializer lookups. This is a bit less than I was hoping for.

PR #4591, opened 2016-10-19
Cleanup record array manager

stefanpenner:


addon/-private/system/record-arrays/adapter-populated-record-array.js
```diff
@@ -20,11 +20,15 @@ var get = Ember.get;
   @extends DS.RecordArray
 */
 export default RecordArray.extend({
-  query: null,
+  init() {
+    this._super(...arguments);
+    this.query = this.query || null;
```
stefanpenner: this is due to links is state provided by internal `loadRecords`, and not via the constructor

PR #4593, opened 2016-10-19
Free up `internalModel`s

### LINKED ISSUE #3296: Unload issue with relationships

hjdivad:
`internalModel`s need to stay around as long as they're connected to a live record via relationships.  Once an entire subgraph is removed however, we can free all of it.

Also add a couple of "late private" fields to `internalModel`'s constructor for shape preservation.

Includes work from @igort and @sly7-7

[Fix #3296]
[Supercede #3301]

### DIFF
--- a/addon/-private/system/model/internal-model.js
+++ b/addon/-private/system/model/internal-model.js
@@ -59,6 +59,16 @@ function extractPivotName(name) {
   );
 }
 
+function areAllModelsUnloaded(internalModels) {
+  for (let i=0; i<internalModels.length; ++i) {
+    let record = internalModels[i].record;
+    if (record && !(record.get('isDestroyed') || record.get('isDestroying'))) {
+      return false;
+    }
+  }
+  return true;
+}
+
 // this (and all heimdall instrumentation) will be stripped by a babel transform
 //  https://github.com/heimdalljs/babel5-plugin-strip-heimdall
 const {
@@ -88,6 +98,7 @@ const {
 );
 
 let InternalModelReferenceId = 1;
+let nextBfsId = 1;
 
 /*
   `InternalModel` is the Model class that we use internally inside Ember Data to represent models.
@@ -111,17 +122,25 @@ export default class InternalModel {
     this.id = id;
     this._internalId = InternalModelReferenceId++;
     this.store = store;
-    this._data = data || new EmptyObject();
     this.modelName = modelName;
-    this.dataHasInitialized = false;
     this._loadingPromise = null;
     this._record = null;
-    this.currentState = RootState.empty;
-    this.isReloading = false;
     this._isDestroyed = false;
     this.isError = false;
-    this.error = null;
     this._isUpdatingRecordArrays = false;
+
+    // During dematerialization we don't want to rematerialize the record.  The
+    // reason this might happen is that dematerialization removes records from
+    // record arrays,  and Ember arrays will always `objectAt(0)` and
+    // `objectAt(len - 1)` to test whether or not `firstObject` or `lastObject`
+    // have changed.
+    this._isDematerializing = false;
+
+    this.resetRecord();
+
+    if (data) {
+      this.__data = data;
+    }
 
     // caches for lazy getters
     this._modelClass = null;
@@ -129,10 +148,12 @@ export default class InternalModel {
     this.__recordArrays = null;
     this._references = null;
     this._recordReference = null;
-    this.__inFlightAttributes = null;
     this.__relationships = null;
-    this.__attributes = null;
     this.__implicitRelationships = null;
+
+    // Used during the mark phase of unloading to avoid checking the same internal
+    // model twice in the same scan
+    this._bfsId = 0;
   }
 
   get modelClass() {
@@ -145,7 +166,7 @@ export default class InternalModel {
 
   get recordReference() {
     if (this._recordReference === null) {
-      this._recordReference = new RecordReference(this.store, this)
+      this._recordReference = new RecordReference(this.store, this);
     }
     return this._recordReference;
   }
@@ -199,6 +220,17 @@ export default class InternalModel {
 
   set _inFlightAttributes(v) {
     this.__inFlightAttributes = v;
+  }
+
+  get _data() {
+    if (this.__data === null) {
+      this.__data = new EmptyObject();
+    }
+    return this.__data;
+  }
+
+  set _data(v) {
+    this.__data = v;
   }
 
   /*
@@ -276,7 +308,7 @@ export default class InternalModel {
   }
 
   getRecord() {
-    if (!this._record) {
+    if (!this._record && !this._isDematerializing) {
       heimdall.increment(materializeRecord);
       let token = heimdall.start('InternalModel.getRecord');
 
@@ -307,8 +339,25 @@ export default class InternalModel {
     return this._record;
   }
 
-  recordObjectWillDestroy() {
+  resetRecord() {
     this._record = null;
+    this.dataHasInitialized = false;
+    this.isReloading = false;
+    this.error = null;
+    this.currentState = RootState.empty;
+    this.__attributes = null;
+    this.__inFlightAttributes = null;
+    this._data = null;
+  }
+
+  dematerializeRecord() {
+    if (this.record) {
+      this._isDematerializing = true;
+      this.record.destroy();
+      this.destroyRelationships();
+      this.updateRecordArrays();
+      this.resetRecord();
+    }
   }
 
   deleteRecord() {
@@ -356,12 +405,107 @@ export default class InternalModel {
     });
   }
 
+  /**
+    Computes the set of internal models reachable from `this` across exactly one
+    relationship.
+
+    @return {Array} An array containing the internal models that `this` belongs
+    to or has many.
+  */
+  _directlyRelatedInternalModels() {
+    let array = [];
+    this.type.eachRelationship((key, relationship) => {
+      if (this._relationships.has(key)) {
+        let relationship = this._relationships.get(key);
+        let localRelationships = relationship.members.toArray();
+        let serverRelationships = relationship.canonicalMembers.toArray();
+
+        array = array.concat(localRelationships, serverRelationships);
+      }
+    });
+    return array;
+  }
+
+
+  /**
+    Computes the set of internal models reachable from this internal model.
+
+    Reachability is determined over the relationship graph (ie a graph where
+    nodes are internal models and edges are belongs to or has many
+    relationships).
+
+    @return {Array} An array including `this` and all internal models reachable
+    from `this`.
+  */
+  _allRelatedInternalModels() {
+    let array = [];
+    let queue = [];
+    let bfsId = nextBfsId++;
+    queue.push(this);
+    this._bfsId = bfsId;
+    while (queue.length > 0) {
+      let node = queue.shift();
+      array.push(node);
+      let related = node._directlyRelatedInternalModels();
+      for (let i=0; i<related.length; ++i) {
+        let internalModel = related[i];
+        assert('Internal Error: seen a future bfs iteration', internalModel._bfsId <= bfsId);
+        if (internalModel._bfsId < bfsId) {
+          queue.push(internalModel);
+          internalModel._bfsId = bfsId;
+        }
+      }
+    }
+    return array;
+  }
+
+
+  /**
+    Unload the record for this internal model. This will cause the record to be
+    destroyed and freed up for garbage collection. It will also do a check
+    for cleaning up internal models.
+
+    This check is performed by first computing the set of related internal
+    models. If all records in this set are unloaded, then the entire set is
+    destroyed. Otherwise, nothing in the set is destroyed.
+
+    This means that this internal model will be freed up for garbage collection
+    once all models that refer to it via some relationship are also unloaded.
+  */
   unloadRecord() {
     this.send('unloadRecord');
+    this.dematerializeRecord();
+    Ember.run.schedule('destroy', this, '_checkForOrphanedInternalModels');
+  }
+
+  _checkForOrphanedInternalModels() {
+    this._isDematerializing = false;
+    if (this.isDestroyed) { return; }
+
+    this._cleanupOrphanedInternalModels();
+  }
+
+  _cleanupOrphanedInternalModels() {
+    let relatedInternalModels = this._allRelatedInternalModels();
+    if (areAllModelsUnloaded(relatedInternalModels)) {
+      for (let i=0; i<relatedInternalModels.length; ++i) {
+        let internalModel = relatedInternalModels[i];
+        if (!internalModel.isDestroyed) {
+          internalModel.destroy();
+        }
+      }
+    }
   }
 
   eachRelationship(callback, binding) {
     return this.modelClass.eachRelationship(callback, binding);
+  }
+
+  destroy() {
+    assert("Cannot destroy an internalModel while its record is materialized", !this.record || this.record.get('isDestroyed') || this.record.get('isDestroying'));
+
+    this.store._removeFromIdMap(this);
+    this._isDestroyed = true;
   }
 
   eachAttribute(callback, binding) {
@@ -400,13 +544,6 @@ export default class InternalModel {
 
   get hasRecord() {
     return !!this._record;
-  }
-
-  destroy() {
-    this._isDestroyed = true;
-    if (this.hasRecord) {
-      return this.record.destroy();
-    }
   }
 
   /*
@@ -714,6 +851,18 @@ export default class InternalModel {
     });
     Object.keys(this._implicitRelationships).forEach((key) => {
       this._implicitRelationships[key].clear();
+      this._implicitRelationships[key].destroy();
+    });
+  }
+
+  destroyRelationships() {
+    this.eachRelationship((name, relationship) => {
+      if (this._relationships.has(name)) {
+        let rel = this._relationships.get(name);
+        rel.destroy();
+      }
+    });
+    Object.keys(this._implicitRelationships).forEach((key) => {
       this._implicitRelationships[key].destroy();
     });
   }
--- a/addon/-private/system/model/model.js
+++ b/addon/-private/system/model/model.js
@@ -807,14 +807,6 @@ const Model = Ember.Object.extend(Ember.Evented, {
     this._super(...arguments);
   },
 
-  willDestroy() {
-    //TODO Move!
-    this._super(...arguments);
-    this._internalModel.clearRelationships();
-    this._internalModel.recordObjectWillDestroy();
-    //TODO should we set internalModel to null here?
-  },
-
   // This is a temporary solution until we refactor DS.Model to not
   // rely on the data property.
   willMergeMixin(props) {
--- a/addon/-private/system/model/states.js
+++ b/addon/-private/system/model/states.js
@@ -459,10 +459,6 @@ const RootState = {
   // you out of the in-flight state.
   rolledBack() { },
   unloadRecord(internalModel) {
-    // clear relationships before moving to deleted state
-    // otherwise it fails
-    internalModel.clearRelationships();
-    internalModel.transitionTo('deleted.saved');
   },
 
   propertyWasReset() { },
@@ -573,10 +569,6 @@ const RootState = {
       },
 
       unloadRecord(internalModel) {
-        // clear relationships before moving to deleted state
-        // otherwise it fails
-        internalModel.clearRelationships();
-        internalModel.transitionTo('deleted.saved');
       },
 
       didCommit() {},
@@ -680,8 +672,6 @@ const RootState = {
 
       setup(internalModel) {
         internalModel.clearRelationships();
-        var store = internalModel.store;
-        store._dematerializeRecord(internalModel);
       },
 
       invokeLifecycleCallbacks(internalModel) {
--- a/addon/-private/system/store.js
+++ b/addon/-private/system/store.js
@@ -2541,15 +2541,13 @@ Store = Service.extend({
     When a record is destroyed, this un-indexes it and
     removes it from any record arrays so it can be GCed.
 
-    @method _dematerializeRecord
+    @method _removeFromIdMap
     @private
     @param {InternalModel} internalModel
   */
-  _dematerializeRecord(internalModel) {
+  _removeFromIdMap(internalModel) {
     let recordMap = this._recordMapFor(internalModel.modelName);
     let id = internalModel.id;
-
-    internalModel.updateRecordArrays();
 
     recordMap.remove(internalModel, id);
   },
--- a/tests/integration/adapter/find-all-test.js
+++ b/tests/integration/adapter/find-all-test.js
@@ -19,6 +19,7 @@ module("integration/adapter/find_all - Finding All Records of a Type", {
       firstName: attr('string'),
       lastName: attr('string')
     });
+    Person.reopenClass({ toString() { return 'Person'; } });
 
     allRecords = null;
 
--- a/tests/integration/records/unload-test.js
+++ b/tests/integration/records/unload-test.js
@@ -7,33 +7,45 @@ import {module, test} from 'qunit';
 
 import DS from 'ember-data';
 
-var attr = DS.attr;
-var belongsTo = DS.belongsTo;
-var hasMany = DS.hasMany;
-var run = Ember.run;
-var env;
+let attr = DS.attr;
+let belongsTo = DS.belongsTo;
+let hasMany = DS.hasMany;
+let run = Ember.run;
+let env;
 
-var Person = DS.Model.extend({
+let Person = DS.Model.extend({
   name: attr('string'),
-  cars: hasMany('car', { async: false })
+  cars: hasMany('car', { async: false }),
+  boats: hasMany('boat', { async: true })
 });
+Person.reopenClass({ toString() { return 'Person'; } });
 
-var Group = DS.Model.extend({
+let Group = DS.Model.extend({
   people: hasMany('person', { async: false })
 });
+Group.reopenClass({ toString() { return 'Group'; } });
 
-var Car = DS.Model.extend({
+let Car = DS.Model.extend({
   make: attr('string'),
   model: attr('string'),
   person: belongsTo('person', { async: false })
 });
+Car.reopenClass({ toString() { return 'Car'; } });
+
+let Boat = DS.Model.extend({
+  name: attr('string'),
+  person: belongsTo('person', { async: false })
+});
+Boat.toString = function() { return 'Boat'; };
 
 module("integration/unload - Unloading Records", {
   beforeEach() {
     env = setupStore({
+      adapter: DS.JSONAPIAdapter,
       person: Person,
       car: Car,
-      group: Group
+      group: Group,
+      boat: Boat
     });
   },
 
@@ -45,7 +57,7 @@ module("integration/unload - Unloading Records", {
 });
 
 test("can unload a single record", function(assert) {
-  var adam;
+  let adam;
   run(function() {
     env.store.push({
       data: {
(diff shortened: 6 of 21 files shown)


addon/-private/system/model/states.js
```diff
@@ -676,8 +676,10 @@ const RootState = {
 
       setup(internalModel) {
         internalModel.clearRelationships();
-        var store = internalModel.store;
-        store._dematerializeRecord(internalModel);
+        // var store = internalModel.store;
+        // store._dematerializeRecord(internalModel);
+
+        // TODO: internalModel.dematerializeRecord() ?
```
hjdivad: @igort do we actually want to dematerialize on `deleted.saved`?  seems suspect

hjdivad:
cc @stefanpenner @pangratz @igorT @sly7-7

addon/-private/system/model/internal-model.js
```diff
@@ -235,8 +241,68 @@ InternalModel.prototype = {
...
+    if (allUnloaded) {
+      relatedInternalModels.forEach((internalModel) => {
+        if (!internalModel.isDestroyed) {
+          internalModel.destroy();
+        }
+      });
+    }
+  },
+
+  destroy: function() {
```
stefanpenner: concise function

addon/-private/system/model/internal-model.js
```diff
@@ -324,6 +383,7 @@ InternalModel.prototype = {
     @private
   */
   pushedData() {
+    this._isDestroyed = false;
```
stefanpenner: this revives internalModels ?

addon/-private/system/model/model.js
```diff
@@ -803,9 +803,9 @@ var Model = Ember.Object.extend(Ember.Evented, {
   willDestroy() {
     //TODO Move!
     this._super(...arguments);
-    this._internalModel.clearRelationships();
-    this._internalModel.recordObjectWillDestroy();
-    //TODO should we set internalModel to null here?
+
+    this._internalModel._resetRecord();
```
stefanpenner: is there a point to resetting the internal model here, doesn't it just  "go dead" once destroyed?

tests/integration/adapter/find-all-test.js
```diff
@@ -19,6 +19,7 @@ module("integration/adapter/find_all - Finding All Records of a Type", {
       firstName: attr('string'),
       lastName: attr('string')
     });
+    Person.reopenClass({ toString: () => 'Person' });
```
stefanpenner: concise function.

tests/integration/records/unload-test.js
```diff
@@ -17,16 +17,25 @@ var Person = DS.Model.extend({
   name: attr('string'),
   cars: hasMany('car', { async: false })
 });
+Person.reopenClass({
+  toString: () => 'Person'
```
stefanpenner: ^

tests/integration/records/unload-test.js
```diff
@@ -17,16 +17,25 @@ var Person = DS.Model.extend({
...
+});
 
 var Group = DS.Model.extend({
   people: hasMany('person', { async: false })
 });
+Group.reopenClass({
+  toString: () => 'Group'
```
stefanpenner: ^

tests/integration/records/unload-test.js
```diff
@@ -17,16 +17,25 @@ var Person = DS.Model.extend({
...
+});
 
 var Car = DS.Model.extend({
   make: attr('string'),
   model: attr('string'),
   person: belongsTo('person', { async: false })
 });
+Car.reopenClass({
+  toString: () => 'Car'
```
stefanpenner: ^

tests/integration/records/unload-test.js
```diff
@@ -280,55 +289,108 @@ test("unloading a record also clears its relationship", function(assert) {
         person.unloadRecord();
       });
 
-      assert.equal(person.get('cars.length'), undefined);
+      assert.expectAssertion(function () {
```
stefanpenner: =>

tests/integration/records/unload-test.js
```diff
@@ -280,55 +289,108 @@ test("unloading a record also clears its relationship", function(assert) {
...
   });
-});
-
-test("unloading a record also clears the implicit inverse relationships", function(assert) {
-  var adam, bob;
-  // disable background reloading so we do not re-create the relationship.
-  env.adapter.shouldBackgroundReloadRecord = () => false;
 
-  run(function() {
+  run(function () {
```
stefanpenner: =>

tests/integration/records/unload-test.js
```diff
@@ -280,55 +289,108 @@ test("unloading a record also clears its relationship", function(assert) {
...
   });
 
-  run(function() {
-    env.store.findRecord('group', 1).then(function(group) {
-      assert.equal(group.get('people.length'), 1, 'The inital length of people is correct');
-      var person = env.store.peekRecord('person', 1);
-      run(function() {
-        person.unloadRecord();
-      });
+  run(function () {
```
stefanpenner: =>

addon/-private/system/model/internal-model.js
```diff
@@ -184,6 +182,14 @@ InternalModel.prototype = {
     this.record = null;
   },
 
+  dematerializeRecord() {
+    if (this.record) {
```
stefanpenner: when do we call dematerializeRecord and there is no record?

addon/-private/system/model/internal-model.js
```diff
@@ -235,8 +241,68 @@ InternalModel.prototype = {
     return this.record;
   },
 
-  unloadRecord() {
+  directlyRelatedInternalModels() {
```
stefanpenner: brief doc explaining why this method exists.

addon/-private/system/model/internal-model.js
```diff
@@ -235,8 +241,68 @@ InternalModel.prototype = {
...
+    var array = [];
+    this.type.eachRelationship((key, relationship) => {
```
stefanpenner: we may want to use `eachRelationshipByName` cp here: https://github.com/emberjs/data/blob/c9d8212c857ca78218ad98d11621819b38dba98f/addon/-private/system/relationships/ext.js#L486 because it is cached (eachRelBy Name can be costly), and seems like its provides what we want here.

addon/-private/system/model/internal-model.js
```diff
@@ -235,8 +241,68 @@ InternalModel.prototype = {
...
+    let array = [];
+    let queue = [];
+    let seen = [];
+    queue.push(this);
+    seen.push(this);
+    while (queue.length > 0) {
+      let node = queue.shift();
+      array.push(node);
+      let related = node.directlyRelatedInternalModels();
+      related.forEach(function(internalModel) {
```
stefanpenner: =>

addon/-private/system/model/internal-model.js
```diff
@@ -235,8 +241,68 @@ InternalModel.prototype = {
...
+        if (seen.indexOf(internalModel) === -1) {
+          queue.push(internalModel);
+          seen.push(internalModel);
+        }
+      });
+    }
+    return array;
+  },
+
+  unloadRecord: function() {
```
stefanpenner: concise function

addon/-private/system/model/internal-model.js
```diff
@@ -235,8 +241,68 @@ InternalModel.prototype = {
     return this.record;
   },
 
-  unloadRecord() {
+  directlyRelatedInternalModels() {
+    var array = [];
```
stefanpenner: let?

addon/-private/system/model/internal-model.js
```diff
@@ -235,8 +241,68 @@ InternalModel.prototype = {
...
+    var record;
+    for (var i=0; i < relatedInternalModels.length; i++) {
+      record = relatedInternalModels[i].record;
+      if (record && !(record.get('isDestroyed') || record.get('isDestroying'))) {
+        allUnloaded = false;
+        break;
+      }
+    }
+    if (allUnloaded) {
+      relatedInternalModels.forEach((internalModel) => {
```
stefanpenner: why for loop on 281, but forEach here

addon/-private/system/model/internal-model.js
```diff
@@ -235,8 +241,68 @@ InternalModel.prototype = {
...
+        if (!internalModel.isDestroyed) {
+          internalModel.destroy();
+        }
+      });
+    }
+  },
+
+  destroy: function() {
+    this._isDestroyed = true;
```
stefanpenner: we should likely mimic embers flags:

* `isDestroying = true` during distruction
* `isDestroyed = true` once destroyed

addon/-private/system/model/internal-model.js
```diff
@@ -235,8 +241,68 @@ InternalModel.prototype = {
     return this.record;
   },
 
-  unloadRecord() {
+  directlyRelatedInternalModels() {
+    var array = [];
+    this.type.eachRelationship((key, relationship) => {
+      if (this._relationships.has(key)) {
+        var related = this._relationships.get(key).members.toArray();
+        array.push(...related);
```
stefanpenner: this may be fine, but there is an upper limit (in the low hundreds of thousands) that one can push onto an array.

addon/-private/system/model/model.js
igorT: No, the internal model lives on, the record goes dead

addon/-private/system/model/internal-model.js
stefanpenner: we should benchmark

addon/-private/system/model/internal-model.js
```diff
@@ -584,6 +644,15 @@ InternalModel.prototype = {
     });
   },
 
+  _resetRecord() {
```
igorT: Can you put a comment saying that this doesn't reset relationships/attributes on purpose, given that constraint, the name is a bit unfortunate

addon/-private/system/store.js
```diff
@@ -794,6 +794,7 @@ Store = Service.extend({
 
     var resolver = Ember.RSVP.defer('Fetching ' + typeClass + 'with id: ' + internalModel.id);
     var pendingFetchItem = {
+      // TODO: s/record/internalModel
```
igorT: TODO

addon/-private/system/store.js
```diff
@@ -2430,12 +2430,14 @@ Store = Service.extend({
     @param {InternalModel} internalModel
   */
   _dematerializeRecord(internalModel) {
```
igorT: Maybe move to the `internalModel`

tests/integration/records/unload-test.js
```diff
@@ -280,55 +289,108 @@ test("unloading a record also clears its relationship", function(assert) {
...
 
-  run(function() {
+  var rematerializedPerson = bob.get('person');
+  assert.equal(rematerializedPerson.get('.id'), '1');
+  // the person is rematerialized; the previous person is *not* re-used
+  assert.notEqual(rematerializedPerson, adam);
+});
+
+
+test('unloading a disconnected subgraph clears the relevant internal models', function(assert) {
```
igorT: Maybe a test for unloading a part of a connected graph doesn't do that?

tests/unit/store/unload-test.js
```diff
@@ -47,7 +49,7 @@ testInDebug("unload a dirty record asserts", function(assert) {
       }
     });
 
-    store.findRecord('record', 1).then(function(record) {
+    store.findRecord('record', 1, { backgroundReload: false }).then(function(record) {
```
pangratz: The `store.findRecord` is unnecessary for this test. Let's use `store.peekRecord` instead.

tests/unit/store/unload-test.js
```diff
@@ -157,18 +158,23 @@ test("can commit store after unload record with relationships", function(assert)
 
     asyncRecords = Ember.RSVP.hash({
       brand: store.findRecord('brand', 1),
-      product: store.findRecord('product', 1)
+      product: store.findRecord('product', 1, { backgroundReload: false })
```
pangratz: `store.peekRecord`

tests/unit/store/unload-test.js
```diff
@@ -157,18 +158,23 @@ test("can commit store after unload record with relationships", function(assert)
...
       records.like = like.save();
       return Ember.RSVP.hash(records);
     }).then(function(records) {
       store.unloadRecord(records.product);
+    });
+  });
+
+  run(function () {
+    promise.then(function () {
       return store.findRecord('product', 1);
```
pangratz: `store.peekRecord`

addon/-private/system/model/internal-model.js
```diff
@@ -95,20 +95,18 @@ export default function InternalModel(type, id, store, _, data) {
   this.store = store;
   this._data = data || new EmptyObject();
   this.modelName = type.modelName;
-  this.dataHasInitialized = false;
+  this._loadingPromise = null;
```
pangratz: Seems unused?

addon/-private/system/model/states.js
hjdivad: per our discussion offline we'll unload the record (which will dematerialize)

addon/-private/system/model/internal-model.js
hjdivad: 👍

addon/-private/system/model/internal-model.js
hjdivad: also this should be `_private`

addon/-private/system/model/internal-model.js
hjdivad: 👍

addon/-private/system/model/internal-model.js
hjdivad: 👍

addon/-private/system/model/internal-model.js
hjdivad: because of `break`

addon/-private/system/model/internal-model.js
hjdivad: 👍

tests/integration/adapter/find-all-test.js
hjdivad: 👍

tests/integration/records/unload-test.js
hjdivad: 👍

tests/integration/records/unload-test.js
hjdivad: 👍

tests/integration/records/unload-test.js
hjdivad: 👍

tests/integration/records/unload-test.js
hjdivad: 👍

addon/-private/system/model/internal-model.js
hjdivad: i'll put in a comment, but even so, can you suggest a better name?

addon/-private/system/model/internal-model.js
hjdivad: it is written to elsewhere;   i have added it here to preserve shape.

Really though we should add a setter and not have other parts of the system write private fields

addon/-private/system/model/internal-model.js
```diff
@@ -235,8 +241,68 @@ InternalModel.prototype = {
...
+    }
+  },
+
+  destroy: function() {
+    this._isDestroyed = true;
+    Ember.run.schedule('destroy', this, this.willDestroy, this);
+  },
+
+  willDestroy() {
+    Ember.assert("Cannot destroy an internalModel while its record is materialized", !this.record);
```
hjdivad: 👍

tests/integration/adapter/find-all-test.js
pangratz: I am curious why this is needed. Which tests are failing without this changes? I would guess only tests wich check for assertions, right? Maybe we should only specify the `toString` in those tests...

wecc:
I think this issue would close #4126

hjdivad:
> I think this issue would close #4126 

yes it should

hjdivad:
appveyor build is complaining about downloading phantom.  Just need someone to restart.

@pangratz this should resolve your concerns
@igort this should resolve your concerns except for deleting implicitly unloading: we should do this in a separate effort.

@stefanpenner this should resolve your concerns except for `eachRelationship` vs `eachRelationshipByName`.  That shouldn't block the PR; we can switch it after doing some benchmarks

bmac:
This looks good to me. 👍 

@hjdivad I think @igorT is the only one who can restart an AppVeyor build. So in the past we've been comfortable merged prs when AppVeyor is red and travis is green, if the reason for the AppVeyor failure has to do with a network error during install and there aren't any build changes in the pr.

addon/-private/system/model/internal-model.js
```diff
@@ -333,14 +359,102 @@ export default class InternalModel {
     return this.record;
   }
 
+
+  /**
+    Computes the set of internal models reachable from `this` acrosxactly one
+    relationship.
+
+    @return {Array} An array containing the internal models that `t` belongs
```
pangratz: ``` s/`t`/`this`/ ```?

addon/-private/system/model/internal-model.js
```diff
@@ -333,14 +359,102 @@ export default class InternalModel {
     return this.record;
   }
 
+
+  /**
+    Computes the set of internal models reachable from `this` acrosxactly one
```
pangratz: acrosxactly?

addon/-private/system/model/internal-model.js
```diff
@@ -333,14 +359,102 @@ export default class InternalModel {
...
+
+
+  /**
+    Computes the set of internal models reachable from this internaodel.
+
+    Reachability is determind over the relatinoship graph (ie a grawhere
+    nodes are internal models and edges are belongs to or has many
+    relationships.
+
+    @return {Array} An array including `this` and all internal modereachable
```
pangratz: modereachable

addon/-private/system/model/internal-model.js
```diff
@@ -333,14 +359,102 @@ export default class InternalModel {
...
+      }
+    });
+    return array;
+  }
+
+
+  /**
+    Computes the set of internal models reachable from this internaodel.
+
+    Reachability is determind over the relatinoship graph (ie a grawhere
```
pangratz: typo: determind . relatinoship, grawhere

addon/-private/system/model/internal-model.js
```diff
@@ -333,14 +359,102 @@ export default class InternalModel {
...
+          array.push(related[i]);
+        }
+      }
+    });
+    return array;
+  }
+
+
+  /**
+    Computes the set of internal models reachable from this internaodel.
```
pangratz: internaodel

addon/-private/system/model/internal-model.js
```diff
@@ -333,14 +359,102 @@ export default class InternalModel {
...
+          internalModel._bfsId = bfsId;
+        }
+      }
+    }
+    return array;
+  }
+
+
+  /**
+    Unload the record for this internal model.  This will cause thecord to be
```
pangratz: thecord

addon/-private/system/model/internal-model.js
```diff
@@ -333,14 +359,102 @@ export default class InternalModel {
...
+    destroyed and freed up for garbage collection.  It will also trer a check
```
pangratz: trer

addon/-private/system/model/internal-model.js
pangratz: Also https://github.com/hjdivad/data/pull/2

hjdivad:
@bmac turns out I can also restart an AppVeyor build by writing comments with such atrocious spelling that @pangratz has to save me with a new commit.

addon/-private/system/model/internal-model.js
```diff
@@ -111,28 +123,28 @@ export default class InternalModel {
...
     // caches for lazy getters
     this.__deferredTriggers = null;
     this._references = null;
     this._recordReference = null;
-    this.__inFlightAttributes = null;
     this.__relationships = null;
-    this.__attributes = null;
     this.__implicitRelationships = null;
+
+    // Used for coloring during BFS
```
stefanpenner: should likely explain just a-bit more. BFS doesn't help us understand why/where this is used. Likely would be enough to say "during marking phase of unloading"

addon/-private/system/model/internal-model.js
```diff
@@ -278,8 +290,22 @@ export default class InternalModel {
     this._triggerDeferredTriggers();
   }
 
-  recordObjectWillDestroy() {
+  resetRecord() {
     this.record = null;
```
stefanpenner: is there actual value in reseting the record during destroy, or can we just let it go dead.

addon/-private/system/model/internal-model.js
```diff
@@ -278,8 +290,22 @@ export default class InternalModel {
...
+    this.dataHasInitialized = false;
+    this.isReloading = false;
+    this.error = null;
+    this.currentState = RootState.empty;
+    this.__attributes = null;
+    this.__inFlightAttributes = null;
+  }
+
+  dematerializeRecord() {
+    if (this.record) {
```
stefanpenner: Curious, should we assert in development if we attempt to dematerialize something already dematerialized? Wouldn't it mean we are doing duplicate work?

addon/-private/system/model/internal-model.js
```diff
@@ -333,14 +359,102 @@ export default class InternalModel {
...
+    relationship.
+
+    @return {Array} An array containing the internal models that `this` belongs
+    to or has many.
+  */
+  _directlyRelatedInternalModels() {
+    let array = [];
+    this.type.eachRelationship((key, relationship) => {
+      if (this._relationships.has(key)) {
+        let related = this._relationships.get(key).members.toArray();
```
stefanpenner: should we create this array, to just iterate then throw it away?

addon/-private/system/model/internal-model.js
hjdivad: 👍

addon/-private/system/model/internal-model.js
hjdivad: The case where we wind up here with `!this.record` is `store.unloadAll(type)` which is public API.

We could change this to assert `!!this.record` and then tweak the store to only unload materialized internal models.  thoughts?

bmac:
@hjdivad do you mind rebasing this pr?

bmac:
ping @hjdivad

runspired:
ping @hjdivad, if you don't have time (as I suspect you don't, hi office budy! 👋 ) just say so and I'll pull down your work and try to move it along.

trumb1mj:
Any updates on this?  I'm running into just this problem.  Is there any brute force method to get by this until it has been fixed?

hjdivad:
@runspired @bmac gonna try to get to it this weekend

@runspired i'll likely need to sync up with you re your recent changes in this area

runspired:
@hjdivad I may have to come also crash at your place :P 

@trumb1mj I don't think you are hitting this issue exactly but a different retention issue that's related.

That said, our intention is to make it so that internal models ARE kept if a relationship still has access to them, while your need is explicitly for them to be unloaded even when one does. David @igorT and myself will need to consider this case.

trumb1mj:
@runspired is there any way to build my relationships to avoid this?  

Also, I don't understand why deleting an object with a many-to-1 would cause any residual references to be left around--the other way makes a bit of sense but I still think we should be able to opt out and just whack things.  Not disagreeing just not quite understanding.

addon/-private/system/model/internal-model.js
```diff
@@ -113,26 +124,29 @@ export default class InternalModel {
     this.store = store;
     this._data = data || new EmptyObject();
     this.modelName = modelName;
-    this.dataHasInitialized = false;
     this._loadingPromise = null;
     this._record = null;
-    this.currentState = RootState.empty;
-    this.isReloading = false;
+    // TODO: see if we can clean this up
+    this._isDestroying = false;
```
igorT: Either way lets leave a comment on why/how it's used

addon/-private/system/model/internal-model.js
```diff
@@ -113,26 +124,29 @@ export default class InternalModel {
...
     this._isDestroyed = false;
     this.isError = false;
-    this.error = null;
     this._isUpdatingRecordArrays = false;
+    this._isDematerializing = false;
```
igorT: Same here

addon/-private/system/model/internal-model.js
```diff
@@ -275,8 +289,9 @@ export default class InternalModel {
     return this._record;
   }
 
+  // TODO: fix the wat of 'get record' and 'getRecord'
   getRecord() {
-    if (!this._record) {
+    if (!this._record && !this._isDematerializing) {
```
igorT: We should comment this

addon/-private/system/model/internal-model.js
```diff
@@ -307,8 +322,25 @@ export default class InternalModel {
...
-  recordObjectWillDestroy() {
+  resetRecord() {
+    // TODO: get laziness working again?
     this._record = null;
+    this.dataHasInitialized = false;
+    this.isReloading = false;
+    this.error = null;
+    this.currentState = RootState.empty;
+    this.__attributes = null;
+    this.__inFlightAttributes = null;
```
igorT: should we clear out data?

addon/-private/system/model/internal-model.js
```diff
@@ -356,14 +388,109 @@ export default class InternalModel {
...
+    while (queue.length > 0) {
+      let node = queue.shift();
+      array.push(node);
+      let related = node._directlyRelatedInternalModels();
+      for (let i=0; i<related.length; ++i) {
+        let internalModel = related[i];
+        if (internalModel._bfsId < bfsId) {
+          queue.push(internalModel);
+          internalModel._bfsId = bfsId;
+        }
```
igorT: would it be worth adding a dev assert if we ever see a case of `internalModel._bfsId > bfsId` ? That would mean bad things are happening

addon/-private/system/model/internal-model.js
```diff
@@ -356,14 +388,109 @@ export default class InternalModel {
...
+    once all models that refer to it via some relationship are also unloaded.
+  */
   unloadRecord() {
     this.send('unloadRecord');
+    this.dematerializeRecord();
+    Ember.run.schedule('destroy', this, '_cleanupOrphanedInternalModels');
+  }
+
+  _cleanupOrphanedInternalModels() {
+    this._isDematerializing = false;
```
igorT: Should comment on why we do this here

addon/-private/system/model/internal-model.js
```diff
@@ -356,14 +388,109 @@ export default class InternalModel {
...
+  destroy() {
+    this._isDestroying = true;
+    Ember.run.schedule('destroy', this, this.willDestroy, this);
+  }
+
+  willDestroy() {
+    assert("Cannot destroy an internalModel while its record materialized", !this.record || this.record.get('isDestroyed') || this.record.get('isDestroying'));
+
+    this.store._removeFromIdMap(this);
+    this._isDestroyed = true;
```
igorT: Should we set isDestroying to false?

igorT: If we still need it

addon/-private/system/model/model.js
```diff
@@ -808,11 +808,8 @@ const Model = Ember.Object.extend(Ember.Evented, {
   },
 
   willDestroy() {
-    //TODO Move!
     this._super(...arguments);
-    this._internalModel.clearRelationships();
-    this._internalModel.recordObjectWillDestroy();
-    //TODO should we set internalModel to null here?
+    this._internalModel.resetRecord();
```
igorT: Seems odd to call this from the record into the internalModel. Can we move it somewhere else?

Kerrick:
Did this seem to leave anybody else with severe performance problems?

In previous versions of Ember Data it was relatively quick to `store.unloadAll('my-model-type')` when I had ~2500 of them loaded, but in Ember Data 2.12 it takes upwards of 25 seconds. (In case it's relevant, my model has 5 `hasMany` relationships, and 3 `belongsTo` relationships.)

image


image

stefanpenner:
@kerrick can you open a fresh issue and provide a reproduction. Sounds like something we should look into and that will make this actionable

tylerturdenpants:
@Kerrick I'm also experiencing exactly the same issue. 2.12 was super fast. 2.13-2.15 is so slow at unloading the records that it seems to hang the browser. I've also noticed given certain relationships lead to a memory leak that quickly grows to the gigabyte range of memory in a minute or two and then crashes.

PR #4594, opened 2016-10-19
Revert "unloadRecord does not remove observers"

pangratz:
Reverts emberjs/data#4277

@ryanpatrickcook Looks like #4277 is not green on current master. Can you reopen a PR based on the latest `master`?

### DIFF
--- a/addon/-private/system/model/model.js
+++ b/addon/-private/system/model/model.js
@@ -122,8 +122,8 @@ var Model = Ember.Object.extend(Ember.Evented, {
     @type {Boolean}
     @readOnly
   */
-  hasDirtyAttributes: Ember.computed('currentState', function() {
-    return this.get('_internalModel.currentState.isDirty');
+  hasDirtyAttributes: Ember.computed('currentState.isDirty', function() {
+    return this.get('currentState.isDirty');
   }),
   /**
     If this property is `true` the record is in the `saving` state. A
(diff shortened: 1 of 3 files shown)


addon/-private/system/model/model.js
```diff
@@ -122,8 +122,8 @@ var Model = Ember.Object.extend(Ember.Evented, {
     @type {Boolean}
     @readOnly
   */
-  hasDirtyAttributes: Ember.computed('currentState', function() {
-    return this.get('_internalModel.currentState.isDirty');
+  hasDirtyAttributes: Ember.computed('currentState.isDirty', function() {
```
stefanpenner: this seems like it could be a `Ember.computed.readOnly('currentState.isDirty')`

ryanpatrickcook:
@pangratz bummer. Looks like it was [failing](https://travis-ci.org/emberjs/data/builds/168950093) on the alpha, beta, and canary builds. Seems like there was a change [here](https://github.com/emberjs/ember.js/commit/a98b4f98a514a48921d45397823c4a25d10007a9) that may have caused my issue.

After the destroy in unloadRecord, it seems to fail on initial access to `get(record, 'isDeleted')`, but subsequent attempts are ok.

screen shot 2016-10-19 at 3 02 26 pm

Will look into. Any help would be appreciated.

pangratz:
@ryanpatrickcook looks like there is similar work going on in #4593, which should address the whole unload situation at a broader level. So I am not sure if it's worth pursuing this further for now... Maybe we should wait until that PR is finished and re-check if the test in your PR is still failing or not.

PR #4600, opened 2016-10-19
make DS.ManyArray lazy

tchak:
This is mostly @igorT typing on my keyboard :)


igorT:
We should add a test that verifies it is indeed lazy

igorT:
Based on measurement this leads to a 15%+ speed improvement when loading hasMany relationships. (one of the scenarios with 5 hasMany's showed a 50% improvement)

tchak:
@igorT test added

addon/-private/system/references/has-many.js
```diff
@@ -45,11 +48,11 @@ HasManyReference.prototype.ids = function() {
...
 
 HasManyReference.prototype.meta = function() {
-  return this.hasManyRelationship.manyArray.meta;
+  return this.hasManyRelationship.meta;
 };
 
 HasManyReference.prototype.push = function(objectOrPromise) {
-  return Ember.RSVP.resolve(objectOrPromise).then((payload) => {
+  return resolve(objectOrPromise).then((payload) => {
     var array = payload;
```
tchak: We have eventually the same problem with `belongsTo`. I think it's best to create an issue and make a separate PR to fix it (if it's really a problem). I have not changed anything significant in this PR regarding references.

stefanpenner: sure, open an issue and link back here, so we can address.

stefanpenner:
This causes a perf issue and breaking apps using glimmer (which detects this scenario)

Specifically, when a relationship is lazily consumed it should not have side-affects, such as invalidating itself.

POC fix to unblock: https://github.com/emberjs/data/pull/4643, but working on a more comprehensive one.

PR #4601, opened 2016-10-20
Naming cleanup

runspired:
Attempting to enforce consistent use of

- internalModel: for InternalModel instances
- record: for DS.Model instances
- modelClass: for extended DS.Model
- kind: for whether something is a belongsTo/hasMany
- modelName: for the name of a modelClass


bmac:
I'm excited for this change. 🎉

tests/unit/model/internal-model-test.js
```diff
@@ -1,24 +0,0 @@
...
-
-testInDebug("Materializing a model twice errors out", function(assert) {
-  assert.expect(1);
-  var internalModel = new DS.InternalModel(MockModelFactory, null, { }, null);
-
-  internalModel.materializeRecord();
-  assert.expectAssertion(function() {
-    internalModel.materializeRecord();
-  }, /more than once/);
-});
```
pangratz: What's the reasoning behind removing this test?

runspired: it's now impossible to trigger this assertion

runspired: @pangratz I think internal model needs tests, but that's a separate PR, I removed this test because it's now not possible to materialize a record more than once.

Issue #4414, opened 2016-06-01
findRecord does not respect reload option when id string does not match result


achidlow:
I'm using token authentication, and have a REST endpoint /users/me that returns the record for the current user based on the authorization token.

When logging out one user and logging in another, even though `findRecord` is called with `{reload: true}`, it doesn't hit up the API.

I've tried various workarounds, the best I could come up with is to find the record with key 'me' in the store and unload it manually.

I've tried stepping through the source code to figure out why this is happening, but I'm very new to front-end development. It looks like the record gets stuck in the `loading` state, so you can't `peekRecord()` it directly either.

``` javascript
import Ember from 'ember';
import DS from 'ember-data';
import ESASession from "ember-simple-auth/services/session";

export default ESASession.extend({
    store: Ember.inject.service(),

    user: Ember.computed('isAuthenticated', function()
    {
        if (this.get('isAuthenticated')) {
            let store = this.get('store');
            // this shouldn't be necessary! ember-data for some reason doesn't like the 'me'
            // key, and the reload option isn't being respected
            const me = store.peekAll('user').findBy('id', 'me');
            if (me) {
                store.unloadRecord(me);
            }
            const promise = store.findRecord('user', 'me', {reload: true});
            return DS.PromiseObject.create({ promise: promise });
        }
    })
});
```

bmac:
@achidlow This sounds like a bug. I assume at some point when you do a `store.findRecord('user', 'me', {reload: true});` your api is returning a response to indicate that there is no `user/me` record. Is that correct?

achidlow:
@bmac thanks for your response!

I've just double checked and this occurring even in the case that `/users/me` only ever returns a user. So for example:
1. The page is initially loaded with the user already logged in (I'm using ember-simple-auth so the user token is persisted to localstorage), which makes a call to `/users/me` to get user info and successfully retrieves user.
2. The user log outs, which doesn't make any API calls, 
3. A different user logs in, which should be making a call to `/users/me` because of the `reload` option, but isn't. 

You can see the above in this screenshot, which was from running without my `unloadRecord` workaround:
![reloadnotworking](https://cloud.githubusercontent.com/assets/5511888/15767414/4c47b210-297a-11e6-800d-dec21a961593.png)

pangratz:
I suspect you're running into the issue where you are requesting a record with id `me` but your server returns a record with a different id (the real one).

The problem is that now there are 2 records within the store:

- a record with id returned by the server (in the loaded state)
[...]

PR #4604, opened 2016-10-20
Warn when findRecord returns a different id than the one requested

pangratz:
This addresses the problem stumbled over in #4414.

Basically we can't `assert` here as this would be a breaking change, since the bug mostly will not occur unless you try to reload the originally requested record again.

cc @igorT

### DIFF
--- a/addon/-private/system/store/finders.js
+++ b/addon/-private/system/store/finders.js
@@ -1,5 +1,5 @@
 import Ember from 'ember';
-import { assert } from "ember-data/-private/debug";
+import { assert, warn } from "ember-data/-private/debug";
 import {
   _bind,
   _guard,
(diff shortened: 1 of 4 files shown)


addon/-private/system/store/finders.js
```diff
@@ -38,6 +38,11 @@ export function _find(adapter, store, typeClass, id, internalModel, options) {
     return store._adapterRun(function() {
       var payload = normalizeResponseHelper(serializer, store, typeClass, adapterPayload, id, 'findRecord');
       assert('Ember Data expected the primary data returned from a `findRecord` response to be an object but instead it found an array.', !Array.isArray(payload.data));
+
+      warn(`You requested a record of type '${typeClass.modelName}' with id '${id}' but the adapter returned a payload with primary data having an id of '${payload.data.id}'. Looks like you want to use store.queryRecord instead http://emberjs.com/api/data/classes/DS.Store.html#method_queryRecord`, payload.data.id === id, {
```
bmac: I'm worried that out of the box `store.queryRecord` doesn't solve this problem out of the box and the documentation linked isn't doesn't point the user toward an actionable fix for the problem this is warning about.

pangratz: Hmm, I see your concern 😔 . But I don't see how this issue could be solved other than `queryRecord`. I will try to update the docs and this warning...

pangratz: See #4609

PR #4605, opened 2016-10-20
Don't notify belongsTo changes if nothing changed

### LINKED ISSUE #4579: querying existing record triggers a change in its belongsTo relationships (twiddle included)
On ember-data 2.7.0 and above, setting an observer on a `belongsTo` triggers without the relationship actually changing.

Here is the twiddle: https://ember-twiddle.com/e5fb5b69d401ca2187840a771b5d7785?fileTreeShown=false&numColumns=2&openFiles=controllers.application.js%2Ctemplates.application.hbs

When you click "reload", you can see a log on the console which shouldn't happen. The belongs to relationship never changed.
The log only happens if ember-data is set to a version above 2.7.

wecc:
Fixes #4579

### DIFF
--- a/tests/integration/records/relationship-changes-test.js
+++ b/tests/integration/records/relationship-changes-test.js
@@ -1,0 +1,108 @@
+import setupStore from 'dummy/tests/helpers/store';
+import Ember from 'ember';
+import DS from 'ember-data';
+import { module, test } from 'qunit';
+
+const { run } = Ember;
+const { attr, belongsTo, Model } = DS;
+
+const Author = Model.extend({
+  name: attr('string')
+});
+
+const Post = Model.extend({
+  author: belongsTo()
+});
+
+let env;
+
+module('integration/records/relationship-changes - Relationship changes', {
+  beforeEach() {
+    env = setupStore({
+      author: Author,
+      post: Post
+    });
+  },
+
+  afterEach() {
+    run(function() {
+      env.container.destroy();
+    });
+  }
+});
+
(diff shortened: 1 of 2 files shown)


BryanCrotaz:
Looks good

tests/integration/records/relationship-changes-test.js
```diff
@@ -0,0 +1,110 @@
...
+import {module, test} from 'qunit';
+
+import DS from 'ember-data';
+
+var env, store, Author, Post;
+var run = Ember.run;
+
+module('integration/records/relationship-changes - Relationship changes', {
+  beforeEach() {
+    Author = DS.Model.extend({
```
stefanpenner: Can we define these models at the module level, that way our tests don't have to constantly recreate these classes. This scenario likely doesn't affect the overall test run speed so much, but in aggregate they really do add up

Exelord:
@wecc 
Hey.
What about this spec?
```js
test('Calling push with hasMany relationship does trigger observer', function(assert) {
  assert.expect(1);

  let observerCount = 0;

  run(function() {
    let post = env.store.push({
      data: {
        id: '2',
        type: 'author'
      }
    });

    post.addObserver('author.posts', function() {
      observerCount++;
    });

    env.store.push({
      data: {
        type: 'post',
        id: '1',
        relationships: {
          author: {
            data: { type: 'author', id: '2' }
          }
        }
      }
    });
  });

  assert.equal(observerCount, 1, 'posts observer should be triggered');
});
```

after yours changes this case just stopped working :/

Exelord:
Sorry guys, this bug is associated with: #4850

PR #4607, opened 2016-10-20
Fix dummy app serializer

igorT:
We were not correctly deserializing relationships coming from the mock server,
because we try to dasherize by default.


igorT:
@runspired ^^

PR #4612, opened 2016-10-21
[DOC] Mark `errorsHashToArray` and `errorsArrayToHash` public

pangratz:
This methods are already present on the `DS` namespace, so this change
marks the officially as public and adds documentation how they can be
used.


pangratz:
@bmac thanks. Done.

PR #4616, opened 2016-10-21
Assert when original and not normalized key is found in payload

### LINKED ISSUE #4253: JSONAPISerializer should warn when not dasherized key names
refs https://github.com/emberjs/data/issues/4252

This seems to be a common source of confusion. I wonder if we can warn for this in development mode.

wecc:
ping @igorT 

Fixes #4253


pangratz:
Related issue #4253

bmac:
👍

PR #4617, opened 2016-10-22
chore(store): cleans up and optimizes code paths around fetching rela…

runspired:
…tionships and single records

Also underscores some store methods that are private that should have been before.

### DIFF
--- a/addon/-private/system/model/internal-model.js
+++ b/addon/-private/system/model/internal-model.js
@@ -709,11 +709,11 @@ export default class InternalModel {
     Preloaded data can be attributes and relationships passed in either as IDs or as actual
     models.
 
-    @method _preloadData
+    @method preloadData
     @private
     @param {Object} preload
   */
-  _preloadData(preload) {
+  preloadData(preload) {
     //TODO(Igor) consider the polymorphic case
     Object.keys(preload).forEach((key) => {
       let preloadValue = get(preload, key);
--- a/addon/-private/system/relationships/state/has-many.js
+++ b/addon/-private/system/relationships/state/has-many.js
@@ -130,9 +130,7 @@ ManyRelationship.prototype.reload = function() {
     this._loadingPromise = promiseManyArray(this.fetchLink(), 'Reload with link');
     return this._loadingPromise;
   } else {
-    this._loadingPromise = promiseManyArray(this.store.scheduleFetchMany(manyArray.toArray()).then(() => {
-      return manyArray;
-    }), 'Reload with ids');
+    this._loadingPromise = promiseManyArray(this.store._scheduleFetchMany(manyArray.currentState).then(() => manyArray), 'Reload with ids');
     return this._loadingPromise;
   }
 };
--- a/addon/-private/system/store.js
+++ b/addon/-private/system/store.js
@@ -49,7 +49,7 @@ const {
   get,
   isNone,
   isPresent,
-  Map,
+  MapWithDefault,
   run: emberRun,
   set,
   Service
@@ -212,7 +212,7 @@ Store = Service.extend({
     this._instanceCache = new ContainerInstanceCache(getOwner(this), this);
 
     //Used to keep track of all the find requests that need to be coalesced
-    this._pendingFetch = Map.create();
+    this._pendingFetch = MapWithDefault.create({ defaultValue() { return []; } });
   },
 
   /**
@@ -664,7 +664,7 @@ Store = Service.extend({
   _findRecord(internalModel, options) {
     // Refetch if the reload option is passed
     if (options.reload) {
-      return this.scheduleFetch(internalModel, options);
+      return this._scheduleFetch(internalModel, options);
     }
 
     var snapshot = internalModel.createSnapshot(options);
@@ -673,7 +673,7 @@ Store = Service.extend({
 
     // Refetch the record if the adapter thinks the record is stale
     if (adapter.shouldReloadRecord(this, snapshot)) {
-      return this.scheduleFetch(internalModel, options);
+      return this._scheduleFetch(internalModel, options);
     }
 
     if (options.backgroundReload === false) {
@@ -682,7 +682,7 @@ Store = Service.extend({
 
     // Trigger the background refetch if backgroundReload option is passed
     if (options.backgroundReload || adapter.shouldBackgroundReloadRecord(this, snapshot)) {
-      this.scheduleFetch(internalModel, options);
+      this._scheduleFetch(internalModel, options);
     }
 
     // Return the cached record
@@ -693,7 +693,7 @@ Store = Service.extend({
     options = options || {};
 
     if (options.preload) {
-      internalModel._preloadData(options.preload);
+      internalModel.preloadData(options.preload);
     }
 
     var fetchedInternalModel = this._findEmptyInternalModel(internalModel, options);
@@ -703,7 +703,7 @@ Store = Service.extend({
 
   _findEmptyInternalModel(internalModel, options) {
     if (internalModel.isEmpty()) {
-      return this.scheduleFetch(internalModel, options);
+      return this._scheduleFetch(internalModel, options);
     }
 
     //TODO double check about reloading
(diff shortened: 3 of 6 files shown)


addon/-private/system/relationships/state/has-many.js
```diff
@@ -130,9 +130,7 @@ ManyRelationship.prototype.reload = function() {
     this._loadingPromise = promiseManyArray(this.fetchLink(), 'Reload with link');
     return this._loadingPromise;
   } else {
-    this._loadingPromise = promiseManyArray(this.store.scheduleFetchMany(manyArray.toArray()).then(() => {
-      return manyArray;
-    }), 'Reload with ids');
+    this._loadingPromise = promiseManyArray(this.store._scheduleFetchMany(manyArray.currentState).then(() => { return manyArray; }), 'Reload with ids');
```
pangratz: `() => manyArray`

igorT:
Seems good, I'll give it a once over once less tired

addon/-private/system/store.js
```diff
@@ -737,57 +737,50 @@ Store = Service.extend({
...
 
     internalModel.loadingData(promise);
 
-    if (!this._pendingFetch.get(typeClass)) {
-      this._pendingFetch.set(typeClass, [pendingFetchItem]);
+    if (!this._pendingFetch.get(modelClass)) {
+      this._pendingFetch.set(modelClass, [pendingFetchItem]);
     } else {
-      this._pendingFetch.get(typeClass).push(pendingFetchItem);
+      this._pendingFetch.get(modelClass).push(pendingFetchItem);
```
pangratz: This could be handled more elegantly using `MapWithDefault`...

addon/-private/system/model/internal-model.js
```diff
@@ -707,11 +707,11 @@ export default class InternalModel {
     Preloaded data can be attributes and relationships passed in either as IDs or as actual
     models.
 
-    @method _preloadData
+    @method preloadData
     @private
     @param {Object} preload
   */
-  _preloadData(preload) {
```
stefanpenner: lets keep the`_`

runspired: @stefanpenner I removed this one because InternalModel is "entirely private api" and this is used by other private modules, I believe that was our stated api naming goal?

addon/-private/system/store.js
runspired: It may also be better to just create a hash via modelName, I don't think we need to go all the way to a map here.

addon/-private/system/relationships/state/has-many.js
runspired: 👍

runspired:
@stefanpenner @pangratz cleaned up

addon/-private/system/store.js
```diff
@@ -737,57 +737,51 @@ Store = Service.extend({
...
+    let pendingFetchItem = {
+      internalModel,
+      resolver,
+      options
     };
-    var promise = resolver.promise;
+    let promise = resolver.promise;
 
     internalModel.loadingData(promise);
+    let cache = this._pendingFetch.getWithDefault();
```
pangratz: What does this?

addon/-private/system/store.js
pangratz: Now that we're using `MapWithDefault`, this can be shortened to `this._pendingFetch.get(modelClass).push(pendingFetchItem)`.

addon/-private/system/store.js
runspired: #facepalm fixed @pangratz

addon/-private/system/store.js
```diff
@@ -800,63 +789,67 @@ Store = Service.extend({
...
+    let store = this;
+    let adapter = store.adapterFor(modelClass.modelName);
+    let shouldCoalesce = !!adapter.findMany && adapter.coalesceFindRequests;
+    let totalItems = pendingFetchItems.length;
+    let internalModels = new Array(totalItems);
+    let seeking = new EmptyObject();
+
+    for (let i = 0; i < totalItems; i++) {
+      internalModels[i] = pendingFetchItems[i].internalModel;
+      seeking[internalModels[i].id] = pendingFetchItems[i];
```
pangratz: Do we want to cache `pendigFetchItems[i]` so the lookup only happens once? This would also allow to get the internalModel's `id` from the cached result without looking it up again.

addon/-private/system/store.js
```diff
@@ -800,63 +789,67 @@ Store = Service.extend({
...
+    }
 
     function _fetchRecord(recordResolverPair) {
-      recordResolverPair.resolver.resolve(store.fetchRecord(recordResolverPair.record, recordResolverPair.options)); // TODO adapter options
+      recordResolverPair.resolver.resolve(store._fetchRecord(recordResolverPair.internalModel, recordResolverPair.options)); // TODO adapter options
```
pangratz: Let's split this into separate statements for better readability.

addon/-private/system/store.js
```diff
@@ -800,63 +789,67 @@ Store = Service.extend({
...
-    function makeMissingRecordsRejector(requestedRecords) {
-      return function rejectMissingRecords(resolvedRecords) {
-        resolvedRecords = Ember.A(resolvedRecords);
-        var missingRecords = requestedRecords.reject((record) => resolvedRecords.includes(record));
-        if (missingRecords.length) {
-          warn('Ember Data expected to find records with the following ids in the adapter response but they were missing: ' + Ember.inspect(Ember.A(missingRecords).mapBy('id')), false, {
-            id: 'ds.store.missing-records-from-adapter'
-          });
+      // reject missing records
+      let missingRecords = [];
```
pangratz: Rename to `missingInternalModels` to better reflect what they are containig

addon/-private/system/store.js
```diff
@@ -800,63 +789,67 @@ Store = Service.extend({
...
+      for (let i = 0, l = expectedInternalModels.length; i < l; i++) {
+        if (!found[expectedInternalModels[i].id]) {
+          missingRecords.push(expectedInternalModels[i]);
```
pangratz: Cache `expectedInternalModels[i]`?

addon/-private/system/store.js
```diff
@@ -800,63 +789,67 @@ Store = Service.extend({
...
-      };
+      }
     }
 
-    function makeRecordsRejector(records) {
-      return function (error) {
-        rejectRecords(records, error);
-      };
-    }
+    function rejectRecords(internalModels, error) {
```
pangratz: `rejectInternalModels`

addon/-private/system/store.js
```diff
@@ -867,27 +860,43 @@ Store = Service.extend({
...
-        var requestedRecords = Ember.A(groupOfRecords);
-        var ids = requestedRecords.mapBy('id');
-        if (ids.length > 1) {
-          _findMany(adapter, store, typeClass, ids, requestedRecords).
-            then(resolveFoundRecords).
-            then(makeMissingRecordsRejector(requestedRecords)).
-            then(null, makeRecordsRejector(requestedRecords));
+        for (let j = 0; j < totalInGroup; j++) {
+          groupedInternalModels[j] = group[j]._internalModel;
+          ids[j] = groupedInternalModels[j].id;
```
pangratz: Cache `group[j]._internalModel`

addon/-private/system/store.js
runspired: I debated on this, the alternative is:

```
let pendingItem = pendingFetchItems[i];
let internalModel = pendingItem.internalModel;
internalModels[i] = internalModel;
seeking[internalModel.id] = pendingItem;
```

This may be more clear but I'm not sure the "cache" actually is any simpler.

addon/-private/system/store.js
runspired:
```
recordResolverPair.resolver.resolve(
  store._fetchRecord(recordResolverPair.internalModel,
  recordResolverPair.options)
); // TODO adapter options
```

addon/-private/system/store.js
runspired: 👍

addon/-private/system/store.js
runspired: same argument as above, I always go back and forth on this. Saving one index access vs one assignment. Perf wise it's likely ignorable, but readability is a bit better.

addon/-private/system/store.js
runspired: 👍

addon/-private/system/store.js
runspired: 👍

addon/-private/system/store.js
pangratz: I meant something like:

```js
let fetchRecord = store._fetchRecord(…);
recordResolverPair.resolver.resolve(fetchRecord);
```

addon/-private/system/store.js
pangratz: I would be in favor of better readability here

addon/-private/system/store.js
```diff
@@ -737,58 +737,47 @@ Store = Service.extend({
...
+  _scheduleFetch(internalModel, options) {
     if (internalModel._loadingPromise) { return internalModel._loadingPromise; }
 
-    var resolver = Ember.RSVP.defer('Fetching ' + typeClass + 'with id: ' + internalModel.id);
-    var pendingFetchItem = {
-      record: internalModel,
-      resolver: resolver,
-      options: options
+    let modelClass = internalModel.type;
+    let resolver = Ember.RSVP.defer('Fetching ' + modelClass.modelName + 'with id: ' + internalModel.id);
```
pangratz: Missing space after `modelClass.modelName + 'with …'`

runspired:
@pangratz updated

runspired:
@bmac rebased

PR #4623, opened 2016-10-22
[BUGFIX beta] Don't invoke findRecord for a new record with client side id

### LINKED ISSUE #3678: shouldBackgroundReloadRecord is background reloading unsaved records
When a record is created and we transition to that route, Ember data does a background reload that results in a 404 from the database and the exception.

``` javascript
    actions: {
        newForm: function () {
            var form = this.store.createRecord("form-design", {
                name: "NewForm",
                description: "Created but not saved",
                rev: "",
                formObjects: []
            });
            this.transitionToRoute("form-design", form.get("id"));
        }
    }
});
```

In order to inhibit the background reload we put a "isNew" check into our adapter.

``` javascript
shouldBackgroundReloadRecord: function (store, snapshot) {
    // We don"t ever want to background reload a new record do we?
    return !snapshot.record.get("isNew");
}
```

I can't think of a use case where ember data would want to do a background reload on records that are not saved.  

Ember data should inhibit background reloading by default for unsaved records.

pangratz:
Addresses #3678

---

Do we want to warn when a `store.findRecord(modelName, newRecord.get("id"))` is called? Something like `You called store.findRecord to find a new record with a client side id. Use the record directly as store.findRecord should not be used for new records`?

### DIFF
--- a/addon/-private/system/store.js
+++ b/addon/-private/system/store.js
@@ -658,6 +658,12 @@ Store = Service.extend({
   },
 
   _findRecord(internalModel, options) {
+    // we can reach this code path when a new record is created with a client
+    // side id and this record is requested via `#findRecord(modelName, newRecord.get('id'))
+    if (internalModel.isNew()) {
+      return Promise.resolve(internalModel);
+    }
+
     // Refetch if the reload option is passed
     if (options.reload) {
       return this.scheduleFetch(internalModel, options);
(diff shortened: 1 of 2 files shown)


bmac:
👍

igorT:
We are closing this in favor of doing this in the adapter instead.

PR #4283, opened 2016-03-28
Add better error messaging for adapters that do not implement createR…

bmac:
…ecord, updateRecord or deleteRecord.

PR #4624, opened 2016-10-23
Test assertions when updateRecord & deleteRecord don't exist on adapter

pangratz:
Followup to #4283 to add tests for asserting that `updateRecord` and `deleteRecord` exist.

cc @bmac

### DIFF
--- a/tests/integration/adapter/store-adapter-test.js
+++ b/tests/integration/adapter/store-adapter-test.js
@@ -24,6 +24,15 @@ var set = Ember.set;
 var run = Ember.run;
 var Person, Dog, env, store, adapter;
 
+function moveRecordOutOfInFlight(record) {
+  run(function() {
+    // move record out of the inflight state so the tests can clean up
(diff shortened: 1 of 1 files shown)


bmac:
👍

tests/integration/adapter/store-adapter-test.js
```diff
@@ -24,6 +24,15 @@ var set = Ember.set;
 var run = Ember.run;
 var Person, Dog, env, store, adapter;
 
+function moveRecordOutOfInFlight(record) {
+  run(function() {
+    // move record out of the inflight state so the tests can clean up
+    // correctly
```
pangratz: It's because of [this assertion](https://github.com/emberjs/data/blob/41193f72435eca264ff3941134d2af1c72f6a861/addon/-private/system/model/states.js#L433-L437) in the state machine.

bmac:
Thanks @pangratz.

PR #4632, opened 2016-10-31
[DOC ds-extended-errors] Add docs for extended errors

lindyhopchris:
Adds documentation for `DS.AdapterError` and the sub-classes added in the `ds-extended-errors` feature.

Relates to #3586 and #3585 

@bmac have tried to add some example usages in as well. Not totally sure if this is what you have in mind, or if the usages are what Ember Data has in mind, so feel free to comment freely!!


bmac:
Thanks @lindyhopchris.

joelpresence:
How do you turn these errors on?  Setting `ds-extended-errors` in environment config has no effect ...

PR #4637, opened 2016-11-09
Simplify Adapter#serialize by using snapshot.serialize() since they d…

bmac:
…uplicate the same logic

cc @wecc

### DIFF
--- a/addon/adapter.js
+++ b/addon/adapter.js
@@ -3,7 +3,6 @@
 */
 
 import Ember from 'ember';
-var get = Ember.get;
 
 /**
   An adapter is an object that receives requests from a store and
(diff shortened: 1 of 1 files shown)


sly7-7:
After a quick search about adapter.serialize, it seems like it's never called in the repository. Maybe we could even deprecate it and remove it for 3.0 ?

bmac:
That sounds good to me @sly7-7. I think the same thing could happen for `store#serialize`

sly7-7:
@bmac I just wonder what we want as a public api for serializing a record. Do we want only keep `record.serialize(options)` ?

bmac:
I think `record.serialize(options)` and `snapshot.serialize(options)` is the way to go.

PR #4641, opened 2016-11-11
Add API docs for the RecordReference

bmac:


addon/-private/system/references/record.js
```diff
@@ -11,28 +18,140 @@ RecordReference.prototype = Object.create(Reference.prototype);
...
+
+   @method id
+   @return {String} The id of the record in this belongsTo relationship.
+*/
 RecordReference.prototype.id = function() {
   return this._id;
 };
 
+/**
+   How the reference will be looked up with it is loaded:
```
pangratz: ...up with _when_ it is loaded:... ?

addon/-private/system/references/record.js
```diff
@@ -11,28 +18,140 @@ RecordReference.prototype = Object.create(Reference.prototype);
...
+
+   * `link`, a URL
```
pangratz: Hmm, I believe that this is not supported at the moment. At least I am not aware of this... What am I missing?

addon/-private/system/references/record.js
```diff
@@ -11,28 +18,140 @@ RecordReference.prototype = Object.create(Reference.prototype);
...
+  For example, `recordReference.push(promise)` will be resolved with a
+  record.
+
+   Example
+
+   ```javascript
+   var userRef = store.getReference('user', 1);
+
+   // provide data for reference
+   userRef.push({ id: 1, username: "@user" }).then(function(user) {
```
pangratz: The data passed to `push` needs to be a JSON-API doc with `{ data: { ... } } `

addon/-private/system/references/record.js
```diff
@@ -11,28 +18,140 @@ RecordReference.prototype = Object.create(Reference.prototype);
...
+*/
 RecordReference.prototype.push = function(objectOrPromise) {
   return Ember.RSVP.resolve(objectOrPromise).then((data) => {
     return this.store.push(data);
   });
 };
 
+/**
+  If the entity referred to by the reference is already loaded, it is
+  present as `reference.value`. Otherwise, the value of this property
```
pangratz: I would change this to: Otherwise the value returned by this function is `null`

locks:
@pangratz can you re-review?

pangratz:
Thanks again @bmac

PR #4643, opened 2016-11-11
[BUGFIX #4649 #4648]

### LINKED ISSUE #4648: HasMany.addRecord notifies more then once
within `ManyRelationship.prototype.addRecord`:
 
we do duplicate work, as
* https://github.com/emberjs/data/blob/master/addon/-private/system/relationships/state/has-many.js#L67
* https://github.com/emberjs/data/blob/master/addon/-private/system/relationships/state/has-many.js#L69

both ultimately notify `notifyHasManyChanged` [here](https://github.com/emberjs/data/blob/master/addon/-private/system/relationships/state/relationship.js#L168) and [here](https://github.com/emberjs/data/blob/aa3a3d5dee27b641f9e6bb831c0d85ecf9730fd6/addon/-private/system/many-array.js#L178)

stefanpenner:
* no longer invalid relationships property when accessing it the first time #4649
* no longer duplicate  “notifyHasManyChange” when adding a record fixes #4648

No longer warn/error in glimmer 2 by accidentally causing duplicate work.

### DIFF
--- a/addon/-private/system/many-array.js
+++ b/addon/-private/system/many-array.js
@@ -129,7 +129,7 @@ export default Ember.Object.extend(Ember.MutableArray, Ember.Evented, {
     this.relationship = this.relationship || null;
 
     this.currentState = Ember.A([]);
-    this.flushCanonical();
+    this.flushCanonical(false);
   },
 
   objectAt(index) {
@@ -141,7 +141,7 @@ export default Ember.Object.extend(Ember.MutableArray, Ember.Evented, {
     return this.currentState[index].getRecord();
   },
 
-  flushCanonical() {
+  flushCanonical(isInitialized = true) {
     let toSet = this.canonicalState;
 
     //a hack for not removing new records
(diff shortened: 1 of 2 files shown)


addon/-private/system/relationships/has-many.js
```diff
@@ -164,8 +164,11 @@ export const HasManyMixin = Ember.Mixin.create({
     //We need to notifyPropertyChange in the adding case because we need to make sure
     //we fetch the newly added record in case it is unloaded
     //TODO(Igor): Consider whether we could do this only if the record state is unloaded
-
     //Goes away once hasMany is double promisified
-    this.notifyPropertyChange(key);
+    //
+    let cache = Ember.meta(this).readableCache();
+    if (typeof cache === 'object' && cache !== null && cache[key] !== undefined) {
```
stefanpenner: This is just a quick experiment, don't mind the lolcode it is not currently my intention to merge this as is.

bmac:
👍 looks good to me.

runspired:
For future note, dropping this line from `ManyArray.internalReplace` changed an undocumented and unintentional behavior that was public facing: https://github.com/emberjs/data/commit/b3e1d67899b06b051a0cc2c12635998215ad6621#diff-f2ea099eb6bd7f8ec1eeff2bd7a32c6bL178

```
this.relationship.notifyHasManyChanged();	
```

Before this, computed properties that subscribed to the relationship key (e.g. `record.foos` not `records.foos.length` or `record.foos.[]`) would dirty when internalReplace was called.  While this is a bug and a performance issue we should fix, I believe this behavior change should go into the changelog. cc @bmac

PR #4644, opened 2016-11-11
[PERF] HasMany doesn't need to inform RecordArrayManager

stefanpenner:
At one point, relationships were managed by the recordArrayManager, this is no longer true. But we still spuriously inform the manager when we flush a relationship


stefanpenner:
It may be that this is actually required https://github.com/emberjs/data/issues/4646 sorta implies that it may be. But if thats the case, we need better tests. We can restore if #4646 ends up informing that this should happen, but \w tests and performantly.

PR #4645, opened 2016-11-11
cleanup system/many-array

stefanpenner:
* ensure shape is stable
* use tidier ES6 syntax when appropriate


fivetanley:
lgtm when travis comes back

PR #4653, opened 2016-11-14
Update the API docs on the JSONAPISerializer

bmac:


addon/serializers/json-api.js
```diff
@@ -391,6 +374,12 @@ const JSONAPISerializer = JSONSerializer.extend({
   },
 
   /**
+    Dasherizes and singularizes the model name in the payload to match
+    the format Ember Data uses internally for the model name.
+
+    For Example: `posts` => `post`, `studentAssesments` =>
```
locks: I believe yuidoc has an `@example` metatag, can you confirm?

locks: Actually, that might not makes much sense. Could you instead write it out in prose? the `=>` is a bit confusing as it's also the syntax for arrow functions.

addon/serializers/json-api.js
```diff
@@ -401,6 +390,11 @@ const JSONAPISerializer = JSONSerializer.extend({
   },
 
   /**
+    Converts the model name to a pluralized version of the model name.
+
+    For Example: `post` => `posts`, `student-assesment` =>
```
locks: See above.

PR #4654, opened 2016-11-15
remove useless private api

sly7-7:
This follows https://github.com/emberjs/data/pull/4637
cc @bmac

### DIFF
--- a/FEATURES.md
+++ b/FEATURES.md
@@ -146,3 +146,7 @@ entry in `config/features.json`.
      }
    });
   ```
+- `ds-deprecate-store-serialize` [#4654](https://github.com/emberjs/data/pull/4654)
+
+  Adds a deprecation warning when using Store#serialize(record) method.
+  You can use record.serialize() instead.
--- a/addon/-private/system/store.js
+++ b/addon/-private/system/store.js
@@ -253,10 +253,17 @@ Store = Service.extend({
 
     @method serialize
     @private
+    @deprecated
     @param {DS.Model} record the record to serialize
     @param {Object} options an options hash
   */
   serialize(record, options) {
+    if (isEnabled('ds-deprecate-store-serialize')) {
+      deprecate('Use of store.serialize is deprecated, use record.serialize instead.', false, {
+        id: 'ds.store.serialize',
+        until: '3.0'
+      });
+    }
     let snapshot = record._internalModel.createSnapshot();
     return snapshot.serialize(options);
   },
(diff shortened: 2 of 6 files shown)


addon/-private/system/store.js
```diff
@@ -233,25 +233,6 @@ Store = Service.extend({
...
-
-    * `includeId`: `true` if the record's ID should be included in
-      the JSON representation
-
-    @method serialize
-    @private
-    @param {DS.Model} record the record to serialize
-    @param {Object} options an options hash
-  */
-  serialize(record, options) {
```
bmac: @sly7-7  do you mind deprecating this instead of removing it? Just in case someone was using it.

pangratz:
since this is a public facing change, I think this should be put behind a feature flag.

@sly7-7 can you update this PR to also include a test for the deprecation? Merci beaucoup!

sly7-7:
@pangratz unless I'm missing something, this method is marked `@private`, so there is no need to feature flag. I will add the test 😄 , I was too lazy to add one, considering the nature of the change.

pangratz:
Basically you're right. But since we're introducing a public facing deprecation, I think it would be good to put this behind a feature flag. If we - despite being private - see that too many applications use this in the wild, we can simply disable the feature.

I think the goal is to use feature flags for all public facing changes, even if they are very small.

sly7-7:
feature flag and test added 😄

addon/-private/system/store.js
```diff
@@ -253,10 +253,17 @@ Store = Service.extend({
 
     @method serialize
     @private
+    @deprecated
     @param {DS.Model} record the record to serialize
     @param {Object} options an options hash
   */
   serialize(record, options) {
+    if (isEnabled('ds-deprecate-store-serialize')) {
+      deprecate('Use of serialize is deprecated, use record.serialize instead.', false, {
```
sly7-7: no pb 😄

FEATURES.md
```diff
@@ -146,3 +146,6 @@ entry in `config/features.json`.
      }
    });
   ```
+- `ds-deprecate-store-serialize` [#4654](https://github.com/emberjs/data/pull/4654)
+
+  Adds a deprecation warning when using Store#serialize method
```
pangratz: Let's also add a short sentence on what should be used instead. Just so people looking at the `FEATURES.md` immediately know what this is about.

sly7-7:
@pangratz I'm a little bit 😞  this takes so much reviews. I hope this is ok now 😄

pangratz:
🎉

PR #4657, opened 2016-11-15
Deprecate unused initializers

### LINKED ISSUE #4093: Deprecate unused initializers
Ember Data has 4 initializers that it supports for legacy reasons but otherwise do nothing. https://github.com/emberjs/data/blob/master/tests/ember-data-initializers.js#L18-L40

@rwjblue [suggested](https://github.com/emberjs/data/pull/4088#issuecomment-172663110) that it may be possible to deprecate them by "hook[ing] into the DAG map that is used for initializers and see if anyone depends on it"

bmac:
Closes #4093 

- [ ] Deprecation RFC https://github.com/emberjs/rfcs/pull/181

### DIFF
--- a/addon/-private/instance-initializers/initialize-store-service.js
+++ b/addon/-private/instance-initializers/initialize-store-service.js
@@ -1,12 +1,53 @@
+import { deprecate } from 'ember-data/-private/debug';
 /*
- Configures a registry for use with an Ember-Data
- store.
+  Configures a registry for use with an Ember-Data
+  store.
 
- @method initializeStoreService
- @param {Ember.ApplicationInstance} applicationOrRegistry
- */
+  @method initializeStoreService
+  @param {Ember.ApplicationInstance} applicationOrRegistry
+*/
 export default function initializeStoreService(application) {
   var container = application.lookup ? application : application.container;
   // Eagerly generate the store so defaultStore is populated.
   container.lookup('service:store');
+
+  let initializers = application.application.constructor.initializers;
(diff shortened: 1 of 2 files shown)


stefanpenner:
👍

app/initializers/data-adapter.js
```diff
@@ -10,5 +10,5 @@ import Ember from 'ember';
 export default {
   name: 'data-adapter',
   before: 'store',
-  initialize: Ember.K
```
locks: Rebase needed?

bmac:
rebased @locks

addon/-private/instance-initializers/initialize-store-service.js
```diff
@@ -1,12 +1,53 @@
...
   // Eagerly generate the store so defaultStore is populated.
   container.lookup('service:store');
+
+  let initializers = application.application.constructor.initializers;
+  deprecateOldEmberDataInitializers(initializers)
+}
+
+
+
+let deprecatedInitializerNames = ['data-adapter', 'injectStore', 'transforms', 'store'];
```
bmac: `store` should not be included in this list.

bmac: Nevermind. This code is correct. I got `store` confused with the `ember-data` initializer.

bmac:
I was given feedback that this should go through an RFC so there is an official document on why this has been deprecated.

https://github.com/emberjs/rfcs/pull/181

PR #4658, opened 2016-11-16
Test for hasMany array removal on deletion

runspired:
Bumped into an edge case in our app and built this test as a sanity check, figured I'd leave it for us to have nice things.


bmac:
Thanks @runspired.

PR #4661, opened 2016-11-16
dirty check to prevent liveRecordArrays being rebuilt too often

runspired:
Every call to `peekAll` currently forces us to completely rebuild the live array. This is a stop gap until recordArrayManager can be improved ( @stefanpenner and I have been brainstorming our way through how to improve it.).

### DIFF
--- a/addon/-private/system/record-array-manager.js
+++ b/addon/-private/system/record-array-manager.js
@@ -72,7 +72,7 @@ export default Ember.Object.extend({
     });
 
     this.liveRecordArrays = MapWithDefault.create({
-      defaultValue: typeClass => this.createRecordArray(typeClass)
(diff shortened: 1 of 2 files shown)


runspired:
@bmac test failure looks to just be a random appveyor crash, I don't think I'm allowed to restart it though.

bmac:
I think @igorT is the only one who can restart appveyor. When travis is green, I've been ignoring appveyor crashes due to network errors like that and merging with with a failing appveyor build if the pr doesn't make any changes to Ember Data's build.

The appveyor check is mostly to make sure windows developers can contribute to Ember Data. So there is a very low risk of breaking the experience for windows devs if changes don't add new dependencies or change the build logic.

PR #4662, opened 2016-11-17
Code Cleanup for store.js

runspired:
I've been working in this file a lot and it got tedious needing to update just small parts of the code formatting each time, so I did a little bit more in bulk.

### DIFF
--- a/addon/-private/system/store.js
+++ b/addon/-private/system/store.js
@@ -44,16 +44,27 @@ import isEnabled from 'ember-data/-private/features';
 export let badIdFormatAssertion = '`id` passed to `findRecord()` has to be non-empty string or number';
 
 const {
+  A,
   _Backburner: Backburner,
+  computed,
   copy,
+  ENV,
+  Error: EmberError,
   get,
+  guidFor,
+  inspect,
   isNone,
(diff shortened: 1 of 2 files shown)


bmac:
Thanks @runspired.

addon/-private/system/store.js
```diff
@@ -1812,15 +1828,15 @@ Store = Service.extend({
...
     pending.forEach((pendingItem) => {
-      var snapshot = pendingItem.snapshot;
-      var resolver = pendingItem.resolver;
-      var record = snapshot._internalModel;
-      var adapter = this.adapterFor(record.type.modelName);
-      var operation;
+      let snapshot = pendingItem.snapshot;
+      let resolver = pendingItem.resolver;
+      let record = snapshot._internalModel;
+      let adapter = this.adapterFor(record.modelClass.modelName);
```
runspired: @bmac I'll fix this and add a failing test, we apparently don't test this. Record.modeClass doesn't exist (yet)

runspired: Actually disregard, this is an internalModel and forward thinking me has already made that work.

PR #4664, opened 2016-11-17
Upgraded IdentityMap and RecordMap

runspired:
When pushing records into the store (no materialization), we only needed to load the associated modelClass for "key" purposes. This PR shifts ember-data to consistently using a normalized `modelName` as the key, making it possible to push internal models, assemble record arrays, and (potentially in the future) even make requests without any access to the modelClass.

Step two of this effort (a separate PR) will be to enable the parsing of `modelClass` to avoid needing to load the classes for related models.  Combined, this gives us a happy story for initial render in apps with pre-loaded data and complicated relationship graphs.

In addition, this paved an easier path forward for #4584

- [x] refactor typeMap => recordMap
- [x] refactor to using modelName instead of guid as the class identifier for recordMap.
- [x] refactor InternalModel to not need `modelClass` at instantiation.
- [x] remove calls to `modelFor` where now possible
- [x] remove `typeMapFor`
- [x] consider deprecation notices for `InternalModel` and `buildInternalModel` in case private APIs were abusing them and now need to refactor from `modelClass` to `modelName`.

- (will be separate PR, but needed to complete this effort) refactor relationship setup to not need to lookup the related modelClass to prevent "pulling the string" on the entire graph.

**UPDATES**

I believe this is ready to ship but want to document a few things:

- `ClassMeta` is not unit tested (yet) but it is implicitly tested as it exposes an identical interface to the pojo that was our `TypeMap` before.
- @stefanpenner rightly points out areas we should be doing more to encapsulate things in this class, but this should be a follow up PR (that also moves to direct unit testing)
- A temporary compromise has been made to give `store` to `new ClassMeta(store)` instead of `owner` in order to support the ability for it to resolve the `modelClass` via `store.modelFor`.  Longer term this should be `owner` `store.modelFor` should instead use `classMeta.modelClass` to resolve the modelClass if necessary.
- with this PR, `_push` does not require `modelClass` at all. The relationship PR is related to this "lazy push" story but is not required for it.

### DIFF
--- a/addon/-private/system/identity-map.js
+++ b/addon/-private/system/identity-map.js
@@ -1,0 +1,49 @@
+import RecordMap from './record-map';
+
+/**
+ `IdentityMap` is a custom storage map for records by modelName
+ used by `DS.Store`.
+
+ @class IdentityMap
+ @private
+ */
+export default class IdentityMap {
+  constructor() {
+    this._map = Object.create(null);
+  }
+
+  /**
+   Retrieves the `RecordMap` for a given modelName,
+   creating one if one did not already exist. This is
+   similar to `getWithDefault` or `get` on a `MapWithDefault`
+
+   @method retrieve
+   @param modelName a previously normalized modelName
+   @returns {RecordMap} the RecordMap for the given modelName
+   */
+  retrieve(modelName) {
+    let recordMap = this._map[modelName];
+
+    if (!recordMap) {
+      recordMap = this._map[modelName] = new RecordMap(modelName);
+    }
+
+    return recordMap;
+  }
+
+  /**
+   Clears the contents of all known `RecordMaps`, but does
+   not remove the RecordMap instances.
+
+   @method clear
+   */
+  clear() {
+    let recordMaps = this._map;
+    let keys = Object.keys(recordMaps);
+
+    for (let i = 0; i < keys.length; i++) {
+      let key = keys[i];
+      recordMaps[key].clear();
+    }
+  }
+}
--- a/addon/-private/system/model/internal-model.js
+++ b/addon/-private/system/model/internal-model.js
@@ -87,6 +87,8 @@ const {
   'updateChangedAttributes'
 );
 
+let InternalModelReferenceId = 1;
+
 /*
   `InternalModel` is the Model class that we use internally inside Ember Data to represent models.
   Internal ED methods should only deal with `InternalModel` objects. It is a fast, plain Javascript class.
@@ -104,13 +106,13 @@ const {
   @class InternalModel
 */
 export default class InternalModel {
-  constructor(modelClass, id, store, data) {
+  constructor(modelName, id, store, data) {
     heimdall.increment(new_InternalModel);
-    this.modelClass = modelClass;
     this.id = id;
+    this._internalId = InternalModelReferenceId++;
     this.store = store;
     this._data = data || new EmptyObject();
-    this.modelName = modelClass.modelName;
+    this.modelName = modelName;
     this.dataHasInitialized = false;
     this._loadingPromise = null;
     this._recordArrays = undefined;
@@ -122,6 +124,7 @@ export default class InternalModel {
     this.error = null;
 
     // caches for lazy getters
+    this._modelClass = null;
     this.__deferredTriggers = null;
     this._references = null;
     this._recordReference = null;
@@ -129,6 +132,10 @@ export default class InternalModel {
     this.__relationships = null;
     this.__attributes = null;
     this.__implicitRelationships = null;
+  }
+
+  get modelClass() {
+    return this._modelClass || (this._modelClass = this.store.modelFor(this.modelName));
   }
 
   get type() {
@@ -778,7 +785,7 @@ export default class InternalModel {
   */
   updateRecordArrays() {
     this._updatingRecordArraysLater = false;
-    this.store.dataWasUpdated(this.modelClass, this);
+    this.store._dataWasUpdated(this);
   }
 
   setId(id) {
--- a/addon/-private/system/record-arrays/record-array.js
+++ b/addon/-private/system/record-arrays/record-array.js
@@ -5,11 +5,10 @@
 import Ember from 'ember';
 import { PromiseArray } from "ember-data/-private/system/promise-proxies";
 import SnapshotRecordArray from "ember-data/-private/system/snapshot-record-array";
-
-const { get, set, RSVP: { Promise } } = Ember;
+const { computed, get, set, RSVP: { Promise } } = Ember;
 
 /**
-  A record array is an array that contains records of a certain type. The record
+  A record array is an array that contains records of a certain modelName. The record
   array materializes records as needed when they are retrieved for the first
   time. You should not create record arrays yourself. Instead, an instance of
   `DS.RecordArray` or its subclasses will be returned by your application's store
@@ -24,14 +23,6 @@ const { get, set, RSVP: { Promise } } = Ember;
 export default Ember.ArrayProxy.extend(Ember.Evented, {
   init() {
     this._super(...arguments);
-
-    /**
-      The model type contained by this record array.
-
-      @property type
-      @type DS.Model
-      */
-    this.type = this.type || null;
 
     /**
       The array of client ids backing the record array. When a
@@ -88,9 +79,21 @@ export default Ember.ArrayProxy.extend(Ember.Evented, {
   },
 
   replace() {
-    let type = get(this, 'type').toString();
-    throw new Error(`The result of a server query (for all ${type} types) is immutable. To modify contents, use toArray()`);
+    throw new Error(`The result of a server query (for all ${this.modelName} types) is immutable. To modify contents, use toArray()`);
   },
+
+  /**
+   The modelClass represented by this record array.
+
+   @property type
+   @type DS.Model
+   */
+  type: computed.readOnly('modelName', function() {
+    if (!this.modelName) {
+      return null;
+    }
+    return this.store.modelFor(this.modelName);
+  }),
 
   /**
     Retrieves an object from the content by index.
@@ -145,10 +148,7 @@ export default Ember.ArrayProxy.extend(Ember.Evented, {
     is finished.
    */
   _update() {
-    let store = get(this, 'store');
-    let modelName = get(this, 'type.modelName');
-
-    return store.findAll(modelName, { reload: true });
+    return this.store.findAll(this.modelName, { reload: true });
   },
 
   /**
@@ -193,9 +193,9 @@ export default Ember.ArrayProxy.extend(Ember.Evented, {
     @return {DS.PromiseArray} promise
   */
   save() {
-    let promiseLabel = 'DS: RecordArray#save ' + get(this, 'type');
-    let promise = Promise.all(this.invoke('save'), promiseLabel).
-      then(() => this, null, 'DS: RecordArray#save return RecordArray');
+    let promiseLabel = `DS: RecordArray#save ${this.modelName}`;
+    let promise = Promise.all(this.invoke('save'), promiseLabel)
+      .then(() => this, null, 'DS: RecordArray#save return RecordArray');
 
     return PromiseArray.create({ promise });
   },
--- a/addon/-private/system/record-arrays/record-array.js
+++ b/addon/-private/system/record-arrays/record-array.js
@@ -88,12 +88,12 @@ export default Ember.ArrayProxy.extend(Ember.Evented, {
    @property type
    @type DS.Model
    */
-  type: computed.readOnly('modelName', function() {
+  type: computed('modelName', function() {
     if (!this.modelName) {
       return null;
     }
     return this.store.modelFor(this.modelName);
-  }),
+  }).readOnly(),
 
   /**
     Retrieves an object from the content by index.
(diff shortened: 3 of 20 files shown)


runspired:
cc @stefanpenner got the first big part done, should have this PR done by EOD, hoping to at least have started the relationship PR that needs to accompany.  Would like thoughts on the new location of `ClassMeta`, especially as I'm envisioning this as a spot to refactor what was my intended use of `Schema`.

addon/-private/system/class-meta.js
```diff
@@ -0,0 +1,57 @@
...
+
+  clear() {
+    if (this._records) {
+      let records = this._records.slice();
+      let record;
+
+      for (let i = 0; i < records.length; i++) {
+        record = records[i];
+        record.unloadRecord();
+        record.destroy(); // maybe within unloadRecord
```
runspired: I copied this over from the existing unloadAll functionality, but I suspect this is actually buggy an we want to reset the length of this._records to `0` here. I left it as is with the guess that `unloadRecord` removes it from the array on it's own (which would also be why we slice).

addon/-private/system/class-meta.js
```diff
@@ -0,0 +1,57 @@
...
+  get modelClass() {
+    if (!this._modelClass) {
+      this._modelClass = this._store.modelFor(this.modelName);
+    }
+
+    return this._modelClass;
+  }
+
+  /* deprecated way of accessing modelClass */
+  get type() {
```
runspired: we still use "type" enough places that I haven't pulled the trigger on an actual deprecation notice yet, want to keep this quieter until we're closer to the end of internals cleanup.

stefanpenner:
I would love to review this thoroughly, although I may not have cycles until tomorrow or sat.

addon/-private/system/class-meta.js
```diff
@@ -0,0 +1,57 @@
...
+  get idToRecord() {
+    return this._idToRecord || (this._idToRecord = new EmptyObject());
+  }
+
+  get records() {
+    return this._records || (this._records = []);
+  }
+
+  get metadata() {
+    return this._metadata || (this._metadata = new EmptyObject());
```
stefanpenner: this may also be like above

addon/-private/system/class-meta.js
stefanpenner: this may be a bit zalgo, as if there is no `modelClass` via `type` call, this will be null.

addon/-private/system/class-meta.js
stefanpenner: ya, I think (but we should double check with someone else) we want to set `this._records.length = 0` just after  `let records = this._records.slice();`

addon/-private/system/model/internal-model.js
```diff
@@ -104,13 +104,12 @@ const {
   @class InternalModel
 */
 export default class InternalModel {
-  constructor(modelClass, id, store, data) {
+  constructor(modelName, id, store, data) {
     heimdall.increment(new_InternalModel);
```
stefanpenner: should the internalModel have a direct reference to its `classMeta` / `internalClass`, that way `get modelClass() ` can simply delegate to it.

addon/-private/system/store.js
```diff
@@ -958,13 +951,15 @@ Store = Service.extend({
...
-    @param {String} type
+    @param {String} modelName
     @param {String|Integer} id
     @since 2.5.0
     @return {RecordReference}
   */
-  getReference: function(type, id) {
-    return this._internalModelForId(type, id).recordReference;
+  getReference(modelName, id) {
+    modelName = this._classKeyFor(modelName);
```
stefanpenner: override argument value makes debugging annoying

stefanpenner:
@runspired given this PR, what remains re: loading modelClass spuriously during cache load?

addon/-private/system/model/internal-model.js
runspired: I agree but wanted to limit the scope of the initial PR

stefanpenner:
@emberjs/ember-data-contributors we would love more sets of eyes on this, we don't want to unilaterally sneak something like this in.

bmac:
I'm 👍  on the direction of this change.

addon/-private/system/class-meta.js
runspired: @bmac I think the issue raised is that if there is (literally) no modelClass to resolve, that it will always be null, and ergo we will perpetually keep trying to resolve it with every "get" of type or modelClass.

addon/-private/system/class-meta.js
```diff
@@ -0,0 +1,57 @@
+import EmptyObject from 'ember-data/-private/system/empty-object';
+
+export default class ClassMeta {
```
runspired: Outside of one new behavior, this object mirrors the pojo it replaces. Don't disagree about unit tests, I just suspect we already implicitly have a bunch that we should congregate and expand upon.

addon/-private/system/class-meta.js
```diff
@@ -0,0 +1,57 @@
...
+    this.modelName = modelName;
+    this._store = store;
+    this._modelClass = null;
+    this._idToRecord = null;
+    this._records = null;
+    this._metadata = null;
+  }
+
+  get idToRecord() {
+    return this._idToRecord || (this._idToRecord = new EmptyObject());
```
runspired: I'm on board with the new direction, I intentionally shaped this class to follow the pojo we were currently using as the typeMap in order to limit the scope of the changes. Can we agree that this is an immediate follow up PR? Or should I go ahead and do this refactor in this PR.

addon/-private/system/class-meta.js
```diff
@@ -0,0 +1,57 @@
...
+  }
+
+  get records() {
+    return this._records || (this._records = []);
```
runspired: unlike above, the way this one works we don't deal with an "what if always resolves to null?" situation.

addon/-private/system/class-meta.js
runspired: unlike above, the way this one works we don't deal with an "what if always resolves to null?" situation.

addon/-private/system/class-meta.js
```diff
@@ -0,0 +1,57 @@
...
+        record.unloadRecord();
+        record.destroy(); // maybe within unloadRecord
+      }
+    }
+
+    this._metadata = null;
+  }
+
+  destroy() {
+    this._store = null;
```
runspired: 👍

addon/-private/system/store.js
runspired: Agreed, hadn't thought about that, will refactor to use a new variable name. I'd also like to bring up (as a separate PR) that it would be nice to officially deprecate support for camelCase modelName usage.

addon/-private/system/store.js
```diff
@@ -1926,50 +1912,56 @@ Store = Service.extend({
...
+    heimdall.increment(_recordMapFor);
+    let recordMaps = this._recordMaps;
+    let recordMap = recordMaps[modelName];
 
-    typeMaps[guid] = typeMap;
+    if (recordMap) { return recordMap; }
 
-    return typeMap;
+    recordMap = new ClassMeta(this, modelName);
+    recordMaps[modelName] = recordMap;
```
runspired: I originally was going to pass in `owner`, but wanted to piggy-pack on the `modelFor` behavior (for now). I think longer term `modelFor` would move into this object itself.

addon/-private/system/store.js
```diff
@@ -2494,18 +2493,17 @@ Store = Service.extend({
...
-    let modelClass = internalModel.type;
-    let typeMap = this.typeMapFor(modelClass);
+    let recordMap = this._recordMapFor(internalModel.modelName);
     let id = internalModel.id;
 
     internalModel.updateRecordArrays();
 
     if (id) {
-      delete typeMap.idToRecord[id];
+      delete recordMap.idToRecord[id];
```
runspired: Yes.

addon/-private/system/store.js
```diff
@@ -2494,18 +2493,17 @@ Store = Service.extend({
...
     }
 
-    let loc = typeMap.records.indexOf(internalModel);
-    typeMap.records.splice(loc, 1);
+    let loc = recordMap.records.indexOf(internalModel);
```
runspired: Yes.

addon/-private/system/model/internal-model.js
stefanpenner: sounds reasonable

addon/-private/system/class-meta.js
stefanpenner: > Can we agree that this is an immediate follow up PR? Or should I go ahead and do this refactor in this PR.

Sounds reasonable to me.

addon/-private/system/class-meta.js
stefanpenner: Although implicitly tested, I suspect it will pay dividends to throughly unit test this. It will help keep us from trolling ourselves while we iterate quickly.

addon/-private/system/class-meta.js
stefanpenner: so its never inspected for empty-ness, always to append/mutate?

addon/-private/system/class-meta.js
stefanpenner: so its never inspected for empty-ness, always to append/mutate?

addon/-private/system/store.js
stefanpenner: seems fine, as long as we don't forget

addon/-private/system/store.js
stefanpenner: seems good.

runspired:
@stefanpenner with these changes, you no longer load a modelClass for anything not related to

- iterating attributes / relationships
- adapter calls
- serializer calls

Iteration currently only happens during create/update/delete data flows, so this means that for what we wanted as a first pass this achieves that.

stefanpenner:
@runspired nice, have you tested this in our app. I want to make sure we didn't forget anything obviously for those offending code-paths.

workmanw:
Not surprisingly, this does seem to break some ember-data-model-fragment tests. I'll look into getting that fixed once this lands. One question though, you said there is also a coming PR for part two of this effort. Do you plan to squeeze that into the same ember-data release? If so, I'll wait to investigate until the effort is complete.

stefanpenner:
@workmanw is there an effort to get model-fragments using public APIs and/or derive the needed public API's for model-fragments to be successful?

workmanw:
@stefanpenner Sorry I missed your response. To be perfectly honest, I wouldn't even know where to begin. I don't have a commanding knowledge of the ember-data internals. I'm not the only maintainer nor the original author, so I can't speak for everyone, but I would be open to trying to make this more durable.

The biggest challenges we have in ember-data-model-fragments are 1) we interact heavily with the internal model and 2) we have to monkey patch part of the ember-data public API. For example, we override `DS.Model.rollbackAttributes`, `DS.Model. createSnapshot`, `DS.model.adapterDidCommit`, etc. Most of the monkey patching is just trying to keep the state of the model and it's fragments in sync. 

If you have idea's on how ember-data's API could be evolved to help facilitate this, I'd love to hear them.

stefanpenner:
@workmanw a good start is thoroughly documenting all the things that require private API usage (and why). That way those more familiar with ember-data can help

addon/-private/system/class-meta.js
runspired: Basically this is a stable array unless it's a brand new meta or the meta has been "reset", in which case it's null

stefanpenner: sounds good

stefanpenner:
@runspired let me know when/if you ready for another code-review

addon/-private/system/record-arrays/record-array.js
```diff
@@ -88,11 +79,23 @@ export default Ember.ArrayProxy.extend(Ember.Evented, {
...
+
+   @property type
+   @type DS.Model
+   */
+  type: computed('modelName', function() {
+    if (!this.modelName) {
+      return null;
+    }
+    return this.store.modelFor(this.modelName);
+  }),
```
stefanpenner: for good reasons, or test bugs?

runspired: Fixed

addon/-private/system/identity-map.js
```diff
@@ -0,0 +1,49 @@
...
+  }
+
+  /**
+   Clears the contents of all known `RecordMaps`, but does
+   not remove the RecordMap instances.
+
+   @method clear
+   */
+  clear() {
+    let recordMaps = this._map;
```
stefanpenner: Do we need to actually clear these, our should we just "release" the dominator?

addon/-private/system/record-arrays/record-array.js
stefanpenner: This doesn't do what you think it is. (the last arg is dropped, but i suspect we want it to actually run)

I suspect you want:

```js
type: computed('modelName', function() {
  // ...
}).readOnly()
```

runspired: Indeed, wanted this.

addon/-private/system/identity-map.js
runspired: I believe we must specifically clear them because clear is responsible for tearing down the internal models.  This is a refactor that preserves existing logic for teardown, but it is absolutely possible we were doing an excessive amount of work on teardown in the past.

stefanpenner: ah ok, lets punt.

runspired: @stefanpenner all but 2 tests still pass with use simply releasing, and they look to be two tests that are testing whether internal-model teardown things are done.

bmac:
This looks good to me @runspired. Do you mind rebasing it when you have a chance?

runspired:
@bmac rebased!

PR #4667, opened 2016-11-18
[PERF] flatten DS.Model

runspired:
- shrinks `reopenClass` usage to one reopen from three separate reopens with mixins.
- eliminates double `Object.extend`
- moves `attr` `belongsTo` `_debugInfo` and `hasMany` mixins directly onto `DS.Model`

cc @krisselden @stefanpenner I'd love to have a review + some extra profiling of the realized effects of this on our app. I know the approach is a win, curious to know just how much.


stefanpenner:
Will get to this today.

stefanpenner:
> cc @krisselden @stefanpenner I'd love to have a review + some extra profiling of the realized effects of this on our app. I know the approach is a win, curious to know just how much.

Often what I do for quick "is this going in the right direction" is i isolate a part of a flame graph I want to see go away. Then compare it explicit before/after. This doesn't translate into high quality #'s, but it helps 1. communicate your intention, and 2. double check that it has remove the part you expected.

stefanpenner:
Alternatively, things like this are often quite skewed in flame graphs (as they happen once for each first instance of a class). More concise benchmarks may be handy.

runspired:
So I benchmarked this as a win, but the benchmarks actually only contained the perf of using and accessing the computeds on a `modelClass`, the lookup cost and the materialization costs were outside the area I was focusing on. Working on adjusting the instrumentation but that suggests to me this is may be a much more significant win than my initial analysis had suggested (fwiw the initial analysis was confusing me as it was much less a win that I felt it would be based on past experience).  New numbers and a cleaned up PR coming soon.

runspired:
I'm closing this in favor of a new PR (forthcoming) that does this in a few more commits. The advantage will be that it'll be a little easier to go back in history and see where the various parts of DS.Model originated from.

PR #4668, opened 2016-11-18
[PERF] use micro-queue

runspired:
@stefanpenner and I were discussing how it's better to use a more specialized micro-queue than to push everything into backburner a few days ago. I was spelunking through a profile and noticed that pushing a bunch of records of one time wasn't batched even though I thought it should be, turns out the run.join + the excess calls to schedule were really hurting us.  In a few limited runs this seemed to help by as much as 30% on the time to push a bunch of records of the same type into the store.

**EDIT**

So upon request, I separated this out into two separate PRs (see #4698 for first change and numbers associated with it).  Here are the numbers for making "just this change" to current master:

Numbers are measuring store._push for 238 records: 34 primary records and 204 related records (6 per primary record). Numbers are in milliseconds and from 35 runs comparing current master to this PR.

**current master**
```json
{
  "min": "17.30",
  "q1": "23.76",
  "q2": "24.86",
  "q3": "27.17",
  "max": "40.92"
}
```

**PR**
```json
{
  "min": "17.58",
  "q1": "24.07",
  "q2": "26.19",
  "q3": "27.41",
  "max": "41.63"
}
```

Obviously this doesn't seem like much of a win at first, if at all, other than a simpler data flow story (we time/batch all the relationship updates together).

However, much like how we didn't see much of an improvement in #4698 either, here are the numbers for once you combine these two techniques.

```json
{
  "min": "16.67",
  "q1": "20.22",
  "q2": "22.41",
  "q3": "24.95",
  "max": "35.48"
}
```

### DIFF
--- a/addon/-private/system/store.js
+++ b/addon/-private/system/store.js
@@ -217,11 +217,24 @@ Store = Service.extend({
     this.recordArrayManager = RecordArrayManager.create({
       store: this
     });
-    this._pendingSave = [];
-    this._instanceCache = new ContainerInstanceCache(getOwner(this), this);
 
-    //Used to keep track of all the find requests that need to be coalesced
+    /*
+      Ember Data uses several specialized micro-queues for organizing
+      and coalescing similar async work.
+
+      These queues are currently controlled by a flush scheduled into
+      ember-data's custom backburner instance.
+     */
+    // used for coalescing record save requests
+    this._pendingSave = [];
+    // used for coalescing relationship setup needs
+    this._pushedInternalModels = [];
+    // stores a reference to the flush for relationship setup
+    this._relationshipFlush = null;
+    // used to keep track of all the find requests that need to be coalesced
     this._pendingFetch = MapWithDefault.create({ defaultValue() { return []; } });
+
+    this._instanceCache = new ContainerInstanceCache(getOwner(this), this);
   },
 
   /**
@@ -1850,25 +1863,27 @@ Store = Service.extend({
     let pending = this._pendingSave.slice();
     this._pendingSave = [];
 
-    pending.forEach((pendingItem) => {
+    for (let i = 0, j = pending.length; i < j; i++) {
+      let pendingItem = pending[i];
       let snapshot = pendingItem.snapshot;
       let resolver = pendingItem.resolver;
-      let record = snapshot._internalModel;
-      let adapter = this.adapterFor(record.modelClass.modelName);
+      let internalModel = snapshot._internalModel;
+      let adapter = this.adapterFor(internalModel.modelClass.modelName);
       let operation;
 
-      if (get(record, 'currentState.stateName') === 'root.deleted.saved') {
+      if (get(internalModel, 'currentState.stateName') === 'root.deleted.saved') {
         return resolver.resolve();
-      } else if (record.isNew()) {
+      } else if (internalModel.isNew()) {
         operation = 'createRecord';
-      } else if (record.isDeleted()) {
+      } else if (internalModel.isDeleted()) {
         operation = 'deleteRecord';
       } else {
         operation = 'updateRecord';
       }
 
       resolver.resolve(_commit(adapter, this, operation, snapshot));
-    });
+    }
+
   },
 
   /**
@@ -1891,8 +1906,8 @@ Store = Service.extend({
     }
     if (data) {
       // normalize relationship IDs into records
-      this._backburner.schedule('normalizeRelationships', this, this._setupRelationships, internalModel, data);
       this.updateId(internalModel, data);
+      this._setupRelationshipsForModel(internalModel, data);
     } else {
       assert(`Your ${internalModel.type.modelName} record was saved to the server, but the response does not have an id and no id has been set client side. Records must have ids. Please update the server response to provide an id in the response or generate the id on the client side either before saving the record or while normalizing the response.`, internalModel.id);
     }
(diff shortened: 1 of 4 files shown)


addon/-private/system/store.js
```diff
@@ -1872,7 +1872,8 @@ Store = Service.extend({
     }
     if (data) {
       // normalize relationship IDs into records
-      this._backburner.schedule('normalizeRelationships', this, this._setupRelationships, internalModel, data);
+      this._pushedInternalModels.push(internalModel, data);
+      this._backburner.schedule('normalizeRelationships', this, this._setupRelationships);
```
runspired: this should also check for an existing flush

stefanpenner:
The idea is, lets use backburner for "the flush"

runspired:
There's issues with this PR (beyond the wrong check for !== null) that I haven't finished debugging. This change leads to adapterRun being called many more times.

addon/-private/system/store.js
```diff
@@ -1872,7 +1872,10 @@ Store = Service.extend({
     }
     if (data) {
       // normalize relationship IDs into records
-      this._backburner.schedule('normalizeRelationships', this, this._setupRelationships, internalModel, data);
+      this._pushedInternalModels.push(internalModel, data);
+      if (this._relationshipFlush === null) {
+        this._relationshipFlush = this._backburner.schedule('normalizeRelationships', this, this._setupRelationships);
```
stefanpenner: rather then forcing the caller to notice this. I would make `_setupRelationships` do this, that way we don't need to leak the scheduling to the caller.

runspired: Addressed

runspired:
So officially I'm seeing about 5% improvement overall with this change on my usual production setup, but it also seems to have significantly smoothed out the spikes and the cpu profile and timelines are much cleaner.

addon/-private/system/store.js
```diff
@@ -1845,24 +1846,29 @@ Store = Service.extend({
...
+      } else if (internalModel.isDeleted()) {
         operation = 'deleteRecord';
       } else {
         operation = 'updateRecord';
       }
 
       resolver.resolve(_commit(adapter, this, operation, snapshot));
+    }
+
+    pending.forEach((pendingItem) => {
```
rwjblue: Seems odd here, is this just leftover?

addon/-private/system/store.js
```diff
@@ -2265,48 +2272,52 @@ Store = Service.extend({
 
     @method _push
     @private
-    @param {Object} data
+    @param {Object} apiDoc
```
rwjblue: I don't immediately grok what `apiDoc` means here, is this a common term? Maybe you could add a short description after the name to make it clearer to folks?

addon/-private/system/store.js
```diff
@@ -2265,48 +2272,52 @@ Store = Service.extend({
...
+      return internalModel;
+    });
   },
 
   _hasModelFor(modelName) {
     return !!getOwner(this)._lookupFactory(`model:${modelName}`);
   },
 
+  _pushedInternalModels: null,
+  _relationshipFlush: null,
```
rwjblue: This should be set to `null` in the constructor (`init`) above.

tests/integration/adapter/build-url-mixin-test.js
```diff
@@ -252,7 +252,7 @@ test('buildURL - buildURL takes the records from findMany', function(assert) {
         }
       }
     });
-    post.get('comments').then(assert.wait(function(post) {
+    post.get('comments').then(assert.wait(function(rel) {
```
pangratz: Unrelated

addon/-private/system/store.js
```diff
@@ -2672,7 +2699,7 @@ function _commit(adapter, store, operation, snapshot) {
       if (adapterPayload) {
         payload = normalizeResponseHelper(serializer, store, modelClass, adapterPayload, snapshot.id, operation);
         if (payload.included) {
-          store.push({ data: payload.included });
+          store._push({ data: null, included: payload.included });
```
pangratz: 👍 good one

addon/-private/system/store.js
pangratz: Maybe `jsonApiDoc`?

tests/integration/adapter/build-url-mixin-test.js
runspired: creeped in from a test failure i was investigating, will scrub

runspired:
Following Meeting feedback, I'll be separating this into two PRs. I'll leave this one for the micro-queue and pull the move of `run.join` to `_push` into a separate PR.

addon/-private/system/store.js
bcardarella: what is the difference between `store.push` and `store._push` here?

runspired: @bcardarella the main difference is that it doesn't materialize the records, however it also will likely become the basis for a new public API that's currently going through the RFC process.

addon/-private/system/store.js
```diff
@@ -2621,7 +2646,7 @@ Store = Service.extend({
       return;
     }
 
-    assert(`A ${relationship.parentType} record was pushed into the store with the value of ${relationship.key} being ${inspect(resourceIdentifier)}, but ${relationship.key} is a belongsTo relationship so the value must not be an array. You should probably check your data payload or serializer.`, !Array.isArray(resourceIdentifier));
+    assert(`A ${relationship.record.modelName} record was pushed into the store with the value of ${relationship.key} being ${inspect(resourceIdentifier)}, but ${relationship.key} is a belongsTo relationship so the value must not be an array. You should probably check your data payload or serializer.`, !Array.isArray(resourceIdentifier));
```
runspired: made this change because `relationship.parentType` is not a real thing which led to this always reading `undefined`.

bmac:
Thanks @runspired.

PR #4670, opened 2016-11-20
Removes Ember.K from codebase

locks:


tests/integration/records/delete-record-test.js
```diff
@@ -1,4 +1,4 @@
-/*eslint no-unused-vars: ["error", { "varsIgnorePattern": "(adam|dave|cersei)" }]*/
+3/*eslint no-unused-vars: ["error", { "varsIgnorePattern": "(adam|dave|cersei)" }]*/
```
cibernox: A bit late, but there is a number here 👀

locks: I had actually fixed this and "pushed", but I forgot to actually `git add` the file so it was in the commit… Thanks @sduquej :)

sduquej: 😅 It was also a bit too late when I saw it. @locks 👍

PR #4672, opened 2016-11-21
JSON API "included" should be an array of resource objects

pcambra:
According to the JSON API specs in http://jsonapi.org/format/#document-resource-objects

"In a compound document, all included resources MUST be represented as an array of resource objects in a top-level included member."

But it seems json-api.js expects having a type and id outside the data structure that wraps jsonapi resource objects.

### DIFF
--- a/addon/serializers/json-api.js
+++ b/addon/serializers/json-api.js
@@ -158,7 +158,7 @@ const JSONAPISerializer = JSONSerializer.extend({
       let ret = new Array(documentHash.included.length);
 
       for (let i = 0; i < documentHash.included.length; i++) {
-        let included = documentHash.included[i];
+        let included = documentHash.included[i].data;
         ret[i] = this._normalizeResourceHelper(included);
       }
 


runspired:
@pcambra this is a misinterpretation of the spec. Resource objects are not required to be wrapped in a `data` object.

runspired:
You can see this here at the link you provided by scrolling down to the example doc: http://jsonapi.org/format/#document-compound-documents

You can see the definition of ResourceObject here: http://jsonapi.org/format/#document-resource-objects

wecc:
Thank you for contributing to Ember Data!

As @runspired correctly stated – Resource Objects does not need to be wrapped inside of `data` if they are side loaded in a Compound Document (in `included`).

I'm going to close this PR as the current behavior should be in alignment with the specification. Feel free to reach out in the `#-ember-data` channel on the Ember Community Slack if you have any questions.

pcambra:
Thanks for the answers, I'm getting this json response from the json api drupal module, I'll follow up with the maintainers of that one.

PR #4675, opened 2016-11-21
[BUGFIX beta] Inverse null relationships should throw if model doesn't exist 2

canufeel:
this is continuation of #4558 which is closed in favour of this one.

`2.6.0` [Introduced](https://github.com/asakusuma/data/blob/ced181ff68f87eecd149c6d15b1f8b6887f17142/addon/-private/system/relationships/state/create.js#L16) a bug such that for `{ inverse: null }` relationships the relationship target is not checked at model creation time. However without `{ inverse: null }` we would get an error when trying to create a model with relationship to another model that doesn't exist. This PR tries to bring in consistent behaviour for such cases - perform a check for model type during model creation and throw an error if the model is not present in the store.

### DIFF
--- a/addon/-private/system/relationships/state/create.js
+++ b/addon/-private/system/relationships/state/create.js
@@ -2,6 +2,7 @@ import Ember from 'ember';
 import ManyRelationship from "ember-data/-private/system/relationships/state/has-many";
 import BelongsToRelationship from "ember-data/-private/system/relationships/state/belongs-to";
 import EmptyObject from "ember-data/-private/system/empty-object";
+import { runInDebug } from 'ember-data/-private/debug';
 
 var get = Ember.get;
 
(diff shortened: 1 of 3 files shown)


bmac:
This looks good to me. @runspired I would appreciate your eyes on this pr.

runspired:
@bmac will review this evening 👍

runspired:
I forgot to come back and say that I reviewed and it looked good. @bmac

runspired:
Specifically, it does not affect or hinder any efforts at laziness, and is properly scoped to debug code paths. Any additional work to ensure laziness of relationship setup need not hold this up.

bmac:
@canufeel do you have time to rebase this pr?

canufeel:
@bmac sure! Would spend some time on it in the next couple days.

canufeel:
@bmac did a rebase

PR #4678, opened 2016-11-22
update documentation for `resetAttribute`

efx:
It was not clear from reading the API documentation that this feature is hidden behind a feature flag. I think it will be helpful to include the following note to provide guidance to users.


pangratz:
Can you update the documentation and name the feature `ds-rollback-attribute`, which will be the case once #4685 is merged.

Thanks!

PR #4684, opened 2016-11-29
Modernizes relationship containers

runspired:
Was spelunking about cleaning up our async story and fell into cleaning up this bit once I realized that `record` was meant to be `internalModel`.  Decided to go all the way and give us a fresh ES6 look.


pangratz:
Thanks!

PR #4686, opened 2016-11-30
[FEATURE ds-improved-ajax] Disable feature

pangratz:
As mentioned in https://github.com/emberjs/website/pull/2727, this feature will be disabled in favor of the corresponding [RFC#171](https://github.com/emberjs/rfcs/pull/171).

### DIFF
--- a/config/features.json
+++ b/config/features.json
@@ -1,5 +1,5 @@
 {
-  "ds-improved-ajax": true,
+  "ds-improved-ajax": null,
   "ds-pushpayload-return": null,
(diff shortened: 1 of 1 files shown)


bmac:
Thanks @pangratz.

denzo:
@pangratz @bmac Should this then be removed from docs? https://emberjs.com/api/data/classes/DS.RESTAdapter.html#method_dataForRequest

Is there a rough timeline on something similar being released?

Issue #4689, opened 2016-12-01
Add blueprints for "ember-cli-mocha >= 0.12.0"


Turbo87:
`ember-cli-mocha@0.12.0` introduced a [new testing API](https://github.com/emberjs/ember-mocha#upgrading) and deprecated the old `describeXXX()` helpers. The blueprints in this repository need to be updated to reflect those changes.

Instead of just updating we should make sure that uses with all versions of `ember-cli-mocha` get the right blueprints. We can accomplish that by checking the installed version of `ember-cli-mocha` using a [VersionChecker](https://github.com/rwjblue/ember-cli-version-checker) and then deciding whether to use the `mocha-files` or `mocha-0.12-files` blueprints folder.

The tracking issue for this is https://github.com/ember-cli/ember-cli-mocha/issues/142

PR #4691, opened 2016-12-02
[BUGFIX beta] Add blueprints for "ember-cli-mocha >= 0.12.0"

Turbo87:
Resolves https://github.com/emberjs/data/issues/4689

### DIFF
--- a/blueprints/test-framework-detector.js
+++ b/blueprints/test-framework-detector.js
@@ -1,4 +1,6 @@
+var fs = require('fs');
 var path = require('path');
+var VersionChecker = require('ember-cli-version-checker');
 
 module.exports = function(blueprint) {
   blueprint.supportsAddon = function() {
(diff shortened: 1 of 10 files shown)


Turbo87:
discovered a small issue with this, do not merge for now... 😞

blueprints/test-framework-detector.js
```diff
@@ -12,7 +14,12 @@ module.exports = function(blueprint) {
     if ('ember-cli-qunit' in dependencies) {
       type = 'qunit';
     } else if ('ember-cli-mocha' in dependencies) {
-      type = 'mocha';
+      var checker = new VersionChecker({ project: this.project });
+      if (fs.existsSync(this.path + '/mocha-0.12-files') && checker.for('ember-cli-mocha', 'npm').satisfies('>=0.12.0')) {
```
pangratz: Why is the check for `/mocha-0.12-files` necessary`? Shouldn't this be the case when the addon of this version is installed?

Turbo87: @pangratz that check allows us to have some blueprints that care about the Mocha version and others that don't. This is not technically needed in this case, but I've reused the code from https://github.com/emberjs/ember.js/pull/14670 to reduce differences.

pangratz:
🚀 🎉 thanks a bunch @Turbo87

PR #4693, opened 2016-12-05
Fixed modelFor documentation. Look up model's class name

maartenparmentier:
```Represents the model's class name as a string. This can be used to look up the model's class name through DS.Store's modelFor method.```

This is wrong and should be `model's class name` since you don't look up the actual model but the name.

### DIFF
--- a/addon/-private/system/model/model.js
+++ b/addon/-private/system/model/model.js
@@ -996,7 +996,7 @@ Model.reopenClass({
 
   /**
    Represents the model's class name as a string. This can be used to look up the model's class name through
-   DS.Store's modelFor method.
+   `DS.Store`'s modelFor method.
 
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: addon/-private/system/model/model.js [hunk 1]
```diff
@@ -995,7 +995,7 @@ Model.reopenClass({
   },
 
   /**
-   Represents the model's class name as a string. This can be used to look up the model through
+   Represents the model's class name as a string. This can be used to look up the model's class name through
    DS.Store's modelFor method.
```
locks: Can you wrap `DS.Store` in a code fence while you're in there? Thanks!

maartenparmentier: @locks done!

pangratz:
Thanks for helping out making the docs more clear @maartenparmentier! ❤️

PR #4696, opened 2016-12-06
fix(benchmarks): benchmarks for store.query no longer included record…

runspired:
… materialization as this is now so lazy that it must be forced, this makes the benchmark force it so that we can continue to measure improvement within materialization

### DIFF
--- a/tests/dummy/app/routes/query/route.js
+++ b/tests/dummy/app/routes/query/route.js
@@ -27,6 +27,7 @@ export default Route.extend({
     let token = heimdall.start('ember-data');
     return this.get('store').query(modelName, params)
       .then((records) => {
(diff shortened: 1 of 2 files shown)


tests/dummy/app/routes/query/route.js
```diff
@@ -24,6 +24,7 @@ export default Route.extend({
     let token = heimdall.start('ember-data');
     return this.get('store').query(modelName, params)
       .then((records) => {
+        records.toArray();
```
bmac: @runspired do you mind adding a comment explaining this line forces the materialization so we don't forget in the future?

runspired:
@bmac updated and rebased.

PR #4699, opened 2016-12-06
include related record on the complex test

runspired:
When benchmarking the micro queues for store._push I realized that I wasn't including related records in the payload anymore, this fixes that for the one scenario in the config but should be fixed for more of them long term.


bmac:
Thanks @runspired.

PR #4701, opened 2016-12-06
[PERF] flatten DS.Model to avoid multi-extend, expensive reopens, and extra mixin detection

runspired:
Replaces #4667 with a version that takes more commits so we preserve more of the history of where code came from.

**Perf**

We see some improvement in both the areas we expect (initial lookup, materialization), there may also be a modest 2-5% improvement in several other areas but I'm skeptical.  I want to run more test runs on this, and I'd like @stefanpenner to put it in our app with me tomorrow and measure in the real world.

**initial-modelFor-lookup**

**Master**

```json
{
  "min": "1.75",
  "q1": "2.18",
  "q2": "2.40",
  "q3": "2.83",
  "max": "3.30",
  "avg": "2.48"
}
```

**PR**

```json
{
  "min": "1.80",
  "q1": "2.08",
  "q2": "2.14",
  "q3": "2.5",
  "max": "3.03",
  "avg": "2.33"
}
```

------------------------------------

**InternalModel.getRecord (238 total calls)**

**Master**

```json
{
  "min": "6.84",
  "q1": "7.67",
  "q2": "9.36",
  "q3": "11.92",
  "max": "15.33",
  "avg": "9.94"
}
```

**PR**

```json
{
  "min": "6.60",
  "q1": "7.63",
  "q2": "8.17",
  "q3": "9.05",
  "max": "11.70",
  "avg": "8.49"
}
```


runspired:
I'm skeptical of the benchmark numbers above, I think I may be getting trolled but haven't figured out where yet. However, I spent some time in the profiler and it's fairly clear that this change is a 40%  win for the initial parse of DS.Model.

**Before**
before-flattened-model

**After**
after-flattened-model

runspired:
Obviously a secondary benefit of all this is a much cleaner flame chart for spotting where to optimize things in the future as well.

stefanpenner:
@runspired thanks for the annotated screenshot, it is very easy to now identify exactly what you saw :)

PR #4704, opened 2016-12-09
Factory cache

runspired:
Needs benchmarking, review, may have been too aggressive in `_modelFor` use, dislike that `modelFactoryFor` is public, should discuss.

Also, one reason for WIP is that this builds off of #4664, may be easier to review once that lands and I rebase ;)

### DIFF
--- a/addon/-private/system/record-arrays/record-array.js
+++ b/addon/-private/system/record-arrays/record-array.js
@@ -92,7 +92,7 @@ export default Ember.ArrayProxy.extend(Ember.Evented, {
     if (!this.modelName) {
       return null;
     }
-    return this.store.modelFor(this.modelName);
+    return this.store._modelFor(this.modelName);
   }).readOnly(),
 
   /**
--- a/addon/-private/system/store.js
+++ b/addon/-private/system/store.js
@@ -215,6 +215,7 @@ Store = Service.extend({
     this._identityMap = new IdentityMap();
     this._pendingSave = [];
     this._instanceCache = new ContainerInstanceCache(getOwner(this), this);
+    this._modelClassCache = new EmptyObject();
 
     /*
       Ember Data uses several specialized micro-queues for organizing
@@ -1231,7 +1232,7 @@ Store = Service.extend({
     assert('Passing classes to store methods has been removed. Please pass a dasherized string instead of '+ inspect(modelName), typeof modelName === 'string');
 
     let modelToken = heimdall.start('initial-modelFor-lookup');
-    let modelClass = this.modelFor(modelName);
+    let modelClass = this._modelFor(modelName);
     heimdall.stop(modelToken);
 
     array = array || this.recordArrayManager.createAdapterPopulatedRecordArray(modelName, query);
@@ -1349,7 +1350,7 @@ Store = Service.extend({
     assert('Passing classes to store methods has been removed. Please pass a dasherized string instead of '+ inspect(modelName), typeof modelName === 'string');
     let trueModelName = this._classKeyFor(modelName);
 
-    let modelClass = this.modelFor(trueModelName);
+    let modelClass = this._modelFor(trueModelName);
     let adapter = this.adapterFor(trueModelName);
 
     assert(`You tried to make a query but you have no adapter (for ${trueModelName})`, adapter);
@@ -1560,7 +1561,7 @@ Store = Service.extend({
     assert('Passing classes to store methods has been removed. Please pass a dasherized string instead of '+ inspect(modelName), typeof modelName === 'string');
     let token = heimdall.start('store.findAll');
     let trueModelName = this._classKeyFor(modelName);
-    let modelClass = this.modelFor(trueModelName);
+    let modelClass = this._modelFor(trueModelName);
     let fetch = this._fetchAll(modelClass, this.peekAll(trueModelName), options);
 
     instrument(() => {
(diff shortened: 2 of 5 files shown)


addon/-private/system/record-map.js
```diff
@@ -0,0 +1,47 @@
...
+  }
+
+  /* deprecated way of accessing modelClass */
+  get type() {
+    throw new Error('RecordMap.type is no longer available');
+  }
+
+  clear() {
+    if (this._records) {
+      let records = this._records.slice();
```
stefanpenner: we don't appear to purge `this._records` membership, should we maybe instead do.

```js
let records = this._records;
this._records = []; // or maybe null
```

That way, we release anything currently retained by `this._records` and don't need to copy the `this._records` array via `slice`

addon/-private/system/record-arrays/record-array.js
```diff
@@ -88,11 +79,23 @@ export default Ember.ArrayProxy.extend(Ember.Evented, {
...
+
+   @property type
+   @type DS.Model
+   */
+  type: computed('modelName', function() {
+    if (!this.modelName) {
+      return null;
+    }
+    return this.store._modelFor(this.modelName);
+  }),
```
stefanpenner: `readOnly`

addon/-private/system/identity-map.js
```diff
@@ -0,0 +1,28 @@
+import RecordMap from './record-map';
+import EmptyObject from './empty-object';
+
+export default class IdentityMap {
+  constructor() {
+    this._map = new EmptyObject();
```
stefanpenner: `Object.create(null)` may be fine, here. EmptyObject was more of a trick for mega mega hotspots to work-around a strange perf issue in older v8's.

In this case, I believe `Object.create(null)` will be best. (less caveats etc)

stefanpenner:
I'm 👍  with this direction, although the title of the PR is a-bit misleading, after reading the PR i feel much better.

We should be careful re: compat, but this change is worth doing!

addon/-private/system/store.js
```diff
@@ -2065,32 +2047,53 @@ Store = Service.extend({
...
-    }
+    let trueModelName = this._classKeyFor(modelName);
+
+    return this._modelFor(trueModelName);
+  },
+
+  /*
+    @private
+   */
+  _modelFor(modelName) {
```
bmac: I'd like to see this remain a public API. `modelFactoryFor` should probably be removed.

addon/-private/system/record-map.js
runspired: I'll look into it. This is the same logic preserved from before, and it's unclear if emptying this has ramifications, I think it does (I think individual records are currently responsible for removing themselves).  This is a data flow story I will untangle soon in order to also help us unlock better change tracking for record arrays.

addon/-private/system/store.js
runspired: It remains public, the differentiator is that "public" APIs must always normalize the modelName while "private" APIs can assume a normalized modelName.  We currently call this so often internally I wanted to avoid that unnecessary step.

runspired: I also agree that `modelFactoryFor` should probably be removed.

runspired:
@stefanpenner 
The requested changes are actually all for #4664 which this PR build on top of / branches from. I will be addressing the feedback there. 

I believe this is also the reason for the confusion around the name. This PR creates a cache for the `modelClass` factories that we lookup. Previously, every call to `modelFor` would do a fresh `owner._lookupFactory` (potentially multiple if finding a mixin) + re-set `factory.modelName`. This turned out to be much more costly than one might expect, and as we currently call `modelFor` roughly 2N + X times per query (where N is number of records returned in a call), this really added up.

addon/-private/system/store.js
runspired: FTR, the reason we can't remove `modelFactoryFor` at the moment is because `adapters` use it to do a `modelFor` lookup without risk of erroring when a model isn't found when an unknown type is encountered in a payload (prints a warning, doesn't error).

stefanpenner:
> The requested changes are actually all for #4664 which this PR build on top of / branches from. I will be addressing the feedback there.

👍 

> I believe this is also the reason for the confusion around the name. This PR creates a cache for the modelClass factories that we lookup. Previously, every call to modelFor would do a fresh owner._lookupFactory (potentially multiple if finding a mixin) + re-set factory.modelName. This turned out to be much more costly than one might expect, and as we currently call modelFor roughly 2N + X times per query (where N is number of records returned in a call), this really added up.


Yup, in the end the name seemed fine. The root cause of the confusion, was just words a words.

addon/-private/system/record-map.js
runspired: Our tests still pass with this in place, so I've made this change.

addon/-private/system/record-arrays/record-array.js
runspired: Change made.

addon/-private/system/identity-map.js
runspired: I switched to `Object.create(null)` but we should probably investigate which is currently better.

addon/-private/system/record-map.js
stefanpenner: @runspired sounds good, making this change should prevent a potential mem leak.

addon/-private/system/store.js
bmac: 👍

bmac:
Overall this pr looks pretty good. Once https://github.com/emberjs/data/pull/4664 lands I think this one should be ready to go.

PR #4706, opened 2016-12-09
Improved performance for findHasMany finder

runspired:
use _push instead of mapping over push result to get internal models


bmac:
Thanks @runspired.

PR #4708, opened 2016-12-09
es6 RecordArrayManager

runspired:
There wasn't really a good reason for this singleton being an Ember.Object (it's not resolved / not in the container / no ember-y things in use), and converting it shaves 33% off it's instantiation time in addition to giving us nice es6 syntax (1.53ms => .97ms).

I additionally moved the instantiation of the `OrderedSet` used to track the `RecordArray`s for an `internalModel` into internal models directly instead of having the manager reach in.  I'm slightly conflicted on this because of the resulting double-underscore for the cached property, but I prefer to leave objects in control of their own properties.

**Before**
es6-manager-before

**After**
es6-manager-after

### DIFF
--- a/tests/dummy/app/routes/query/route.js
+++ b/tests/dummy/app/routes/query/route.js
@@ -24,8 +24,6 @@ export default Route.extend({
     let modelName = params.modelName;
     delete params.modelName;
 
-    heimdall.enableTimelineFeatures();
-
     let token = heimdall.start('ember-data');
     return this.get('store').query(modelName, params)
       .then((records) => {
--- a/tests/unit/record-arrays/record-array-test.js
+++ b/tests/unit/record-arrays/record-array-test.js
@@ -192,14 +192,25 @@ test('#_removeInternalModels', function(assert) {
   assert.deepEqual(content, [], 'now contains no models');
 });
 
+class FakeInternalModel {
+  constructor(record) {
+    this._record = record;
+    this.__recordArrays = null;
+  }
+
+  get _recordArrays() {
+    return this.__recordArrays;
+  }
+
+  getRecord() { return this._record; }
+
+  createSnapshot() {
+    return this._record;
+  }
+}
+
 function internalModelFor(record) {
-  return {
-    _recordArrays: undefined,
-    getRecord() { return record; },
-    createSnapshot() {
-      return record;
-    }
-  };
+  return new FakeInternalModel(record);
 }
 
 test('#save', function(assert) {
@@ -237,7 +248,7 @@ test('#destroy', function(assert) {
   let internalModel1 = internalModelFor(model1);
 
   // TODO: this will be removed once we fix ownership related memory leaks.
-  internalModel1._recordArrays = {
+  internalModel1.__recordArrays = {
     delete(array) {
       didDissociatieFromOwnRecords++;
       assert.equal(array, recordArray);
@@ -308,7 +319,7 @@ test('#destroy', function(assert) {
   let internalModel1 = internalModelFor(model1);
 
   // TODO: this will be removed once we fix ownership related memory leaks.
-  internalModel1._recordArrays = {
+  internalModel1.__recordArrays = {
     delete(array) {
       didDissociatieFromOwnRecords++;
       assert.equal(array, recordArray);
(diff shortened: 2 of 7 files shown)


stefanpenner:
This may be ok here. Although it is unclear if others don't reopen/extend this class.  If that isn't the case, and the output isn't unreasonably large. This *may* be fine.

---

@runspired / @krisselden and myself should powow soon to fixup the cost centers insides of creation (of first object when mixins need application). I believe there is a relatively clear path to improving that, that would just have mega leverage across the whole ecosystem.

runspired:
@stefanpenner this isn't something others can extend, but it is something that could theoretically have been reached into, found, and reopened.  Is there a list of ember-data addons that we generally search for private usage in? cc @bmac

tests/dummy/app/routes/query/route.js
```diff
@@ -24,6 +24,7 @@ export default Route.extend({
     let modelName = params.modelName;
     delete params.modelName;
 
+    console.time('ember-data');
```
stefanpenner: ^ should we leave this?

tests/dummy/app/routes/query/route.js
```diff
@@ -33,6 +34,7 @@ export default Route.extend({
         // and clutter our benchmarks and make it harder to time.
         records.toArray();
         heimdall.stop(token);
+        console.timeEnd('ember-data');
```
stefanpenner: ^ should we leave this?

tests/unit/record-arrays/record-array-test.js
```diff
@@ -192,14 +192,25 @@ test('#_removeInternalModels', function(assert) {
...
 });
 
+class FakeInternalModel {
+  constructor(record) {
+    this._record = record;
+    this.__recordArrays = null;
+  }
+
+  get _recordArrays() {
+    return this.__recordArrays;
```
stefanpenner: who sets `__recordArrays` ?

runspired: perhaps no one at the moment, but our tests often introspect state so moving it to a similar getter will let us test correctly

tests/dummy/app/routes/query/route.js
runspired: I've found it invaluable and it's in our dummy app so it should be ok.

runspired: however I also just want to bake this into heimdall.

tests/unit/record-arrays/record-array-test.js
stefanpenner: ok, sounds good.

tests/dummy/app/routes/query/route.js
stefanpenner: 👍

krisselden:
> (1.53ms => .97ms).

was this one sample?

krisselden:
@stefanpenner didn't you reopen this at Yapp for the subscribe stuff?

stefanpenner:
> @stefanpenner didn't you reopen this at Yapp for the subscribe stuff?

Unclear, I do also share your concern this is risky relative to the value. (making mixin stuff faster, would have nice leverage...)

krisselden:
Also, the proto() is only doing work here because of lookupFactory extend.

runspired:
As far as wins go, this one is pretty small, but one of hundreds of paper cuts we die from when out together.

I don't actually think the win here is in this small perf gain as much as it is in building a core set of primitives that don't depend on Ember unless they truly need to. If Ember.Object and lookupFactory do have their costs lowered, I'm still unconvinced that this would belong remaining one.

bmac:
I just checked out the [47 Ember Data Adapter addons](https://emberobserver.com/categories/ember-data-adapters) from ember observer into a directory and grepped for `RecordArrayManager` and found no matches. I think its unlikely that anyone is reopening the RecordArrayManager.

chadhietala:
@runspired This great! Is there any way to get some more exhaustive benchmarks?

runspired:
@chadhietala the benchmark runs for repeated iterations are roughly similar. Since we strip the classCallCheck I'd be surprised if the compiled output were larger vs equal or smaller than the compiled output for the Ember.Object version.

fivetanley:
Would heavily prefer this to be feature flagged.

bmac:
@fivetanley What is your thinking on why the feature flag for this change. In theory it should be invisible to users and doesn't provide and new functionality?

runspired:
@bmac @fivetanely I think this is the sort of thing where if we send it out, and it breaks people's apps, its small and we just revert it. Don't think it warrants a flag.

runspired:
cc @bmac @fivetanley for review and suggested next steps.

bmac:
@runspired this just needs a rebase and should be good to merge.

runspired:
@bmac rebased :)

Issue #4709, opened 2016-12-09
Empty Content-Type


mwpastore:
In Ember/Data 2.11, in development mode with `ember server --proxy <..>`, it looks like the response headers from the server are being lower-cased, which causes [this statement](https://github.com/emberjs/data/blob/bf3c7f8708b7e86f0bf05c6f56bae6e8c9cdc4a8/addon/adapters/rest.js#L1186) to set the `payloadContentType` to "Empty Content-Type". 

Here's an API response via `curl`:

```sh
$ curl -i -H 'Accept: application/vnd.api+json' -H 'Content-Type: application/vnd.api+json' -d '{}' -X PATCH :4567/authors/3
```
```
HTTP/1.1 403 Forbidden 
Content-Type: application/vnd.api+json
Content-Length: 206
Server: WEBrick/1.3.1 (Ruby/2.3.3/2016-11-21)
Date: Fri, 09 Dec 2016 20:58:37 GMT
Connection: Keep-Alive
```
```json
{
  "errors": [
    {
      "id": "ac645b7e-238a-416d-b431-496f8bc6fc05",
      "title": "Forbidden Error",
      "detail": "You are not authorized to perform this action",
      "status": "403"
    }
  ]
}
```

Here is `headers` in `generatedDetailedMessage()` via console.log:

```
Vary: "Accept-Encoding"
X-Powered-By: "Express"
connection: "close"
content-length: "206"
content-type: "application/vnd.api+json"
date: "Fri, 09 Dec 2016 21:00:04 GMT"
server: "WEBrick/1.3.1 (Ruby/2.3.3/2016-11-21)"
```

PR #4710, opened 2016-12-09
Stringify JSONAPI errors and payload Content-Type. Fixes #4342 and #4709

### LINKED ISSUE #4342: headers are case insensitive yet ember-data looks for case sensitive headers

mwpastore:


mwpastore:
This needs considerably more thought.

PR #4715, opened 2016-12-14
converts Snapshot to an es6 class instead of using prototype assign

runspired:
Was working on cleaning up some things and just felt this could use a facelift. This isn't for perf, but I'm going to try to isolate it in a profile anyway because I'm personally curious the overall effects of:

- not setting things on the prototype that we're just going to override in the constructor anyway
- (obvs babelified) es6 class definitions vs. prototype definition using the `Object.assign` pattern

### DIFF
--- a/addon/-private/system/snapshot.js
+++ b/addon/-private/system/snapshot.js
@@ -5,7 +5,9 @@
 import Ember from 'ember';
 import EmptyObject from "ember-data/-private/system/empty-object";
 
-var get = Ember.get;
+const {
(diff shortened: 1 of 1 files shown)


runspired:
@stefpenner this is also something that would be a pretty good object to look at the perf of `new EmptyObject()` vs `Object.create(null)` on, as it uses it 4x and populates two of them immediately.

runspired:
cc @bmac for final review / merging

Issue #4362, opened 2016-04-29
Add preconfigured ember twiddle to 'reporting a bug' section


adam-knights:
The reporting a bug section has js bins etc but Ember Twiddle can be alot easier to use. @pangratz had a basic one https://ember-twiddle.com/0e1a24aabb8fa7c1fdd8/copy?fileTreeShown=false&numColumns=2&openFiles=routes.application.js%2Ctemplates.application.hbs
that I found easy to play with and add to.

So something similarly maintained would be good, and then added to this section.

I also noticed Ember CLI now have a template that appears when you try to create an issue - linking to the preconfigured bins in one of those may also save alot of time/waiting for a repo

wecc:
This ember-twiddle link looks good to me! Would you like to open a PR adding it to `CONTRIBUTING.md`?

wecc:
I'm closing this, adding Ember Twiddle is already tracked in #4238. I copied the link from this issue to that issue.

PR #4716, opened 2016-12-14
[DOC canary] Updating CONTRIBUTING.md to use ember-twiddle as examples

pete-the-pete:
PR for #4238 

Created ember-twiddle versions for the reporting bug section of contributing.md.  The first two links are recreations of the existing js-fiddles, and the third is the one mentioned in #4362


bmac:
Thanks @pete-the-pete.

PR #4717, opened 2016-12-14
remove usage and deprecate InternalModel's type property

sly7-7:
I hope this moves to the right direction. (#4584)
I wonder if we would do the same for the Snapshot and RecordArray classes

cc/ @runspired

### DIFF
--- a/addon/-private/system/model/internal-model.js
+++ b/addon/-private/system/model/internal-model.js
@@ -1,5 +1,5 @@
 import Ember from 'ember';
-import { assert, runInDebug } from "ember-data/-private/debug";
+import { assert, deprecate, runInDebug } from "ember-data/-private/debug";
 import RootState from "ember-data/-private/system/model/states";
 import Relationships from "ember-data/-private/system/relationships/state/create";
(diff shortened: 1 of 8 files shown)


stefanpenner:
oops didn't notice this, these should be fixed up. If any are missing please reopen/resubmit

sly7-7:
@stefanpenner No worries, I will check if any are missing

sly7-7:
@stefanpenner I just checked, if we still want to deprecate internalModel.type, there is one place missing: https://github.com/emberjs/data/blob/master/addon/-private/system/relationships/state/create.js#L18

I did not find any other missing replacement, even if for some label, `internalModel.modelClass` was replaced by `internalModel.modelName`, but I assume this is merely the the same thing for labels.

Just let me now if you want me to resubmit a PR deprecating and removing `internalModel.type`

stefanpenner:
I'm not sure how to remove the one you linked to easily

sly7-7:
why not just using [internalModel.modelClass](https://github.com/emberjs/data/blob/master/addon/-private/system/model/internal-model.js#L158) ?

PR #4722, opened 2016-12-21
tweaks to perf instrumentation

runspired:
A recent feature flag change cause some of my instrumentation to follow different code paths and resulted in different than expected results.  These tweaks both account for that and are a bit more resilient.  I've back-ported them to various branches I'm using for comparisons to keep measurements fairly comparable as we iterate.


stefanpenner:
house keeping.

PR #4723, opened 2016-12-21
Simplify Backburner Usage

runspired:
TL;DR spin up less, schedule less, batch more things, cleaner profiles :)

In a query returning 238 records 34 of which are primary records, this resulted in:

- ember’s run.schedule 105 times => 2x
- ember’s run.join 2 times => 1x
- ED’s schedule 274 times => 2x
- ED’s join 274 times => 2x

~~Right now this fails on the tests around support for preloading relationships when doing findRecord, digging in on that.~~

~~Also includes some additional instrumentation stuff that's part of a different PR + some additions to that instrumentation I'll need to back out.~~

~~I appear to have pushed this far enough we get failing tests around pushing hasMany's with invalid values. I think something likely just ended up dropping out of the same run loop, investigating.~~

~~There's still a little bit of low hanging fruit on this.~~

### DIFF
--- a/addon/-private/system/model/internal-model.js
+++ b/addon/-private/system/model/internal-model.js
@@ -366,7 +366,6 @@ export default class InternalModel {
 
   setupData(data) {
     heimdall.increment(setupData);
-    let token = heimdall.start('setupData');
     let changedKeys = this._changedKeys(data.attributes);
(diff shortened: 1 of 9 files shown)


runspired:
cc @bmac for final review / merging

PR #4726, opened 2016-12-23
[SUPER WIP] Relationship State Overhaul

runspired:
This is the beginnings of revisiting the objects / patterns used to manage the relationship layer in ember-data.

- [ ] treat `Relationship` as an abstract base class
- [ ] simpler `BelongsToRelationship` class
- [ ] explicit `ImplictRelationship` class
- [ ] simpler `HasManyRelationship` class

-- potential changes for this effort that may be bumped to a second PR after this one --
- [ ] remove `ManyArray` (upgraded `HasManyRelationship` should be able to handle this entirely)
- [ ] swap out `ArrayProxy` for a more custom solution.

**Current HasMany Complexity**
image_uploaded_from_ios_720

### DIFF
--- a/addon/-private/system/relationships/state/-relationship.js
+++ b/addon/-private/system/relationships/state/-relationship.js
@@ -1,0 +1,448 @@
+import { warn, assert } from 'ember-data/-private/debug';
+import Ember from 'ember';
+import { PromiseObject } from 'ember-data/-private/system/promise-proxies';
+import { assertPolymorphicType } from 'ember-data/-private/debug';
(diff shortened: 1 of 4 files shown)


BryanCrotaz:
Is this part of Single Source of Truth?

stefanpenner:
@BryanCrotaz I believe this is just general relationship cleanup, preventing duplicate work and stuff.

runspired:
@BryanCrotaz @stefanpenner no, but it gets us closer to being able to implement it cleanly.

runspired:
A better version of this work is in #4882

PR #4728, opened 2016-12-27
[CLEANUP beta] Linter changes for addon, addon/adapters and addon/serializers

pete-the-pete:
Minor clean up for linter issues.

### DIFF
--- a/addon/adapters/json-api.js
+++ b/addon/adapters/json-api.js
@@ -144,7 +144,7 @@ import { instrument, deprecate } from 'ember-data/-private/debug';
   @namespace DS
   @extends DS.RESTAdapter
 */
-var JSONAPIAdapter = RESTAdapter.extend({
+const JSONAPIAdapter = RESTAdapter.extend({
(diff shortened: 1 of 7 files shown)


bmac:
Hi @pete-the-pete. Thanks for this pr. Would you be opposed to making some changes so `let` is preferred over `const`? I believe this help bring Ember Data's coding conventions in line with the conventions used it Ember.

I believe Ember follows a convention of "liberally-let" per madhatted.com/2016/1/25/let-it-be and only allows `const` in module scope.

pete-the-pete:
Sure, i'll update soon.  Is it worth it to make the actual eslintrc more similar?

https://github.com/emberjs/data/blob/master/.eslintrc.js vs https://github.com/emberjs/ember.js/blob/master/.eslintrc.js

bmac:
making the .eslintrc.js more similar sounds like a good idea. 👍

bmac:
Thanks @pete-the-pete.

PR #4729, opened 2016-12-28
[CLEANUP Beta] Linter changes for addon/-private (not -private/system)

pete-the-pete:
Minor clean up for linter issues.

### DIFF
--- a/addon/-private/adapters/build-url-mixin.js
+++ b/addon/-private/adapters/build-url-mixin.js
@@ -1,6 +1,6 @@
 import Ember from 'ember';
 
-var get = Ember.get;
+const get = Ember.get;
 
 /**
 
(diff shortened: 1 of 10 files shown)


addon/-private/adapters/build-url-mixin.js
```diff
@@ -401,7 +401,7 @@ export default Ember.Mixin.create({
     }
 
     // No path provided
-    var url = [];
+    const url = [];
```
igorT: Can we not use const for things that are not constants please

addon/-private/adapters/build-url-mixin.js
```diff
@@ -434,7 +434,7 @@ export default Ember.Mixin.create({
     @return {String} path
   **/
   pathForType(modelName) {
-    var camelized = Ember.String.camelize(modelName);
+    const camelized = Ember.String.camelize(modelName);
```
igorT: Same here

pete-the-pete: yes, I am going back through these PRs per @bmac's suggestion:
"I believe Ember follows a convention of "liberally-let" per madhatted.com/2016/1/25/let-it-be and only allows const in module scope."

I'll update once the builds are working again.

PR #4743, opened 2017-01-10
[BUGFIX canary] Fix _lookupFactory deprecation for Ember canary

locks:


addon/-private/utils.js
```diff
@@ -41,7 +41,18 @@ function getOwner(context) {
...
 
+/*
+  ember-factory-for is a new feature in Ember 2.12 that finally provides a public API for looking factories up.
+  This function serves as a super simple polyfill to avoid triggering deprecations.
+ */
+function factoryFor(owner) {
+  if (owner.factoryFor) {
+    return owner.factoryFor
+  }
+}
```
bmac: Is this uses?

addon/-private/utils.js
```diff
@@ -43,5 +43,6 @@ function getOwner(context) {
 
 export {
   modelHasAttributeOrRelationshipNamedType,
-  getOwner
+  getOwner,
+  factoryFor
```
fivetanley: this appears to not be a function anymore?

locks: oops, I only removed part of the 💩

addon/-private/utils.js
```diff
@@ -12,11 +12,11 @@ function modelHasAttributeOrRelationshipNamedType(modelClass) {
   return get(modelClass, 'attributes').has('type') || get(modelClass, 'relationshipsByName').has('type');
 }
 
-/*
+/**
```
bmac: Do you mind removing the extra `*`s that were added in this file? YUIDoc sees these as API docs even though these functions are never made available to the user.

addon/-private/system/store.js
```diff
@@ -2034,7 +2034,14 @@ Store = Service.extend({
     // container = < 1.11
     let owner = getOwner(this);
 
-    let mixin = owner._lookupFactory('mixin:' + normalizedModelName);
+    let mixin;
+
+    if (owner.factoryFor) {
+      mixin = owner.factoryFor(`mixin:${normalizedModelName}`) && owner.factoryFor(`mixin:${normalizedModelName}`).class;
```
pangratz: Should this be cached?

rwjblue: Seems odd to go through the `factoryFor` twice like this, I'd suggest a refactor to:

```js
let MaybeMixin = owner.factoryFor(`mixin:${normalizedModelName}`);
mixin = MaybeMixin && MaybeMixin.class;
```

addon/-private/system/store.js
```diff
@@ -2107,7 +2114,11 @@ Store = Service.extend({
     let trueModelName = this._classKeyFor(modelName);
     let owner = getOwner(this);
 
-    return owner._lookupFactory(`model:${trueModelName}`);
+    if (owner.factoryFor) {
+      return owner.factoryFor(`model:${trueModelName}`) && owner.factoryFor(`model:${trueModelName}`).class;
```
pangratz: This could be cached too

rwjblue: Same comment as above, we shouldn't call `factorFor` twice in a row like this...

addon/-private/system/store.js
```diff
@@ -2034,7 +2034,13 @@ Store = Service.extend({
     // container = < 1.11
     let owner = getOwner(this);
 
-    let mixin = owner._lookupFactory('mixin:' + normalizedModelName);
+    if (owner.factoryFor) {
```
bmac: Looks like a `let mixin;` got lost in the squash.

locks: ✅

bmac:
Thanks @locks.

workmanw:
I was in the process of fixing up ember-data-model-fragments for `ember-data-2.12.0-beta`. I noticed that the beta channel for ember-data is missing this commit. I'm definitely in no hurry, I just wanted to call out that it's labeled `[BUGFIX canary]` but needs to be in the beta channel.

rwjblue:
Ahh, good catch @workmanw. It was likely still canary when the PR was proposed, but once we branched beta for 2.12 the commit prefix didn't get updated. 

@bmac - Can this be pulled in for 2.12 beta cycle?

bmac:
Will do.

bmac:
This pr has been pulled into the beta cycle and included in the https://github.com/emberjs/data/releases/tag/v2.12.0-beta.2 release.

rwjblue:
Thank you @bmac!

workmanw:
I'm happy to report ember-data-model-fragments' test pass on all channels. Thanks!!

PR #4760, opened 2017-01-18
Update deprecate arguments

estshy:
According to docs (https://github.com/emberjs/ember.js/blob/v2.10.0/packages/ember-debug/lib/deprecate.js#L97) ember deprecate requires 3 arguments.

### DIFF
--- a/addon/-private/system/store.js
+++ b/addon/-private/system/store.js
@@ -1783,7 +1783,7 @@ Store = Service.extend({
     @return {boolean}
   */
   recordIsLoaded(modelName, id) {
-    deprecate(`Use of recordIsLoaded is deprecated, use hasRecordForId instead.`, true, {
+    deprecate(`Use of recordIsLoaded is deprecated, use hasRecordForId instead.`, false, {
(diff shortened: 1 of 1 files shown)


addon/-private/system/store.js
```diff
@@ -1783,7 +1783,7 @@ Store = Service.extend({
     @return {boolean}
   */
   recordIsLoaded(modelName, id) {
-    deprecate(`Use of recordIsLoaded is deprecated, use hasRecordForId instead.`, {
+    deprecate(`Use of recordIsLoaded is deprecated, use hasRecordForId instead.`, true, {
```
rwjblue: This should be `false` not `true`.  These deprecations (AFAICT) are never being triggered.  

`deprecate('foo', true, { id: 'asdf', until: 'asdf'})` doesn't actually deprecate anything, but `deprecate('foo', false, { id: 'asdf', until: 'asdf'})` does.

estshy: @rwjblue That's correct.

bmac:
Thanks @estshy.

PR #4762, opened 2017-01-19
[DOC] - add bower install to README

sbatson5:
This may be implied, but it does need to be done before running tests for the first time.


locks:
Thanks!

Issue #4733, opened 2017-01-03
Document caching behavior of store methods.


rwjblue:
It is my understanding that `this.store.query` and `this.store.queryRecord` will never be cached (they will always do a request), but the documentation does not state that (at least during my quick perusal).

We should update the documentation for all of the primary methods to indicate what will happen for repeated calls (do they return prior results, do they always trigger an XHR, etc):

- [x] `store.query` in https://github.com/emberjs/data/pull/4763
- [x] `store.queryRecord` in https://github.com/emberjs/data/pull/4892
- [x] `store.findAll` noop
- [x] `store.findRecord` noop
- [x] `store.peekAll` noop
- [x] `store.peekRecord` in https://github.com/emberjs/data/pull/4900

This API documentation should probably also link to the corresponding `shouldBackgroundReload` type docs that explain how to tweak/modify the behavior.

ryanlabouve:
Can I help with doing this? Maybe I could start with `store.query` and `store.queryRecord`?

locks:
@ryanlabouve please do! 🙏

runspired:
@rwjblue They are cached, they just never return from cache. Yes, we should document this :)

GCheung55:
@runspired if they never return from cache then what's the reason for caching?

locks:
@GCheung55 because you might want to `peekRecord` or `peekAll` afterwards, and if the record is already in the store, you get it back :D

runspired:
@GCheung55 so that `belongsTo`, `hasMany`, `peekAll` `peekRecord` and `findRecord` continue to work like they should :P

ryanlabouve:
Starting with [query](https://github.com/emberjs/data/blob/master/addon/-private/system/store.js#L1220), I'm wanting to double check my understand of what's happening. If this is right, I'll work on working and PR:

>The results of `query` are cached [(source)](https://github.com/emberjs/data/blob/master/addon/-private/system/store/finders.js#L158) in the store, however each time this method is called a new request is made [(source)](https://github.com/emberjs/data/blob/master/addon/-private/system/store/finders.js#L144).

Should I add something about why these are cached? (as @runspired points out above)?

ryanlabouve:
We looked at [peekAll](https://emberjs.com/api/data/classes/DS.Store.html#method_peekAll) and it seems like it answers the caching questions well. 

@jonpitch is going to PR on peekRecord.

ryanlabouve:
[...]

PR #4763, opened 2017-01-20
Document cache/request behavior of `query` #4733

ryanlabouve:
Basing this on https://github.com/emberjs/data/blob/master/addon/-private/system/store/finders.js#L144

Would love feedback on how this could be better.


igorT:
I would maybe just move it higher up, but seems good to me

bmac:
Thanks @ryanlabouve.

PR #4765, opened 2017-01-24
[DOC] Make model.unloadRecord public

knownasilya:
`store.unloadRecord` says that it's there for symmetry, meaning the model should have a similar public method.

cc @bmac

### DIFF
--- a/addon/-private/system/model/model.js
+++ b/addon/-private/system/model/model.js
@@ -587,7 +587,7 @@ const Model = Ember.Object.extend(Ember.Evented, {
   },
 
   /**
-    Unloads the record from the store. Only non-dirty records can be unloaded.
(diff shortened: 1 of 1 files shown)


addon/-private/system/model/model.js
```diff
@@ -587,8 +587,9 @@ const Model = Ember.Object.extend(Ember.Evented, {
   },
 
   /**
+    Unloads the record from the store. Only non-dirty records can be unloaded.
```
bmac: I'd remove the second sentence and add "This will cause the record to be destroyed and freed up for garbage collection.".

knownasilya:
Should I update that on the store as well?

bmac:
Thanks @knownasilya

PR #4767, opened 2017-01-25
Updating gitignore to remove .idea

pete-the-pete:
Wanted to add these so I don't accidentally commit them.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,7 +3,6 @@ bundle
 tmp/
 tests/source/
 dist/
-.idea
 *.iml
 
 benchmarks/results/*.json


jherdman:
Generally speaking one should add a global `.gitignore` for this sort of thing https://help.github.com/articles/ignoring-files/

locks:
I agree with @jherdman here. Ideally `.idea` wouldn't be in the gitignore either ;P

pete-the-pete:
Ah, makes sense.  Should I remove .idea? or just close this?

bmac:
@pete-the-pete lets remove `.idea`.

bmac:
Thanks @pete-the-pete.

PR #4769, opened 2017-01-25
[DOC release] Add missing docs for `preload` option.

GCheung55:


addon/-private/system/store.js
```diff
@@ -495,6 +495,27 @@ Store = Service.extend({
...
+
+    ```javascript
+    store.findRecord('comment', 2, { preload: { post: 1 } });
+    ```
+
+    If you have access to the post model you can also pass the model itself:
+
+    ```javascript
+    store.findRecord('post', 1).then(function (myPostModel) {
+      store.findRecord('comment', 2, {post: myPostModel});
```
locks: can you put some spaces inside the `{}` like in the first code block?

GCheung55: @locks Done and done!

bmac:
Thanks @GCheung55.

Issue #4764, opened 2017-01-24
Default date transform does not conform to ISO 8601 standard for dates with timezones


tomlagier:
This is the default date transform: https://github.com/emberjs/data/blob/master/addon/-private/ext/date.js#L34

Per https://www.w3.org/TR/NOTE-datetime, timezones should be specified as `TZD  = time zone designator (Z or +hh:mm or -hh:mm)`

The regex on the above line only parses Z or +hhmm or -hhmm - note the missing `:` after the last non-capture group (`(?:(\d{2}...`).

Looks like a simple mistake, but it breaks out of the box normalization of dates on servers with timezones. Let me know if you would like a PR to fix or if there is some context here that I'm missing.  In the meantime, we can create our own normalize function to properly create dates out of those strings.

bmac:
Hi @tomlagier. We would accept a pr to fix the issue if you have the time.

tomlagier:
PR is up at https://github.com/emberjs/data/pull/4771

bmac:
Closed via https://github.com/emberjs/data/pull/4771

PR #4771, opened 2017-01-25
[BUGFIX beta] Ensure ISO-8601 regex correctly matches timestamps 

tomlagier:
This fixes the case that a timestamp with a timezone offset matches the regex and correctly calls Date.UTC rather than native Date.parse.

See https://github.com/emberjs/data/issues/4764

### DIFF
--- a/tests/unit/transform/date-test.js
+++ b/tests/unit/transform/date-test.js
@@ -46,9 +46,19 @@ test("#deserialize with different offset formats", function(assert) {
   var dateStringColon = '2013-03-15T23:22:00.000+00:00';
   var dateStringShortOffset = '2016-12-02T17:30:00.000+00';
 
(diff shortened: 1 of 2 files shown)


tests/unit/transform/date-test.js
```diff
@@ -46,9 +46,19 @@ test("#deserialize with different offset formats", function(assert) {
   var dateStringColon = '2013-03-15T23:22:00.000+00:00';
   var dateStringShortOffset = '2016-12-02T17:30:00.000+00';
 
+  assert.expect(6);
+
+  var _dateUTC = Date.UTC;
```
fivetanley: we should wrap this in a `try`/`finally` so that the original Date.UTC always gets restored.

PR #4775, opened 2017-01-26
[DOC] shorthand methods for transform

knownasilya:


addon/transform.js
```diff
@@ -14,10 +14,11 @@ import Ember from 'ember';
 
   // Converts centigrade in the JSON to fahrenheit in the app
   export default DS.Transform.extend({
-    deserialize: function(serialized, options) {
+    deserialize(serialized, options) {
       return (serialized *  1.8) + 32;
     },
-    serialize: function(deserialized, options) {
+
```
locks: Trailing whitespace ;/

PR #4778, opened 2017-01-26
[DOC] shorthand methods and quotes for json-api serializer

knownasilya:


locks:
Can't kick the AppVeyor build cc @igorT

addon/serializers/json-api.js
```diff
@@ -27,7 +27,7 @@ const dasherize = Ember.String.dasherize;
   ```js
```
locks: can you update this with the actual file path, like the rest?

addon/serializers/json-api.js
```diff
@@ -24,30 +24,32 @@ const dasherize = Ember.String.dasherize;
 
   This serializer normalizes a JSON API payload that looks like:
 
-  ```js
+  ```app/models/player.js
+  import DS from 'ember-data';
 
-    // models/player.js
-    import DS from "ember-data";
+  const { attr, belongsTo } = DS;
```
locks: can you revert the destructuring? ;X

knownasilya: k :)

knownasilya: Seems inconsistent, since they are used in other examples. Should those be removed?

locks: Hm, I personally think so, but we should get some data team consensus.

locks: (I thought the docs didn't use destructuring, like Ember/Guides)

PR #4789, opened 2017-02-03
Do not access container if Ember.getOwner exists.

bmac:
Fixes broken tests on master


bmac:
Thanks for the review @rwjblue.

Issue #4795, opened 2017-02-06
unloadAll no longer removes records from the store


trumb1mj:
In 2.13.0-canary+8ae5670c08 there appears be a regression.  Running unloadAll on vanilla, unrelated data no longer removes records from the store.

sly7-7:
Closing, see https://github.com/emberjs/data/pull/4796#issuecomment-281282668

PR #4796, opened 2017-02-06
Added unloadAll unit test to show behavior of ember data not removing…

trumb1mj:
… records from the store.  Related to #4795

### DIFF
--- a/tests/unit/store/unload-test.js
+++ b/tests/unit/store/unload-test.js
@@ -67,7 +67,7 @@ testInDebug("unload a dirty record asserts", function(assert) {
 });
 
 test('unload a record', function(assert) {
-  assert.expect(2);
+  assert.expect(4);
 
   run(function() {
     store.push({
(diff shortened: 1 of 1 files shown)


tests/unit/store/unload-test.js
```diff
@@ -79,21 +79,62 @@ test('unload a record', function(assert) {
...
+    });
+
+    store.findRecord('record', 1).then(function(record) {
+      assert.equal(get(record, 'id'), 1, "found record with id 1");
+    });
+
+    var results = store.peekAll("record")
+    assert.equal(results.get("length"), 1, "found record with peekAll");
+
+    run(function() {
```
sly7-7: There is a problem here creating a nested run loop.
Rewriting this test like this makes it pass
BTW, for me you examples describe the same behaviors a in https://github.com/emberjs/data/blob/master/tests/integration/records/unload-test.js#L85
```
test('unloadAll records', function(assert) {
  assert.expect(3);

  run(function() {
    store.push({
      data: {
        type: 'record',
        id: '1',
        attributes: {
          title: 'toto'
        }
      }
    });
  });

  let record = store.peekRecord('record', 1);
  assert.equal(get(record, 'id'), 1, "found record with id 1");

  var results = store.peekAll("record")
  assert.equal(results.get("length"), 1, "found record with peekAll");

  run(() => {
    store.unloadAll("record");
  });

  assert.equal(results.get("length"), 0, "not found record with peekAll");
});
```

sly7-7:
@trumb1mj Thank you for your contribution, since the tests you add here are already described in this file https://github.com/emberjs/data/blob/master/tests/integration/records/unload-test.js I'm just closing this issue. Please reopen if you think I'm wrong.

runspired:
@sly7-7 there is an issue here in that peekAll is not properly clearing, haven't totally figured out a fix for it yet though.

sly7-7:
@runspired by `peekAll` you meant `unloadAll`, or I definitely miss something here

runspired:
@sly7-7 unloadAll does not properly unload peekAll arrays atm on canary, still digging into why exactly but it seems to be that the work on dematerialization uncovered that we were doing a poor job of removing deleted records from live arrays.

sly7-7:
@runspired I can't reproduce something in twiddle (ember-data canary not working), but in my app, using ember-data canary, I encounter the following case, playing in the console:

crs = store.peekAll('concept-relation');

crs.get('length');  // 59

Ember.run(() => { $E.unloadAll('concept-relation'); }) // crashes here: https://github.com/emberjs/data/blob/master/addon/-private/system/relationships/state/has-many.js#L215

The manyArray still has a length, but records are null.

Not sure if this is related to the issue you mention, but definitely a bug too

PR #4800, opened 2017-02-10
failing test for saving new record

fivetanley:
The error message says

> Your car record was saved to the server, but the response does not have an id and no id has been set client side. Records must have ids. Please update the server response to provide an id in the response or generate the id on the client side either before saving the record or while normalizing the response

But unfortunately no serializer methods are called. 🤔


fivetanley:
After talking with @runspired, `normalizeResponse` should be called here (and is).

Issue #4807, opened 2017-02-20
Dependency Injections are broken when using ember 2.12 and newer.


workmanw:
Originally filed as: https://github.com/emberjs/ember.js/issues/14941
Ember-Release (2.11.x): [Working Twiddle](https://ember-twiddle.com/e3a9f016848e28174059c52fe42944d2/75f5cfe06f78f0fee31f747ce66b5963e001d31e?openFiles=templates.application.hbs%2C)
Ember-Beta (2.12.x): [Broken Twiddle](https://ember-twiddle.com/e3a9f016848e28174059c52fe42944d2/7f6eb3760848bb88616fef26e854dc8fc3a055fe?openFiles=templates.application.hbs%2C)

It seems that using `owner.inject` targeting a `model` type is broken w/ ember-beta. Example:

```javascript
// app/initializers/i18n.js
export default {
  name: 'i18n',
  after: 'ember-i18n',
  initialize: function(app) {
    app.inject('model', 'i18n', 'service:i18n')
  }
};
```

Additionally, other ember types, such as component and controller, work as expected.

--------

@rwjblue helped me on slack to track this down a bit. We believe the cause of the problem lies with the implementation of `modelForFactory` ([store#modelForFactory](https://github.com/emberjs/data/blob/e9704ca5efa6e5bd8e0ca9e1b4014c9a33db386e/addon/-private/system/store.js#L2123-L2127)).

Prior to Ember 2.11, ember-data used `_lookupFactory` and internally that would do all of the dependency injections via an expensive `.extend()` process. One benefit of `factoryFor` is that it does not do this and thus saves perf. But the consequence is that when instantiating a class extracted from a "factory manager" those injections still need to be setup.

rwjblue:
Here are a few of the steps on the code safari that @workmanw and I just went through:

* [store.modelForFactory](https://github.com/emberjs/data/blob/e9704ca5efa6e5bd8e0ca9e1b4014c9a33db386e/addon/-private/system/store.js#L2123-L2127) returns the raw `.class` property instead of the full result from `owner.factoryFor`
* [InternalModel#getRecord](https://github.com/emberjs/data/blob/04b53683bbfd22b6840bc4c7d5c4c0c45719cdef/addon/-private/system/model/internal-model.js#L333) calls `._create` on the result of `store.modelFactoryFor` to instantiate the model instance.  Unfortunately, this `._create` is completely unaware of injections and is simply calling `._create` on the raw unmodified `.class`.

[...]

PR #4808, opened 2017-02-20
Avoid mutating model factory in _modelForMixin.

rwjblue:
With the upcoming `factory-for` work it is likely that mutating the result of `modelFactoryFor` will throw an error.  This is a minor refactoring to avoid needing to mutate the return value (since we can simply do the work before registering).

Addresses one of the issues mentioned in https://github.com/emberjs/data/issues/4807#issuecomment-281091810.


bmac:
Thanks @rwjblue.

PR #4809, opened 2017-02-21
bump ember-cli

sly7-7:
Unfortunately, for now ember-cli-qunit must stay on 2.x. I think we should move from ember-dev to ember-qunit-assert-helpers


btecu:
`ember-cli-eslint` is used, so you don't need those `.jshintrc` files.

bmac:
Thanks @sly7-7.

PR #4810, opened 2017-02-21
[Fixes #4807] realize class + factory seperation

### LINKED ISSUE #4807: Dependency Injections are broken when using ember 2.12 and newer.

stefanpenner:
*note: this combined with embers factoryFor makes `MODEL_FACTORY_INJECTIONS` irrelevant.*

Fixes https://github.com/emberjs/data/issues/4807

### DIFF
--- a/addon/-private/system/store.js
+++ b/addon/-private/system/store.js
@@ -2087,6 +2087,12 @@ Store = Service.extend({
     @private
    */
   _modelFor(modelName) {
+    let maybeFactory = this._modelFactoryFor(modelName);
+    // for factorFor factory/class split
+    return maybeFactory.class ? maybeFactory.class : maybeFactory;
+  },
+
+  _modelFactoryFor(modelName) {
     heimdall.increment(modelFor);
     let factory = this._modelClassCache[modelName];
 
@@ -2101,9 +2107,13 @@ Store = Service.extend({
         throw new EmberError(`No model was found for '${modelName}'`);
       }
 
-      assert(`'${inspect(factory)}' does not appear to be an ember-data model`, (typeof factory._create === 'function') );
+      // interopt with the future
+      let klass = getOwner(this).factoryFor ? factory.class : factory;
 
-      factory.modelName = factory.modelName || modelName;
+      assert(`'${inspect(klass)}' does not appear to be an ember-data model`, klass.isModel);
+
+      // TODO: deprecate this
+      klass.modelName = klass.modelName || modelName;
 
       this._modelClassCache[modelName] = factory;
     }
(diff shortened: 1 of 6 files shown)


addon/-private/system/store.js
```diff
@@ -2101,9 +2107,13 @@ Store = Service.extend({
...
 
-      assert(`'${inspect(factory)}' does not appear to be an ember-data model`, (typeof factory._create === 'function') );
+      // interopt with the future
+      let klass = getOwner(this).factoryFor ? factory.class : factory;
+
+      assert(`'${inspect(klass)}' does not appear to be an ember-data model`, klass.isModel);
 
-      factory.modelName = factory.modelName || modelName;
+      // TODO: deprecate this
+      klass.modelName = klass.modelName || modelName;
```
chadhietala: Can you do `Ember.deprecate` instead of the TODO?

stefanpenner: We don't currently have a path forward yet. I would prefer to deprecate once this is the case.

stefanpenner: But will find a path forward for this soon!

stefanpenner: @chadhietala I'm doing a follow up PR that reduces ember-datas dependence on `modelName`. That will be a pre-cursor to deprecating it.

chadhietala: SG

stefanpenner: i've removed many more internal usages of `modelName`, there are some more low hanging fruit but addressing it fully will be some effort (although I hope we can do asap). 

Some notes: https://github.com/emberjs/data/issues/4814

PR #4813, opened 2017-02-22
misc cleanup

stefanpenner:
* avoid needly internalModel -> type conversion, when the information is already present on the internalModel
* make asserts more consistent (ES6 interpolation etc)
* rename trueModelName to normalizedModelName

### DIFF
--- a/addon/-private/system/store.js
+++ b/addon/-private/system/store.js
@@ -340,9 +340,9 @@ Store = Service.extend({
     @return {DS.Model} record
   */
   createRecord(modelName, inputProperties) {
-    assert("You need to pass a model name to the store's createRecord method", isPresent(modelName));
-    assert(`Passing classes to store methods has been removed. Please pass a dasherized string instead of ${inspect(modelName)}`, typeof modelName === 'string');
-    let trueModelName = this._classKeyFor(modelName);
+    assert(`You need to pass a model name to the store's createRecord method`, isPresent(modelName));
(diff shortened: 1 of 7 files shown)


addon/-private/system/store.js
```diff
@@ -746,7 +746,7 @@ Store = Service.extend({
 
     let fetchedInternalModel = this._findEmptyInternalModel(internalModel, options);
 
-    return promiseRecord(fetchedInternalModel, "DS: Store#findRecord " + internalModel.typeKey + " with id: " + get(internalModel, 'id'));
+    return promiseRecord(fetchedInternalModel, `DS: Store#findRecord ${internalModel.modeName} with id: ${internalModel.id}`);
```
runspired: modelName* not modeName

addon/-private/system/store.js
```diff
@@ -820,13 +820,15 @@ Store = Service.extend({
...
-    let resolver = RSVP.defer('Fetching ' + modelClass.modelName + ' with id: ' + internalModel.id);
+    let { id, modelName } = internalModel;
+    let resolver = RSVP.defer(`Fetching ${modelName}' with id: ${id}`);
     let pendingFetchItem = {
       internalModel,
       resolver,
       options
     };
+
+    let modelClass = internalModel.type; // TODO: is this needed?
```
runspired: no, this can be refactored to `modelName`.  If so, `this._pendingFetch` should be refactored as well as we don't beed it to expecting `modelClass`

stefanpenner: Ya likely later PR, i think the review gods will be sad with the size of this one.

PR #4821, opened 2017-02-24
Pay go setup relationships

hjdivad:
The goal here is to do as little work as we can with relationship payloads until the relationship is actually accessed.  In an extreme case, where many relationship payloads are pushed, but no relationships are accessed, all we want to do is stash the payload.

So what we want is
  - to spend less time on relationships we don't access
  - to not spend more time (even if we spend it later on) for relationships we do access

Testing on my early 2015 13" macbook pro I get a substantial improvement on pushing relationships that aren't accessed.  

----

For my test case of 2800 models and 4800 relationships lazy relationships is about 2.2x as fast when no relationship is accessed.

master
master-lazy

pr
pr-lazy

----

For my test case of 2800 models and 4800 relationships lazy relationships costs about as much as master

master
master-eager-push
master-eager-expand

pr
pr-eager-push
pr-eager-expand


tests/integration/relationships/inverse-relationships-test.js
```diff
@@ -463,8 +463,8 @@ testInDebug("Inverse relationships that don't exist throw a nice error for a bel
   }, /We found no inverse relationships by the name of 'testPost' on the 'user' model/);
 });
 
-test("inverseFor is only called when inverse is not null", function(assert) {
-  assert.expect(2);
+test("inverseFor short-circuits when inverse is null", function(assert) {
```
stefanpenner: Now that we have both `inverseFor` and `_inverseFor` should we test both of them?

addon/-private/system/model/internal-model.js
```diff
@@ -505,6 +505,7 @@ export default class InternalModel {
     assert("Cannot destroy an internalModel while its record is materialized", !this.record || this.record.get('isDestroyed') || this.record.get('isDestroying'));
 
     this.store._removeFromIdMap(this);
+    this.store._relationshipsPayloads.unload(this.modelName, this.id);
```
stefanpenner: should we unify 507 and 508 into `this.store._internalModelDestroyed(internalModel)` that way, we don't need to leak `_removeFromIdMap` and `_relationshipsPayload` private members to `internalModel`

addon/-private/system/model/internal-model.js
```diff
@@ -518,6 +519,8 @@ export default class InternalModel {
 
   setupData(data) {
     heimdall.increment(setupData);
+    this.store._relationshipsPayloads.push(this.modelName, this.id, data.relationships);
```
stefanpenner: thoughts on `_internalModelHasData`  or something?

addon/-private/system/model/internal-model.js
```diff
@@ -988,6 +991,8 @@ export default class InternalModel {
   */
   adapterDidCommit(data) {
     if (data) {
+      this.store._relationshipsPayloads.push(this.modelName, this.id, data.relationships);
```
stefanpenner: https://github.com/emberjs/data/pull/4821/files#r104052721

addon/-private/system/relationships/relationships-payloads.js
```diff
@@ -0,0 +1,446 @@
...
+            type: 'user'
+          }
+        },
+        // tells us this payload was populated from the inverse side; so if
+        // we're a many side of a relationship we are not fully loaded
+        inverse: true,
+      }
+    }
+
+  @class RelationshipsPayloads
```
stefanpenner: likely should be marked as `@private`

addon/-private/system/relationships/relationships-payloads.js
```diff
@@ -0,0 +1,446 @@
...
+  }
+
+  /**
+    Find the invese of a given relationship.
+
+    @return {[String, String]} A tuple of the inverse model and inverse relationship names
+  */
+  _inverseFor(modelName, relationshipName) {
+    // TODO: lightschema this
+    let modelClass = this._store._modelFor(modelName);
```
stefanpenner: should this by `modelFor`?

addon/-private/system/relationships/relationships-payloads.js
```diff
@@ -0,0 +1,446 @@
...
+          return true;
+        }
+      }
+      return false;
+    } else {
+      return this._hasRecordForId(this._store, data.type, data.id);
+    }
+  }
+
+  _hasRecordForId(store, type, id) {
```
stefanpenner: this method does not appear to rely on `this`, we should likely move it off the prototype and just to a module local helper function. Or, we should make it use `this._store` internally. 

When I see methods on `this` I feel it suggests it needs access to `this`.

addon/-private/system/relationships/relationships-payloads.js
```diff
@@ -0,0 +1,446 @@
...
+      } else {
+        inversePayloads[inverseId] = inverseEntry;
+      }
+    }
+  }
+
+  get _lhsRelationshipIsMany() {
+    if (this.__lhsRelatinoshipIsMany === undefined) {
+      if (this._lhsModelName) {
+        let modelClass = this._store.modelFor(this._lhsModelName);
```
stefanpenner: I believe both LHS and RHS modelClass's are loaded when RelationshipPayload is instantiated, should we just pass them in at construction? At that point, do we already have the relationship descriptor as well, or is that what we want lazy?

addon/-private/system/relationships/relationships-payloads.js
hjdivad: 👍

addon/-private/system/relationships/relationships-payloads.js
```diff
@@ -0,0 +1,446 @@
...
+        inverse: true,
+      }
+    }
+
+  @class RelationshipsPayloads
+*/
+export default class RelationshipsPayloads {
+  constructor(store) {
+    this._store = store;
+    this._map = {};
```
hjdivad: 👍

addon/-private/system/relationships/relationships-payloads.js
```diff
@@ -0,0 +1,446 @@
...
+        user: DS.belongsTo('user')
+      });
+
+      relationshipPayloads.get('user', 'hobbies') === relationshipPayloads.get('hobby', 'user');
+
+    @private
+  */
+  _getRelationshipPayloads(modelName, relationshipName) {
+    // TODO: lightschema this
+    let modelClass = this._store._modelFor(modelName);
```
hjdivad: It's strictly internal, so by the time we get here `modelName` should already be normalized.  I'm fine with normalizing here (ie calling `this._store.modelFor` instead of `this._store._modelFor`) for simplicity's sake.

hjdivad: 👍

addon/-private/system/relationships/relationships-payloads.js
hjdivad: https://github.com/emberjs/data/pull/4821#discussion_r104754041

addon/-private/system/relationships/relationships-payloads.js
hjdivad: 👍

addon/-private/system/relationships/relationships-payloads.js
hjdivad: > I believe both LHS and RHS modelClass's are loaded when RelationshipPayload is instantiated, should we just pass them in at construction? At that point, do we already have the relationship descriptor as well, or is that what we want lazy?

This is not necessarily the case.  `inverseFor` does not load the inverse class.  Even when we check for the inverse relationship, we won't cause the inverse class to be loaded.

tests/integration/relationships/inverse-relationships-test.js
hjdivad: Looking more closely, we don't need to introduce `_inverseFor`; the point was just to separate the API from the work so we can test that the expensive path is not taken, but this separation already exists between `inverseFor` and `_findInverseFor`, so I've updated the tests to assert that `inverseFor` short-circuits `_findInverseFor` rather than `_inverseFor`.

addon/-private/system/model/internal-model.js
hjdivad: if we do that should we also add `_pushRelationshipPayload` to avoid the Law of Demeter on `_relationshipsPayloads.push`?

addon/-private/system/model/internal-model.js
hjdivad: seems good.  motivation is to avoid the Law of Demeter violation presumably

addon/-private/system/model/internal-model.js
hjdivad: 👍

addon/-private/system/relationships/relationships-payloads.js
hjdivad: > This is not necessarily the case. inverseFor does not load the inverse class. Even when we check for the inverse relationship, we won't cause the inverse class to be loaded.

Okay what I said is just wrong.  I check this but obviously did not do so correctly.  I'll push the classes down to `RelationshipPayloads`

hjdivad: I'm just passing in the relationship meta since all we want off of it is to know whether or not this side of the relationship is a many side.

hjdivad:
@stefanpenner i've responded to your review + done some additional perf tweaks.  I've kept the changes separate, in https://github.com/emberjs/data/pull/4821/commits/c53997e46e7429b761809fa5e8fc49343489e992 if you want to look at only the changes

addon/-private/system/model/model.js
```diff
@@ -1308,27 +1308,21 @@ Model.reopenClass({
...
-    if (options && options.inverse === null) {
-      return null;
-    }
-
-    return this._inverseFor(name, store);
-  },
-
-  _inverseFor(name, store) {
     let inverseMap = get(this, 'inverseMap');
     if (inverseMap[name]) {
```
stefanpenner: should this also check for `null` and if it is `null` return `null`, rather then failing this branch and looking it up again on 1315 only to find `null` again?

addon/-private/system/relationships/state/create.js
```diff
@@ -57,23 +57,19 @@ export default class Relationships {
     if (!relationship) {
       let relationshipsByName = get(internalModel.type, 'relationshipsByName');
       let rel = relationshipsByName.get(key);
-      let relationshipPayloadEntry = internalModel.store._relationshipsPayloads.get(internalModel.modelName, internalModel.id, key);
+      let relationshipPayload = internalModel.store._relationshipsPayloads.get(internalModel.modelName, internalModel.id, key);
 
-      if (rel) {
-        relationship = relationships[key] = createRelationshipFor(internalModel, rel, internalModel.store);
-      }
+      if (!rel) { return undefined; }
```
stefanpenner: should we move this return to before line 60, that way we don't bother getting the relationshipPayload at all?

addon/-private/system/relationships/state/create.js
```diff
@@ -57,23 +57,19 @@ export default class Relationships {
...
 
-      if (relationshipPayloadEntry !== undefined) {
-        let { payload, inverse } = relationshipPayloadEntry;
-        if (!inverse) {
-          relationship.push(payload);
-        } else {
-          relationship.pushFromInverse(payload);
-        }
+      if (relationshipPayload) {
+        this.internalModel.store._backburner.join(function() {
```
stefanpenner: `=>`

stefanpenner: i'm surprised this `join` is required, whats it all about?

addon/-private/system/relationships/state/relationship.js
```diff
@@ -155,19 +157,31 @@ export default class Relationship {
...
     }
     this.flushCanonicalLater();
     this.setHasData(true);
   }
 
+  setupInverseRelationship(internalModel, initial) {
+    if (this.inverseKey) {
+      let relationships = internalModel._relationships;
+      let relationshipExisted = relationships.has(this.inverseKey);
+      let relationship = relationships.get(this.inverseKey);
```
stefanpenner: should we move this line to 72, that way if the branch is not taken we don't lazily create the relationship at all.

tests/integration/relationships/has-many-test.js
```diff
@@ -2839,6 +2839,7 @@ test("deleted records should stay deleted", function(assert) {
     message.destroyRecord()
   });
 
+  self.halt=true;
```
stefanpenner: ^

tests/integration/relationships/has-many-test.js
```diff
@@ -2853,7 +2854,12 @@ test("deleted records should stay deleted", function(assert) {
         }
       }]
     });
+    self.halt=false;
```
stefanpenner: ^

addon/-private/system/model/model.js
hjdivad: 👍

addon/-private/system/relationships/state/create.js
hjdivad: 👍

addon/-private/system/relationships/state/create.js
hjdivad: This is the thing that batches canonical flushes.  In master this batching happens [within _push](https://github.com/emberjs/data/blob/03440c14e14ab7923ae2ed7db8fbbec6d3fa89e6/addon/-private/system/store.js#L2315-L2342)

tests/integration/relationships/has-many-test.js
hjdivad: 👍

tests/integration/relationships/has-many-test.js
hjdivad: 👍

addon/-private/system/model/internal-model.js
```diff
@@ -518,6 +518,8 @@ export default class InternalModel {
 
   setupData(data) {
     heimdall.increment(setupData);
+    this.store._internalModelHasRelationshipData(this.modelName, this.id, data.relationships);
```
igorT: This sounds like a boolean returning method, maybe something like received or process or something would be better?

addon/-private/system/model/model.js
```diff
@@ -1312,9 +1312,17 @@ Model.reopenClass({
...
       return inverseMap[name];
     } else {
-      let inverse = this._findInverseFor(name, store);
-      inverseMap[name] = inverse;
-      return inverse;
+      let relationship = get(this, 'relationshipsByName').get(name);
+      if (!relationship) {
+        return inverseMap[name] = null;
```
igorT: Can we not assign & return at the same time

addon/-private/system/model/model.js
```diff
@@ -1312,9 +1312,17 @@ Model.reopenClass({
...
+      }
+
+      let options = relationship.options;
+      if (options && options.inverse === null) {
```
igorT: Can you leave a comment here please

addon/-private/system/relationships/relationships-payloads.js
```diff
@@ -0,0 +1,417 @@
+import Ember from 'ember';
+let get = Ember.get;
+
+/**
+  Manages relationship payloads for a given store.  Acts as a single source of
```
igorT: Can you expand on this comment to say that this is only until the relationship is initialized

addon/-private/system/relationships/relationships-payloads.js
```diff
@@ -0,0 +1,417 @@
...
+          }
+        }
+      }
+
+    @method
+  */
+  get(modelName, id, relationshipName) {
+    let modelClass = this._store._modelFor(modelName);
+    let relationshipsByName = get(modelClass, 'relationshipsByName');
+    let relationshipPayloads = this._getRelationshipPayloads(modelName, relationshipName, modelClass, relationshipsByName, false);
```
igorT: Can you leave a comment as to why this signature is so long

addon/-private/system/relationships/relationships-payloads.js
```diff
@@ -0,0 +1,417 @@
...
+      relationshipsPayloads.push('user', 1, userPayload.data.relationships);
+
+    @method
+  */
+  push(modelName, id, relationshipsData) {
+    if (!relationshipsData) { return; }
+
+    let modelClass = this._store._modelFor(modelName);
+    let relationshipsByName = get(modelClass, 'relationshipsByName');
+    for (let key in relationshipsData) {
```
igorT: Is this safe to do?

addon/-private/system/relationships/relationships-payloads.js
```diff
@@ -0,0 +1,417 @@
...
+      let relationshipPayloads = this._getRelationshipPayloads(modelName, key, modelClass, relationshipsByName, true);
```
igorT: Maybe passing the relationship would simplify things?

addon/-private/system/relationships/relationships-payloads.js
```diff
@@ -0,0 +1,417 @@
...
+
+    let key = `${modelName}:${relationshipName}`;
+    if (!this._cache[key] && init) {
+      return this._initializeRelationshipPayloads(modelName, relationshipName, modelClass, relationshipsByName);
+    }
+
+    return this._cache[key];
+  }
+
+  _initializeRelationshipPayloads(modelName, relationshipName, modelClass, relationshipsByName) {
```
igorT: This method could use comments or breaking apart

addon/-private/system/relationships/relationships-payloads.js
```diff
@@ -0,0 +1,417 @@
...
+    this._store = store;
+    this._lhsModelName = modelName;
+    this._lhsRelationshipName = relationshipName;
+    this._lhsRelationshipMeta = relationshipMeta;
+    this._rhsModelName = inverseModelName;
+    this._rhsRelationshipName = inverseRelationshipName;
+    this._rhsRelationshipMeta = inverseRelationshipMeta;
+
+    this._lhsPayloads = {};
+    if (modelName !== inverseModelName || relationshipName !== inverseRelationshipName) {
```
igorT: Can you explain the goal behind this conditional

addon/-private/system/relationships/relationships-payloads.js
```diff
@@ -0,0 +1,417 @@
...
+      this._rhsPayloads = {};
+      this._isReflexive = false;
+    } else {
+      // Edge case when we have a reflexive relationship to itself
+      //  eg user hasMany friends inverse friends
+      this._rhsPayloads = this._lhsPayloads;
+      this._isReflexive = true;
+    }
+
+    // either canoical on push or pending & flush
```
igorT: This comment doesn't really tell me very much

addon/-private/system/relationships/relationships-payloads.js
```diff
@@ -0,0 +1,417 @@
...
+    this._pendingPayloads = [];
+  }
+
+  get(modelName, id, relationshipName) {
+    this._flushPending();
+
+    if (modelName === this._lhsModelName && relationshipName === this._lhsRelationshipName) {
```
igorT: Can you leave a comment explaining the logic behind mapping to lhs or rhs

addon/-private/system/relationships/relationships-payloads.js
```diff
@@ -0,0 +1,417 @@
...
+      return this._lhsPayloads[id];
+    } else {
```
igorT: let's add a debug check that asserts this is actually === rhsRelationshipName

addon/-private/system/relationships/relationships-payloads.js
```diff
@@ -0,0 +1,417 @@
...
+  }
+
+  push(modelName, id, relationshipName, relationshipData) {
+    this._pendingPayloads.push([modelName, id, relationshipName, relationshipData]);
+  }
+
+  unload(modelName, id, relationshipName) {
+    this._flushPending();
+
+    if (modelName === this._lhsModelName && relationshipName === this._lhsRelationshipName) {
```
igorT: Maybe extract into a method?

addon/-private/system/relationships/relationships-payloads.js
```diff
@@ -0,0 +1,417 @@
...
+    if (!this._inverseLoaded(payloads[id])) {
+      this._removeInverse(id, payloads[id], inversePayloads);
+      delete payloads[id];
+    }
+  }
+
+  _flushPending() {
+    if (this._pendingPayloads.length === 0) { return; }
+
+    let work = this._pendingPayloads.splice(0, this._pendingPayloads.length);
```
igorT: can we rename this to payloadsToBeProcessed or something like that

addon/-private/system/relationships/relationships-payloads.js
```diff
@@ -0,0 +1,417 @@
...
+
+      // TODO: maybe delay this slightly?
+      let inverseRelationshipData = {
+        data: {
+          id: id,
+          type: modelName
+        }
+      }
+
+      if (modelName === this._lhsModelName && relationshipName === this._lhsRelationshipName) {
```
igorT: Can you leave a comment explaining the intent behind this blob

addon/-private/system/relationships/relationships-payloads.js
```diff
@@ -0,0 +1,417 @@
...
+        this._populateInverse(relationshipData, inverseRelationshipData, this._rhsPayloads, this._rhsRelationshipIsMany);
+      } else {
+        this._removeInverse(id, this._rhsPayloads[id], this._lhsPayloads);
+        this._rhsPayloads[id] = relationshipData;
+        this._populateInverse(relationshipData, inverseRelationshipData, this._lhsPayloads, this._lhsRelationshipIsMany);
+      }
+    }
+  }
+
+  _inverseLoaded(payload) {
```
igorT: Seems bad that the invariant from unloading where graphs are either unloaded together or not at all, is changed here

addon/-private/system/relationships/relationships-payloads.js
```diff
@@ -0,0 +1,417 @@
...
+          return true;
+        }
+      }
+      return false;
+    } else {
+      return hasRecordForId(this._store, data.type, data.id);
+    }
+  }
+
+  _populateInverse(relationshipData, inversePayload, inversePayloads, inverseIsMany) {
```
igorT: When we rename the classes, can we rename this arguments as well please

addon/-private/system/model/internal-model.js
hjdivad: `_internalModelDidGetGetRelationshipData`?

addon/-private/system/relationships/relationships-payloads.js
hjdivad: ah yeah, i'll switch this to `Object.keys().forEach`

addon/-private/system/relationships/relationships-payloads.js
hjdivad: unfortunately we'd need to also pass the inverse relationship, which we don't always need to compute

addon/-private/system/model/model.js
hjdivad: 👍

addon/-private/system/model/model.js
hjdivad: 👍

addon/-private/system/relationships/relationships-payloads.js
hjdivad: 👍

addon/-private/system/relationships/relationships-payloads.js
hjdivad: 👍

addon/-private/system/relationships/relationships-payloads.js
hjdivad: I've added some comments in the then clause to explain, as well as expanded the comments in the else clause.

```js
 // then clause

      // The common case of a non-reflexive relationship, or a reflexive
      // relationship whose inverse is not itself

// else clause

      // Edge case when we have a reflexive relationship to itself
      //  eg user hasMany friends inverse friends
      //
      //  In this case there aren't really two sides to the relationship, but
      //  we set `_rhsPayloads = _lhsPayloads` to make things easier to reason
      //  about

```

addon/-private/system/relationships/relationships-payloads.js
hjdivad: updated

addon/-private/system/relationships/relationships-payloads.js
hjdivad: 👍

addon/-private/system/relationships/relationships-payloads.js
hjdivad: 👍

addon/-private/system/relationships/relationships-payloads.js
hjdivad: 👍

addon/-private/system/relationships/relationships-payloads.js
hjdivad: sure

addon/-private/system/relationships/relationships-payloads.js
hjdivad: 👍

addon/-private/system/relationships/relationships-payloads.js
hjdivad: k i've renamed `inversePayloads` to `inverseIdToPayloads` here and elsewhere

addon/-private/system/relationships/relationship-payloads.js
```diff
@@ -0,0 +1,283 @@
...
+      } else {
+        assert(`${modelName}:${relationshipName} is not either side of this relationship, ${this._lhsModelName}:${this._lhsRelationshipName}<->${this._rhsModelName}:${this._rhsRelationshipName}`, this._isRHS(modelName, relationshipName));
+        previousPayload = this._rhsPayloads[id];
+        idToPayloads = this._rhsPayloads;
+        inverseIdToPayloads = this._lhsPayloads;
+        inverseIsMany = this._lhsRelationshipIsMany;
+      }
+
+      // actually flush this individual payload
+      //
```
igorT: Can you leave a comment saying why we remove and then repopulate

addon/-private/system/relationships/relationship-payloads.js
```diff
@@ -0,0 +1,283 @@
...
+
+  _populateInverse(relationshipData, inversePayload, inverseIdToPayloads, inverseIsMany) {
+    if (!relationshipData.data) {
+      // This id doesn't have an inverse, eg a belongsTo with a payload
+      //-{ data: null }, so there's nothing to populate
+      return;
+    }
+
+    if (Array.isArray(relationshipData.data)) {
+      for (let i=0; i<relationshipData.data.length; ++i) {
```
igorT: Lets look into using maps or something nicer here

addon/-private/system/relationships/relationship-payloads.js
```diff
@@ -0,0 +1,283 @@
...
+        this._addToInverse(inversePayload, inverseId, inverseIdToPayloads, inverseIsMany);
+      }
+    } else {
+      let inverseId = relationshipData.data.id;
+      this._addToInverse(inversePayload, inverseId, inverseIdToPayloads, inverseIsMany);
+    }
+  }
+
+  _addToInverse(inversePayload, inverseId, inverseIdToPayloads, inverseIsMany) {
+    if (this._isReflexive && inversePayload.data.id === inverseId) {
```
igorT: Is there a test for this case?

addon/-private/system/relationships/relationship-payloads.js
```diff
@@ -0,0 +1,283 @@
...
+        let inverseId = relationshipData.data[i].id;
+        this._addToInverse(inversePayload, inverseId, inverseIdToPayloads, inverseIsMany);
+      }
+    } else {
+      let inverseId = relationshipData.data.id;
+      this._addToInverse(inversePayload, inverseId, inverseIdToPayloads, inverseIsMany);
+    }
+  }
+
+  _addToInverse(inversePayload, inverseId, inverseIdToPayloads, inverseIsMany) {
```
igorT: Lets add a comment saying this is called by populateInverse/short description of what the purpose of it is

addon/-private/system/relationships/relationship-payloads.js
```diff
@@ -0,0 +1,283 @@
...
+    if (this._isReflexive && inversePayload.data.id === inverseId) {
+      // eg <user:1>.friends = [{ id: 1, type: 'user' }]
+      return;
+    }
+
+    let existingPayload = inverseIdToPayloads[inverseId];
+    let existingData = existingPayload && existingPayload.data;
+
+    if (Array.isArray(existingData)) {
```
igorT: Can we rewrite this to have the existence check upfront or add comments

addon/-private/system/relationships/relationship-payloads.js
```diff
@@ -0,0 +1,283 @@
...
+    this._store = store;
+    this._lhsModelName = modelName;
+    this._lhsRelationshipName = relationshipName;
+    this._lhsRelationshipMeta = relationshipMeta;
+    this._rhsModelName = inverseModelName;
+    this._rhsRelationshipName = inverseRelationshipName;
+    this._rhsRelationshipMeta = inverseRelationshipMeta;
+
+    // a map of id -> payloads for the left hand side of the relationship.
+    this._lhsPayloads = {};
```
igorT: Check if this should be object create null

addon/-private/system/relationships/relationship-payloads.js
```diff
@@ -0,0 +1,283 @@
...
+      inversePayload.data = data.filter((x) => x.id !== id);
+    } else {
+      inversePayloads[inverseId] = {
+        data: null
+      };
+    }
+  }
+}
+
+function hasRecordForId(store, type, id) {
```
igorT: In case this sticks around can we verify this is needed.

addon/-private/system/relationships/state/belongs-to.js
```diff
@@ -36,6 +36,18 @@ export default class BelongsToRelationship extends Relationship {
     this.flushCanonicalLater();
   }
 
+  setInitialCanonicalRecord(record) {
```
igorT: Can we move this to a separate PR

addon/-private/system/relationships/state/create.js
```diff
@@ -51,14 +51,22 @@ export default class Relationships {
...
-      if (rel) {
-        relationship = relationships[key] = createRelationshipFor(internalModel, rel, internalModel.store);
+      if (!rel) { return undefined; }
+
+      let relationshipPayload = internalModel.store._relationshipsPayloads.get(internalModel.modelName, internalModel.id, key);
+
+      relationship = relationships[key] = createRelationshipFor(internalModel, rel, internalModel.store);
+
+      if (relationshipPayload) {
+        this.internalModel.store._backburner.join(() => {
```
igorT: Check if this is needed

addon/-private/system/relationships/state/relationship.js
```diff
@@ -155,19 +157,31 @@ export default class Relationship {
...
-        }
-        record._implicitRelationships[this.inverseKeyForImplicit].addCanonicalRecord(this.record);
-      }
+      this.setupInverseRelationship(record, false);
     }
     this.flushCanonicalLater();
     this.setHasData(true);
   }
 
+  setupInverseRelationship(internalModel, initial) {
```
igorT: We don't need `initial` here

addon/-private/system/relationships/state/relationship.js
```diff
@@ -155,19 +157,31 @@ export default class Relationship {
...
+    if (this.inverseKey) {
+      let relationships = internalModel._relationships;
+      let relationshipExisted = relationships.has(this.inverseKey);
+      let relationship = relationships.get(this.inverseKey);
+      if (relationshipExisted || this.isPolymorphic) {
```
igorT: Can you leave comments explaining 
1. The laziness
2. why we are not lazy in the polymorphic case

addon/-private/system/relationships/state/relationship.js
```diff
@@ -155,19 +157,31 @@ export default class Relationship {
...
+        relationship.addCanonicalRecord(this.internalModel);
+      }
+    } else {
+      let relationships = internalModel._implicitRelationships;
+      let relationship = relationships[this.inverseKeyForImplicit];
+      if (!relationship) {
```
igorT: Same here

addon/-private/system/store.js
```diff
@@ -2403,14 +2412,15 @@ Store = Service.extend({
     heimdall.increment(_setupRelationships);
     let setupToken = heimdall.start('store._setupRelationships');
     let pushed = this._pushedInternalModels;
+    let modelNameToInverseMap = Object.create(null);
```
igorT: Lets see if there are perf benefits to caching this across runs

igorT: Can you leave a comment as to why we do this

addon/-private/system/store.js
```diff
@@ -2787,19 +2797,82 @@ function _commit(adapter, store, operation, snapshot) {
   }, label);
 }
 
-function setupRelationships(store, internalModel, data) {
+function inverseRelationshipInitialized(store, internalModel, data, key, modelNameToInverseMap) {
```
igorT: lets rename this to isInverseRelationshipInitialized

addon/-private/system/store.js
```diff
@@ -2787,19 +2797,82 @@ function _commit(adapter, store, operation, snapshot) {
...
+    // can't check inverse for eg { comments: { links: { related: URL }}}
+    return false;
+  }
+
+  let inverseMap = modelNameToInverseMap[internalModel.modelName]
+  if (!inverseMap) {
+    inverseMap = modelNameToInverseMap[internalModel.modelName] = get(internalModel.type, 'inverseMap');
+  }
+  let inverseData = inverseMap[key];
+  if (inverseData === undefined) {
```
igorT: lets rename this to something like `inverseRelationshipMetadata`/something not containing the word `Data`

addon/-private/system/store.js
```diff
@@ -2787,19 +2797,82 @@ function _commit(adapter, store, operation, snapshot) {
...
-  internalModel.type.eachRelationship((key, descriptor) => {
-    if (!data.relationships[key]) {
-      return;
-    }
+  let relationships = internalModel._relationships;
 
-    let relationship = internalModel._relationships.get(key);
-    relationship.push(data.relationships[key]);
-  });
+  for (let key in data.relationships) {
```
igorT: lets try switching this back to `eachRelationship`

addon/-private/system/relationships/relationship-payloads.js
hjdivad: 👍

addon/-private/system/relationships/relationship-payloads.js
hjdivad: https://jsperf.com/map-fn-vs-map-arrow-vs-fori/1

unless i've messed something up with the benchmarks, fori looks to be about 2x faster than map

addon/-private/system/relationships/relationship-payloads.js
hjdivad: already a unit test in relationship payloads, but i've added one for hasmany as well

addon/-private/system/relationships/relationship-payloads.js
hjdivad: 👍

addon/-private/system/relationships/relationship-payloads.js
hjdivad: sure

addon/-private/system/relationships/relationship-payloads.js
hjdivad: yep it should be; i've made the change

addon/-private/system/relationships/relationship-payloads.js
hjdivad: it's gone

addon/-private/system/relationships/state/belongs-to.js
hjdivad: ok

addon/-private/system/relationships/state/relationship.js
hjdivad: 👍

addon/-private/system/relationships/state/relationship.js
hjdivad: 👍

addon/-private/system/relationships/state/relationship.js
hjdivad: The semantics here have not changed at all: https://github.com/emberjs/data/blob/2edaf160a8a437cc989b2e6c980cc9c7d525c2c2/addon/-private/system/relationships/state/relationship.js#L160-L165

addon/-private/system/relationships/state/relationship.js
hjdivad: No; this PR preserves the invariant that if a relationship is initialized, its inverse is also initialized.

We could change the invariant, but it would add complexity in the case where a relationship A gets its initial state from the payload of its inverse B, B is initialized and then mutated before A is initialized.

This is something we could do but it seems better to leave that for a separate PR

addon/-private/system/store.js
hjdivad: 👍

addon/-private/system/store.js
hjdivad: 👍

addon/-private/system/store.js
hjdivad: yep this seems good; also with the setInitial work the backburner.join run in relationships/store/create seems unnecessary

addon/-private/system/relationships/state/belongs-to.js
hjdivad: moved to a separate commit within this PR in case we want to revert separately

hjdivad:
@igorT @stefanpenner this is ready to merge.

stefanpenner:
:shipit:

addon/-private/system/store.js
```diff
@@ -2791,14 +2806,81 @@ function _commit(adapter, store, operation, snapshot) {
...
 
-    let relationship = internalModel._relationships.get(key);
-    relationship.push(data.relationships[key]);
+    let relationshipRequiresNotification = relationships.has(relationshipName) ||
+      isInverseRelationshipInitialized(store, internalModel, data, relationshipName, modelNameToInverseMap);
+
+    if (relationshipRequiresNotification) {
+      let relationshipData = data.relationships[relationshipName];
+      relationships.get(relationshipName).push(relationshipData);
+    } else {
```
igorT: We were gonna remove this else right?

hjdivad: ah you're right, LMK push this fix

rlivsey:
Looks like this has broken for JSONAPI payloads which contain `links` for hasMany associations.

When accessing a relationship in a template I now get the following error, going to before this PR was merged works fine:

```
Assertion Failed: You modified "meeting.attendees" twice on <minutebase@model:meeting::ember1351:772c5e6e-aceb-11e6-ae79-770161ca087d> in a single render. It was rendered in "component:mb-pages/dashboard/meeting" and modified in "component:mb-pages/dashboard/meeting". This was unreliable and slow in Ember 1.x and is no longer supported. See https://github.com/emberjs/ember.js/issues/13948 for more details.
Error
    at assert (http://minutebase.dev:4202/assets/vendor.js:21770:13)
    at assert (http://minutebase.dev:4202/assets/vendor.js:33521:34)
    at Object.assertNotRendered (http://minutebase.dev:4202/assets/vendor.js:39296:11)
    at Object.propertyDidChange (http://minutebase.dev:4202/assets/vendor.js:37912:30)
    at Class.propertyDidChange (http://minutebase.dev:4202/assets/vendor.js:51064:19)
    at Class.notifyPropertyChange (http://minutebase.dev:4202/assets/vendor.js:51078:12)
    at InternalModel.notifyPropertyChange (http://minutebase.dev:4202/assets/vendor.js:140423:21)
    at ManyRelationship.updateLink (http://minutebase.dev:4202/assets/vendor.js:147451:19)
    at ManyRelationship.push (http://minutebase.dev:4202/assets/vendor.js:147529:16)
    at Relationships.get (http://minutebase.dev:4202/assets/vendor.js:146885:24)
```

Looks like `updateLink` is notifying property change and causing Glimmer to re-evaluate the relationship and so it explodes.

Tried to make a Twiddle, but the example Twiddle appears to be broken at the moment.

stefanpenner:
@rlivsey can you open a new issue so this isn't lost?

rlivsey:
@stefanpenner yup, was trying to make a repro first but will open an issue once I get the kids to bed!

PR #4823, opened 2017-02-25
[BUGFIX] ensure all of DS is initialized together

stefanpenner:
* DS object will always present fully. Today during initialization it remains partially initializers for a period (missing everything from https://github.com/emberjs/data/blob/master/addon/index.js#L91-L154)
* this also fixes the (unknown mixin) stuff for all DS classes

Although this does force the entire DS namespaces to be initialized immediately. This shouldn’t cause a performance issue, as it the full namespace was always initialized, just in two phases.

### DIFF
--- a/app/initializers/ember-data.js
+++ b/app/initializers/ember-data.js
@@ -1,5 +1,5 @@
 import setupContainer from 'ember-data/setup-container';
-import 'ember-data/-private/core';
+import 'ember-data/index';
 
 /*
 


stefanpenner:
strange the failure seems to be uglify related...

bmac:
@stefanpenner the build seems to pass with `"loader.js": "v4.1.0"` but get the uglify error with `v4.2.2`.

bmac:
More debugging has lead me to believe this line is what uglify is chocking on. https://github.com/ember-cli/loader.js/blob/master/lib/loader/loader.js#L7 I'm not sure why its getting run on the unprocessed loader.js thought. I think I'll lock Ember Data to using `v4.1.0` of loader.js for now.

stefanpenner:
@bmac im not sure what script is running that causes that issue (haven't had a chance to look). My guess is it is pulling in loader.js directly, rather then letting the addon provide it.

bmac:
Looks like you are right. I've opened a pr with a temporary fix https://github.com/emberjs/data/pull/4824.

stefanpenner:
@bmac rebased on-top of your fix.

PR #4833, opened 2017-03-01
upgrade qunit

stefanpenner:
This includes other nice things, but the major nice thing is a multi-select dropdown for modules. And the ability to run multiple modules in isolation, not just one at at time.

screen shot 2017-03-01 at 10 11 44 am

- [ ] `ember-dev` isn't yet QUnit 3 compat: https://github.com/emberjs/ember-dev/issues/168

### DIFF
--- a/package.json
+++ b/package.json
@@ -64,7 +64,7 @@
     "ember-cli-inject-live-reload": "^1.4.1",
     "ember-cli-internal-test-helpers": "^0.8.1",
     "ember-cli-pretender": "^1.0.1",
-    "ember-cli-qunit": "^2.1.0",
+    "ember-cli-qunit": "^3.1.1",
     "ember-cli-release": "^0.2.9",
     "ember-cli-shims": "^1.0.2",
     "ember-cli-sri": "^2.1.0",


workmanw:
I think one of the issues you might have is the https://github.com/emberjs/ember-dev library isn't really built for QUnit 2.0. I created an addon recently that I hope would be mostly a drop in replacement ([ember-qunit-assert-helpers](https://github.com/workmanw/ember-qunit-assert-helpers)). I'd be happy to collaborate on this PR. I think @sly7-7 may have brought this up in the slack #dev-ember-data channel.

sly7-7:
When I upgraded ember-cli, this was the only problem I had, and @workmanw just explained this.
Actually I think the first thing to do is to replace ember-dev by ember-qunit-assert-helpers. I would also work on this migration if we choose that way.
@stefanpenner do you agree ?

stefanpenner:
@sly7-7 yup sounds reasonable

bmac:
I'm going to close this. I believe it has been super seeded by https://github.com/emberjs/data/pull/5371.

PR #4836, opened 2017-03-03
cleanup + reject with useful reasons

stefanpenner:
* es6 cleanup
* promise cleanup
* remove dangling promises / errors
* throw useful error message when a payload does not include an expected record instead of `undefined`

### DIFF
--- a/addon/-private/system/store.js
+++ b/addon/-private/system/store.js
@@ -906,10 +906,11 @@ Store = Service.extend({
 
     function rejectInternalModels(internalModels, error) {
       for (let i = 0, l = internalModels.length; i < l; i++) {
-        let pair = seeking[internalModels[i].id];
+        let internalModel = internalModels[i];
+        let pair = seeking[internalModel.id];
 
         if (pair) {
-          pair.resolver.reject(error);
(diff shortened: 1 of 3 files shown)


stefanpenner:
Test failure is just on one of the appveyor things, and is unrelated to ember-data.

addon/-private/system/store.js
```diff
@@ -906,10 +906,11 @@ Store = Service.extend({
 
     function rejectInternalModels(internalModels, error) {
       for (let i = 0, l = internalModels.length; i < l; i++) {
-        let pair = seeking[internalModels[i].id];
+        let internalModel = internalModels[i];
+        let pair = seeking[internalModel.id];
 
         if (pair) {
-          pair.resolver.reject(error);
+          pair.resolver.reject(error || new Error(`Expected: '${internalModel}' to be present in the adapter provided payload, but it was not found.`));
```
stefanpenner: this is the only change to ember-data, previously `undefined` was throw, now it is an error with a actionable message.

bmac:
Thanks @stefanpenner.

PR #4838, opened 2017-03-04
Stefs cleanup (and some test suite memory leak fixes)

stefanpenner:
Noticed 8 containers leaked to the end of the test suite, this PR fixes 5 of them (2 of them are yet upstream). This leaves us with 1, and that one appears strange/tricky (will investigate next).

also found these leaks:

* https://github.com/tildeio/rsvp.js/pull/446 <-- caused us to leak 2 containers
* https://github.com/stefanpenner/es6-promise

---

The goal of this, is to be sure our own test suite can be used (at-least manually) to detect leaks/tests. So if we can have our test suite not leak, it will be easier to catch leaks going forward (like the 2 RSVP / ES6 promise leaks).

### DIFF
--- a/tests/integration/debug-adapter-test.js
+++ b/tests/integration/debug-adapter-test.js
@@ -8,10 +8,12 @@ var App, store, debugAdapter;
 var get = Ember.get;
 var run = Ember.run;
 
-module("DS.DebugAdapter", {
+module('DS.DebugAdapter', {
   beforeEach() {
     Ember.run(function() {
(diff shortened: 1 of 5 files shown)


stefanpenner:
The remaining leak may not be a leak on our end, as the profiler is unable to calculate path to the retainer...

screen shot 2017-03-03 at 4 40 10 pm

I've walk as many paths as i could manually, and inspected the code in question and cant seem to find a reasonable reason for it to still be around....

PR #4840, opened 2017-03-06
[DOC] Fix markup for example code of `DS.Model#inverseFor`

tricknotes:
At http://emberjs.com/api/data/classes/DS.Model.html#method_inverseFor

Before:

After:


locks:
Thanks!

PR #4842, opened 2017-03-08
Stable has many promise proxy

stefanpenner:
- [x] fixes memory leak: https://github.com/emberjs/data/issues/4826
- [x] prevents needless re-render, allowing glimmer to diff rather then constantly having to blow away `each`'s which `each` across relationships.

### DIFF
--- a/addon/-private/system/relationships/state/has-many.js
+++ b/addon/-private/system/relationships/state/has-many.js
@@ -10,6 +10,7 @@ export default class ManyRelationship extends Relationship {
     this.belongsToType = relationshipMeta.type;
     this.canonicalState = [];
     this.isPolymorphic = relationshipMeta.options.polymorphic;
+    this._manyArray = null;
   }
 
   get manyArray() {
--- a/addon/-private/system/relationships/state/has-many.js
+++ b/addon/-private/system/relationships/state/has-many.js
@@ -1,8 +1,8 @@
-import { assert, assertPolymorphicType } from "ember-data/-private/debug";
-import { PromiseManyArray, promiseManyArray } from "../../promise-proxies";
-import Relationship from "./relationship";
-import OrderedSet from "../../ordered-set";
-import ManyArray from "../../many-array";
+import { assert, assertPolymorphicType } from 'ember-data/-private/debug';
+import { PromiseManyArray } from '../../promise-proxies';
+import Relationship from './relationship';
+import OrderedSet from '../../ordered-set';
+import ManyArray from '../../many-array';
 
 export default class ManyRelationship extends Relationship {
   constructor(store, record, inverseKey, relationshipMeta) {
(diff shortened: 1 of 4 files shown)


stefanpenner:
cc @BryanCrotaz

addon/-private/system/relationships/state/has-many.js
```diff
@@ -12,6 +12,25 @@ export default class ManyRelationship extends Relationship {
...
+
+  get _loadingPromise() { return this.__loadingPromise; }
+  _updateLoadingPromise(promise, content) {
+    if (this.__loadingPromise) {
+      if (content) {
+        this.__loadingPromise.set('content', content)
+      }
+      this.__loadingPromise.set('promise', promise)
+    } else {
+      this.__loadingPromise = new PromiseManyArray({
```
sly7-7: should we call `PromiseManyArray.create()` or `new PromiseManyArray()` ?

tests/unit/promise-proxies-test.js
```diff
@@ -23,3 +23,90 @@ test('.reload should NOT leak the internal promise, rather return another promis
...
+
+test('.reload should be stable', function(assert) {
+  assert.expect(19);
+
+  let content = Ember.A();
+
+  content.reload = () => Ember.RSVP.Promise.resolve(content);
+  let promise = Ember.RSVP.Promise.resolve(content);
+
+  let array = DS.PromiseManyArray.create({
```
sly7-7: like above

addon/-private/system/relationships/state/has-many.js
```diff
@@ -12,6 +12,25 @@ export default class ManyRelationship extends Relationship {
     this.belongsToType = relationshipMeta.type;
     this.canonicalState = [];
     this.isPolymorphic = relationshipMeta.options.polymorphic;
+    this._manyArray = null;
+    this.__loadingPromise = null;
+  }
+
+  get _loadingPromise() { return this.__loadingPromise; }
```
bmac: Why have both `_loadingPromise` and `__loadingPromise`?

addon/-private/system/relationships/state/has-many.js
stefanpenner: Either should be fine

addon/-private/system/relationships/state/has-many.js
stefanpenner: Wanted to make sure it was readOnly without going through the update method. (So people don't accidentally bypass)

tests/unit/model/relationships/has-many-test.js
```diff
@@ -759,6 +759,62 @@ test("DS.hasMany is async by default", function(assert) {
...
+test('DS.hasMany is stable', function(assert) {
+  const Tag = DS.Model.extend({
+    name: DS.attr('string'),
+    people: DS.hasMany('person')
+  });
+
+  const Person = DS.Model.extend({
+    name: DS.attr('string'),
+    tag: DS.belongsTo('tag', { async: false })
+  });
```
hjdivad: Adding `toString`s on these makes debugging a lot nicer.

stefanpenner: I can update all of them in a later PR

PR #4847, opened 2017-03-08
Improve naming around internal constructs (record vs internal model)

stefanpenner:
* This was super confusing before
* This does not change any public APIs


stefanpenner:
@runspired / @hjdivad / @bmac / @igorT  r?

PR #4851, opened 2017-03-09
Don't calculate `_changedKeys` pointlessly

hjdivad:
If the internal model has no record, we won't do anything with the changes, so let's not waste cycles calculating them.

### DIFF
--- a/addon/-private/system/model/internal-model.js
+++ b/addon/-private/system/model/internal-model.js
@@ -518,9 +518,15 @@ export default class InternalModel {
 
   setupData(data) {
     heimdall.increment(setupData);
-    let changedKeys = this._changedKeys(data.attributes);
+    let changedKeys;
+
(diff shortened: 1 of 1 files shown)


addon/-private/system/model/internal-model.js
```diff
@@ -518,9 +518,15 @@ export default class InternalModel {
...
+
+    if (this.hasRecord) {
+      changedKeys = this._changedKeys(data.attributes);
+    }
+
     assign(this._data, data.attributes);
     this.pushedData();
+
     if (this.hasRecord) {
       this.record._notifyProperties(changedKeys);
```
stefanpenner: should we just move the above `this._changedKeys` to this conditional.

```js
this.record._notifyProperties(this._changedKeys(data.attributes);
```

PR #4852, opened 2017-03-11
[PERF] optimize updates to hasMany state

### LINKED ISSUE #4150: has-many cannonicalState refresh not optimal
Hello,

By looking at the code to find a bug, I stepped upon the `computeChange` which seems to do a lot of quite heavy work I'm not sure is really needed :

``` javascript
    for (i = 0; i < length; i++) {
      record = records[i];
      this.removeCanonicalRecord(record);
      this.addCanonicalRecord(record, i);
    }
```

If I understand correctly, this is to avoid having duplicate in the canonicalState, which is OK, but maybe there is a better way of doing than actually removing the record and adding it back in case of duplicate ? 

The main problem here is it will trigger most of the time (adding and removing references will do that), and it really seems overkill ( removeCanonicalRecord does a lot of things ). 

If is for sorting purpose, I think it should be not to hard to find another way.

BryanCrotaz:
resolves #4150

On master:
With 5000 records, adding 1000 on the end with a `store.push` takes 250ms in `has-many.computeChanges`
Removing them again takes another 250ms

With this PR:
adding takes 37ms
removing takes 9ms

### DIFF
--- a/addon/-private/system/diff-array.js
+++ b/addon/-private/system/diff-array.js
@@ -2,6 +2,7 @@
   @namespace
   @method diffArray
   @private
+  @for DS
   @param {Array} oldArray the old array
   @param {Array} newArray the new array
   @return {hash} {
--- a/addon/-private/system/relationships/state/has-many.js
+++ b/addon/-private/system/relationships/state/has-many.js
@@ -5,6 +5,8 @@ import Relationship from './relationship';
 import OrderedSet from '../../ordered-set';
 import ManyArray from '../../many-array';
 
+import diffArray from '../../diff-array';
+
 export default class ManyRelationship extends Relationship {
   constructor(store, internalModel, inverseKey, relationshipMeta) {
     super(store, internalModel, inverseKey, relationshipMeta);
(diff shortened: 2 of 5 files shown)


BryanCrotaz:
@stefanpenner you might like this one

stefanpenner:
Will try to review this evening thanks.

BryanCrotaz:
thoughts?

BryanCrotaz:
nudge

BryanCrotaz:
[nudge] @stefanpenner  @igorT @bmac

runspired:
@BryanCrotaz we should revisit this soon

runspired:
@BryanCrotaz main thing would be to rebase and see where we are (do we now fail tests for instance). I am very happy to help you keep this moving forward so that it does not stall again.

runspired:
I've rebased the commit, and I find that it happily fixes our `ASSERT KNOWN LIMITATION EXISTS` test for hasMany losing local changes after a delete.  That said, I'm not sure this approach will always cause us to keep local changes, and what's more I suspect this leaks `_removedInternalModels`.

I'm going to add some more tests and investigate those two points.

PR #4856, opened 2017-03-13
[BUGFIX #4497] query/queryRecord/filter now support adapter options

### LINKED ISSUE #4497: adapterOptions not usable for `query` and `queryRecord`
`adapterOptions` is a great way to communicate one-off settings to an adapter; however, it does not appear that it is usable for `query` or `queryRecord`.

[Here is a demonstration](https://ember-twiddle.com/f8542e541f70a3cfb213469edd862867?fileTreeShown=false&openFiles=routes.application.js%2C) on how passing `adapterOptions` into the  `query` argument for `store.query` and `store.queryRecord` will just result in the options being parameterized (which according to the docs is exactly what should be happening -- it mentions directly passing on anything present there).

I would assume there are a few paths forward:
- allow `adapterOptions` to be passed into the `query` argument for `store.query` and `store.queryRecord`, automatically strip that key and provide it to the snapshot, and then update the docs to document that special behavior
- allow a third `options` argument to `store.query` and `store.queryRecord`
- ????

stefanpenner:
[fixes #4497]

### DIFF
--- a/addon/-private/system/store.js
+++ b/addon/-private/system/store.js
@@ -1268,11 +1268,16 @@ Store = Service.extend({
     assert(`You need to pass a query hash to the store's query method`, query);
     assert(`Passing classes to store methods has been removed. Please pass a dasherized string instead of ${modelName}`, typeof modelName === 'string');
 
+    let options = {};
+
+    options.adapterOptions = query.adapterOptions;
+    delete query.adapterOptions;
+
     let normalizedModelName = normalizeModelName(modelName);
-    return this._query(normalizedModelName, query);
+    return this._query(normalizedModelName, query, null, options);
   },
 
-  _query(modelName, query, array) {
+  _query(modelName, query, array, options) {
     let token = heimdall.start('store._query');
     assert(`You need to pass a model name to the store's query method`, isPresent(modelName));
     assert(`You need to pass a query hash to the store's query method`, query);
@@ -1288,7 +1293,7 @@ Store = Service.extend({
     assert(`You tried to load a query but you have no adapter (for ${modelName})`, adapter);
     assert(`You tried to load a query but your adapter does not implement 'query'`, typeof adapter.query === 'function');
 
-    let pA = promiseArray(_query(adapter, this, modelName, query, array));
+    let pA = promiseArray(_query(adapter, this, modelName, query, array, options));
     instrument(() => {
       pA.finally(() => { heimdall.stop(token); });
     });
@@ -1400,10 +1405,15 @@ Store = Service.extend({
 
     let adapter = this.adapterFor(normalizedModelName);
 
+    let options = {};
+
+    options.adapterOptions = query.adapterOptions;
+    delete query.adapterOptions;
+
     assert(`You tried to make a query but you have no adapter (for ${normalizedModelName})`, adapter);
     assert(`You tried to make a query but your adapter does not implement 'queryRecord'`, typeof adapter.queryRecord === 'function');
 
-    return promiseObject(_queryRecord(adapter, this, modelName, query).then(internalModel => {
+    return promiseObject(_queryRecord(adapter, this, modelName, query, options).then(internalModel => {
       // the promise returned by store.queryRecord is expected to resolve with
       // an instance of DS.Model
       if (internalModel) {
--- a/addon/-private/system/store.js
+++ b/addon/-private/system/store.js
@@ -1261,20 +1261,22 @@ Store = Service.extend({
     @method query
     @param {String} modelName
     @param {any} query an opaque query to be used by the adapter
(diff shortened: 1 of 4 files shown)


runspired:
Thanks! Reviewing now.

addon/-private/system/store.js
```diff
@@ -1255,11 +1255,16 @@ Store = Service.extend({
     assert(`You need to pass a query hash to the store's query method`, query);
     assert(`Passing classes to store methods has been removed. Please pass a dasherized string instead of ${modelName}`, typeof modelName === 'string');
 
+    let options = {};
+
+    options.adapterOptions = query.adapterOptions;
+    delete query.adapterOptions;
```
hjdivad: should we test that this is removed from `query`?

addon/-private/system/store.js
```diff
@@ -1386,10 +1391,15 @@ Store = Service.extend({
 
     let adapter = this.adapterFor(normalizedModelName);
 
+    let options = {};
+
+    options.adapterOptions = query.adapterOptions;
+    delete query.adapterOptions;
```
hjdivad: should we test that this is removed from `query`?

addon/-private/system/store.js
hjdivad: or is this already handled by an existing test?

stefanpenner: addressed

addon/-private/system/store.js
stefanpenner: addressed

addon/-private/system/store.js
hjdivad: 🤘

addon/-private/system/store.js
hjdivad: 🤘

stefanpenner:
we may want to consider moving `adapterOptions` to a third arg, not to be conflated with `query`

runspired:
> we may want to consider moving adapterOptions to a third arg, not to be conflated with query

👍

runspired:
@hjdivad and I discussed this OOB and decided that moving to a new optional last param for `adapterOptions` would be best. We gated use of that param for now by enforcing the `{ adapterOptions }` structure, which we can loosen or expand later as needed.

This has weirdness for `store.filter` where it becomes the 4th param, but the alternatives were equally messy and that API has been soft-deprecated for a very long time.

runspired:
Test failures on ember-canary are unrelated, it appears we have some breaking change to dig into unless there's a bug in Ember.

stefanpenner:
👯‍♂️

addon/-private/system/store/finders.js
```diff
@@ -143,14 +143,14 @@ export function _query(adapter, store, modelName, query, recordArray) {
 
   if (createRecordArray) {
     recordArray = recordArray || store.recordArrayManager.createAdapterPopulatedRecordArray(modelName, query);
-    promise = Promise.resolve().then(() => adapter.query(store, modelClass, query, recordArray));
+    promise = Promise.resolve().then(() => adapter.query(store, modelClass, query, recordArray, options));
   } else {
     promise = Promise.resolve().then(() => adapter.query(store, modelClass, query));
```
loganrosen: Was it intentional that `options` isn't passed along if `createRecordArray` is false? There isn't such a distinction in [the documentation](https://api.emberjs.com/ember-data/3.10/classes/DS.Store/methods/createRecord?anchor=query) about when an included `adapterOptions` hash will be passed to `adapter.query`.

stefanpenner: I think it was a mistake actually.

@runspired / @hjdivad / @igorT can you sanity check that?

stefanpenner: Speaking with @runspired he reminded me what this was intentionally, as without a record array, we cannot pass `options` as a fourth argument without breaking user-land arity checks. Also, creating the record array eagerly and providing it here, had some performance implications.

The plan at the time, was to eventually change this API to use an object, instead of positional arguments, to avoid this problem.

runspired: Additional note: if the arity check says that `adapter.query` uses less than 4 arguments (recordArray is 4th, options is 5th), we don't pass either along.

I'm not sure we were as worried about breaking user-land arity checks as we were about the performance costs of materializing the snapshotRecordArray when we didn't need to.

If your `query` implementation has an arity of 4/5 and is still not receiving options, then let's dive into that; however, simply adding it into the function signature in prep for usage should be enough to receive any options that were passed in.

loganrosen: @runspired and I dug into this more, and there's an issue with the arity check where it doesn't properly detect the argument length for adapter.query. This is captured in #6232.

PR #4862, opened 2017-03-14
Babel 6

rwjblue:


rwjblue:
Current failures are due to some node@4 incompatibilities in babel6-plugin-strip-heimdall that have been fixed upstream. I just need to confirm it is released and update here.

rwjblue:
Everything is working locally (I believe), but I'd love for folks to test (using node@6 or higher for now) and report back.

rwjblue:
@runspired - Can you release an updated babel6-plugin-strip-heimdall with node@4 support?

runspired:
@rwjblue yep!

rwjblue:
OK, this is ready for review.  It has feature parity with prior babel@5 configuration, and fixes a bug in our globals build (which makes our 2.13.0-beta.1 bower builds completely broken).

The goal with this work is to get ember-cli@2.13 to only include addons that are using babel@6, as such I'd like to get this cherry-picked into beta branch once reviewed.

rwjblue:
@bmac - I have kept "semantic" commits so far, but I don't mind squashing if that makes cherry-picking easier...

package.json
```diff
@@ -45,6 +45,17 @@
...
+    "babel-plugin-transform-es2015-block-scoping": "^6.23.0",
+    "babel-plugin-transform-es2015-classes": "^6.23.0",
+    "babel-plugin-transform-es2015-computed-properties": "^6.22.0",
+    "babel-plugin-transform-es2015-constants": "^6.1.4",
+    "babel-plugin-transform-es2015-destructuring": "^6.23.0",
+    "babel-plugin-transform-es2015-modules-amd": "^6.24.0",
+    "babel-plugin-transform-es2015-parameters": "^6.23.0",
+    "babel-plugin-transform-es2015-shorthand-properties": "^6.22.0",
+    "babel-plugin-transform-es2015-spread": "^6.22.0",
+    "babel-plugin-transform-es2015-template-literals": "^6.22.0",
```
runspired: is the assumption here that we build and release the globals version ourselves and thus these don't need to be dependencies?

runspired: @rwjblue this was the only thing I spotted I had any question on, otherwise +1 LGTM

stefanpenner:
Something seems fishy:

`yarn` for me isn't deduping the dependencies of all the babel plugins correctly, i'm ending up with nearly 80,000 files in node_modules. Most of them from duplicated `core-js` and `babel-runtimes` (all the same size).

My warm/colde initial builds are also quite slow likely do to hash-for-dep having to traverse all these dependencies.

stefanpenner:
When installing the babel6 branch in a project via

```js
yarn add "emberjs/data#babel-6"
```

All appears well. Not quite sure in-project I was having issues.

package.json
rwjblue: Yes, the only place we use these directly is in our own build of the addon itself (e.g. from `ember-cli-build.js` in this repo which is not used by any other consumers).

rwjblue:
> class syntax uses verbose (and hard to debug) array of pojo descriptors. Something #4864 addresses in babel 5. Maybe loose mode is missing?
>
>```
>    _createClass(RecordArrayManager, [{
>      key: 'recordDidChange',
>      value: function recordDidChange(internalModel) {
>        // TODO: change name
>        // TODO: track that it was also a change
>        this.internalModelDidChange(internalModel);
>      }
>    }, {
>```

Yes, we are not using loose.  I'll fix.

> do we expect? I thought @rwjblue added some loose mode for this guy to babel 6.
> ```js
>  Object.defineProperty(exports, "__esModule", {
>    value: true
>  });

Yes, we still expect `__esModule` (the thing I removed was `interopRequireDefault` / `interopRequireWildcard`), but we could likely use loose mode.

stefanpenner:
> do we expect? I thought @rwjblue added some loose mode for this guy to babel 6.
> 
>  Object.defineProperty(exports, "__esModule", {
>    value: true
>  });
> Yes, we still expect __esModule (the thing I removed was interopRequireDefault / interopRequireWildcard), but we could likely use loose mode.

👍

rwjblue:
OK, updated and ensured that the asset output is in `loose` mode (to match the changes from https://github.com/emberjs/data/pull/4864).

The appveyor failure seems related to an SSL issue, but one of the try results had already passed at this point so I think its fine...

---

@bmac I'd like for this to land into ember-data@2.13.0-beta.2, would you like me to squash the commits down to make that easier?

bmac:
Yes please @rwjblue.

stefanpenner:
just did a quick check, output looks good.

If you can double check build and yarn install times doesn't regress, we should be good!

rwjblue:
Before changes:

* `yarn` time (after a `rm -rf node_modules`): 27.44s
* `du -csh node_modules`: 369M
* `ember build` time (cold): 32.80s
* `ember build` time (warm): 6.89s
* `ember serve` rebuild time (after `touch addon/index.js`): 239ms
* `ember build -prod` time (cold): 36.23s
* Asset size of `dist/globals/ember-data.prod.js`: 691.6KB (137.69 KB gzipped)
* Asset size of `dist/globals/ember-data.min.js`: 164.95KB (37.54 KB gzipped)

After changes:

* `yarn` time (after a `rm -rf node_modules`): 25.03s
* `du -csh node_modules`: 339M
* `ember build` time (cold): 31.76s
* `ember build` time (warm): 9.29s
* `ember serve` rebuild time (after `touch addon/index.js`): 211ms
* `ember build -prod` time (cold): 36.97s
* Asset size of `dist/globals/ember-data.prod.js`: 562.04KB (111.86 KB gzipped)
* Asset size of `dist/globals/ember-data.min.js`: 164.76KB (37.83 KB gzipped)

---

It seems basically the same or slightly better in almost all of the things I measured.

bmac:
Thanks @rwjblue. 🎉

workmanw:
Just wanted to give a heads up that this change caused me to get some cryptic errors that took a little bit of time to track down.

```
The Broccoli Plugin: [broccoli-persistent-filter:Babel] failed with:
ReferenceError: [BABEL] ember-data/-private/adapters.js: Using removed Babel 5 option: /Users/wesley/Code/buttercup/src/buttercup-epcot/apps/buttercup/package.json.optional - Put the specific transforms you want in the `plugins` option
```

So I tracked this down to the fact that we had the following in our package.json:

```
"babel": {
  "optional": [
    "es7.decorators"
  ]
}
```

I guess maybe this is kind of legacy? The right was seems to be to include the following in my `ember-cli-build.js`.

```
var app = new EmberApp(defaults, {
  babel: {
    optional: ['es7.decorators']
  },
  // ...
});
```

I'm not sure how many people have this in their package.json, but if it's a sizable amount it might be worth mentioning in some upgrade notes / blog post. 

Just wanted to call it out.

rwjblue:
Interesting!! I thought we disabled the normal `.babelrc` discovery behavior, but it looks like we don't. I honestly had no clue that in an ember-cli app you could put `babel` key in your `package.json` and have it do anything...

workmanw:
>  I honestly had no clue that in an ember-cli app you could put babel key in your package.json and have it do anything...

That makes two of us. Another engineer on my team added decorators and when I saw it I immediately thought, why is that in the `package.json`? Especially because we had other babel options in the `ember-cli-build.js`. But debugging through the code, it looks like `babel-core` has an `OptionManager` which recursively walks the directory ancestor tree, parsing any package.json files it finds and merging them into the over all options. 👣 🔫 

Either way we had found that `package.json` trick somewhere. So presumably we're not the only ones using it. ¯\\\_(ツ)_/¯

runspired:
cc @stefanpenner I think you had to do some things to get the babel options from the parent for the rollup/babel re-ordering, any chance that changes the semantics of what is applied when?

PR #4863, opened 2017-03-15
Tidy up parse-reponse-headers

stefanpenner:
This aims to avoid the following overhead:

* forEach + closure  <-- this seems to be a-bit noisy on boot
* [field, …value] doesn’t really transpile nicely

More aggressive approach is possible, maybe someone else has time. This addresses the issues I noticed...


stefanpenner:
@krisselden / @runspired

stefanpenner:
quick benchmark harness i wrote for this one: https://github.com/stefanpenner/perf-scratch/commit/e3b2fe1dd5179a1577566ad511ff93c7674057f6

PR #4864, opened 2017-03-15
Loose es6 class

stefanpenner:
make ES6.Classes loose
* makes profiling easier (methods are not called value)
* appears to be a slight performance improvement (although not the motivator)

### DIFF
--- a/index.js
+++ b/index.js
@@ -124,6 +124,7 @@ module.exports = {
 
     this.options.babel = this.options.babel || {};
     add(this.options.babel, 'blacklist', ['es6.modules', 'useStrict']);
+    add(this.options.babel, 'loose', ['es6.classes']);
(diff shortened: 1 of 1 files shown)


stefanpenner:
should we consider bugfix beta?

rwjblue:
@stefanpenner - https://github.com/emberjs/data/pull/4862 plans to land into beta.  I've updated to use `loose` mode there also, and hope to land that PR today. Any objections to closing this in favor of that? I don't really mind either way, just pointing out what I'm working on...

stefanpenner:
@rwjblue lets use your PR (if its landed in the next few days). We should keep this around, in-case @bmac thinks its appropriate for a bugfix beta.

rwjblue:
@stefanpenner - Seems good.  https://github.com/emberjs/data/pull/4862 is ready for review, and also intends to target beta branch.

stefanpenner:
#4862 will still supercede, this will simplify my until then

PR #4865, opened 2017-03-16
[FEAT] InternalModelClass - separates "schema" from DS.Model

runspired:
@stefanpenner Ready for an early review.

I still need to port the tests that test this functionality in DS.Model to testing InternalModelClass directly.

### DIFF
--- a/addon/-private/system/model/internal-model-class.js
+++ b/addon/-private/system/model/internal-model-class.js
@@ -151,13 +151,26 @@ export default class InternalModelClass {
   }
 
   inverseFor(name, store) {
-    let inverseMap = get(this, 'inverseMap');
-    if (inverseMap[name]) {
+    let inverseMap = this.inverseMap;
+
+    if (inverseMap[name] !== undefined) {
       return inverseMap[name];
     } else {
-      let inverse = this._findInverseFor(name, store);
-      inverseMap[name] = inverse;
-      return inverse;
+      let relationship = this.relationshipsByName.get(name);
+      if (!relationship) {
+        inverseMap[name] = null;
+        return null;
+      }
+
+      let options = relationship.options;
+      if (options && options.inverse === null) {
+        // populate the cache with a miss entry so we can skip getting and going
+        // through `relationshipsByName`
+        inverseMap[name] = null;
+        return null;
+      }
+
+      return inverseMap[name] = this._findInverseFor(name, store);
     }
   }
 
@@ -575,7 +588,7 @@ export default class InternalModelClass {
    Unused internally
    */
   eachRelatedType(callback, binding) {
-    let relationshipTypes = get(this, 'relatedTypes');
+    let relationshipTypes = this.relatedTypes;
 
     for (let i = 0; i < relationshipTypes.length; i++) {
       let type = relationshipTypes[i];
@@ -587,7 +600,7 @@ export default class InternalModelClass {
    Used once by json-serializer
    */
   eachTransformedAttribute(callback, binding) {
-    get(this, 'transformedAttributes').forEach((type, name) => {
+    this.transformedAttributes.forEach((type, name) => {
       callback.call(binding, name, type);
     });
   }
--- a/addon/-private/system/model/internal-model-class.js
+++ b/addon/-private/system/model/internal-model-class.js
@@ -1,0 +1,615 @@
+import Ember from 'ember';
+import {
+  relationshipFromMeta
+} from 'ember-data/-private/system/relationship-meta';
+import { assert, warn } from "ember-data/-private/debug";
+
+const {
+  get,
+  Map,
+  A
+} = Ember;
+
+function findPossibleInverses(type, inverseType, name, relationshipsSoFar) {
+  let possibleRelationships = relationshipsSoFar || [];
+  let relationshipMap = get(inverseType, 'relationships');
+
+  if (!relationshipMap) { return possibleRelationships; }
+
+  let relationships = relationshipMap.get(type.modelName).filter(relationship => {
+    let optionsForRelationship = inverseType.metaForProperty(relationship.name).options;
+
+    if (!optionsForRelationship.inverse) {
+      return true;
+    }
+
+    return name === optionsForRelationship.inverse;
+  });
+
+  if (relationships) {
+    possibleRelationships.push.apply(possibleRelationships, relationships);
+  }
+
+  //Recurse to support polymorphism
+  if (type.superclass) {
+    findPossibleInverses(type.superclass, inverseType, name, possibleRelationships);
+  }
+
+  return possibleRelationships;
+}
+
+
+export default class InternalModelClass {
+  constructor(modelClass) {
+    let attributes;
+    let relationships;
+    let descriptors = get(modelClass, '_computedProperties');
+    let relationshipMap;
+    let properties = descriptors.length ? Object.create(null) : null;
+
+    for (let i = 0; i < descriptors.length; i++) {
+      let value = descriptors[i];
+      let key = value.name;
+
+      if (value && value.meta) {
+        let propertyMeta = value.meta;
+        propertyMeta.key = key;
+
+        if (propertyMeta.isAttribute === true) {
+          if (!attributes) {
+            attributes = Object.create(null);
+          }
+
+          propertyMeta.kind = 'attribute';
+          properties[key] = propertyMeta;
+          attributes[key] = propertyMeta;
+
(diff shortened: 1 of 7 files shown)


addon/-private/system/model/model.js
```diff
@@ -1,54 +1,29 @@
 import Ember from 'ember';
-import { assert, deprecate, warn } from "ember-data/-private/debug";
-import { PromiseObject } from "../promise-proxies";
-import Errors from "../model/errors";
+import { assert, deprecate } from 'ember-data/-private/debug';
+import { PromiseObject } from '../promise-proxies';
+import Errors from '../model/errors';
 import isEnabled from '../../features';
 import RootState from '../model/states';
+<<<<<<< 2edaf160a8a437cc989b2e6c980cc9c7d525c2c2
```
runspired: ugh, not sure how these crept it during my rebase (silly editor telling me file had been fixed)

addon/-private/system/model/internal-model-class.js
```diff
@@ -0,0 +1,648 @@
...
+} from 'ember-data/-private/system/relationship-meta';
+import { assert, warn } from "ember-data/-private/debug";
+
+let COUNT_DATA = Object.create(null);
+
+function count(str) {
+  COUNT_DATA[str] = COUNT_DATA[str] ? COUNT_DATA[str] + 1 : 1;
+}
+
+window.getInternalModelClassCountData = function() {
```
josemarluedke: I was wondering if this `window` will this play FastBoot?

runspired: This should be removed. I've been using it for tracking when/where internally we use some of these hooks so that we can optimize them better and potentially deprecate a few of them.

BryanCrotaz:
interesting
we were doing some work with polymorphic types to be able to `find` model or descendants. This requires knowing the `extends` tree for the model hierarchy, which we couldn't work out how to do - does this PR make that possible?

tchak:
LGTM

stefanpenner:
CI appears un-happy.

addon/-private/system/model/model.js
```diff
@@ -1631,11 +1498,13 @@ Model.reopenClass({
    @param {any} binding the value to which the callback's `this` should be bound
    */
   eachRelationship(callback, binding) {
-    get(this, 'relationshipsByName').forEach((relationship, name) => {
-      callback.call(binding, name, relationship);
-    });
+    get(this, '__internalModelClass').eachRelationship(callback, binding);
   },
 
+  __internalModelClass: computed(function() {
```
stefanpenner: I don't think we should introduce another property on DS.Model. This is accessible via `this._internalModel.modelClass`, I think we should use that, rather then the CP indirection.

* it will be quicker then a CP
* less things on the DS.Model the better

addon/-private/system/model/model.js
```diff
@@ -1824,9 +1648,7 @@ Model.reopenClass({
    @static
    */
   eachAttribute(callback, binding) {
```
stefanpenner: I suspect we likely just want a proxyMethod for all these.

```js
eachAttribute: proxyToModelClass('eachAttribute')
```

addon/-private/system/model/internal-model-class.js
```diff
@@ -0,0 +1,615 @@
...
+
+  return possibleRelationships;
+}
+
+
+export default class InternalModelClass {
+  constructor(modelClass) {
+    let attributes;
+    let relationships;
+    let descriptors = get(modelClass, '_computedProperties');
```
stefanpenner: Is there a reason why descriptor parsing is done eagerly in the constructor?

addon/-private/system/model/internal-model-class.js
```diff
@@ -0,0 +1,615 @@
...
+    return this._properties[name];
+  }
+
+  eachRelationship(cb, context) {
+    let relationships = this._relationships;
+
+    if (relationships) {
+      let exec = cb.bind(context);
+
+      for (let name in relationships) {
```
stefanpenner: we seem to only iterate relationships for there membership, should `this._relationships` just be an array then?

it will then:

* iterate faster
* be easy to skip/avoid/dealloc if empty

addon/-private/system/model/internal-model-class.js
```diff
@@ -0,0 +1,615 @@
...
+      for (let name in attributes) {
+        let attribute = attributes[name];
+
+        exec(name, attribute);
+      }
+    }
+  }
+
+  inverseFor(name, store) {
+    let inverseMap = get(this, 'inverseMap');
```
stefanpenner: `inverseMap` is a `getter` do we need `get(..`?

addon/-private/system/model/internal-model-class.js
```diff
@@ -0,0 +1,615 @@
...
+    }
+
+    return this.__transformedAttributesMap;
+  }
+
+  /**
+   Unused internally
+   */
+  eachRelatedType(callback, binding) {
+    let relationshipTypes = get(this, 'relatedTypes');
```
stefanpenner: relatedTypes is a getter now, we can skip the `get` helper

addon/-private/system/model/internal-model-class.js
```diff
@@ -0,0 +1,615 @@
...
+      let type = relationshipTypes[i];
+      callback.call(binding, type);
+    }
+  }
+
+  /**
+   Used once by json-serializer
+   */
+  eachTransformedAttribute(callback, binding) {
+    get(this, 'transformedAttributes').forEach((type, name) => {
```
stefanpenner: `get` helper is not required

addon/-private/system/model/model.js
runspired: As discussed offline, this is a static prop, and for reasons (including ember-inspector) we can't rely on setting this via `store.modelFor` lookup.  Personally I prefer this would be hidden (via WeakMap or such) from end users, but we aren't left with very good options at the moment for that.

addon/-private/system/model/internal-model-class.js
runspired: I experimented with doing it lazily, but in nearly all cases currently we access additional information almost immediately following lookup.

addon/-private/system/model/internal-model-class.js
runspired: I will experiment with that, there may have been reasons associated with needing to support the legacy Map api.

addon/-private/system/model/internal-model-class.js
runspired: Probably not.

addon/-private/system/model/internal-model-class.js
runspired: 👍

addon/-private/system/model/internal-model-class.js
runspired: 👍

addon/-private/system/model/model.js
stefanpenner: 👍

addon/-private/system/model/internal-model-class.js
stefanpenner: 👍

addon/-private/system/model/model.js
```diff
@@ -1630,11 +1497,11 @@ Model.reopenClass({
...
-  eachRelationship(callback, binding) {
-    get(this, 'relationshipsByName').forEach((relationship, name) => {
-      callback.call(binding, name, relationship);
-    });
-  },
+  eachRelationship: proxyMethodToInternalClass('eachRelationship'),
+
+  __internalModelClass: computed(function() {
+    return new InternalModelClass(this);
+  }),
```
stefanpenner: `readOnly`

runspired: 👍

addon/-private/system/model/internal-model-class.js
runspired: @stefanpenner now that the other PRs are merging I'll revisit this one for some cleanup soon :)

addon/-private/system/model/model.js
```diff
@@ -1168,6 +1135,18 @@ const Model = Ember.Object.extend(Ember.Evented, {
   }
 });
 
+function proxyPropToInternalClass() {
+  return computed(function(key) {
```
stefanpenner: `readOnly`

runspired:
Closing in favor of the ModelData RFC #5344

Issue #4804, opened 2017-02-15
pushPayload does not change hasMany relationship when going to empty state


danielspaniel:
I have a model which hasMany things, then I pushPayload with data that says .. I have no things anymore.  Ember data ignores that and keeps the hasMany things. 

The key activities are that:
 -  I have a model with no things
 -  then I set some things  model.set('things',  things)
 -  then I pushPayload to remove them. 

Here is a twiddle to show the problem:

https://ember-twiddle.com/26204b7fa7fc185dde464418cd33df95?openFiles=routes.application.js%2C

hjdivad:
@danielspaniel https://ember-twiddle.com/53d305379774c4756d576107c98ae6d4?openFiles=routes.application.js%2C

In the above gist you can comment / uncomment line 48 (ie the `this.store.pushPayload` line) to see things either cleared or not cleared.

There are two changes as compared to your original twiddle.

1. the initial relationships data is populated, the relationships are not merely sieloaded in the included payload
2. the manyarray itself is used and not overwritten by the live array from `peekAll` (ie the `person.set('things', things)` line is commented).


Please let me know if I have misunderstood the issue your twiddle aimed to highlight, but I believe the core error in the above was in the payload and not an ember data bug.

runspired:
The issue with the first twiddle is that the relationship is being added as "dirty state", it won't be cleared because no save attempt has been made (so obviously a server update would not know about it).

danielspaniel:
The reason I flag this as a bug is because even in a dirty state if you pushPayload with values for the hasMany then the relationship will update itself to add those new rows. 
By your logic in a dirty state pushing data to a model would always have no affect.

Exelord:
Any news here?

Dirk-27:
I ran into the same issue with ember-data Version 3.8: empty relationship arrays in the payload didn't clear the relationship. What is the state of this issue?

wuarmin:
I have the same issue

runspired:
@Dirk-27 the state of this issue is there is nothing more we can do for `push` api's. It is equally incorrect to discard local mutations as it is to not discard them: it just boils down to what your app wants.

While not explicitly enabled by the new cache design in emberjs/rfcs#854 once `@ember-data/graph` is made public rollback will be a public API. Custom cache implementations can also choose today to do one behavior vs the other.

runspired:
[...]

PR #4872, opened 2017-03-16
[BUGFIX] Notify hasMany relation if changed

### LINKED ISSUE #4816: PushPayload do not update belongsTo relations
Hey!
Here is a twiddle of the issue with a description: https://ember-twiddle.com/17151235360da0e8b7964fd324fdc299?openFiles=templates.components.push-payload.hbs%2Ctemplates.components.push-payload.hbs

Exelord:
It will fix #4816 #4804 
I just need to fix one test. Maybe you can help me :)

Edit: All tests fixed :)

### DIFF
--- a/addon/-private/system/relationships/state/has-many.js
+++ b/addon/-private/system/relationships/state/has-many.js
@@ -111,6 +111,8 @@ export default class ManyRelationship extends Relationship {
   flushCanonical() {
     if (this._manyArray) {
       this._manyArray.flushCanonical();
+    } else {
+      this.notifyHasManyChanged();
     }
     super.flushCanonical();
   }
--- a/tests/integration/records/relationship-changes-test.js
+++ b/tests/integration/records/relationship-changes-test.js
@@ -10,7 +10,8 @@ const { attr, belongsTo, hasMany, Model } = DS;
 let env, store;
 
 const Author = Model.extend({
-  name: attr('string')
+  name: attr('string'),
+  posts: hasMany()
 });
 
 const Post = Model.extend({
@@ -948,4 +949,113 @@ test('Calling push with same belongsTo relationship does not trigger observer', 
   });
 
   assert.equal(observerCount, 0, 'author observer should not be triggered');
+});
+
+test('Calling push with non empty hasMany relationship does trigger observer', function(assert) {
+  assert.expect(1);
+
+  let observerCount = 0;
+
+  run(function() {
+    let author = env.store.push({
+      data: {
+        id: '2',
+        type: 'author',
+        relationships: {
+          posts: {
+            data: [{ type: 'post', id: '2' }]
+          }
+        }
+      }
+    });
+
+    author.addObserver('posts', function() {
+      observerCount++;
+    });
+
+    env.store.push({
+      data: {
+        type: 'post',
+        id: '1',
+        relationships: {
+          author: {
+            data: { type: 'author', id: '2' }
+          }
+        }
+      }
+    });
+  });
+
+  assert.equal(observerCount, 1, 'posts observer should be triggered');
+});
+
+test('Calling push with same hasMany relationship does not trigger observer', function(assert) {
+  assert.expect(1);
+
+  let observerCount = 0;
+
+  run(function() {
+    let author = env.store.push({
+      data: {
+        id: '2',
+        type: 'author',
+        relationships: {
+          posts: {
+            data: [{ type: 'post', id: '2' }]
+          }
+        }
+      }
+    });
+
+    author.addObserver('posts', function() {
+      observerCount++;
+    });
+
+    env.store.push({
+      data: {
+        type: 'post',
+        id: '2',
+        relationships: {
+          author: {
+            data: { type: 'author', id: '2' }
+          }
+        }
+      }
+    });
+  });
+
+  assert.equal(observerCount, 0, 'posts observer should not be triggered');
+});
+
+test('Calling push with empty hasMany relationship does trigger observer', function(assert) {
+  assert.expect(1);
+
+  let observerCount = 0;
+
+  run(function() {
+    let author = env.store.push({
+      data: {
+        id: '2',
+        type: 'author'
+      }
+    });
+
+    author.addObserver('posts', function() {
+      observerCount++;
+    });
+
+    env.store.push({
+      data: {
+        type: 'post',
+        id: '1',
+        relationships: {
+          author: {
+            data: { type: 'author', id: '2' }
+          }
+        }
+      }
+    });
+  });
+
+  assert.equal(observerCount, 1, 'posts observer should be triggered');
 });
--- a/tests/integration/records/relationship-changes-test.js
+++ b/tests/integration/records/relationship-changes-test.js
@@ -951,7 +951,7 @@ test('Calling push with same belongsTo relationship does not trigger observer', 
   assert.equal(observerCount, 0, 'author observer should not be triggered');
 });
 
-test('Calling push with non empty hasMany relationship does trigger observer', function(assert) {
+test('Calling push with hasMany relationship does trigger relation observer with already loaded objects', function(assert) {
   assert.expect(1);
 
   let observerCount = 0;
@@ -969,7 +969,7 @@ test('Calling push with non empty hasMany relationship does trigger observer', f
       }
     });
 
-    author.addObserver('posts', function() {
+    author.addObserver('posts.[]', function() {
       observerCount++;
     });
 
@@ -989,7 +989,7 @@ test('Calling push with non empty hasMany relationship does trigger observer', f
   assert.equal(observerCount, 1, 'posts observer should be triggered');
 });
 
-test('Calling push with same hasMany relationship does not trigger observer', function(assert) {
+test('Calling push with hasMany relationship does trigger relation observer with the same loaded objects', function(assert) {
   assert.expect(1);
 
   let observerCount = 0;
@@ -1007,7 +1007,7 @@ test('Calling push with same hasMany relationship does not trigger observer', fu
       }
     });
 
-    author.addObserver('posts', function() {
+    author.addObserver('posts.[]', function() {
       observerCount++;
     });
 
@@ -1027,7 +1027,7 @@ test('Calling push with same hasMany relationship does not trigger observer', fu
   assert.equal(observerCount, 0, 'posts observer should not be triggered');
 });
 
-test('Calling push with empty hasMany relationship does trigger observer', function(assert) {
+test('Calling push with hasMany relationship does trigger relation observer with not loaded objects', function(assert) {
   assert.expect(1);
 
   let observerCount = 0;
@@ -1040,7 +1040,7 @@ test('Calling push with empty hasMany relationship does trigger observer', funct
       }
     });
 
-    author.addObserver('posts', function() {
+    author.addObserver('posts.[]', function() {
       observerCount++;
     });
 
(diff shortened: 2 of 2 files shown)


tests/integration/records/relationship-changes-test.js
```diff
@@ -949,3 +950,112 @@ test('Calling push with same belongsTo relationship does not trigger observer',
...
+        type: 'author',
+        relationships: {
+          posts: {
+            data: [{ type: 'post', id: '2' }]
+          }
+        }
+      }
+    });
+
+    author.addObserver('posts', function() {
```
BryanCrotaz: should be `addObserver('posts.[]')`

the array hasn't been replaced (`posts`), it's been modified (`posts.[]`)

addon/-private/system/relationships/state/has-many.js
```diff
@@ -112,6 +112,8 @@ export default class ManyRelationship extends Relationship {
     if (this._manyArray) {
       this._manyArray.flushCanonical();
     }
+
+    this.notifyHasManyChanged();
```
BryanCrotaz: No - the array hasn't necessarily changed - the above code is there to identify whether anything has changed and to notify the minimum possible change.

This addition should not be made. It nullifies the optimisation.

tests/integration/records/relationship-changes-test.js
```diff
@@ -949,3 +950,112 @@ test('Calling push with same belongsTo relationship does not trigger observer',
...
```
BryanCrotaz: should be `addObserver('posts.[]', ...`

tests/integration/records/relationship-changes-test.js
```diff
@@ -949,3 +950,112 @@ test('Calling push with same belongsTo relationship does not trigger observer',
...
+          }
+        }
+      }
+    });
+  });
+
+  assert.equal(observerCount, 1, 'posts observer should be triggered');
+});
+
+test('Calling push with same hasMany relationship does not trigger observer', function(assert) {
```
BryanCrotaz: test name should say that you're pushing the inverse relationship

tests/integration/records/relationship-changes-test.js
```diff
@@ -949,3 +950,112 @@ test('Calling push with same belongsTo relationship does not trigger observer',
 
   assert.equal(observerCount, 0, 'author observer should not be triggered');
 });
+
+test('Calling push with non empty hasMany relationship does trigger observer', function(assert) {
```
BryanCrotaz: test name should say that you're pushing the inverse relationship

tests/integration/records/relationship-changes-test.js
```diff
@@ -949,3 +950,112 @@ test('Calling push with same belongsTo relationship does not trigger observer',
...
+
+  run(function() {
+    let author = env.store.push({
+      data: {
+        id: '2',
+        type: 'author'
+      }
+    });
+
+    author.addObserver('posts', function() {
```
BryanCrotaz: again, observe `posts.[]`

tests/integration/records/relationship-changes-test.js
```diff
@@ -949,3 +950,112 @@ test('Calling push with same belongsTo relationship does not trigger observer',
...
+          }
+        }
+      }
+    });
+  });
+
+  assert.equal(observerCount, 0, 'posts observer should not be triggered');
+});
+
+test('Calling push with empty hasMany relationship does trigger observer', function(assert) {
```
BryanCrotaz: name is misleading - that's not what you're doing at all

Exelord:
@BryanCrotaz what about now? :)

addon/-private/system/relationships/state/has-many.js
```diff
@@ -111,6 +111,8 @@ export default class ManyRelationship extends Relationship {
   flushCanonical() {
     if (this._manyArray) {
       this._manyArray.flushCanonical();
+    } else {
+      this.notifyHasManyChanged();
```
BryanCrotaz: it hasn't necessarily changed here

tests/integration/records/relationship-changes-test.js
```diff
@@ -949,3 +950,112 @@ test('Calling push with same belongsTo relationship does not trigger observer',
 
   assert.equal(observerCount, 0, 'author observer should not be triggered');
 });
+
+test('Calling push with hasMany relationship does trigger relation observer with already loaded objects', function(assert) {
```
BryanCrotaz: you're not pushing a relationship - you're pushing an object that changes the inverse hasMany

tests/integration/records/relationship-changes-test.js
```diff
@@ -949,3 +950,112 @@ test('Calling push with same belongsTo relationship does not trigger observer',
...
+          }
+        }
+      }
+    });
+  });
+
+  assert.equal(observerCount, 1, 'posts observer should be triggered');
+});
+
+test('Calling push with hasMany relationship does trigger relation observer with the same loaded objects', function(assert) {
```
BryanCrotaz: ^

tests/integration/records/relationship-changes-test.js
```diff
@@ -949,3 +950,112 @@ test('Calling push with same belongsTo relationship does not trigger observer',
...
+          }
+        }
+      }
+    });
+  });
+
+  assert.equal(observerCount, 0, 'posts observer should not be triggered');
+});
+
+test('Calling push with hasMany relationship does trigger relation observer with not loaded objects', function(assert) {
```
BryanCrotaz: ^

tests/integration/records/relationship-changes-test.js
Exelord: In the same file, there are 3 other tests related to belongsTo, testing the same behavior and they are named:
```js
test('Calling push with same belongsTo relationship does not trigger observer', function(assert) {
test('Calling push with updated belongsTo relationship trigger observer', function(assert) {
test('Calling push with relationship triggers willChange and didChange with detail when replacing different length in middle', function(assert) {
```

How should be named my tests according to your proposal?

BryanCrotaz: Pushing object that modifies hasMany via inverse relationship fires observer

BryanCrotaz: Feel free to modify the other test names to match!

addon/-private/system/relationships/state/has-many.js
Exelord: In case when hasMany relationship was not loaded previously it will notify relation about change.
Thanks to this, these two tests are working correctly. And as I know the same behavior is working with belongsTo (that tests are calling)

screen shot 2017-03-17 at 08 40 46

BryanCrotaz:
If it did would anyone be listening? If no one can be listening then there's no need to notify

Exelord:
Becouse it can be loaded by pushPayload and then the relation will change even if the relation wasnt loaded . Check the tests. It broke a lot of code in our project just because in one case the hasMany wasn't observerd. And there is a test for that which should pass the same as belongsTo do.

stefanpenner:
I can appreciate that this bugfix may have caused some grief (especially tests which do not render, and thus to not access these computed properties). But without the fix, the grief was very serious and pronounced. 

> Check the tests.

This test do not appear correct, if no-one has ever accessed `post` relationship it does not exist, and should not be observable. Prior behavior that violated this we truelly unfortunate. Both in program correctness, and in performance.

Let me provide a quick example:

```js
let User = Ember.Object.extend({
  veryLargeFilteredList: computed('someInput', function() {
    // do expensive work, and return the result (but only if someone accesses it)
   }),
  
   // it would be very unfortunate, if the above CP was always forced to compute, merely by observing it. Even though it is unclear anyone actually needs it....
   // to balance the needs, we can simply access the CP eager in init, or via on('init')
   veryLargeFilteredListDidChange: Ember.observer('veryLargeFilteredList.[]', function() {
   // when the list changes, inform me.
 })
});
```

If someone really needs to observe something, (and mimic the buggy behavior, but explicitly):

```js
let User = Ember.Object.extend({
  init() {
     this._super(...arguments);
     // explicitly tell the system, that this CP should always be computed 
     // * opt-in, is much easier then opt-out
     // * out-in, ensures that by default unneeded work does not happen
     this.get('veryLargeFilteredList');
  },
  veryLargeFilteredList: computed('someInput', function() {
    // do expensive work, and return the result (but only if someone accesses it)
   }),
  
   veryLargeFilteredListDidChange: Ember.observer('veryLargeFilteredList.[]', function() {
   // when the list changes, inform me.
 })
});
```

---

More on whats going on:

An observer observing a something should not have side-affects, if observing a property forced that property to be created that would in-fact be a side-affect. One should safely be able to observe information that is not yet present (not forcing it to become present), but then react only once it is used and is present.

```js
init() {
  this._super(...arguments);


  // computed properties are accessor, and accessor are by default lazy.
  // if we want them to be eager, we can invoke them in the constructor.
  // as they are now computed, they are observable.
  this.get('fullName'); 
}
```

```js
fullNameDidChange: observer('fullName', function() {
  let fullName = this.get('fullName'); // accesses fullName ensure it has been computed.
  // something
}).on('init') // informs the system that the observer should immediately fire after the constructor runs
```

If you want to observe relationships such as this, you must first access said relationship

---

*sidenote: In general, I wouldn't recommend relying heavily on observers. They are a low level concept (important, but unwieldy). I speak more about this problem (and how to be successful with cp/observers) here -> https://www.youtube.com/watch?v=vvZEddrClAQ.*

Exelord:
I see now. I fully agree with the observers purpose that you present and now understand my bad. They shouldn't change and propagate changes to computed properties.

However, I'm still confident with that test case which in my opinion should pass: `Pushing object that modifies already loaded hasMany via inverse relationship fires observer`. Why?

1. With the same behavior as in `belongsTo` we expect that changing relation to other value will notify an observer, ie. Changing Post's author will notify 'post.author' about change; also in hasMany context (inverse relation)
2. OK. We shouldn't observe a relation that we will not use. But in a case when you load data to store not through `findRecord`, `findAll` or fetching the relation explicit, we should be able still to keep a behavior compatibility. Eg. loading a relations data by ajax request and pushing to the store, should notify observer directly because we have no way to setup computed on it. (We can, but if we will do that we will fire a fetch request which we want to avoid [cuz we are loading our data alternativly])

I found another solution to solve that problem on user side by adding
```js
    author.hasMany('posts').hasManyRelationship.manyArray;
```
before the test's observer which will force the many array to create.

It also works if we make it in hasMany state:
```js
  flushCanonical() {
    if (this._manyArray) {
      this._manyArray.flushCanonical();
    } else {
      this.manyArray;
    }
    super.flushCanonical();
  }
```

But as I expect you will not be happy with that cuz it will load (the lazy part) in much more cases which we want to avoid. :/
So, maybe you have another idea how to tackle this problem?

Exelord:
Is the topic die?

stefanpenner:
> Is the topic die?

No but the PR does not appear to be correct right, mind updating?

Your last comment seems reasonable, but Unfortunately I have not yet had any cycles to investigate (but do intend do).

runspired:
I think this is may be magically fixed by the relationship cleanup PR, we had an issue where ManyArray was created in initial state always, even when it shouldn't be. By fixing this, I think if I understand correctly you get notified when it is loaded later were you to be observing it.

PoslinskiNet:
Is there anything we can do to move on with this fix?

runspired:
Check my analysis on #5245, I believe these are identical and that furthermore the underlying issue is folks trying to listen on unconsumed computed property changes, which is a code smell.

PR #4882, opened 2017-03-20
[Feat] simplified relationship state

runspired:
# Simplified Relationship State

**Original Commits** https://github.com/runspired/data/tree/original/simple-relationship-state this PR has since been flattened for rebase to continue tracking master.

## Description

This is a backwards compatible revisiting of Ember-Data's relationship state objects with the intent of simplifying data flows, improving our ability to reason about the relationship layer, and further optimize relationship management in the future.

## How to Review

Each commit in this PR is a discrete step along the path of this refactor, and (with I think one exception) left ember-data in a working state with tests passing.  To follow the logic, follow the PRs.

## Summary

- Creates a new `ImplicitRelationship` class which absorbed most of the previous `OrderedSet` behavior of `Relationship`
- Converts the `Relationship` class to largely acting as a shared interface that `BelongsToRelationship` `HasManyRelationship` and `ImplicitRelationship` must implement.
- Simplifies the state management of `BelongsToRelationship` and `HasManyRelationship`, each now has a `currentState` and `canonicalState` property. For `belongs-to` these are `InternalModel`s while for `has-many` these are arrays of `InternalModel`s.
- Moves management of `currentState` for the `HasManyRelationship` out of `ManyArray` and into the relationship class. This brings `ManyArray` much closer to being a simple proxy Array.
- Cleans up some test code encountered during this process
- Cleans up some use of `record` => `internalModel`
- Uses `UniqueArray` to make a few performance improvements to places where we were blindly concat-ing arrays leading to duplicate work before.
- More intelligently iterates over relationships in `Store._setupRelationships`

## Motivations

- Make relationship state simpler to manage and easier reason about.
- Enable better optimizations within the relationship layer

## Performance Characteristics

With this refactor we

- Allocate fewer objects (especially far fewer `OrderedSet`s which are only used now for `ImplicitRelationship`s).
- Produce many fewer closures when iterating
- No longer move back and forth between `OrderedSet` and `Array`
- use strict assignment and `===` for managing `BelongsToRelationship` state
- no longer duplicate `diffing` work for each state change within each relationship class layer as well as within `ManyArray`
- broadcast fewer duplicate or unnecessary state changes

These changes were compared against `#master` to validate performance changes we expect to see.  The test used the `localhost:4200/api/complexes?included=foo%2Cbaz&limit=34` endpoint which returns 34 primary records and 204 included records of 3 total model types.

**Before**
screen shot 2017-03-20 at 5 16 06 pm

**After**
screen shot 2017-03-20 at 5 14 26 pm

**Results of This PR against LinkedIn Feed**
plot


addon/-private/system/unique-array.js
```diff
@@ -0,0 +1,21 @@
+export default class UniqueArray {
+  constructor(key) {
+    this.key = key;
+    this.seen = Object.create(null);
```
krisselden: consider not initializing seen until array is > say 20 items.

addon/-private/system/model/internal-model.js
```diff
@@ -426,17 +427,30 @@ export default class InternalModel {
...
-
-        array = array.concat(localRelationships, serverRelationships);
+        let additions;
+
+        switch (relationship.kind) {
+          case 'belongs-to':
+            array.push(relationship.currentState, relationship.canonicalState);
+            return;
+          case 'has-many':
+            additions = [].concat(relationship.currentState, relationship.canonicalState);
```
krisselden: do 2 pushes instead of concat

addon/-private/system/relationships/state/implicit.js
```diff
@@ -0,0 +1,200 @@
...
+    heimdall.increment(newRelationship);
+
+    this.members = new OrderedSet();
+    this.canonicalMembers = new OrderedSet();
+  }
+
+  destroy() {
+    if (!this.inverseKey) { return; }
+
+    let allMembers =
```
krisselden: can we not copy to an array just to iterate over?

addon/-private/system/relationships/state/implicit.js
```diff
@@ -0,0 +1,200 @@
...
+  }
+
+  removeInternalModels(internalModels) {
+    heimdall.increment(removeInternalModels);
+    internalModels.forEach((internalModel) => this.removeRecord(internalModel));
+  }
+
+  addInternalModels(internalModels, idx) {
+    heimdall.increment(addInternalModels);
+    internalModels.forEach(internalModel => {
```
krisselden: if this is a hot path consider avoiding forEach

addon/-private/system/snapshot.js
```diff
@@ -279,11 +279,11 @@ export default class Snapshot {
     }
 
     hasData = get(relationship, 'hasData');
-    members = get(relationship, 'members');
+    currentState = relationship.currentState;
 
     if (hasData) {
       results = [];
-      members.forEach((member) => {
+      currentState.forEach((member) => {
```
krisselden: if hot path, avoid forEach.

addon/-private/system/model/internal-model.js
runspired: I was debating this, any particular reason why?

addon/-private/system/relationships/state/implicit.js
runspired: Seeing as this doesn't unique the sets either it's doubly inefficient.

runspired: (also fwiw this is the legacy path that everything used to do, now only implicit does it, I haven't spent much time optimizing implicit's code yet)

addon/-private/system/unique-array.js
runspired: 👍

testem.js
```diff
@@ -8,7 +8,7 @@ module.exports = {
     "PhantomJS"
   ],
   "launch_in_dev": [
-    "PhantomJS",
+  //  "PhantomJS",
```
runspired: this crept in, should remove it.

addon/-private/system/store.js
```diff
@@ -2396,28 +2398,50 @@ Store = Service.extend({
...
+          if (data.relationships[key]) {
+            let relationship = internalModel._relationships.get(key);
+            relationship.push(data.relationships[key]);
+          }
+        }
+      }
     }
 
-    pushed.length = 0;
+    this._pushedInternalModels = Object.create(null);
```
pangratz: Do we need to reset `_hasPushedInternalModels` as well? As far as I can tell, otherwise the `_setupRelationships` is only invoked once?

runspired: Thanks for the catch! We do need to. I have a fix for this in a different branch where I was rebasing and will bring it back over to here.

addon/-private/system/relationships/state/relationship.js
```diff
@@ -18,4 +18,9 @@ export default class Relationship {
     this.hasLoaded = false;
     this.meta = null;
   }
+
+  destroy() {
+    this.currentState = null;
```
igorT: Check if this is still there

addon/-private/system/relationships/state/belongs-to.js
```diff
@@ -6,10 +6,14 @@ import {
...
 } from '../../promise-proxies';
-import Relationship from "./implicit";
-const ImplicitRelationship = Relationship;
+import ImplicitRelationship from './implicit';
+import Relationship from './relationship';
 
 export default class BelongsToRelationship extends Relationship {
+  constructor(store, internalModel, inverseKey, relationshipMeta) {
+    super(store, internalModel, inverseKey, relationshipMeta);
+    this.kind = 'belongs-to';
```
igorT: I think we camelCase this

addon/-private/system/relationships/state/belongs-to.js
```diff
@@ -69,60 +70,45 @@ export default class BelongsToRelationship extends Relationship {
...
-      this.members.addWithIndex(newRecord, idx);
-      this.notifyRecordRelationshipAdded(newRecord, idx);
-      if (this.inverseKey) {
-        newRecord._relationships.get(this.inverseKey).addRecord(this.record);
-      } else {
-        if (!newRecord._implicitRelationships[this.inverseKeyForImplicit]) {
-          newRecord._implicitRelationships[this.inverseKeyForImplicit] = new ImplicitRelationship(this.store, newRecord, this.key,  { options: {} });
-        }
-        newRecord._implicitRelationships[this.inverseKeyForImplicit].addRecord(this.record);
+    // TODO implicit-legacy @runspired is this needed?
```
igorT: Leftover comment

addon/-private/system/relationships/state/belongs-to.js
```diff
@@ -172,12 +187,13 @@ export default class BelongsToRelationship extends Relationship {
...
 
   fetchLink() {
-    return this.store.findBelongsTo(this.internalModel, this.link, this.relationshipMeta).then((record) => {
-      if (record) {
-        this.addRecord(record);
-      }
-      return record;
-    });
+    return this.store.findBelongsTo(this.internalModel, this.link, this.relationshipMeta)
+      .then((internalModel) => {
```
igorT: Can you keep existing indentation please

addon/-private/system/relationships/state/relationship.js
```diff
@@ -1,7 +1,13 @@
+import { assert, warn } from 'ember-data/-private/debug';
+import _normalizeLink from '../../normalize-link';
+
+let REL_ID = 0;
+
 export default class Relationship {
   constructor(store, internalModel, inverseKey, relationshipMeta) {
     let asyncOptionValue = relationshipMeta.options.async;
 
+    this.rel_id = REL_ID++;
```
igorT: Lets leave a comment as to why we are doing this

addon/-private/system/model/internal-model.js
```diff
@@ -31,6 +31,28 @@ const {
 
 const assign = Ember.assign || Ember.merge;
 
+class UniqueArray {
```
igorT: Can you document this please, and maybe take it out in a separate file?

addon/-private/system/model/internal-model.js
```diff
@@ -31,6 +31,28 @@ const {
...
+  constructor(key) {
+    this.key = key;
+    this.seen = Object.create(null);
+    this.items = [];
+  }
+
+  push(...additions) {
```
igorT: Can we make this more like a util to make it clear it's not a stateful array

runspired:
hrm, that wasn't failing locally. Might have missed something in the commit.

addon/-private/system/model/internal-model.js
```diff
@@ -429,17 +429,30 @@ export default class InternalModel {
     to or has many.
   */
   _directlyRelatedInternalModels() {
-    let array = [];
-    this.type.eachRelationship((key, relationship) => {
+    let uniqueSet = new OrderedSet();
```
stefanpenner: all sets are unique, lets just name this variable `set`

addon/-private/system/ordered-set.js
```diff
@@ -16,6 +16,18 @@ OrderedSet.prototype = Object.create(EmberOrderedSet.prototype);
...
+      this.add(value);
+    }
+  }
+
+  return this.size;
+};
+
 OrderedSet.prototype.addWithIndex = function(obj, idx) {
   let guid = guidFor(obj);
   let presenceSet = this.presenceSet;
```
stefanpenner: This is unfortunately private API that will change. Ember-data shouldn't be reaching into private state, rather it should use `set.has`

stefanpenner: I think the only thing is to add this to ember's, and have ember's lazily create `presenceSet` for old versions of ED.

runspired: This part is pre-existing, but I agree. I feel we really just need to build an OrderedSet with good perf and tailored to ember-data, given that it exists inside of Ember mostly for ember-data to begin with.

addon/-private/system/model/internal-model.js
hjdivad: best to change it.  `uniqueSet` is confusing (like `hashMapWithKeysAndValues`)

stefanpenner:
rebasing

addon/-private/system/store.js
```diff
@@ -2417,20 +2419,63 @@ Store = Service.extend({
...
     // instances than types we want to minimize the cost of looking up the
     // inverse map and the overhead of Ember.get adds up.
     let modelNameToInverseMap = Object.create(null);
+    let modelNameToRelationshipsMap = Object.create(null);
 
-    for (let i = 0, l = pushed.length; i < l; i += 2) {
-      // This will convert relationships specified as IDs into DS.Model instances
-      // (possibly unloaded) and also create the data structures used to track
-      // relationships.
+    for (let i = 0; i < pushed.length; i += 2) {
```
stefanpenner: this produces quite sketchy ES5 output, lots of nested functions to handle the `let` TDZ

```js
  _setupRelationships: function _setupRelationships() {
    var pushed = this._pushedInternalModels;
    var store = this;

    // Cache the inverse maps for each modelClass that we visit during this
    // payload push.  In the common case where we are pushing many more
    // instances than types we want to minimize the cost of looking up the
    // inverse map and the overhead of Ember.get adds up.
    var modelNameToInverseMap = Object.create(null);
    var modelNameToRelationshipsMap = Object.create(null);

    var _loop2 = function _loop2(i) {
      var internalModel = pushed[i];
      var data = pushed[i + 1];
      var modelName = internalModel.modelName;
      var relationshipsByName = modelNameToRelationshipsMap[modelName];

      if (relationshipsByName === undefined) {
        relationshipsByName = modelNameToRelationshipsMap[modelName] = get$3(internalModel.modelClass, 'relationshipsByName');
      }

      var relationshipNames = relationshipsByName._keys.list;
      var relationships = internalModel._relationships;

      var _loop3 = function _loop3(keyIndex) {
        var relationshipName = relationshipNames[keyIndex];
        var relationshipData = data.relationships[relationshipName];

        if (relationshipData !== undefined) {
          var relationshipRequiresNotification = relationships.has(relationshipName) || isInverseRelationshipInitialized(store, internalModel, data, relationshipName, modelNameToInverseMap);

          if (relationshipRequiresNotification) {
            relationships.get(relationshipName).push(relationshipData);
          }

          // in debug, assert payload validity eagerly
        }
      };

      for (var keyIndex = 0; keyIndex < relationshipNames.length; keyIndex++) {
        _loop3(keyIndex);
      }
    };

    for (var i = 0; i < pushed.length; i += 2) {
      _loop2(i);
    }

    pushed.length = 0;
  }
```

addon/-private/system/relationships/relationship-payloads.js
```diff
@@ -324,7 +324,7 @@ export default class RelationshipPayloads {
       return;
     }
 
-    if (Array.isArray(data)) {
+    if (data.length) { // dirty-check for "is-array"
```
stefanpenner: whats wrong with `Array.isArray` ?

addon/-private/system/ordered-set.js
```diff
@@ -16,6 +16,18 @@ OrderedSet.prototype = Object.create(EmberOrderedSet.prototype);
 OrderedSet.prototype.constructor = OrderedSet;
 OrderedSet.prototype._super$constructor = EmberOrderedSet;
 
+OrderedSet.prototype.pushMany = function pushMany(additions) {
```
stefanpenner: unit test

addon/-private/system/relationships/state/belongs-to.js
```diff
@@ -42,52 +40,101 @@ export default class BelongsToRelationship extends Relationship {
...
     // notifying our internalModel that we've "changed" and excessive thrash on
     // setting up inverse relationships
-    this.canonicalMembers.add(internalModel);
-    this.members.add(internalModel);
-    this.inverseInternalModel = this.canonicalState = internalModel;
-    this.setupInverseRelationship(internalModel);
+    this.currentState = this.canonicalState = internalModel;
+    this.setupInverseRelationship(internalModel, true);
+
+    // this.flushCanonicalLater();
```
stefanpenner: should these be removed?

addon/-private/system/relationships/state/implicit.js
```diff
@@ -0,0 +1,209 @@
...
+    //TODO remove once we have proper diffing
+    let newInternalModels = [];
+    for (let i = 0; i < list.length; i++) {
+      if (list[i].isNew()) {
+        newInternalModels.push(list[i]);
+      }
+    }
+
+    //TODO(Igor) make this less abysmally slow
+    this._currentState = this.canonicalState.copy();
```
stefanpenner: should we use the `currentState` setter?

addon/-private/system/relationships/relationship-payloads.js
bmeurer: I'd recommend to stick to `Array.isArray`, which gives you predictable performance, unless you now for sure that `data` is always an array.

addon/-private/system/relationships/state/implicit.js
runspired: it doesn't matter here in theory but I did mean to change this to use it once I added the setter.

addon/-private/system/relationships/state/belongs-to.js
runspired: yes

addon/-private/system/ordered-set.js
runspired: 👍

addon/-private/system/store.js
```diff
@@ -2417,20 +2419,63 @@ Store = Service.extend({
...
+        if (relationshipData !== undefined) {
+          let relationshipRequiresNotification = relationships.has(relationshipName) ||
+            isInverseRelationshipInitialized(store, internalModel, data, relationshipName, modelNameToInverseMap);
+
+          if (relationshipRequiresNotification) {
+            relationships.get(relationshipName).push(relationshipData);
+          }
+
+          // in debug, assert payload validity eagerly
+          runInDebug(() => {
```
bmeurer: If you'd strip the `runInDebug` before running Babel on it, then Babel might be able to avoid the closures.

runspired: 👍

addon/-private/system/relationships/relationship-payloads.js
runspired: 👍

runspired:
We should still do this, but with #5344 coming and other work including many cleanup commits and lazy-relationships landing over the past year, this has diverged too far from Master. A clean reboot will come soon :)

PR #4884, opened 2017-03-21
Defer serializer loading

hjdivad:
There's no need to load serializers if and until the adapter fetch promise resolves.


hjdivad:
appveyor failures look to be environmental

PR #4885, opened 2017-03-21
Lazy snapshot.type

hjdivad:
Make snapshot.type and snapshotRecordArray.type load classes lazily, so we don't pay the cost of loading the classes if we don't need to.


hjdivad:
appveyor failure looks environmental; there's no tests failing there

PR #4886, opened 2017-03-22
[BUGFIX DEPRECATE] fixes and deprecates when a user re-exports a model multiple places for use

runspired:
@bmac If good this should be backported to release and beta.

This twiddle demonstrates the issue: https://ember-twiddle.com/035a072e1818f48c09ac9a38f13b1773?openFiles=twiddle.json%2C

All Ember versions prior to 2.12 work with all ED versions. Beginning in 2.12 (I think we fussed with double extend) this begins breaking in ED.

We can temporarily fix this with a deprecation by extending the class ourselves.

### DIFF
--- a/addon/-private/system/store.js
+++ b/addon/-private/system/store.js
@@ -2124,7 +2124,7 @@ Store = Service.extend({
        */
       if (klass.modelName && klass.modelName !== modelName) {
         throw new Error(
-          `Ember Data found a Model for ${modelName} but it the class ` +
+          `Ember Data found a Model for ${modelName} but the class ` +
           `already had the modelName ${klass.modelName}.\n\n` +
           'This likely means that you are re-exporting a Model.\n\n' +
           `Replace:\n\texport { default } from './${klass.modelName}';\nwith:\n\t` +
--- a/addon/-private/system/store.js
+++ b/addon/-private/system/store.js
@@ -2137,8 +2137,12 @@ Store = Service.extend({
           until: '3.0'
         });
 
-        const extendedKlass = klass.extend();
-        owner.register(`model:${modelName}`, extendedKlass);
+        // simple (cheaper) extend that @krisselden says we can do
+        //  but which means `modelFor(<modelName>).extend() will not work
+        class extendedKlass extends klass {}
+        const typeStr = `model:${modelName}`;
+        owner.unregister(typeStr);
+        owner.register(typeStr, extendedKlass);
 
         const factory = this.modelFactoryFor(modelName);
         klass = hasFactoryFor ? factory.class : factory;
--- a/addon/-private/system/store.js
+++ b/addon/-private/system/store.js
@@ -2111,7 +2111,8 @@ Store = Service.extend({
       }
 
       // interopt with the future
-      let klass = getOwner(this).factoryFor ? factory.class : factory;
+      const hasFactoryFor = getOwner(this).factoryFor;
+      let klass = hasFactoryFor ? factory.class : factory;
 
       assert(`'${inspect(klass)}' does not appear to be an ember-data model`, klass.isModel);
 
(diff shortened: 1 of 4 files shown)


runspired:
I like the suggestion from @hjdivad. Since @stefanpenner is out, perhaps @krisselden or @rwjblue has thoughts on the best approach here?

runspired:
@hjdivad updated

addon/-private/system/store.js
```diff
@@ -2107,13 +2107,40 @@ Store = Service.extend({
...
+      if (klass.modelName && klass.modelName !== modelName) {
+        const Message =
+          `Ember Data found a Model for ${modelName} but the class ` +
+          `already had the modelName ${klass.modelName}.\n\n` +
+          'This likely means that you are re-exporting a Model.\n\n' +
+          `Replace:\n\texport { default } from './${klass.modelName}';\nwith:\n\t` +
+          `import Model from './${klass.modelName}';\n\n\texport default Model.extend();`;
+
+        if (hasFactoryFor) {
+          klass = factory.class = klass.extend();
```
rwjblue: It seems like a bug to me that ember allows setting factory.class like this...

runspired: if it is, then I don't think we have a path forward.

krisselden: factoryFor isn’t supposed to allow that, I thought we were using proxies to error in debug for that.

rwjblue: @krisselden - We do exactly that, for all other properties other than `create` and `class`. We should fix this (e.g. throw if class or create are attempted to be set).

rwjblue: @runspired - We could still do what you want I think. Either:

* do the double extend when we detect modelName is not a match, register, then look up again for them...
* create an cache our own "fake factory" here...

runspired: I think I prefer the register approach given it is "more correct" in terms of intent. Would be weird if ED and the registry were out of sync.

runspired: cc @rwjblue

stefanpenner: @rwjblue / @krisselden i totally agree with you guys, the issue is ember-data's reliance on `constructor.modelName`. I have removed as much of this dependence as can (I believe) without public API change...

addon/-private/system/store.js
```diff
@@ -2107,13 +2107,40 @@ Store = Service.extend({
...
-      // TODO: deprecate this
-      klass.modelName = klass.modelName || modelName;
+      // TODO: deprecate klass.modelName
+      /*
+        In test environments it is common to re-use the same base model in multiple
+        tests and register it each time. Since that seems convenient and good
+        to continue allowing this ignores the case where modelName is set but is
+         the same.
+       */
+      if (klass.modelName && klass.modelName !== modelName) {
```
stefanpenner: In theory (for the test case), we could detect if the original container has been destroyed or changed and if so, just seamlessly transition without a warning?

addon/-private/system/store.js
stefanpenner: I believe if users have tests that pollute the constructor, they should subclass in the own tests. I don't believe we should subclass for them.

addon/-private/system/store.js
runspired: ugh, I hadn't even considered tests mutating this yet. :'( 

Thoughts on how would we differentiate between "modelName set by previous container" and "modelName set by previous location's lookup"?

stefanpenner: if the `owner` is different? Maybe we store an owner version.

---

* I think tests should just double extend themselves (this is aligned with how the language works)
* we should just explicitly not support double re-exports of models until `modelName` goes away entirely.

runspired:
I believe this is what folks are hitting in #6143

PR #4892, opened 2017-03-27
[DOC] Documents the caching behavior of queryRecord

ryanlabouve:
This just adds some documentation describing the caching behavior of `queryRecord`

Reference: https://github.com/emberjs/data/issues/4733


bmac:
Thanks @ryanlabouve.

Issue #4755, opened 2017-01-16
remove relationship guard in relationship/state/relationship.js destroy


hjdivad:
This guard exists solely to help with bad state during test teardown.  The issue is that relationships can wind up in a bad state.  Remove the assertion and then clean up things that cause this kind of bad state.

One known example is that the type check for pushing onto a has many happens too late: in particular it happens after `members` has already been mutated.

PR #4893, opened 2017-03-27
Remove unnecessary relationship guard

drewchandler:
Fixes https://github.com/emberjs/data/issues/4755


bmac:
Thanks @drewchandler.

Issue #4754, opened 2017-01-16
Change setupStore helper default serializer to JSONAPISerializer


hjdivad:
It's currently `JSONSerializer` which causes oddities during test refactoring, as many tests push json-api documents into the store and you sometimes want to refactor this to be the fulfillment value of an adapter method.

hd719:
Hello I am new to open source and I would like to contribute to this project. 

-Thank you!

hjdivad:
hi @hd719 it's awesome that you want to help

This is an issue around cleaning up the tests.

In particular, when setting up a mock store, the default serializer is [DS.JSONSerializer](https://github.com/emberjs/data/blob/c99af930749cd697ab35ab2bab5a1f74137462b2/tests/helpers/store.js#L52) which is inconvenient because the internal representation is essentially json-api, so it would be much nicer to make it `DS.JSONAPISerializer`.

So doing this cleanup consists of

1. run the tests, verify they're green.
2. change that line to make the default serializer `DS.JSONAPISerializer`
3. see what tests break

Just doing this is already useful, but the next step is to clean up the tests that break.  Likely some of them will simply depend on manually specifying a serializer if they were relying on the current default of `DS.JSONSerializer`.

Here's an example of manually registering a serializer: https://github.com/emberjs/data/blob/master/tests/integration/store/query-record-test.js#L76  (note that `env` here is the result of `setupStore`, created in the [beforeEach](https://github.com/emberjs/data/blob/master/tests/integration/store/query-record-test.js#L21-L23) of this test).

eshtadc:
I didn't see this ever get a PR with all the test changes so I went ahead with one.  Chose to update test formats since I figured they were testing the "default" except in cases where it really seemed to have an intention of testing JSONSerializer specifically.  One of my first ember-data PRs so happy for any feedback :)

hjdivad:
#4894

hjdivad:
#5003

bmac:
Closed by https://github.com/emberjs/data/pull/5003

PR #4894, opened 2017-03-27
Change setupStore helper default serializer to JSONAPISerializer

tacocat:
This is an incomplete PR working towards the change requested in https://github.com/emberjs/data/issues/4754, but it requires a good chunk of tests to be updated to resolve with json api format.

Before updating the remaining tests, I wanted to double check this is the intended direction.


bmac:
@tacocat Thanks I like the direction of these changes and would be happy to merge it if/when you get the time to complete the remaining tests.

You may also need to rebase this pr on top of the latest master to get the tests passing on travis. I believe there was a bug on master when you originally forked the branch.

PR #4895, opened 2017-03-27
[PERF] single-pass over the initializers array

runspired:
Before we made three passes:

- map
- filter
- forEach

Now we make one.


bmac:
@runspired do you mind rebasing this when you get a chance?

stefanpenner:
thanks @run

Issue #4758, opened 2017-01-16
Assert when ManyArray.objectAt after one of the records has been unloaded


hjdivad:
Although it would be nice for this to work, we do not support sparse arrays.  It would probably be better to error rather than to have undefined behaviour (as we have today).

However, it is likely better as a first step to just warn, and then to see how pervasive the warning becomes, especially in test teardown.

mixonic:
Fixed in https://github.com/emberjs/data/pull/4896

runspired:
Additional fix in https://github.com/emberjs/data/pull/4912

stefanpenner:
#4896 does not fix this issue, I believe #4896 is actually performing the wrong check. Out of bounds access does not signify an unloaded record, rather it is just out of bounds access and should return undefined (no warning).

Rather the check here should be `internalModel !== undefined && internalModel.someFlagSayingIsHasBeenUnloadedAndIForgotTheName`

stefanpenner:
After chatting with @hjdivad we feel this is actually no longer as issue. If a relationship has a member that is unloaded, the entire relationship has also been dematerialized and must be refetched anyways.

PR #4896, opened 2017-03-27
Ember.warn when ManyArray.objectAt() is undefined

bagby:
First pass at #4758 . Not sure that there's a commit prefix that is applicable here, it's more to help with debugging, and to provide some data points as to the prevalence of the problem.

### DIFF
--- a/addon/-private/system/many-array.js
+++ b/addon/-private/system/many-array.js
@@ -2,7 +2,7 @@
   @module ember-data
 */
 import Ember from 'ember';
-import { assert } from "ember-data/-private/debug";
+import { assert, warn } from "ember-data/-private/debug";
 import { PromiseArray } from "./promise-proxies";
 import { _objectIsAlive } from "./store/common";
 import diffArray from './diff-array';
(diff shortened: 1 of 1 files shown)


bmac:
Thanks @bagby.

addon/-private/system/many-array.js
```diff
@@ -136,7 +136,12 @@ export default Ember.Object.extend(Ember.MutableArray, Ember.Evented, {
   objectAt(index) {
     let object = this.currentState[index];
     //Ember observers such as 'firstObject', 'lastObject' might do out of bounds accesses
-    if (object === undefined) { return; }
+    if (object === undefined) {
```
runspired: this needs to bounds check that `index` is not `length-1` or `0` because of the greediness of `firstObject` and `lastObject`.

runspired: https://github.com/emberjs/data/pull/4912

hjdivad: for future travelers the issue @runspired is referring to is https://github.com/emberjs/ember.js/issues/14843

stefanpenner:
Thank you for the contribution, unfortunately @hjdivad (who I am pairing with right now) forgot that https://github.com/emberjs/data/issues/4758 itself is no longer valid.

The PR as is, is slightly in-correct as warning for out of bounds wasn't the original intention, rather it was non out of bounds check, but the internalModel returned being unloaded is the case where the warning should have occured.

@hjdivad and @igorT changed the semantics, so that this case #4758 intended to fix does not occur, as the entire manyArray in this case has been destroyed.

PR #4897, opened 2017-03-27
Deprecate buildInternalModel, add _buildInternalModel.

yethon:
Prefixing the `buildInternalModel` private function with an underscore, based on this (slightly outdated) ticket [3250](https://github.com/emberjs/data/issues/3250).


bmac:
Thanks @yethon.

Issue #4752, opened 2017-01-16
rename `relationship.destroy`


hjdivad:
The semantics of `relationship.destroy` are odd given its name.

PR #4898, opened 2017-03-27
[BUGFIX beta] Rename `relationship.destroy` to `removeInverseRelationships`

sduquej:
Fixes https://github.com/emberjs/data/issues/4752

The semantics of destroy have changed since https://github.com/emberjs/data/commit/2f18405a4785486f3bd7c865ca1096ea44e2c184. This PR renames the method to convey better what it does.

### DIFF
--- a/addon/-private/system/model/internal-model.js
+++ b/addon/-private/system/model/internal-model.js
@@ -875,12 +875,12 @@ export default class InternalModel {
       if (this._relationships.has(name)) {
         let rel = this._relationships.get(name);
         rel.clear();
-        rel.destroy();
+        rel.removeInverseRelationships();
       }
     });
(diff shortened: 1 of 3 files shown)


bmac:
Looks good @sduquej. Would you mind rebasing this pr? I just want to make sure it passes CI since its renaming a method.

sduquej:
Done! 🤞

bmac:
Thanks @sduquej.

runspired:
We should revert this PR, `destroy()` was the correct method name as it was invoked when tearing down the instance, and it is not expected that you can use a relationship after it is called.

runspired:
Furthermore, this was definitely not a bugfix, or needed for beta.

PR #4901, opened 2017-03-27
[BUGFIX beta] Removes `get record` from InternalModel.

### LINKED ISSUE #4756: Check use of `internalModel.record`
`IntenalModel` has a `get record` as well as a `getRecord` which is a bit confusing.

Likely we can remove the getter and replace it with `InternalModel.hasRecord` but this remains to be verified.

esbanarango:
This PR removes `get record` from InternalModel 'cause there is already a `getRecord` method.

Fixes #4756

### DIFF
--- a/addon/-private/system/references/record.js
+++ b/addon/-private/system/references/record.js
@@ -112,7 +112,7 @@ RecordReference.prototype.push = function(objectOrPromise) {
    @return {DS.Model} the record for this RecordReference
 */
 RecordReference.prototype.value = function() {
-  return this.internalModel.record;
+  return this.internalModel._record;
 };
 
 /**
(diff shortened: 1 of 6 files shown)


addon/-private/system/references/record.js
```diff
@@ -112,7 +112,7 @@ RecordReference.prototype.push = function(objectOrPromise) {
    @return {DS.Model} the record for this RecordReference
 */
 RecordReference.prototype.value = function() {
-  return this.internalModel.record;
+  return this.internalModel._record;
```
sduquej: `getRecord()`?

bmac: @esbanarango do you mind changing this line to use `getRecord()`?

esbanarango: @bmac changed. 👍

bmac:
Thanks @esbanarango.

PR #4909, opened 2017-03-29
Deprecate didUpdateAll, add _didUpdateAll.

yethon:
Prefixing the didUpdateAll private function with an underscore, based on this (slightly outdated) ticket [3250](https://github.com/emberjs/data/issues/3250).


stefanpenner:
@yethon nice thanks!

PR #4910, opened 2017-03-30
[BUGFIX] initial-relationship-setup must handle null

runspired:


addon/-private/system/relationships/state/has-many.js
```diff
@@ -180,13 +180,15 @@ export default class ManyRelationship extends Relationship {
...
 
   setInitialInternalModels(internalModels) {
-    let args = [0, this.canonicalState.length].concat(internalModels);
-    this.canonicalState.splice.apply(this.canonicalState, args);
-    internalModels.forEach(internalModel => {
-      this.canonicalMembers.add(internalModel);
-      this.members.add(internalModel);
-      this.setupInverseRelationship(internalModel);
-    });
+    if (internalModels && internalModels.length) {
```
igorT: Why would it be true but not have a length?

runspired: I was just being overly cautious

runspired: I have only ever hit undefined or an array with items in practice

igorT: If it's not undefined but has no length, l think we are in a bad state. We should assert against that in debug maybe, but we shouldn't guard, that would just hide a bug

runspired: Dropped the length check, I suspect asserting length here would be asserting at the wrong layer.

PR #4912, opened 2017-04-03
[BUGFIX] guard hasMany#object-at-undefined against out-of-bounds access

runspired:
`firstObject` and `lastObject` are greedy, causing this warning to trigger often unnecessarily.

### DIFF
--- a/addon/-private/system/many-array.js
+++ b/addon/-private/system/many-array.js
@@ -137,9 +137,9 @@ export default Ember.Object.extend(Ember.MutableArray, Ember.Evented, {
     let object = this.currentState[index];
     //Ember observers such as 'firstObject', 'lastObject' might do out of bounds accesses
     if (object === undefined) {
-      warn(`ManyArray#objectAt(index) return undefined for index '${index}'. See https://github.com/emberjs/data/issues/4758`, false, {
-        id: 'ds.many-array.object-at-undefined'
-      });
+      warn(`ManyArray#objectAt(index) return undefined for index '${index}'. See https://github.com/emberjs/data/issues/4758`,
+           (index ===0 || index === this.currentState.length - 1),
+           { id: 'ds.many-array.object-at-undefined' });
       return;
     }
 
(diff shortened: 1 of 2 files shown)


runspired:
@stefanpenner I'm not sure how I would test this in a sustainable way. Any test we write now is a test that wont fail but which we do want to fail later when firstObject and lastObject are no longer greedy.

hjdivad:
@runspired https://github.com/runspired/data/pull/2 seems fine.  I agree that tests that fail when `firstObject` is no longer eager would be kinda nice, but testing this behaviour with appropriate comments seems better than not having tests at all; I don't think it's worth letting the perfect be the enemy of the good here.

runspired:
@hjdivad thanks! I wasn't saying we shouldn't take action, just that the added value over just documenting the reasoning in a comment didn't seem to be very high to me. Both is still good :D

hjdivad:
> @hjdivad thanks! I wasn't saying we shouldn't take action, just that the added value over just documenting the reasoning in a comment didn't seem to be very high to me. Both is still good :D

For re-adding the warning if/when `firstObject` is no longer eager, I agree, but there is the issue that comments are pretty bad at catching regressions 😀

stefanpenner:
Can we get a failing test in ember, I would like to address this at the source ASAP, such a test would very much help.

hjdivad:
> Can we get a failing test in ember, I would like to address this at the source ASAP, such a test would very much help.

@stefanpenner i'll put something together

hjdivad:
@stefanpenner https://github.com/emberjs/ember.js/pull/15110

updates for failing test: https://github.com/emberjs/ember.js/pull/15110/commits/8b851162583963832e6f6662415adbe9ee4341a2

suggested fix: https://github.com/emberjs/ember.js/pull/15110/commits/dabcbc29ca605490f74f220c470d1d339c71d7fb

That PR needs work before it could be merged, but the first order of business is to decide the tradeoff mentioned there, namely between

1. Having `objectAt` called eagerly for `firstObject` and `lastObject`, causing a great deal of grief for anybody trying to subclass `MutableArray` or
2. Not detecting that `[A,B].insertAt(0, A]` does not change `firstObject`

stefanpenner:
Something is fishy, I don't believe #4896 is correct and should likely be reverted which leads me to believe this PR is not required at all.

More details: https://github.com/emberjs/data/issues/4758#issuecomment-291702047

stefanpenner:
https://github.com/emberjs/data/issues/4758#issuecomment-291704259 for context, but we will be rolling back #4896

PR #4913, opened 2017-04-04
[BUGFIX canary] don't prematurely nullify props

runspired:
don't prematurely nullify props on the container-instance-cache. the cache's `destroy` method is called during the store's `willDestroy` meaning it may still be required before the Application has completed cleanup.


stefanpenner:
@runspired good catch. I don't believe this should be nulled out at all, the container-instance-cache.js isn't the root retainer. Rather the real container is the dominator, and it already gets cleaned up correctly.

PR #4914, opened 2017-04-05
Revert "Ember.warn when ManyArray.objectAt() is underfined"

stefanpenner:
This reverts commit 44ffbff63e4cae33ab519b23d51394645d7ccab1.

More details: https://github.com/emberjs/data/pull/4896#issuecomment-291706797

### DIFF
--- a/addon/-private/system/many-array.js
+++ b/addon/-private/system/many-array.js
@@ -2,7 +2,7 @@
   @module ember-data
 */
 import Ember from 'ember';
-import { assert, warn } from "ember-data/-private/debug";
+import { assert } from "ember-data/-private/debug";
 import { PromiseArray } from "./promise-proxies";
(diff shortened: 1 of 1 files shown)


stefanpenner:
FYI: @runspired

runspired:
I have thoughts, we may not totally want to revert, need to read up.

runspired:
Ok, read through the other comments and I agree this doesn't cover the unloaded case, and the sparse case is now handled separately.  We should continue with the revert.

As a parting note: off of master our app was hitting this very often. This suggests to me the sparse case is not totally handled correctly yet; however, in #4882 I noticed that `ManyArray` was created eagerly at times we did not intend and fixed that.  It may be that hitting the out of bounds check and the overly eager materialization of the ManyArray are the same problem.

PR #4917, opened 2017-04-05
[BACKPORT 4913] For release

stefanpenner:
cc @runspired / @bmac


addon/-private/system/store/container-instance-cache.js
```diff
@@ -1,8 +1,8 @@
 /* global heimdall */
 import Ember from 'ember';
+<<<<<<< HEAD
 import EmptyObject from "ember-data/-private/system/empty-object";
```
krisselden: Bad conflict resolution

PR #4919, opened 2017-04-10
[DOC] Fix findHasMany parameter order

timmorey:
Updated the documentation to reflect the actual order of parameters, which is incorrectly rendered in the api documentation (see https://emberjs.com/api/data/classes/DS.RESTAdapter.html#method_findHasMany).

I presume that the method signature shown in the doc (`findHasMany (store, snapshot, relationship, url)` for this method) is generated from these `@param` directives.  I copied the signature out of the doc and found myself confused by the parameter values.


bmac:
Thanks @timmorey.

PR #4922, opened 2017-04-12
[BUGFIX release] restore internalModels GUID_KEY’s

stefanpenner:
Without this, adding an internalModel to an orderd set will result in a new [GUID_KEY] prop being added, but using defineProperty https://github.com/emberjs/ember.js/blob/master/packages/ember-utils/lib/guid.js#L187. Both causing a shape change, and a potentially costly defProp.

Restoring this restores that, and addresses a category of OrderdSet related performance issues.

@rwjblue / @runspired

### DIFF
--- a/addon/-private/system/model/internal-model.js
+++ b/addon/-private/system/model/internal-model.js
@@ -119,7 +119,10 @@ export default class InternalModel {
   constructor(modelName, id, store, data) {
     heimdall.increment(new_InternalModel);
     this.id = id;
-    this._internalId = InternalModelReferenceId++;
+
+    // this ensure ordered set can quickly identify this as unique
+    this[Ember.GUID_KEY] = InternalModelReferenceId++ + 'internal-model';
+
     this.store = store;
     this.modelName = modelName;
     this._loadingPromise = null;


addon/-private/system/model/internal-model.js
```diff
@@ -119,7 +119,10 @@ export default class InternalModel {
   constructor(modelName, id, store, data) {
     heimdall.increment(new_InternalModel);
     this.id = id;
-    this._internalId = InternalModelReferenceId++;
+
+    // this ensure ordered set can quickly identify this as unique
+    this[Ember.GUID_KEY] = InternalModelReferenceId++ + 'internal-model';
```
stefanpenner: Was originally added as part of making stuff faster: https://github.com/emberjs/data/commit/ca9393fb01f23038eb15c479c56e02be4a466558

Left a comment to hope it survives

runspired: The comment will help this survive a lot better, the assumption before was that this shaping was to help the cases where internalModel was being abused by DS.Model (no longer is), meta, computeds and other public interface things.

stefanpenner: understood, the secret handshake this is doing is not obvious at all. Im not surprised it was lost.

PR #4920, opened 2017-04-11
Svelted some our builds a bit (3KB savings)

runspired:
This takes current master's `addon/` modules from 160.04KB and 84 modules in prod to 157.01 KB and 79 modules

It also adds the script `lib/analyze-modules.js` that can be used for a granular breakdown of the ember-data modules in a minified production version of vendor.js

Example:

```
Processing ember-data modules from: dist/assets/vendor-601321e0c286e71176ce7910ad0436e4.js


Ember Data Asset Size Report
================================================
	Total: 157.01 KB
	Modules: 79
================================================
15.14 KB	9.6 %	ember-data/-private/system/store
14.47 KB	9.2 %	ember-data/-private/system/model/internal-model
 9.51 KB	6.1 %	ember-data/adapters/rest
 8.67 KB	5.5 %	ember-data/serializers/json
 6.92 KB	4.4 %	ember-data/-private/system/model/model
 5.87 KB	3.7 %	ember-data/-private/system/model/states
 5.71 KB	3.6 %	ember-data/-private/system/relationships/state/relationship
 5.61 KB	3.6 %	ember-data/serializers/json-api
 5.51 KB	3.5 %	ember-data/-private/system/relationships/state/has-many
 4.77 KB	3.0 %	ember-data/serializers/rest
 4.53 KB	2.9 %	ember-data/-private/system/record-array-manager
 4.52 KB	2.9 %	ember-data/serializers/embedded-records-mixin
 3.96 KB	2.5 %	ember-data/-private/system/relationships/state/belongs-to
 3.47 KB	2.2 %	ember-data/-private/system/store/finders
 3.36 KB	2.1 %	ember-data/index
 2.97 KB	1.9 %	ember-data/-private/system/snapshot
 2.81 KB	1.8 %	ember-data/-private/system/relationships/relationship-payloads
 2.32 KB	1.5 %	ember-data/-private/system/many-array
 2.13 KB	1.4 %	ember-data/-private/system/record-arrays/record-array
 2.02 KB	1.3 %	ember-data/-private/system/references/has-many
 1.99 KB	1.3 %	ember-data/adapters/errors
 1.95 KB	1.2 %	ember-data/-private/system/model/errors
 1.95 KB	1.2 %	ember-data/adapters/json-api
 1.91 KB	1.2 %	ember-data/-private/adapters/build-url-mixin
 1.86 KB	1.2 %	ember-data/-private/system/debug/debug-adapter
[...]

PR #4924, opened 2017-04-14
[PERF] Model.prototype.didDefineProperty only in dev

stefanpenner:
`meta.parentType` (which is produced by) `didDefineProperty` is only required for development assertions. As development assertions are stripped in prod, this `didDefineProperty` can also be stripped

### DIFF
--- a/addon/-private/system/model/model.js
+++ b/addon/-private/system/model/model.js
@@ -1050,46 +1050,6 @@ const Model = Ember.Object.extend(Ember.Evented, {
   notifyBelongsToChanged(key) {
     this.notifyPropertyChange(key);
   },
-
-  /**
-   This Ember.js hook allows an object to be notified when a property
-   is defined.
-
-   In this case, we use it to be notified when an Ember Data user defines a
-   belongs-to relationship. In that case, we need to set up observers for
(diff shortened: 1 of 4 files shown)


addon/-private/system/model/model.js
```diff
@@ -1924,4 +1884,47 @@ if (isEnabled('ds-rollback-attribute')) {
   });
 }
 
+runInDebug(() => {
+  Model.reopen({
```
runspired: I believe this should be reopenClass as this was a static method?

runspired: actually scratch that I think I read the diff wrong on the original location.

stefanpenner:
` ember test --environment=production` needs some TLC

runspired:
@stefanpenner this is fixed by another of my PR's, the issue is that the env is "test" at runtime but "production" at build time, so our code stripping needs to take account a little differently.

runspired:
@stefanpenner actually, looks like you merged it: https://github.com/emberjs/data/pull/4920

stefanpenner:
Also tidying up the other side of this in ember https://github.com/emberjs/ember.js/pull/15141 (no cross dependencies exist)

Issue #4927, opened 2017-04-15
ember-data canary breaks builds


kratiahuja:
Seems like this [commit](https://github.com/emberjs/data/commit/a2785b2442b824947f59b94f6a2d0399b5dd708f) introduced `let` and `const` which is not supported in Node 4 in strict mode. This is causing `ember-cli-fastboot` tests to break. See [travis logs](https://travis-ci.org/ember-fastboot/ember-cli-fastboot/jobs/222429161)

Going to send a PR to fix this. 

cc: @runspired

PR #4928, opened 2017-04-15
Don't use `let` or `const`

kratiahuja:
It is breaking `ember-cli-fastboot` tests. Fixes: #4927 

cc: @runspired @stefanpenner @hjdivad

### DIFF
--- a/lib/transforms/babel-plugin-remove-imports.js
+++ b/lib/transforms/babel-plugin-remove-imports.js
@@ -1,8 +1,8 @@
 var path = require('path');
 
 function removeImports() {
-  let importDeclarationsToRemove;
-  let filteredImports;
+  var importDeclarationsToRemove;
+  var filteredImports;
 
   return {
     name: 'remove-filtered-imports',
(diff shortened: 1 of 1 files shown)


runspired:
Thanks @kratiahuja. I forget we can't have nice things ;)

kratiahuja:
On that note, why is `ember-data` only being tested against Node 6 (I thought ember-cli supports Node 4)? [Travis config shows only Node 6] (https://github.com/emberjs/data/blob/master/.travis.yml#L5).

runspired:
@bmac seems to have introduced the node 6 change.

bmac:
I changed travis to use node 6 because master was broken during the contributor workshop and I wanted to unblock contributors. Sounds like I need to revert back to node 4 and debug the real issue.

rwjblue:
let and const are fine in node 4, you just have to ensure `'use strict';` is at the top of the file.

rwjblue:
@bmac - the original reason there was a failure around the time of EmberConf was that a new version of ember-try introduced some issues with npm@2 (which is used by node 4). That was fixed within a couple days of EmberConf, so it should be good now...

PR #4933, opened 2017-04-19
Preheat GUID_KEY on attr/belongsTo/hasMany meta for Ember.Map perf

runspired:
While investigating the perf of a loop over each attribute and relationship, I noticed that we spent a particularly egregious amount of time in `Map.set` (in the screenshot this is the call to `set` within `(anonymous)`.  This was due to us always transitioning Meta in a more expensive manner than is necessary.

This PR preheats meta similar to how we preheat GUID_KEY for Ember.OrderedSet on `InternalModel`
 
screen shot 2017-04-19 at 1 17 47 am

### DIFF
--- a/addon/-private/system/relationships/belongs-to.js
+++ b/addon/-private/system/relationships/belongs-to.js
@@ -2,6 +2,8 @@ import Ember from 'ember';
 import { assert, warn } from "ember-data/-private/debug";
 import normalizeModelName from "../normalize-model-name";
 
+let PreheatGuid = 0;
+
 /**
   `DS.belongsTo` is used to define One-To-One and One-To-Many
   relationships on a [DS.Model](/api/data/classes/DS.Model.html).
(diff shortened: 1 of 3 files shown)


runspired:
After looking into this further, the issue is not `meta` but rather that we pass the `String` `name` as the key for the map, which forces the map to perform many expensive checks before settling on it being a String and doing this: https://github.com/emberjs/ember.js/blob/v2.12.0/packages/ember-utils/lib/guid.js#L160

This perf issue would be mitigated by removing the use of `Map` from these ModelClass schema uses or providing an alternate happy path off of them for perf sensitive folks.

stefanpenner:
@runspired which map (in ember-data) was this for?

stefanpenner:
this one? https://github.com/emberjs/data/blob/83c2e95524986b3adec7e80f86d4d0d1c0ef3841/addon/-private/system/relationships/ext.js#L16-L18

stefanpenner:
@runspired if its that one, yes lets just kill the `map`

stefanpenner:
We should likely also fast-path this in maps/sets. Ill try to remember

runspired:
@stefanpenner I hit this in `relationshipsByName` but most of the maps created by Model are using string keys.

PR #4925, opened 2017-04-14
[SVELTE] rollup -private for winnings

runspired:
This PR de-cycles our import/export structure into and out-of `-private`, allowing us to use rollup to condense the `-private` directory.

It took 2 liberties with file structure changes to differentiate between public and private:

- the deprecated instance-initializer is now in `addon/instance-initializers`
- the initializers have been flattened into `setup-container` (and could be made a bit nicer still)

Additionally, `Transform` and Adapter `Errors` are now re-exported from `-private`, as they are occasionally used in the internals. I don't think transforms and errors belong there and only a very tiny portion of them is actually used internally. As they are entirely public interface we should consider how to divide / extract.

*TODO*

- [x] ensure globals build has not regressed
- [-] ~ensure globals build gets the same winnings~
- [x] ensure production builds still strip all code correctly
- [x] build time differences (cold, warm, rebuild)
- [x] output load time differences (via kris's method) <-- just due diligence (then we can decide if we want to invest more in this approach for other things (like apps) and also make sure we don't troll ourslves)
- [ ] ship it!

**Before**

```
ember-data Asset Size Report
=====================
        Total: 160.06 KB
        gZip: 34.56 KB
        Modules: 84
=====================
  Bytes         Percent   Gzip Bytes    Percent Module Name
-----------------------------
15.28 KB        9.5 %    4.66 KB        13.5 %  ember-data/-private/system/store
14.52 KB        9.1 %    3.90 KB        11.3 %  ember-data/-private/system/model/internal-model
 9.54 KB        6.0 %    2.86 KB        8.3 %   ember-data/adapters/rest
 8.70 KB        5.4 %    2.36 KB        6.8 %   ember-data/serializers/json
 6.95 KB        4.3 %    2.24 KB        6.5 %   ember-data/-private/system/model/model
 5.54 KB        3.5 %    1.77 KB        5.1 %   ember-data/-private/system/relationships/state/has-many
 4.80 KB        3.0 %    1.60 KB        4.6 %   ember-data/serializers/rest
 5.64 KB        3.5 %    1.49 KB        4.3 %   ember-data/serializers/json-api
 5.74 KB        3.6 %    1.48 KB        4.3 %   ember-data/-private/system/relationships/state/relationship
 5.90 KB        3.7 %    1.45 KB        4.2 %   ember-data/-private/system/model/states
 4.56 KB        2.8 %    1.37 KB        4.0 %   ember-data/-private/system/record-array-manager
[...]

PR #4936, opened 2017-04-22
[SVELTE] Transforms & remove Date.parse polyfill

runspired:
~This PR branches off of `svelte/rollup-private` and will need that to land first, the relevant commit is https://github.com/emberjs/data/pull/4936/commits/89f4c9d369ae631f9966e630be8d5fec1097ba84~ **Landed** :)

- This PR brings Transforms back into the public folder (they had moved into the Private folder during the first phase of rollup).

- The motivation for the original move was due to the location of initializers (later fixed in that PR) as well as the private `Date.parse` polyfill.

- In addition to bringing these back into the public folder, this removes the `Date.parse` polyfill as much as possible, adding a new deprecation for a corner case which our polyfill handled but which is not part of the spec.

**Removing the `Date.parse` polyfill has been debated before, here's a short TL;DR**

Our custom Date parsing logic pre-dated and polyfilled the ES5 behavior; however, we now only support ES5 browsers.

In ES2015, initially there was a breaking change which prevented us from removing our custom parsing logic, however, that breaking change was reverted and not implemented by major browsers.  We can now safely rely on the native `Date.parse`.

Previous discussions of our own here: https://github.com/emberjs/data/pull/2685

See: https://github.com/tc39/ecma262/pull/138
and: https://github.com/tc39/ecma262/issues/87

**However**, it turns our the polyfill we are using does not conform to the ECMA variant of ISO 8601, with a minor difference in that it allows for a "shorthand" timezone offset which is part of the greater spec but not ECMA. 

This means that going immediately to the native `Date.parse` is not possible without a deprecation, which has been detected and added here: https://github.com/emberjs/data/pull/4936/commits/74aef8a0cbef375a4b019ddb2cd8470ea6c1e66f#diff-6da4887e7f026f285304a8bbb0f06e68R64


stefanpenner:
I like the idea, https://github.com/emberjs/data/pull/4925 is waiting for you though. Lets get that one in, so we can look at this one.

tests/unit/transform/date-test.js
```diff
@@ -39,28 +38,22 @@ test('#deserialize', function(assert) {
   assert.equal(transform.deserialize(undefined), null);
 });
 
-test('#deserialize with different offset formats', function(assert) {
+testInDebug('#deserialize with different offset formats', function(assert) {
```
stefanpenner: I think we should likely always test this, but only test the expect deprecation in debug.

stefanpenner: @runspired thoughts?
 Write

runspired: I agree, but I'm not sure (based on how that stripping happens) how to set this up at a glance. I will think.

PR #4937, opened 2017-04-22
[BUGFIX beta] Update dependencies to use Babel 6.

rwjblue:
In ember-cli@2.13 we intend to ship only one version of babel by default (though having multiple versions does work perfectly fine).

These deps have just been released, the only breaking change (to them) was bumping the min engine version to Node 4 which we have already done here.

@bmac - Is it possible to land this into 2.13.0 before shipping it?

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,8 +34,8 @@
     "ember-cli-string-utils": "^1.0.0",
     "ember-cli-test-info": "^1.0.0",
     "ember-cli-version-checker": "^1.1.4",
-    "ember-inflector": "^1.9.4",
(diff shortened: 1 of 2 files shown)


bmac:
Looks like there are some dependency errors in the build but we should be able to fix those and get this merged in before the `2.13` release.

rwjblue:
I tweaked the `peerDependencies` entry for `ember-inflector`.  That should fix at least one of the issues that Travis had...

rwjblue:
Sweet, looks like that was the issue.

@bmac - Passing now.

bmac:
Thanks @rwjblue.

PR #4939, opened 2017-04-25
Handle invalid response (422) in rest adapter

edance:
Not sure if this is the full fix but this change fixed the issue for me.

According to the documentation for the Rest Serializer, errors with 422 should look like this:

```javascript
{
  "errors": {
    "msg": "Something went wrong"
  }
}
```

This resulted in an error being thrown:

"Assertion Failed: `AdapterError` expects json-api formatted errors array."

### DIFF
--- a/addon/adapters/rest.js
+++ b/addon/adapters/rest.js
@@ -963,7 +963,7 @@ const RESTAdapter = Adapter.extend(BuildURLMixin, {
     if (this.isSuccess(status, headers, payload)) {
       return payload;
     } else if (this.isInvalid(status, headers, payload)) {
-      return new InvalidError(payload.errors);
+      return new InvalidError([payload.errors]);
     }
 
     let errors          = this.normalizeErrorResponse(status, headers, payload);


stefanpenner:
I believe `payload.errors` should be an array of pojos, where each pojo is a different error.

stefanpenner:
rest adapter wants a pojo: https://github.com/emberjs/data/blob/2fabe2e86dfa6561d9d0038d4726e5fcc4658320/addon/adapters/rest.js#L179-L189
json API adapter wants an array of pojos

addon/adapters/rest.js
```diff
@@ -963,7 +963,7 @@ const RESTAdapter = Adapter.extend(BuildURLMixin, {
     if (this.isSuccess(status, headers, payload)) {
       return payload;
     } else if (this.isInvalid(status, headers, payload)) {
-      return new InvalidError(payload.errors);
+      return new InvalidError([payload.errors]);
```
stefanpenner: I believe this actually must be run through `this.normalizeErrorResponse`  -> https://github.com/emberjs/data/blob/2fabe2e86dfa6561d9d0038d4726e5fcc4658320/addon/adapters/rest.js#L1151-L1171

Like:

```js
new InvalidError(this.normalizeErrorResponse(status, headers, payload);
```

HuggableSquare:
This actually happens with any type of error, the `AdapterError` `extend` method is looking for `errors` to be an array, and if it isn't it throws an error, even though as far as I can tell it will work even it's an object.

addon/adapters/rest.js
fivetanley: I think this is probably the best approach.

runspired:
not sure there's anything for us to do here. We largely have moved off the paradigms this was looking to patch.

PR #4941, opened 2017-04-25
only coerce non-string id's

stefanpenner:


runspired:
I'm slightly surprised the type I'd check beats adding an empty string to another string.

PR #4943, opened 2017-04-25
hash-for-dep only the plugin (not all of ember-data) as this plugin h…

stefanpenner:
…as no dependencies.


stefanpenner:
windows CI failure is NPM related, and the second windows build (identical) passed

PR #4946, opened 2017-04-26
dummy app routes should work with and without heimdall

stefanpenner:
- [ ] ensure these routes are tested so they don't regress


stefanpenner:
cc @runspired

PR #4947, opened 2017-04-26
remove needless change events when creating a recordArrays

stefanpenner:
When possible we should create new record arrays with there content, rather then creating them, and later adding there content. This prevents un-needed change events from firing. Some Older API's make this a tad tricky, but modern usages should not be penalized.

- [x] liveRecordArrays
- [x] adapterPopulatedRecordArrays


addon/-private/system/record-array-manager.js
```diff
@@ -470,3 +479,50 @@ function remove(array, item) {
 
   return false;
 }
+
+function updateLiveRecordArray(array, internalModels) {
+  let modelsToAdd = [];
+  let modelsToRemove = [];
+
+  for (let i = 0; i < internalModels.length; i++) {
```
runspired: Does our new `diffArray` method work here or is the data structure different?

stefanpenner: Interesting. I didn't think of that, I should try

tests/integration/record-arrays/adapter-populated-record-array-test.js
```diff
@@ -161,6 +161,75 @@ test('recordArray.replace() throws error', function(assert) {
   }, Error('The result of a server query (on person) is immutable.'), 'throws error');
 });
 
+test('only pass record array if explicitly in arity', function(assert) {
```
stefanpenner: Will update

stefanpenner:
fun fun

```sh
Error: The addon `ember-cli-htmlbars` requires the npm package `ember-cli` to be above 0.1.2, but you have undefined.
```

rwjblue:
Hmm. Can you double check and confirm we have ember-cli-htmlbars@1.3.2? I released a bad patch release earlier (1.3.1)...

stefanpenner:
@rwjblue the issue did start with 1.3.2 but it appears to not be the fault of 1.3.2. Something either in the dep checker/ember-cli or blueprint-test helpers is causing the grief.

stefanpenner:
@rwjblue TL;DR

These tests https://github.com/emberjs/data/blob/master/node-tests/blueprints/transform-test.js#L45 create a project at `tmp/<something>/my-app` that have a node_modules that contains only `ember-cli-mocha`, but the package.json contains ember-cli-htmlbars. Due to the node resolution algorithm ember-cli-htmlbars is being discovered in the node_module directory of ember-data itself.

I believe modifyPackages may not be working correctly..

stefanpenner:
finally green, although appveyor test are still hanging (unrelated to this PR, but I think we know whatsup in ember-cli land causing this...)

PR #4950, opened 2017-04-26
[PERF] Guard against poorly optimized babel output

runspired:
This PR will fail to build until we have fixed all the locations in which our babel compilation step is producing sub-optimal output to handle `let` and `const` closures.

When this PR is green, we can merge it and gain perfs.

cc @stefanpenner

### DIFF
--- a/addon/-private/system/store.js
+++ b/addon/-private/system/store.js
@@ -933,29 +933,31 @@ Store = Service.extend({
 
       let groups = adapter.groupRecordsForFindMany(this, snapshots);
 
-      for (let i = 0, l = groups.length; i < l; i++) {
-        let group = groups[i];
-        let totalInGroup = groups[i].length;
-        let ids = new Array(totalInGroup);
-        let groupedInternalModels = new Array(totalInGroup);
+      for (var i = 0, l = groups.length; i < l; i++) {
+        var group = groups[i];
+        var totalInGroup = groups[i].length;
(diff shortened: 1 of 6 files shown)


stefanpenner:
the thrown error is good (from ci):

```
The Broccoli Plugin: [broccoli-persistent-filter:Babel] failed with:
SyntaxError: ember-data/serializers/rest.js: Compiling let/const in this block would add a closure (throwIfClosureRequired).
  254 |     let keys = Object.keys(payload);
  255 | 
> 256 |     for (let i = 0, length = keys.length; i < length; i++) {
      |                                                            ^
  257 |       let prop = keys[i];
  258 |       let modelName = prop;
  259 |       let forcedSecondary = false;
```

stefanpenner:
@runspired should i fix?

runspired:
@stefanpenner yes. However, it stops the build on the first error encountered, you may just encounter the next one :)

stefanpenner:
looking now

stefanpenner:
this made me finally fix: https://github.com/stefanpenner/broccoli-persistent-filter/pull/106/commits/d7ab2d1708c16b4f851c32e4f6611e0412bfb702 :trollface:

stefanpenner:
oops broke something...

stefanpenner:
> oops broke something...

fixed locally, lets see what `try` says.

runspired:
Thanks @stefanpenner :)

PR #4956, opened 2017-04-29
Add failing test for serializing embedded recursive relationship

neborn:
This is a failing test that provides an example of an issue with calls to serialize resulting in a stack overflow if the model contains a recursive many-to-many relationship.

I don't think this case really needs to be supported. Ultimately we ended up adding `inverse: null` to the relationship as a resolution. But a more descriptive error message would have been helpful in tracking down the issue.

Thank you for your consideration.

cc. @stefanpenner

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,77 @@
 
 ### Master
 
+
+### Release 2.13.0 (April 28, 2017)
+- [#4860](https://github.com/emberjs/data/pull/4860) Refactor the detection / warnings around ember-cli-shims.
+- [#4853](https://github.com/emberjs/data/pull/4853) [Fixes #4853] use Object.create(null) directly
+- [#4858](https://github.com/emberjs/data/pull/4858) [PERF] let GUID_KEYS just be set on the model instances
+- [#4937](https://github.com/emberjs/data/pull/4937) [BUGFIX beta] Update dependencies to use Babel 6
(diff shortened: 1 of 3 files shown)


stefanpenner:
@neborn awesome thanks. I'll take a look early this week

stefanpenner:
@neborn interesting, this isn't even an actual cycle and its stack overflowing.... Looking further. (thanks for the reproduction!)

stefanpenner:
Ah ... its the fact the hasMany is automatically inferred on both sides, coupled with embedded that causes the cycle to exist.

stefanpenner:
Some thoughts:

* detect this exact scenario (embedded self-referential relationships with inferred inverse) and throw
* catch stack overflows and try to augment them with more useful information.

stefanpenner:
I think `relationship.type === snapshot.modelName && relationship.options.inverse !== null` check covers the cases I believe are issues.

stefanpenner:
will fix CI in the AM

stefanpenner:
Not sure whats up, can't seem to replicate locally... :sadpanda:

stefanpenner:
we should change the assertions to `if (DEBUG) { /*...*/ }` to get around the mysterious issue...

runspired:
closing as based on the convo the original issue was never replicated to be able to fix and this code is for legacy things we are working to remove. Newer patterns don't involve EmberData providing serializers, not serializing cycles becomes a matter of an app using a `Set` or `Map` and tracking seen.

PR #4958, opened 2017-04-29
[BUGFIX beta] Move broccoli-stew to dependencies

bmac:


runspired:
I dont think this is necessary, I believe the correct change is to remove it in `index.js`

cc @stefanpenner

PR #4959, opened 2017-04-29
[BUGFIX] remove forgotten broccoli-stew import

runspired:
cc @bmac


bmac:
👍

runspired:
@bmac I'm not sure why this (and other) test runs are failing. Only canary seems to have a problem but even there I don't see an obvious culprit?

stefanpenner:
I'll take a look at the failure. Thanks for catching this, my bad :(

PR #4960, opened 2017-04-29
ensure we don't compile typeof checks poorly

runspired:
When doing a `typeof` comparison against anything other than a string, Babel compiles the check into a much more expensive function call.  Some examples of this which are present in our code include:

```js
const type = typeof thing;
```

and

```js
switch (typeof Foo)
```

Example output:

```js
var _typeof = typeof Symbol === "function" && typeof Symbol.iterator === "symbol" ? function (obj) { return typeof obj; } : function (obj) { return obj && typeof Symbol === "function" && obj.constructor === Symbol && obj !== Symbol.prototype ? "symbol" : typeof obj; };

var type = typeof thing === "undefined" ? "undefined" : _typeof(thing);
```

To deal with this, because we do not rely on iterator or Symbol in our code, I added `transform-es2015-typeof-symbol` to the list of excluded babel plugins.


runspired:
@hjdivad test failures are unrelated, all PRs are failing. I spent some time digging in over the weekend and it looks to be because our yarn lock file installs a different set of dependencies locally than we get when npm installing in CI.  I haven't totally unwound which dependencies need to be officially bumped in order to make the failures stop.

stefanpenner:
@runspired 👍

PR #4965, opened 2017-05-03
[BUGFIX beta]  Skip test which doesn't play nicely with the latest em…

bmac:
…ber-cli and is failing on master

This is a temporary change to unblock Ember Data

See https://github.com/ember-cli/ember-cli/pull/7019 for a more long term fix


stefanpenner:
@bmac looks like the same issue my be affecting another test

Issue #4968, opened 2017-05-04
ember-data 2.13+ throws error when booting ember-engines


rwjblue:
The deprecations added as part of https://github.com/emberjs/data/pull/4657 are written in a way that assumes the argument passed into the `instance-initializer` has an `application` property. This is true for normal applications, but is not true for engines (which still run our `instance-initializers`). 

Specifically [this line] errors (because `application.application` is `null`):

```
  deprecateOldEmberDataInitializers(application.application.constructor.initializers);
```

rwjblue:
I'm looking at fixing now...

rwjblue:
Submitted PR's to fix ( #4969 and #4970).

sly7-7:
Likely fixed 😄  Thanks @rwjblue

PR #4970, opened 2017-05-04
[BUGFIX release] Ensure Engines can boot without error.

rwjblue:
The argument passed into `instance-initializers` is _either_ an `Ember.ApplicationInstance` instance _or_ a `Ember.EngineInstance` instance. The `.application` property is only present on `Ember.ApplicationInstance`'s.

This change adds some inline comments around the various conditionals and swaps the logic to use `instance.base` when present and avoid erroring if neither `.base` or `.application` is present.

---

This mirrors the functionality changes made in #4969, but targets `release` branch.

Closes https://github.com/emberjs/data/pull/4968

### DIFF
--- a/addon/-private/instance-initializers/initialize-store-service.js
+++ b/addon/-private/instance-initializers/initialize-store-service.js
@@ -4,15 +4,20 @@ import { deprecate } from 'ember-data/-private/debug';
   store.
 
   @method initializeStoreService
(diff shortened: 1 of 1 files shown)


bmac:
Travis was red but the only failure was from an unrelated thats already fixed on master. I'm going to merge this and cherry-pick the build fix from master to the release branch.

rwjblue:
:+1: Thank you!

Issue #4972, opened 2017-05-04
Saving a new record that returns a existing record now fails.


arenoir:
So this is certainly an edge case but in the past was a nice feature.  

I have a model `address` quite simple 4 attributes `lineOne`, `city`, `state` and `zip`.  When creating a new address the server checks to see if a duplicate exists and if so returns the duplicate with the 201 created http code.

This all works fine unless the duplicate is already loaded.  If it is loaded I get the following error.

```
You cannot update the id index of an InternalModel once set. Attempted to update 66.
```

In versions of ember data 2.11 and prior this worked without issue.

arenoir:
I am thinking this being an edge case I will just break into a custom ajax request.

hernanvicente:
Similar problem here, we were having this trouble too, in our case our ember application has a bank account model and the user was able to add and delete accounts, but in our API if an account was  deleted by the user the backend handled to only deactivate the record in the database.  

Later if the user creates again the same account, the API activates the record in the database and returns it with 201 status code and we're getting the same error, then we solved it, unloading the record from the store inside the destroyRecord function.

It's worth mentioning that we haven't this problem with ember data 2.8.1 and it starts to fail when we did upgrade to 2.13.1

Here a chunk of our controller code

````
import Ember from 'ember';

export default Ember.Controller.extend({
  actions: {
    save(bankAccount) {
      let self = this;
      bankAccount.save().then(function() {
        self.transitionToRoute('bank-accounts.index');
      }, function(response){});
    },
    delete(bankAccount) {
      let self = this;
      bankAccount.destroyRecord().then(function() {
        self.store.unloadRecord(bankAccount);
        self.transitionToRoute('bank-accounts.index');
      }, function(response){});
    }
  }
});
````

For us the trick was to unload the duplicated record that was already loaded into the store.

arenoir:
I'm reopening this because I think it should be discussed more

stevesims:
I'm seeing this `Assertion Failed` error too - although the circumstances are slightly different.

With our app the server will send push updates to the client.  I suspect I'm seeing this when the pushed version of the data gets sent back before the `save` of the record gets it's results sent back (or processed).  Unfortunately though I can't verify that is exactly what is actually happening, however I do know that the `id` being reported in the assertion failure does not match up to any records in the store *before* the `save` is called.

arenoir:
[...]

PR #4973, opened 2017-05-04
allow saving a new record to return a existing record

arenoir:
This fixes an issue #4972 that I am having. I am thinking this should probably be an addon something like adding findOrCreate to the store.


arenoir:
Abandoning in favor of custom ajax request.

PR #4981, opened 2017-05-11
[DOC release] Document DS.HasManyReference.ids

bantic:
The existing inline doc incorrectly marks the method name as
`remoteType`.


pangratz:
Vielen Dank!

PR #5008, opened 2017-06-09
[BUGFIX BETA]: Fixed export regression. `ember-data/transform` to be default.

workmanw:
@stefanpenner made reference to this in the original PR (https://github.com/emberjs/data/commit/24f41209b7ab9978bcdf029e50e3e62564d0b50f#commitcomment-22121947), but I think it was just overlooked. Either way this created a regression for us when using: `import Transform from 'ember-data/transform';`.

### DIFF
--- a/addon/transform.js
+++ b/addon/transform.js
@@ -1,1 +1,1 @@
-export { Transform } from './-private';
+export { default } from './transforms/transform';


addon/transform.js
```diff
@@ -1 +1 @@
-export { Transform } from './-private';
+export { Transform as Default } from './-private';
```
stefanpenner: shouldn't this be lowercase?

workmanw: 😬 Yea ... I'm not sure how that worked. I guess the tests don't exercise that.

workmanw:
@stefanpenner So it looks like the `transform` code was moved again after the first commit. Not it resides in `ember-data/transforms/transform`, but we also want to map it to `ember-data/transform`.

Because it's a default export from `ember-data/transforms/transform` and really the goal is to just map the contents, I chose to just do:
```
export * from './transforms/transform';
```

Alternative I think I could have done:
```
export { default } from './transforms/transform';
```

Let me know if you have a preference.

workmanw:
Err. Actually I'm not actually sure if `export { default } from './transforms/transform';` is valid. It may happen to just work because of how babel transpiles. It's unclear to me by looking at the spec ( https://tc39.github.io/ecma262/#sec-exports ). @stefanpenner do you know by chance? Worst case is I could import with one statement and reexport default as a second.

Who knew one line of code could be so complicated 😜

runspired:
@workmanw this is fixed elsewhere I thought it landed

workmanw:
@runspired Hmm. It's not in master or beta. I also checked the recent most PRs. I don't see any sign of it. Maybe it was in a close (unmerged) PR?

bmac:
Thanks @workmanw.

bendemboski:
The bug that this PR fixed just got released in 2.14.0, but this fix isn't in that release, it's in the 2.15 track, so this regression exists in the latest release :(

bmac:
Thanks for the heads up @bendemboski. I missed this when cherry picking commits for the release. I'll release a  2.14.1 with this commit as soon as I get a chance.

bmac:
Ember Data [2.14.1](https://github.com/emberjs/data/releases/tag/v2.14.1) has been published with this fix.

PR #5009, opened 2017-06-11
cleanup `index.js` and drop `0.12` node

bekzod:


package.json
```diff
@@ -111,7 +111,7 @@
     "ember-inflector": "^2.0.0"
   },
   "engines": {
-    "node": ">= 0.12.0"
```
rwjblue: FWIW, in practice we have already done this bump (we include deps that are only Node 4+ compatible since at least 2.12).

bekzod: thanks for clarifying, sorry I wrote bad commit message, we still should keep `engines` section right ?

rwjblue: Yep, the changes are good. I was just commenting to make it clear to the other maintainers that Node 4 was essentially already required...

bmac:
Thanks @bekzod.

PR #5011, opened 2017-06-16
Fix and tests for unloadRecord => findRecord issue

runspired:
The underlying issue was that the internalModel was not cleaning up it's `destroy` call and flags when being rematerialized immediately.

### DIFF
--- a/addon/-private/system/model/internal-model.js
+++ b/addon/-private/system/model/internal-model.js
@@ -24,6 +24,7 @@ const {
   isEmpty,
   isEqual,
   setOwner,
+  run,
   RSVP,
   RSVP: { Promise }
 } = Ember;
@@ -136,6 +137,7 @@ export default class InternalModel {
     // `objectAt(len - 1)` to test whether or not `firstObject` or `lastObject`
     // have changed.
     this._isDematerializing = false;
+    this._scheduledDestroy = null;
 
     this.resetRecord();
 
@@ -490,11 +492,22 @@ export default class InternalModel {
   unloadRecord() {
     this.send('unloadRecord');
     this.dematerializeRecord();
-    Ember.run.schedule('destroy', this, '_checkForOrphanedInternalModels');
+    if (this._scheduledDestroy === null) {
+      this._scheduledDestroy = run.schedule('destroy', this, '_checkForOrphanedInternalModels');
+    }
+  }
+
+  cancelDestroy() {
+    assert(`You cannot cancel the destruction of an InternalModel once it has already been destroyed`, !this.isDestroyed);
+
+    this._isDematerializing = false;
+    run.cancel(this._scheduledDestroy);
+    this._scheduledDestroy = null;
   }
 
   _checkForOrphanedInternalModels() {
     this._isDematerializing = false;
+    this._scheduledDestroy = null;
     if (this.isDestroyed) { return; }
 
     this._cleanupOrphanedInternalModels();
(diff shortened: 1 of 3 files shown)


tests/integration/records/unload-test.js
```diff
@@ -419,3 +419,48 @@ test('unloading a disconnected subgraph clears the relevant internal models', fu
...
+
+  const internalModel = record._internalModel;
+  assert.equal(internalModel.currentState.stateName, 'root.loaded.saved', 'We are loaded initially');
+
+  // we test that we can sync call unloadRecord followed by findRecord
+  run(function() {
+    store.unloadRecord(record);
+    assert.equal(record.isDestroying, true, 'the record is destroying');
+    assert.equal(internalModel.currentState.stateName, 'root.empty', 'We are unloaded after unloadRecord');
+    store.findRecord('person', '1');
```
stefanpenner: this should likely cancel the above `unloadRecords`'s `_checkForOrphanedInternalModels` and unset `_isDematerializing`, so it can be back in a pristine place, ready for a findRecord

runspired: 👍

tests/integration/records/unload-test.js
```diff
@@ -419,3 +419,198 @@ test('unloading a disconnected subgraph clears the relevant internal models', fu
...
+  assert.equal(internalModel.currentState.stateName, 'root.loaded.saved', 'We are loaded initially');
+
+  run(function() {
+    let timer = null;
+    store.unloadRecord(record);
+    timer = internalModel._scheduledDestroy;
+    assert.notEqual(timer, null, 'We scheduled destroy');
+    store.unloadRecord(record);
+    let  newTimer = internalModel._scheduledDestroy;
+    assert.equal(newTimer, timer, 'We scheduled destroy only once');
```
stefanpenner: is there a way for us to test the observable behavior of this rather then the implementation? Specifically, i suspect this will cause issues \w resurrection, and it should be possible to test that directly.

tests/integration/records/unload-test.js
```diff
@@ -419,3 +419,198 @@ test('unloading a disconnected subgraph clears the relevant internal models', fu
...
+      }
+    });
+  });
+
+  const internalModel = record._internalModel;
+  assert.equal(internalModel.currentState.stateName, 'root.loaded.saved', 'We are loaded initially');
+
+  run(function() {
+    store.unloadRecord(record);
+    let timer = internalModel._scheduledDestroy;
```
stefanpenner: we should aim to test observable behavior. This seems to tied to implementation details.

runspired: How about two calls to `scheduleDestroy` followed by a `cancelDestroy` and testing that `isDestroyed` is false?

stefanpenner: yup, something like that sounds good.

addon/-private/system/model/internal-model.js
```diff
@@ -490,11 +492,24 @@ export default class InternalModel {
...
+    }
+  }
+
+  cancelDestroy() {
+    assert(`You cannot cancel the destruction of an InternalModel once it has already been destroyed`, !this.isDestroyed);
+
+    if (this._isDematerializing === true) {
+      this._isDematerializing = false;
+      run.cancel(this._scheduledDestroy);
+      this._scheduledDestroy = null;
```
stefanpenner: Any reason to not put this outside the above `if` statement? I suspect always nulling this one here, will prevent us from accidentally not nulling it out sometimes.

runspired: Good catch. The intent was to avoid unnecessary work if we weren't in a state that needed to be cleaned up, but this isn't the correct check for that. Moved to just always resetting all three things.

stefanpenner:
@bmac we likely need to bugfix release this guy.

PR #5022, opened 2017-06-19
Ensure `ember-data/-private` module is emitted properly for ember-cli < 2.12

rwjblue:
ember-cli@2.12 and higher have the addon itself responsible for transpiling modules so that the output of the `treeForAddon` hook is expected to be AMD.

Prior versions of ember-cli (i.e. < 2.12) do second pass (that is not possible to disable) to transpile the output of `treeForAddon` **again** (to transpile modules).

This commit changes two things:

* It leverages `ember-cli-babel@6`'s built in ability to detect the various ember-cli version scenarios and either transpile `ember-data/-private` down to AMD or not correctly
* It emits the output file from `broccoli-rollup` into `modules/ember-data/-private.js` (previously was `modules/-private.js`). In ember-cli < 2.12, we have to ensure that the emitted file is in the correct location for that second pass module transpilation.


rwjblue:
This will need to be ported forward to `beta` and `master` (it targets `release` branch now). I made the changes directly against 2.14.1, and master has diverged somewhat significantly (in a good way) so my patch didn't apply.

index.js
```diff
@@ -130,21 +135,25 @@ module.exports = {
...
+    var publicTree = babel.transpileTree(this.debugTree(withoutPrivate, 'babel-public:input'));
+
+    publicTree = this.debugTree(publicTree, 'babel-public:output');
 
     privateTree = new Rollup(privateTree, {
       rollup: {
         entry: '-private/index.js',
         targets: [
-          { dest: '-private.js', format: 'amd', moduleId: 'ember-data/-private' }
+          { dest: 'ember-data/-private.js', format: babel.shouldCompileModules() ? 'amd' : 'es', moduleId: 'ember-data/-private' }
```
runspired: should `dest` be subject to the same conditional as `format`?

index.js
```diff
@@ -8,6 +8,7 @@ var Rollup = require('broccoli-rollup');
 var Babel = require('broccoli-babel-transpiler');
 var merge   = require('broccoli-merge-trees');
 var version = require('./lib/version');
+var BroccoliDebug = require('broccoli-debug');
```
runspired: this seems forgotten and likely to fail as it's not a dep

index.js
rwjblue: In 2.12 and higher we simply concat regardless of source path, so this is fine.  In < 2.12, we need to be in the right place so that the transpiled AMD module has the right `moduleId`.

If you notice down below, the public modules are emitted into `modules/ember-data` already (in all cases). This should be fine for both sides (and the tests are passing).

runspired: Thanks :)

Issue #4351, opened 2016-04-22
_changedKey function : possibility to deep compare object / array


olivierchatry:
Hello,

I'm not sure if it is relevant or not, but pushing object values will always being notified as "changed" by _changedKeys because the function only compare the reference value. It would be nice to have a hook ( maybe in the "transforms" object ?  that would allow to provide a isEqual function. 

What do you think ?

olivierchatry:
Sorry, just traced the code and found out that you can provide a "isEqual" function to the object you compare and it will use it. I would still prefer to have that in "transforms" object as it would make more sense, but I'm sure I can work that around.

olivierchatry:
Actually, scrap that, there is a problem after ember get the changed key, the values are anyway always assigned: 
system/model/internal-model.js

```.js
    setupData: function setupData(data) {
      var changedKeys = this._changedKeys(data.attributes);
      assign(this._data, data.attributes);
      this.pushedData();
      if (this.record) {
        this.record._notifyProperties(changedKeys);
      }
      this.didInitalizeData();
    },
```

I think ember should just assign the values that have changed.

wecc:
Would you be able to reproduce the issue or summarize what the actual vs. expected result is?

stefanpenner:
This seems to have gone stale, if @wecc question is answered we can likely reopen.

Note, there are also plans to make the difffing/change-tracking pluggable in the future

PR #5026, opened 2017-06-21
Add `deepCompare` property option to prevent triggering observers

Bajena:
This was first spotted by @olivierchatry here: https://github.com/emberjs/data/issues/4351. This issue was a pretty big problem in our project, so I spent some time to come up with a workaround. 
Currently we are monkey-patching the `internalModel` class in our project, but maybe it might be useful also for other people.

I decided to add it as a property option, because:
- I wanted this to be non-intrusive
- Sometimes deep comparison may be pretty heavy if the backend responds with big arrays/objects

Attributes in `_changedKeys` method are compared using the standard Ember.isEqual method,
however, if a property is an array/plain object, then the attribute
comparison will always return true and any observers added for this property
will fire. In order to prevent this `deepCompare` option can be used when
defining the property. For example:

 ```javascript
 DS.Model.extend({
  emails: DS.attr({ deepCompare: true })
});
 ```

P.S. This is my first PR here, so I hope for your understanding ;)


runspired:
I'm skeptical of the stringify approach to comparison being adequate; however I generally think this would be a good thing for us to do. I don't think it needs to be an option though, we should likely just move towards a simpler compare mechanism by default. I'll bring this up in a meeting soon to see if we have a collective opinion.

Bajena:
Thanks for answer @runspired let me know if I can help anyhow with this 👍

PR #5029, opened 2017-06-22
loosen header parsing slightly

stefanpenner:


stefanpenner:
tests will fail, due to: https://github.com/emberjs/data/pull/5030  I may merge if that is the only test that failed.

PR #5031, opened 2017-06-22
Ensure `ember-data/-private` module is emitted properly for ember-cli…

stefanpenner:
… < 2.12.

ember-cli@2.12 and higher have the addon itself responsible for transpiling
modules so that the output of the `treeForAddon` hook is expected to be
AMD.

Prior versions of ember-cli (i.e. < 2.12) do second pass (that is not possible
to disable) to transpile the output of `treeForAddon` **again** (to transpile
modules).

This commit changes things around leveraging ember-cli-babel@6's built in
ability to detect the various ember-cli version scenarios and either
transpile `ember-data/-private` down to AMD or not correctly.


index.js
```diff
@@ -52,7 +53,9 @@ module.exports = {
   init() {
     this._super.init && this._super.init.apply(this, arguments);
 
-    let bowerDeps = this.project.bowerDependencies();
+    this.debugTree = BroccoliDebug.buildDebugCallback('ember-data');
+
+    var bowerDeps = this.project.bowerDependencies();
```
rwjblue: 🔪 `var`

runspired:
🎉

PR #5038, opened 2017-06-26
remove deprecated `idToRecord` and cleanup

bekzod:


addon/-private/system/internal-model-map.js
```diff
@@ -112,21 +100,15 @@ export default class InternalModelMap {
    @method clear
    */
   clear() {
-    if (this._models) {
```
bekzod: this always true

addon/-private/system/internal-model-map.js
```diff
@@ -112,21 +100,15 @@ export default class InternalModelMap {
...
+
+    for (let i = 0; i < models.length; i++) {
+      let model = models[i];
+      model.unloadRecord();
     }
 
     this._metadata = null;
   }
 
-  destroy() {
```
bekzod: this is not used

Issue #4986, opened 2017-05-16
unloadRecord of child record causes hasMany relationship to be invalid using ember-data@2.13


sudowork:
## Example repo and integration test to reproduce the issue
- [ember-data 2.13.1 (broken)](https://github.com/sudowork/unloadrecord-ember-data-bug/tree/ember-data213)
- [ember-data 2.12.2 (working)](https://github.com/sudowork/unloadrecord-ember-data-bug/tree/ember-data212)

_Update_: I wrote a test case to test this behavior https://github.com/sudowork/data/commit/89a34ba7b59535bc679e8b6f7ea7108cc13fbf45. I bisected the issue to have started with 6d96eda226f15738b23032bc25c009e081070359 (from #4593).

_Update 2_: Converted the repo I had to an ember-twiddle now that I can load 2.13: [https://ember-twiddle.com/df7fa30d2eb3d32efe16185f4f376128?openFiles=controllers.todo-list.js%2C](ember-twiddle)

## Steps to reproduce:

```javascript
// models/todo-list.js
import DS from 'ember-data';
import { hasMany } from 'ember-data/relationships';

export default DS.Model.extend({
  todos: hasMany('todo', { async: false }),
});
```

1. Assume you have a model `todo-list` which `hasMany` `todo` items as `todos` already loaded in the store and is currently in scope.
1. Call [`todos.removeObject(todo)`](https://github.com/sudowork/unloadrecord-ember-data-bug/commit/03f7d0e7393927ce725469937bf35482b78ea550#diff-78fe8209f0a401e3c9acab9bacd673ccR29) to remove the item from the relationship
1. Call [`this.store.unloadRecord(todo)`](https://github.com/sudowork/unloadrecord-ember-data-bug/commit/03f7d0e7393927ce725469937bf35482b78ea550#diff-78fe8209f0a401e3c9acab9bacd673ccR31) to clean up the record from the store
1. Explicitly try accessing [`get(this, 'model.todos')`](https://github.com/sudowork/unloadrecord-ember-data-bug/commit/03f7d0e7393927ce725469937bf35482b78ea550#diff-78fe8209f0a401e3c9acab9bacd673ccR34) again to load the `todo` records. (Expected result: `[]`, Actual result: Assertion error thrown).

[...]

PR #5040, opened 2017-06-27
tests for unload + resurrect + remote relationship state

stefanpenner:


sly7-7:
@stefanpenner What do you think about adding a test of unloading the child record of the relationship. This would perhaps emphasize the bug found in https://github.com/emberjs/data/issues/4986 ?

stefanpenner:
@sly7-7 +1 (although I don't have cycles right now to do that, could you give it a whirl?

stefanpenner:
@sly7-7 ah, thanks for pointing that out (sorry so many github issues and PR to keep track of).

PR #5042, opened 2017-06-28
fix notifyPropertyChange error causing backtracking re-render error

### LINKED ISSUE #4942: hasMany association with links causes backtracking re-render error
Running against Canary I'm getting the following error when a has many relationship with links is used in a template:

```
Assertion Failed: You modified "meeting.attendees" twice on <minutebase@model:meeting::ember1351:772c5e6e-aceb-11e6-ae79-770161ca087d> in a single render. It was rendered in "component:mb-pages/dashboard/meeting" and modified in "component:mb-pages/dashboard/meeting". This was unreliable and slow in Ember 1.x and is no longer supported. See https://github.com/emberjs/ember.js/issues/13948 for more details.
Error
    at assert (http://minutebase.dev:4202/assets/vendor.js:21770:13)
    at assert (http://minutebase.dev:4202/assets/vendor.js:33521:34)
    at Object.assertNotRendered (http://minutebase.dev:4202/assets/vendor.js:39296:11)
    at Object.propertyDidChange (http://minutebase.dev:4202/assets/vendor.js:37912:30)
    at Class.propertyDidChange (http://minutebase.dev:4202/assets/vendor.js:51064:19)
    at Class.notifyPropertyChange (http://minutebase.dev:4202/assets/vendor.js:51078:12)
    at InternalModel.notifyPropertyChange (http://minutebase.dev:4202/assets/vendor.js:140423:21)
    at ManyRelationship.updateLink (http://minutebase.dev:4202/assets/vendor.js:147451:19)
    at ManyRelationship.push (http://minutebase.dev:4202/assets/vendor.js:147529:16)
    at Relationships.get (http://minutebase.dev:4202/assets/vendor.js:146885:24)
```

Looks like `updateLink` is notifying property change and causing Glimmer to re-evaluate the relationship and so it explodes.

I narrowed it down to when #4821 was merged, which makes sense as I assume that the links was previously resolved in advance but not happens when it's accessed.

Here's a [twiddle](https://ember-twiddle.com/e9ddeefbf25aea4fa2c27d9fe8e5fd01?openFiles=mirage.config.js%2C) which will hopefully demonstrate it, I can't get it working against ember-data canary as I get an error of "No application initializer named 'store'"

eriktrom:
for async relationships referenced in templates

- works with json api adapter (w/ links)
- works w/ django rest adapter
- works active-model-adapter

fixes #4942

cc @stefanpenner @runspired (i believe u 2 have the most context on this)

### DIFF
--- a/addon/-private/system/model/internal-model.js
+++ b/addon/-private/system/model/internal-model.js
@@ -724,25 +724,27 @@ export default class InternalModel {
 
   notifyHasManyAdded(key, record, idx) {
     if (this.hasRecord) {
-      this._record.notifyHasManyAdded(key, record, idx);
+      Ember.run.join(() => this._record.notifyHasManyAdded(key, record, idx));
     }
   }
 
   notifyHasManyRemoved(key, record, idx) {
     if (this.hasRecord) {
-      this._record.notifyHasManyRemoved(key, record, idx);
+      Ember.run.join(() => this._record.notifyHasManyRemoved(key, record, idx));
     }
   }
 
   notifyBelongsToChanged(key, record) {
     if (this.hasRecord) {
-      this._record.notifyBelongsToChanged(key, record);
+      Ember.run.join(() => this._record.notifyBelongsToChanged(key, record));
     }
   }
 
   notifyPropertyChange(key) {
     if (this.hasRecord) {
+      Ember.beginPropertyChanges();
       this._record.notifyPropertyChange(key);
+      Ember.endPropertyChanges();
     }
   }
 
--- a/addon/-private/system/model/model.js
+++ b/addon/-private/system/model/model.js
@@ -1040,7 +1040,9 @@ const Model = Ember.Object.extend(Ember.Evented, {
   },
 
   notifyBelongsToChanged(key) {
+    Ember.beginPropertyChanges();
     this.notifyPropertyChange(key);
+    Ember.endPropertyChanges();
   },
   /**
    Given a callback, iterates over each of the relationships in the model,
@@ -1112,7 +1114,9 @@ const Model = Ember.Object.extend(Ember.Evented, {
     //TODO(Igor): Consider whether we could do this only if the record state is unloaded
 
     //Goes away once hasMany is double promisified
+    Ember.beginPropertyChanges();
     this.notifyPropertyChange(key);
+    Ember.endPropertyChanges();
   },
 
   eachAttribute(callback, binding) {
--- a/addon/-private/system/store.js
+++ b/addon/-private/system/store.js
@@ -370,7 +370,9 @@ Store = Service.extend({
 
     // TODO @runspired this should also be coalesced into some form of internalModel.setState()
     internalModel.eachRelationship((key, descriptor) => {
-      internalModel._relationships.get(key).setHasData(true);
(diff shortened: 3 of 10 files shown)


eriktrom:
yo so these will pass but 

`findBelongsTo` and `findHasMany` *is NEVER called* in the tests from commit 7273f78

however, the re-render issues in a large app i upgraded last Friday, and which I don't fully understand by any means, are completely gone.

thus, i leave this as a helpful 'hi' i tried, maybe it'll help, maybe not, close me if not, all good

eriktrom:
k - one more note, that app does not use json api links - so most likely, this fixes async relationships in the rest adapter but there is something inherited in json api adapter that needs updating (or this is just totally wrong)

eriktrom:
um - i think i fixed it - that said, the logic was 7 months old - all i did was wrap the `notifyPropertyChange(key)` (original code) in `Ember.beginPropertyChanges()/Ember.endPropertyChanges()`

seems odd, works though 🤣

eriktrom:
@stefanpenner - related and seems important note - our app has issues when the dependent key inside a model is the hasMany relationships live record array - i believe that the attempt to reduce change events from these 2 commits is the root cause:

https://github.com/emberjs/data/commit/8c306d6ff581587238f1a88ae6a7724f47ec0068
https://github.com/emberjs/data/commit/7ed1c8a09d0e0737d0b425010bdba9a6715f7c9e

in this pr I **tried** hard to go with the flow of 'perf ED' - i started by reverting commits - then got this working (for re-renders at least) with no perf loss.. (i believe, big claim, hehe)

however, I'm worried about the side effects that apps like ours might run into for edge cases, especially when they go slightly outside the lines of 'ideal'. If upgrading becomes a risk..

i'll drop into dev-ember-data - say friday after lunch - can show u the codez... and have bandwidth to help move this forward - as example, one flow that does not use a dependent key on a hasMany live record array works, the other, boom. Also, save/destroyRecord/change route/save (w/ same id) goes boom too, which made me laugh out loud. unloadRecord also has a bug, un-related to 2.14, came in via 2.13... 

tl;dr - after a scare this morning, prior to showing the app to a potential customer, we're on 2.12 - so down to fix it w/ u

fyi, cheerio

eriktrom:
let me know if this is merge ready, i'll rebase to include only the first commit

runspired:
@eriktrom I believe we want to do some root-cause-analysis on this. Your tests look to be addressing something similar to what @stefanpenner addressed in #5048, could be related. The additional run-loops here only hide the issue, they won't solve it.

A couple of questions:

- Does this error only happen when a newly created record is involved in the hasMany?
- if it happens elsewhere (as the issues opened suggest `links` may be the issue) can you PR a failing test for the more general case to examine?

eriktrom:
> I believe we want to do some root-cause-analysis on this
> The additional run-loops here only hide the issue, they won't solve it.

yeah, that's pretty much what i figured... and why i left it as Ember.run.join, instead of the de-structured way, for example

> A couple of questions: ... 

Regarding further investigation - let me go back through, i'll write a failing test(s) and we can go from there. I'll keep working off master to ensure it's not already fixed. Thanks.

eriktrom:
closing in favor of a root cause analysis fix

PR #5045, opened 2017-06-30
[CLEANUP beta] Clean up some of the method signatures in RESTAdapter

nlfurniss:
Very small cleanup I noticed could be done after doing my investigation for `ember-fetch`. No fixes or features, so no new tests. Old tests passing locally.

### DIFF
--- a/addon/adapters/rest.js
+++ b/addon/adapters/rest.js
@@ -963,13 +963,13 @@ const RESTAdapter = Adapter.extend(BuildURLMixin, {
     @return {Object | DS.AdapterError} response
   */
   handleResponse(status, headers, payload, requestData) {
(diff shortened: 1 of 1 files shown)


stefanpenner:
@bmac looks ok to me, do you think this may be an intimate API others are relying on?

nlfurniss:
@hjdivad, @bmac any opinion on this?

hjdivad:
@stefanpenner only `normalizeErrorResponse` is even marked as `@private` in the jsdoc.

`isSuccess` and `isInvalid` seem to be public (whether intentional or not);  I don't think we can really drop the args for these

nlfurniss:
Ahh good point. Doesn't seem worth it in that case

PR #5048, opened 2017-07-02
[BUGFIX release] createRecord should only setup relationships it has …

stefanpenner:
…data for

relationships should in theory now only be accessed if they are used (`set` or `get`). This should mitigate the large number of test failures see during upgrade due to to https://github.com/emberjs/data/blob/e6cb564bdb33da4b3c15c3c668dc4b1fdafdf190/addon/-private/system/relationships/state/create.js#L20

If this doesn't actually fix the issue in enough cases, we can always turn that assert into a warn or something. But I would love a failing test for the exact case that remains (if one exists) to be 100% sure. As I would prefer to not loose a good assertion unless we must.

### DIFF
--- a/tests/integration/relationships/belongs-to-test.js
+++ b/tests/integration/relationships/belongs-to-test.js
@@ -955,26 +955,49 @@ test("belongsTo hasData sync not loaded", function(assert) {
   });
 });
 
-test("belongsTo hasData async created", function(assert) {
-  assert.expect(1);
+test("belongsTo hasData NOT created", function(assert) {
+  assert.expect(2);
 
   Book.reopen({
     author: belongsTo('author', { async: true })
   });
 
   run(() => {
+    let author = store.createRecord('author');
     let book = store.createRecord('book', { name: 'The Greatest Book' });
     let relationship = book._internalModel._relationships.get('author');
+
+    assert.equal(relationship.hasData, false, 'relationship does not have data');
+
+    book = store.createRecord('book', {
+      name: 'The Greatest Book',
+      author
+    });
+
+    relationship = book._internalModel._relationships.get('author');
+
     assert.equal(relationship.hasData, true, 'relationship has data');
   });
 });
 
 test("belongsTo hasData sync created", function(assert) {
-  assert.expect(1);
+  assert.expect(2);
 
   run(() => {
-    let book = store.createRecord('book', { name: 'The Greatest Book' });
+    let author = store.createRecord('author');
+    let book = store.createRecord('book', {
+      name: 'The Greatest Book'
+    });
+
     let relationship = book._internalModel._relationships.get('author');
+    assert.equal(relationship.hasData, false, 'relationship does not have data');
+
+    book = store.createRecord('book', {
+      name: 'The Greatest Book',
+      author
+    });
+
+    relationship = book._internalModel._relationships.get('author');
     assert.equal(relationship.hasData, true, 'relationship has data');
   });
 });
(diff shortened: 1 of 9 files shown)


tests/integration/relationships/belongs-to-test.js
```diff
@@ -970,26 +970,49 @@ test("belongsTo hasData sync not loaded", function(assert) {
...
 
-  run(function() {
-    var book = store.createRecord('book', { name: 'The Greatest Book' });
-    var relationship = book._internalModel._relationships.get('author');
+  run(() => {
+    let author = store.createRecord('author');
+    let book = store.createRecord('book', { name: 'The Greatest Book' });
+    let relationship = book._internalModel._relationships.get('author');
+
+    assert.equal(relationship.hasData, false, 'relationship has data');
```
hjdivad: assertion message is wrong: `relationship does not have data`

tests/integration/relationships/belongs-to-test.js
```diff
@@ -970,26 +970,49 @@ test("belongsTo hasData sync not loaded", function(assert) {
...
-    var book = store.createRecord('book', { name: 'The Greatest Book' });
-    var relationship = book._internalModel._relationships.get('author');
+  run(() => {
+    let author = store.createRecord('author');
+    let book = store.createRecord('book', {
+      name: 'The Greatest Book'
+    });
+
+    let relationship = book._internalModel._relationships.get('author');
+    assert.equal(relationship.hasData, false, 'relationship has data');
```
hjdivad: see ^

tests/integration/relationships/has-many-test.js
```diff
@@ -2547,25 +2547,44 @@ test("hasMany hasData sync not loaded", function(assert) {
...
 
-  run(function() {
-    var chapter = store.createRecord('chapter', { title: 'The Story Begins' });
-    var relationship = chapter._internalModel._relationships.get('pages');
+  run(() => {
+    let chapter = store.createRecord('chapter', { title: 'The Story Begins' });
+    let page = store.createRecord('page');
+
+    let relationship = chapter._internalModel._relationships.get('pages');
+    assert.equal(relationship.hasData, false, 'relationship has data');
```
hjdivad: see ^

tests/integration/relationships/has-many-test.js
```diff
@@ -2547,25 +2547,44 @@ test("hasMany hasData sync not loaded", function(assert) {
...
 
 test("hasMany hasData sync created", function(assert) {
-  assert.expect(1);
+  assert.expect(2);
+
+  run(() => {
+    let chapter = store.createRecord('chapter', { title: 'The Story Begins' });
+    let relationship = chapter._internalModel._relationships.get('pages');
+
+    assert.equal(relationship.hasData, false, 'relationship has data');
```
hjdivad: see ^

stefanpenner:
seems to be blocked by:

```js
SyntaxError: ember-data/adapters/json-api.js: @ember/debug does not have a deprecate import
   7 | import RESTAdapter from "./rest";
   8 | import { isEnabled } from '../-private';
>  9 | import { deprecate } from '@ember/debug';
     | ^
  10 | import { instrument } from 'ember-data/-debug';
```

@rwjblue is investigating.

stefanpenner:
@runspired r?

runspired:
LGTM

stefanpenner:
@bmac got cycles for a backport + release? (I'm currently on vacation, but if you don't have cycles I can handle it)

bmac:
@stefanpenner I'm on it.

bmac:
Released as 2.14.4.

stefanpenner:
@bmac awesome thanks dude, you rock!

thomasdurin:
Hi,

Since this changes unset relationships are not serialized, but I was wondering if this match the JSON-API specs ?

Doesn't JSON-API specs always require to specify all relationships, including those that are null ?

PR #4987, opened 2017-05-16
Add test for removing and unloading a record from a hasMany relationship

sudowork:
In relation to #4986.

Note: This test is intentionally failing right now. It is passing if run against a branch prior to 6d96eda226f15738b23032bc25c009e081070359.

PR #5052, opened 2017-07-03
add test for hasMany + addObject + unloadRecord

stefanpenner:
complementary to: https://github.com/emberjs/data/pull/4987

This test does legitimately fail

### DIFF
--- a/tests/integration/relationships/has-many-test.js
+++ b/tests/integration/relationships/has-many-test.js
@@ -134,6 +134,87 @@ test("When a hasMany relationship is accessed, the adapter's findMany method sho
   });
 });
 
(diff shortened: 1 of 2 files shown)


tests/integration/relationships/has-many-test.js
```diff
@@ -134,6 +134,87 @@ test("When a hasMany relationship is accessed, the adapter's findMany method sho
   });
 });
 
+test("hasMany + canonical vs currentState + unloadRecord", function(assert) {
+  assert.expect(6);
+
+  let postData = {
```
sly7-7: userData seems better ?

stefanpenner: 👍

stefanpenner:
this is now fixed

PR #5053, opened 2017-07-04
[BUGFIX] has-many relationship: set hasData when fetching link

### LINKED ISSUE #4845: HasManyReference property _isLoaded return false when it had been loaded once (with empty result)
I have an issue with HasManyReference property "_isLoaded". It should return true when relationship had been loaded at least one time, right? But it returns false, when last relationship request returned empty array. (works correct when last request returned more then 0 elements)

https://github.com/emberjs/data/blob/v2.11.0/addon/-private/system/references/has-many.js#L111

csantero:
I believe this fixes #4845. The problem is that `setHasData` was only being called when at least one internal model was being added, which is not the case when the server returns an empty array.

### DIFF
--- a/tests/integration/references/has-many-test.js
+++ b/tests/integration/references/has-many-test.js
@@ -662,6 +662,40 @@ test("load() fetches link when remoteType is link", function(assert) {
   });
 });
 
+test("load() fetches link when remoteType is link but an empty set of records is returned", function(assert) {
+  env.adapter.findHasMany = function(store, snapshot, link) {
+    assert.equal(link, "/families/1/persons");
+
+    return Ember.RSVP.resolve({ data: [] });
+  };
+
+  let family;
+  run(() => {
+    family = env.store.push({
+      data: {
+        type: 'family',
+        id: 1,
+        relationships: {
+          persons: {
+            links: { related: '/families/1/persons' }
+          }
+        }
+      }
+    });
+  });
+
+  let personsReference = family.hasMany('persons');
+  assert.equal(personsReference.remoteType(), "link");
+
+  return run(() => {
+    return personsReference.load().then((records) => {
+      assert.ok(records instanceof DS.ManyArray, "push resolves with the referenced records");
+      assert.equal(get(records, 'length'), 0);
+      assert.equal(get(personsReference.value(), 'length'), 0);
+    });
+  });
+});
+
 test("load() - only a single find is triggered", function(assert) {
   var done = assert.async();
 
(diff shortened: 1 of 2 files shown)


csantero:
Not sure what's wrong with Travis. `npm test` passes locally.

tests/integration/references/has-many-test.js
```diff
@@ -662,6 +662,44 @@ test("load() fetches link when remoteType is link", function(assert) {
...
+  var personsReference = family.hasMany('persons');
+  assert.equal(personsReference.remoteType(), "link");
+
+  run(function() {
+    personsReference.load().then(function(records) {
+      assert.ok(records instanceof DS.ManyArray, "push resolves with the referenced records");
+      assert.equal(get(records, 'length'), 0);
+      assert.equal(get(personsReference.value(), 'length'), 0);
+
+      done();
```
stefanpenner: lets not use `done`, instead use promises directly. That way when the promise rejects, the test fails.

tests/integration/references/has-many-test.js
```diff
@@ -662,6 +662,44 @@ test("load() fetches link when remoteType is link", function(assert) {
...
+test("load() fetches link when remoteType is link but an empty set of records is returned", function(assert) {
+  var done = assert.async();
+
+  env.adapter.findHasMany = function(store, snapshot, link) {
+    assert.equal(link, "/families/1/persons");
+
+    return Ember.RSVP.resolve({ data: [] });
+  };
+
+  var family;
```
stefanpenner: let

tests/integration/references/has-many-test.js
```diff
@@ -662,6 +662,44 @@ test("load() fetches link when remoteType is link", function(assert) {
...
+  run(function() {
```
stefanpenner: `=>`

tests/integration/references/has-many-test.js
```diff
@@ -662,6 +662,44 @@ test("load() fetches link when remoteType is link", function(assert) {
...
+          }
+        }
+      }
+    });
+  });
+
+  var personsReference = family.hasMany('persons');
+  assert.equal(personsReference.remoteType(), "link");
+
+  run(function() {
```
stefanpenner: `=>`

stefanpenner: `return` (so promises work)

tests/integration/references/has-many-test.js
```diff
@@ -662,6 +662,44 @@ test("load() fetches link when remoteType is link", function(assert) {
...
+    personsReference.load().then(function(records) {
```
stefanpenner: `return` (so promises work)

stefanpenner: `=>`

stefanpenner:
@csantero this looks reasonable

@igorT  / @runspired  / @hjdivad /  @bmac  r?

csantero:
@stefanpenner I've updated the test.

hjdivad:
Looks great, thanks for the contribution @csantero

Issue #5050, opened 2017-07-03
Accessing `include` query parameter with `coalesceFIndRequests = true`


buschtoens:
When `coalesceFindRequests = true` the `include` option is not passed through to the snapshot. This means, that it is impossible to access this option in the `findMany` method:

```js
export default JSONAPIAdapter.extend({
  coalesceFindRequests: true,

  findMany(store, type, ids, snapshots) {
    console.log(snapshots[0].include); // undefined
    return this._super(...arguments);
  }
});
```

This is because of these lines in [addon/-private/system/store.js](https://github.com/emberjs/data/blob/f360683474a9068f4731fb64fac7b5abfe6a1eaf/addon/-private/system/store.js#L929-L932)
```js
_flushPendingFetchForType(pendingFetchItems, modelName) {
  // ...
  let snapshots = new Array(totalItems);
  for (let i = 0; i < totalItems; i++) {
    snapshots[i] = internalModels[i].createSnapshot();
  }
  // ...
}
```

and [addon/-private/system/store/finders.js](https://github.com/emberjs/data/blob/f360683474a9068f4731fb64fac7b5abfe6a1eaf/addon/-private/system/store/finders.js#L52-L53)

```js
export function _findMany(adapter, store, modelName, ids, internalModels) {
  let snapshots = Ember.A(internalModels).invoke('createSnapshot');
  // ...
}
```

Two questions arise:

1) Why are the `pendingItem.options` not passed to `createSnapshot()` in `store.js`?
2) Why are new snapshots created in `_findMany` in `finders.js`?

I made some naive changes to both files and was able to pass the options to the `createSnapshot()` method in `store.js` and then use these snapshots in `finders.js`.

```js
let snapshots = new Array(totalItems);
for (let i = 0; i < totalItems; i++) {
  let internalModel = internalModels[i];
  let pendingItem = seeking[internalModel.id];
  snapshots[i] = internalModel.createSnapshot(pendingItem.options);
}
```

```js
export function _findMany(adapter, store, modelName, ids, internalModels, snapshots) {
  snapshots = snapshots || Ember.A(internalModels).invoke('createSnapshot');
  // ...
}
```

I don't know much about the inner workings of Ember Data and how this might interfere with other parts of the library, but these changes introduced no apparent bugs / errors in our existing relatively complicated app.

I assume that this is just an oversight, as passing `include` for coalesced find requests isn't *really exactly* defined behavior.

buschtoens:
[...]

PR #5054, opened 2017-07-04
store: pass original snapshots to adapter#findMany

buschtoens:
This test demos the bug described in #5050.

As suggested by @runspired on Slack, I also added a bugfix.

I only have a rudimentary idea about Ember Data internals, so please excuse me, if this is not the correct solution to the problem. I feel that it is kind of convoluted, but that stems from the fact that `store#_flushPendingFetchForType(pendingFetchItems, modelName)` converts to and from internal models and snapshots multiple times, as is stated in [a code comment](https://github.com/emberjs/data/blob/e1699cbb488445ad578b02a053e7f621bbbb3f85/addon/-private/system/store.js#L919-L928):

> TODO: Improve records => snapshots => records => snapshots
>
> We want to provide records to all store methods and snapshots to all
> adapter methods. To make sure we're doing that we're providing an array
> of snapshots to adapter.groupRecordsForFindMany(), which in turn will
> return grouped snapshots instead of grouped records.
>
> But since the _findMany() finder is a store method we need to get the
> records from the grouped snapshots even though the _findMany() finder
> will once again convert the records to snapshots for adapter.findMany()

### DIFF
--- a/addon/-private/system/store/finders.js
+++ b/addon/-private/system/store/finders.js
@@ -52,13 +52,8 @@ export function _find(adapter, store, modelClass, id, internalModel, options) {
   }, `DS: Extract payload of '${modelName}'`);
 }
 
-export function _findMany(adapter, store, modelName, ids, internalModels, snapshots) {
+export function _findMany(adapter, store, modelName, ids, snapshots) {
   if (DEBUG) { incrementRequestCount(); }
-  if (snapshots) {
-    snapshots = A(snapshots);
-  } else {
-    snapshots = A(internalModels).invoke('createSnapshot');
-  }
   let modelClass = store.modelFor(modelName); // `adapter.findMany` gets the modelClass still
   let promise = adapter.findMany(store, modelClass, ids, snapshots);
   let label = `DS: Handle Adapter#findMany of '${modelName}'`;
--- a/addon/-private/system/store/finders.js
+++ b/addon/-private/system/store/finders.js
@@ -1,4 +1,3 @@
-import { A } from '@ember/array';
 import { Promise } from 'rsvp';
 import { assert, warn } from '@ember/debug';
 import { DEBUG } from '@glimmer/env';
@@ -55,7 +54,7 @@ export function _find(adapter, store, modelClass, id, internalModel, options) {
 export function _findMany(adapter, store, modelName, ids, snapshots) {
   if (DEBUG) { incrementRequestCount(); }
   let modelClass = store.modelFor(modelName); // `adapter.findMany` gets the modelClass still
-  let promise = adapter.findMany(store, modelClass, ids, A(snapshots));
+  let promise = adapter.findMany(store, modelClass, ids, snapshots);
   let label = `DS: Handle Adapter#findMany of '${modelName}'`;
 
   if (promise === undefined) {
--- a/addon/-private/system/store/finders.js
+++ b/addon/-private/system/store/finders.js
@@ -1,4 +1,3 @@
-import { A } from '@ember/array';
 import { Promise } from 'rsvp';
 import { assert, warn } from '@ember/debug';
 import { DEBUG } from '@glimmer/env';
--- a/addon/-private/system/store/finders.js
+++ b/addon/-private/system/store/finders.js
@@ -52,9 +52,13 @@ export function _find(adapter, store, modelClass, id, internalModel, options) {
   }, `DS: Extract payload of '${modelName}'`);
 }
 
-export function _findMany(adapter, store, modelName, ids, internalModels) {
+export function _findMany(adapter, store, modelName, ids, internalModels, snapshots) {
   if (DEBUG) { incrementRequestCount(); }
-  let snapshots = A(internalModels).invoke('createSnapshot');
+  if (snapshots) {
+    snapshots = A(snapshots);
+  } else {
+    snapshots = A(internalModels).invoke('createSnapshot');
+  }
   let modelClass = store.modelFor(modelName); // `adapter.findMany` gets the modelClass still
   let promise = adapter.findMany(store, modelClass, ids, snapshots);
   let label = `DS: Handle Adapter#findMany of '${modelName}'`;
--- a/addon/-private/system/store/finders.js
+++ b/addon/-private/system/store/finders.js
@@ -1,3 +1,4 @@
+import { A } from '@ember/array';
 import { Promise } from 'rsvp';
 import { assert, warn } from '@ember/debug';
 import { DEBUG } from '@glimmer/env';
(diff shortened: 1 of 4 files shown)


runspired:
@buschtoens I'd love to revisit this

buschtoens:
@runspired Absolutely. This is still nagging me. How can I help?

@ryanto and @samselikoff might also be interested in this. I'm sure this could cause trouble with ember-data-storefront.

runspired:
@buschtoens let's start by rebasing. I apologize but I might be asking you to rebase a few times in the next few days, a lot of things are in-flight to master atm.

buschtoens:
@runspired that's cool with me. I'll try to get to it tomorrow.

tests/unit/store/adapter-interop-test.js
```diff
@@ -113,6 +113,33 @@ test('Calling Store#findRecord multiple times coalesces the calls into a adapter
...
+  const Type = DS.Model.extend();
+  let store = createStore({
+    adapter: Adapter,
+    test: Type
+  });
+
+  return run(() => {
+    return Ember.RSVP.all([
+      store.findRecord('test', 1, { include: 'someResource' }),
+      store.findRecord('test', 2, { include: 'someResource' })
```
buschtoens: Next step would be requesting different `include` strings and fixing [`groupRecordsForFindMany`](https://www.emberjs.com/api/ember-data/3.0/classes/DS.JSONAPIAdapter/methods/groupRecordsForFindMany?anchor=groupRecordsForFindMany) to take that into account, I guess?

buschtoens: Also, I should probably already request different `include`s here and assert, that these _different_ `include`s are correctly passed to `findMany`.

runspired: I'll give this a review this evening :)

samselikoff:
thanks @buschtoens , will try out our test suite on this

runspired:
@buschtoens I haven't forgotten you :)

buschtoens:
@samselikoff That's nice of you. Thank you :wink: 

@samselikoff I just noticed that my comment was phrased ambiguously. Just for the sake of clarity: I meant that the issue described in #5050 probably causes problems with ember-data-storefront, not this fix.

Did you try to run ember-data-storefront's test suite with `coalesceFindRequests` enabled?

```js
store.loadRecord('post', 1, { include: 'comments' });
store.loadRecord('post', 2, { include: 'comments' });
```

The snippet above likely won't work, since the `include` gets lost underway.

addon/-private/system/store/finders.js
```diff
@@ -52,9 +52,13 @@ export function _find(adapter, store, modelClass, id, internalModel, options) {
...
 }
 
-export function _findMany(adapter, store, modelName, ids, internalModels) {
+export function _findMany(adapter, store, modelName, ids, internalModels, snapshots) {
   if (DEBUG) { incrementRequestCount(); }
-  let snapshots = A(internalModels).invoke('createSnapshot');
+  if (snapshots) {
+    snapshots = A(snapshots);
+  } else {
+    snapshots = A(internalModels).invoke('createSnapshot');
```
runspired: in what scenario do we reach here without snapshots?

tests/unit/store/adapter-interop-test.js
runspired: I agree with your proposed test enhancements :)

I think we pass on the grouping for now though.

runspired:
@buschtoens any change to address the feedback?

addon/-private/system/store/finders.js
buschtoens: I think I've merely added this as a safeguard, so that add-ons relying on this intimate API don't break. 

I'll check if Ember Data itself accesses this from different places, tho. I can also remove this, if you like.

buschtoens:
@runspired sorry, I'll push an update in the next 24 hours. I was too caught up enjoying the rare sunny weather in Germany. 😅☀️🏖️🌈

addon/-private/system/store/finders.js
```diff
@@ -52,9 +52,8 @@ export function _find(adapter, store, modelClass, id, internalModel, options) {
   }, `DS: Extract payload of '${modelName}'`);
 }
 
-export function _findMany(adapter, store, modelName, ids, internalModels) {
+export function _findMany(adapter, store, modelName, ids, snapshots) {
   if (DEBUG) { incrementRequestCount(); }
-  let snapshots = A(internalModels).invoke('createSnapshot');
   let modelClass = store.modelFor(modelName); // `adapter.findMany` gets the modelClass still
   let promise = adapter.findMany(store, modelClass, ids, snapshots);
```
buschtoens: `snapshots` is a native array here, but was converted to an `EmberArray` in the now deleted [line 57](https://github.com/emberjs/data/pull/5054/files#diff-65c0e9ac4f030e1eff959291b8c5032aL57).

```js
let snapshots = A(internalModels).invoke('createSnapshot');
```

I should probably change this line from

```js
let promise = adapter.findMany(store, modelClass, ids, snapshots);
```

to

```js
let promise = adapter.findMany(store, modelClass, ids, A(snapshots));
```

tests/unit/store/adapter-interop-test.js
```diff
@@ -109,6 +109,65 @@ test('Calling Store#findRecord multiple times coalesces the calls into a adapter
...
+        data: snapshots.map(({ id }) => ({ id, type: type.modelName }))
+      });
+    },
+
+    findRecord(store, type, id, snapshot) {
+      assert.equal(
+        snapshot.include,
+        includedResourcesForIds[snapshot.id],
+        `Snapshot #${snapshot.id} retains the 'include' adapter option in #findRecord`
+      );
```
buschtoens: Depending on how the actual groupRecordsForFindMany method is implemented, it's possible that a single record of a coalesced group is retrieved via findRecord. In order to assert that findRecord works correctly as well, I've added a custom groupRecordsForFindMany method and added an assertion in here as well.

addon/-private/system/store/finders.js
buschtoens: Apparently I should have [run the whole test suite](https://travis-ci.org/emberjs/data/jobs/369972161#L572-L591). Haha.

Seems like the `build-url-mixin` relies on `snapshots` being an `EmberArray`.

buschtoens: Actually it was just the test, that relied on it. I've gone ahead and remove that `EmberArray` wrapper and fixed the test.

tests/integration/adapter/build-url-mixin-test.js
```diff
@@ -222,7 +222,7 @@ test('buildURL - buildURL takes the records from findMany', function(assert) {
 
   adapter.buildURL = function(type, ids, snapshots) {
     if (Array.isArray(snapshots)) {
-      return "/posts/" + snapshots.get('firstObject').belongsTo('post', { id: true }) + '/comments/';
+      return "/posts/" + snapshots[0].belongsTo('post', { id: true }) + '/comments/';
```
buschtoens: Context for this change: https://github.com/emberjs/data/pull/5054#pullrequestreview-114264010

runspired:
@sly7-7 this could be something for you to pickup.

sly7-7:
@runspired Ok, I can look at that. From what I understand, it basically needs a rebase, because I think your comments/feedback have been adressed ?

runspired:
> @runspired Ok, I can look at that. From what I understand, it basically needs a rebase, because I think your comments/feedback have been adressed ?

I believe so. The rebase is more of a "look at what was changed and re-implement" though because these files have substantially changed in location since this was last updated. Start there and lets see if tests are still passing.

sly7-7:
From what I see, I even think the underlying issue is already resolved (passing the adapterOptions map to _findMany), but I kept the test though. Tests are currently running locally :)

PR #5058, opened 2017-07-10
[BUGFIX] keep local state after a deletion

runspired:
If this approach is deemed good, this should be backported to beta and release.

This addresses an issue uncovered by the work to improve relationships: deleting a relationship results in a call to `flushCanonical` which erases any other local changes made. We can prevent local edits being removed for the common case (delete returning no additional data).

cc @igorT @hjdivad @stefanpenner for review

### DIFF
--- a/addon/-private/system/model/internal-model.js
+++ b/addon/-private/system/model/internal-model.js
@@ -715,7 +715,7 @@ export default class InternalModel {
     heimdall.increment(send);
     let currentState = this.currentState;
 
-    if (typeof currentState[name] !== 'function') {
+    if (!currentState[name]) {
       this._unhandledEvent(currentState, name, context);
     }
 
@@ -725,12 +725,6 @@ export default class InternalModel {
   notifyHasManyAdded(key, record, idx) {
     if (this.hasRecord) {
       this._record.notifyHasManyAdded(key, record, idx);
-    }
-  }
-
-  notifyHasManyRemoved(key, record, idx) {
-    if (this.hasRecord) {
-      this._record.notifyHasManyRemoved(key, record, idx);
     }
   }
 
@@ -768,7 +762,7 @@ export default class InternalModel {
     }
 
     if (this.isNew()) {
-      this.clearRelationships(true);
+      this.clearRelationships();
     }
 
     if (this.isValid()) {
@@ -885,28 +879,38 @@ export default class InternalModel {
     @method clearRelationships
     @private
   */
-  clearRelationships(resetAll = false) {
+  clearRelationships() {
     this.eachRelationship((name, relationship) => {
       if (this._relationships.has(name)) {
         let rel = this._relationships.get(name);
 
-        if (resetAll === true) {
-          rel.clear();
-          rel.removeInverseRelationships();
-        } else {
-          rel.unloadInverseInternalModel(this);
-        }
+        rel.clear();
+        rel.removeInverseRelationships();
       }
     });
     Object.keys(this._implicitRelationships).forEach((key) => {
       let rel = this._implicitRelationships[key];
 
-      if (resetAll === true) {
-        rel.clear();
-        rel.removeInverseRelationships();
-      } else {
-        rel.unloadInverseInternalModel(this);
+      rel.clear();
+      rel.removeInverseRelationships();
+    });
+  }
+
+  /*
+    @method unloadDeletedRecordFromRelationships
+   */
+  unloadDeletedRecordFromRelationships() {
+    this.eachRelationship((name) => {
+      if (this._relationships.has(name)) {
+        let rel = this._relationships.get(name);
+
+        rel.unloadDeletedInverseInternalModel(this);
       }
+    });
+    Object.keys(this._implicitRelationships).forEach((key) => {
+      let rel = this._implicitRelationships[key];
+
+      rel.unloadDeletedInverseInternalModel(this);
     });
   }
 
--- a/addon/-private/system/model/internal-model.js
+++ b/addon/-private/system/model/internal-model.js
@@ -715,7 +715,7 @@ export default class InternalModel {
     heimdall.increment(send);
     let currentState = this.currentState;
 
-    if (!currentState[name]) {
+    if (typeof currentState[name] !== 'function') {
       this._unhandledEvent(currentState, name, context);
     }
 
@@ -768,7 +768,7 @@ export default class InternalModel {
     }
 
     if (this.isNew()) {
-      this.clearRelationships();
+      this.clearRelationships(true);
     }
 
     if (this.isValid()) {
@@ -885,17 +885,28 @@ export default class InternalModel {
     @method clearRelationships
     @private
   */
-  clearRelationships() {
+  clearRelationships(resetAll = false) {
     this.eachRelationship((name, relationship) => {
       if (this._relationships.has(name)) {
         let rel = this._relationships.get(name);
-        rel.clear();
-        rel.removeInverseRelationships();
+
+        if (resetAll === true) {
+          rel.clear();
+          rel.removeInverseRelationships();
+        } else {
+          rel.unloadInverseInternalModel(this);
+        }
       }
     });
     Object.keys(this._implicitRelationships).forEach((key) => {
-      this._implicitRelationships[key].clear();
-      this._implicitRelationships[key].removeInverseRelationships();
+      let rel = this._implicitRelationships[key];
+
+      if (resetAll === true) {
+        rel.clear();
+        rel.removeInverseRelationships();
+      } else {
+        rel.unloadInverseInternalModel(this);
+      }
     });
   }
 
--- a/addon/-private/system/model/internal-model.js
+++ b/addon/-private/system/model/internal-model.js
@@ -762,7 +762,7 @@ export default class InternalModel {
     }
 
     if (this.isNew()) {
-      this._resetAllRelationshipsToEmpty();
+      this.removeFromInverseRelationships(true);
     }
 
     if (this.isValid()) {
@@ -876,50 +876,36 @@ export default class InternalModel {
   }
 
   /*
-    This method should only be called by records in the `isNew()` state.
-    It will completely reset all relationships to an empty state and remove
-    the record from any associated inverses as well.
+   This method should only be called by records in the `isNew()` state OR once the record
+   has been deleted and that deletion has been persisted.
 
-    @method _resetAllRelationshipsToEmpty
-    @private
-  */
-  _resetAllRelationshipsToEmpty() {
-    this.eachRelationship((name, relationship) => {
-      if (this._relationships.has(name)) {
-        let rel = this._relationships.get(name);
+   It will remove this record from any associated relationships.
 
-        rel.clear();
-        rel.removeInverseRelationships();
-      }
-    });
-    Object.keys(this._implicitRelationships).forEach((key) => {
-      let rel = this._implicitRelationships[key];
-
-      rel.clear();
-      rel.removeInverseRelationships();
-    });
-  }
-
-  /*
-    This method should only be called once the record has been deleted
-    and that deletion has been persisted. It will remove this record from
-    any associated relationships.
+   It will completely reset all relationships to an empty state and remove
+   the record from any associated inverses as well.
 
     @method removeFromInverseRelationships
+    @param {Boolean} isNew whether to unload from the `isNew` perspective
     @private
    */
-  removeFromInverseRelationships() {
+  removeFromInverseRelationships(isNew = false) {
     this.eachRelationship((name) => {
       if (this._relationships.has(name)) {
         let rel = this._relationships.get(name);
 
         rel.removeDeletedInternalModelFromInverse();
+        if (isNew === true) {
+          rel.clear();
+        }
       }
     });
     Object.keys(this._implicitRelationships).forEach((key) => {
       let rel = this._implicitRelationships[key];
 
       rel.removeDeletedInternalModelFromInverse();
+      if (isNew === true) {
+        rel.clear();
+      }
     });
   }
 
--- a/addon/-private/system/model/internal-model.js
+++ b/addon/-private/system/model/internal-model.js
@@ -905,10 +905,10 @@ export default class InternalModel {
     and that deletion has been persisted. It will remove this record from
     any associated relationships.
 
-    @method unloadDeletedRecordFromRelationships
+    @method removeFromInverseRelationships
     @private
    */
-  unloadDeletedRecordFromRelationships() {
+  removeFromInverseRelationships() {
     this.eachRelationship((name) => {
       if (this._relationships.has(name)) {
         let rel = this._relationships.get(name);
--- a/addon/-private/system/model/internal-model.js
+++ b/addon/-private/system/model/internal-model.js
@@ -885,14 +885,14 @@ export default class InternalModel {
         let rel = this._relationships.get(name);
 
         rel.clear();
-        // rel.removeInverseRelationships();
+        rel.removeInverseRelationships();
       }
     });
     Object.keys(this._implicitRelationships).forEach((key) => {
       let rel = this._implicitRelationships[key];
 
       rel.clear();
-      // rel.removeInverseRelationships();
+      rel.removeInverseRelationships();
     });
   }
 
--- a/addon/-private/system/model/internal-model.js
+++ b/addon/-private/system/model/internal-model.js
@@ -881,8 +881,8 @@ export default class InternalModel {
 
    It will remove this record from any associated relationships.
 
-   If `isNew` is true (default false), it will completely reset all relationships
-   to an empty state as well.
+   If `isNew` is true (default false), it will also completely reset all
+    relationships to an empty state as well.
 
     @method removeFromInverseRelationships
     @param {Boolean} isNew whether to unload from the `isNew` perspective
@@ -893,7 +893,7 @@ export default class InternalModel {
       if (this._relationships.has(name)) {
         let rel = this._relationships.get(name);
 
-        rel.removeDeletedInternalModelFromInverse();
+        rel.removeCompletelyFromInverse();
         if (isNew === true) {
           rel.clear();
         }
@@ -902,13 +902,17 @@ export default class InternalModel {
     Object.keys(this._implicitRelationships).forEach((key) => {
       let rel = this._implicitRelationships[key];
 
-      rel.removeDeletedInternalModelFromInverse();
+      rel.removeCompletelyFromInverse();
       if (isNew === true) {
         rel.clear();
       }
     });
   }
 
+  /*
+    Notify all inverses that this internalModel has been dematerialized
+    and destroys any ManyArrays.
+   */
   destroyRelationships() {
     this.eachRelationship((name, relationship) => {
       if (this._relationships.has(name)) {
@@ -992,6 +996,8 @@ export default class InternalModel {
   }
 
   /*
+    Used to notify the store to update FilteredRecordArray membership.
+
     @method updateRecordArrays
     @private
   */
--- a/addon/-private/system/model/internal-model.js
+++ b/addon/-private/system/model/internal-model.js
@@ -913,13 +913,13 @@ export default class InternalModel {
       if (this._relationships.has(name)) {
         let rel = this._relationships.get(name);
 
-        rel.removeDeletedInternalModelFromInverse(this);
+        rel.removeDeletedInternalModelFromInverse();
       }
     });
     Object.keys(this._implicitRelationships).forEach((key) => {
       let rel = this._implicitRelationships[key];
 
-      rel.removeDeletedInternalModelFromInverse(this);
+      rel.removeDeletedInternalModelFromInverse();
     });
   }
 
--- a/addon/-private/system/model/internal-model.js
+++ b/addon/-private/system/model/internal-model.js
@@ -881,8 +881,8 @@ export default class InternalModel {
 
    It will remove this record from any associated relationships.
 
-   It will completely reset all relationships to an empty state and remove
-   the record from any associated inverses as well.
+   If `isNew` is true (default false), it will completely reset all relationships
+   to an empty state as well.
 
     @method removeFromInverseRelationships
     @param {Boolean} isNew whether to unload from the `isNew` perspective
--- a/addon/-private/system/model/internal-model.js
+++ b/addon/-private/system/model/internal-model.js
@@ -762,7 +762,7 @@ export default class InternalModel {
     }
 
     if (this.isNew()) {
-      this.clearRelationships();
+      this._resetAllRelationshipsToEmpty();
     }
 
     if (this.isValid()) {
@@ -876,10 +876,14 @@ export default class InternalModel {
   }
 
   /*
-    @method clearRelationships
+    This method should only be called by records in the `isNew()` state.
+    It will completely reset all relationships to an empty state and remove
+    the record from any associated inverses as well.
+
+    @method _resetAllRelationshipsToEmpty
     @private
   */
-  clearRelationships() {
+  _resetAllRelationshipsToEmpty() {
     this.eachRelationship((name, relationship) => {
       if (this._relationships.has(name)) {
         let rel = this._relationships.get(name);
@@ -897,20 +901,25 @@ export default class InternalModel {
   }
 
   /*
+    This method should only be called once the record has been deleted
+    and that deletion has been persisted. It will remove this record from
+    any associated relationships.
+
     @method unloadDeletedRecordFromRelationships
+    @private
    */
   unloadDeletedRecordFromRelationships() {
     this.eachRelationship((name) => {
       if (this._relationships.has(name)) {
         let rel = this._relationships.get(name);
 
-        rel.unloadDeletedInverseInternalModel(this);
+        rel.removeDeletedInternalModelFromInverse(this);
       }
     });
     Object.keys(this._implicitRelationships).forEach((key) => {
       let rel = this._implicitRelationships[key];
 
-      rel.unloadDeletedInverseInternalModel(this);
+      rel.removeDeletedInternalModelFromInverse(this);
     });
   }
 
--- a/addon/-private/system/model/internal-model.js
+++ b/addon/-private/system/model/internal-model.js
@@ -885,14 +885,14 @@ export default class InternalModel {
         let rel = this._relationships.get(name);
 
         rel.clear();
-        rel.removeInverseRelationships();
+        // rel.removeInverseRelationships();
       }
     });
     Object.keys(this._implicitRelationships).forEach((key) => {
       let rel = this._implicitRelationships[key];
 
       rel.clear();
-      rel.removeInverseRelationships();
+      // rel.removeInverseRelationships();
     });
   }
 
--- a/addon/-private/system/model/model.js
+++ b/addon/-private/system/model/model.js
@@ -1118,8 +1118,8 @@ const Model = Ember.Object.extend(Ember.Evented, {
   notifyHasManyRemoved(key) {
     // We need to notifyPropertyChange in the removing case to flush ui state
     //  Goes away if we remove/improve the outer PromiseProxy on async relationships
-    // this.notifyPropertyChange(key);
-    this.notifyPropertyChange(`${key}.length`);
+    //  Notifying `${key}.length` does not work here
+    this.get(key).notifyPropertyChange('length');
   },
 
   eachAttribute(callback, binding) {
--- a/addon/-private/system/model/model.js
+++ b/addon/-private/system/model/model.js
@@ -1118,7 +1118,8 @@ const Model = Ember.Object.extend(Ember.Evented, {
   notifyHasManyRemoved(key) {
     // We need to notifyPropertyChange in the removing case to flush ui state
     //  Goes away if we remove/improve the outer PromiseProxy on async relationships
-    //  Notifying `${key}.length` does not work here
+    //  Notifying `${key}` or `${key}.length` does not work here, notifying `content`
+    //  after `get(key)` also works but this wont always be an async relationship.
     this.get(key).notifyPropertyChange('length');
   },
 
--- a/addon/-private/system/model/model.js
+++ b/addon/-private/system/model/model.js
@@ -1115,6 +1115,13 @@ const Model = Ember.Object.extend(Ember.Evented, {
     this.notifyPropertyChange(key);
   },
 
+  notifyHasManyRemoved(key) {
+    // We need to notifyPropertyChange in the removing case to flush ui state
+    //  Goes away if we remove/improve the outer PromiseProxy on async relationships
+    // this.notifyPropertyChange(key);
+    this.notifyPropertyChange(`${key}.length`);
+  },
+
   eachAttribute(callback, binding) {
     this.constructor.eachAttribute(callback, binding);
   }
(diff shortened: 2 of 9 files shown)


addon/-private/system/model/internal-model.js
```diff
@@ -885,17 +885,28 @@ export default class InternalModel {
     @method clearRelationships
     @private
   */
-  clearRelationships() {
+  clearRelationships(resetAll = false) {
```
stefanpenner: I would prefer two methods then a single method filled with if statements

tests/unit/model/relationships/has-many-test.js
```diff
@@ -906,6 +906,569 @@ test('it is possible to add a new item to a relationship', function(assert) {
...
+  assert.equal(get(pets, 'length'), 1, 'precond - relationship has only one pet to start');
+
+  run(() => {
+    pets.pushObjects([rambo, rebel]);
+  });
+
+  assert.equal(get(pets, 'length'), 3, 'precond2 - relationship now has three pets');
+
+  run(() => {
+    shen.destroyRecord({})
```
stefanpenner: Return

tests/unit/model/relationships/has-many-test.js
```diff
@@ -906,6 +906,569 @@ test('it is possible to add a new item to a relationship', function(assert) {
...
+        }
+      ]
+    });
+  });
+
+  return run(() => {
+    const person = store.peekRecord('person', '1');
+    const petsProxy = person.get('pets');
+
+    petsProxy.then((pets) => {
```
stefanpenner: Return

stefanpenner:
I'll take some time tonight or tomorrow to review. 

I'm not sure the current semantics for unload purging inverse makes sense in the first place, as it seems at odds with resurrection...

I'll compare your tests to another approach I was playing with (one that I think corrects the above resurrection vs unload issue) and see what I missed.

addon/-private/system/model/internal-model.js
runspired: I agree, will refactor it tomorrow

runspired:
So the making here could be better but the idea is that the delete has already occurred and been committed and ergo the method uses the term "unload" because the relationship itself will be in the correct state afterwards, no additional signaling needed. 

Overall it turns out that clear is called well before unloadRecord (is called during the delete), and our issue with the clearing is/was not due to the dematerialization change itself but other relationship bug fixes.

We need not fear resurrection here because these methods are called to remove a persisted deletion from the relationship.

addon/-private/system/relationships/state/has-many.js
```diff
@@ -111,6 +111,32 @@ export default class ManyRelationship extends Relationship {
...
+
+    if (manyArray) {
+      const currentState = manyArray.currentState;
+      const idx = currentState.indexOf(internalModel);
+
+      if (idx !== -1) {
+        currentState.splice(idx, 1);
+
+        manyArray.set('length', currentState.length);
+        manyArray.notifyPropertyChange('length');
```
stefanpenner: why is the above `set` not sufficient to notify?

addon/-private/system/relationships/state/has-many.js
```diff
@@ -111,6 +111,32 @@ export default class ManyRelationship extends Relationship {
...
+
+        this.internalModel.notifyHasManyRemoved(this.key, internalModel, idx);
```
stefanpenner: if the proxy is just wrapping the hasMany, why wouldn't the above `set`propagate?

addon/-private/system/model/internal-model.js
```diff
@@ -885,17 +885,28 @@ export default class InternalModel {
...
     this.eachRelationship((name, relationship) => {
       if (this._relationships.has(name)) {
         let rel = this._relationships.get(name);
-        rel.clear();
-        rel.removeInverseRelationships();
+
+        if (resetAll === true) {
+          rel.clear();
+          rel.removeInverseRelationships();
```
stefanpenner: curious, do we actually want to remove the entire inverse here, or just remove our own internalModel form the inverse? If its just remove ourselves, then the `clear` flag feels less bad and the above recommendation to two methods can be disregarded.

Specifically:

```js
if (resetAll === true) {
  rel.clear();
}

rel.unloadInverseInternalModel(this);
```

Also, would the following name change improve clarity?

`rel.unloadInverseInternalModel(this);` -> `rel.unloadInternalModelFromInverse(this);`

--

This would seem to tidy up how we deal with inverses, spuriously destroying them must lead to other issues...

addon/-private/system/model/internal-model.js
```diff
@@ -885,17 +885,28 @@ export default class InternalModel {
...
       }
     });
     Object.keys(this._implicitRelationships).forEach((key) => {
-      this._implicitRelationships[key].clear();
-      this._implicitRelationships[key].removeInverseRelationships();
+      let rel = this._implicitRelationships[key];
+
+      if (resetAll === true) {
+        rel.clear();
+        rel.removeInverseRelationships();
```
stefanpenner: ^ same

addon/-private/system/model/internal-model.js
```diff
@@ -715,7 +715,7 @@ export default class InternalModel {
     heimdall.increment(send);
     let currentState = this.currentState;
 
-    if (!currentState[name]) {
+    if (typeof currentState[name] !== 'function') {
```
runspired: It can be, just something I noticed as I was stepping through.

addon/-private/system/model/internal-model.js
runspired: We still end up with the name spaghetti with `unloadInternalModelFromInverseFromOwn` and `unloadInternalModelFromInverseFromInverses`

addon/-private/system/relationships/state/has-many.js
runspired: for `async` relationships an issue was surfaced in which `set` was not enough to notify the `PromiseProxy` that `length` had changed. This particular notify was a shot in the dark at fixing it that did not work and should be removed. See comment in `model.js` for more on the issue surfaced, you may have an idea as to what isn't wired correctly.

addon/-private/system/relationships/state/has-many.js
runspired: unknown, see comments left in `model.js`

addon/-private/system/model/internal-model.js
stefanpenner: I believe always doing `unloadInverseInternalModel(this)` will nearly fix https://github.com/emberjs/data/pull/5052 the remaining assertion failures will be based on what we decide resurrection + unload + retained by inverse relationship actually means. The code-base is currently slightly confused... specifically:

the following assertion https://github.com/emberjs/data/pull/5052/files#diff-f5099af3b5cad3fc343f3f4b73eb3c97R208  will be true if resurrection does apply in this case (reviving 2 + 6) OR resurrection does not apply and the contacts list will become  ` ['3','4','5','7']`

I think @hjdivad will need to chime in.

addon/-private/system/model/internal-model.js
stefanpenner: we may not need to split the methods, depending on the outcome of our discussion -> https://github.com/emberjs/data/pull/5058#discussion_r126597331

addon/-private/system/relationships/state/has-many.js
stefanpenner: removing this `notifyPropertyChange` appears to have no affect on the tests. it seems safe to remove.

addon/-private/system/relationships/state/has-many.js
stefanpenner: https://github.com/emberjs/ember.js/pull/15487 may fix, and if so the following would polyfill it.

```js
let maybeLengthDescriptor = Ember.ArrayProxy.proto().length;

if (typeof maybeLengthDescriptor === 'object' &&
  maybeLengthDescriptor !== null &&
  maybeLengthDescriptor.isDescriptor === true &&
  maybeLengthDescriptor._dependentKeys === undefined) {
  maybeLengthDescriptor._dependentKeys = ['arrangedContent.length'];
}
```

stefanpenner: scratch that. 

The issue is, ManyArray emulates Enumerable, and thus must provide it's own `arrayContentWillChange` and `arrayContentDidChange` events. It's internal methods do this correctly, but your manual and external `currentState.splice` + `manyArray.set('length` does not. Rather then reaching into the internal state, you may want to use `manyArray.internalReplace` or add your own `manyArray.splice.

```js
splice(idx, count) {
    this.arrayContentWillChange(idx, count, 0);
    this.currentState.splice(idx, count);

    this.set('length', this.currentState.length);
    this.arrayContentDidChange(idx, count, 0);
  }
```

addon/-private/system/model/model.js
```diff
@@ -1115,6 +1115,14 @@ const Model = Ember.Object.extend(Ember.Evented, {
     this.notifyPropertyChange(key);
   },
 
+  notifyHasManyRemoved(key) {
```
stefanpenner: once https://github.com/emberjs/data/pull/5058/files#r126607364 is addressed, this goes away.

addon/-private/system/relationships/state/has-many.js
stefanpenner: if https://github.com/emberjs/data/pull/5058/files#r126607364 is addressed, most of this goes away.

addon/-private/system/relationships/state/has-many.js
```diff
@@ -111,6 +111,32 @@ export default class ManyRelationship extends Relationship {
...
+    const canonicalIndex = this.canonicalState.indexOf(internalModel);
+
+    if (canonicalIndex !== -1) {
+      this.canonicalState.splice(canonicalIndex, 1);
+    }
+
+    const manyArray = this._manyArray;
+
+    if (manyArray) {
+      const currentState = manyArray.currentState;
```
stefanpenner: this block can likely be replaced with: `manyArray.removeInternalModels([internalModel])`

addon/-private/system/relationships/state/has-many.js
runspired: doh, that makes some sense of why this seemed so odd.

I'm still a bit confused why this (1) didn't seem to be an issue unless there was the extra proxy from the `async` wrapper involved and (2) the Enumerable docs suggest just `set` is fine (https://www.emberjs.com/api/classes/Ember.Enumerable.html) and (3) just for learning why is `this.arrayContentWillChange` absolutely necessary to trigger change events here?

stefanpenner: > the Enumerable docs suggest just ...

Enumerable docs appear in-complete `arrayContent{Will,Did}Change` stuff seems to be missing.

> I'm still a bit confused why this (1) didn't seem to be an issue unless there was the extra proxy from the async wrapper

As soon as there is a ArrayProxy (in this case: PromiseManyArray) `length` had the chance of getting out of sync.

addon/-private/system/relationships/state/relationship.js
```diff
@@ -266,6 +266,31 @@ export default class Relationship {
...
+    this.notifyRecordRelationshipRemoved(internalModel);
+  }
+
+  unloadInverseInternalModelFromInverses(internalModel) {
+    if (!this.inverseKey) { return; }
+
+    let allMembers =
+      // we actually want a union of members and canonicalMembers
+      // they should be disjoint but currently are not due to a bug
+      this.members.toArray().concat(this.canonicalMembers.toArray());
```
stefanpenner: rather then making 3 arrays only to iterate them, maybe we should do something like:

```js
const unload = inverseInternalModel => {
   let relationship = inverseInternalModel._relationships.get(this.inverseKey);
   relationship.unloadInverseInternalModelFromOwn(internalModel);
}

this.members.forEach(unload);
this.canonicalMembers(unload);
```

addon/-private/system/model/internal-model.js
```diff
@@ -885,14 +885,14 @@ export default class InternalModel {
         let rel = this._relationships.get(name);
 
         rel.clear();
-        // rel.removeInverseRelationships();
+        rel.removeInverseRelationships();
```
runspired: @hjdivad commenting these out didn't make any tests fail o.O

addon/-private/system/model/internal-model.js
```diff
@@ -889,13 +883,34 @@ export default class InternalModel {
...
+
+      rel.clear();
+      rel.removeInverseRelationships();
+    });
+  }
+
+  /*
+    @method unloadDeletedRecordFromRelationships
+   */
+  unloadDeletedRecordFromRelationships() {
```
hjdivad: per our chat out of band, let's rename this to not conflate this case with `unloadRecord`

hjdivad: Also let's add some comments here as breadcrumbs to our future selves who will clean this up.

This does fix an important bug, but currently there are two codepaths whose different semantics are not obvious within the same level of abstraction, ie that `rel.clear()` causes a `flushCanonical` but `unloadDeletedInverseInternalModel ` does not is not clear, and that this is the entire point is also not clear.

addon/-private/system/relationships/state/relationship.js
```diff
@@ -266,6 +269,33 @@ export default class Relationship {
     this.flushCanonicalLater();
   }
 
+  unloadDeletedInverseInternalModel(internalModel) {
```
hjdivad: what's the case where `internalModel != this.internalModel`?

hjdivad: Per our discussion out of band, let's fix the inconsistency in naming here.  `unloadDeletedInverseInternalModel` and `unloadDeletedInverseInternalModelFromOwn` both mention an "inverse internal model" but are referring to different sides of the relationship.

tests/unit/model/relationships/has-many-test.js
```diff
@@ -906,6 +906,561 @@ test('it is possible to add a new item to a relationship', function(assert) {
...
+
+  const person = store.peekRecord('person', '1');
+  const pets = run(() => person.get('pets'));
+
+  const shen = pets.objectAt(0);
+  const rambo = store.peekRecord('pet', '2');
+  const rebel = store.peekRecord('pet', '3');
+
+  assert.equal(get(shen, 'name'), 'Shenanigans', 'precond - relationships work');
+  assert.equal(get(pets, 'length'), 1, 'precond - relationship has only one pet to start');
```
hjdivad: let's change these assertions to `deepEqual` an array of ids rather than just checking the length

tests/unit/model/relationships/has-many-test.js
```diff
@@ -906,6 +906,561 @@ test('it is possible to add a new item to a relationship', function(assert) {
...
+
+  run(() => {
+    pets.pushObjects([rambo, rebel]);
+  });
+
+  assert.equal(get(pets, 'length'), 3, 'precond2 - relationship now has three pets');
```
hjdivad: let's change these assertions to `deepEqual` an array of ids rather than just checking the length

tests/unit/model/relationships/has-many-test.js
```diff
@@ -906,6 +906,561 @@ test('it is possible to add a new item to a relationship', function(assert) {
...
+
+  run(() => {
+    return shen.destroyRecord({})
+      .then(() => {
+        shen.unloadRecord();
+      });
+  });
+
+  assert.equal(get(pets, 'length'), 2, 'relationship now has two pets');
```
hjdivad: let's change these assertions to `deepEqual` an array of ids rather than just checking the length

tests/unit/model/relationships/has-many-test.js
```diff
@@ -906,6 +906,561 @@ test('it is possible to add a new item to a relationship', function(assert) {
...
+    const person = store.peekRecord('person', '1');
+    const petsProxy = run(() => person.get('pets'));
+
+    return petsProxy.then((pets) => {
+      const shen = pets.objectAt(0);
+      const rambo = store.peekRecord('pet', '2');
+      const rebel = store.peekRecord('pet', '3');
+
+      assert.equal(get(shen, 'name'), 'Shenanigans', 'precond - relationships work');
+      assert.equal(get(pets, 'length'), 1, 'precond - relationship has only one pet to start');
```
hjdivad: let's change these assertions to `deepEqual` an array of ids rather than just checking the length

tests/unit/model/relationships/has-many-test.js
```diff
@@ -906,6 +906,561 @@ test('it is possible to add a new item to a relationship', function(assert) {
...
+      assert.equal(get(petsProxy, 'length'), 1, 'precond - proxy has only one pet to start');
+
+      pets.pushObjects([rambo, rebel]);
+
+      assert.equal(get(pets, 'length'), 3, 'precond2 - relationship now has three pets');
```
hjdivad: let's change these assertions to `deepEqual` an array of ids rather than just checking the length

tests/unit/model/relationships/has-many-test.js
```diff
@@ -906,6 +906,561 @@ test('it is possible to add a new item to a relationship', function(assert) {
...
+      assert.equal(get(petsProxy, 'length'), 3, 'precond2 - proxy now reflects three pets');
+
+      return shen.destroyRecord({})
+        .then(() => {
+          shen.unloadRecord();
+
+          assert.equal(get(pets, 'length'), 2, 'relationship now has two pets');
+          assert.equal(get(petsProxy, 'length'), 2, 'proxy now reflects two pets');
```
hjdivad: let's change these assertions to `deepEqual` an array of ids rather than just checking the length

addon/-private/system/relationships/state/belongs-to.js
```diff
@@ -63,6 +63,25 @@ export default class BelongsToRelationship extends Relationship {
     this.notifyBelongsToChanged();
   }
 
+  unloadDeletedInverseInternalModelFromOwn(internalModel) {
```
hjdivad:
```js
foo.set(belongsTo, new);
new.destroyRecord();

foo.get(belongsTo) === null;
``

We should not restore canonical when deleting new; we should let belongsto be null.  That's consistent with [current behaviour](https://ember-twiddle.com/7f94a0e81ea38d9022a5007ef8be15c8?openFiles=adapters.application.js%2C) and in any case, restoring here is a bit wat

File: tests/unit/model/relationships/has-many-test.js [hunk 22]
```diff
@@ -906,6 +906,561 @@ test('it is possible to add a new item to a relationship', function(assert) {
...
+          shen.unloadRecord();
+
+          dog = person.get('dog.content');
+          assert.ok(dog === rambo, 'The currentState of the belongsTo was preserved after the delete');
+        });
+    });
+  });
+});
+
+test('deleting an item that is the current state of a belongsTo restores canonicalState', function(assert) {
```
hjdivad: why?

1. this is a change in semantics (see [this twiddle](https://ember-twiddle.com/7f94a0e81ea38d9022a5007ef8be15c8?openFiles=adapters.application.js%2C))
2. this is surprising behaviour

tests/unit/model/relationships/has-many-test.js
```diff
@@ -906,6 +906,561 @@ test('it is possible to add a new item to a relationship', function(assert) {
...
+/*
+  This test, when passing, affirms that a known limitation of ember-data still exists.
+
+  When pushing new data into the store, ember-data is currently incapable of knowing whether
+  a relationship has been persisted. In order to update relationship state effectively, ember-data
+  blindly "flushes canonical" state, removing any `currentState` changes. A delete that sideloads
+  the parent record's hasMany is a situation in which this limitation will be encountered should other
+  local changes to the relationship still exist.
+ */
+test('returning new hasMany relationship info from a delete supplements local state', function(assert) {
```
hjdivad: s/supplements/supersedes

Supplements is exactly what it does not do ^_^

tests/unit/model/relationships/has-many-test.js
```diff
@@ -906,6 +906,561 @@ test('it is possible to add a new item to a relationship', function(assert) {
...
+
+  const person = store.peekRecord('person', '1');
+  const pets = run(() => person.get('pets'));
+
+  const shen = store.peekRecord('pet', '1');
+  const rambo = store.peekRecord('pet', '2');
+  const rebel = store.peekRecord('pet', '3');
+
+  assert.equal(get(shen, 'name'), 'Shenanigans', 'precond - relationships work');
+  assert.equal(get(pets, 'length'), 2, 'precond - relationship has only one pet to start');
```
hjdivad: as before, let's `deepEqual` ids instead of just checking length in this test

addon/-private/system/relationships/state/relationship.js
```diff
@@ -266,6 +269,41 @@ export default class Relationship {
...
 
+  /*
+    Call this method once a record deletion has been persisted
+    to purge it from both current and canonical state of all
+    relationships.
+
+    @method removeDeletedInternalModelFromInverse
+    @private
+   */
+  removeDeletedInternalModelFromInverse(internalModel) {
```
hjdivad: what's the case where `internalModel != this.internalModel`?

tests/unit/model/relationships/has-many-test.js
runspired: wrote the test for what we want to happen, and it should fail once what we want does happen, should fix that test description to account for this

tests/unit/model/relationships/has-many-test.js
runspired: ok

addon/-private/system/model/internal-model.js
hjdivad: hmm yeah it looks like the unload tests assert against the id map and the relationship payloads, but not against relationships (either our side or inverse).

addon/-private/system/relationships/state/relationship.js
runspired: there isn't one afaik, just seemed better to pass this in with the original implementation. using `this.internalModel` is probably a bit safer

addon/-private/system/model/internal-model.js
```diff
@@ -882,20 +876,50 @@ export default class InternalModel {
...
+
+  /*
+    This method should only be called once the record has been deleted
+    and that deletion has been persisted. It will remove this record from
+    any associated relationships.
+
+    @method unloadDeletedRecordFromRelationships
+    @private
+   */
+  unloadDeletedRecordFromRelationships() {
```
igorT: We use unload as the term for unloading records, maybe we should call this method something else, like remove?

addon/-private/system/relationships/state/belongs-to.js
```diff
@@ -63,6 +63,19 @@ export default class BelongsToRelationship extends Relationship {
     this.notifyBelongsToChanged();
   }
 
+  removeDeletedInternalModelFromOwn(internalModel) {
+    super.removeDeletedInternalModelFromOwn(internalModel);
```
igorT: I thought we moved away from calling super in relationships?

addon/-private/system/model/internal-model.js
hjdivad: Absolutely.

@runspired I guess we missed an instance of the renaming

addon/-private/system/model/internal-model.js
```diff
@@ -882,20 +876,50 @@ export default class InternalModel {
...
-    @method clearRelationships
+    This method should only be called by records in the `isNew()` state.
+    It will completely reset all relationships to an empty state and remove
+    the record from any associated inverses as well.
+
+    @method _resetAllRelationshipsToEmpty
     @private
   */
-  clearRelationships() {
+  _resetAllRelationshipsToEmpty() {
```
igorT: What is the difference in behavior between the `isNew` and the deletion case?

addon/-private/system/relationships/state/belongs-to.js
runspired: We never landed that

addon/-private/system/model/internal-model.js
runspired: confirm, will rename

addon/-private/system/relationships/state/relationship.js
```diff
@@ -266,6 +269,42 @@ export default class Relationship {
...
 
+  /*
+    Call this method once a record deletion has been persisted
+    to purge it from both current and canonical state of all
+    relationships.
+
+    @method removeDeletedInternalModelFromInverse
+    @private
+   */
+  removeDeletedInternalModelFromInverse() {
```
stefanpenner: I believe `Deleted` should be removed from this record name. In general a method name should describe `what` it does, not why or when it should be invoked.

addon/-private/system/relationships/state/relationship.js
```diff
@@ -266,6 +269,42 @@ export default class Relationship {
...
+        relationship.removeDeletedInternalModelFromOwn(internalModel);
+        seen[id] = true;
+      }
+    };
+
+    this.members.forEach(unload);
+    this.canonicalMembers.forEach(unload);
+  }
+
+  removeDeletedInternalModelFromOwn(internalModel) {
```
stefanpenner: I believe Deleted should be removed from this record name. In general a method name should describe what it does, not why or when it should be invoked.

addon/-private/system/relationships/state/relationship.js
```diff
@@ -266,6 +269,42 @@ export default class Relationship {
...
+  removeDeletedInternalModelFromInverse() {
+    if (!this.inverseKey) { return; }
+
+    // we actually want a union of members and canonicalMembers
+    // they should be disjoint but currently are not due to a bug
+    let seen = Object.create(null);
+    const internalModel = this.internalModel;
+
+    const unload = inverseInternalModel => {
+      const id = guidFor(inverseInternalModel);
```
stefanpenner: As this is internal `inverseInternalModel.id` should be sufficient.

addon/-private/system/relationships/state/relationship.js
```diff
@@ -266,6 +269,42 @@ export default class Relationship {
...
+    };
+
+    this.members.forEach(unload);
+    this.canonicalMembers.forEach(unload);
+  }
+
+  removeDeletedInternalModelFromOwn(internalModel) {
+    this.canonicalMembers.delete(internalModel);
+    this.members.delete(internalModel);
+    this.notifyRecordRelationshipRemoved(internalModel);
```
stefanpenner: does this method do anything? I only see two implementations and they appear to be noops. 

https://github.com/emberjs/data/search?utf8=%E2%9C%93&q=notifyRecordRelationshipRemoved&

addon/-private/system/model/internal-model.js
stefanpenner: How about: `removeFromInverseRelationships`, example:

```js
internalModel.removeFromInverseRelationships();
```

The method would now describes concisely *what* is does not *when* to use it. The *when* and *why* should be provided via docs, or derived contextually from where it is used.

addon/-private/system/model/states.js
```diff
@@ -671,7 +671,7 @@ const RootState = {
       isDirty: false,
 
       setup(internalModel) {
-        internalModel.clearRelationships();
+        internalModel.unloadDeletedRecordFromRelationships();
```
stefanpenner:
```js
internalModel.removeFromInverseRelationships()
```

addon/-private/system/model/internal-model.js
```diff
@@ -728,12 +728,6 @@ export default class InternalModel {
     }
   }
 
-  notifyHasManyRemoved(key, record, idx) {
```
stefanpenner: 👍  this was dead code right?

addon/-private/system/model/internal-model.js
```diff
@@ -882,20 +876,50 @@ export default class InternalModel {
...
+
+    @method unloadDeletedRecordFromRelationships
+    @private
+   */
+  unloadDeletedRecordFromRelationships() {
+    this.eachRelationship((name) => {
+      if (this._relationships.has(name)) {
+        let rel = this._relationships.get(name);
+
+        rel.removeDeletedInternalModelFromInverse();
```
stefanpenner: should `destroyRelationships` also get this treatment?

addon/-private/system/model/internal-model.js
runspired: 👍

addon/-private/system/relationships/state/relationship.js
runspired: @stefanpenner what if there is no `id` yet?

addon/-private/system/relationships/state/relationship.js
runspired: Will follow guidance on your other comment for renaming

addon/-private/system/relationships/state/relationship.js
runspired: I believe this was put here for something with dematerialization that was never completed maybe?

addon/-private/system/model/internal-model.js
runspired: yes

addon/-private/system/relationships/state/belongs-to.js
igorT: :(

addon/-private/system/relationships/state/has-many.js
```diff
@@ -111,6 +111,26 @@ export default class ManyRelationship extends Relationship {
     super.removeCanonicalInternalModelFromOwn(internalModel, idx);
   }
 
+  removeDeletedInternalModelFromOwn(internalModel) {
```
hjdivad: seems related to `removeInternalModelFromOwn`, why diverge?

hjdivad:
Summary of issues

1. lose local changes in hasMany on a delete of a canonical record and the server returns no new canonical data.
2. lose local changes in hasMany more generally during flushCanonical as flushCanonical does not diff
3. unload clears inverse relationships eagerly, preventing intra-loop rematerialization
4. unload perf issue possibly overwalking during gc

---

This pr addresses 1.

cc @runspired @igorT @stefanpenner

addon/-private/system/relationships/state/relationship.js
stefanpenner: ah, thats right. Seems good

addon/-private/system/relationships/state/relationship.js
stefanpenner: @runspired this function is quite similar to `removeInternalModelFromOwn` what is the divergence?

stefanpenner: The divergence (but feel free to confirm):

* `removeInternalModelFromOwn`:  is removing from current state (awaiting future canonical flush)
* `removeDeletedInternalModelFromOwn`: is removing from both current state (`members`) and canonical state (`canonicalMembers`)

The naming of these methods sorta blows, but naming is hard. So at the very least we may need to document them somewhat.

Some ideas:

```js
Removes a given internal model from the relationships currentState, but not from that relationships canonicalState. This results in an optimistic delete, and subsequent confirmation (typically server side) will update the canonicalState.

@method removeInternalModelFromOwn
@param internalModel
```


```js
Removes a given internal model from both a relationships currentState and canonical state. This typically occurs when an internalRecord's deletion is persisted.

@method removeDeletedInternalModelFromOwn
@param internalModel
```

addon/-private/system/relationships/state/has-many.js
stefanpenner: i believe the divergence is: https://github.com/emberjs/data/pull/5058/files#r127379185

runspired: `removeInternalModelFromOwn` removes only current state while `removeDeletedInternalModelFromOwn` removes from current and canonical state.

PR #5060, opened 2017-07-12
Set hasData for preloaded empty hasMany

hjdivad:
See #5053 for a similar fix in the fetchLink case.  See #4845 for the initially reported bug.


hjdivad:
@stefanpenner @bmac trivial follow-up to #5053

PR #5066, opened 2017-07-14
Revert "[BUGFIX] keep local state after a deletion"

stefanpenner:
Reverts emberjs/data#5058

going to squash this


stefanpenner:
cc @runspired

PR #5070, opened 2017-07-15
[DOC release] Update doc for deprecated function.

MichalBryxi:
- store#filter() is now [deprecated](https://github.com/emberjs/data/blob/master/addon/-private/system/store.js#L1785)
 - Don't suggest it's usage


stefanpenner:
https://github.com/emberjs/data/pull/5071

PR #5074, opened 2017-07-18
disable node-tests in windows for now

stefanpenner:


stefanpenner:
I'm sick of seeing the windows tests fail... I'll bring it back after the ember-cli issue is addressed and released..

BryanCrotaz:
There appear to be no tests at all on Windows
```
npm start
chrome.exe http://localhost:4200/tests
```
result: No tests run.

Is this issue the reason why? The description appears to imply that only node tests wouldn't be run.

stefanpenner:
@BryanCrotaz hmm that is quite strange...

PR #5077, opened 2017-07-21
[BUGFIX release] fix destroyRecord followed by unloadRecord and impli…

stefanpenner:
…cit inverse relationships

- [x] confirm it works

### DIFF
--- a/addon/-private/system/model/internal-model.js
+++ b/addon/-private/system/model/internal-model.js
@@ -899,8 +899,12 @@ export default class InternalModel {
         }
       }
     });
-    Object.keys(this._implicitRelationships).forEach((key) => {
-      let rel = this._implicitRelationships[key];
+
+    let implicitRelationships = this._implicitRelationships;
+    this.__implicitRelationships = null;
+
+    Object.keys(implicitRelationships).forEach((key) => {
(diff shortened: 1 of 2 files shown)


addon/-private/system/model/internal-model.js
```diff
@@ -899,8 +899,12 @@ export default class InternalModel {
         }
       }
     });
-    Object.keys(this._implicitRelationships).forEach((key) => {
-      let rel = this._implicitRelationships[key];
+
+    let implicitRelationships = this._implicitRelationships;
+    this.__implicitRelationships = null;
```
bmac: This feels a little insider baseball to me but I don't have a better suggestion.

stefanpenner: @bmac well, we are in the same class were it is private... Ya It's not amazing. Other ideas end up being junky too. I may circle back after this gets backported, and try to tidy this pattern up.

hjdivad: i think the basic approach is fine and i suspect it seems a bit iffy mainly because these data structures already have a lot of complicated state.  I suspect landing the simplify relationships work will make this kind of resetting not look so bad

PR #5079, opened 2017-07-21
[BACKPORT 2 release] Release fix unload after destroy

stefanpenner:
…cit inverse relationships


stefanpenner:
Ah this failure is the same as the last backport. We switched default serializers in tests, so my deleteRecord response is considered malformed. Will fix shortly

Issue #5080, opened 2017-07-21
Error: Cannot find module 'babel-plugin-transform-es2015-block-scoping'


Turbo87:
with `npm@2` which is the default in Node 4 which is recommended to be used by Ember addons at the moment:

```
Error: Cannot find module 'babel-plugin-transform-es2015-block-scoping'
    at Function.Module._resolveFilename (module.js:325:15)
    at Function.Module._load (module.js:276:25)
    at Module.require (module.js:353:17)
    at require (internal/module.js:12:17)
    at Object.<anonymous> (/home/travis/build/simplabs/ember-cli-pixijs/node_modules/ember-data/lib/stripped-build-plugins.js:7:29)
```

see https://github.com/emberjs/data/blob/v2.14.5/lib/stripped-build-plugins.js#L7

tl;dr if you want to `require()` something then it should be a dependency of your project

This is a duplicate of https://github.com/emberjs/data/issues/5024 since I can't reopen the original issue and this problem is apparently not resolved even though the original issue was closed for some reason.

Turbo87:
Resolved by https://github.com/emberjs/data/pull/5081

Issue #5024, opened 2017-06-21
Cannot find module 'babel-plugin-transform-es2015-block-scoping'


devdemi:
Hi!

Since `ember-data` version was upgraded to `v2.14.2` when I execute `ember server` I get net error message:

> Cannot find module 'babel-plugin-transform-es2015-block-scoping'
> Error: Cannot find module 'babel-plugin-transform-es2015-block-scoping'
>     at Function.Module._resolveFilename (module.js:325:15)
>     at Function.Module._load (module.js:276:25)
>     at Module.require (module.js:353:17)
>     at require (internal/module.js:12:17)
>     at Object.<anonymous> (/usr/src/fe/node_modules/ember-data/lib/stripped-build-plugins.js:7:29)
>     at Module._compile (module.js:409:26)
>     at Object.Module._extensions..js (module.js:416:10)
>     at Module.load (module.js:343:32)
>     at Function.Module._load (module.js:300:12)
>     at Module.require (module.js:353:17)
>     at require (internal/module.js:12:17)
> 
Really I can't find in `node_modules/ember-data/node_modules/` `babel-plugin-transform-es2015-block-scoping` module.
I found out in `ember-data`'s package.json next dependecy `"babel-plugin-transform-es2015-block-scoping": "^6.24.1"` but early it was "^6.23.0". A lot of another ember's modules has dependency `"babel-plugin-transform-es2015-block-scoping": "^6.23.0"`
When add `"babel-plugin-transform-es2015-block-scoping": "^6.24.1"` to my project's `package.json` then it works well.

I have no ideas how to resolve this problem. Please advice what I need to do.

---

Output from `ember version --verbose && npm --version && yarn --version`:
```
ember-cli: 2.8.0
http_parser: 2.7.0
node: 4.5.0
v8: 4.5.103.37
uv: 1.9.1
zlib: 1.2.8
ares: 1.10.1-DEV
icu: 56.1
modules: 46
openssl: 1.0.2h
os: linux x64
2.15.9

```

dgavey:
I was able to fix this on ember-drag-drop by upgrading the version of node used by Travis from ancient to 6.

Turbo87:
Resolved by https://github.com/emberjs/data/pull/5081

PR #5081, opened 2017-07-21
Convert "babel-plugin-transform-es2015-block-scoping" from dev to prod dependency

Turbo87:
Resolves https://github.com/emberjs/data/issues/5080 and https://github.com/emberjs/data/issues/5024

/cc @stefanpenner @bmac @runspired


bmac:
Thanks @Turbo87.

Turbo87:
@bmac would be great if we could get a patch release somewhat soonish :)

bmac:
Looks like @stefanpenner just released 2.14.6.

PR #5088, opened 2017-07-24
{MASTER} [BUGFIX release] handle dupe relationship entries

stefanpenner:
If a relationship was setup with duplicate entries, it would enter an
invalid state. Specifically, this.canonicalMembers and
this.canonicalState would be out of sync. Resulting in some sad things.

This was most likely introduced by https://github.com/emberjs/data/commit/f8304b23c792a1e74cefb63140e6ee1917cbdfa1#commitcomment-23256408

### DIFF
--- a/addon/-private/system/relationships/state/has-many.js
+++ b/addon/-private/system/relationships/state/has-many.js
@@ -201,17 +201,25 @@ export default class ManyRelationship extends Relationship {
   }
 
   setInitialInternalModels(internalModels) {
-    if (!internalModels) {
+    if (Array.isArray(internalModels) === false || internalModels.length === 0) {
       return;
     }
 
-    let args = [0, this.canonicalState.length].concat(internalModels);
-    this.canonicalState.splice.apply(this.canonicalState, args);
-    internalModels.forEach(internalModel => {
+    let forCanonical = [];
+
+    for (let i = 0; i< internalModels.length; i++) {
+      let internalModel = internalModels[i];
+      if (this.canonicalMembers.has(internalModel)) {
+        continue;
+      }
+
+      forCanonical.push(internalModel);
       this.canonicalMembers.add(internalModel);
       this.members.add(internalModel);
       this.setupInverseRelationship(internalModel);
-    });
+    }
+
+    this.canonicalState.splice(0, this.canonicalState.length, ...forCanonical);
   }
 
   fetchLink() {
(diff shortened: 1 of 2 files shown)


addon/-private/system/relationships/state/has-many.js
```diff
@@ -201,17 +201,25 @@ export default class ManyRelationship extends Relationship {
   }
 
   setInitialInternalModels(internalModels) {
-    if (!internalModels) {
+    if (Array.isArray(internalModels) === false || internalModels.length === 0) {
```
hjdivad: Seems okay but this feel pretty deep internally for checking input validity; IIRC we already have a dev assertion to this effect at the store API.

Can we move this kind of validation up to public APIs?

tests/unit/model/relationships/has-many-test.js
```diff
@@ -399,6 +399,72 @@ test('hasMany with explicit initial null works even when the inverse was set to
   });
 });
 
+test('hasMany with duplicates from payload', function(assert) {
+  assert.expect(1);
+
+  const Tag = DS.Model.extend({
+    name: DS.attr('string'),
+    people: DS.hasMany('person', { async: false })
+  });
```
hjdivad: +`toString`?

tests/unit/model/relationships/has-many-test.js
```diff
@@ -399,6 +399,72 @@ test('hasMany with explicit initial null works even when the inverse was set to
...
+
+  const Person = DS.Model.extend({
+    name: DS.attr('string'),
+    tag: DS.belongsTo('tag', { async: false })
+  });
```
hjdivad: +`toString`?

addon/-private/system/relationships/state/has-many.js
stefanpenner: not in a backport

stefanpenner: Also, the above was more of a refinement on the existing code. But yes, a future cleanup PR sounds good.

tests/unit/model/relationships/has-many-test.js
```diff
@@ -399,6 +399,84 @@ test('hasMany with explicit initial null works even when the inverse was set to
...
 
+test('hasMany with duplicates from payload', function(assert) {
+  assert.expect(1);
+
+  const Tag = DS.Model.extend({
+    name: DS.attr('string'),
+    people: DS.hasMany('person', { async: false })
+  });
+
+  Tag.reopenClass({
```
hjdivad: +:heart:

Issue #5025, opened 2017-06-21
`FilteredRecordArray` has null values after `unloadRecord` w/ 2.13 and newer.


workmanw:
This is very similar to #4986, however it has a slightly different cause so I wanted to file it as a separate issue. 

### What's happening
I created specific twiddle to demonstrated:

* [ember-data-2.12.2 (working)](https://ember-twiddle.com/50f002a0e1a36ce249b7b3ccd4062eee?openFiles=controllers.todo-list.js%2C)
* [ember-data-2.13.2 (broken)](https://ember-twiddle.com/fb8383d82a1b92a8be61e618aa8862ab?openFiles=controllers.todo-list.js%2C)

I would direct your attention to the `removeTodo` action in the `todo-list` controller. What's happening is that when I call `unloadRecord` the filtered array replaces the value of the model that was just unloaded with `null`.

This `null` value exists only until the end of the runloop because there is a `_flush` function scheduled to run in the `action` queue to clean these up.

### Why is this happening?

With 2.13 there was a change made (#4593) that set's the internal model's record to `null` when it's reset or dematerialized (this is new functionality and important to understanding the issue). See: [internal-model.js#L360](https://github.com/emberjs/data/blob/ac0192184/addon/-private/system/model/internal-model.js#L360). 

Also, because the filtered array's content is a list of internal models, it has to look up the record via `objectAtContent` which produces a `null` values for unloaded records. See: [record-array.js#L107-L110](https://github.com/emberjs/data/blob/ac019218465456991119089cf32482f09f3db898/addon/-private/system/record-arrays/record-array.js#L107-L110). So that means when someone tries to use `objectAt` or `toArray` it'll get a list with a null value in place of the item that was just unloaded.

### What did it do before?

With 2.12 the record was not nullified from the internal model. So while the array would produce a stale value (i.e. an unloaded record) until the end of the runloop, it was more desirable than a null value. 

### Why is this a problem?

In my case, after the unload action happens and it schedules a `_flush` task in the `actions` queue, however it moves on to other tasks in the `render` queue. That's when views and computed properties access the array, hit the null value that they weren't expecting and throw an exception.

I could make my application more durable to assume that values in these filtered arrays could be null, but it feels like the right answer is they should never been null.

stefanpenner:
@workmanw if you have cycles, do you mind submitting a failing test?

workmanw:
[...]

PR #5095, opened 2017-07-25
Added failing test for issue #5025. 

workmanw:
Ping @stefanpenner -- As promised, a failing test. 

One thing to note is this assertion:

```javascript
    assert.equal(get(people, 'length'), 2, 'Unloaded record removed from the array');
```

This would technically have failed before the previous regression because instead of nullifying the record in the list, it just left the stale value until the end of the loop. If the goal is to restore that behavior, then we need to change the test criteria. 

And I would be happy to do that, let me know.

-------

For reference a test of the previous functionality would look like:

```javascript
run(() => {
  people.objectAt(0).unloadRecord();
  
  assert.equal(get(people, 'length'), 3, 'Unload does not complete until the end of the loop');
  assert.ok(get(people.objectAt(0), 'name'), 'Scumbag Dale', 'Dale is still the first object until the end of the loop');
});

assert.equal(get(people, 'length'), 2, 'Unloaded record removed from the array');
assert.ok(get(people.objectAt(0), 'name'), 'Scumbag Katz', 'Katz shifted down after the unload');
assert.ok(get(people.objectAt(1), 'name'), 'Scumbag Bryn', 'Bryn shifted down after the unload');
```

### DIFF
--- a/tests/integration/filter-test.js
+++ b/tests/integration/filter-test.js
@@ -837,3 +837,28 @@ test('destroying filteredRecordArray unregisters models from being filtered', fu
 
   assert.equal(filterFn.summary.called.length, 1, 'expected the filter function not being called anymore');
 });
+
+test('unloading records in the filter', function(assert) {
+  run(() => store.push({ data }));
+
+  let people = run(() => {
+    return store.filter('person', hash => {
+      if (hash.get('name').match(/Scumbag/)) {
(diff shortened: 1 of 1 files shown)


runspired:
I ported this test into #5378 and confirmed that the fix for #5350 also fixes this issue.

PR #5097, opened 2017-07-27
[BUGFIX release] restore unloadRecord + createRecord

stefanpenner:
The following should be possible:

```js
const record = store.find(‘record’, 1);
record.unloadRecord();
store.createRecord(‘record’, 1);
```

before this commit, the `createRecord` would fail,
as the `record` would not yet be fully unloaded, 
and its ID would still be reserved.

### DIFF
--- a/addon/-private/system/store.js
+++ b/addon/-private/system/store.js
@@ -2566,15 +2566,23 @@ Store = Service.extend({
 
     assert(`You can no longer pass a modelClass as the first argument to store._buildInternalModel. Pass modelName instead.`, typeof modelName === 'string');
 
-    let recordMap = this._internalModelsFor(modelName);
+    let internalModels = this._internalModelsFor(modelName);
+    let existingInternalModel = internalModels.get(id);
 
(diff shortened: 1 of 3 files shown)


addon/-private/system/store.js
```diff
@@ -2566,15 +2566,23 @@ Store = Service.extend({
 
     assert(`You can no longer pass a modelClass as the first argument to store._buildInternalModel. Pass modelName instead.`, typeof modelName === 'string');
 
-    let recordMap = this._internalModelsFor(modelName);
+    let internalModels = this._internalModelsFor(modelName);
+    let existingInternalModel = internalModels.get(id);
 
-    assert(`The id ${id} has already been used with another record for modelClass '${modelName}'.`, !id || !recordMap.get(id));
+    if (existingInternalModel) {
```
rwjblue: Mentioned this in chat, but I think this should be `if (existingInternalModel && existingInternalModel._isDematerializing) {` (otherwise we destroy everything and the assert just below is useless).

rwjblue:
👍 LGTM

PR #5102, opened 2017-07-29
[BUGFIX release] unloadRecord + findRecord + orphaned relationships

stefanpenner:
Given unloadRecord followed by a synchronous findRecord, we should be sure
any orphaned relationships are discarded.

### DIFF
--- a/addon/-private/system/model/internal-model.js
+++ b/addon/-private/system/model/internal-model.js
@@ -521,6 +521,11 @@ export default class InternalModel {
       this.cancelDestroy();
     }
     this._checkForOrphanedInternalModels();
+    if (this.isDestroyed || this.isDestroying) { return; }
+
+    // just in-case we are not one of the orphaned, we should still
+    // still destroy ourselves
+    this.destroy();
   }
 
   _checkForOrphanedInternalModels() {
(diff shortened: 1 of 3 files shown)


addon/-private/system/model/internal-model.js
```diff
@@ -521,6 +521,11 @@ export default class InternalModel {
       this.cancelDestroy();
     }
     this._checkForOrphanedInternalModels();
+    if (this.isDestroyed || this.isDestroying) { return; }
+
+    // just in-case we are not one of the orphaned, we should still
+    // still destroy outselves
```
stefanpenner: typo: outselves -> ourselves

addon/-private/system/store.js
```diff
@@ -1128,14 +1128,13 @@ Store = Service.extend({
     let internalModel = this._internalModelsFor(modelName).get(trueId);
 
     if (!internalModel) {
```
rwjblue: Ideally we can refactor this to avoid `if (!` with an else clause (and instead flipping the blocks around to remove the `!`). Definitely fine for a follow up PR though.

stefanpenner: nah i can doit as part of this one.

rwjblue: Awesome, thank you. It is much easier for me  to read this way.

stefanpenner:
released as v2.14.9

PR #5110, opened 2017-08-01
[BUGFIX beta] Implement `cacheKeyForTree` hook.

rwjblue:
This hook was added in [ember-cli/rfcs#90](https://github.com/ember-cli/rfcs/blob/master/complete/0090-addon-tree-caching.md), and essentially guarantees that a given tree returned by an addon is considered stable (or not).

In the case of Ember Data, we do not have different tree output based on our parent (project or addon).

Specifically, this implementation allows both an app _and_ a lazy engine to depend on `ember-data` without duplicating the `ember-data` assets in both the `assets/vendor.js` and `engine-dist/<engine-name>/assets/engine-vendor.js`.

Fixes https://github.com/ember-engines/ember-engines/issues/438.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,7 @@
 ---
 language: node_js
 sudo: false
+dist: trusty
 node_js:
   - "7"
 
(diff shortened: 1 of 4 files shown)


rwjblue:
Rebased.

thec0keman:
@rwjblue will this be released in 2.15?  would it be possible to get this released as a part of 2.14?  Thanks!

PR #5117, opened 2017-08-09
Give Model a static toString method

DingoEatingFuzz:
This changes the dreaded `(unknown mixin)` to something like `model:person`.

This will make things like this error much more useful:

> Error: Assertion Failed: You defined the 'run' relationship on (unknown mixin), but multiple possible inverse relationships of type (unknown mixin) were found on (unknown mixin). Look at http://emberjs.com/guides/models/defining-models/#toc_explicit-inverses for how to explicitly specify inverses

### DIFF
--- a/addon/-private/system/model/model.js
+++ b/addon/-private/system/model/model.js
@@ -1860,6 +1860,16 @@ Model.reopenClass({
     get(this, 'transformedAttributes').forEach((type, name) => {
       callback.call(binding, name, type);
     });
+  },
+
+  /**
+   Returns the name of the model class.
+
+   @method toString
+   @static
+   */
+  toString() {
+    return `model:${get(this, 'modelName')}`;
   }
 });
 
--- a/tests/unit/model-test.js
+++ b/tests/unit/model-test.js
@@ -19,7 +19,6 @@ module('unit/model - DS.Model', {
       isDrugAddict: DS.attr('boolean'),
       isArchived: DS.attr()
     });
-    Person.toString =  () => 'person';
 
     env = setupStore({
       adapter: DS.JSONAPIAdapter,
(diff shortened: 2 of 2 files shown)


rwjblue:
If you rebase (to get https://github.com/emberjs/data/commit/49a66fe3b10bdddb5b53ff7912b209eecb808e03) this should pass in CI...

rwjblue:
@runspired - Wanted to get your thoughts here too...

runspired:
I thought Stef had recently fixed this, hrm

stefanpenner:
This is fixed, if you load models via the container and a modern resolver. The above should only benefit tests. And if so, should likely be a test specific reopen.

DingoEatingFuzz:
@stefanpenner, can you elaborate?

This code:
```js
  model() {
    return RSVP.hash({
      nodes: this.get('store').findAll('node'),
      agents: this.get('store').findAll('agent'),
    }).then(({ nodes }) => {
      console.log(nodes.get('firstObject').toString());
      console.log(nodes.get('firstObject').constructor.toString());
    });
  },
```

Yields this output:

```
<nomad-ui@model:node::ember844:69b5c85f-5356-4365-b75a-e37ca76adfc3>
(subclass of DS.Model)
```

With these dependencies:

```
DEBUG: Ember           : 2.14.1
DEBUG: Ember Data      : 2.14.4
```

Was this fixed after 2.14.4, or is there another explanation?

stefanpenner:
> <nomad-ui@model:node::ember844:69b5c85f-5356-4365-b75a-e37ca76adfc3>
> (subclass of DS.Model)

ah..

stefanpenner:
This seems reasonable, unless we want to remove `modelName` from classes, as that attribute does pollute the class...

I'll have to think about this some more.

rwjblue:
IMHO, deprecating `modelName` is a separate concern, we should just make sure that we don't box ourselves in a corner with it. IMHO, the same thing could be solved without `modelName` via `_debugContainerKey` (for example)...

stefanpenner:
> IMHO, the same thing could be solved without modelName via _debugContainerKey (for example)...

On the class?

runspired:
I don't believe it is possible for us to move off of `modelName` being on `DS.Model`  without moving to ES Classes.  Moving to ES Classes is on the roadmap anyway, and would mean that the class name would be available (and the default) for this situation.  Due to this, I personally do not see any issue with moving this PR forward with the use of `modelName`.

@rwjblue @stefanpenner what do you think?

rwjblue:
Moving away from modelName just means we need some other mechanism for the same rough thing, whenever we add that we can easily update this `toString` implementation...

PR #5123, opened 2017-08-13
Ensure we are properly testing multiple Ember versions.

rwjblue:
Prior to this commit (since c2ca068) we were only testing ember-source@2.11. Even though we had a `config/ember-try.js` with many scenarios, none of them properly removed the `ember-source` version (even though ba7f46221 was specifically trying to do this :sob:).

The fundamental issue here is:

* `ember-try` supports removing packages, but requires that you specify them
  in the `dependencies` or `devDependencies` objects.  Since my changes in
  ba7f46221 added `"ember-source": null` to the root of the `npm` config
  it was completely ignored :disappointed:.
* `ember-source` emits a warning to the console when it is present **and**
  the bower package for `ember` is in use.  Unfortunately, this warning is
  a bit misleading (it doesn't tell you that the `ember-source` from `npm`
  will always win). The `ember-source` addon should be updated to throw an
  error when this misconfiguration is detected.
* `ember-cli` will **always** use `ember-source` if it is present, and
  completely ignore whatever is in `bower.json`. IMHO, this default was not
  correct and `ember-cli` should have done the same thing that `ember-data`
  did during its migration from bower to `npm`: if both packages are present
  default to `bower.json` version with a warning.

---

There are almost certainly failing tests in the various scenarios. I pushed this in a branch so that others can help track down and fix the various issues.

### DIFF
--- a/.bowerrc
+++ b/.bowerrc
@@ -1,4 +1,0 @@
-{
-  "directory": "bower_components",
-  "interactive": false
-}
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,7 @@
 ---
 language: node_js
 sudo: false
+dist: trusty
 node_js:
   - "7"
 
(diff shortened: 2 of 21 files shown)


rwjblue:
I'm very sorry I didn't properly handle this back in ba7f46221 😭.

rwjblue:
Migrated from a warning to an error in https://github.com/emberjs/ember.js/pull/15582.

rwjblue:
Looks like nearly all of the failures on ember-release, ember-beta, ember-canary, and ember-alpha channels are related to the [Ember.MODEL_FACTORY_INJECTIONS](https://emberjs.com/deprecations/v2.x/#toc_ember-model_factory_injections-removed) deprecation.  https://github.com/emberjs/data/pull/5127 should fix those.

stefanpenner:
I'm working on this.

stefanpenner:
fixed some more stuff, mostly re: 1.13.x. Still more to debug.

rwjblue:
Looking good @stefanpenner, looks like all but Ember 2.4 and 1.13 are passing now.

PR #5128, opened 2017-08-15
Avoid MODEL_FACTORY_INJECTION deprecations in tests on Ember >= 2.14.

stefanpenner:


tests/integration/relationships/belongs-to-test.js
```diff
@@ -11,7 +11,6 @@ const { attr, hasMany, belongsTo } = DS;
 const { hash } = RSVP;
 
 let env, store, User, Message, Post, Comment, Book, Chapter, Author, NewMessage;
-const injectionValue = Ember.MODEL_FACTORY_INJECTIONS;
 
 module("integration/relationship/belongs_to Belongs-To Relationships", {
   beforeEach() {
```
stefanpenner: @rwjblue previously, you ran `setupModelFactoryInjections` here mistakenly. Turns out the code was only resetting the injection stuff here, not actually setting it up. Instead I simply removed it.

PR #5133, opened 2017-08-16
Defer to `record._{changedKeys,assign}` if present

hjdivad:
This is a small step towards enabling a swappable model class which is useful for people who want to use ember data (and indeed even inter-op with DS.Model) but have less well-defined APIs or APIs that are large enough where defining their own DS.Model for each

It is private for now to enable experimentation.

### DIFF
--- a/addon/-private/system/model/internal-model.js
+++ b/addon/-private/system/model/internal-model.js
@@ -2,7 +2,7 @@ import { assign, merge } from '@ember/polyfills';
 import { set, get } from '@ember/object';
 import { copy } from '@ember/object/internals';
 import EmberError from '@ember/error';
-import { isEqual, isEmpty } from '@ember/utils';
+import { isEmpty } from '@ember/utils';
 import { setOwner } from '@ember/application';
 import { run } from '@ember/runloop';
 import RSVP, { Promise } from 'rsvp';
@@ -585,7 +585,8 @@ export default class InternalModel {
       changedKeys = this._changedKeys(data.attributes);
     }
 
-    emberAssign(this._data, data.attributes);
+    this._assignAttributes(data.attributes);
+
     this.pushedData();
 
     if (this.hasRecord) {
@@ -1067,9 +1068,9 @@ export default class InternalModel {
     this.didCleanError();
     let changedKeys = this._changedKeys(data);
 
-    emberAssign(this._data, this._inFlightAttributes);
+    this._assignAttributes(this._inFlightAttributes);
     if (data) {
-      emberAssign(this._data, data);
+      this._assignAttributes(data);
     }
 
     this._inFlightAttributes = null;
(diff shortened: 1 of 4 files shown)


hjdivad:
@rwjblue the PR we were just discussing

addon/-private/system/model/internal-model.js
```diff
@@ -583,10 +583,19 @@ export default class InternalModel {
     let changedKeys;
 
     if (this.hasRecord) {
-      changedKeys = this._changedKeys(data.attributes);
+      if (typeof this._record._changedKeys === 'function') {
```
rwjblue: Discussed offline with @hjdivad, but I'd like to see these checks be done inside the `InternalModel.prototype._changedKeys` method itself. It would clean up some of this extra branching logic here and below, and also help prevent issues from future refactors that might add additional calls to `_changedKeys` but forget to do this `this._record._changedKeys` checking manually.

rwjblue:
I'm also curious if it might make sense to update `InternalModel.prototype.getRecord` to set `this._changedKeys = this._record._changedKeys` if present. I think that should be saved for a separate refactor (with some profiling evidence to confirm it doesn't have negative impact).

addon/-private/system/model/internal-model.js
```diff
@@ -583,10 +583,19 @@ export default class InternalModel {
...
+        changedKeys = this._record._changedKeys(data.attributes);
+      } else {
+        changedKeys = this._changedKeys(data.attributes);
+      }
+    }
+
+    if (this.hasRecord && this._record._assignAttributes) {
```
rwjblue: This could follow a similar pattern to what I mentioned RE: `_changedKeys`. We could add `this._assignAttributes` method and have it do the conditional?  Not sure that is useful though if this is literally the only place it happens...

hjdivad: I'm checking to see if this is actually necessary; might drop it if not otherwise i agree i'll switch to the other pattern

addon/-private/system/model/internal-model.js
```diff
@@ -1218,6 +1219,10 @@ export default class InternalModel {
     @private
   */
   _changedKeys(updates) {
+    if (this.hasRecord && typeof this._record._changedKeys === 'function') {
```
stefanpenner: If we go this route, I think we should consider not branching here. Rather invoking some symbol on `this._record` which is then implemented by DS.Model as well. Thoughts?

hjdivad: I agree your suggestion is a better approach

hjdivad: @stefanpenner we can drop the function check, but we'll still need to branch depending on `this.hasRecord`

hjdivad:
@stefanpenner i've updated to branch only on `this.hasRecord`; seem better to you?

hjdivad:
@stefanpenner per our conversation OOB we don't need to branch in `changedKeys` because as you pointed out we shouldn't (and indeed don't) even call it in `setupData` unless `this.hasRecord`.

hjdivad:
superseded by model data work

Issue #5134, opened 2017-08-17
id relationship does not update after the record is provided


luxferresum:
If we load an id-relationship to an not yet loaded record and then later add load that record the belongstoreference is not updated.

```
  async model() {

    // first load foos with id-relations to not yet loaded resources
    // the model has resource: hasMany('resource')
    // and the relationship is resource : { data: { id: 'r1', type: 'resource' } }
    // however resource r1 is NOT sideloaded
    const foos = await this.store.findAll('foo');
    
    // now we load all resources, including all that were referenced by foos
    // this means after this the resource 'r1' is loaded!
    await this.store.findAll('resource');

    // and now the relationship is null on ember-data 2.14.10.
    // this works on ember-data 2.13
    return foos.get('firstObject').belongsTo('resource').value();
  }
```

A git repo to reproduce can be found [here](https://github.com/luxferresum/ember-data-problem-empty-rels).

stefanpenner:
failing test: https://github.com/emberjs/data/pull/5135

luxferresum:
This seems to be fixed on master and 2.16. I haven't checked other versions.

PR #5135, opened 2017-08-17
failing test for reference when id is procided

luxferresum:
This is a failing test for #5134.

This is my first Contribution to ember-data. If I should change something please tell me!


luxferresum:
This seems to be fixed on master and `2.16`. I haven't checked other version. I will close the issue but leave this open because I think it would be good to have the test as regression test.

luxferresum:
Is anything holding this back?

bmac:
Thanks for the ping @luxferresum. I just re-trigger the travis build. If it passes again I will merge.

bmac:
Thanks @luxferresum.

PR #5142, opened 2017-08-23
[BUGFIX release] Load many2many relationships correctly

### LINKED ISSUE #5140: 2.14.7 regression store.query breaks some many2many relationships
This issues doesn't exist in 2.13.2 - 2.14.6, it appears for the first time in 2.14.7 and still exists in 2.15.0-beta.4.

Sorry for the meandering ticket and all the leaky domain stuff from using my own API, not really sure what details might be important here. 

I'm doing some pre-loading of relationships in my router to save network requests. In my route I have:
```
afterModel(model) {
  return store.query('objective', {filters: {courses}});
}
```
Which loads all the `objectives` that would be loaded with a `course.get('objectives')`. The goal is for them to then be available in the store without needing to fetch each one individually.

This works as long as each `objective` returned only has a single course in its relationship like
``` 
{
  id: 1
  courses: [1]
}
```

but it seems to fail if there are more than one:
``` 
{
  id: 1
  courses: [1, 2]
}
```

The result ends up being that the loaded `course` model somehow gets confused and thinks it is only linked to some of the objectives. Operations like `{{#each course.objective}}` or `{{course.objectives.length}}` or even `course.hasMany('objectives').ids()` all provide incorrect data.

This *only* happens when the route is visited directly (by refreshing the page or typing in the URL). If a `link-to` is followed (even one that uses an `id` and not an object for the model then the relationships all load. 

I've tried to create a small reproduction of this, unfortunately I couldn't make it work with ember-twiddle but I have breaking code here:

https://github.com/jrjohnson/reproduction-ember-data-query-does-not-update-store
and a running app example at:
https://grader-duck-81038.netlify.com/

I've created three routes:
1. Working - no `afterModel` hook used, works fine
2. Broken - `afterModel` hook used, breaks
3. Working with fewer related items  uses `afterModel` hook, but loads a different `course` object with `objectives` that are only related to itself.

(edited: added info that 2.15.0-beta.4 was tested)
(edited: I'm not sure where I got it wrong originally but this issues actually showed up in 2.14.7 for the first time.)

jrjohnson:
Fair warning, I really don't know what this fix does. I think the original commit I'm patching had a small logic error that this fixes, but I hope @stefanpenner (who did the original work) can verify that. This change does fix the issue I was seeing though so I'm hopeful it is correct.

When a many2many relationship exists and more than one related item is
included in the relationship ensure that both sides load completely.

This issue was introduced in e430edee24591afa562167c6f5775b61d1031b75

Fixes #5140

### DIFF
--- a/addon/-private/system/relationships/state/has-many.js
+++ b/addon/-private/system/relationships/state/has-many.js
@@ -209,7 +209,7 @@ export default class ManyRelationship extends Relationship {
 
     for (let i = 0; i< internalModels.length; i++) {
       let internalModel = internalModels[i];
-      if (this.canonicalMembers.has(internalModel)) {
+      if (forCanonical.indexOf(internalModel) !== -1) {
         continue;
       }
 
--- a/tests/unit/model/relationships/has-many-test.js
+++ b/tests/unit/model/relationships/has-many-test.js
@@ -477,6 +477,136 @@ test('hasMany with duplicates from payload', function(assert) {
   });
 });
 
+test('many2many loads both sides #5140', function(assert) {
+  assert.expect(3);
+
+  const Tag = DS.Model.extend({
+    name: DS.attr('string'),
+    people: DS.hasMany('person', { async: false })
+  });
+
+  Tag.reopenClass({
+    toString() {
+      return 'tag';
+    }
+  });
+
+  const Person = DS.Model.extend({
+    name: DS.attr('string'),
+    tags: DS.hasMany('tags', { async: false })
+  });
+
+  Person.reopenClass({
+    toString() {
+      return 'person';
+    }
+  });
+
+  let env = setupStore({ tag: Tag, person: Person });
+  let { store } = env;
+
+  run(() => {
+    // first we push in data with the relationship
+    store.push({
+      data: [
+        {
+          type: 'person',
+          id: 1,
+          attributes: {
+            name: 'David J. Hamilton'
+          },
+          relationships: {
+            tags: [{
+              data: {
+                type: 'tag',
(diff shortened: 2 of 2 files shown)


addon/-private/system/relationships/state/has-many.js
```diff
@@ -209,7 +209,7 @@ export default class ManyRelationship extends Relationship {
 
     for (let i = 0; i< internalModels.length; i++) {
       let internalModel = internalModels[i];
-      if (this.canonicalMembers.has(internalModel)) {
```
stefanpenner: I'm super confused how these end up being different. So  thank you for providing a test which I can debug :)

sly7-7: @stefanpenner I wondered that too, and I think could be a side effect here: https://github.com/emberjs/data/blob/master/addon/-private/system/relationships/state/relationship.js#L167 which add internal models to the relationship canoninal members. So this may leads to `this.canonicalMembers` and `forCanonical`beeing different

stefanpenner: yup, that seems to be it.

stefanpenner:
@jrjohnson thanks for the PR, after debugging your failing test I now see the issue! I have adapted your fix slightly (just a simplification) -> https://github.com/emberjs/data/pull/5144

jrjohnson:
Great! I'm glad it put you on the right track. My fix was pretty much a shot in the dark.

stefanpenner:
@jrjohnson your fix and test really made this super actionable. I wish all bug reports came this way :)

PR #5144, opened 2017-08-24
[BUGFIX release] Load many2many relationships correctly

stefanpenner:
When a many2many relationship exists and more than one related item is
included in the relationship ensure that both sides load completely.

This issue was introduced in e430edee24591afa562167c6f5775b61d1031b75

---

note: this is https://github.com/emberjs/data/pull/5142 with some simplification


stefanpenner:
backport beta: https://github.com/emberjs/data/pull/5145

PR #5147, opened 2017-08-24
[BUGFIX release] fix relationship merging

stefanpenner:
When relationship information is merged, the newest information should win.
Unfortunately prior to this PR, relationship containing only `link` information, would mistakenly purge any existing `data` information on that relationship.

1. Update One
```js
{
    data: { /* existing data */ },
    Link: { /* existing data */ },
}
```

2. Update Two

```js
{
    Link: { /* new data */ },
}
```

Previously resulted in:

```js
{
   link: { /* new data */ },
   data: null // oops
}
```

But now (correctly) results in:

```js
{
   link: { /* new data */ },
   data: { /* existing data */ }
}
```

### DIFF
--- a/tests/integration/relationships/nested-relationship-test.js
+++ b/tests/integration/relationships/nested-relationship-test.js
@@ -1,0 +1,148 @@
+import setupStore from 'dummy/tests/helpers/store';
+import Ember from 'ember';
+
+import {module, test} from 'qunit';
+
+import DS from 'ember-data';
+
+const { get, run } = Ember;
+const { attr, hasMany, belongsTo } = DS;
+
+let env, store, serializer, Elder, MiddleAger, Kid;
+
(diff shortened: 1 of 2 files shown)


tests/integration/relationships/nested-relationship-test.js
```diff
@@ -0,0 +1,148 @@
...
+  Server loading tests
+*/
+
+test('Sideloaded nested relationships load correctly', function(assert) {
+  run(() => {
+    serializer.pushPayload(store, {
+      data: [
+        {
+          id: '1',
+          type: 'kids',
```
hjdivad: Do we need this at all?

Can't we minimize the test to only have `middle-ager` and `elder`?

I think now that we've figured out the root cause the kid relationship adds noise to the test.

PR #5150, opened 2017-08-25
[BUGFIX release] Enumerate relationships provided, not all on a given…

stefanpenner:
… schema

### DIFF
--- a/addon/-private/system/store.js
+++ b/addon/-private/system/store.js
@@ -2429,9 +2429,10 @@ Store = Service.extend({
     // payload push.  In the common case where we are pushing many more
     // instances than types we want to minimize the cost of looking up the
     // inverse map and the overhead of Ember.get adds up.
-    let modelNameToInverseMap = Object.create(null);
+    let modelNameToInverseMap;
 
     for (let i = 0, l = pushed.length; i < l; i += 2) {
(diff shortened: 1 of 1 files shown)


addon/-private/system/store.js
```diff
@@ -2429,9 +2429,10 @@ Store = Service.extend({
     // payload push.  In the common case where we are pushing many more
     // instances than types we want to minimize the cost of looking up the
     // inverse map and the overhead of Ember.get adds up.
-    let modelNameToInverseMap = Object.create(null);
+    let modelNameToInverseMap;
 
     for (let i = 0, l = pushed.length; i < l; i += 2) {
+      modelNameToInverseMap = modelNameToInverseMap || Object.create(null);
```
chadhietala: Won't `modelNameToInverseMap` always be `undefined` here?

chadhietala: Is it worth preventing the `ToBoolean` conversion here e.g. `(modelNameToInverseMap !== undefined)`

stefanpenner: I believe the cost of `ToBoolean` coercion in this context is negligible. I could default `modelNameToInverseMap` to `false` though, if you would prefer. That would be a minimal enough change.  Let me know

chadhietala: No this seems fine.

PR #5151, opened 2017-08-25
[CLEANUP] PERF prevent duplicate recordArrayManager signals on push o…

stefanpenner:
…f a new record

* if the record is new, ensure only 1 recordWasLoaded is called (after setup)
* if the record is existing, ensure only 1 recordDidChange is called (after setup)
* decouple InternalModel slightly more from recordArrayManager
* remove now defunct code

### DIFF
--- a/addon/-private/system/store.js
+++ b/addon/-private/system/store.js
@@ -2017,9 +2017,15 @@ Store = Service.extend({
     heimdall.increment(_load);
     let internalModel = this._internalModelForId(data.type, data.id);
 
+    let isUpdate = internalModel.currentState.isEmpty === false;
+
     internalModel.setupData(data);
 
-    this.recordArrayManager.recordDidChange(internalModel);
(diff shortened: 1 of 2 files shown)


addon/-private/system/store.js
```diff
@@ -2017,9 +2017,15 @@ Store = Service.extend({
     heimdall.increment(_load);
     let internalModel = this._internalModelForId(data.type, data.id);
 
+    let isUpdate = internalModel.currentState.isEmpty === false;
+
     internalModel.setupData(data);
 
-    this.recordArrayManager.recordDidChange(internalModel);
+    if (isUpdate) {
+      this.recordArrayManager.recordDidChange(internalModel);
```
stefanpenner: previously, when pushing a new record this was firing, even though a new record has no updates. It merely starts off its existence in that state...

This resulted in every new record going through the record-array manager twice on initial push...

Issue #5111, opened 2017-08-02
peekAll contains null entries after unloadRecord


Kilowhisky:
This is related to #4963 and #5025 (probably). #4963 affected me though that exact issue appears to have been resolved. I'm thinking that #4963 was hiding #5025 from me before. Now i'm upgrading to 2.14.9 from 2.12 and i'm getting another issue. 

This code is run after i unload a number of records from the store. 
```
        let store = this.get('store');
        let records = store.peekAll('asset-position');
        let orderedBy = records.sortBy('utc');
```

the `sortBy('utc')` fails because the peekAll record array still contains references to unloaded records.
```
ember.debug.js:26513 Assertion Failed: Cannot call get with 'utc' on an undefined object.
Error
    at Object.assert (http://localhost:4200/skyrouter3/assets/vendor.js:29571:15)
    at get (http://localhost:4200/skyrouter3/assets/vendor.js:40963:64)
    at http://localhost:4200/skyrouter3/assets/vendor.js:54265:43
    at InternalArray.sort (native)
    at Array.sort (native)
    at Class.sortBy (http://localhost:4200/skyrouter3/assets/vendor.js:54261:29)
    at Class.loadMoreAssetPositionsLastUtcReset (http://localhost:4200/skyrouter3/assets/sky-router-3.js:42129:37)
    at Object.applyStr (http://localhost:4200/skyrouter3/assets/vendor.js:61345:18)
    at sendEvent (http://localhost:4200/skyrouter3/assets/vendor.js:38746:22)
    at Class.trigger (http://localhost:4200/skyrouter3/assets/vendor.js:54394:33)
```

If this is too close to those other issues please close. 

EDIT:
For show:
```
        let records = store.peekAll('asset-position');
        console.log(records.get('length'));
        console.log(records.filter(y => y != null).get('length');
```
This outputs `1844` and `700`. Big difference.

workmanw:
We just ran into a bug today where this happened in our application.

stefanpenner:
I think a similar to fix to the others can be applied here. Specifically, `unloadRecord` needs to act like client-side delete when it comes to recordArrays

workmanw:
@stefanpenner Failing test: #5157

nag5000:
I have the same bug, but in a little different situation:
```js
// /routes/courses.js
model() {
  return this.store.peekAll('course');
}

// /controllers/courses.js
draftCourses: computed('model.@each.isDraft', function() {
[...]

PR #5157, opened 2017-08-28
Added failing test for #5111

workmanw:


urbany:
Is this bug still present in ED@2.15.0 ?

workmanw:
@urbany Unfortunately so. I just rebased against master and pushed so you'll see a travis failure here shortly. I also ran the tests locally and this test is still failing.

tests/integration/peek-all-test.js
```diff
@@ -89,3 +89,25 @@ test("Calling store.peekAll() after creating a record should return correct data
...
+    store.createRecord('person', { name: "Tomster" });
+    store.createRecord('person', { name: "Zoey" });
+  });
+
+  assert.equal(get(store.peekAll('person'), 'length'), 2, 'should contain one person');
+
+  run(() => {
+    allPeople.objectAt(0).unloadRecord();
+
+    assert.equal(get(allPeople, 'length'), 2, 'Unload does not complete until the end of the loop');
```
fivetanley: i think we want this to happen before the loop ends to reduce # of wtfs

workmanw: I definitely agree that that would be more ideal. I had an extended conversation with @stefanpenner about that on slack. IIRC his near term objective was to get thing backing to how they were at 2.12, then later down the line rethink some of this stuff. That was several months back so it may have changed. But that was the thought process behind this test.

runspired:
I've confirmed that #5378 will resolve this issue, and that PR has a test covering this case.

PR #5165, opened 2017-09-06
[DOC] Update README build step script name

bantic:
Changes step 3 of "Building Ember Data" to instruct using the npm script
"production" rather than "build:production", which doesn't exist.


locks:
😗 👌

Issue #5167, opened 2017-09-06
multiple calls to peekAll and unloadAll put store into inconsistent state


bantic:
This seems similar to #5111, perhaps a duplicate.

**edit**: First observed using ember data version `2.14.11` and reproduced on `2.16.0-canary`.

I have 2 models, one of which `hasMany` of the other, and I noticed that after repeating a flow that involved calling `unloadAll` to remove the child elements and then `pushPayload` to add new ones, the store started reporting inconsistent results for both `peekAll` and `parentModel.get('childModels.length')`.

For example, if the models are Person and Car and Person has-many cars (and car belongs-to person), the following code puts the store into the inconsistent state:

```javascript
let person = store.createRecord('person');
store.push({ data: { type: 'car', id: '1'}});
store.peekAll('car').forEach(car => car.set('person', person));
console.log(person.get('cars.length')); // 1, as expected

store.unloadAll('car');
console.log(store.peekAll('car').get('length')); // 0, expected
console.log(person.get('cars.length')); // 0, expected

// load the car again
store.push({data: { type: 'car', id: '1'}});
store.peekAll('car').forEach(car => car.set('person', person));
console.log(store._internalModelsFor('car')); // 0! (should be: 1)
console.log(person.get('cars.length')); // 0! (should be: 1)

store.unloadAll('car');
console.log(store.peekAll('car').get('length')); // 1! (should be: 0)
```

This unexpected behavior appears to be related to using `peekAll` + `forEach` to set the parent property on the peeked records. If the lines in the code snippet above that call `car.set('person', person)` are removed, things behave as expected.

bantic:
Failing test: #5168

workmanw:
It seems related to me also. I also have #5157 which could be a different flavor of the same problem.

mqje:
Hello, I think have a similar issue with the `unloadAll()` method. I am trying to unload some 'type' with this method :
```
const self = this;
['type1', 'type2', 'type3'].forEach((type) => {
self.get('store').unloadAll(type);
});
```
But it doesn't seems to unload my record at all, the only way I find out to unload my record is :
```
self.get('store').unloadAll();
```
This method work, but I'm losing some other record in my store too, which is something not intended.

Also, it worked with ember@2.12.
I am currently on 2.15, and it doesn't work.

runspired:
I confirmed that this is fixed by #5378 and added a test in case it were to resurface.

PR #5168, opened 2017-09-06
Add failing test for #5167

bantic:
Demonstrates the behavior described in #5167.

Removing these lines from `loadCar` (and the assertions of the value of `person.get('cars.length')`)
makes the tests pass:

```
  run(() => {
    env.store.peekAll('car').forEach(car => car.set('person', person));
  });
```


bantic:
Added comments in the test to point out which tests fail.

runspired:
This is covered by the tests added in #5378

PR #5176, opened 2017-09-15
fix node 4

kellyselden:
This file introduced let and const without strict mode, and is now failing on node 4.

cc @runspired


runspired:
Sorry about that! Thought we had finally hit an LTS beyond those concerns.

runspired:
cc @locks you'll also want to backport this

kellyselden:
You might want to switch CI to node 4 to catch this in the future.

PR #5178, opened 2017-09-15
Ignore included resources without that don't have a corresponding ember-data model

### LINKED ISSUE #4568: TypeError: Cannot read property 'type' of null
If our server json api response includes a model that our ember application does not know about we got a `TypeError: Cannot read property 'type' of null` error.

Would a more desired behavior be to ignore unknown models? Let me know your thoughts, would be happy to help with a fix.
#### Stack trace

```
TypeError: Cannot read property 'type' of null
    at r._pushInternalModel (https://example.com/assets/vendor-ff470fb65f1009ec8ea7bcea14d830f0.js:26:2600)
    at r.push (https://example.com/assets/vendor-ff470fb65f1009ec8ea7bcea14d830f0.js:26:2332)
    at https://example.com/assets/vendor-ff470fb65f1009ec8ea7bcea14d830f0.js:25:23950
    at Object.run (https://example.com/assets/vendor-ff470fb65f1009ec8ea7bcea14d830f0.js:5:11070)
    at r._adapterRun (https://example.com/assets/vendor-ff470fb65f1009ec8ea7bcea14d830f0.js:26:3526)
    at https://example.com/assets/vendor-ff470fb65f1009ec8ea7bcea14d830f0.js:25:23853
    at y (https://example.com/assets/vendor-ff470fb65f1009ec8ea7bcea14d830f0.js:18:26952)
    at b (https://example.com/assets/vendor-ff470fb65f1009ec8ea7bcea14d830f0.js:18:27047)
    at v (https://example.com/assets/vendor-ff470fb65f1009ec8ea7bcea14d830f0.js:18:26855)
    at https://example.com/assets/vendor-ff470fb65f1009ec8ea7bcea14d830f0.js:13:13635
```
#### Versions Reproduced

ember-data: 2.8.1
ember-data: ^2.4.0

lukemelia:
When the JSON API serializer encounters a resource with a type that doesn't resolve
to one of the application's Ember Data models, it replaces the object with `null` and
emits a warning. Prior to this commit, if that `null` was one of the sideloaded
(i.e. `included`) resources, the store would throw trying to ask for the `null`'s type.
This commit makes the serializer drop any nulls out of the list of included resources
to avoid this problem.

To summarize, if your API attempts to include resources that you don't have models for,
a warning will be emitted and they will be ignored.

Fixes #4568

### DIFF
--- a/addon/serializers/json-api.js
+++ b/addon/serializers/json-api.js
@@ -152,11 +152,13 @@ const JSONAPISerializer = JSONSerializer.extend({
     }
 
     if (Array.isArray(documentHash.included)) {
-      let ret = new Array(documentHash.included.length);
-
+      let ret = new Array();
       for (let i = 0; i < documentHash.included.length; i++) {
         let included = documentHash.included[i];
-        ret[i] = this._normalizeResourceHelper(included);
+        let normalized = this._normalizeResourceHelper(included);
+        if (normalized !== null) { // can be null when unknown type is encountered
+          ret.push(normalized);
+        }
       }
 
       documentHash.included = ret;
--- a/tests/integration/serializers/json-api-serializer-test.js
+++ b/tests/integration/serializers/json-api-serializer-test.js
@@ -138,6 +138,71 @@ testInDebug('Warns when normalizing an unknown type', function(assert) {
   }, /Encountered a resource object with type "UnknownType", but no model was found for model name "unknown-type"/);
 });
 
+testInDebug('Warns when normalizing payload with unknown type included', function(assert) {
+  var documentHash = {
+    data: {
+      type: 'users',
+      id: '1',
+      attributes: {
+        'first-name': 'Yehuda',
+        'last-name': 'Katz'
+      },
+      relationships: {
+        company: {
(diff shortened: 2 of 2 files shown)


addon/serializers/json-api.js
```diff
@@ -153,10 +153,18 @@ const JSONAPISerializer = JSONSerializer.extend({
...
         let included = documentHash.included[i];
-        ret[i] = this._normalizeResourceHelper(included);
+        let normalized = this._normalizeResourceHelper(included);
+        if (!normalized) {
+          needsCompacting = true;
+        }
+        ret[i] = normalized;
+      }
+      if (needsCompacting) {
+        ret = ret.filter(Boolean);
```
runspired: I think we should just use push and an explicit check for not null vs preallocating length here.

lukemelia: Sounds good @runspired. I was concerned about changing the performance characteristics of the happy path, but I agree that is clearer. I'll make the change now.

lukemelia: ✅

bmac:
Thanks @lukemelia.

PR #5183, opened 2017-09-18
[DOCS release] Add missing "relationship" field to RESTAdapter.findBelongsTo docs

gabrielgrant:


gabrielgrant:
It's not clear to me why AppVeyor is failing (looking through the long seems to only show passing tests). Is there something I can do to fix that?

bmac:
Thanks @gabrielgrant.

PR #5185, opened 2017-09-21
support multiple adapters in belongsTo and hasMany

larsklevan:
Currently the adapter to use for loading records in a belongsTo or hasMany is determined by the model that has the association. This makes it so that if you have different adapters for the two models the wrong one is used to load the associated models, resulting in incorrect headers, etc.

For example if I have an `Author` configured to use `RESTAdapter` with a hasMany for `Book` that uses `JSONAPIAdapter` you will get the `JSONAPIAdapter` if you do `store.findRecord('book', 1)` but the `RESTAdapter` if you do `author.get('books')`. In this case the result is that when calling `findRecord` you get the expected `Accept` header of `application/vnd.api+json` where when you use the `hasMany` you get `application/json` instead.

[Ember Twiddle](https://ember-twiddle.com/4b4d6f29d238fb611be705b02b85c90d?openFiles=templates.application.hbs%2C)

### DIFF
--- a/addon/-private/system/store.js
+++ b/addon/-private/system/store.js
@@ -1184,7 +1184,7 @@ Store = Service.extend({
     @return {Promise} promise
   */
   findHasMany(internalModel, link, relationship) {
-    let adapter = this.adapterFor(internalModel.modelName);
+    let adapter = this.adapterFor(relationship.type);
 
     assert(`You tried to load a hasMany relationship but you have no adapter (for ${internalModel.modelName})`, adapter);
     assert(`You tried to load a hasMany relationship from a specified 'link' in the original payload but your adapter does not implement 'findHasMany'`, typeof adapter.findHasMany === 'function');
@@ -1201,7 +1201,7 @@ Store = Service.extend({
     @return {Promise} promise
   */
   findBelongsTo(internalModel, link, relationship) {
-    let adapter = this.adapterFor(internalModel.modelName);
+    let adapter = this.adapterFor(relationship.type);
 
     assert(`You tried to load a belongsTo relationship but you have no adapter (for ${internalModel.modelName})`, adapter);
     assert(`You tried to load a belongsTo relationship from a specified 'link' in the original payload but your adapter does not implement 'findBelongsTo'`, typeof adapter.findBelongsTo === 'function');
--- a/tests/unit/store/finders-test.js
+++ b/tests/unit/store/finders-test.js
@@ -97,7 +97,7 @@ test('findHasMany does not load a serializer until the adapter promise resolves'
 
   let deferedFind = defer();
 
-  this.env.registry.register('adapter:person', DS.Adapter.extend({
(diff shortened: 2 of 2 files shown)


larsklevan:
Any feedback or questions on this?

pangratz:
Hey @larsklevan,

thanks for your first contribution and many thanks for your patience! I finally had the time to take a look at this PR, so here's my feedback.

---

_DISCLAIMER_

Sorry, I was kind of in the mood for a more detailed answer. TL;DR; is a suggestion for an alternative implementation, which is backwards compatible and brings in the new functionality via an addon.

---

So the use case you're having is a valid one, and IMHO there should be support for this out of the box. Just a little bit of explanation on why the adapter of the record of relationship (and not the adapter for the type of relationship) is used at the moment: the idea behind it is that the adapter for the record knows best how to fetch the relationship. But since people are stumbling over a different use case, it's definitely time to improve this.

At the moment your suggestions are technically a breaking change, since from now on a potentially different adapter could be used; but we need to support backwards compatibility. Also, if a new public facing feature is added to ember data, it should go through the [RFC process](https://github.com/emberjs/rfcs), so the community can weigh in with suggestions and alternative perspectives.

The issue is that the RFC process can take a bit of time until everything is settled and agreed on. Until such an RFC is accepted, we can do something still. What do you think of refactoring this so only internal hooks are added. By this we don't expose new public API (no hence no RFC needed (for now)) and guarantee backwards compatibility. Then we could create an addon which exposes a possible new public API and makes the feature available by a simple `ember install ember-data-adapter-find-relationship-hooks`.

So, here's the plan:

### interim solution with internal only hooks

```js
// ember-datas/addon/-private/system/store.js

export default Service.extend({

  findHasMany(internalModel, link, relationship) {
    let adapter this._adapterForFindHasMany(...arguments);
    ...
  },

  _adapterForFindHasMany(internalModel, link, relationship) {
    return this.adapterFor(internalModel.modelName);
  },

  ...

});
```

### Addon

Offers a more public API

```js
// ember-data-adapter-find-relationship-hooks/index.js

export default Mixin.create({

  _adapterForFindHasMany(internalModel, link, relationship) {
    if (this.adapterForFindHasMany) {
      let record = internalModel._record;
      return this.adapterForFindHasMany(record, link, relationship);
    },

  },

});
```

which can be used in an app via

```js
// app/services/store.js
import Store from "ember-data/store";
import AdapterFindRelationshipHooks from "ember-data-adapter-find-relationship-hooks";

export default Store.extend(AdapterFindRelationshipHooks);
```

### RFC

Once the above 2 are done, the RFC only suggests to merge `ember-data-adapter-find-relationship-hooks` into core and deprecate the addon. If it is approved: great. If not: hey, we have a solution which can be used easily.

---

What do you think? Still motivated to help out?

Thanks again,
have a good one!

larsklevan:
@pangratz thanks for the feedback! I'll take a shot at it.

PR #5187, opened 2017-09-23
Use let->var

### LINKED ISSUE #5186: FastBoot tests breaking on ember-data canary
Seems like this [PR](https://github.com/emberjs/data/pull/5114) introduces usage of `let` which is only allowed in strict mode: https://travis-ci.org/ember-fastboot/ember-cli-fastboot/jobs/279033050 . This is breaking fastboot. 

cc: @runspired

kratiahuja:
Fixes #5186 

Is there any reason why ember-data is running tests against Node 7 only? If we run on Node 4, this issue would have been caught. Should I go ahead and update travis?

cc: @runspired @stefanpenner

### DIFF
--- a/lib/transforms/babel-plugin-remove-imports.js
+++ b/lib/transforms/babel-plugin-remove-imports.js
@@ -1,3 +1,5 @@
+'use strict';
+
 function PluginRemoveFilteredImports() {
   var importDeclarationsToRemove;
   var filteredImports;


rwjblue:
Since we already require Node 4, I think the better solution here would be to add `'use strict';` to the top of the file.

kratiahuja:
@rwjblue just switched to `'use strict';` . Totally forgot to do that!

kellyselden:
Duplicate of #5176.

kratiahuja:
Sorry didn't realize there was an open PR. I'll go ahead and close this.

PR #5199, opened 2017-10-03
Support for multiple records on top of same data

dnachev:
Right now, Ember Data has one-to-one mapping between an InternalModel and the actual `DS.Model`, which the application uses. There are two use cases, where this has is a drawback:
- GraphQL fragments and any other technology, which allows projection over the data
- The `ember-data-model-fragments`, which uses `_fragments` property to store the extra models, which have been created for a given internal model

If we refactor InternalModel to support multiple records than these use cases can use semi-public API to build upon instead of monkey patching Ember Data.

I'm looking for any comments on the feasibility of this refactor and whether it is something, we would like to have in Ember Data.

Note: The PR is not complete as there are:
- [ ] more tests to be written
- [ ] figure out what to do with the triggers
- [ ] fix the firing of `unloadRecord`
- [ ] put it behind a feature flag

### DIFF
--- a/addon/-private/system/internal-model-map.js
+++ b/addon/-private/system/internal-model-map.js
@@ -103,7 +103,7 @@ export default class InternalModelMap {
 
     for (let i = 0; i < models.length; i++) {
       let model = models[i];
-      model.unloadRecord();
+      model.unloadRecords();
     }
 
     this._metadata = null;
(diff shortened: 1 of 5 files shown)


locks:
This sounds like it needs an RFC, @dnachev?

dnachev:
Closing this PR in favor of implementing a solution in an external addon.

PR #5206, opened 2017-10-04
[BUGFIX beta] Fix broccoli-babel-transpiler cache warnings

xomaczar:
Latest ember-data@2.16 was still having "opting out of cache" warnings.
cc @rwjblue


bmac:
Thanks @xomaczar.

bmac:
Released as Ember Data 2.16.1.

PR #5213, opened 2017-10-06
[BUGFIX beta] proxy `meta` on PromiseArray

### LINKED ISSUE #5129: Meta data not proxied
If I create a normal promise proxy object, `{{myPromiseProxyObject.meta.foo}}` will proxy it's `content` value. But with an ED record array this doesn't work. The corresponding value in `content.meta` is present however.. leading me to expect this to work.

https://ember-twiddle.com/0396aa4bca83a7011625826f8143674b?openFiles=templates.application.hbs%2C

Noticed because my pagination computed properties were not re-computing when meta data was clearly changing.

pangratz:
This closes #5129.

I am unsure if this is the correct way to fix this or if something else is fishy...

Maybe it's because of this 🤔:

https://github.com/emberjs/data/blob/5fceead8acb53723bdf38daabad00aff093f390a/addon/-private/system/record-arrays/adapter-populated-record-array.js#L51-L52

### DIFF
--- a/addon/-private/system/promise-proxies.js
+++ b/addon/-private/system/promise-proxies.js
@@ -2,6 +2,7 @@ import Ember from 'ember';
 import { assert } from '@ember/debug';
 
 const { get , RSVP: { Promise }} = Ember;
+const { computed: { reads } } = Ember;
 
 /**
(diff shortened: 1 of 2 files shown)


bmac:
Thanks @pangratz.

Issue #5207, opened 2017-10-05
RecordReference & BelongsToReference value() is inconsistent


xomaczar:
If the record does not exist in the store, we get the following:
```
let recordReference = this.store.getReference('book', '1');
recordReference.value() === undefined
```
while belongsToReference
```
let author = this.store.peekRecord('author', 1);
author.belongsTo('book').value() === null
```
The docs treat these too cases the same

pangratz:
Good catch! This seems like a bug since the documentation mentions it should return `null`:

https://github.com/emberjs/data/blob/5fceead8acb53723bdf38daabad00aff093f390a/addon/-private/system/store.js#L981-L984

Wanna submit a bugfix PR?

xomaczar:
Sure - so null everywhere?

pangratz:
Yes. Have you found other `undefined` values for `value()`?

PR #5218, opened 2017-10-10
Update assert against nulls

pangratz:
Does this make sense? Basically this is to prevent us from future :trollface:ing like reported in #5207. I am unsure if this is a good idea or if we should update the tests to use `assert.ok` / `assert.notOk` instead 🤔 . Basically just did a search and replace...

### DIFF
--- a/tests/integration/store/query-test.js
+++ b/tests/integration/store/query-test.js
@@ -39,7 +39,7 @@ test("meta is proxied correctly on the PromiseArray", function(assert) {
     result = store.query('person', {});
   });
 
-  assert.equal(result.get('meta.foo'), undefined);
+  assert.notOk(result.get('meta.foo'), 'precond: meta is not yet set');
 
   run(function() {
     defered.resolve({ data: [], meta: { foo: 'bar' } });
--- a/tests/integration/adapter/json-api-adapter-test.js
+++ b/tests/integration/adapter/json-api-adapter-test.js
@@ -968,7 +968,7 @@ test('fetching a belongsTo relationship link that returns null', function(assert
       return post.get('author');
     }).then(author => {
       assert.equal(passedUrl[1], 'http://example.com/post/1/author');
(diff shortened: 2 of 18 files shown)


rwjblue:
```
950) [PhantomJS 2.1] integration/store/query: meta is proxied correctly on the PromiseArray
```

^ is failing the build on Ember 1.13 and 2.4 (but passes on 2.8+).

pangratz:
Good catch @rwjblue! Seems like this is failing since #5213. Any quick idea on what's wrong with that PR and earlier Ember versions?

pangratz:
I can't reproduce locally via `ember try:one ember-1-13 --- ember test` 😔

bmac:
https://github.com/emberjs/data/pull/5213 looked green on travis when I merged it...

tests/integration/store/query-test.js
```diff
@@ -39,7 +39,7 @@ test("meta is proxied correctly on the PromiseArray", function(assert) {
     result = store.query('person', {});
   });
 
-  assert.equal(result.get('meta.foo'), undefined);
+  assert.strictEqual(result.get('meta.foo'), undefined);
```
bmac: @pangratz I spent some time digging into this and couldn't figure out while its failing. I blame phantom weirdness. Lets leave this as `assert.equal` so the tests pass since the fact that this returns undefined is Ember behavior and not Ember Data.

bmac: Correction, I was just able to reproduce it. Turns out in Ember 1.13 `Ember.get({foo: null}, 'foo.bar')` returns `null` but in Ember 2.0 `Ember.get({foo: null}, 'foo.bar')` returns `undefined`.

pangratz: Uff, thanks for tracking that one down!

I've updated this assertion to be `assert.notOk`.

PR #5221, opened 2017-10-11
Fix docs link in DS.Serializer

luketheobscure:
When rendered, this links to https://www.emberjs.com/api/ember-data/2.15/classes/DS.JSONSerializer.html - which 404's. Correct link doesn't have the `.html`.


bmac:
Thanks @luketheobscure.

PR #5223, opened 2017-10-12
[BUGFIX] Cleanup dependencies.

rwjblue:
The first commit removes some test only dependencies, and the second does some general cleanup/updating of the packages deps.


rwjblue:
The failure in Travis was the one being fixed in #5224

rwjblue:
Rebased on top of #5224.

Turbo87:
@rwjblue looks like CI is still falling with a broccoli error 🤔

since both commits target different branches should we split the PR up?

rwjblue:
Ya, good point (will help us land this quicker). I pared this one down to just the test dependency changes.

PR #5225, opened 2017-10-12
Remove blueprints for Mocha < 0.12.

rwjblue:
ember-cli-mocha@0.12.0 was released quite some time ago (on 2016-11-23), and by all accounts has been well adopted. Supporting both mocha test blueprints indefinitely is untenable, and now seems like as good a time as any to drop the blueprints for the older syntax.

Note: this does *not* remove support for anything, it merely makes it slightly less convienient to create new test files for adapter/model/serializer/transform objects when your application is using Mocha < 0.12.

/cc @turbo87


rwjblue:
Yep, that was exactly my thinking.

bmac:
Thanks @rwjblue and @Turbo87.

Issue #5211, opened 2017-10-06
A bidirectional relationship `parent.children` becomes `hasLoaded: true` when a child that references the parent is pushed to the store


lolmaus:
## Affected versions

Ember Data 2.14+. This includes current stable Ember Data 2.16.1.

## Steps to reproduce

Given two models:

```js
// parent
export default Model.extend({
  children: hasMany('child')
});
```

```js
// child
export default Model.extend({
  parent: belongsTo('parent')
});
```

Push two records into the store:

```js
    const parent = store.push({
      data: {
        id: '1',
        type: 'parent',
        
        relationships: {
          children: {
            links: {
              "self": "https://example.com/parents/1/relationships/children",
              "related": "https://example.com/parents/1/children"
            }
          }
        }
      }
    })

    const child = store.push({
      data: {
        id: '1',
        type: 'child',
        relationships: {
          parent: {
            data: {
              id: '1',
              type: 'parent'
            }
          }
        }
      }
    })
```

## Expected behavior (Ember Data 2.13 and below work correctly)

The `parent.children` relationship is not considered loaded:

```js
parent.hasMany('children').hasManyRelationship.hasLoaded // => false
```

When you do `parent.get('children')`, a network request is made.

## Actual behavior (Ember Data 2.14 and above)

The `parent.children` relationship is considered loaded:

```js
parent.hasMany('children').hasManyRelationship.hasLoaded // => true
```

When you do `parent.get('children')`, a network request is *not* made, which breaks our app in many ways.

## Reproduction

https://ember-twiddle.com/3331c5dfc361920818fa3325842f5b3e?openFiles=controllers.application.js%2C (see console output, then try switching to `2.12.0` in `twiddle.json`)

## Credit

This bug has been discovered by @vvainio.

lolmaus:
This happens not only after a push. You can `findRecord` or `query` for a child, and if it references a parent, the parent will start thinking that it has only one child.

lolmaus:
This one seems to be either related or duplicate: #5209.

vvainio:
Here's another Twiddle demonstrating this issue. In our use-case, we are fetching some data in the route's model hook, and then side loading the rest. This pattern has worked at least up until `2.12.2`.

https://ember-twiddle.com/45d37cc9b2525b32f3cd6dbda1c96ecf?fileTreeShown=false&numColumns=2&openFiles=routes.application.js%2Ctemplates.application.hbs

BryanCrotaz:
[...]

Issue #5209, opened 2017-10-05
Links related relationship on modelA nulls belongsTo on modelB


DingoEatingFuzz:
### Data Model

1. modelA has a `hasMany('modelB')`
2. modelB has a `belongsTo('modelA')`
3. modelB payload looks like this: `model_a_id: 'model-a-1'`
4. modelA payload does not include the set of modelB ids. Instead, it is related through a related link: `/api/model-a/:id` returns the model and `/api/model-a/:id/model-b` returns the set of related modelB.

### Steps to reproduce

1. Find a modelB
2. Do not find the modelB's modelA
3. Find modelB's modelA independently
4. Observe that `modelB.get('modelA')` is a fulfilled promise with `null` for content

### Expected result

modelB should maintain its relationship with modelA despite modelA being loaded with unknown IDs for its hasMany modelB.

### Actual result

The relationship is null and the original ID from the payload is now lost.

### Notes

This _only_ happens in this order. If modelB's modelA is already loaded, it seems to be fine.

pangratz:
Can you update [this ember-twiddle](https://ember-twiddle.com/0e1a24aabb8fa7c1fdd8/copy?fileTreeShown=false&numColumns=2&openFiles=routes.application.js%2Ctemplates.application.hbs) so it demonstrates the issue you're having? That would help greatly track down the issue your having! Thanks!

lolmaus:
I've filed a very similar issue: #5211, it's either related or duplicated. I've provided a reproduction twiddle.

williamhector:
Here is a twiddle that I believe is the same issue as the OP:
https://ember-twiddle.com/2cf56ec2cd082d5c5a109e51993de730?openFiles=adapters.organisation.js%2C

The application route loads a given user, loads a given organisation, then loads the user's organisation (should be the same as the given org).
The organisation's users relationship (adminUsers) is never loaded, so while it might be related, I believe it's not a duplicate.

If you omit the "links" from the payload in `adapters/organisation.js`, the issue doesn't occur.
Or, if you call `user.get('organisation')` before `store.findRecord('organisation', id)` in `routes/application.js` then the issue doesn't occur.

BryanCrotaz:
I can confirm this behaviour in a real world app

BryanCrotaz:
I'm working on a PR

BryanCrotaz:
I have a failing test based on @williamhector 's twiddle...

BryanCrotaz:
[...]

PR #5230, opened 2017-10-15
[BUGFIX] enable lazy-relationship payloads to work with polymorphic relationships

### LINKED ISSUE #5037: store.push can leave belongsTo content as null
push a jsonapi blob with:

```
... relationships: {
  parent:
    data: {
      id: something,
      type: my-type
    }
}
```

obj.get('parent') === null

### LINKED ISSUE #5055: pushing polymorphic records to store only keeps last subtype from hasMany inverse
This is alittle bit hard to explain but the twiddle shows it.

If you have a model A that hasMany of a polymorphic type B 
( and lets say we have subtypes B1, B2 ). 

When you push polymorphic record B1 that is associated to A, 
then this first record B1 is correctly added to the A's hasMany list. 
If you push another B1 type record then that is added correctly. 
As soon as you switch B types to, let's say B2, then B2 is added but the previous B1's are removed, 
 and now the hasMany will only accept B2 subtypes.

This twiddle shows the issue:

https://ember-twiddle.com/df81a058c787d97db31366f939a01591?openFiles=routes.application.js%2C

 So, as long as you push the same polymorphic type, everything works fine.
 If you push a different polymorphic record than the first, it clears the previous subtypes in the hasMany array and adds new one, and will now ONLY allow this new type in the array.
 If you use createRecord everything works fine.
 
Looks like the reason createRecord works is because it takes a different path for adding it's data and relationships than push does. Push is not initializing the relationship right away ( waits for you to  do a call like B1.get(A-relationship). 
That means that if you say A.get(B-relationship) will always be not quite ready ( with push ). Seems like there is over optimization going on because the behaviour should be same for create or push ... I reckon.

runspired:
*Once reviewed/approved, this should be backported to 2.14, 2.15, and 2.16 beta.*

Resolves #5037 #5055 #5208 #5209 #5211 #5267 #5268 

**Problem**

Ember Data 2.14 introduced lazy-relationship parsing; however, this parsing used left-side/right-side keying that was incompatible with polymorphic relationships.  This resulted in payloads for newly encountered polymorphic types causing all previous payloads to be dereferenced.

In addition, a number of bugs were discovered where we were incorrectly setting up `RelationshipPayloads` as if there was an inverse when there was none, or where the inverse resolved to the key `:`.  This PR fixes that from the perspective of RelationshipPayloads but part of the problem was a bug in our calculation of `inverseFor` which is resolved with https://github.com/emberjs/data/pull/5261

**Solution**

This PR resolves these issues with three primary changes:

1) `relationship-info` is now [calculated and stored separately](https://github.com/emberjs/data/pull/5230/files#diff-1c53e096d92ba3a647478f3f51185a01R182) from the `RelationshipPayloads` objects.  This not only should be a performance win for us by enabling us to avoid calculating out the info in several scenarios where we would have needed to repeatedly redo the work before, but helps to resolve lookup for scenarios where there is no inverse or the inverse is polymorphic.  Additionally it cleans up many areas of the lazy-relationship code substantially to have this separation of concerns.

2) `RelationshipPayloads` are now always [cached on the manager by the resolved `base-model-name` instead of `model-name`](https://github.com/emberjs/data/pull/5230/files#diff-1c53e096d92ba3a647478f3f51185a01R173).  This enables us to confidently always utilize the correct `RelationshipPayloads` even when encountering a polymorphic type.

3) The payloads held for a relationship in `RelationshipPayloads` are now stored in [a two-level cache by `model-name` and `id`](https://github.com/emberjs/data/pull/5230/files?w=1#diff-ce1590fc9040b7fec82c5c59304875adR82) instead of a single-level cache of just `id`. This enables situations in which polymorphic siblings do not share a common unique-id constraint to work properly.

**Tests**

- one-side polymorphic belongsTo / hasMany
- both-sides polymorphic belongsTo / hasMany
- multi-table polymorphism (e.g. when IDs are not unique across all polymorphic types)
- reflexive polymorphic relationships
- self-referential non-reflexive polymorphic relationships


addon/-private/system/relationships/relationship-payloads.js
```diff
@@ -44,10 +44,32 @@ export default class RelationshipPayloads {
...
+
     this._lhsModelName = modelName;
     this._lhsRelationshipName = relationshipName;
     this._lhsRelationshipMeta = relationshipMeta;
 
+    this._rhsIsPolymorphic = inverseRelationshipMeta !== null &&
+      inverseRelationshipMeta.options !== undefined &&
+      inverseRelationshipMeta.options.polymorphic === true;
+
+    assert(`Both side of a relationship cannot be polymorphic`, !this._lhsIsPolymorphic || !this._rhsIsPolymorphic);
```
hjdivad: as we discussed out of band, it's fine for both sides to be polymorphic

addon/-private/system/relationships/relationship-payloads-manager.js
```diff
@@ -205,6 +209,22 @@ export default class RelationshipPayloadsManager {
     let lhsKey = `${modelName}:${relationshipName}`;
     let rhsKey = `${inverseModelName}:${inverseRelationshipName}`;
```
hjdivad: as discussed out of band, in both cases we want the key to be the model name defined by the inverse relationship, when an inverse relationship exists

addon/-private/system/relationships/relationship-payloads.js
```diff
@@ -44,10 +44,32 @@ export default class RelationshipPayloads {
   constructor(store, modelName, relationshipName, relationshipMeta, inverseModelName, inverseRelationshipName, inverseRelationshipMeta) {
```
hjdivad: maybe s/`modelName`/`baseModelName` or something here as we want this to always be the base model name

addon/-private/system/relationships/relationship-payloads.js
```diff
@@ -126,14 +159,28 @@ export default class RelationshipPayloads {
     }
   }
 
+  _isSide(potentialMatches, modelName, relationshipName) {
```
hjdivad: 👍

addon/-private/system/relationships/relationship-payloads-manager.js
hjdivad: so we should always create `RelationshipPayloads` with the base name and then add the actual model name

tests/unit/system/relationships/relationship-payload-manager-test.js
```diff
@@ -674,3 +678,72 @@ test('get can retrieve payloads with self-links in reflexive relationships', fun
   let entry = this.relationshipPayloadsManager.get('user', 1, 'friends');
   assert.deepEqual(entry, friendsPayload, 'self-link in reflexive relationship');
 });
+
+test('handles relationships where one side is polymorphic', function(assert) {
```
hjdivad: as discussed out of band, need more tests

- push subtype then push basetype
- push basetype then push subtype
- both sides poly

hjdivad:
@runspired possibly also #5208

tests/unit/system/relationships/relationship-payload-manager-test.js
danielspaniel: Any way I can keep this moving?  Is it possible for me to merge tests into this feature branch? Never tried that before

hjdivad: @danielspaniel you can open pull requests against runspired/data to branch `fix-polymporphic-payloads` and when they're merged it'll update this pr.  I've done it before and it works surprisingly well.

@danielspaniel if you have time to contribute tests for this that would be super helpful.  Thanks ^_^

danielspaniel: I have all tests ready but @runspired  added this to RelationshipPayloads constructor: 
```
assert(`Both side of a relationship cannot be polymorphic`, !this._lhsIsPolymorphic || !this._rhsIsPolymorphic);
```

That is showstopper, since previously in ember data you could do that .. so I think that assertion is perhaps guarding agains something I don't know what.  

But I removed that assertion and the test failed for both sides polymorphic, so maybe that had something to do with it .. not sure

danielspaniel: This feature ( lazy relationship loading ) is super complex and was implemented with very limited tests in the first place.  
 I am thinking it might be better to dump the feature,  rather than keep patching it.

  Was this your feature @hjdivad ?  Is there a great value to it.  If there is I am not seeing it. 
Would you be up for dumping it/ removing it?

danielspaniel: So, if a feature is not done properly and introduces very serious bugs,  are you telling me you are reluctant to back out the feature because you would lose the nice performance gains? 
 That makes no sense. 
First off, whoever wrote the feature should take responsibility and either fix it or remove it. 
I am willing to help out as much as I can, but the way it was done really looks like spaghetti to me, so I can't do much to save it. 
Secondly, if I wrote this feature, I would be embarrassed to keep it, if it broke ember-data. Does not really matter what kind of fun it offers if it is broken.  

How long should this broken feature sit in ember-data and just kill off polymorphic relationship users? 
Is it right to just wait around for some other work someone else is doing?  Who knows how long that other thing Igor is working on will be done.  Seems irrelevant to me, especially since he is not involved in this discussion much. It has only been you, me , Chris and Stephan ( saying he liked the original bug posting  )

runspired: @danielspaniel Sorry for being a bit quiet. Last week when @hjdivad and I discussed this PR OOB we discussed that assert.  Polymorphic to Polymorphic in ED has always been very buggy, and I wasn't sure it really worked at all and I thought it'd be too difficult to achieve here; however, in that conversation David and I also arrived at a solution for supporting it. So long as we always resolve to the baseModelName on both sides, the approach this PR outlines will work for both sides being polymorphic as well.  I will update the PR with that approach.

runspired: @danielspaniel can you PR the tests you've added to my branch?

danielspaniel: sure thing .. gimme a few moments

danielspaniel: I submitted PR to your branch of ED just now

runspired:
Updated to handle polymorphic-to-polymorphic relationships and with the help of @danielspaniel we now have more tests. I would like to clean up the testing a bit more but this is ready for another review pass by @hjdivad.

The only piece that concerns me is that we now MUST lookup the inverse greedily in a situation we did not need to before here: https://github.com/emberjs/data/pull/5230/files#diff-1c53e096d92ba3a647478f3f51185a01R178

I'm also sure now that we've got an implementation that covers the appropriate cases that this is due for some further refactoring and simplification.

BryanCrotaz:
Haven't looked at source in detail but have you tested with multiple levels of model extension?

E.g. A extends B extends C, with relationships introduced at each level?

runspired:
@BryanCrotaz those cases introduce no additional complexity or scenarios. A relationship must explicitly declare that it is polymorphic, and it must use the base type's type when doing so.

BryanCrotaz:
@runspired :) will try it out

danielspaniel:
This fix is still not complete enough for polymorphic relationships.  I found that  polymorphic self referential relationships  don't work  .. 2 options 

 - I can add the failing tests to this branch, @runspired can send more brainwaves at the problem  
 or 
 - we can just give up now, rip out this lazy relationships manager and save everyone who loves polymorphism   from dying an early death from alcoholism

runspired:
Another failing test please

danielspaniel:
I guess you beat me to it .. I was working on it but you added a whole ton of self-reflexive tests that were the ones that were breaking for me

danielspaniel:
actually check out my test .. it still fails  ( and I just rebased )

runspired:
I'm afk but reading those asserts I believe they assert the incorrect thing.

danielspaniel:
not sure why you think that .. but let me know when you get a chance.

runspired:
@danielspaniel ported and fixed up your test asserts, found the issue was with self-referential but non-reflexive relationships (as you setup).  Added fix.

danielspaniel:
Your getting ever so very close to fixing all this @runspired .. you are still hanging in there which is pretty awesome. 

ok .. how's about this for some philosophical ramblings. 

 lets say we have polymorphic hats and the data is like this: 

```javascript
  const bigHatData = { data: { id: 1, type: "big-hat", attributes: { type: "big-hat" }, relationships: { "user": { "data": { "type": "user", "id": "1" } } } } };
  const smallHatData = { data: { id: 1, type: "small-hat", attributes: { type: "small-hat" }, relationships: { "user": { "data": { "type": "user", "id": "1" } } } } };
```
 
 upon close inspection one sees that the two models have the same id ( 1 ) but they are different sub-types. 

 If you push the user and then these two hats .. the test to get the user hats fails ( only finds 1 .. not 2 ) 

 yours/our tests are doing this 

```javascript 
  const bigHatData = { data: { id: 1, type: "big-hat", attributes: { type: "big-hat" }, relationships: { "user": { "data": { "type": "user", "id": "1" } } } } };
  const smallHatData = { data: { id: 2, type: "small-hat", attributes: { type: "small-hat" }, relationships: { "user": { "data": { "type": "user", "id": "1" } } } } };
```
  using a different id for every subtype   and therefore they pass .. but I think that the first way should pass as well .. because it used to in the past and I think polymorphic models can have same id for different type 

 you want me to PR this failing test ?  

 or you want to give up now and get a free carton of fig newtons and whatever is behind curtain number 2

runspired:
This is a hard one because this case has actually never worked properly. It seems to work on the surface but has long plagued us as a source of bug reports.

danielspaniel:
I hear you, but I not sure what to tell the databases that are storing out the data this way .. since they going to laugh at me if I ask for different id's tomorrow

runspired:
@danielspaniel I would push some tests for this case, I will see what I can do but can't promise much atm.

runspired:
@danielspaniel just a thought, but worst case it would be pretty easy for you to rectify the ID conflicts client-side by normalizing them to `${modelName}-${id}` on the way in and then stripping the modelName while serializing on the way out.

danielspaniel:
I guess I could, but so would everyone else .. and currently without the lazy relationship thing,  if you just create stuff like that , it works .. since create bypasses the loading lazy thing-a-majigger.

runspired:
I think @igorT @stefanpenner and @hjdivad need to weigh in on our intent to support conflicting IDs on polymorphic type.

danielspaniel:
ok .. fair enough ..

danielspaniel:
just remember as you discuss the fate of the polymorphic universe that polymorphic types with same ids like 

 'big-hat' id: 1
 'small-hat' id: 1 

 Are supported in Ember Data and work fine ( I have been using polymorphic models like this for many years in ember-data and never had a problem. )  
 They work right now in ED 2.14+  but ONLY if you don't use lazy relationship thing-a-bob. 
 If you create the models with createRecord it works ( in other words )

runspired:
@danielspaniel my point was not that they don't sometimes work today, just that they don't consistently work and have been a source of many bug reports.  A very simple error case is the following:

```js
/*
  if the request returns an entry of type `hat`,
  we end up with two identity map entries and the
  next request for findRecord or a call to `peekRecord`
  will return an empty model.
*/
store.findRecord('big-hat', '1')
```

Additional bugs include any relationship that has both a `big-hat` and `small-hat` with the same ID will via most entry-points end up with only one of the two entries.

danielspaniel:
That particular example is interesting @runspired . maybe that case where they seek to find a subtype should assert an error if it returns a different subtype ( one possibility ) 

I think it is just going to be hard to drop supporting same id's in subtypes because polymorphism is either single table or multi table and multi table will have same id's. How can you tell people to change their database  tables in order to use ember data? 

Seems like the best thing would be to enforce proper usage as I was suggesting above. Though I don't know all the cases that are causing the current meltdowns. 

Maybe ember data has to change the way it does things ( holds it's data ) in the long run to avoid these kinds of bugs, but the issue will always be there since people will still be using multi table polymorphic records. 

Just dropping the support in lazy relationships and keeping it elsewhere ( createRecord ) seems like not the greatest idea, but I know full well the frustration you are facing in trying to fix up something for a feature that might change in the future.

runspired:
@danielspaniel if we agree that for now you can make the IDs not conflict fairly easily on the client-side but also that we will continue to work on a solution to make multi-table work, would that be acceptable? In addition to a general desire to have a good solution, I likely have a project coming up soon that will require work on ember-data in this area.

danielspaniel:
I can't do the client side id munging you suggested for my application, so I would vote that down. But it's also others as well who have the same problem. You would force all ember-data users to do some hack. 

 All just to save the lazy relationship thing?  Why are you trying so hard to save that feature?  I think you're not a quitter and admire you for it.  But removing that broken feature solves the problem immediately.  Then we can work on ember data for the future to shore up the polymorphic issues ( which I am both down and up for )

The lazy relationships feature just adds another point of failure to ember-data and might be better off if we sent it to a remote island to live out it's days surfing and getting drunk.

danielspaniel:
But maybe there is a compromise?  
 
If the lazy relationships could somehow be made optional than that would work as an interim solution. 
If there could be environment setting or a way to set that in an initializer ( set to run before ember-data fires its engines ) to say =>  lazy-relationships: false     I think everyone would be happy.

danielspaniel:
I am not usually one to get excited about anything except grass growing in my lawn, but thanks for starting the  WIP to get to a conclusion on being able to support multi table inheritance.  Very very exciting.  👯

danielspaniel:
This branch now completes the fixes required for it to work with polymorphic and regular models. I see no other issues and the tests have solid coverage.

danielspaniel:
Can I help in any way to resolve this issue ( sometime soon .. by end of year? )

runspired:
@danielspaniel we discussed on the weekly ember-data call today ensuring that this is properly reviewed and landed before we release the next major, and that we will attempt to backport all the way to when lazy-relationships first landed.  Effectively this means that we expect to land and release this within the next week.

danielspaniel:
super .. thanks for update

addon/-private/system/relationships/relationship-payloads-manager.js
```diff
@@ -172,7 +172,28 @@ export default class RelationshipPayloadsManager {
       return this._initializeRelationshipPayloads(modelName, relationshipName, modelClass, relationshipsByName);
     }
 
-    return this._cache[key];
+    let cache = this._cache[key];
+
+    if (cache === undefined) {
```
igorT: please add comments here as to what we are doing

addon/-private/system/relationships/relationship-payloads.js
```diff
@@ -172,14 +235,17 @@ export default class RelationshipPayloads {
       let idToPayloads;
       let inverseIdToPayloads;
       let inverseIsMany;
+
+      const payloadId = this.payloadId(modelName, id);
+
```
igorT: Can you key on the type id pair instead of just the id please

addon/-private/system/relationships/relationship-payloads.js
```diff
@@ -1,22 +1,56 @@
...
+    if (types[modelName] !== undefined) {
+      delete types[modelName][id];
+    }
+  }
+}
+
 /**
-  Manages the payloads for both sides of a single relationship, across all model
-  instances.
+ Manages the payloads for both sides of a single relationship, across all model
```
igorT: Are the whitespace changes on purpose?

BryanCrotaz:
Does this overlap with my PR #5268? It seems as though we're fixing the same problem in very different ways.

runspired:
@bryancrotaz 100% overlap

BryanCrotaz:
My solution seems simpler but then I think I must therefore be missing something...

runspired:
Check the various test cases added :P

addon/-private/system/relationships/relationship-payloads.js
```diff
@@ -79,71 +91,91 @@ export default class RelationshipPayloads {
...
-    @method
-  */
+   @method
+   */
   unload(modelName, id, relationshipName) {
     this._flushPending();
 
     if (this._isLHS(modelName, relationshipName)) {
-      delete this._lhsPayloads[id];
+      delete this.lhs_payloads.delete(modelName, id);
```
runspired: this is a mistake?

runspired: seems sketch this passes tests

addon/-private/system/relationships/relationship-payloads.js
```diff
@@ -79,71 +91,91 @@ export default class RelationshipPayloads {
...
-    @return {boolean} true iff `modelName` and `relationshipName` refer to the
-    left hand side of this relationship, as opposed to the right hand side.
+   @return {boolean} true iff `modelName` and `relationshipName` refer to the
+   left hand side of this relationship, as opposed to the right hand side.
 
-    @method
-  */
+   @method
+   */
   _isLHS(modelName, relationshipName) {
```
runspired: Can you please rewrite/add to the comments here

addon/-private/system/relationships/relationship-payloads.js
```diff
@@ -79,71 +91,91 @@ export default class RelationshipPayloads {
...
-    return modelName === this._lhsModelName && relationshipName === this._lhsRelationshipName;
+    let relInfo = this._relInfo;
+    let isSelfReferential = relInfo.isSelfReferential;
+    let isRelationship = relationshipName === relInfo.lhs_relationshipName;
+
+    if (isRelationship === true) {
+      return isSelfReferential === true || // itself
+        modelName === relInfo.lhs_baseModelName || // base or non-polymorphic
+        relInfo.lhs_modelNames.indexOf(modelName) !== -1; // polymorphic
```
runspired: please add a comment here

danielspaniel:
I guess this never made it to 2.18 .. any idea when this might get merged ?

cibernox:
I believe I'm hitting this. Is there any more work to be done?

BryanCrotaz:
Has this stalled? What’s left to do? Can we all help?

bmac:
I've back-ported this fix to `2.14.11`, `2.15.4`, `2.16.4`, `2.17.1`, `2.18.2`, and `3.0.2`.

danielspaniel:
This was a good weekend. Thanks for the merge and backports.

Issue #5233, opened 2017-10-17
Inconsistency with environment checks


IceDragon200:
Versions:

Ember 2.16.2
Ember Data 2.16.3

https://github.com/emberjs/data/blob/722e684943d13ad855aee5cc9023f6dae16946cc/lib/stripped-build-plugins.js#L62

https://github.com/emberjs/data/blob/7d1b25182a5fe67afa6e1103790e9b5147aafbb4/index.js#L32

In the stripped-build-plugins, a equality test is done, but in the index a regular expression (without bounds) is used.

This resulted in:

```
Uncaught Error: Could not find module `ember-data/-debug` imported from `ember-data/-private`
```

When trying to use an environment that had 'production' in it's name, but was not === 'production'.

Easiest way to break it:

```bash
ember build --environment=kinda_production
```

This issue can be resolved by changing:

```javascript
if (environment === 'production' || process.env.INSTRUMENT_HEIMDALL === 'true') { 
```

to 

```javascript
if (/production/.test(environment) || process.env.INSTRUMENT_HEIMDALL === 'true') { 
```

runspired:
@IceDragon200 a PR to fix this would be welcomed, you are correct we should support "production-like" environments throughout.

IceDragon200:
@runspired Created :)

rwjblue:
To be clear, ember-cli only supports three specific build environments: `test`, `development`, and `production`. Deployment environments are a very different thing (the ember-cli-deploy folks actually have a [nice section on their site](http://ember-cli-deploy.com/docs/v1.0.x/configuration/#additional-environments) about this exact thing).

pangratz:
Fixed and addressed by #5234 and #5239.

PR #5234, opened 2017-10-17
Check if environment is production-like in stripped-build-plugins

IceDragon200:
On request, a PR, to fix the production like environment issue

## Related

#5233

### DIFF
--- a/lib/stripped-build-plugins.js
+++ b/lib/stripped-build-plugins.js
@@ -59,7 +59,7 @@ module.exports = function(environment) {
     console.warn('NOT STRIPPING HEIMDALL');
   }
 
-  if (environment === 'production' || process.env.INSTRUMENT_HEIMDALL === 'true') {
+  if (/production/.test(environment) || process.env.INSTRUMENT_HEIMDALL === 'true') {
     postTransformPlugins.push([StripClassCallCheck]);
     uniqueAdd(filteredImports, 'ember-data/-debug', [
       'assertPolymorphicType'


rwjblue:
Seems good for us to be consistent here, but there are only three possible values for `process.env.EMBER_ENV` that are valid: `test`, `development`, and `production`.

I'd very slightly prefer to change the `index.js` check to be `=== 'production'` (instead of changing the other to use a regexp)...

runspired:
@rwjblue I agree that ember-cli doesn't support it in theory but I also know that historically we've loosely matched `production` and many teams have relied on that for configuring production-like staging environments (often began before ember-cli-deploy paved the path for nice things).

IceDragon200:
@rwjblue, @runspired Well prior to updating to 2.16 (coming from 2.9)  it was working fine with the production-like environments, it kinda threw a wrench in my day after the upgrade.

Maybe keeping it loose for now but adding a deprecation warning if a production-like environment is found:

```javascript
if (env !== 'production' && /production/.test(env)) {
 warn("production like environments are deprecated, if using ember-cli-deploy, please configure your build using 'production'");
}
```

This would ensure folks like me, who expected the old behaviour are given fair warning and a path for migration before completely flipping the table on them.

Sorry about the late reply, it was a tricky week

bmac:
@IceDragon200 that seems like a reasonable compromise. Do you need this change back-ported into 2.16?

IceDragon200:
@bmac Well I've already bitten the bullet and changed my code to use environment variables, so, not for me, but for anyone else that encounters this problem.

tl;dr Sure go ahead :)

runspired:
@IceDragon200 fwiw what I was saying was that as long as both checks match this should resolve your issue. Granted `===` means you won't get the code stripping in production-like environments, but it won't error either.  I agree with @bmac that a deprecation for production-like is reasonable.

bmac:
I added a pr for introducing the production like warning. https://github.com/emberjs/data/pull/5239

Issue #5219, opened 2017-10-10
inverse link payload is overriden when new data payload is pushed


sly7-7:
I find myself in a use case where a link hasMany relationship was not loaded (ie the link was not called. I've finally found the root cause, and wondered if it is expected.
In this line, the payload is totally overriden instead of merged if it existed before (which can be the case if it was a link).
https://github.com/emberjs/data/blob/1fe7e462546a13d9b9ce4b8f4dff388870d631aa/addon/-private/system/relationships/relationship-payloads.js#L296

cc/ @hjdivad

urbany:
Can you provide an example? I didn't understand the scenario.

ppcano:
@sly7-7 Based on your description, #5235 may be the same case.

sly7-7:
@ppcano Thank you, your test is indeed the same use case. I'm closing this one.

BryanCrotaz:
Fixed in #5257

PR #5235, opened 2017-10-19
Add a failing test for hasMany with a link

ppcano:
Tests for #5219 

> A hasMany relationship with a link will trigger the link request even if an inverse related object is pushed to the store

This was introduced at `2.14`


BryanCrotaz:
See if PR #5257 fixes this

ppcano:
@BryanCrotaz Awesome, it fixes it.

BryanCrotaz:
This test added to PR #5219

PR #5238, opened 2017-10-22
Move initialize-store-service.js out of the instance-initializers dir…

### LINKED ISSUE #5118: Can't load initializers in globals build
When I try to use canary ember-data in ember-twiddle I get an error.

```
Uncaught Error: Assertion Failed: An instance initializer cannot be registered without an initialize function
  assert @ ember.debug.js:5989
  assert @ ember.debug.js:17740
  (anonymous) @ ember.debug.js:5115
  registerInstanceInitializers @ ember-data.js:18728
  exports.default @ ember-data.js:18705
  (anonymous) @ ember-data.js:18741
  (anonymous) @ ember-data.js:18751
```

The problems seems to be rooted at this line, which I believe only runs in the globals build:

https://github.com/emberjs/data/blob/8e600fe3c4d295be3d80b75ea0dca793c8ed889d/lib/javascripts.js#L57

bmac:
…ectory.

That directory seems to be suggesting to ember-load-initializers that
it is a full initializer object and causing an exception to be thrown
in the globals build.

Fixes #5118 

@rwjblue would you mind reviewing this and letting me know if this is the right approach? I originally tried renaming `addon/instance-initializers/initialize-store-service.js` to `addon/instance-initializers/ember-data.js` export a full object and had `app/instance-initializers/ember-data.js` report that object (following the pattern used in the instance-initializers generated from the blueprints). However I ran into a BroccoliMergeTrees error and I didn't feel confident in my broccoli skills to debug the Ember Data's build atm.


rwjblue:
> I originally tried renaming addon/instance-initializers/initialize-store-service.js to addon/instance-initializers/ember-data.js export a full object and had app/instance-initializers/ember-data.js report that object (following the pattern used in the instance-initializers generated from the blueprints)

I _vaguely_ think that this is the approach I would take, but either is fine IMHO...

PR #5242, opened 2017-10-25
[BUGFIX beta] Normalize model names during `push`

hjdivad:
Model names are normalized (so eg `store.peekRecord('fooBar',1) === store.peekRecord('foo-bar', 1)`) at most of the store API (`peekRecord`,`findRecord` &c.) but notably absent is `push`.

This adds back model name normalization to push


hjdivad:
@bmac @fivetanley @runspired this is re the conversation we were having yesterday

hjdivad:
good point @rwjblue

PR #5243, opened 2017-10-26
[DOC] Fixup DS.Store.pushPayload example

sduquej:
The application serializer is used when a model name is not given. We already have one example where a model name is given and specify that the corresponding serializer will be used.


bmac:
Thanks @sduquej.

PR #4976, opened 2017-05-09
Add test for hasMany relationship

nadavshatz:
In relation to #4974

PR #5245, opened 2017-10-27
[DISCUSSION] Provide an alternative to using `hasMany.[]` to react to reference changes.

### LINKED ISSUE #4974: Computed `record.myHasMany.[]` doesn't fire if nothing accesses `record.myHasMany`
It looks like its no longer possible to listen on hasMany array length (`model.foos.[]`) and synchronously get the size of it.
```js
// works in ember-data 2.10 but not later
fooLength: Ember.computed('model.foos.[]', function() {
  return this.get('model').hasMany('foos').ids().length;
}),
```
I created [twiddle ](https://ember-twiddle.com/edbbdab5cf53bb07174058008be7cc03?openFiles=tests.unit.services.store-test.js%2C)- you can see that if you change ember-data to older version it works.

runspired:
- Adds test for and resolves #4974 which has been a bug since ember-data `2.11`
- Improves failing test from #4976 but shows that this is a different issue.

This PR contains the sad tale of how making the `manyArray` for a record lazy has combined with the nature of how computed properties and property chains observe values to mega-troll our end user's ability to introspect state.  The fix contained here is dirty, and we should consider alternatives.

## The Bug

Imagine that you want to observe changes to the references (`{ type, id }`) held by a `hasMany`, and react to them without also triggering a fetch of the relationship.  Ember offers a tool that allows for observing changes to an array, while ember-data offers a tool for fetching a relationship by reference to avoid fetching it. 

*Example model*

```js
export default Model.extend({
  others: hasMany('other-model', { async: false, inverse: 'mine' }),
  
  hasOthers: computed('others.[]', function() {
    return this.hasMany('others').ids().length > 0;
  }),
  
  othersCount: computed('others.[]', function() {
    return this.hasMany('others').ids().length;
  })
});
```

As additional records are pushed into the store changing the content of `others`, these computed properties will not trigger.

[Twiddle](https://ember-twiddle.com/4a2d2c9571d5f5f40a386ce6b0a324df?openFiles=models.my-model.js%2C).

## The Underlying Problem

Because `manyArray` is crated lazily, the observation chain for `others.[]` sees an `undefined` value until something calls `get('others')`.  Because of this, until something accesses the relationship, changes to the array go unnoticed because there is no array to watch.

It could be argued that a computed property with a dependency that goes un-fetched is a code smell.  However, we have not provided another means for easily working with relationship references without forcing the fetch of the relationship.

It could also be argued that working on data in the store by reference is a separate view of the world than working on data via the model layer, and thus combining these separate views in this way is simply not supported. However, this would leave users that need to react to changes by reference but do not desire to force the relationship to load with no good avenue forward.

## The Dirty Fix

We could make accessing `hasManyReference.prototype.ids()` have the side-effect of materializing the `manyArray`. This solves the general case, but increases the mental overhead and can create spooky-action at a distance.

## A Long Term Proposal

We should make working with relationships by reference a first class feature of ember-data.

### DIFF
--- a/addon/-private/system/references/has-many.js
+++ b/addon/-private/system/references/has-many.js
@@ -147,6 +147,7 @@ HasManyReference.prototype.link = function() {
    @return {Array} The ids in this has-many relationship
 */
 HasManyReference.prototype.ids = function() {
+  this.hasManyRelationship.manyArray;
   let members = this.hasManyRelationship.members.toArray();
 
   return members.map(function(internalModel) {
--- a/addon/-private/system/many-array.js
+++ b/addon/-private/system/many-array.js
@@ -143,7 +143,6 @@ export default EmberObject.extend(MutableArray, Evented, {
   },
 
   flushCanonical(isInitialized = true) {
-    debugger;
     // It’s possible the parent side of the relationship may have been unloaded by this point
     if (!_objectIsAlive(this)) {
       return;
--- a/addon/-private/system/many-array.js
+++ b/addon/-private/system/many-array.js
@@ -143,6 +143,7 @@ export default EmberObject.extend(MutableArray, Evented, {
   },
 
   flushCanonical(isInitialized = true) {
+    debugger;
     // It’s possible the parent side of the relationship may have been unloaded by this point
     if (!_objectIsAlive(this)) {
       return;
--- a/addon/-private/system/model/model.js
+++ b/addon/-private/system/model/model.js
@@ -1116,6 +1116,7 @@ const Model = EmberObject.extend(Evented, {
     //TODO(Igor): Consider whether we could do this only if the record state is unloaded
 
     //Goes away once hasMany is double promisified
+    // debugger;
     this.notifyPropertyChange(key);
   },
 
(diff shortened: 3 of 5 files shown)


addon/-private/system/references/has-many.js
```diff
@@ -147,6 +147,7 @@ HasManyReference.prototype.link = function() {
    @return {Array} The ids in this has-many relationship
 */
 HasManyReference.prototype.ids = function() {
+  this.hasManyRelationship.manyArray;
```
erichonkanen: just curious what the effect of this is? is that what forces it to recognize the related update?

runspired: `manyArray` is lazily created on access, by accessing it, `others.[]` now points at an array instead of undefined.

erichonkanen: ok I see nice

PoslinskiNet:
@runspired :red_circle:

jvenezia:
Would love to see this feature coming.

I'm doing that in the meantime...

```
  notes: DS.hasMany('note'),

  notesCount: computed('notes.[]', function () {
    this.hasMany('notes').hasManyRelationship.manyArray; // forces computed property to run after relationship metadata is loaded
    return this.hasMany('notes').ids().length;
  }),
```

runspired:
Closing as this was mostly intended for discussion.

Issue #5237, opened 2017-10-22
Support for natural key's


erichonkanen:
Hello,

We've run into an issue with a model that uses a natural key for the `id`.

The issue can be summed up as:

```If a record is created, then deleted, then a new record created with same natural key as previously deleted record, we get error.```

```Assertion Failed: 'natural-key-model' was saved to the server, but the response returned the new id 'foo|bar', which has already been used with another record.``` (ember-data 2.17.0-beta.1)

It seems like the ember-data store is retaining a reference to the supposedly deleted record which prevents a new one from being created.

It is explained more [here](https://discuss.emberjs.com/t/reusing-ids-for-soft-deleted-records/13715/3)

Does ember-data support natural key's or is it something ever talked about? Or, am I off and the issue from above link is unrelated to natural key?

Thanks

bmac:
Hi @erichonkanen. This seems like a bug and it is something Ember Data should support.

erichonkanen:
@bmac thanks for the input, I'm not entirely familiar with the intimate details of ember-data but do you think a work-around is possible? also where in the code would a new feature for this target? I'd be interested to get more familiar and possibly contribute

urbany:
This looks like a duplicate of the `unloadRecord` open issues.
@erichonkanen Here is a work around for now: https://github.com/emberjs/data/issues/4972#issuecomment-336662752

erichonkanen:
@urbany ok I'll take a look thanks... to be honest it's frustrating because I have two separate issues going on with ember-data and it has complicated things:

1) We cannot move beyond `2.12.x` due to [an issue with computed properties on related fields.](https://github.com/emberjs/data/issues/4974)

2) Because we are stuck at `2.12.x`, we can't properly address this issue around natural/recurring id as the codebase between edge and 2.12.x has changed and produces different errors... sigh

Nevertheless I'll take a look! thanks

erichonkanen:
was able to upgrade to ember-data `2.17.x` so will continue to look at this issue now that using latest

erichonkanen:
[...]

Issue #5136, opened 2017-08-17
Loading a previously unloaded record doesn't update relationships


workmanw:
Unfortunately I've hit another unloadRecord bug. When I unload a record that is on the `belongsTo` side of a `one-to-many` it is correctly removed from the `hasMany` array. However when I push that same record back into the store, it is not added back to the `hasMany` array.

Ember: 2.14.1
Ember-Data: 2.14.10

Reproduction repo: [workmanw/unload-relationships-bug](https://github.com/workmanw/unload-relationships-bug). 
Code:  [app/controllers/index.js](https://github.com/workmanw/unload-relationships-bug/blob/master/app/controllers/index.js). 
Live example: https://workmanw.github.io/unload-relationships-bug
Failing Test: #5137

To reproduce:
1) Load Record
2) Unload Record
3) Load Record again.

You'll notice that the "All Comments" section shows the record was reloaded (re-pushed) into the store, however the relationship on the `post` is not updated to reflect this.

One note here, I'm using `async: false` on this relationship. If I use `async: true` it creates a crash on unload.

workmanw:
Just a little more poking around. 

This functionality last worked correctly in `ember-data-2.12.2`. With `ember-data-2.13.0` it was not possible to unload records and update relationships. That was fixed in `ember-data-2.14.7` via #5097, however this issue was left lingering.

stefanpenner:
@workmanw thanks for the failing test

workmanw:
You're welcome! Let me know if there is anything else I can do to help.

stefanpenner:
@workmanw mind adding another test for:

>  If I use async: true it creates a crash on unload.

That way we can fix both these issues and ensure they don't come back

stefanpenner:
> Unfortunately I've hit another unloadRecord bug.

ooo my favorite ... :trollface:

workmanw:
@stefanpenner Sure thing, I'll add another test now.

workmanw:
@stefanpenner Actually, as I'm trying to write another test, I'm realizing that it may not be a valid use case. The reason is when you unload an async relationship, it still leaves the `hasMany` setup so that it can fetch it again if needed. Consider this (a little pseudo code):

```
let adam = store.peekRecord('person', 1);
adam.hasMany('boats').ids();   //  '[]'

store.push(/* a boat for adam */);
adam.hasMany('boats').ids();   // '[20]';

store.unloadRecord('boat', 20);
[...]

PR #5249, opened 2017-11-02
Fix client-side delete + resurrection

hjdivad:
Previously relationships were only cleared for new records.  Now they
are cleared for destroyed records as well, allowing destroyed records to
be pushed back into the store and properly update their inverse
relationships.

### DIFF
--- a/addon/-private/system/model/internal-model.js
+++ b/addon/-private/system/model/internal-model.js
@@ -772,7 +772,7 @@ export default class InternalModel {
     }
 
     if (this.isNew()) {
-      this.removeFromInverseRelationships(true);
+      this.removeFromInverseRelationships();
     }
 
     if (this.isValid()) {
@@ -891,19 +891,13 @@ export default class InternalModel {
 
    It will remove this record from any associated relationships.
 
-   If `isNew` is true (default false), it will also completely reset all
-    relationships to an empty state as well.
-
     @method removeFromInverseRelationships
-    @param {Boolean} isNew whether to unload from the `isNew` perspective
     @private
    */
-  removeFromInverseRelationships(isNew = false) {
+  removeFromInverseRelationships() {
     this._relationships.forEach((name, rel) => {
       rel.removeCompletelyFromInverse();
-      if (isNew === true) {
-        rel.clear();
-      }
+      rel.clear();
     });
 
     let implicitRelationships = this._implicitRelationships;
@@ -913,9 +907,7 @@ export default class InternalModel {
       let rel = implicitRelationships[key];
 
       rel.removeCompletelyFromInverse();
-      if (isNew === true) {
-        rel.clear();
-      }
+      rel.clear();
     });
   }
 
--- a/tests/integration/adapter/client-side-delete-test.js
+++ b/tests/integration/adapter/client-side-delete-test.js
@@ -1,0 +1,88 @@
+import { resolve } from 'rsvp';
+import { run } from '@ember/runloop';
+import setupStore from 'dummy/tests/helpers/store';
+
+import { module, test } from 'qunit';
+
+import DS from 'ember-data';
+
+module('integration/adapter/store-adapter - client-side delete', {
+  beforeEach() {
+    this.Bookstore = DS.Model.extend({
+      books: DS.hasMany('book', { async: false, inverse: 'bookstore' })
+    });
+    this.Book = DS.Model.extend({
+      bookstore: DS.belongsTo('bookstore', { inverse: 'books' })
+    });
+
+    this.env = setupStore({
+      bookstore: this.Bookstore,
+      book: this.Book
+    });
+    this.store = this.env.store;
+    this.adapter = this.env.adapter;
+  },
+
+  afterEach() {
+    run(this.env.container, 'destroy');
+  }
+});
+
+test('client-side deleted records can be added back from an inverse', function(assert) {
+  this.adapter.deleteRecord = function (store, modelClass, snapshot) {
+    if (snapshot.adapterOptions.clientSideDelete) {
+      return resolve();
+    }
+
+    assert.ok(false, 'unreachable');
+  }
+
+  run(() => this.store.push({
+    data: {
+      id: '1',
+      type: 'bookstore',
+      relationships: {
+        books: {
+          data: [{
+            id: '1', type: 'book'
+          }, {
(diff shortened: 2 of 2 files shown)


hjdivad:
#5136 / #5137 are expecting `unloadRecord` to client-side delete (ie remove themselves from non-unloaded inverse relationships).

This use case is supported per documentation via a client-side delete, ie
- delete the record
- have the adapter not send a request

However, some client-side delete use cases require resurrection (ie client-side deleting the record and then pushing the same record back into the store and repopulating inverses).

This behaviour was changed in #5058, which aimed to keep local state in relationships after record deletion

hjdivad:
cc @runspired @workmanw

hjdivad:
@workmanw i believe I've captured your use case in public api in the test, but have also verified in your repo https://github.com/workmanw/unload-relationships-bug

If you can confirm this solves your issue that would be great but in either case thanks for your diligence and patience.

stefanpenner:
Glad we can drop this hack.

runspired:
I suspect I didn't add an adequate enough test for why that extra guard was needed. I'll dig in this morning.

runspired:
After reviewing the tests +1 LGTM

hjdivad:
@runspired the reason this doesn't regress the local-changes preservation is because it's captured by [flushCanonical](https://github.com/emberjs/data/blob/master/addon/-private/system/relationships/state/relationship.js#L324-L331)

runspired:
@hjdivad records in the `isNew` state are covered, but not records not in that state. I do believe I have a test though that is passing anyway.

TL;DR what I'm saying is:

- `a hasMany B [b1]`
- push existing B (`b2`, not `isNew`) onto `a`
- now `a hasMany B [b1, b2]` with canonical state `[b1]` and current state `[b1, b2]`
- delete `b1`
- flushCanonical
- bug would be if the resulting state is `a hasMany B []` instead of `a hasMany B [b2]`

workmanw:
So using this PR with the suggested changes for "client side delete" and "unloading" does work around the issues of #5136. 

```javascript
comment.destroyRecord({ adapterOptions: { clientSideDelete: true } }).then(() => {
  comment.unloadRecord();
});
```

In my application, we have a base model, so I'd probably just create a utility function for this behavior. E.g. `comment.clientSideDelete()`. But that got me thinking, if this is the official work around to what was previous functionality, why not add this behavior in ember-data. It seems like that would ease the transition for many people who are currently stuck.

erichonkanen:
this may possibly fix the issue we've been having which I submitted here... https://github.com/emberjs/data/issues/5237

hjdivad:
@workmanw we're going to restore the previous semantics of `unloadRecord` for sync relationships.

Your suggestion of having some kind of "skip adapter" functionality is a good one; as it happens @stefanpenner was discussing the possibility of doing this as well for something that's common and currently not super ergonomic with the public API.

erichonkanen:
hey so I just installed this thinking it might address the issue of "natural key" which I submitted in #5237 but it didn't solve the problem...

[this is the assertion](https://github.com/emberjs/data/blob/ce606b1b6946a700248f92e35222cca195c8e80e/addon/-private/system/store.js#L1978) that is triggered when I try to create a record using a natural id which was previously used for a now destroyed record...

it seems like this is related but I did a quick run through and I still get the assertion. any feedback appreciated

erichonkanen:
@hjdivad any movement on this or thoughts around my previous comment?

hjdivad:
@erichonkanen The only thing I'm going to do before merging this pr is trace the delete with local additions case to confirm this is tested as expected.

I believe your issue #5237 has the same root cause as #5136 which is the next thing I'm aiming to address.

workmanw:
I was hoping to see this make it's way into 2.17, but admittedly that might have been a little ambitious. Can we set a goal of having this resolved in 2.18?

hjdivad:
Walked through this with @rwjblue and we've confirmed the regression that flag was meant to capture is indeed still tested and works.

Additionally, the case that @runspired [outlined](https://github.com/emberjs/data/pull/5249#issuecomment-341543678) also still works, but only because deletes won't trigger `flushCanonical`.

It remains the case that [ManyArray.flushCanonical](https://github.com/emberjs/data/blob/ed4e45f7ca4a28a478293857442164e295eccf5b/addon/-private/system/many-array.js#L145) discards local changes for non-new records, but that's not the issue that the `isNew` flag was meant to address.

hjdivad:
Failing test for the above issue: https://github.com/emberjs/data/pull/5269

workmanw:
@hjdivad et al, thanks a lot for all your hard work on this. It really is appreciated.

Hopefully I'm not pestering too much, but I'm curious what the next steps are toward resolving: #5136 . We had talked on slack and on https://github.com/workmanw/unload-relationships-bug/pull/1 about how this was the critical step to resolving these issues. I believe the net out of that convo was once this issue was resolved, then it could be worked around by introduced adding custom adapter options on the application side.

Example:

```javascript
// adapters/application.js
deleteRecord(store, type, snapshot) {
  let options = snapshot.adapterOptions || {};
  if (options.clientSideDelete) {
    return resolve();
  }
  return this._super(store, type, snapshot);
}
```
```javascript
// controllers/comment.js
actions: {
  unloadComment() {
    let comment = this.get('comment')
    if (comment) {
      comment.destroyRecord({ adapterOptions: { clientSideDelete: true } }).
        then(() => comment.unloadRecord());
    }
  }
}
```

I'm curious what the general thoughts are on the state of this issue. Will application owners need to make these changes to restore functionality, or will there be further improvement to make this easier? 

Thanks again!

erichonkanen:
Not to hijack but the issue above regarding loading/unloading is (I think) what is preventing us from moving past 2.12.x and is something we are really hoping gets resolved... our functionality works great on 2.12.x but there are other issues and obv we want to be on the latest

hjdivad:
> I'm curious what the general thoughts are on the state of this issue. Will application owners need to make these changes to restore functionality, or will there be further improvement to make this easier?

@workmanw no, we're going to restore the client-side delete semantics of `unloadRecord` for sync inverse relationships.

@rwjblue and I had some WIP on this Friday.

workmanw:
@hjdivad @rwjblue Awesome. Thanks for the update and all your hard work on this stuff!

Issue #5254, opened 2017-11-09
Replace Ember.Map/OrderedSet with native Map/Set classes


Turbo87:
`ember-data` is currently using `Ember.Map` and `Ember.OrderedSet`, which are close approximations of the native `Map` and `Set` classes introduced in ES2015. It seems that `ember-data` is the only major user of these classes and Ember is not even using them itself anymore and would like to get rid of them.

Since 3.0 is around the corner it might make sense to reevaluate the usefulness of these polyfills as IE11 already has somewhat usable native `Map` and `Set` support (see http://kangax.github.io/compat-table/es6/#test-Map).

As some public methods of `ember-data` return instances of `Map` and `MapWithDefault` this would probably have to be considered a breaking change 🤔

rwjblue:
I submitted #5255 to migrate away from `Ember.Map` and `Ember.MapWithDefault` (to native `Map`), however after discussing with @hjdivad I believe that we will still have a need for `@ember/ordered-set` ([the addon](https://github.com/emberjs/ember-ordered-set/issues/11)) as a custom data store for the ember-data relationship data (e.g. an array like with fast membership lookup for large relationships).

Turbo87:
@rwjblue as mentioned in https://github.com/emberjs/ember-ordered-set/issues/14 it seems that the differences between `OrderedSet` and `Set` are quite small, what is the specific difference that makes it necessary to keep `OrderedSet` around?

hjdivad:
@Turbo87 https://github.com/emberjs/data/blob/master/addon/-private/system/ordered-set.js#L19

The order is not always insertion order.  When inserting into a `hasMany` it is possible to insert at an arbitrary index.  The relationships code currently assumes its set `members` has the same order as the many array.

Turbo87:
@hjdivad okay, so we are actually talking about `EmberData.OrderedSet` 😉

Turbo87:
this is somewhat important though because in that case the code for `OrderedSet` could just live inside of Ember Data instead of having a dedicated addon just for that class

rwjblue:
@Turbo87 - Confirm, but I was just going to use the addon simply because it was already done and tested (thanks!)...

PR #5255, opened 2017-11-09
[BUGFIX beta] Use native Map if present.

rwjblue:
When native `Map` is present, use it. Otherwise fallback to `Ember.Map`.

This positions ember-data for the eventual deprecation and dropping of `Ember.Map` and `Ember.MapWithDefault` in Ember itself (and using native classes should generally be better for us long term than shipping custom _slightly_ different versions).

Note there are a few specific differences between native `Map` (and the `MapWithDefault` implementation that extends from the native `Map` when possible) and Ember's `Ember.Map` that we should be aware of:

* `Ember.Map` has custom `copy` and `isEmpty` methods which are not present in native `Map`
* `Ember.Map` adds a static `create` method (which simply instantiates itself with `new Ember.Map()`)
* `Ember.Map` does not accept constructor arguments
* `Ember.Map` does not have:
  * `@@species`
  * `@@iterator`
  * `entries`
  * `values`

Addresses `Ember.Map` and `Ember.MapWithDefault` usages from #5254.


addon/-private/system/map-with-default.js
```diff
@@ -0,0 +1,60 @@
...
+        let defaultValue = this.defaultValue(key);
+        this.set(key, defaultValue);
+        return defaultValue;
+      }
+    }
+  }
+  ```
+*/
+
+export default class MapWithDefault {
```
hjdivad: Should we add `copy` and `isEmpty` to warn/error with a useful message for people who may be relying on this?

It seems unlikely, but on the other hand, `MapWithDefault` is returned by some public API

rwjblue:
I'm going to try and pick this back up today/tomorrow to cleanup and address the points @hjdivad mentioned above...

rwjblue:
OK, updated this to make it (AFAICT) completely backwards compatible. Also, documented inline to make it clear _why_ the somewhat weirdo approach is required.

@hjdivad / @bmac / et. al.

bmac:
Thanks @rwjblue.

PR #5256, opened 2017-11-12
fix relationships with links

BryanCrotaz:
Fix issue #5209 and #5211 

when one end of a relationship is marked `{inverse:null}` keep the two relationships separate in the cache and don't create the `":"` cache entry
minor optimisation too in `relationship-payloads.js`

In RelationshipPayloads, don't overwrite an inverse relationship if it's got links instead of data

Added ember-cli-babel polyfill to allow async/await in tests for clarity


BryanCrotaz:
Shame, but removing async/await fixed it

BryanCrotaz:
Can't get clean code, so creating a new PR #5257

PR #5257, opened 2017-11-12
[BUGFIX] Fix issues with links

BryanCrotaz:
Fix issue #5209 and #5211

when one end of a relationship is marked {inverse:null} keep the two relationships separate in the cache and don't create the ":" cache entry

minor optimisation in relationship-payloads.js - don't scan an array multiple times

In RelationshipPayloads, don't overwrite an inverse relationship if it's got links instead of data

Fixup nested relationships test which now needs to download its links

This should be backported to 2.14,15,16

### DIFF
--- a/addon/-private/system/relationships/relationship-payloads-manager.js
+++ b/addon/-private/system/relationships/relationship-payloads-manager.js
@@ -203,7 +203,7 @@ export default class RelationshipPayloadsManager {
     }
 
     let lhsKey = `${modelName}:${relationshipName}`;
-    let rhsKey = `${inverseModelName}:${inverseRelationshipName}`;
+    let rhsKey = inverseRelationshipMeta ? `${inverseModelName}:${inverseRelationshipName}` : null;
 
     // populate the cache for both sides of the relationship, as they both use
     // the same `RelationshipPayloads`.
@@ -211,16 +211,20 @@ export default class RelationshipPayloadsManager {
     // This works out better than creating a single common key, because to
     // compute that key we would need to do work to look up the inverse
     //
-    return this._cache[lhsKey] =
-      this._cache[rhsKey] =
-      new RelationshipPayloads(
-        this._store,
-        modelName,
-        relationshipName,
-        relationshipMeta,
-        inverseModelName,
-        inverseRelationshipName,
-        inverseRelationshipMeta
-      );
+    const newRelationship = new RelationshipPayloads(
+      this._store,
+      modelName,
+      relationshipName,
+      relationshipMeta,
+      inverseModelName,
+      inverseRelationshipName,
+      inverseRelationshipMeta
+    );
+
+    this._cache[lhsKey] = newRelationship;
+    if (rhsKey && !this._cache[rhsKey]) {
+      this._cache[rhsKey] = newRelationship;
+    }
+    return newRelationship;
   }
 }
--- a/tests/integration/relationships/json-api-links-test.js
+++ b/tests/integration/relationships/json-api-links-test.js
@@ -1,0 +1,166 @@
+import { run } from '@ember/runloop';
+import { Promise } from 'rsvp';
+import setupStore from 'dummy/tests/helpers/store';
+
+import {
+  reset as resetModelFactoryInjection
+} from 'dummy/tests/helpers/model-factory-injection';
+import { module, test } from 'qunit';
+
+import DS from 'ember-data';
+
+const { hasMany, belongsTo } = DS;
+
+let env, User, Organisation;
+
+module("integration/relationship/json-api-links Relationships loaded by links", {
+  beforeEach() {
+  },
+
+  afterEach() {
+    resetModelFactoryInjection();
+    run(env.container, 'destroy');
+  }
+});
+
+test("Loading link with inverse:null on other model caches the two ends separately", function (assert) {
+  User = DS.Model.extend({
+    organisation: belongsTo('organisation', { inverse: null })
+  });
+
+  Organisation = DS.Model.extend({
+    adminUsers: hasMany('user')
+  });
+
+  env = setupStore({
+    user: User,
+    organisation: Organisation
+  });
+
+  env.registry.optionsForType('serializer', { singleton: false });
+  env.registry.optionsForType('adapter', { singleton: false });
+
+  const store = env.store;
+
+  User = store.modelFor('user');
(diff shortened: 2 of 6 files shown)


ppcano:
This fixes #5235 (initially reported in #5219). This bug is blocking our ED upgrade.

@BryanCrotaz Thanks a lot; feel free to include the #5235 failing test.

sly7-7:
@bmac @hjdivad :shipit: ?

tests/integration/relationships/json-api-links-test.js
```diff
@@ -0,0 +1,166 @@
...
+              id,
+              type: 'user',
+              relationships: {
+                organisation: {
+                  data: { id: 1, type: 'organisation' }
+                }
+              }
+            }
+          })
+        }, 10);
```
bmac: I think we can skip the `new Promise`/`run.later(..., 10)` and simplify this method with just `return Promise.resolve(...)`
```js
return Promise.resolve({
        data: {
          id,
          type: 'user',
          relationships: {
            organisation: {
              data: { id: 1, type: 'organisation' }
            }
          }
        }
      });
```

tests/integration/relationships/json-api-links-test.js
```diff
@@ -0,0 +1,166 @@
...
+              relationships: {
+                'admin-users': {
+                  links: {
+                    related: '/org-admins'
+                  }
+                }
+              }
+            }
+          })
+        }, 10);
```
bmac: Same here

tests/integration/relationships/nested-relationship-test.js
```diff
@@ -35,6 +37,25 @@ module('integration/relationships/nested_relationships_test - Nested relationshi
...
+              resolve({
+                data: [{
+                  id: 1,
+                  type: 'kid'
+                }]
+              });
+            } else {
+              resolve({ data: [] });
+            }
+          }, 10);
```
bmac: Same story here. Returning `Promise.resolve(...)` in each branch of this conditional should allow us to remove the run.later.

bmac:
Overall this change looks good. I just had a few comments for some cleanup in the tests.

BryanCrotaz:
@bmac changes made

tests/integration/relationships/nested-relationship-test.js
```diff
@@ -35,6 +37,19 @@ module('integration/relationships/nested_relationships_test - Nested relationshi
...
+        let data = [];
+        if (url === '/middle-agers/1/kids') {
+          data.push({
+            id: 1,
+            type: 'kid'
+          });
+        }
+        return resolve({ data });
+      }
+    }));
```
pangratz: I am not sure about this approach for fixing the failing test, as it seems this introduces a change in the semantics: before no link was called, now we do when we get the `kids` relationship on line [138/153](https://github.com/emberjs/data/pull/5257/files#diff-e4e8cc9dae6301ff76229fabdb9e3434R153):

```js
assert.ok(middleAger.get('kids').includes(kid));
```

This essentially boils down to if a relationship should be considered loaded when link & data is provided 😔 .

For this PR it might be best to change the assertion so we check the content of the relationship in a side-effect free way:

```js
assert.ok(middleAger.hasMany('kids').value().includes(kid));
```

addon/-private/system/relationships/relationship-payloads-manager.js
```diff
@@ -203,24 +203,28 @@ export default class RelationshipPayloadsManager {
...
+      relationshipMeta,
+      inverseModelName,
+      inverseRelationshipName,
+      inverseRelationshipMeta
+    );
+
+    this._cache[lhsKey] = newRelationship;
+    if (rhsKey && !this._cache[rhsKey]) {
+      this._cache[rhsKey] = newRelationship;
+    }
```
pangratz: This makes the above comment outdated, as we are not always populating the cache with the _same_ payload. It might makes sense to update the comment to reflect that...

pangratz:
@hjdivad @runspired would you have time to take a look at these changes?

runspired:
@pangratz I don't think this is necessary if we fix the inverse bug but I'll reconfirm.

mattraydub:
Any idea why this hasn't been merged/accepted yet?

runspired:
Mostly it's the wrong fix.

mattraydub:
thanks @runspired . Is there a valid fix elsewhere? Or is it still being considered / worked on. I had cherry picked this commit into the production version of ember-data we are using, and its been working fine, however I'd prefer to stay in line with what is going to get released.

runspired:
@mattraydub yes, https://github.com/emberjs/data/pull/5230

bmac:
@BryanCrotaz has https://github.com/emberjs/data/pull/5230 resolved this issue?

BryanCrotaz:
In feature lockdown at the moment for a big complex release of 8 interrelated Ember apps, but when that horror show is complete I'll upgrade to #5230

pangratz:
@bmac AFAICT the fix in #5230 hasn't addressed the issues with `links`, which this PR is solving...

BryanCrotaz:
Haven’t hadn’t a chance to test it yet

runspired:
@pangratz there may be a separate issue but I dug in months ago and it most definitely did resolve this.

pangratz:
I ran the test from #5235 against latest `master` and it's still failing (because the relationship payload is overwritten [here](https://github.com/emberjs/data/blob/v3.0.1/addon/-private/system/relationships/relationship-payloads.js#L295-L301) and the `links` are thrown out).

runspired:
@pangratz that test is wrong. Neither side specifies inverse: null. Does it still fail if you fix the test? Overwriting in the non null case is the expected behavior.

runspired:
Also cc @hjdivad, I suspect we could do a nicer thing here and merge links so as to keep both unless links is explicitly overwritten by the API

pangratz:
@runspired uh la la. I can confirm that `inverse: null` fixes the failing test in #5235.

sly7-7:
It's not clear to me why we must add `inverse: null` in the test, could someone explain that to me please?

runspired:
@sly7-7 if you do not specify inverse null, then we auto-resolve the other side of the relationship to create a link between the two classes. This results in either side being able to set the state of the relationship, which the test is asking us to not do.

PR #5260, opened 2017-11-14
[doc] Update urlForFindRecord example

acorncom:
The argument order gets reversed internally, but that is currently left as an exercise to the reader to figure out ... 😛


pangratz:
Thanks @acorncom!

PR #5261, opened 2017-11-21
[BUGFIX] inverseFor should respect inverse: null

runspired:
While working on #5230 I discovered that our current resolution strategy for `inverseFor` would improperly allow relationships that specify `inverse: null` to still be matched.  This addresses that, adds a test, and fixes two tests that depended upon this bug.

### DIFF
--- a/addon/-private/system/model/model.js
+++ b/addon/-private/system/model/model.js
@@ -35,7 +35,7 @@ function findPossibleInverses(type, inverseType, name, relationshipsSoFar) {
   let relationships = relationshipMap.get(type.modelName).filter(relationship => {
     let optionsForRelationship = inverseType.metaForProperty(relationship.name).options;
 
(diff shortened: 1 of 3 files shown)


runspired:
This supplants https://github.com/emberjs/data/pull/5257

igorT:
:+1: on the code
Can you please refactor the tests so they are not intermingled with the json serializer assertions? Currently it's easy to remove the added test lines as superfluous and not see any changes

sly7-7:
Though I'm not sure, I think this fix a bug I still encounter in 2.18. Would it be possible to backport this in 2.18 ? For now I tried to update to 3.0.0 beta, and this release seems to finally fix all the problem I encountered after 2.12.2. Really great job, I love you guys.

aaronbhansen:
@runspired I was also wondering if this would be available to backport to 2.18? That would be great to have.

PR #5265, opened 2017-11-27
[FEATURE] Return pushPayload result by default

Exelord:
Hey, after almost 2 years, are we ready to move further with https://github.com/emberjs/data/pull/4110 ? :)

This PR removes `ds-pushpayload-return` feature flag and enables this behavior by default.

I would love to merge it, so if any further work is required just give me a sign.

### DIFF
--- a/FEATURES.md
+++ b/FEATURES.md
@@ -17,11 +17,6 @@ entry in `config/features.json`.
   `methodForRequest`, `urlForRequest`, `headersForRequest` and `bodyForRequest`
   in the `DS.RESTAdapter`.
 
-- `ds-pushpayload-return` [#4110](https://github.com/emberjs/data/pull/4110)
-
-  Enables `pushPayload` to return the model(s) that are created or
-  updated via the internal `store.push`.
-
 - `ds-payload-type-hooks` [#4318](https://github.com/emberjs/data/pull/4318)
 
   Adds two new hooks `modelNameFromPayloadType` and `payloadTypeFromModelName`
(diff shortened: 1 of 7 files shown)


workmanw:
As the original author of `ds-pushpayload-return` I'm :+1: on this. But I think about 8 months or so ago someone told me this feature was a no go because it restricted future performance improvements that could eventually be made to `pushPayload`. 

That said, hopefully a core team member can chime in and tell us definitively.

sly7-7:
@bmac Are there any discussions on this one ? I'm personnally 👍 but, not aware of the `restricted future performance improvements`thing...

bmac:
Its been a while but I believe the issue was there was a plan to defer the work of creating records until they were needed by the user. Returning them here would force them to always be created by push payload.

I think @igorT had some ideas about making introducing a new api to make this more granular so the user could be specific about if they wanted to "Just push these records into the store and update the source of truth" or "push these records into the store and give me something back".

Exelord:
OK... So we have to wait for the new API :/

sly7-7:
@Exelord For now a quite straightforward workaround could be: https://github.com/emberjs/data/pull/5384#issuecomment-374263288

Exelord:
It's cool but it still expect that I know what is in the response :/ which I dont, tahts why pushPayload was the only way here

runspired:
@Exelord actually it doesn't assume.  In fact quite the opposite. `store.pushPayload` assumed an _incredible_ amount about the payloads it was receiving, usually incorrectly. Users had to massage the payload into a format that was totally different format from what their API gave them and from what their serializer expected just to use the feature. Meanwhile the util method based over `normalizeResponse` give you the exact information you need when you need it for how to properly convert the data to json-api format.

Issue #5267, opened 2017-11-29
inverse relationships not set up when pushing jsonapi blob with polymorphic model


BryanCrotaz:
Single source of truth isn't working for polymorphic relationships.

Push a `hasMany` with a descendant model. You'd expect that the matching `belongsTo` on the descendant model points back to the parent model. Instead it's null.

This works if the child model is the type specified in the `hasMany`, but not if it's a descendant model.

```
store.push({
  data: { post with relationship to descendant comment1 },
  included: { descendant comment1 with missing post relationship }
})
```

```
post.get('comments') => [comment1]
comment1.get('post') => **null**
```

Workaround: add `post` relationship into `comment1` in jsonapi

Working on a PR with test

BryanCrotaz:
Test added in PR #5268

BryanCrotaz:
@stefanpenner I'm not sure of the best way to fix this. The cache is keyed on `baseClass:relationship` so the entry isn't being found when you search for `descendant:relationship`.

We could add multiple cache entries for each of the different descendants, but I'm not sure how to discover them.

BryanCrotaz:
I think I found a way... see PR for details

BryanCrotaz:
Probably this has no value given #5230, but worth comparing for method

bmac:
I believe this has been resolved by https://github.com/emberjs/data/pull/5230. Feel free to reopen if it is still an issue.

PR #5268, opened 2017-11-30
Fix missing inverse

BryanCrotaz:
Fix issue #5267 

Initially add test to show problem


BryanCrotaz:
@stefanpenner needs some input from you. Could you comment on the direction I've taken here, storing polymorphic relationships against the base class and not the descendant class? It fixes the problem I found but breaks a couple of tests. Before I go and find out why, I wanted to check that this is the right way to go.

BryanCrotaz:
Probably this has no value given #5230, but worth comparing for method

runspired:
Closing as #5230 has landed

PR #5269, opened 2017-12-01
[BUGFIX FEAT] Preserve local hasMany changes during flushCanonical

hjdivad:
resolves #5269 

When new data comes in via `store.push` for a many array, and
`ManyArray.prototype.flushCanonical` is invoked:
  - new, unpersisted records added to the relationship are correctly not discarded, but
  - persisted records added to the relationship are erroneously discarded
  - new, unpersisted removals from the relationship are erroneously discarded

paired with @rwjblue, fix and additional test by @runspired

### DIFF
--- a/addon/-private/system/diff-array.js
+++ b/addon/-private/system/diff-array.js
@@ -1,3 +1,4 @@
+import OrderedSet from './ordered-set';
 /**
   @namespace
   @method diffArray
@@ -11,7 +12,7 @@
       removedCount: <integer>       // 0 if no change
     }
 */
-export default function diffArray(oldArray, newArray) {
+export function diffArray(oldArray, newArray) {
   const oldLength = oldArray.length;
   const newLength = newArray.length;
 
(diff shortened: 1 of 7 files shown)


runspired:
Kicked this off, it's still failing. I'm wondering if it wouldn't be best with these "long term goal" tests to merge them and skip them. If testem would work with `Qunit.todo` correctly that would be ideal.

runspired:
This test may be fixed by #4852

addon/-private/system/diff-array.js
```diff
@@ -57,3 +58,95 @@ export default function diffArray(oldArray, newArray) {
...
+  if (array) {
+    for (let i=0, l=array.length; i<l; i++) {
+      set.add(array[i]);
+    }
+  }
+
+  return set;
+}
+
+export function computeChanges(oldArray, newArray, removalSet) {
```
hjdivad: let's document the semantics of this function

hjdivad: let's also add a test for non-contiguous changes

addon/-private/system/diff-array.js
```diff
@@ -57,3 +58,95 @@ export default function diffArray(oldArray, newArray) {
...
+          changeset.changes = true;
+          break;
+        }
+      }
+    }
+
+    return changeset;
+  }
+
+  let removedMembersSet = setForArray(oldArray.slice(diff.firstChangeIndex, diff.firstChangeIndex + diff.removedCount));
```
hjdivad: Is this actually the set of removed members or rather the range (as a set) that contains all removed members?

Since we scan the range it's not a correctness issue but it makes the code more confusing

addon/-private/system/diff-array.js
```diff
@@ -57,3 +58,95 @@ export default function diffArray(oldArray, newArray) {
...
+  let changeBlockSet = setForArray(newArray.slice(diff.firstChangeIndex, diff.firstChangeIndex + diff.addedCount));
```
hjdivad: What is the change block?  Isn't this the minimal contiguous range that includes all the added records?

I'm trying to see why the name isn't symmetrical with `removedMembersSet`

addon/-private/system/diff-array.js
hjdivad: speaking of which we should probably not have it in `diff-array` as it's semantics are more specific than array diffing

if we don't want to keep it in `relationships/state/has-many` maybe move it to something like `relationships/state/utils` and perhaps rename it to `computeHasManyChanges(previousCanonicalState, newCanonicalState, locallyRemoved)`

addon/-private/system/diff-array.js
```diff
@@ -57,3 +58,95 @@ export default function diffArray(oldArray, newArray) {
...
+
+  // remove members that were removed but not re-added
+  removedMembersSet.forEach(member => {
+    if (!changeBlockSet.has(member)) {
+      changeset.removals.push(member);
+    }
+  });
+
+  let flushCanonicalLater = false;
```
hjdivad: when is this var changed?

hjdivad:
> This test may be fixed by #4852

The PR is green; if we don't need the local changes we'll still want the tests i think

runspired:
Closing this in favor of a new attempt at a later point that takes into consideration `RecordData`.

runspired:
Actually, going to reopen this one and PR the `todo` tests while eliminating the attempted implementation.

runspired:
Closing since stale, likely mostly addressed with #7493

PR #5273, opened 2017-12-06
WIP client-side-delete semantics `unloadRecord`

### LINKED ISSUE #5136: Loading a previously unloaded record doesn't update relationships

hjdivad:
Fix #5136, #5137

### DIFF
--- a/addon/-private/system/relationships/state/belongs-to.js
+++ b/addon/-private/system/relationships/state/belongs-to.js
@@ -58,6 +58,7 @@ export default class BelongsToRelationship extends Relationship {
   }
 
   inverseDidDematerialize() {
+    super.inverseDidDematerialize(this.inverseInternalModel);
     this.notifyBelongsToChanged();
   }
 
@@ -72,6 +73,13 @@ export default class BelongsToRelationship extends Relationship {
       this.inverseInternalModel = null;
       this.notifyBelongsToChanged();
     }
+  }
+
+
+  removeCompletelyFromInverse() {
+    super.removeCompletelyFromInverse();
+
+    this.inverseInternalModel = null;
   }
 
   flushCanonical() {
@@ -115,6 +123,12 @@ export default class BelongsToRelationship extends Relationship {
     this.notifyBelongsToChanged();
   }
 
+  removeAllInternalModelsFromOwn() {
+    super.removeAllInternalModelsFromOwn();
+    this.inverseInternalModel = null;
+    this.notifyBelongsToChanged();
+  }
+
   notifyBelongsToChanged() {
     this.internalModel.notifyBelongsToChanged(this.key);
   }
@@ -123,6 +137,11 @@ export default class BelongsToRelationship extends Relationship {
     if (!this.canonicalMembers.has(internalModel)) { return;}
     this.canonicalState = null;
     super.removeCanonicalInternalModelFromOwn(internalModel);
+  }
+
+  removeAllCanonicalInternalModelsFromOwn() {
+    super.removeAllCanonicalInternalModelsFromOwn();
+    this.canonicalState = null;
   }
 
   findRecord() {
--- a/addon/-private/system/relationships/state/has-many.js
+++ b/addon/-private/system/relationships/state/has-many.js
@@ -11,7 +11,12 @@ export default class ManyRelationship extends Relationship {
     this.belongsToType = relationshipMeta.type;
     this.canonicalState = [];
     this.isPolymorphic = relationshipMeta.options.polymorphic;
+    // The ManyArray for this relationship
     this._manyArray = null;
+    // The previous ManyArray for this relationship.  It will be destroyed when
+    // we create a new many array, but in the interim it will be updated if
+    // inverse internal models are unloaded.
+    this._retainedManyArray = null;
     this.__loadingPromise = null;
   }
 
@@ -33,6 +38,8 @@ export default class ManyRelationship extends Relationship {
   }
 
   get manyArray() {
+    assert(`Error: relationship ${this.parentType}:${this.key} has both many array and retained many array`, this._manyArray === null || this._retainedManyArray === null);
+
     if (!this._manyArray) {
       this._manyArray = ManyArray.create({
         canonicalState: this.canonicalState,
@@ -43,7 +50,13 @@ export default class ManyRelationship extends Relationship {
         meta: this.meta,
         isPolymorphic: this.isPolymorphic
       });
+
+      if (this._retainedManyArray !== null) {
+        this._retainedManyArray.destroy();
+        this._retainedManyArray = null;
+      }
     }
+
     return this._manyArray;
   }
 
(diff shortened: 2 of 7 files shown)


hjdivad:
Follow-on for #5136 #5137

hjdivad:
cc @igorT

tests/integration/records/unload-test.js
```diff
@@ -1488,3 +1488,8 @@ test('many:many async unload', function (assert) {
     assert.equal(person1Friends.isDestroyed, true, 'previous ManyArray is destroyed in the runloop after refetching');
   });
 });
+
+test('1 sync : 1 async');
```
rwjblue: Fwiw I do t think QUnit supports this syntax (assuming you are intending to basically create a skip?)...

hjdivad: @rwjblue creating a todo, not a skip.

addon/-private/system/relationships/state/belongs-to.js
```diff
@@ -15,6 +15,20 @@ export default class BelongsToRelationship extends Relationship {
...
 
+  _inverseIsAsync() {
+    if (!this.inverseKey || !this.inverseInternalModel) {
+      return false;
+    }
+    return this.inverseInternalModel._relationships.get(this.inverseKey).isAsync;
+  }
+
+  _inverseIsSync() {
+    if (!this.inverseKey || !this.inverseInternalModel) {
```
igorT: the model == null case should probably be handled by the caller?

addon/-private/system/relationships/state/has-many.js
```diff
@@ -12,9 +12,17 @@ export default class ManyRelationship extends Relationship {
     this.canonicalState = [];
     this.isPolymorphic = relationshipMeta.options.polymorphic;
     this._manyArray = null;
+    this._retainedManyArray = null;
     this.__loadingPromise = null;
   }
 
+  _inverseIsAsync() {
```
igorT: This could live on the parent class

addon/-private/system/relationships/state/has-many.js
```diff
@@ -44,6 +52,10 @@ export default class ManyRelationship extends Relationship {
         isPolymorphic: this.isPolymorphic
       });
     }
+    if (this._retainedManyArray !== null) {
```
igorT: Should this ever happen if manyArray is not null?

addon/-private/system/relationships/state/has-many.js
```diff
@@ -78,10 +90,16 @@ export default class ManyRelationship extends Relationship {
...
-      this._manyArray = null;
+  inverseDidDematerialize(inverseInternalModel) {
+    if (this.isAsync) {
+      if (this._manyArray) {
+        this._retainedManyArray = this._manyArray;
+        this._manyArray = null;
+      }
+      this._removeInternalModelFromManyArray(this._retainedManyArray, inverseInternalModel);
+    } else {
+      this.removeInternalModelFromOwn(inverseInternalModel);
```
igorT: This seems to be common for both hasMany and belongsTo?

addon/-private/system/relationships/state/has-many.js
```diff
@@ -143,7 +167,26 @@ export default class ManyRelationship extends Relationship {
       return;
     }
     super.removeInternalModelFromOwn(internalModel, idx);
-    let manyArray = this.manyArray;
+    // TODO: explain this madness
```
igorT: ❤️

addon/-private/system/relationships/state/relationship.js
```diff
@@ -78,30 +78,27 @@ export default class Relationship {
     this.hasLoaded = false;
   }
 
-  get parentType() {
-    return this.internalModel.modelName;
+  _inverseIsAsync() {
+    // TODO: clean this up; it's for implicit relationships
+    return this.isAsync;
```
igorT: this seems incorrect?

addon/-private/system/relationships/state/has-many.js
```diff
@@ -143,7 +167,26 @@ export default class ManyRelationship extends Relationship {
...
+    this._removeInternalModelFromManyArray(this.manyArray, internalModel, idx);
+    this._removeInternalModelFromManyArray(this._retainedManyArray, internalModel, idx);
```
igorT: we should maybe do this in the async case?

addon/-private/system/relationships/state/has-many.js
igorT: lets add an assert

fivetanley:
@hjdivad I tried this out in our app. We're still stuck in this scenario:

```javascript
import Ember from 'ember';

  actions: {
    unload() {
      const clients = this.store.peekAll('client');

      // there is one record that matches this, so one record unloaded
      clients.filterBy('isCool').forEach((client) => {
        client.unloadRecord();
      });
      
      // here clients is [Model, Model, null]
      // null should not be in the array
      
      // error here because filterBy calls Ember.get on null to try to find hasDirtyAttributes.
      clients.filterBy('hasDirtyAttributes').forEach(client => {
        client.unloadRecord();
      });
    }
  }
});
```

It looks like some `null`s are still getting stuck in the RecordArray created for `this.store.peekAll('client')`.

hjdivad:
Thanks for the followup @fivetanley.  I'll see about reproducing this in a test if you haven't already.

addon/-private/system/relationships/state/has-many.js
hjdivad: we discussed this OOB, but LMK if you still think there's an issue here

hjdivad:
@igorT this should be good to merge; issues we discussed are handled.

hjdivad:
@fivetanley this doesn't resolve the filter issue you mentioned [in your previous comment](https://github.com/emberjs/data/pull/5273#issuecomment-356679354); that will be addressed in a follow-on issue.

sly7-7:
@fivetanley I wonder if in your use case, calling toArray() after filter would change anything ?

workmanw:
This isn't tagged with the release channel. Can we please get this in 2.18 LTS?

workmanw:
Shameless bump @hjdivad @rwjblue -- Can we get this PR and #5249 into 2.18.1? I know DS doesn't have the same level LTS support as Ember, but it would be nice to have a place to jump to in the 2.x series.

sly7-7:
Or at least in 3.0 beta ?

elgordino:
FYI This is included in 3.1.0-beta.1 https://github.com/emberjs/data/releases

bmac:
I just released 2.18.1 and 3.0.1 with this pr. Thanks to @sly7-7 for pinging me about including this pr.

workmanw:
@bmac Thank you!!

jlami:
@hjdivad What is the status of the followup on the problem @fivetanley was having?
I think this might be the same problem as I'm having. see also #5353 and #5354 

Could somebody explain the reasoning behind the distinction in handling of the canonical state of sync and async relationships for unload? I'm inclined the client side should not be effected by this, as this is more of a server behaviour in my mind.

Issue #5274, opened 2017-12-12
Fix fetch hasMany using links when parent & child records are already loaded


mattraydub:
Given a Parent model with "children" (hasMany) and Child model with "parent" (belongsTo) and proper inverses setup, when loading (store.findRecord) the Parent record, then loading one of the Child records (store.findRecord), then accessing parent.get('children') only returns the 1 previously fetch Child record instead of using the links, and fetching the entire relationship from the server.

Looks like the issue exists somewhere in the loading of the child payload. Since the payload includes a reference to its parent relationship, the store notices it already has that parent and updates the inverse relationship "children" with this 1 record in the `store._relationshipsPayloads`. So before fetching its children the parent already knows about this child in its relationshipsPayload. Not sure if this is necessary since it really needs to use the link to fetch the rest of the children from the server, in order to have the proper records loaded into that relationship.

Will be posting my breaking test shortly.

This is similar to #5209 but not exactly the same.

mattraydub:
This was fixed by PR #5257

jamesarosen:
@mattraydub you said this was fixed by #5257, but that hasn't been merged. Do you mean another PR? Or should we reopen this?

mattraydub:
This issue presents a problem with a breaking test. The PR #5257 fixes this problem, so I closed so that no one continues to try to fix it. Not sure if they will ever merge that PR, but I cherry-picked it into the version I am using. I guess we could reopen, but just wanted to explain my point-of-view.

psbanka:
@mattraydub - I would advocate for re-opening this ticket so that it has visibility to the greater team. Clearly we don't want to be running on a forked version of `ember-data`

PR #5275, opened 2017-12-12
breaking test for issue #5274

mattraydub:
see #5274 for explanation


mattraydub:
A different PR #5257 fixes the breaking test here

PR #5277, opened 2017-12-13
[BUGFIX] Fix store.createRecord with belongsTo when model has @each o…

mattraydub:
When passing belongsTo relationship during create (on Model with property using @each observer) the InternalModel.getRecord actually recurses on itself thus calling the modelFactory create 2 times. This is only happening when there is something watching a path on the created model using the @each observer.

This test and fix ensures the create is only called once.

### DIFF
--- a/addon/-private/system/model/internal-model.js
+++ b/addon/-private/system/model/internal-model.js
@@ -338,10 +338,6 @@ export default class InternalModel {
         adapterError: this.error
       };
 
-      if (typeof properties === 'object' && properties !== null) {
-        emberAssign(createOptions, properties);
-      }
-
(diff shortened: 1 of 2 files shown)


runspired:
Digging into this PR. I'm not familiar with the issue but seems like a thing we should fix if broken, and there's at least one other bug report I've seen float by lately that looked similar.

runspired:
This was fixed in #5369

runspired:
Just wanted to circle back and say thank you very much @mattraydub ! I find your failing tests super helpful :) We should rope you in to doing more of them :D

mattraydub:
@runspired glad I was able to help

PR #5280, opened 2017-12-15
[doc beta]document extractRelationship in json-api.js

toddjordan:
In response to https://github.com/ember-learn/ember-api-docs/issues/399

`extractRelationship` has different args for `JSONAPISerializer` than `JSONSerializer`.  Since `JSONAPISerializer.extractRelationship` is not documented the apidocs show `JSONSerializer.extractRelationship` instead, leading to confusion.  This pr simply adds documentation reflecting correct arguments to `extractRelationship`, and also adds `extractRelationships` documentation (so source link goes to the correct implementation.

#### Before:
image

#### After:
image

### DIFF
--- a/addon/serializers/json-api.js
+++ b/addon/serializers/json-api.js
@@ -301,6 +301,15 @@ const JSONAPISerializer = JSONSerializer.extend({
     return attributes;
   },
 
+  /**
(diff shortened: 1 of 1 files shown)


bmac:
Thanks @toddjordan

PR #5281, opened 2017-12-18
Fix the ember-try canary scenario by polyfilling WeakMap

bmac:
@rwjblue would you mind reviewing this and confirming that adding this to `ember-cli-build.js` will only impact the test builds and not the builds of consumers of Ember Data?


rwjblue:
Confirm, `ember-cli-build.js` in addons only affects the addon itself.

PR #5282, opened 2017-12-18
Add guard & more helpful error massage when calling _setupRelationships when store is destroyed

danwenzel:
This PR adds a guard against trying to set up relationships after the store has been destroyed, providing a more helpful error message.

#### Background:

We ran into a test failure which showed the following error:

```
TypeError: Cannot read property 'push' of null
    at Class._setupRelationshipsForModel
```

similar to this issue: https://github.com/samselikoff/ember-cli-mirage/issues/1220

After some digging, it turned out that the store had been destroyed too early in the test, and this line was causing the error: https://github.com/emberjs/data/blob/master/addon/-private/system/store.js#L2418

The error happened because the store sets `_pushedInternalModels` to `null` when destroyed: https://github.com/emberjs/data/blob/master/addon/-private/system/store.js#L2706, so trying to later call `push` on it causes an error.

This PR adds an assertion that makes it clearer what's going on, making for faster debugging.


danwenzel:
Not sure why the test is failing in Phantom...  It's passing just fine for me locally

addon/-private/system/store.js
```diff
@@ -2415,6 +2415,11 @@ Store = Service.extend({
       return;
     }
 
+    if (this.isDestroyed) {
+      assert('Attempting to set up relationships after store has been destroyed');
```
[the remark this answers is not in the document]
danwenzel: Done

danwenzel:
Alright, so I noticed that the failing test was in the production environment, so I added a conditional in the test file.  Let me know if there's a better way to do that.

tests/integration/store-test.js
```diff
@@ -202,12 +203,16 @@ test("destroying the store correctly cleans everything up", function(assert) {
...
 
   run(store, 'destroy');
-
   assert.equal(personWillDestroy.called.length, 1, 'expected person to have recieved willDestroy once');
   assert.equal(carWillDestroy.called.length, 1, 'expected car to recieve willDestroy once');
   assert.equal(carsWillDestroy.called.length, 1, 'expected person.cars to recieve willDestroy once');
   assert.equal(adapterPopulatedPeopleWillDestroy.called.length, 1, 'expected adapterPopulatedPeople to recieve willDestroy once');
   assert.equal(filterdPeopleWillDestroy.called.length, 1, 'expected filterdPeople.willDestroy to have been called once');
+  if (config.environment !== 'production') {
+    assert.throws(() => {
```
rwjblue: I think we generally use `assert.expectAssertion(() => {}, /some matcher/)` for this.  That custom assertion is provided by https://github.com/workmanw/ember-qunit-assert-helpers/ and is aware of development vs production...

danwenzel: @rwjblue - If I do that, then the production test still fails, because it moves on and hits the original `TypeError: Cannot read property 'push' of null`.

This is why I had originally built the assertion into a conditional.  It would allow an early return - in production, the assertion would be skipped, but you wouldn't hit that error.

danwenzel: Or another idea:  Maybe we want to just `throw` instead of `assert`?  i.e. maybe we _want_ this better error message in production too? This case is currently throwing an error in prod either way.

danwenzel: @rwjblue ⬆️  Any thoughts on this?

danwenzel: Pinging again... @rwjblue ⬆️   Any thoughts?

danwenzel: Thanks @runspired !  Any thoughts on the other issues above?  If we want to leave production alone, we'll have to think of some way to only call that method in the non-prod tests.

runspired: @danwenzel see https://github.com/emberjs/data/blob/master/tests/unit/store/push-test.js#L7

danwenzel: Thanks, @runspired , that worked perfectly!  All passing now

danwenzel:
@runspired @rwjblue - Ok to ship this?

tests/integration/store-test.js
```diff
@@ -209,6 +209,13 @@ test("destroying the store correctly cleans everything up", function(assert) {
   assert.equal(filterdPeopleWillDestroy.called.length, 1, 'expected filterdPeople.willDestroy to have been called once');
 });
 
+testInDebug("_setupRelationshipsForModel generates helpful error in non-prod.", function(assert) {
+  run(store, 'destroy');
+  assert.throws(() => {
+    store._setupRelationshipsForModel(null, { relationships: {} });
+  }, /Attempting to set up relationships after store has been destroyed/);
+});
```
runspired: We should design a test scenario similar to a real-world scenario that would have caused this to happen.  I'm especially interested in such a test because:

1. It may be that we need to adjust our waiters https://github.com/emberjs/data/pull/5422
2. _setupRelationshipsForModel should still occur in the same run-loop as whatever scheduled it. This indicates to me that this either might be the wrong location to assert, or that we could improve the message to tell people exactly what they need to do to resolve the error.

danwenzel: Hi @runspired.  I agree that we should eventually dig more into the root issue, but this PR at least make the situation 1 step better - Someone encountering this in dev or test will know more about what's going on and fix it faster.

I unfortunately won't have time to dig deeper for a while, so maybe we could ship this as-is?

runspired: @danwenzel if you can jot down some notes for me about roughly the setup you encountered in the wild, and give me commit access to your fork, I will continue to improve this :)

danwenzel: Hi @runspired , sorry for the delay.

I can't remember the details of my setup, but this gives a pretty good example of the issue: https://github.com/samselikoff/ember-cli-mirage/issues/1220.

I've given you access to my fork.  Let me know if you need anything else

runspired:
Will revisit this soon, I have a pretty good idea of how to design a good test for it, and likely we can simply do the "right" thing and no-op here vs. error. 

Basically, the waiters in #5545 and an assert about using `store.push` when the store is destroyed should be enough, any other async internal timing things should be on ember-data to cleanup.

runspired:
After revisiting I confirmed that there are only two ways for _setupRelationships to be called: via `commit` and via `push/_push`. #5545 has guarded and tested `commit`, I will follow up with a new PR that guards `push/_push`.

PR #5289, opened 2017-12-18
Remove support/deprecation for normalizeHash in 3.0.0

bmac:


addon/serializers/rest.js
```diff
@@ -161,13 +161,6 @@ const RESTSerializer = JSONSerializer.extend({
     @return {Object}
   */
   normalize(modelClass, resourceHash, prop) {
```
rwjblue: Should we remove this whole override instead? It seems like it only calls `this._super` after the changes, which might mean we can remove the whole function...

Issue #5292, opened 2017-12-19
QUnit Blueprints Quest Issue


alexander-alvarez:
__Background__
https://github.com/emberjs/ember.js/issues/15933

__Tasks__
- [x] (:lock: @alexander-alvarez) Extend [test-framework detector](https://github.com/emberjs/data/blob/master/blueprints/test-framework-detector.js) to use rfc-232 style generators where applicable.
- [x] Add `<type>-test`  "simplified QUnit" blueprints
  - [x] (:lock: @alexander-alvarez) adapter-test
  - [x] (:lock: @alexander-alvarez) model-test
  - [x] (:lock: @alexander-alvarez) serializer-test
  - [x] (:lock: @alexander-alvarez) transform-test
- [ ] Extract the above tests blueprints into a standalone polyfill addon (similar to https://github.com/alexander-alvarez/ember-qunit-nested-module-blueprints-polyfill but for ember-data)

---
To pick up one of these items, please comment below and we'll update the checklist with :lock: @your-name-here...

alexander-alvarez:
https://github.com/alexander-alvarez/data/compare/node-test-cleanups...alexander-alvarez:test-framework-detectors-fixup is ready to go once https://github.com/emberjs/data/pull/5293 is merged

rwjblue:
The first couple of items are landed. The others are up for grabs...

rwjblue:
@alexander-alvarez - I believe the main blueprint types have been done, and all thats left is the polyfill addon. Can you confirm?

alexander-alvarez:
@rwjblue yup :+1:

runspired:
@alexander-alvarez what remains here?

alexander-alvarez:
@runspired  Just the last item
> Extract the above tests blueprints into a standalone polyfill addon

rwjblue:
TBH, I think this is probably fine to leave off. At the time we needed a polyfill addon because 3.0 was a ways off, now its out (as is 3.1).

PR #5298, opened 2017-12-21
blueprints/serializer-test: Add RFC232 variants

alexander-alvarez:
See https://github.com/emberjs/data/issues/5292

Had questions on what we wanted the default to be so I implemented both 
@Turbo87 @bmac

### DIFF
--- a/blueprints/serializer-test/qunit-rfc-232-files/tests/unit/__path__/__test__.js
+++ b/blueprints/serializer-test/qunit-rfc-232-files/tests/unit/__path__/__test__.js
@@ -1,0 +1,24 @@
+import { module, test } from 'qunit';
+import { setupTest } from 'ember-qunit';
+import { run } from '@ember/runloop';
+
+module('serializer:<%= dasherizedModuleName %>', '<%= friendlyTestDescription %>', function(hooks) {
+  setupTest(hooks);
+
+  // Replace this with your real tests.
+  test('it exists', function(assert) {
+    let store = this.owner.lookup('service:store');
+    let serializer = this.owner.factoryFor('serializer:<%= dasherizedModuleName %>').create({ store });
+
+    assert.ok(serializer);
+  });
+
+  test('it serializes records', function(assert) {
+    let store = this.owner.lookup('service:store');
+    let record = run(() => store.createRecord('<%= dasherizedModuleName %>', {}));
+
+    let serializedRecord = record.serialize();
+
+    assert.ok(serializedRecord);
+  });
+});
--- a/blueprints/serializer-test/qunit-rfc-232-files/tests/unit/__path__/__test__.js
+++ b/blueprints/serializer-test/qunit-rfc-232-files/tests/unit/__path__/__test__.js
@@ -2,7 +2,7 @@ import { module, test } from 'qunit';
 import { setupTest } from 'ember-qunit';
 import { run } from '@ember/runloop';
 
-module('serializer:<%= dasherizedModuleName %>', '<%= friendlyTestDescription %>', function(hooks) {
+module('<%= friendlyTestDescription %>', function(hooks) {
   setupTest(hooks);
 
   // Replace this with your real tests.
--- a/blueprints/serializer-test/qunit-rfc-232-files/tests/unit/__path__/__test__.js
+++ b/blueprints/serializer-test/qunit-rfc-232-files/tests/unit/__path__/__test__.js
@@ -8,7 +8,7 @@ module('<%= friendlyTestDescription %>', function(hooks) {
   // Replace this with your real tests.
   test('it exists', function(assert) {
     let store = this.owner.lookup('service:store');
-    let serializer = store.serializerFor('serializer:<%= dasherizedModuleName %>');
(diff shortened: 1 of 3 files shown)


blueprints/serializer-test/qunit-rfc-232-files/tests/unit/__path__/__test__.js
```diff
@@ -0,0 +1,24 @@
+import { module, test } from 'qunit';
+import { setupTest } from 'ember-qunit';
+import { run } from '@ember/runloop';
+
+module('serializer:<%= dasherizedModuleName %>', '<%= friendlyTestDescription %>', function(hooks) {
```
rwjblue: Should only be two arguments here (`friendlyTestDescription` and the `function(hooks) {}`...

alexander-alvarez:
@rwjblue fixed... sorry about that... thoughts on any of the two testing styles here?

blueprints/serializer-test/qunit-rfc-232-files/tests/unit/__path__/__test__.js
```diff
@@ -0,0 +1,24 @@
...
+import { setupTest } from 'ember-qunit';
+import { run } from '@ember/runloop';
+
+module('<%= friendlyTestDescription %>', function(hooks) {
+  setupTest(hooks);
+
+  // Replace this with your real tests.
+  test('it exists', function(assert) {
+    let store = this.owner.lookup('service:store');
+    let serializer = this.owner.factoryFor('serializer:<%= dasherizedModuleName %>').create({ store });
```
rwjblue: I think this should be `.lookup` (not `factoryFor().create()`) because serializers are singleton...

blueprints/serializer-test/qunit-rfc-232-files/tests/unit/__path__/__test__.js
```diff
@@ -0,0 +1,24 @@
...
+    assert.ok(serializer);
+  });
+
+  test('it serializes records', function(assert) {
+    let store = this.owner.lookup('service:store');
+    let record = run(() => store.createRecord('<%= dasherizedModuleName %>', {}));
+
+    let serializedRecord = record.serialize();
+
+    assert.ok(serializedRecord);
```
rwjblue: This should probably use some sort of `assert.deepEqual`

blueprints/serializer-test/qunit-rfc-232-files/tests/unit/__path__/__test__.js
alexander-alvarez: you need the store initialized on the serializer though, which lookup doesn't automatically do
is that something that should be addressed upstream?

blueprints/serializer-test/qunit-rfc-232-files/tests/unit/__path__/__test__.js
alexander-alvarez: Fine with whatever -- I simply copied the old ones https://github.com/emberjs/data/blob/master/blueprints/serializer-test/qunit-files/tests/unit/__path__/__test__.js#L14

blueprints/serializer-test/qunit-rfc-232-files/tests/unit/__path__/__test__.js
rwjblue: Haha, this seems sorta bonkers. I submitted https://github.com/emberjs/data/pull/5300 to simplify greatly 😺

rwjblue: @alexander-alvarez - Even without #5300, we should do `store.serializerFor(<%= dasherizedModelName %>)` here as that is the public API for getting a given things serializer. There are a few fallbacks that Ember Data handles nicely for you (e.g. use application if you don't have a model specific one) and we should go through the same common API.

alexander-alvarez: understood :+1:

blueprints/serializer-test/qunit-rfc-232-files/tests/unit/__path__/__test__.js
alexander-alvarez: I was playing around with adding `assert.deepEqual({}, serializedRecord);` in my repo, and realized that if you've already implemented the model for which a serializer is being generated, at the moment we can't generate a test case that will pass when generated + when using the equality check.

I was under the impression that generated test cases should always pass -- do we want to continue to have this behavior?

alexander-alvarez:
```
559) [PhantomJS 2.1] integration/relationships/has_many - Has-Many Relationships: Accessing a hasMany backed by a link multiple times triggers only one request
     Same promise is returned both times
```
not sure the test failure is related @rwjblue

rwjblue:
Restarted!

alexander-alvarez:
@rwjblue didn't realize -- it's failing on canary
```
Scenario ember-canary: FAIL
Command run: ember test
```

bmac:
Looks like the canary issue has be resolved and this is now passing again. @rwjblue has your issue been addressed?

PR #5300, opened 2017-12-22
Simplify `store.adapterFor` and `store.serializerFor`.

rwjblue:
* Make adapter and serializer types singleton
* Simplify `adapterFor` and `serializerFor`
* Remove `ContainerInstanceCache`


rwjblue:
I created a tracking issue (https://github.com/lytics/ember-data-model-fragments/issues/285) in ember-data-model-fragments...

rwjblue:
I chatted with @bmac offline, and there was some concern around the change making adapters and serializers singletons. Specifically, the move _away_ from being singletons was done in order to support having different adapters and serializer instances with multiple store services (where each store got its own adapter and serializer instances). There were no tests to that effect, however it would be fairly easy to revert that small subsection here.

I am concerned about lack of proper test coverage if this is actually feature to the system...

bmac:
@rwjblue Just reviewed it and this change looks good to me. I think we should land it without fce9eedb8fc7da37c54bba141435505aaa8da96c. I just created https://github.com/emberjs/data/pull/5301 to try to add a test for the behavior around adapters and serializers being singletons.

PR #5301, opened 2017-12-22
Add a test to ensure multiple stores get unique instances of serializ…

bmac:
…ers and adapters

### DIFF
--- a/tests/integration/multiple-stores-test.js
+++ b/tests/integration/multiple-stores-test.js
@@ -1,6 +1,7 @@
 import { run } from '@ember/runloop';
 import setupStore from 'dummy/tests/helpers/store';
 import { module, test } from 'qunit';
+import { get } from '@ember/object';
 
 import DS from 'ember-data';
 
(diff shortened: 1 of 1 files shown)


rwjblue:
Thank you for putting this together!

tests/integration/multiple-stores-test.js
```diff
@@ -136,3 +136,23 @@ test("embedded records should be created in multiple stores", function(assert) {
...
 });
+
+
+test("each store should have a unique instance of the serializers", function(assert) {
+  env.registry.register('serializer:home-planet', DS.RESTSerializer.extend({}));
+
+  let serializer_a = env.store_a.serializerFor('home-planet');
+  let serializer_b = env.store_b.serializerFor('home-planet');
+
+  assert.notEqual(serializer_a, serializer_b);
```
rwjblue: Can you add more assertions here?

Specifically, we don't _actually_ care that they aren't the same instance (not really at least), what we care about is that `.store` on them is the right thing.

```js
assert.equal(get(serializer_a, 'store'), env.store_a);
assert.equal(get(serializer_b, 'store'), env.store_b);
```

bmac: Updated

PR #5303, opened 2017-12-28
Fix tests with Ember Canary

bmac:


tests/integration/relationships/belongs-to-test.js
```diff
@@ -542,12 +542,27 @@ test("relationshipsByName does not cache a factory", function(assert) {
         "model factory based on relationship type matches the model based on store.modelFor");
 });
 
+function getDescriptor(object, key) {
+  var meta = Ember.meta(object);
+  var mixins = (meta && meta._mixins) || {};
+  for (let klass of Object.values(mixins)) {
```
rwjblue: FWIW, I think this only works because we use the babel-polyfill. We should generally avoid `for...of` (because it requires native `Symbol`) and I don't believe that `Object.values` is available on most platforms...

I'll try to tweak/fixup in a separate PR...

PR #5318, opened 2018-01-07
Use Ember.computed in the headers examples to avoid ember eslint warning

bmac:


fivetanley:
@bmac can you tag this with a `[doc beta]` so it gets pulled in?

PR #5323, opened 2018-01-10
Stop using deprecated vendor-prefix hook to inject enableoptionalfeat…

bmac:
…ures url check.


pangratz:
Looks good, thanks @bmac!

PR #5324, opened 2018-01-12
Don't serialize new has many relationships

ryanto:
Similar to #5317 but for hasMany.

Currently when saving a model with an unsaved `hasMany` a null id is used in the JSON:API document. 

For example

```js
let company = this.store.peekRecord('company', 1);

let user = this.store.createRecord('user');
company.get('users').addObject(user);

company.save();
```

Generates this post payload (note that we're using serializer option `attrs.users.serialize = true`)

```js
{
  data: {
    type: 'company',
    id: '1',
    relationships: {
      users: [
        data: {
          type: 'users',
          id: null
        }
      ]
    }
  }
}
```

This is an invalid payload. To fix this, we'll filter out new records from using in the serialized belongsTo.

### DIFF
--- a/addon/serializers/json-api.js
+++ b/addon/serializers/json-api.js
@@ -516,27 +516,31 @@ const JSONAPISerializer = JSONSerializer.extend({
     if (this.shouldSerializeHasMany(snapshot, key, relationship)) {
       let hasMany = snapshot.hasMany(key);
       if (hasMany !== undefined) {
+        // only serialize has many relationships that are not new
+        let nonNewHasMany = hasMany.filter(item => item.record && !item.record.get('isNew'));
 
-        json.relationships = json.relationships || {};
+        if (nonNewHasMany.length > 0) {
+          json.relationships = json.relationships || {};
 
-        let payloadKey = this._getMappedKey(key, snapshot.type);
-        if (payloadKey === key && this.keyForRelationship) {
-          payloadKey = this.keyForRelationship(key, 'hasMany', 'serialize');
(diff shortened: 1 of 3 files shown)


sly7-7:
Looks good if we don't want to automatically save the new record. 
I just wonder in that scenario if we could add an option to send the newly created `user` together with the `company`. Not sure it that even makes sense.

ryanto:
@sly7-7 Yup! I think you're describing a common use case for most applications would want. I'm expecting JSONAPI operations to be the answer to that problem: https://github.com/json-api/json-api/pull/1254

runspired:
@ryanto not super sure why this failed CI. Could you rebase and let it kick off again?

ryanto:
All set, I'll keep an eye on CI.

FYI when I pulled the latest and ran `yarn install` it looks like it added an entry to the lockfile.

runspired:
Thanks @ryanto !

juggy:
There is problem with this bug fix. An empty relationship is not serialized anymore, which is a different behavior than before. 

If you have a relation and remove it and save the record, to inform the backend of that relation change, you need to send that empty array back.

juggy:
Well the format should be the one of 3.1... 
```
...
"relationships": {
  "my_relations: {
    "data": []
  }
}
```

ryanto:
PR is here: https://github.com/emberjs/data/pull/5466

ryanto:
@juggy A fix has been merged, let me know if you run into any other issues.

PR #5332, opened 2018-01-22
[BUGFIX beta] Fix development build relationship caching.

rwjblue:
The previous implementation was intending to cache relationships when in the production environment, but avoid caching them in development.  Unfortunately, this code was factored in a way that relied on mutating internal private state of the underlying `Ember.ComputedProperty` instance when `Ember.testing` was true (which was used as a way to determine prod vs non-prod).

When this code was introduced (2014-12-02 in https://github.com/emberjs/data/pull/2530), setting `_cacheable` on a `ComputedProperty` instance would have worked as intended, however a refactor in Ember (ironically one day prior in https://github.com/emberjs/ember.js/pull/9489, but on Ember's canary channel) deprecated using `_cacheable` or `.cacheable()` (in favor of opting out of caching via `.volatile()`). Later support for `_cacheable` was completely removed in https://github.com/emberjs/ember.js/pull/12036 leading up to Ember 2.0.0.

This PR refactors the code a bit, so that we no longer need to mutate the computed property after creation and allows us to avoid some gnarliness...

### DIFF
--- a/addon/-private/system/relationships/ext.js
+++ b/addon/-private/system/relationships/ext.js
@@ -2,7 +2,6 @@ import { A } from '@ember/array';
 import { computed } from '@ember/object';
 import MapWithDefault from '@ember/map/with-default';
 import Map from '@ember/map';
-import Ember from 'ember';
 import { assert } from '@ember/debug';
 import {
   typeForRelationshipMeta,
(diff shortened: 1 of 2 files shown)


rwjblue:
Much easier to review with: https://github.com/emberjs/data/pull/5332/files?w=1

PR #5339, opened 2018-01-27
Prevent build issues when linking ember-source.

rwjblue:
Prior to this, we used our ember-data's own dirname as the default `baseDir` for any of our custom babel plugins. This baseDir is used to traverse all JS files in all deps (recursively) and ensure that if any of them are changed, the transpilation cache is invalidated.

Once we swapped to using ember-source as an npm package this meant that we would be stating/hashing all files inside of `node_modules/ember-source/**`. When using `npm link ember-source` this leads to a JS max heap errors (because of `node_modules/ember-source/tmp/**` being _MASSIVE_).

This fixes that specific issue, by using the **correct** basedir for each individual plugin.

/cc @mmun

### DIFF
--- a/lib/stripped-build-plugins.js
+++ b/lib/stripped-build-plugins.js
@@ -1,13 +1,15 @@
 'use strict';
 
 var path = require('path');
-var fs            = require('fs');
-var FilterImports = require('babel-plugin-filter-imports');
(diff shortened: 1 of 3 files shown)


rwjblue:
This is helping me work through and fix the CI failures, merging...

PR #5340, opened 2018-01-28
[BUGFIX beta] Update DS.Errors#unknownProperty to return `undefined`.

rwjblue:
In Ember 3.1+ any object that has an `unknownProperty` will throw an error if its `unknownProperty` returns any value other than `undefined` and the property is accessed via `obj.someProperty` (instead of `obj.get('someProperty')` or `Ember.get(obj, 'someProperty')`).


rwjblue:
Should be a BUGFIX beta

PR #5342, opened 2018-01-28
Refactor Travis setup to leverage stages...

rwjblue:
Using the new(ish) stages feature from Travis allows us to fail early (and avoid wasting CI resources) when a "simple" `ember test` with locked deps does not pass.

As you can see in the screenshot below:

screenshot

The first stage is `test`, and it uses our specific `yarn.lock` to run `ember test`.  If that build fails, nothing else is ran.  If it succeeds 4 more builds are kicked off (production tests, optional-features, node-tests, floating dependencies). If any of those builds fail, nothing else is ran. Then it moves on to testing various Ember versions...

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,9 @@ node_js:
   - "4"
   - "6"
 
+addons:
+  chrome: stable
+
 # hack to get after_success skipped in node 4
 matrix:
   exclude:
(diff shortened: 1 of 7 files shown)


rwjblue:
Rebased on top of latest master to get Chrome passing (needed changes in https://github.com/emberjs/data/pull/5340 to fix a new proxy related assertion)...

bmac:
Thanks @rwjblue.

PR #5345, opened 2018-01-31
Check if adapter#query is a wrapped function

ryanto:
Currently the `if` statement always returns false because adapter#query is a wrappedFunction, which has length == 0.

This PR checks if the query is a wrappedFunction and if it is we'll check the length of that function.

I'm also wondering if we can get rid of the `if` statement and have the recordArray created and given to the query function in all cases. My guess is that the conditional guards around having to create and then later add models to the record array because there is a problem with the events that fire? @stefanpenner do you have any thoughts?

### DIFF
--- a/addon/-private/system/store/finders.js
+++ b/addon/-private/system/store/finders.js
@@ -138,7 +138,10 @@ export function _query(adapter, store, modelName, query, recordArray) {
   let modelClass = store.modelFor(modelName); // adapter.query needs the class
 
   let promise;
-  if (adapter.query.length > 3) {
+  let createRecordArray = adapter.query.length > 3 ||
+    (adapter.query.wrappedFunction && adapter.query.wrappedFunction.length > 3);
+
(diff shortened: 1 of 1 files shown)


bmac:
@stefanpenner do you have some time to look at this pr?

ryanto:
Fyi, I turned out not needing this. We ended up using a different approach. Not sure if the PR would help anyone else though.

runspired:
We discussed this at the weekly meeting today and roughly our opinion is this:

- we should fix the current intent in the near term, which means merging this PR
- this check is likely to be brittle, especially since Ember will soon be moving the superWrapper into a WeakMap where we would not be able to check it, in the not so long term we need to remove the check
- to remove it, we need to cleanup the SnapshotRecordArray / adapterOptions mess, this arity check is present to enable us to often avoid a costly allocation
- we are currently plotting an overhaul of the request layer, which will allow us to plot a path out of this craziness

PR #5346, opened 2018-01-31
[BUGFIX] Object.values is not available in IE11 and some other randoms environments yet

runspired:


runspired:
I'm unsure why on my machine yarn insists that glob needs 3 resolutions instead of 2, but I'll remove the yarn lock commit from this PR so it's not a distraction

runspired:
Closing because this was against the wrong branch

PR #5347, opened 2018-01-31
[BUGFIX] Object.values is not available in IE11 and some other random environments yet

runspired:
What it says in the title

### DIFF
--- a/addon/-private/system/model/model.js
+++ b/addon/-private/system/model/model.js
@@ -1849,8 +1849,10 @@ if (isEnabled('ds-rollback-attribute')) {
       @method rollbackAttribute
     */
     rollbackAttribute(attributeName) {
-      if (attributeName in this._internalModel._attributes) {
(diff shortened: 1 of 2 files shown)


addon/-private/system/model/model.js
```diff
@@ -1849,8 +1849,10 @@ if (isEnabled('ds-rollback-attribute')) {
       @method rollbackAttribute
     */
     rollbackAttribute(attributeName) {
-      if (attributeName in this._internalModel._attributes) {
-        this.set(attributeName, this._internalModel.lastAcknowledgedValue(attributeName));
+      let modelData = this._internalModel._modelData;
+
+      if (attributeName in modelData.__attributes) {
+        this.set(attributeName, modelData.lastAcknowledgedValue(attributeName));
```
hjdivad: this should delegate to `modelData.rollbackAttribute(attributeName)` similar to `rollbackAttributes`

PR #5350, opened 2018-02-06
[BUGFIX beta] Fix unloadAll to also unload live records

dwickern:
`unloadAll` should also unload records which were fetched using `findAll`. Currently the unloaded records are destroyed but the record array isn't cleared. So calling `findAll` a second time after unloading will return the destroyed records _and_ fresh ones (potentially duplicates of the destroyed records).


addon/-private/system/record-array-manager.js
```diff
@@ -76,6 +76,13 @@ export default class RecordArrayManager {
     this.internalModelDidChange(internalModel);
   }
 
+  recordWasUnloaded(internalModel) {
+    let array = this._liveRecordArrays[internalModel.modelName];
```
joelalejandro: @dwickern Since `array` isn't changing, maybe it's best to use `const` here?

dwickern: It was to match the existing code style

joelalejandro: Ah, I see. :+1:

sly7-7:
LGTM, bu @hjdivad could you review this please ?
Also maybe prefix the commit with [BUGFIX beta] ?

runspired:
I will look into this issue and review this PR as part of #5378. I suspect the underlying issue is a combination of one we have already fixed with the issue seen in `peekAll`.

dwickern:
@runspired the test in this PR fails against your branch

runspired:
I dug in and while this fix appears to work it does not address an underlying issue:

`store.findAll` appears to put records into a  fundamentally different state than `store.push`. I'm looking into why this is, as they should be in the same state.

runspired:
The issue appears to be that records need to have been accessed to be unloaded correctly. `findAll` puts records into the record-array, but does not materialize them. Using `store.push` has the side-effect of materializing them. I was able to build simpler test reproductions once I narrowed in on this.

runspired:
I found the underlying issue and resolved it here: https://github.com/emberjs/data/pull/5378/commits/55fd92467809f00fb25371fa89b64502745e6eaa

Thanks for the failing test and fix attempt, it was extremely helpful in narrowing in on the root cause.

PR #5351, opened 2018-02-07
Test for #5274 - fetch hasMany using links when parent & child records are already loaded

mattraydub:
see #5274 for explanation


mattraydub:
A different PR #5257 fixes the breaking test here, however that PR has not been merged or accepted, and seems to have some resistance

PR #5352, opened 2018-02-13
[BUGFIX] record.data is intimate API that we need to deprecate

runspired:
This just fixes the accessor. We should actually deprecate this though.

### DIFF
--- a/addon/-private/system/model/model.js
+++ b/addon/-private/system/model/model.js
@@ -1129,7 +1129,8 @@ const Model = EmberObject.extend(Evented, {
  */
 Object.defineProperty(Model.prototype, 'data', {
   get() {
-    return this._internalModel._data;
+    // TODO deprecate this!!!!!!!!!!! it's private but intimate
+    return this._internalModel._modelData._data;
   }
 });
 


runspired:
@rwjblue the answer to that is complicated, mostly bc the answer to that should have always been "dont". It's possible we decide to do something like `record.modelData` but that's likely for another RFC to decide.

runspired:
cc @igorT since this may count as a point of feedback for the ModelData RFC

hjdivad:
this is clearly an improvement although i agree a followup pr for deprecation / better path forward is a good idea

hjdivad:
@runspired i don't quite follow your reply; this seems likely something we should clearly deprecate?

runspired:
@hjdivad users reach into `data` because they want to access underlying properties without the proxy / dirty state / causing a load (in the case of relationships) / some other reason.  While this has always been mega-bad, I don't think we can persay deprecate this without a documented pattern going forward that addresses why folks were doing this in the first place.

addon/-private/system/model/model.js
```diff
@@ -1129,7 +1129,8 @@ const Model = EmberObject.extend(Evented, {
  */
 Object.defineProperty(Model.prototype, 'data', {
   get() {
-    return this._internalModel._data;
+    // TODO deprecate this!!!!!!!!!!! it's private but intimate
+    return this._internalModel._modelData._data;
```
stefanpenner: can i haz test?

hjdivad:
> @hjdivad users reach into data because they want to access underlying properties without the proxy / dirty state / causing a load (in the case of relationships) / some other reason. While this has always been mega-bad, I don't think we can persay deprecate this without a documented pattern going forward that addresses why folks were doing this in the first place.

The references API should handle the relationships case.

We'll want to review some concrete examples of users doing this to understand the actual use cases involved.

Issue #5343, opened 2018-01-30
model.unloadRecord causes Uncaught TypeError: Cannot read property 'eachAttribute' of null


alexmiddeleer:
```
at new Snapshot (-private.js:5069)
    at InternalModel.createSnapshot (-private.js:6894)
    at Class._flushPendingFetchForType (-private.js:10771)
    at cb (ember-metal.js:5546)
    at OrderedSet.forEach (ember-metal.js:5358)
    at MapWithDefault.forEach (ember-metal.js:5554)
    at Class.flushAllPendingFetches (-private.js:10687)
    at invoke (backburner.js:258)
    at Queue.flush (backburner.js:151)
```
This is after calling `findRecord` `then` `unloadRecord` on the result. We are upgrading to ember 2.18 and ember-data 2.18, and this appears to be a regression.

Twiddle reproduction:
https://ember-twiddle.com/c16533267ff13353b25498fac4d05ca3?fileTreeShown=false&numColumns=2&openFiles=controllers.application.js%2Ctemplates.application.hbs

Possibly a duplicate of https://github.com/emberjs/data/issues/4963, but that was supposedly fixed a while ago.

The code that doesn't work is this:
```
this.get('store').findRecord('person', 1).then((person) => {
   person.unloadRecord();
});
```

jlami:
Unloading the record from the `controller.model` seems to not give this error.
It actually really needs to be just after the `findRecord`.
I also have this problem in 2.17.

medokin:
Repoduced in ember-data 3.0.1

olivierchatry:
I have the same problems. If I do that it works :
```

**Comment by buschtoens** [comment]
I get the same error, when I try to load a relationship by ID, that `404`s.

For instance:

```js
{
  id: '1',
  type: 'post',
  relationships: {
    'author': { id: '2', type: 'author' }
  }
}
```

Now, if I access the `author` relationship and therefore trigger Ember Data to load it and the request fails with a `404`, I get the same error:

```
ncaught TypeError: Cannot read property 'eachAttribute' of null
    at new Snapshot (vendor.js:181070)
    at InternalModel.createSnapshot (vendor.js:182969)
    at Class._flushPendingFetchForType (vendor.js:186533)
    at cb (vendor.js:45496)
    at OrderedSet.forEach (vendor.js:45309)
    at MapWithDefault.forEach (vendor.js:45504)
    at Class.flushAllPendingFetches (vendor.js:186450)
[...]

Issue #5328, opened 2018-01-18
unloadRecord after v2.12


Vincz:
Hey :)

I have this simple piece of code (removing record from the store while notified by a websocket).

```
handleRemove: function(content) {
        var store = this.get("store");
        for (var model in content) {
            if (!this.get("ignoredModels").includes(model)) {
                var modelId = content[model]["id"];
                var object = store.peekRecord(store.convert(model), modelId);
                if (object) {
                    object.unloadRecord();
                }
            }
        }
    }
```

In ember-data 2.12 everything works as expected. 
But after this version, I'm unable to track the origin (if someone know how to do it ?), but the store tries to re-fetch some unloaded models (and fail with a 404 from the server).
I'm also encoutering sometimes the `Cannot read property 'eachAttribute' of null ` error that others seems to have on unload.

henrymazza:
#4986 says it fixed it, but I couldn't see it. I'm experiencing this with 2.18.0. The record vanishes leaving ManyArrays full of nulls to break glimmer.

henrymazza:
Actually right **before** unloadRecord.

bastimeyer:
I've just tried upgrading to ED 3.0.1 and am still experiencing this issue.
See this simple test: https://gist.github.com/bastimeyer/ecb84112ba24082dfa58fdd3596ef978

runspired:
unloadRecord should not be use for client-side delete. I would recommend building a solution overtop of `deleteRecord` instead.

sly7-7:
@runspired I thought the recent work by @hjdivad around unloadRecord was to bring back this client-side delete. I'm just confused now.

runspired:
@sly7-7 it brought it back for a small subset of cases where it could work.

sly7-7:
In that case, I think I must rework my client-side delete process of the object in my app. But does deleteRecord won't let some "zombies" records, and so my application would leak ?

JonathanBristow:
If we shouldn't use unloadRecord how would you advise removing a record from the store?
The record I'm working with is loaded as part of a hasMany relationship.

henrymazza:
Indeed. If a record gets deleted on server how you make ED understand that? The right way should be to issue a `reload()`, but the server will return a 404 and ED will just blow up. If I issue a `destroyRecord()` the server will return 404 too. `deleteRecord` will leave a zombie object in my collections. That is, you can’t differentiate between something that is new, was deleted or never existed. 

[...]

PR #5353, opened 2018-02-14
unload should invalidate async hasMany: failing test

jlami:
This simulates the problems I'm having when trying to debug #5343 

As far as I could tell the new behaviour for async relationships is not to mark unloaded items as deleted, but to use `invalidate+refetch` as mentioned here: https://github.com/emberjs/data/commit/6e8a23656cb073596cd951bdab9ca0541d1bcbe3
But I'm having trouble understanding whether the invalidate should be done by the ember-data user, or by ember-data.
I feel this should be ember-data's job, but this test has one line in comment where the user could `hasMany(....).reload()` to refetch the data to fix the problem. But that feels like a bad solution.

### DIFF
--- a/tests/integration/records/unload-test.js
+++ b/tests/integration/records/unload-test.js
@@ -2008,3 +2008,113 @@ test('1 sync : many async unload sync side', function(assert) {
     })
   );
 });
+
+test('unload should invalidate async hasMany', function(assert) {
+  let isUnloaded = false;
+  env.adapter.coalesceFindRequests = false;
+
+  env.adapter.findRecord = (store, type, id) => {
+    assert.equal(type, Boat, 'findRecord(_, type) is correct');
+
+    if (id === '2' && isUnloaded) {
+      throw "not found";
+    }
+
+    let relationships = {
+      person: {
+        data: {
+          type: 'person',
+          id: '1'
+        }
+      }
+    };
+
+    return {
+      data: {
+        type: 'boat',
+        id: id,
+        relationships
+      }
+    }
+  };
+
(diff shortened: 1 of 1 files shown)


jlami:
Might be related to #5328

jlami:
@sly7-7 I think the problem here is partly due to me not understanding the new async handling. The first assert of the count not being correct after the unload could probably be ignored. But the final exception might be the real problem.

Where ember-data now correctly is able to reload an dematerialized record that is still in the ManyArray, it fails to handle the case when a record is deleted on the server side. The rejected fetch will result in a complete failure now, while this could actually be a valid response. So maybe this test should be renamed into 'async relationship can handle server side deletion' or something like that?

jlami:
Does anybody have any idea on how a rejection in a loading ManyArray should be handled in async relationships? The async part makes it valid to have changes on the server side once the data is asked for. So I think ember-data should play well with any record that is rejected.

The findMany part only gives warnings if you don't include deleted items as far as I can tell: https://github.com/emberjs/data/blob/master/addon/-private/system/store.js#L892
But this path is only used if coalesceRecords is true I believe.

We don't coalesce records in our plugin and this path seems to throw exceptions that are not caught and I'm not really sure if this will actually update the ManyArray if it would be used with a `model.get('manyArray').catch(...)`. And even if this would work, computed properties don't work this way. So I really think the ManyArray should handle this when reloading. Maybe here: https://github.com/emberjs/data/blob/master/addon/-private/system/relationships/state/has-many.js#L290

runspired:
@bmac we can close this as fixed by #5377

PR #5356, opened 2018-02-16
Removing reference to Ember.Logger

lupestro:
This is part of the work for RFC 297, the deprecation of Ember.Logger


lupestro:
The two errors from Travis were timeouts. I could use some help finding out why they happened.

bmac:
@lupestro I think I fixed the timeout issue in https://github.com/emberjs/data/pull/5360. Do you mind rebasing this pr?

bmac:
Looks like this is now passing after restarting travis. Thanks @lupestro.

PR #4931, opened 2017-04-18
[BUGFIX beta] [fixes #4509] createRecord initializes correctly

stefanpenner:
* by the name Model.prototype.init is called, internal model’s data has been loaded
* do no emit extra change events during `createRecord` as the attributes are the initial state.

Issue #4509, opened 2016-08-23
`init` event fires before leaving empty state


fivetanley:
Consider the following model definition:

``` javascript
import Model from "ember-data/model";
import attr from "ember-data/attr";
import { belongsTo, hasMany } from "ember-data/relationships";
import Ember from 'ember';

export default Model.extend({
  title: attr(),
  propSetFromObservingTitle: attr(),

  setPropFromTitle: Ember.observer('title', function() {
      this.set('propSetFromObservingTitle', this.get('title') + 'hi');
  }).on('init')
});
```

In this example, we have two properties that are `DS.attr`'s. We have `propSetFromObservingTitle` because we basically want a computed property that _also_ gets serialized to the server.

This code works with normal ember objects, but when we run the following code:

``` javascript
this.store.createRecord('post', {title: 'foo'});
```

We get the following exception:

```
VM423 ember.debug.js:31321 Error: Attempted to handle event `didSetProperty` on <post:null> while in state root.empty. Called with {name: propSetFromObservingTitle, oldValue: undefined, originalValue: undefined, value: undefinedhi}.
    at new Error (native)
    at Error.EmberError (https://cdnjs.cloudflare.com/ajax/libs/ember.js/2.6.2/ember.debug.js:15976:21)
    at InternalModel._unhandledEvent (https://cdnjs.cloudflare.com/ajax/libs/ember-data.js/2.7.0/ember-data.js:2623:13)
    at InternalModel.send (https://cdnjs.cloudflare.com/ajax/libs/ember-data.js/2.7.0/ember-data.js:2501:14)
    at Class.set (https://cdnjs.cloudflare.com/ajax/libs/ember-data.js/2.7.0/ember-data.js:13361:31)
    at Descriptor.computedPropertySet [as _set] (https://cdnjs.cloudflare.com/ajax/libs/ember.js/2.6.2/ember.debug.js:14654:28)
    at Descriptor.computedPropertySetWithSuspend [as setWithSuspend] (https://cdnjs.cloudflare.com/ajax/libs/ember.js/2.6.2/ember.debug.js:14634:19)
    at Descriptor.computedPropertySetEntry [as set] (https://cdnjs.cloudflare.com/ajax/libs/ember.js/2.6.2/ember.debug.js:14612:17)
    at Object.set (https://cdnjs.cloudflare.com/ajax/libs/ember.js/2.6.2/ember.debug.js:20348:12)
[...]

PR #5359, opened 2018-02-16
failed test: createRecord creates two records on ember 3.0

dwickern:
After upgrading ember from 2.18.2 to 3.0, _with ember-data unchanged_ at version 2.18.1, a call to `createRecord` creates 2 records.

It happens under these conditions:
1. The call to `createRecord` sets up a relationship which has an inverse
2. The inverse array is being observed (e.g. used in a template)

Here's a shortened example of the problem:
```js
const Post = Model.extend({
  comments: hasMany('comment')
});
const Comment = Model.extend({
  post: belongsTo('post')
});

// non-working example
let post = this.get('model');
let comment = this.store.createRecord('comment', { post });
comment === post.get('comments.firstObject'); // false!! actually 2 comments were created
```

It works if you set up the relationship _after_ construction:
```js
// workaround
let post = this.get('model');
let comment = this.store.createRecord('comment');
comment.set('post', post); // don't pass directly to createRecord
comment === post.get('comments.firstObject'); // true
```

It also works if nobody is observing the `comments` array.

The stack trace explains why: `createRecord` updates the inverse record array synchronously before actually creating the record
```
create (container.js:420)
create (container.js:178)       <<<============
getRecord (-private.js:6597)
objectAt (-private.js:4180)
objectAt (array.js:50)
objectAtContent (array_proxy.js:110)
objectAt (array_proxy.js:149)
objectAt (array.js:50)
addObserverForContentKey (array.js:600)
arrayDidChange (array.js:554)
arrayContentDidChange (array.js:105)
_emberMetal.Mixin.create._Mixin$create.arrayContentDidChange (array.js:281)
exports.default._object.default.extend._EmberObject$extend._arrangedContentArrayDidChange (array_proxy.js:221)
applyStr (ember-utils.js:528)
sendEvent (ember-metal.js:257)
arrayContentDidChange (array.js:108)
_emberMetal.Mixin.create._Mixin$create.arrayContentDidChange (array.js:281)
internalReplace (-private.js:4224)
_addInternalModels (-private.js:4240)
addInternalModel (-private.js:4412)
addInternalModel (-private.js:3745)
addInternalModel (-private.js:4733)
setInternalModel (-private.js:4646)
set (-private.js:12386)
computedPropertySet (ember-metal.js:3456)
computedPropertySetWithSuspend (ember-metal.js:3438)
computedPropertySetEntry (ember-metal.js:3416)
Class (core_object.js:100)
_ClassMixinProps.create (core_object.js:277)
create (container.js:420)
create (container.js:178)       <<<============
getRecord (-private.js:6597)
createRecord (-private.js:9922)
```

### DIFF
--- a/tests/integration/relationships/one-to-many-test.js
+++ b/tests/integration/relationships/one-to-many-test.js
@@ -1452,3 +1452,35 @@ test("Rollbacking attributes of a created record works correctly when the belong
   assert.equal(user.get('accounts.length'), 0, "User does not have the account anymore");
   assert.equal(account.get('user'), null, 'Account does not have the user anymore');
 });
+
+test("createRecord updates inverse record array which has observers", function(assert) {
+
+  env.adapter.findAll = () => {
+    return {
+      data: [{
+        id: '2',
+        type: 'user',
+        attributes: {
+          name: 'Stanley'
+        }
+      }]
+    }
+  };
+
+  return store.findAll('user').then(users => {
+    assert.equal(users.get('length'), 1, 'Exactly 1 user');
+
+    let user = users.get('firstObject');
+    assert.equal(user.get('messages.length'), 0, 'Record array is initially empty');
+
+    // set up an observer
+    user.addObserver('messages.@each.title', () => {});
+    user.get('messages.firstObject');
+
+    let message = run(() => store.createRecord('message', { user, title: 'EmberFest was great' }));
+    assert.equal(user.get('messages.length'), 1, 'The message is added to the record array');
+
+    let messageFromArray = user.get('messages.firstObject');
+    assert.equal(message, messageFromArray, 'Only one message should be created');
+  });
+});


dwickern:
I also see a deprecation error on this line which doesn't seem right
```js
user.get('messages.firstObject')
// Getting the '@each' property on object <DS.Errors:ember9921> is deprecated
```

mmun:
@dwickern That deprecation error arrises because when assert.equal is false it tries to serialize the objects. Not sure the best way to address that. We could try making `@each` non-enumerable.

mmun:
So after some initial sleuthing, that same call stack appears when using Ember 2.11 as well. It's not clear to me why two records are created.

dwickern:
@mmun Thanks for looking into this. Sure enough, I see the same reentrant call into `getRecord` on 2.11. My test passes on 2.11-2.18 so something changed in 3.0 to make the bug visible.

mmun:
@dwickern I'll keep looking later today.

mmun:
So I've got an update. I haven't been able to do a full root cause as it is simply too complicated for me and I don't have the time to step through it at all. However, what I can say is that it looks like messy interaction between array events being triggered during initialization caused by Ember Data when it is splicing into inverse relationships during createRecord.

It's not obvious to me whether this is a limitation of the current object model or whether it is a bug in Ember Data. Fortunately, we can avoid this issue relatively easily by changing https://github.com/emberjs/data/blob/1732cd5beeda795a08fb6f70260fe830699ed454/addon/-private/system/store.js#L361

```js
let record = internalModel.getRecord();
setProperties(record, properties);
```

This will allow the record to be fully instantiated and assigned into the internal model's `_record` field before we do any complex logic like splicing into inverse relationships.

A couple implementation notes: 1) we probably want to remove the properties parameter internally for getRecord. 2) with my proposed implementation we end up assigning `id` to the record twice. Not sure how big of a deal that is.

@dwickern Would you be interested in implementing the fix I suggested?

mmun:
Here are some issues where this behaviour was initially introduced:

- https://github.com/emberjs/data/issues/4509
- https://github.com/emberjs/data/pull/4931

I believe it is acceptable to revert that PR until it can be implemented correctly (i.e. not creating two records recursively). I think it would also be fine if someone wants to attempt to fix the current implementation.

mmun:
I bisected the ember.js commit down to what I expected it to be: https://github.com/emberjs/ember.js/pull/16166.

Still, I'm not sure whether this is more of Ember or Ember data's fault, since the trivial refactor of moving the property assignments described in https://github.com/emberjs/data/pull/5359#issuecomment-366880425 to avoid the double create fixes the issue in https://github.com/emberjs/ember.js/issues/16258 as well.

dwickern:
The fix from https://github.com/emberjs/data/pull/5359#issuecomment-366880425 LGTM, I'll open a PR unless you want to look into it further

mmun:
@dwickern I am worried that, while the fix I described will correctly this bug, it is undoing something that was done intentionally to solve https://github.com/emberjs/data/issues/4509.

I believe it ought to be possible to fix this bug without compromising but I haven't been able to figure it out yet. I've pinged some other core team members to take a look. If we can't get to the bottom of it in the next couple weeks than I think we'll have to do the fix I described.

PR #5364, opened 2018-02-27
[BUGFIX release] Setup internal model's _record before assigning relationships

mmun:
Fixes #5359 and https://github.com/emberjs/ember.js/issues/16258.

Before this PR, it was possible to accidentally reify inverse relationships into an invalid state when combining `createRecord` with `@each` observers/CPs on an inverse relationship (see test). This would cause internalModel's `getRecord` to be called recursively with unexpected results. This problem became amplified in Ember 3.x because ArrayProxies now cache the result of `objectAt`, which lead us to caching the wrong invalid state.

This fix ensures that `_internalModel._record` is setup immediately, before any reification happens, to avoid ever entering an invalid state.

### DIFF
--- a/addon/-private/system/model/internal-model.js
+++ b/addon/-private/system/model/internal-model.js
@@ -343,7 +343,7 @@ export default class InternalModel {
       // instances with the injections applied
       let createOptions = {
         store: this.store,
-        _internalModel: this,
+        __setupInternalModel: this,
         id: this.id,
         currentState: this.currentState,
         isError: this.isError,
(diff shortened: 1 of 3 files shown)


addon/-private/system/model/internal-model.js
```diff
@@ -343,7 +343,7 @@ export default class InternalModel {
       // instances with the injections applied
       let createOptions = {
         store: this.store,
-        _internalModel: this,
+        __setupInternalModel: this,
```
rwjblue: Can you add a comment here also? So that it's clear that this `__setupInternalModel` property must exist before any user specified properties. In particular, we care the most about user specified properties that are relationships.

rwjblue:
A possibly simpler solution to this would be to revert the change made in https://github.com/emberjs/data/pull/4931. That obviously has the side-effect of not exposing the user-created properties to `init`, but I believe that fix only _partially_ solved the issue reported in https://github.com/emberjs/data/issues/4509 (it fixed the case where the user called `.createRecord` but not the "normal" case with a `.findRecord`).

mmun:
@rwjblue Good point. Let's do that.

PR #5365, opened 2018-02-28
Remove private api

BryanCrotaz:
`begin...endPropertyChanges` is private API and subject to change in the near future. `changeProperties` is the recommended (@rwjblue) replacement. Plus it has try...finally built in to avoid mismatched pairs of calls, which would cause observers app-wide to stop working.


BryanCrotaz:
After I lost 4 days of my life trying to work out why all my observers had stopped working, I thought I was duty bound to stop anyone else falling in the hole!

PR #5366, opened 2018-02-28
fix initial polymorphic attribute

sly7-7:
Maybe I'm missing something, but it seems weird for a relationship to be polymorphic by default, right ?

### DIFF
--- a/addon/-private/system/relationships/state/has-many.js
+++ b/addon/-private/system/relationships/state/has-many.js
@@ -10,7 +10,6 @@ export default class ManyRelationship extends Relationship {
     super(store, internalModel, inverseKey, relationshipMeta);
     this.belongsToType = relationshipMeta.type;
     this.canonicalState = [];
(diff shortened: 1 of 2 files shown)


bmac:
@hjdivad would you have time to review this pr?

rwjblue:
Would you mind adding a regression test, to help ensure we don't make this mistake again...

sly7-7:
The thing is I don't even know if this property is used at all. I think (maybe wrongly) if it was used, some existing tests would already be failing. I will investigate more asap.

sly7-7:
@runspired Ok, I also removed the initialization of the `isPolymorphic` attribute from the HasManyRelationship class. In that case, without the right initialization to 'false' in the parent class, some tests are failling.

runspired:
Thanks! so the TLDR for anyone following along was that our tests passed bc the extended class (has-many) also initialized `isPolymorphic` clobbering our mistake. By fixing the mistake we removed the dual initialization.

Issue #5358, opened 2018-02-16
Base ember-data twiddles are broken


NLincoln:
Both the `RESTAdapter` and `JSONAPIAdapter` base twiddles (found [here](https://github.com/emberjs/data/blob/master/CONTRIBUTING.md#reporting-a-bug)) log the following error to the console:

```
ember-data.js:258 Uncaught Error: Could not find module `ember-data/-private/global` imported from `(require)`
    at missingModule (ember-data.js:258)
    at findModule (ember-data.js:269)
    at requireModule (ember-data.js:35)
    at ember-data.js:18643
    at ember-data.js:18650
    at ember-data.js:18651
missingModule @ ember-data.js:258
findModule @ ember-data.js:269
requireModule @ ember-data.js:35
(anonymous) @ ember-data.js:18643
(anonymous) @ ember-data.js:18650
(anonymous) @ ember-data.js:18651

loader-e7feae4459c970132902cb2a2663d88d.js:1 
Uncaught Error: Could not find module `ember-data` imported from `(require)`
    at n (loader-e7feae4459c970132902cb2a2663d88d.js:1)
    at s (loader-e7feae4459c970132902cb2a2663d88d.js:1)
    at requireModule (loader-e7feae4459c970132902cb2a2663d88d.js:1)
    at VM1096 about:srcdoc:212
n @ loader-e7feae4459c970132902cb2a2663d88d.js:1
s @ loader-e7feae4459c970132902cb2a2663d88d.js:1
requireModule @ loader-e7feae4459c970132902cb2a2663d88d.js:1
(anonymous) @ VM1096 about:srcdoc:212
```

This is effecting bug reports that use the base ember-data twiddle, such as #5295

sly7-7:
@bmac @locks Hi you two :) Do you have any idea where to start to make this work again ?

PR #5367, opened 2018-02-28
Update ember twiddle links to working versions

bmac:
Fixes https://github.com/emberjs/data/issues/5358


sly7-7:
@bmac, just out of curiosity, in case it happens again, where did you find those new links ?

PR #5369, opened 2018-03-01
[BUGFIX beta] Fix createRecord creating two records

dwickern:
Fixes #5359 and emberjs/ember.js#16258

Explained in #5364:
> Before this PR, it was possible to accidentally reify inverse relationships into an invalid state when combining createRecord with @each observers/CPs on an inverse relationship (see test). This would cause internalModel's getRecord to be called recursively with unexpected results. This problem became amplified in Ember 3.x because ArrayProxies now cache the result of objectAt, which lead us to caching the wrong invalid state.

This applies the workaround from https://github.com/emberjs/data/pull/5359#issuecomment-366880425 to create and cache the record _before_ setting any state which might update inverses.

cc @mmun

### DIFF
--- a/addon/-private/system/model/internal-model.js
+++ b/addon/-private/system/model/internal-model.js
@@ -334,7 +334,7 @@ export default class InternalModel {
     return this.currentState.dirtyType;
   }
 
-  getRecord(properties) {
+  getRecord() {
     if (!this._record && !this._isDematerializing) {
       heimdall.increment(materializeRecord);
       let token = heimdall.start('InternalModel.getRecord');
@@ -349,10 +349,6 @@ export default class InternalModel {
         isError: this.isError,
         adapterError: this.error
       };
-
-      if (typeof properties === 'object' && properties !== null) {
-        emberAssign(createOptions, properties);
-      }
 
       if (setOwner) {
         // ensure that `getOwner(this)` works inside a model instance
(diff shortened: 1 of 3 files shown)


mmun:
FYI, the CI failure is unrelated to this PR.

dwickern:
I'll add 2 records were being created on ember 2.x as well due to #4931. The extra record was discarded (leaked?) until the ArrayProxy change in ember 3.0 made the behavior visible.

Thanks @mmun for all your work on this!

Do you want a separate PR backporting to ember-data 2.18.x?

rwjblue:
@dwickern - I think the issues with master should be resolved, would you mind a rebase?

mmun:
@rwjblue I rebased.

aaronbhansen:
I'm getting this on Ember 2.18.3. I applied this patch locally and it fixed my problem. @rwjblue @mmun any issue with me opening a PR for this same fix on 2.18?

See this image. 
image

Both `order_selection` and `$E` are `order/selection` models. They have different emberIds, but the same internal model.

My component has a reference to the `order_selection` from the initial event, but the store has a different model. the `order_selection` my component is referencing isn't actually found in the store anywhere.

runspired:
@aaronbhansen this fix introduced other regressions, there would be many, many commits necessary to get the complete fix, as each attempt introduced a new regression :trollface: I'm not sure it is a good idea for us to attempt to backport this one given there is a relatively easy app-side fix:

```
let newRecord = store.createRecord(type, props);
newRecord.setProperties(relationships);
```

aaronbhansen:
I'll see if that works and update our app code for now. Thanks for the quick response

PR #5370, opened 2018-03-02
Fix failing test on master by adding an invalid inverse to trigger th…

bmac:
…e expected error

@runspired I'd like you eyes on this since I don't have a great mental model of all the moving pieces in this part of the code. It seems like Ember Data is already detecting there is no valid inverse for the `comment` relationship so it never generates an inverseMeta. So I added an explicit inverse that is invalid (because the other side has `inverse: null`) to trigger the expected error.


runspired:
@bmac just to follow up:

The thing this test was intended to capture (once `inverse: null` is present) was indeed [addressed here](https://github.com/emberjs/data/pull/5261/files).  This is the correct fix for this test.

However, investigating this made me realize there is still an uncovered case, I will PR a new test for it.

bmac:
Thanks for the follow up @runspired.

PR #5371, opened 2018-03-06
Upgrade to ember-cli-qunit 4.0 and replace ember-dev with ember-qunit…

bmac:
…-assert-helpers


tests/integration/relationships/inverse-relationships-test.js
```diff
@@ -640,9 +640,11 @@ testInDebug("Inverse null relationships with models that don't exist throw a nic
 
   let env = setupStore({ user: User });
 
-  assert.throws(() => {
-    run(() => env.store.createRecord('user', { post: {}}));
-  }, /No model was found for 'post'/);
+  run(() => {
```
rwjblue: This change seems somewhat odd

bmac: Indeed, not sure what was going on there. I just moved the runloop back inside the `expectAssertion` and everything passes so I'm going to commit that change.

rwjblue:
@bmac - looks like I created a merge conflict in yarn.lock, sorry 😢

PR #5372, opened 2018-03-07
[BUGFIX beta] Use @ember/ordered-set to avoid deprecations.

rwjblue:
Ember will deprecate usage of `Ember.OrderedSet` (likely in 3.2 or 3.3), and as part of that process the implementation has been extracted to an addon `@ember/ordered-set`. The addon intelligently avoids shipping duplicated implementations, and is compatible with newer _and_ older Ember versions.

Closes https://github.com/emberjs/data/issues/5254

### DIFF
--- a/addon/-private/system/ordered-set.js
+++ b/addon/-private/system/ordered-set.js
@@ -1,7 +1,5 @@
+import EmberOrderedSet from '@ember/ordered-set';
 import { guidFor } from '@ember/object/internals';
-import Ember from 'ember';
-
-const EmberOrderedSet = Ember.OrderedSet;
 
 export default function OrderedSet() {
   this._super$constructor();
--- a/package.json
+++ b/package.json
@@ -20,6 +20,7 @@
   "author": "",
   "license": "MIT",
   "dependencies": {
+    "@ember/ordered-set": "^1.0.0",
     "amd-name-resolver": "0.0.7",
     "babel-plugin-ember-modules-api-polyfill": "^1.4.2",
     "babel-plugin-feature-flags": "^0.3.1",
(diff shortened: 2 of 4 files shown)


sly7-7:
I encounter this warning when building the app: 
`'@ember/ordered-set' is imported by tmp/rollup-cache_path-O3J4AVcJ.tmp/-private/system/ordered-set.js, but could not be resolved – treating it as an external dependency`
running on ember-cli: 3.0.0

bmac:
@sly7-7 do you have `@ember/ordered-set` installed in your `node_modules` folder?

rwjblue:
Maybe I missed adding it as an external to the roll up part of the build?

rwjblue:
Should have gone in this list: https://github.com/emberjs/data/blob/dc2523230b222c8554813f4416459faef79f3e89/index.js#L180

😭 sorry

bmac:
@sly7-7 is [on it](https://github.com/emberjs/data/pull/5383).

broerse:
I encounter this warning when building the app:
'@ember/ordered-set' is imported by tmp/rollup-cache_path-O3J4AVcJ.tmp/-private/system/ordered-set.js, but could not be resolved – treating it as an external dependency
running on ember-cli: 3.1.0 but mot in 3.0.0

bmac:
That warning is fixed by https://github.com/emberjs/data/pull/5383. Looks like I missed cherry-picking that commit into the 3.1 release. I'll release a 3.1.1 patch tomorrow with that fix.

bmac:
Ember Data 3.1.1 has been published with a fix for this warning.

broerse:
@bmac Yes it is fixed again. Thanks!

jherdman:
This deprecation seems to be raised in 3.3.1 as well.

runspired:
@jherdman I think it is a different deprecation (the one where we aren't using `new` which is resolved by #5557 and which needs backported to 3.1, 3.2, 3.3

PR #5376, opened 2018-03-13
[bugfix beta] Fetch cancels unload

### LINKED ISSUE #5343: model.unloadRecord causes Uncaught TypeError: Cannot read property 'eachAttribute' of null

hjdivad:
When fetching records, cancel internal model destruction if it is
scheduled, ie if the model is dematerializing because it was unloaded.

It is not possible to construct snapshots for dematerializing internal
models.  Prior to this commit, this could happen when fetching a record
in the same runloop that it was unloaded.  A straightfoward way of
getting into this state was via

```
store.findRecord('book', 1).then(b => b.unloadRecord())
```

when the model is already cached in the store.  Under these conditions,
the fetch is scheduled, then the promise fulfills with the cached record
and is unloaded and *then* the scheduled fetch is flushed.

[fix #5343]

### DIFF
--- a/addon/-private/system/store.js
+++ b/addon/-private/system/store.js
@@ -855,6 +855,18 @@ Store = Service.extend({
       seeking[internalModel.id] = pendingItem;
     }
 
+    for (let i = 0; i < totalItems; i++) {
+      let internalModel = internalModels[i];
+      // We may have unloaded the record after scheduling this fetch, in which
+      // case we must cancel the destory.  This is because we require a record
+      // to build a snapshot.  This is not fundamental: this cancelation code
+      // can be removed when snapshots can be created for internal models that
+      // have no records.
+      if (internalModel.hasScheduledDestroy()) {
+        internalModels[i].cancelDestroy();
+      }
+    }
+
     function _fetchRecord(recordResolverPair) {
       let recordFetch = store._fetchRecord(
         recordResolverPair.internalModel,
--- a/tests/integration/records/unload-test.js
+++ b/tests/integration/records/unload-test.js
@@ -2089,3 +2089,31 @@ test('unload invalidates link promises', function(assert) {
     })
   );
 });
+
+test('fetching records cancels unloading', function(assert) {
+  env.adapter.findRecord = (store, type, id) => {
+    assert.equal(type, Person, 'findRecord(_, type) is correct');
+    assert.deepEqual(id, '1', 'findRecord(_, _, id) is correct');
+
+    return  {
+      data: {
+        id: 1,
+        type: 'person'
+      }
+    }
+  };
+
+  run(() =>
+    env.store.push({
+      data: {
+        id: 1,
+        type: 'person'
+      }
+    })
+  );
+
+  return run(() =>
+    env.store.findRecord('person', 1, { backgroundReload: true })
+      .then(person => person.unloadRecord())
+  );
+});


addon/-private/system/store.js
```diff
@@ -921,6 +925,9 @@ Store = Service.extend({
       // will once again convert the records to snapshots for adapter.findMany()
       let snapshots = new Array(totalItems);
       for (let i = 0; i < totalItems; i++) {
+        if (internalModels[i].hasScheduledDestroy()) {
+          internalModels[i].cancelDestroy();
```
igorT: Why are we canceling in two places? Does the test catch both instances where canceling is required?

sly7-7:
I tested the scenario in which I reproduced #5343, and this seems to work. Thanks a lot @hjdivad

addon/-private/system/store.js
hjdivad: per OOB discussion, moving all cancelation to flush

sly7-7:
@bmac A very kindlyi-smoothy-loving ping to ship a new beta... 😺

bmac:
@sly7-7 I just published [v3.1.0-beta.2](https://github.com/emberjs/data/releases/tag/v3.1.0-beta.2).

sly7-7:
Thank you So much @bmac

phantomphildius:
@bmac thank you this also fixes the same issue we were having!

jaswilli:
I just tested this on a `2.18.2` app and it fixes it there as well. Any chance of `2.18.3` bugfix release? 😄

aalasolutions:
Please release 2.18.4 ember data with this fix. 2.18 is still in LTS branch and we need this fix

rwjblue:
@aalasolutions - FWIW, there have not been _any_ ember-data versions that are LTS. The only project that has released any LTS versions to date has been Ember itself (ember-source npm package).

aalasolutions:
@rwjblue hmm, right. meanwhile I have totally ditched unloadRecord and in my component, i filtered the records which are not required. So I guess I can live with that for now

bmac:
I just cherry-picked this commit to the 2.18 branch and release Ember Data 2.18.4.

(Sorry @rwjblue I should have pinged you to give you a heads up first).

PR #5378, opened 2018-03-15
[BUGFIX] resolve issues with RecordArray sync for peekAll

### LINKED ISSUE #5025: `FilteredRecordArray` has null values after `unloadRecord` w/ 2.13 and newer.

### LINKED ISSUE #5111: peekAll contains null entries after unloadRecord

### LINKED ISSUE #5167: multiple calls to peekAll and unloadAll put store into inconsistent state

### LINKED ISSUE #5175: 2.15.0 - `unloadAll()` not remove records from `store`
```
DEBUG: -------------------------------
DEBUG: Ember      : 2.15.0
DEBUG: Ember Data : 2.15.0
DEBUG: jQuery     : 3.2.1
DEBUG: -------------------------------
```

```
> store.peekAll('directions').get('length')
0

> store.findAll('directions')
XHR finished loading: SEARCH "https://host.ru/api/v2/rest/directions".

> store.peekAll('directions').get('length')
2

>store.unloadAll()
undefined

> store.peekAll('directions').get('length')
2
```

But when unload record-by-record - all Ok:
```
> store.peekAll('directions').get('length')
0

> store.findAll('directions')
XHR finished loading: SEARCH "https://host.ru/api/v2/rest/directions".

> store.peekAll('directions').get('length')
2

> store.peekAll('directions').objectAt(0).unloadRecord()
undefined

> store.peekAll('directions').get('length')
1

>store.peekAll('directions').objectAt(0).unloadRecord()
undefined

> store.peekAll('directions').get('length')
0
```

### LINKED ISSUE #5271: RecordArrayManager._syncLiveRecordArray trigger unexpect didChange cause EMBER_GLIMMER_BACKTRACKING_RERENDER
### Twiddle first:
[https://ember-twiddle.com/eddfe31a08d04ea9301b7a7ec77d2413](https://ember-twiddle.com/eddfe31a08d04ea9301b7a7ec77d2413)

Please run this twiddle and check console for error detail.

### Based on my research, this issue apears to be related to `RecordArrayManager.prototype._syncLiveRecordArray` method

```javascript
RecordArrayManager.prototype._syncLiveRecordArray = function _syncLiveRecordArray(array, modelName) {
    (true && !(typeof modelName === 'string') && Ember.assert("recordArrayManger.syncLiveRecordArray expects modelName not modelClass as the second param", typeof modelName === 'string'));

   
    var hasNoPotentialDeletions = Object.keys(this._pending).length === 0;
    var map = this.store._internalModelsFor(modelName);
    var hasNoInsertionsOrRemovals = get$10(map, 'length') === get$10(array, 'length');

    /*
      Ideally the recordArrayManager has knowledge of the changes to be applied to
      liveRecordArrays, and is capable of strategically flushing those changes and applying
      small diffs if desired.  However, until we've refactored recordArrayManager, this dirty
      check prevents us from unnecessarily wiping out live record arrays returned by peekAll.
      */
    if (hasNoPotentialDeletions && hasNoInsertionsOrRemovals) {
      return;
    }
    
     // MY COMMENT: hasNoPotentialDeletions is false because some async relationships are trying to fetch data
    var internalModels = this._visibleInternalModelsByType(modelName);
    var modelsToAdd = [];
    for (var i = 0; i < internalModels.length; i++) {
      var internalModel = internalModels[i];
      var recordArrays = internalModel._recordArrays;
      if (recordArrays.has(array) === false) {
        recordArrays.add(array);
        modelsToAdd.push(internalModel);
      }
    }

    // MY COMMENT: even if modelsToAdd is an empty Array, it triggers didChange event
    array._pushInternalModels(modelsToAdd);
  };

### PR DESCRIPTION
**Pull request by runspired**
This PR resolves issues with ensuring that RecordArray is correctly synced for peekAll.

Resolves #5271 
Resolves #5167 
Resolves #5175
Resolves #5111 / #5157 

It also includes tests and a fix for #5350 and #5025 / #5095  wherein non-materialized records would remain in live-record-arrays and filtered record arrays upon unload.

Resolves #5025 

cc @tylerturdenpants @igorT @hjdivad @workmanw

### DIFF
--- a/addon/-private/system/model/internal-model.js
+++ b/addon/-private/system/model/internal-model.js
@@ -293,6 +293,7 @@ export default class InternalModel {
     // models to rematerialize their records.
 
     return this._isDematerializing ||
+      this.hasScheduledDestroy() ||
       this.isDestroyed ||
       this.currentState.stateName === 'root.deleted.saved' ||
       this.isEmpty();
@@ -381,9 +382,10 @@ export default class InternalModel {
       this._isDematerializing = true;
       this._record.destroy();
       this.destroyRelationships();
-      this.updateRecordArrays();
       this.resetRecord();
     }
+
+    this.updateRecordArrays();
   }
 
   deleteRecord() {
--- a/addon/-private/system/record-array-manager.js
+++ b/addon/-private/system/record-array-manager.js
@@ -62,7 +62,6 @@ export default class RecordArrayManager {
     this._liveRecordArrays = Object.create(null);
     this._pending = Object.create(null);
     this._adapterPopulatedRecordArrays = [];
-    this._nextManagerFlush = null;
   }
 
   recordDidChange(internalModel) {
@@ -94,7 +93,7 @@ export default class RecordArrayManager {
       return;
     }
 
-    this._nextManagerFlush = emberRun.schedule('actions', this, this._flush);
+    emberRun.schedule('actions', this, this._flush);
   }
 
   _flushPendingInternalModelsForModelName(modelName, internalModels) {
--- a/addon/-private/system/record-array-manager.js
+++ b/addon/-private/system/record-array-manager.js
@@ -62,6 +62,7 @@ export default class RecordArrayManager {
     this._liveRecordArrays = Object.create(null);
     this._pending = Object.create(null);
     this._adapterPopulatedRecordArrays = [];
+    this._nextManagerFlush = null;
   }
 
   recordDidChange(internalModel) {
@@ -93,7 +94,42 @@ export default class RecordArrayManager {
       return;
     }
 
-    emberRun.schedule('actions', this, this._flush);
+    this._nextManagerFlush = emberRun.schedule('actions', this, this._flush);
+  }
+
+  _flushPendingInternalModelsForModelName(modelName, internalModels) {
+    let modelsToRemove = [];
+
+    for (let j = 0; j < internalModels.length; j++) {
+      let internalModel = internalModels[j];
+      // mark internalModels, so they can once again be processed by the
+      // recordArrayManager
+      internalModel._pendingRecordArrayManagerFlush = false;
+      // build up a set of models to ensure we have purged correctly;
+      if (internalModel.isHiddenFromRecordArrays()) {
+        modelsToRemove.push(internalModel);
+      }
+    }
+
+    // process filteredRecordArrays
+    if (this._filteredRecordArrays[modelName]) {
+      let recordArrays = this.filteredRecordArraysFor(modelName);
+      for (let i = 0; i < recordArrays.length; i++) {
+        this.updateFilterRecordArray(recordArrays[i], modelName, internalModels);
+      }
+    }
+
+    let array = this._liveRecordArrays[modelName];
+    if (array) {
+      // TODO: skip if it only changed
+      // process liveRecordArrays
+      this.updateLiveRecordArray(array, internalModels);
+    }
+
+    // process adapterPopulatedRecordArrays
+    if (modelsToRemove.length > 0) {
+      removeFromAdapterPopulatedRecordArrays(modelsToRemove);
+    }
   }
 
   _flush() {
@@ -101,40 +137,9 @@ export default class RecordArrayManager {
 
     let pending = this._pending;
     this._pending = Object.create(null);
-    let modelsToRemove = [];
 
     for (let modelName in pending) {
-      let internalModels = pending[modelName];
-      for (let j = 0; j < internalModels.length; j++) {
-        let internalModel = internalModels[j];
-        // mark internalModels, so they can once again be processed by the
-        // recordArrayManager
-        internalModel._pendingRecordArrayManagerFlush = false;
-        // build up a set of models to ensure we have purged correctly;
-        if (internalModel.isHiddenFromRecordArrays()) {
-          modelsToRemove.push(internalModel);
-        }
-      }
-
-      // process filteredRecordArrays
-      if (this._filteredRecordArrays[modelName]) {
-        let recordArrays = this.filteredRecordArraysFor(modelName);
-        for (let i = 0; i < recordArrays.length; i++) {
-          this.updateFilterRecordArray(recordArrays[i], modelName, internalModels);
-        }
-      }
-
-      let array = this._liveRecordArrays[modelName];
-      if (array) {
-        // TODO: skip if it only changed
-        // process liveRecordArrays
-        this.updateLiveRecordArray(array, internalModels);
-      }
-
-      // process adapterPopulatedRecordArrays
-      if (modelsToRemove.length > 0) {
-        removeFromAdapterPopulatedRecordArrays(modelsToRemove);
-      }
+      this._flushPendingInternalModelsForModelName(modelName, pending[modelName]);
     }
   }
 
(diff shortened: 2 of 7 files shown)

### CODE REVIEW

**Comment by tylerturdenpants** [comment]
@runspired, it works so far!  I'll check back in when I get more coverage of my app

**Comment by HeroicEric** [comment]
I tested this with our app that has been stuck at 2.13.2 for a while and this lets us go all the way to 3.x 😄

**Comment by sly7-7** [comment]
I really have to dig and try to narrow my problem then, I still encounter a 
```
ember-console.js:29 TypeError: Cannot read property 'eachAttribute' of null
    at new Snapshot (-private.js:5377)
    at InternalModel.createSnapshot (-private.js:7173)
    at _findHasMany (-private.js:9104)
    at Class.findHasMany (-private.js:11343)
    at ManyRelationship.fetchLink (-private.js:4871)
    at ManyRelationship.findLink (-private.js:4132)
    at ManyRelationship.getRecords (-private.js:4913)
    at Class.get (-private.js:13357)
    at ComputedPropertyPrototype.get (ember-metal.js:4064)
    at get (ember-metal.js:3436)
```
In my scenario, there are several unloadAll and a call to a route.refresh whereas models are still displayed on the screen.

runspired:
@sly7-7 hrm :/ Makes me suspicious that #5376 has not caught everything it needed to.

tylerturdenpants:
@sly7-7 I had the same issue way back in 2.13. I think I resolved it by routing away from where the models were in use and then performed the unload. 

As an aside, @runspired, what will it take to get this merged?  This has been the most promising fix in the last few months. I owe you big time. You’re a huge value to this community.

btecu:
Will this make it into 3.1?

sly7-7:
@tylerturdenpants 
`@sly7-7 I had the same issue way back in 2.13. I think I resolved it by routing away from where the models were in use and then performed the unload.`

Thank you for the tip, I will see if I can try that as a workaround, but for sure there is 'something' wrong 'somewhere'

runspired:
@sly7-7 I have some thoughts on what your bug is, will try to write a test to see if it fails.

Kilowhisky:
Found a related issue: #5395

tylerturdenpants:
@bmac In what version is this slated to land?

sly7-7:
@tylerturdenpants It should be in the just released 3.1. You will encounter a warning though: https://github.com/emberjs/data/pull/5372#issuecomment-376305559 but as @bmac said patch will be soon released too 😄

tylerturdenpants:
Oh nice @sly7-7. Wish you luck on your issues. I’m still encountering the “Id already used” but is much easier to patch.

bmac:
Since this was just merged I didn't cherry pick it into 3.1 just yet. Currently its only on `3.2.0-beta.1` but I'd be willing to make a patch release to 3.1 in a few days if nothing major gets reported on the beta release.

PR #5380, opened 2018-03-16
Remove use of `modules` directory for ember-cli >= 3.0

runspired:
ember-cli no longer requires the modules directory to be used and will deprecate it soon

cc @rwjblue @kellyselden


twokul:
lgtm

Issue #5253, opened 2017-11-08
Ember Data FEATURES aren't merged with Ember FEATURES


alexander-alvarez:
There a feature in `ds-check-should-serialize-relationships` which has been `true` since `ember-data 2.12` https://github.com/emberjs/data/blob/v2.12.0/config/features.json#L7

My interpretation of https://guides.emberjs.com/v2.16.0/configuring-ember/feature-flags/#toc_flagging-details leads me to believe that I should be able to use said feature in a non-beta (and non-canary) release of Ember. 

I haven't gotten this to work

https://github.com/alexander-alvarez/ds-features-not-enabled/commit/d89e7f3b9dd3884db3284b652cb071e4bbf59928 **reproduces** it on `ember: 2.15.x` and `ember-data: 2.15.x`

**My diagnosis**

This check https://github.com/emberjs/data/blob/master/addon/serializers/json-api.js#L582 fails

When I dump `Ember.FEATURES` I get
```json
{
  "ember-libraries-isregistered":false,
  "ember-improved-instrumentation":false,
  "ember-metal-weakmap":false,
  "ember-glimmer-allow-backtracking-rerender":false,
  "ember-routing-router-service":true,
  "ember-engines-mount-params":true,
  "ember-module-unification":false,
  "glimmer-custom-component-manager":false,
  "mandatory-setter":true,
  "ember-glimmer-detect-backtracking-rerender":true
}
```
which are the features enabled in https://github.com/emberjs/ember.js/blob/v2.15.1/features.json

Should it not also contain the features in ember-data's `features.json`?

runspired:
@rwjblue we removed all of our feature flags, but we should still get this fixed before we add any new ones.

runspired:
Feature flag infra was later improved to not require merging with Ember's.

PR #5384, opened 2018-03-16
[FEAT] remove all stale feature flags

runspired:
As discussed in the face-to-face, our feature flags have gone very stale and we
we decided to go a different direction on each. This is a debt-collection PR to remove
our stale feature flags and reset the feature-flag clock.

All feature flags that were previously `null` were removed with their `off` branch left behind.
All feature flags that were previously `true` were removed with their `on` branch left behind.


runspired:
Starting a list of tickets to update when this merges:

- https://github.com/emberjs/data/issues/5253

jonnii:
@runspired sad about `ds-pushpayload-return`, it's pretty useful for people who use websockets =(

runspired:
@jonnii are you saying you are using that feature!?

Are you aware of `store.push` ?

jonnii:
@runspired I didn't realize `push` returned the record it created, you can ignore my comment!

urbany:
Related: https://github.com/emberjs/data/pull/5265

sly7-7:
@runspired All green, so I guess this is good to merge ?

runspired:
@sly7-7 an example websocket push that is much less brittle than even the `pushPayload` use today

```
// ... payload is raw payload received
let store = this.get('store');
let ModelClass = store.modelFor('foo');
let serializer = store.serializerFor('foo');
let normalized = store.normalizeResponse(store, ModelClass, payload, null, 'query');

return store.push(normalized);
```

PR #5386, opened 2018-03-17
[FEAT fetch] use ember-fetch instead of jQuery

### LINKED ISSUE #5320: [QUEST] Making jQuery Optional
This issue tracks emberjs/ember.js#16058 for Ember Data.

Currently, ember.js is in the progress of making jQuery optional. This does not deprecate jQuery in Ember, but it makes using Ember without jQuery a first-class option. Today, it's possible, but the Ember tests don't run without jQuery, and users need to keep track themselves of which APIs might inadvertently rely on jQuery.

This issue tracks making jQuery optional in the Ember Data component of Ember.

As a general principle, there are a few different kinds of reliance on jQuery:

- Actually using jQuery internally. For internal-only use, we should switch to a standard API (like `fetch()` in Ember Data).
- Reliance on details of jQuery as a result of the fact that we're using jQuery internally. For example, because Ember Data uses jQuery.ajax internally, people can use `ajaxSend` and other public jQuery APIs to configure the fetch. In this case, the strategy we're using is to keep around the jQuery code, but allow users to opt into a more standard mode explicitly. In general, we would like to try to detect when users are attempting to use jQuery in development mode and give them good assertions.

So in short, we need to remove the use of jQuery in purely internal places, and create a mode for using Ember Data without jQuery on an opt-in basis in places where jQuery is leaked into the public API.

To get started, I'd recommend using an internal flag like jQueryDisabled that you can use as a conditional to opt into the no-jquery semantics. Ember Core will soon create a flag you can use explicitly (we need an RFC) that you can swap to as soon as we have it. But there's no reason to wait.

@tchak would you mind commenting with concrete steps we need here, and I'll add them to the issue?

---

Work to be done:

- [x] Remove jQuery usage from tests
- [x] Remove use of `$.parseJSON`
- [x] Figure out how to feature-flag an optional dependency (cc @rwjblue)
  - [x] avoid introducing `ember-fetch` bloat to people opting out of jQuery
- [x] Create a new feature flag for using fetch in place of `$.ajax`
- [x] Figure out how to avoid introducing unnecessary bloat

tchak:
depends on #5375, #5385 and #5375 
will fix #5320

### DIFF
--- a/addon/adapters/rest.js
+++ b/addon/adapters/rest.js
@@ -4,13 +4,17 @@
   @module ember-data
 */
 
-import $ from 'jquery';
+import Ember from 'ember';
+import fetch, { Response } from 'fetch';
+//FIXME: We need to add a direct export of `serializeQueryParams`
+import { serializeQueryParams } from 'ember-fetch/mixins/adapter-fetch';
 
 import { Promise as EmberPromise } from 'rsvp';
 import { get, computed } from '@ember/object';
 import { getOwner } from '@ember/application';
 import { run } from '@ember/runloop';
 import Adapter from '../adapter';
+import { assign } from '@ember/polyfills';
 import {
   parseResponseHeaders,
   BuildURLMixin,
@@ -24,7 +28,6 @@ import {
   TimeoutError,
   AbortError,
 } from '../-private';
-import { instrument } from 'ember-data/-debug';
 import { warn } from '@ember/debug';
 import { DEBUG } from '@glimmer/env';
 
@@ -296,6 +299,11 @@ const RESTAdapter = Adapter.extend(BuildURLMixin, {
 
   fastboot: computed(function() {
     return getOwner(this).lookup('service:fastboot');
+  }),
+
+  useFetch: computed(function() {
+    let ENV = getOwner(this).resolveRegistration('config:environment');
+    return ((ENV && ENV._JQUERY_INTEGRATION) === false) || (Ember.$ === undefined);
   }),
 
   /**
@@ -983,16 +991,33 @@ const RESTAdapter = Adapter.extend(BuildURLMixin, {
     let hash = adapter.ajaxOptions(url, type, options);
 
     return new Promise(function(resolve, reject) {
-      hash.success = function(payload, textStatus, jqXHR) {
-        heimdall.stop(token);
-        let response = ajaxSuccessHandler(adapter, payload, jqXHR, requestData);
-        run.join(null, resolve, response);
-      };
-      hash.error = function(jqXHR, textStatus, errorThrown) {
-        heimdall.stop(token);
-        let error = ajaxErrorHandler(adapter, jqXHR, errorThrown, requestData);
-        run.join(null, reject, error);
-      };
+      if (get(this, 'useFetch')) {
+        hash.success = function(response) {
+          heimdall.stop(token);
+          determineBodyPromise(response, requestData).then(payload => {
+            let response = fetchSuccessHandler(adapter, payload, response, requestData);
+            run.join(null, resolve, response);
+          });
+        };
+        hash.error = function(response, errorThrown) {
+          heimdall.stop(token);
+          determineBodyPromise(response, requestData).then(payload => {
+            let error = fetchErrorHandler(adapter, payload, response, errorThrown, requestData);
+            run.join(null, reject, error);
+          });
+        };
+      } else {
+        hash.success = function(payload, textStatus, jqXHR) {
+          heimdall.stop(token);
+          let response = ajaxSuccessHandler(adapter, payload, jqXHR, requestData);
+          run.join(null, resolve, response);
+        };
+        hash.error = function(jqXHR, textStatus, errorThrown) {
+          heimdall.stop(token);
+          let error = ajaxErrorHandler(adapter, jqXHR, errorThrown, requestData);
+          run.join(null, reject, error);
+        };
+      }
 
       adapter._ajax(hash);
     }, 'DS: RESTAdapter#ajax ' + type + ' to ' + url);
@@ -1004,7 +1029,7 @@ const RESTAdapter = Adapter.extend(BuildURLMixin, {
     @param {Object} options jQuery ajax options to be used for the ajax request
   */
   _ajaxRequest(options) {
-    $.ajax(options);
+    Ember.$.ajax(options);
   },
 
   /**
@@ -1022,8 +1047,20 @@ const RESTAdapter = Adapter.extend(BuildURLMixin, {
     }
   },
 
+  _fetchRequest(options) {
+    fetch(options.url, options).then(response => {
+      if (response.ok) {
+        options.success(response);
+      } else {
+        options.error(response);
+      }
+    }).catch(error => options.error(new Response(), error));
+  },
+
   _ajax(options) {
-    if (get(this, 'fastboot.isFastBoot')) {
+    if (get(this, 'useFetch')) {
+      this._fetchRequest(options);
+    } else if (get(this, 'fastboot.isFastBoot')) {
       this._najaxRequest(options);
     } else {
       this._ajaxRequest(options);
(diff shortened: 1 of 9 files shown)


addon/adapters/rest.js
```diff
@@ -4,12 +4,16 @@
   @module ember-data
 */
 
-import $ from 'jquery';
+import Ember from 'ember';
+import fetch, { Response } from 'fetch';
+//FIXME: We need to add a direct export of `serializeQueryParams`
+import { serializeQueryParams } from 'ember-fetch/mixins/adapter-fetch';
```
nlfurniss: I've been wondering where the best place for this to live is. In a jQuery-less world, `$.param()` is gone and Ember should have some replacement.

tchak: I feel like `ember-fetch` is good place for now. I am planing on doing a PR to move it out of `ember-fetch/mixins/adapter-fetch`. But a broader discussion is needed about `fetch` usage: `fetch` on it own is not a suitable replacement for `$.ajax` in my opinion. `fetch` is too low level.

tchak: Another option might be to use `URLSearchParams` provided by `ember-url` (with polyfills). It is a standard, but might be less convenient then a `$.param()`.

btecu:
@tchak What's the status? Seems to be the last step for being able to use ED without jQuery.

package.json
```diff
@@ -48,6 +48,7 @@
     "ember-cli-string-utils": "^1.1.0",
     "ember-cli-test-info": "^1.0.0",
     "ember-cli-version-checker": "^2.1.2",
+    "ember-fetch": "^5.1.3",
```
runspired: I think we should we make this a `peerDependency` instead to avoid trolling folks with multiple versions, and a `devDependency` for our own use.

addon/adapters/rest.js
NullVoxPopuli: what's better? `URLSearchParams` from `ember-url`, or `serializerQueryParams` from the fetch adapter?

runspired:
@tchak @nullvoxpopuli think this needs updated for code style, there's also some broken tests

NullVoxPopuli:
I'll take a look. :)

addon/adapters/rest.js
tchak: Using `URLSearchParams` is not an option after investigation – it won't serialise nested params. I think wee should expose something like `import { param } from 'ember-fetch';`

tchak:
I probably won't be available today match, but I should be able to help if needed tomorrow

addon/adapters/rest.js
NullVoxPopuli: https://github.com/ember-cli/ember-fetch/pull/130

tchak:
There is a couple of external tests failing. Ember Observer seems like a timeout. Ilios Frontend looks like some genuine errors. I will have to look in to it more closely.

runspired:
@tchak may also be flakiness, it passed when I restarted the job. cc @jrjohnson

jrjohnson:
I can't see the failure, but we've been struggling terribly with flaky tests the last couple of days for reasons we haven't tracked down yet.  If the error was something like https://github.com/ilios/frontend/issues/4046 then this may be the issue.

NullVoxPopuli:
build passed! (but hasn't reported back to github for some reason)

runspired:
@NullVoxPopuli anything remaining? I noticed this:

> I think there should probably be a good amount of tests for fetch, which don't exist yet.
also, with ember fetch, there is this function: mungOptionsForFetch
> 
> So, I think for this feature to be complete, we need to find all the request url / headers / whatever related tests and make fetch versions of them.

package.json
NullVoxPopuli: this still needs to be removed from here. this was probably my bad when I did partial work on this.

NullVoxPopuli:
@runspired good catch -- 

this file: https://github.com/emberjs/data/pull/5386/files#diff-038a1249383849dd6526ee97d193e925R90

the ajax file needs a fetch equivalent.

also, are coverage artifacts outputted somewhere? my _hunch_ is that a lot of the new fetch stuff is missing most coverage

NullVoxPopuli:
tests and ember-fetch dep removed: https://github.com/tchak/data/pull/3

tchak:
Thank you @NullVoxPopuli! PR rebased

NullVoxPopuli:
is anything pending on this?

tchak:
I think we should land https://github.com/ember-cli/ember-fetch/pull/130 before this. If not we will create a dependency hell where some version of ED will require older versions of fetch than others.

tchak:
I think we should land https://github.com/ember-cli/ember-fetch/pull/130 before this. If not we will create a dependency hell where some version of ED will require older versions of fetch than others.

tchak:
I think we should land https://github.com/ember-cli/ember-fetch/pull/130 before this. If not we will create a dependency hell where some version of ED will require older versions of fetch than others.

addon/adapters/rest.js
NullVoxPopuli: @tchak ember-fetch has been updated!

nlfurniss:
@tchak ember-cli/ember-fetch#130 landed in v6.0.0

runspired:
@tchak now that that fetch PR has landed, do we need to specify somewhere here that 6.0 is the min viable version? Anything we need to update? If not, I'll look at fixing the lock file and landing this sweetness :)

tchak:
@runspired I just rebased the branch and bumped `ember-fetch` to v6. I am not entirely sure how we can enforce the minimum fetch version.

mansona:
@tchak I know it might seem a bit hacky but what if you did something in the index.js that used [resolve](https://www.npmjs.com/package/resolve) to check what version ember-fetch is actually installed and throw an error if it's too low?

NullVoxPopuli:
do peerDependencies not solve this?

mansona:
@NullVoxPopuli I think it's just a warning when it comes to peerDependencies 🤔 but I could be wrong

runspired:
@xg-wang is correct, and likely that’s what we should do.

runspired:
I’m a bit hesitant to do something that forces folks to install ember-fetch as much as I want to eliminate jQuery. We should likely make whether it is required “optional” and only use version checker to enforce if the flag is not set to drop it.

tchak:
Si it should throw an error only if no `jQuery` support is enabled and `ember-fetch` is absent or lower than required version, correct?

runspired:
@tchak I was thinking more broad: as in it should likely be the default that without jquery we expect fetch, but folks should be able to opt out of both.

xg-wang:
```
shouldIncludeChildAddon(childAddon) {
      if (childAddon.name === 'ember-fetch') {
        return !HAS_JQUERY;
      } else {
        return this._super.shouldIncludeChildAddon.apply(this, arguments);
      }
    }
```
I think this could do it. To check top level & sibling ember-fetch existence should be done in ember-fetch self's cli hook.

dcyriller:
> I’m a bit hesitant to do something that forces folks to install ember-fetch as much as I want to eliminate jQuery
> folks should be able to opt out of both

@runspired how do you see the opt out strategy? I guess the fallback would be on a non-polyfilled version of fetch (what about fastboot?). In `config/environment.js`, we could toggle the native fetch strategy this way?
```js
module.exports = function(environment) {
  var ENV = {
    'ember-data': {
      useNativeFetch: true // defaults to false
    }
  }
}
```

@tchak I've done a PR (targeting yours) to throw an error in case of dependency mismatch.

simonihmig:
The quest to remove jQuery by default as per [RFC386](https://github.com/emberjs/rfcs/pull/386) is making [good progress](https://github.com/emberjs/ember.js/issues/17476). Part of it is also resolving this issue. 

So is there a chance we can land this in the near future? Ideally in a forthcoming 3.9 release? (which is the version of Ember which will add the additional jQuery deprecations)

NullVoxPopuli:
where can we view the status of this?

runspired:
> @runspired how do you see the opt out strategy? I guess the fallback would be on a non-polyfilled version of fetch (what about fastboot?). In config/environment.js, we could toggle the native fetch strategy this way?

@dcyriller I think i'm going to skip on this requirement for now and add it as a feature later if folks need. Is all that is needed to land a rebase at this point?

runspired:
Closing in favor of #5900

Issue #5395, opened 2018-03-26
unloadRecord() jumps the gun on peekAll arrays using dependentArrayWillChange observers


Kilowhisky:
I'm not sure exactly who's at fault here or if anything can be done but it most closely appears to be ember data's live record array's fault.

If you have a live record array recovered from using a peekAll

`store.peekAll('comments')`

and you attach array observers to the peekAll array
```
this.dependentArray.addArrayObserver(this, {
            willChange: 'dependentArrayWillChange',
            didChange: 'dependentArrayDidChange'
        });
```

When you issue an unload request over more than 1 record
```
while (countToRemove > 0) {
                let record = orderedBy.popObject();
                        store.unloadRecord(record);
                    });
                    countToRemove--;
 }
```

The `dependentArrayWillChange` will be called for every single removal BUT after the first call the dependent array will have the records to be remove set to null.

```
dependentArrayWillChange(observedArray, start, removeCount /*, addCount */) {
            if (removeCount > 0) {
                console.log(observedArray, start, removeCount, observedArray.get('length'));
                console.log(observedArray.toArray());
                let removedSlice = observedArray.slice(start, start + removeCount);
                console.log(removedSlice); 
            }
        }
```

In the above code the first removal will properly have the record that was removed. The second time the method is called the removed slice and any other records that were pending to be removed will be set to `null` or `undefined`.

So if you had a array of 5 and you unload 3 of them the `dependentArrayWillChange` will fire 3 times. The first time the array will be in an unmodified state as the change has not occurred. The second time it is called the `sourceArray` will look like this

0: {object}
1: {object}
2: null
3: null
4: null or undefined (which is odd)

After the whole run loop completes the arrays are cleaned up and the nulls are removed. But at the time the array observers are fired the array state is all sorts of wrong. 

By the way If i run my unloads in their own run loop the problem is solved
```
while (countToRemove > 0) {
                let record = orderedBy.popObject();
                Ember.run(() => {
                      store.unloadRecord(record);
                });
                countToRemove--;
 }
```

Here is what i see going on. My guess is ember is batching unloads or batching observer calls.

1) Fully populated live record array of 5 records is found.
2) App asks in a while loop to unload 3 of the 5 records
3) ???
4) ArrayWillChange is fired for the first record to be removed
5) ArrayDidChange is fired and shows that 3 of the 5 records are now set to null
6) ArrayWillChange is fired for the second record to be removed
7) ArrayDidChange is fired for the second record to be removed
[...]

PR #5400, opened 2018-03-27
add test for  #5395

Kilowhisky:
Adds failing test for #5395

### DIFF
--- a/tests/integration/record-arrays/peeked-records-test.js
+++ b/tests/integration/record-arrays/peeked-records-test.js
@@ -431,3 +431,72 @@ test('unloading filtered records', function(assert) {
   assert.equal(get(people, 'length'), 1, 'Unloaded record removed from the array');
   assert.equal(get(people.objectAt(0), 'name'), 'Scumbag Joe', 'Joe shifted down after the unload');
 });
+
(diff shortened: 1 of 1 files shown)


runspired:
As ember has deprecated array observers in 4.0, and as there are many patterns for reacting to record-array changes efficiently without array-observers, and as the relationship layer has been refactored to generally be far more efficient causing over notifications to go down (the source of some of the trouble leading to the issue for which this PR was opened), and as the craziness of unloadRecord is mostly being eliminated due to internal conversion to identifiers (such that unloaded records can always be fully released), I'm going to close this.

Long winded I know but that's because fundamentally I care about why this PR was opened in the first place: performant array operations matter a lot. There are a lot of good patterns out there though that don't require the array-observer, and we're dedicated to improving base performance where possible. If @Kilowhisky or anyone else watching this issue still faces performance problems using octane paradigms and ember-data 3.28+ please reach out. At a minimum we should create a performance benchmark to add to the performance tests we run in CI so that we can monitor and focus on improving the base performance of the scenario, and if it's evident more is needed to achieve the necessary performance characteristics for large data sets then we'll look into what we can do to make subscribing to granular updates for a type first-class.

PR #5402, opened 2018-03-28
Document changing the request method.

rimian:
The `methodForRequest` method does not seem to be documented here! 

I'm not sure how the documentation works exactly. Is this enough?

### DIFF
--- a/addon/adapters/rest.js
+++ b/addon/adapters/rest.js
@@ -284,10 +284,10 @@ const Promise = EmberPromise;
   });
   ```
   ### Method Customization
-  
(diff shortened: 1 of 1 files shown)


runspired:
@rimian Thank you so much for taking the time to write some docs for us!

Unfortunately, `methodForRequest` is only available for the now removed `ds-improved-ajax` feature flag. If you'd like to PR it to the new (temporary) home for this feature flag while we plot a new course, you can find that under development here: https://github.com/runspired/ds-improved-ajax

PR #5403, opened 2018-03-28
[Model Data] Fix belongs to relationship when `didCommit` with null

dnachev:
Not sure this is the correct fix, but it fixes the problem for me.


addon/-private/system/relationships/state/belongs-to.js
```diff
@@ -84,6 +84,7 @@ export default class BelongsToRelationship extends Relationship {
     //temporary fix to not remove newly created records if server returned null.
     //TODO remove once we have proper diffing
     if (this.inverseModelData && this.inverseModelData.isNew() && !this.canonicalState) {
+      this.willSync = false;
```
dnachev: The `willSync` flag is maintained in the super `Relationship` class and it will be reset when `flushCanonical` is called. However, in this special case, where the relationship is new and the server don't return any value for the relationship, the flag will remain as `willSync = true` and any subsequent attempts to `flushCanonical` the relationship will be ignored.

runspired:
The failing test is the optional feature for belongsTo embedded. I suspect rebasing the model-data branch after landing @bmac's feature flag test removal work would clean this up, but I'll dig in deeper.

runspired:
The failing test is fixed on master

PR #5404, opened 2018-03-28
Deprecate `store.updateId`

hjdivad:
It was always private; so this is a cautious deprecation.

### DIFF
--- a/addon/-private/system/store.js
+++ b/addon/-private/system/store.js
@@ -1975,7 +1975,10 @@ Store = Service.extend({
   },
 
   updateId(internalModel, data) {
-    // TODO Deprecate this? Completely remove it?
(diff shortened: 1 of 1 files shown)


runspired:
I re-kicked the scenario that hit a timeout. The failed test on the ember-canary scenario is fixed on master already.

addon/-private/system/store.js
```diff
@@ -1965,6 +1965,10 @@ Store = Service.extend({
 
   updateId(internalModel, data) {
     // TODO Deprecate this? Completely remove it?
```
rwjblue: Can / should we kill this comment too (now that we have done what it suggests)?

hjdivad:
test failures are the ones already on the branch that are being worked on.

PR #5406, opened 2018-03-30
[BUGFIX release] Fix `Model.modelName` inheritance with Ember 3.2+.

rwjblue:
On Ember 3.2 and higher, Ember uses "real" inheritance when an `Ember.Object` is `.extend()`ed. This means that in 3.2 and higher the `ModelClassHere.modelName` property is now being inherited (where previously each `.extend()` reset the value to `null`).

The fix here ensures that each `ModelClass.extend()` gets its own `modelName`...


runspired:
Changing the check to the below will fix the tests for pre ember canary:

```js
if (!klass.modelName || !klass.hasOwnProperty('modelName')) {
        klass.modelName = modelName;
      }
```

PR #5231, opened 2017-10-16
failing tests for record being removed from store on destroyRecord

fivetanley:
tests for #4995

PR #5408, opened 2018-04-02
[FEAT BUGFIX] enable canonical state updates to deleted records

### LINKED ISSUE #4995: destroyRecord regression in 2.13.x
I started experiencing a regression after upgrading to 2.13.x from 2.12.2. The error happens when calling destroyRecord, then later querying records. Here is a test case route that reproduces the issue for me:

```
  model() {
    this.store.findRecord('my-model', '123')
      .then(m => m.destroyRecord())
      .then(() => this.store.query('my-model', {}));
  }
});
```

The gotcha in our system is that the DELETE handler does not immediately delete the record, but marks the record for deletion. So it will still be returned in query calls until it's finished being deleted.

The error that I am getting in 2.13.1 is
```
Attempted to handle event `pushedData` on <my-model:123> while in state root.deleted.saved. 

Error
    at InternalModel._unhandledEvent (http://localhost:4200/ui/assets/vendor-541f5b0fb0f07999ce5d23075731eb0c.js:140691:13)
    at InternalModel.send (http://localhost:4200/ui/assets/vendor-541f5b0fb0f07999ce5d23075731eb0c.js:140560:14)
    at InternalModel.pushedData (http://localhost:4200/ui/assets/vendor-541f5b0fb0f07999ce5d23075731eb0c.js:140502:12)
    at InternalModel.setupData (http://localhost:4200/ui/assets/vendor-541f5b0fb0f07999ce5d23075731eb0c.js:140465:12)
    at Class._load (http://localhost:4200/ui/assets/vendor-541f5b0fb0f07999ce5d23075731eb0c.js:148474:21)
    at Class._pushInternalModel (http://localhost:4200/ui/assets/vendor-541f5b0fb0f07999ce5d23075731eb0c.js:148836:32)
    at http://localhost:4200/ui/assets/vendor-541f5b0fb0f07999ce5d23075731eb0c.js:148779:39
    at Backburner.run (http://localhost:4200/ui/assets/vendor-541f5b0fb0f07999ce5d23075731eb0c.js:17614:23)
    at Backburner.join (http://localhost:4200/ui/assets/vendor-541f5b0fb0f07999ce5d23075731eb0c.js:17640:23)
    at Class._push (http://localhost:4200/ui/assets/vendor-541f5b0fb0f07999ce5d23075731eb0c.js:148763:52)
```

runspired:
resolves #4995
replaces #5231 

**Problem**

Records in the `root.deleted.saved` state currently error if new data is received from the server for the record while the record is still in the store.

For instance, a user may desire the following flow, which is now possible with this PR:

```
=> signal deletion (send actual `DELETE` request)
=> update ui state to isDeleting
=> receive notice of actual deletion
=> update ui state to isDeleted
=> unload
```

**Use Case**

Sometimes deletions are "soft deletes", and sometimes deletions may be queued on the server and the deleted state not reflect immediately in other requests.

**2.13 Regression**

Prior to 2.13 we would unload records whose deletion had been persisted. While this enabled folks to push server updates back into the store for either use case above, this would create a new record in the `root.loaded.saved` state, which is also unexpected.

**This Fix**

This fix enables canonical (but not dirty) state updates to `root.deleted.saved` records to still occur.  Records are still available to the UI until the user opts to call `unloadRecord`. This means that end users may themselves decide whether they want to unload and treat these records as `undeleted` again, or maintain the deleted state.

**Soft Delete**

User's desiring `soft delete` aka `archiving` with the ability to resurrect a model should not be using `deleteRecord` or `destroyRecord`. While this PR will make soft-deletions via this mechanism work *slightly* better up until the point that resurrection is desired, users should instead fire an action to update a flag of their own making, and only fire a DELETE for a permanent deletion.

### DIFF
--- a/tests/unit/model-test.js
+++ b/tests/unit/model-test.js
@@ -16,11 +16,13 @@ module('unit/model - DS.Model', {
   beforeEach() {
     Person = DS.Model.extend({
       name: DS.attr('string'),
-      isDrugAddict: DS.attr('boolean')
+      isDrugAddict: DS.attr('boolean'),
+      isArchived: DS.attr()
     });
     Person.toString =  () => 'person';
 
     env = setupStore({
+      adapter: DS.JSONAPIAdapter,
       person: Person
     });
     store = env.store;
@@ -564,6 +566,52 @@ test('supports pushedData in root.deleted.uncommitted', function(assert) {
     store.push(hash);
     assert.equal(get(record, 'currentState.stateName'), 'root.deleted.uncommitted',
       'record accepts pushedData is in root.deleted.uncommitted state');
+  });
+});
+
+test('supports pushedData in root.deleted.saved', function(assert) {
+  let { adapter } = env;
+
+  adapter.shouldBackgroundReloadRecord = () => false;
+  adapter.deleteRecord = () => {
+    return Ember.RSVP.resolve();
+  };
+
+  let record = run(() => store.push({
+    data: {
+      type: 'person',
+      id: '1',
+      attributes: {
+        isArchived: false
+      }
+    }
+  }));
+
+  run(() => {
+    record.destroyRecord().then(() => {
+      let currentState = record._internalModel.currentState;
+
+      assert.ok(currentState.stateName === 'root.deleted.saved',
+        'record is in a persisted deleted state');
+      assert.equal(get(record, 'isDeleted'), true);
+      assert.ok(store.peekRecord('person', '1') !== null, 'the deleted person is not removed from store (no unload called)');
+    });
+  });
+
+  run(() => {
+    try {
+      store.push({
+        data: {
+          type: 'person',
+          id: '1',
+          attributes: {
+            isArchived: true
+          }
+        }
+      })
+    } catch (e) {
+      assert.ok(false, e);
+    }
   });
 });
 
--- a/tests/unit/model-test.js
+++ b/tests/unit/model-test.js
@@ -599,19 +599,15 @@ test('supports canonical updates via pushedData in root.deleted.saved', function
   });
 
   run(() => {
-    try {
-      store.push({
-        data: {
-          type: 'person',
-          id: '1',
-          attributes: {
-            isArchived: true
-          }
+    store.push({
+      data: {
+        type: 'person',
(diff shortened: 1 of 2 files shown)


tests/unit/model-test.js
```diff
@@ -567,6 +569,52 @@ test('supports pushedData in root.deleted.uncommitted', function(assert) {
...
+        data: {
+          type: 'person',
+          id: '1',
+          attributes: {
+            isArchived: true
+          }
+        }
+      })
+    } catch (e) {
+      assert.ok(false, e);
```
hjdivad: what does this try/catch achieve?

tests/unit/model-test.js
```diff
@@ -567,6 +569,106 @@ test('supports pushedData in root.deleted.uncommitted', function(assert) {
...
+      assert.ok(store.peekRecord('person', '1') !== null, 'the deleted person is not removed from store (no unload called)');
+    });
+  });
+
+  run(() => {
+    try {
+      set(record, 'isArchived', true);
+      assert.ok(false, 'Was unable to dirty a deleted record');
+    } catch (e) {
+      assert.ok(true, e.message);
```
hjdivad: let's assert the specific error rather than merely that there was an error

tests/unit/model-test.js
```diff
@@ -567,6 +569,106 @@ test('supports pushedData in root.deleted.uncommitted', function(assert) {
...
+        data: {
+          type: 'person',
+          id: '1',
+          attributes: {
+            isArchived: true
+          }
+        }
+      });
+    } catch (e) {
+      assert.ok(false, e);
```
hjdivad: what does this try/catch achieve?

tests/unit/model-test.js
runspired: Will do

tests/unit/model-test.js
runspired: I think I can use `expectAssertion` here instead

runspired: @hjdivad oh wait, actually, I can't unless there's a corollary? I basically want a "expect no assertion"

PR #5409, opened 2018-04-03
Failing: Pushing a polymorphic belongsTo in a has many relationship

sly7-7:


tests/integration/relationships/has-many-test.js
```diff
@@ -1528,6 +1528,25 @@ test("Polymorphic relationships with a hasMany is set up correctly on both sides
...
+  var post, comment;
+
+  run(function () {
+    post = env.store.push({ data: { type: 'post', id: 1 } });
+    comment = env.store.push({ data: { type: 'comment', id: 1 } });
+    let comments = post.get('comments');
+    comments.pushObject(comment);
+  });
+
+  assert.equal(comment.get('message'), post, 'The polymorphic belongsTo is set up correctly');
```
runspired: Unfortunately, `assert.equal` uses an equality algorithm that totally fails on computed properties. The tests fare failing because `assert.equal` is being called with a model which ultimately has `store` injected on it which has `defaultAdapter` as a computed property. I would refactor this test to one capable of a strict reference equality check.

sly7-7: Thanks for the explanation, I will use test the equality on the id instead.

tests/integration/relationships/has-many-test.js
```diff
@@ -1528,6 +1528,25 @@ test("Polymorphic relationships with a hasMany is set up correctly on both sides
   assert.equal(get(email, 'posts.length'), 1, "The inverse has many is set up correctly on the email side.");
 });
 
+test("Can push a polymorphic belongsTo, the relationship is set up correctly on both sides", function(assert) {
+  assert.expect(2);
+
+  Comment.reopen({
+    message: DS.belongsTo('message', { polymorphic: true, async: false })
```
runspired: is the intent for this to be a reflexive polymorphic relationship? If not, I suspect you meant to set this to `belongsTo('post'`

runspired: (If you do intent a reflexive polymorphic relationship, then the test code below would need changes)

runspired:
After discussing out of band, we agreed that the test setup is invalid and that a valid setup for this test would result in duplicating an existing (passing) test.

That said, of note is the confusion here on where to define a relationship.

The test intended for the `belongsTo` to be polymorphic, but in order for this to work the inverse relationship needs to be defined on the base-type.

The given setup was attempting:

```
Message

Post extends Message hasMany Comment

Comment hasOne Message
```

It needed to be:

```
Message hasMany Comment

Post extends Message

Comment hasOne Message
```

PR #5410, opened 2018-04-03
[FEAT BUGFIX] resolves issues with links and data in relationships

### LINKED ISSUE #3115: isReloading flag not set on hasMany relationships
When you reload a model, its `isReloading` flag is set. But, when you reload a relationship, there is no flag. The flag would be helpful for adapters that suppress API calls for relationships that were previously loaded.

### LINKED ISSUE #4099: Ember Data tries to reload an already-loaded `belongsTo`
Ember data models:

``` js
// models/author.js
import DS from 'ember-data';

  name: DS.attr(),
  posts: DS.hasMany()

});

// models/post.js
import DS from 'ember-data';

  author: DS.belongsTo()

});

```

I'm using a JSON:API backend (and adapter on the frontend). My application route (with the `ds-finder-include` feature flag enabled):

``` js
// routes/application.js
  model() {
    return this.store.findRecord('author', 1, {
      include: 'posts'
    });
  }
});
```

which correctly sends out the request

```
GET http://localhost:3000/authors/1?include=posts
```

which responds with https://gist.github.com/samselikoff/20012aa58075db6124de.

In my application template,

``` hbs
<h2>I am {{model.name}}, here are my posts:</h2>

<ul>
  {{#each model.posts as |post|}}
    <li>{{post.id}} (written by {{post.author.name}})</li>
  {{/each}}
</ul>
```

`{{post.author.name}}` correctly displays the author's name, but triggers n+1 requests to `/posts/1/author`, `/posts/2/author` and so on. So, the inverse relationship is loaded correctly in Ember Data, but ED is still making an ajax request.

@wecc tracked it down to https://github.com/emberjs/data/blob/58262301c97df58998f307d6b75a4d4cbd2176c8/addon/-private/system/relationships/state/belongs-to.js#L120-L146

### LINKED ISSUE #4292: Known belongsTo ids cannot use a link in a RESTSerializer
In a 'normal' situation there are two possibilities to obtain a belongsTo relation when using the RESTSerializer:
1.  Add the relation id to the response of owner object; Ember will automatically fetch the model by `/name-of-model-plural/{:id}`.
2. Add a `links` object in the normalize function of own created serializer to obtain the data. Ember will fetch the model by given link.

So far, so good. But something goes wrong if both the relation id is available in the response and `links` object is also setted in your own created serializer.

Ember does still call `/name-of-model-plural/{:id}`, instead of calling `/own-given-link/{:id}`.
See [Ember Twiddle](https://ember-twiddle.com/9a0dd17be2804beb280fa87b71d41c9a?fileTreeShown=false&numColumns=2&openFiles=application.route.js%2Cserializers.file.js) for a live example.

### LINKED ISSUE #4712: hasMany resolution will not use related links if any data:[] is present
If I have this json-api response:
```
{
  "data": [
    {
      "type": "expel-alerts",
      "id": "64839987-80d5-4de7-9a74-62d472ba0562",
      "attributes": {
        "detail": "collaborate with collaborators",
        "title": "3 endpoints in Finance domain beaconing to known C2 122.33.52.122",
        "severity": "3",
        "activity-start": "2016-11-23T04:55:45.000Z",
        "activity-end": "2016-11-25T23:55:45.000Z",
        "type": "network"
      },
      "relationships": {
        "vendor-alerts": {
          "data": [
            {
              "type": "vendor-alerts",
              "id": "a91db472-99f7-4a8d-8878-7365e689a346"
            }
          ],
          "links": {
            "self": "/expel-alerts/64839987-80d5-4de7-9a74-62d472ba0562/relationships/vendor-alerts",
            "related": "/expel-alerts/64839987-80d5-4de7-9a74-62d472ba0562/vendor-alerts"
          }
        }
      }
    }
  ],
  "meta": {
    "count": 1
  }
}
```
In the ember inspector if $E is set to the expelAlert model object, , if I do $E.get('vendorAlerts') , ember-data will attempt to do /api/vendor-alerts/a91db472-99f7-4a8d-8878-7365e689a346 to retrieve the vendorAlert model (ignoring the related links entry)

If, the JSONAPI response (for just the relationships area) is:
```
      "relationships": {
        "vendor-alerts": {
          "links": {
            "self": "/expel-alerts/64839987-80d5-4de7-9a74-62d472ba0562/relationships/vendor-alerts",
            "related": "/expel-alerts/64839987-80d5-4de7-9a74-62d472ba0562/vendor-alerts"
          }
        }
      }
```
then $E.get('vendorAlerts') will hit /api/expel-alerts/64839987-80d5-4de7-9a74-62d472ba0562/vendor-alerts  (respecting the related links entry)

I would think(?) it would always prioritize the links entry.  (especially if there were say, 5000 vendor-alerts.. you wouldn't want 5000 individual ajax requests, nor would the coalesce true ?filter[etc]=1,2,...,5000 be right either)

### LINKED ISSUE #4889: Weird behaviour when including relationships in response
I'm currently trying to load quite a few records and their related records in one request. Sometimes Ember Data seems to fetch related records even though they were included in the request. Another weird thing is, that switching to a different model it seems like reloaded models change or unset their relations.

I've made a little twiddle for reproduction:
https://ember-twiddle.com/bbeb5040ebe8e342ecf854199e1bb0a6

Visiting this link, you'll see a list of `Playlists`, clicking one of those fetches one of those playlists with all their related data in one request (using mirage). On the first visit, it seems to make unnecessary requests to already included records of `PlaylistTrack`.

Switching between Playlists also seem to have some weird behavior, where `Track` relations to `Artist` seem to unset them on subsequent requests.

### LINKED ISSUE #4954: hasMany, belongsTo inconsistent about adapter vs serializer
### Summary

In a hasMany or belongsTo relationship, fetching the associated record(s) will use the source adapter and the foreign serializer.

### Example

I have two models: `Person` and `Group`:
```js
// app/models/group.js:
people: DS.hasMany('person')

// app/models/person.js:
group: DS.belongsTo('group')
```

The `Person` API is classic JSON. The `Group` API is JSON-API. So I declare adapters and serializers accordingly:

```js
// app/adapters/group.js
import DS from 'ember-data';
export default DS.JSONAPIAdapter;

// app/adapters/person.js
import ActiveModelAdapter from 'active-model-adapter';
export default ActiveModelAdapter;

// app/serializers/group.js
import DS from 'ember-data';
export default DS.JSONAPISerializer;

// app/serializers/person.js
import { ActiveModelSerializer } from 'active-model-adapter';
export default ActiveModelSerializer;
```

If I do a `store.findAll('person')` or a `store.find('group', 178)`, this works great.

But if I do `person.get('group')` or `group.get('people')`, this breaks terribly. The problem is that `person.get('group')` uses `adapter:person` (not `adapter:group`), but `serializer:group`. Thus, it's receiving active-model-style JSON, but expecting JSON-API.

If I do `store.query('group', { include: 'people' })`, it breaks in a different way. The JSON-API response for groups comes back with JSON-API documents for the people. It passes these documents to `serizlier:person`, which expects active-model-style JSON documents.

### LINKED ISSUE #4991: Parent nodes in tree-like recursive structure always appear as child of their children
I think I found a bug in Ember Data, but at the same time I'm very wary of thinking that. However, I don't know what else to make of the behavior I am getting.

Long story short, I created [this repository](https://github.com/gnapse/ember-data-dummy-poc) where the behavior in question can be reproduced. Some details are also given here (mdehoog/ember-data-has-many-query#24) since this was initially thought to be caused by using that ember addon, but the repository I created allows to reproduce the problem without that addon being involved.

In an attempt to briefly describe the problem here for completeness, here it goes:

I'm using Ember Data with a [jsonapi](jsonapi.org/format/) api, and I have a tree-like nested/recursive model `Role`, where each role has many subordinate roles under it in an organization (the children), and also each role has a manager (the parent node).

JsonApi allows to include the URLs to fetch relationships. If the URL given for a certain role, to fetch its subordinates, contains a query param, then for some reason the instances of subordinates obtained when fetching from that URL now include their parent also in the list of their subordinates.

This is perhaps better understood with an example:

```
CEO (id = 1)
  - CTO
    - Software Division Manager
    - Hardware Division Manager
```

If the URL to fetch the CEO's subordinates is `/roles/1/subordinates` then everything works correctly. However, if the URL is `/roles/1/subordinates?depth=1` (regardless of the meaning of this parameter, assume the parameter does not actually change what the api responds) then the roles obtained from that request now include not only their actual subordinates, but they also include their parent node in the list of subordinates.

For instance, in the tree above, after having obtained the subordinates of the CEO, in this case that's the list with only one element, the CTO. If we ask the CTO to give us its subordinates, it'll return a list with three roles: the two roles actually under it in the structure, and the CEO as well.

### LINKED ISSUE #4997: belongsTo computed uses diferrent url to fetch related then BelongsToReference.load
BelongsTo computed retrieves model using:
```
this._internalModel._relationships.get(key).getRecord();
```

but BelongsToReference.load internally incorporates decision:
```
 if (this.remoteType() === "id") {
    return this.belongsToRelationship.getRecord();
  }

  if (this.remoteType() === "link") {
    return this.belongsToRelationship.findLink().then((internalModel) => {
      return this.value();
    });
  }
```
Above results in different url used to fetch the record when related record id and type is suplied by parent record JSON API.

### LINKED ISSUE #5143: 2.14.x. links relationships id is reloaded. 2.13.2 is OK.
## **_With 2.13.2 everything ok._**

**From 2.14.x** I have a strange phenomenon:

I'm in the category 1 page, in that page I have a simply

```
{{#each model.posts key="id" as |post index|}}
    {{#link-to 'post' model.id post.id}}Post{{/link-to}}`
{{/each}}
```

when I go in the post page I have this:

`{{#link-to "category" model.category}}Back to category{{/link-to}}`

sometimes that link is **_`/#`_**.

In my router.js I have this:

```
...
this.route('post', { path: ':category_id/:post_id' });
...

```
So after investigating a bit I realized that the blame is of the new "ember-data 2.14.x (and 2.15.x of course, I tried)" and of its bad relationship with the related JSONAPI links (many bugs with JSONAPI related links just fixed in ember-data 2.14.10).

The first time I enter that post page the link "Back to category" is **_`/#`_** until it loads (**again** I can say, because in the category page it already knows about `post` and its `category.id`, isn't it?).

I can't reproduce on ember-twiddle because I'm not enough good to. But I realized a twiddle with mirage without links related relationships. Maybe some of you can insert the missing links part in mirage? Here is the link: https://ember-twiddle.com/94b21b033f50339cfa5de0040787f12f

I hope we can fix soon and most important we can understand and test more from this mistake. :)

Maybe related to: https://github.com/emberjs/data/pull/5124, https://github.com/emberjs/data/issues/5062, https://github.com/emberjs/data/issues/4982.

And maybe @mmun (author of bugfix https://github.com/emberjs/data/pull/5119 in 2.14.10) knows something more and better?

### LINKED ISSUE #5209: Links related relationship on modelA nulls belongsTo on modelB

### LINKED ISSUE #5211: A bidirectional relationship `parent.children` becomes `hasLoaded: true` when a child that references the parent is pushed to the store

### LINKED ISSUE #5274: Fix fetch hasMany using links when parent & child records are already loaded

### LINKED ISSUE #5423: "Attempted to handle event `reloadRecord` on <xyz> while in state root.loading" when reloading a belongsTo association multiple times simultaneously
I have a situation where I render a same component twice on the same page. This component reloads a belongsTo association at initialization. So something like:

```
// template.hbs

{{x-component model=model}}

{{! ...further on the page... !}}

{{x-component model=model}}
```

```javascript
// x-component/component.js

export default Component.extend({
  init() {
    this._super(...arguments);

    this.get('model').belongsTo('assoc').reload();
  }
});
```

This throws an error

> Attempted to handle event `reloadRecord` on <xyz> while in state root.loading.

I figured the issue is that since the components are rendered on the same page, the two reloads of the association are fired at "around" the same time. However, further down the stack inside Ember Data, one reload necessarily happens before the other and changes the state of the internal model. When the second reload is processed, it fails at https://github.com/emberjs/data/blob/v3.0.2/addon/-private/system/model/internal-model.js#L749 because the previous one has changed the state.

I tried observing the state before the reload, i.e.

```javascript
// x-component/component.js

export default Component.extend({
  init() {
    this._super(...arguments);
   
    // the following line always prints 'false'
    console.log(this.get('model').belongsTo('assoc').internalModel.isLoading());
   
    // the following line always prints 'root.loaded.saved'
    console.log(this.get('model').belongsTo('assoc').internalModel.currentState.stateName);
    
    this.get('model').belongsTo('assoc').reload();
  }
});
```

but it looks like the `reload` calls really happen at around the same time, or in any case before Ember Data has changed the state due to one reload being processed.

I could of course reload the association in the parent instead of inside the component but ideally I'd like to keep this logic contained within the component.

Any thoughts on how to approach this?

runspired:
**Description**

This PR includes two fixes for links and a suite of new tests to accompany them.

The first addresses a shortcoming of the lazy-relationships work which caused `meta` and `links` information to be lossy. We now "merge forward" meta and links information on relationship payloads when necessary.  We also now handle data for `has-many` relationships discovered via their inverse in a special way (see `_partialData`). This avoids us losing information about whether the membership should be considered complete or not. 

The second addresses the situation in which both `data` and `links` are present for a relationship. We now ALWAYS prefer to fetch/re-fetch relationship data via a link if present.  To achieve this, we now do a minimal amount of book-keeping on whether the relationship should be considered available locally, or must be refetched.

**General Warning**

This PR cleans up some previously poorly spec'd behavior, which means that it may cause some minor breakage to apps using `links` that had found work-arounds for the `mixing links and data` issue.

Because it fixes our ability to use `links` for relationships in general, folks that implemented workarounds for relationships expecting their links to never be used might be particularly surprised.  Ideally things will `"just work"™` in most situations for these folks, and they can delete their workarounds.

**Resolves and/or supplants**

- Resolves #5423 
- Resolves #5351
- Resolves #5274 
- Resolves #5257 
- Resolves #5235 
- Resolves #5211 
- Resolves #5209 
- Resolves #5143 
- Resolves #4997 
- Resolves #4991 
- Resolves #4954  
- Resolves #4889 
- Resolves #4712 
- Resolves #4292 
- Resolves #4099
- Resolves #3115  

It additionally advances the status of #2905 and #2162

### DIFF
--- a/addon/-private/system/promise-proxies.js
+++ b/addon/-private/system/promise-proxies.js
@@ -82,6 +82,21 @@ export function promiseArray(promise, label) {
   });
 }
 
+export function proxyToContent(method) {
+  return function() {
+    return get(this, 'content')[method](...arguments);
+  };
+}
+
+export const PromiseBelongsTo = PromiseObject.extend({
+  reload() {
+    assert('You are trying to reload an async belongsTo before it has been created', get(this, 'content'));
+    this.get('_belongsToState').reload();
+
+    return this;
+  }
+});
+
 /**
   A PromiseManyArray is a PromiseArray that also proxies certain method calls
   to the underlying manyArray.
@@ -99,13 +114,6 @@ export function promiseArray(promise, label) {
   @namespace DS
   @extends Ember.ArrayProxy
 */
-
-export function proxyToContent(method) {
-  return function() {
-    return get(this, 'content')[method](...arguments);
-  };
-}
-
 export const PromiseManyArray = PromiseArray.extend({
   reload() {
     assert('You are trying to reload an async manyArray before it has been created', get(this, 'content'));
--- a/addon/-private/system/promise-proxies.js
+++ b/addon/-private/system/promise-proxies.js
@@ -89,8 +89,6 @@ export function proxyToContent(method) {
 }
 
 export const PromiseBelongsTo = PromiseObject.extend({
-  meta: reads('content.meta'),
-
   reload() {
     assert('You are trying to reload an async belongsTo before it has been created', this.get('content') !== undefined);
     this.get('_belongsToState').reload();
--- a/addon/-private/system/promise-proxies.js
+++ b/addon/-private/system/promise-proxies.js
@@ -1,7 +1,7 @@
 import ObjectProxy from '@ember/object/proxy';
 import PromiseProxyMixin from '@ember/object/promise-proxy-mixin';
 import ArrayProxy from '@ember/array/proxy';
-import { get } from '@ember/object';
+import { get, computed } from '@ember/object';
 import { reads } from '@ember/object/computed';
 import { Promise } from 'rsvp';
 import { assert } from '@ember/debug';
@@ -82,13 +82,20 @@ export function promiseArray(promise, label) {
   });
 }
 
-export function proxyToContent(method) {
-  return function() {
-    return get(this, 'content')[method](...arguments);
-  };
-}
-
 export const PromiseBelongsTo = PromiseObject.extend({
+
+  // we don't proxy meta because we would need to proxy it to the relationship state container
+  //  however, meta on relationships does not trigger change notifications.
+  //  if you need relationship meta, you should do `record.belongsTo(relationshipName).meta()`
+  meta: computed(function() {
+    assert(
+      'You attempted to access meta on the promise for the async belongsTo relationship ' +
+      `${this.get('_belongsToState').internalModel.modelName}:${this.get('_belongsToState').key}'.` +
+       '\nUse `record.belongsTo(relationshipName).meta()` instead.',
+      false
+    );
+  }),
+
   reload() {
     assert('You are trying to reload an async belongsTo before it has been created', this.get('content') !== undefined);
     this.get('_belongsToState').reload();
@@ -114,9 +121,14 @@ export const PromiseBelongsTo = PromiseObject.extend({
   @namespace DS
   @extends Ember.ArrayProxy
 */
-export const PromiseManyArray = PromiseArray.extend({
-  link: reads('content.link'),
 
+export function proxyToContent(method) {
+  return function() {
+    return get(this, 'content')[method](...arguments);
+  };
+}
+
+export const PromiseManyArray = PromiseArray.extend({
   reload() {
     assert('You are trying to reload an async manyArray before it has been created', get(this, 'content'));
     this.set('promise', this.get('content').reload());
--- a/addon/-private/system/promise-proxies.js
+++ b/addon/-private/system/promise-proxies.js
@@ -89,8 +89,10 @@ export function proxyToContent(method) {
 }
 
 export const PromiseBelongsTo = PromiseObject.extend({
+  meta: reads('content.meta'),
+
   reload() {
-    assert('You are trying to reload an async belongsTo before it has been created', get(this, 'content'));
+    assert('You are trying to reload an async belongsTo before it has been created', this.get('content') !== undefined);
     this.get('_belongsToState').reload();
 
     return this;
@@ -115,9 +117,11 @@ export const PromiseBelongsTo = PromiseObject.extend({
   @extends Ember.ArrayProxy
 */
 export const PromiseManyArray = PromiseArray.extend({
+  link: reads('content.link'),
+
   reload() {
     assert('You are trying to reload an async manyArray before it has been created', get(this, 'content'));
-    this.set('promise', this.get('content').reload())
+    this.set('promise', this.get('content').reload());
     return this;
   },
 
--- a/addon/-private/system/references/has-many.js
+++ b/addon/-private/system/references/has-many.js
@@ -258,7 +258,7 @@ export default class HasManyReference extends Reference {
   }
 
   _isLoaded() {
-    let hasRelationshipDataProperty = get(this.hasManyRelationship, 'hasRelationshipDataProperty');
+    let hasRelationshipDataProperty = get(this.hasManyRelationship, 'hasAnyRelationshipData');
     if (!hasRelationshipDataProperty) {
       return false;
     }
--- a/addon/-private/system/references/has-many.js
+++ b/addon/-private/system/references/has-many.js
@@ -258,8 +258,8 @@ export default class HasManyReference extends Reference {
   }
 
   _isLoaded() {
-    let hasData = get(this.hasManyRelationship, 'hasData');
-    if (!hasData) {
+    let hasRelationshipDataProperty = get(this.hasManyRelationship, 'hasRelationshipDataProperty');
+    if (!hasRelationshipDataProperty) {
       return false;
     }
 
--- a/addon/-private/system/relationships/relationship-payloads.js
+++ b/addon/-private/system/relationships/relationship-payloads.js
@@ -9,7 +9,11 @@ import { assert } from '@ember/debug';
  */
 function mergeForwardPayload(oldPayload, newPayload) {
   if (oldPayload && oldPayload.data !== undefined && newPayload.data === undefined) {
-    // newPayload.data = oldPayload.data;
+    newPayload.data = oldPayload.data;
+  }
+
+  if (newPayload.data === undefined && oldPayload && oldPayload._partialData !== undefined) {
+    newPayload._partialData = oldPayload._partialData;
   }
 
   if (oldPayload && oldPayload.meta !== undefined && newPayload.meta === undefined) {
@@ -121,13 +125,16 @@ export default class RelationshipPayloads {
    */
   get(modelName, id, relationshipName) {
     this._flushPending();
+    let payload;
 
     if (this._isLHS(modelName, relationshipName)) {
-      return this.lhs_payloads.get(modelName, id);
+      payload = this.lhs_payloads.get(modelName, id);
     } else {
       assert(`${modelName}:${relationshipName} is not either side of this relationship, ${this._relInfo.lhs_key}<->${this._relInfo.rhs_key}`, this._isRHS(modelName, relationshipName));
-      return this.rhs_payloads.get(modelName, id);
+      payload = this.rhs_payloads.get(modelName, id);
     }
+
+    return payload;
   }
 
   /**
@@ -367,6 +374,9 @@ export default class RelationshipPayloads {
         // if we also have links, which this would indicate
         if (existingData) {
           existingData.push(inversePayload.data);
+        } else {
+          existingPayload._partialData = existingPayload._partialData || [];
+          existingPayload._partialData.push(inversePayload.data);
         }
       } else {
         mergeForwardPayload(existingPayload, inversePayload);
@@ -377,7 +387,7 @@ export default class RelationshipPayloads {
       //
       if (inverseIsMany) {
         inversePayloadMap.set(resourceIdentifier.type, resourceIdentifier.id, {
-          data: [inversePayload.data]
+          _partialData: [inversePayload.data]
         });
       } else {
         inversePayloadMap.set(resourceIdentifier.type, resourceIdentifier.id, inversePayload);
@@ -405,7 +415,10 @@ export default class RelationshipPayloads {
    */
   _removeInverse(id, previousPayload, inversePayloadMap) {
     let data = previousPayload && previousPayload.data;
-    if (!data) {
+    let partialData = previousPayload && previousPayload._partialData;
+    let maybeData = data || partialData;
+
+    if (!maybeData) {
       // either this is the first time we've seen a payload for this id, or its
       // previous payload indicated that it had no inverse, eg a belongsTo
       // relationship with payload { data: null }
@@ -415,10 +428,10 @@ export default class RelationshipPayloads {
       return;
     }
 
-    if (Array.isArray(data)) {
+    if (Array.isArray(maybeData)) {
       // TODO: diff rather than removeall addall?
-      for (let i=0; i<data.length; ++i) {
-        const resourceIdentifier = data[i];
+      for (let i=0; i<maybeData.length; ++i) {
+        const resourceIdentifier = maybeData[i];
         this._removeFromInverse(id, resourceIdentifier, inversePayloadMap);
       }
     } else {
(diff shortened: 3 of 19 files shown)


addon/-private/system/relationships/state/has-many.js
```diff
@@ -336,6 +332,20 @@ export default class ManyRelationship extends Relationship {
     }
   }
 
+  localStateIsEmpty() {
+    let manyArray = this.manyArray;
+    let internalModels = manyArray.currentState;
+    let manyIsArrayLoaded = manyArray.get('isLoaded');
```
hjdivad: s/manyIsArrayLoaded/manyArrayIsLoaded perhaps

addon/-private/system/relationships/state/relationship.js
```diff
@@ -77,6 +77,7 @@ export default class Relationship {
     this.meta = null;
     this.hasData = false;
     this.hasLoaded = false;
+    this.hasLocalData = false;
```
hjdivad: What's the case where `hasData && !hasLocalData`?

ie how does "local" data differ from data?

Does this capture explicit `data` in a json-api resource as opposed to data loaded from a link in a json-api resource that initially contained no `data`?

addon/-private/system/relationships/state/has-many.js
runspired: confirm, was dyslexic.

addon/-private/system/relationships/state/relationship.js
sly7-7: Maybe (probably) because I'm french, but I would (wrongly?) better understand `relationshipHasDataProperty`.

runspired:
I added 25 new tests to ensure full coverage of the links+data issue. We have 8 failures but I think they should be pretty easy to resolve.

addon/-private/system/relationships/relationship-payloads.js
```diff
@@ -257,23 +257,40 @@ export default class RelationshipPayloads {
...
       // * undefined is NOT considered new information, we should keep original state
       // * anything else is considered new information, and it should win
+      let isMatchingIdentifier = this._isMatchingIdentifier(
+        relationshipData && relationshipData.data,
+        previousPayload && previousPayload.data
+      );
+
       if (relationshipData.data !== undefined) {
-        this._removeInverse(id, previousPayload, inversePayloadMap);
+        if (!isMatchingIdentifier) {
```
sly7-7: Am I missing something or it should be called with two arguments ?

addon/-private/system/references/has-many.js
```diff
@@ -258,8 +258,8 @@ export default class HasManyReference extends Reference {
   }
 
   _isLoaded() {
-    let hasData = get(this.hasManyRelationship, 'hasData');
-    if (!hasData) {
+    let hasRelationshipDataProperty = get(this.hasManyRelationship, 'hasAnyRelationshipData');
+    if (!hasRelationshipDataProperty) {
```
runspired: this should probably be noted for a follow up PR to be changed. No tests failed but now that we've clarified what the flags are it is rather obvious that this is not correct.

addon/-private/system/promise-proxies.js
```diff
@@ -82,6 +82,28 @@ export function promiseArray(promise, label) {
...
+  //  however, meta on relationships does not trigger change notifications.
+  //  if you need relationship meta, you should do `record.belongsTo(relationshipName).meta()`
+  meta: computed(function() {
+    assert(
+      'You attempted to access meta on the promise for the async belongsTo relationship ' +
+      `${this.get('_belongsToState').internalModel.modelName}:${this.get('_belongsToState').key}'.` +
+       '\nUse `record.belongsTo(relationshipName).meta()` instead.',
+      false
+    );
+  }),
```
runspired: sidenote: this is actually true for hasMany as well. It does proxy `meta`, however it rarely picks up the change notification (if ever).  We should address this soon.

bmac:
Thanks @runspired.

BryanCrotaz:
Wow. Nicely done.

sandstrom:
Awesome @runspired 🥇 🎉

PR #5411, opened 2018-04-03
Remove deprecations scheduled for 3.0

runspired:
Leaves one behind for array responses and queryRecord that will be a bit trickier to cleanup


addon/-private/system/model/errors.js
```diff
@@ -117,7 +98,6 @@ export default ArrayProxy.extend(Evented, {
     this.on('becameValid', target, becameValid);
   },
```
runspired: cc @stefanpenner tests broke when I tried to remove `_registerHandlers`, suspect it's just some tests that need updating but would love you to poke at this if you have a moment.

PR #5412, opened 2018-04-03
Fix npmignore and cleanup test artifacts before publishing

runspired:
cc @rwjblue @kategengler @bmac


rwjblue:
Sorry 😭 about missing this during the Travis setup refactor ....

Issue #5247, opened 2017-11-01
store.createRecord should be async?


rwjblue:
While working on [ember-qunit-codemod](https://github.com/rwjblue/ember-qunit-codemod) (which converts existing QUnit tests to the structure proposed in emberjs/rfcs#232) I have ran into something a bit surprising to me around `store.createRecord(...)` and I _think_ it may be something that we should address (or at least discuss).

Currently, ember-data uses batching in a few locations:

* creation/updating of the various live record arrays https://github.com/emberjs/data/blob/ce606b1b6946a700248f92e35222cca195c8e80e/addon/-private/system/record-array-manager.js#L96 

* Transitioning an internal model into _any_state (via `InternalModel.prototype.transitionTo`) https://github.com/emberjs/data/blob/ce606b1b6946a700248f92e35222cca195c8e80e/addon/-private/system/model/internal-model.js#L847 https://github.com/emberjs/data/blob/ce606b1b6946a700248f92e35222cca195c8e80e/addon/-private/system/model/internal-model.js#L1017-L1019

To me, this effectively means that anything that runs through these paths is actually async (which roughly seems like "everything"?) and probably should return a promise.

---

Concretely, the issues I'm facing in testing land are:

* `store.createRecord(...)` must be wrapped in a `Ember.run` to avoid triggering the auto-run assertion, in the prior `moduleForModel` API this was hidden from users behind `this.subject()`, but the new API is specifically intended to remove these extra "test only API's" and instead have users do what they would actually do in normal app code

* calling `run(() => store.createRecord(...))` in test behaves _significantly_ different than what most users would do in "normal" app code (`store.createRecord(...)`)

bmac:
We talked about this at a recent Ember Data core meeting and it was agreed that `createRecord` should be made sync. We suspect the performance improvements with deferring relationship and record array updates don't make a big of an impact for `createRecord` then they do with pushing many records from the server.

rwjblue:
👍 thank you for summarizing

Issue #2566, opened 2014-12-09
defaultValue should make models dirty


courthead:
In the JSBin below, I've got an `App.Book` model with a default value for the `title` attribute, and a book instance retrieved from the server that has no value for its `title` attribute:

http://emberjs.jsbin.com/xaniheguto/2/edit?html,js,output

The book's title is automatically set via the `defaultValue` option. However, the model isn't marked dirty and `model.rollback()` doesn't set the title to `undefined`. I might be alone here, but I would expect any model that's changed in memory to be marked dirty so I know whether or not it differs from the state on the server, and the fact that `defaultValue` is an exception to that behavior is a surprise.

One use-case might be if you're using Firebase, Dropbox Datastore API, etc as a back-end. You can't write server code to validate your data, so you have to do it on the front-end. In this particular case, you'd have no practical way of knowing that the book's title on the server is actually `null`.

Twinkletoes:
I disagree - if you want a default value in the UI, use a controller. For my money, defaultValue should only be used when creating new records. It shouldn't override what we get from the server.

bmac:
It is surprising to me that `defaultValue` is used when the record is loaded from the server and not created locally. Unfortunately, I don't think we can change this behavior without breaking semver so its going to have to stay.

I'm going to close this issue. Feel free to re-open if you disagree.

knownasilya:
What about making the change for 3.0?

courthead:
+1 for changing this in 3.0

runspired:
To restate the original request here:

`defaultValue` on a newly created record should cause the newly created record to be dirty.  I'm not sure if I agree with this, but I do understand how it could be useful.

My concern is that the use cases for "create a record with default values and save it" sounds suspiciously like "API actions".  For instance, likes on a like button, in which the record itself is barely cared about.

Now, it may be that an "API action" should actually be added as a relationship to something else, in which case this was valid, but more often than not I see this in the "discardable local record" situation.  And, to go with our "like" example, were this a like there would be a relationship. Ergo, I don't see this feature as useful for anything other than "disregarded API actions", for which the following pattern is better, using the example of `analytics` gathering.

```
this.get('store').adapterFor('analytics').triggerAction('page-load', data);
```

cc @wecc

wecc:
Dumping this here after discussion with @runspired, @stefanpenner and @tchak:

The current `defaultValue` implementation is not really "default – as in a UI concern. That's why they won't mark records as dirty or get serialized.

[...]

PR #5413, opened 2018-04-03
[BUGFIX] Fix availability of properties in createRecord init

runspired:
This restores the changes from #4931 that were removed by #5369, ensuring that in as many cases as possible we "do the expected thing" and give access to properties in `init`.

Historically, so long as `_internalModel` has data, any available record-data is available in `init`. This PR ensures that the same is true for properties passed to `createRecord`.

We achieve this by buffering the changes for dirty-state sent to `ManyArray`.  In the process, we also cleanup the responsibilities of `attr` `belongsTo` and `hasMany` descriptors who previously were responsible for the tricky logic of where to find state and how to set state on internal-model. This should help inform and improve the APIs of https://github.com/emberjs/rfcs/pull/293

### DIFF
--- a/addon/-private/system/model/internal-model.js
+++ b/addon/-private/system/model/internal-model.js
@@ -350,37 +350,39 @@ export default class InternalModel {
         isError: this.isError,
         adapterError: this.error
       };
-      let hasCreateProperties = typeof properties === 'object' && properties !== null;
-      let initializedRelationships;
 
-      if (hasCreateProperties) {
-        initializedRelationships = [];
+      if (properties !== undefined) {
+        assert(`You passed '${properties}' as properties for record creation instead of an object.`, typeof properties === 'object' && properties !== null);
+        let classFields = this.getFields();
+        let relationships = this._relationships;
+        let propertyNames = Object.keys(properties);
 
-        if ('id' in properties) {
-          this.setId(properties.id);
-          delete properties.id;
+        for (let i = 0; i < propertyNames.length; i++) {
+          let name = propertyNames[i];
+          let fieldType = classFields.get(name);
+          let propertyValue = properties[name];
+
+          if (name === 'id') {
+            this.setId(propertyValue);
+            continue;
+          }
+
+          switch (fieldType) {
+            case 'attribute':
+              this.setDirtyAttribute(name, propertyValue);
+              break;
+            case 'belongsTo':
+              this.setDirtyBelongsTo(name, propertyValue);
+              relationships.get(name).setHasData(true);
+              break;
+            case 'hasMany':
+              this.setDirtyHasMany(name, propertyValue);
+              relationships.get(name).setHasData(true);
+              break;
+            default:
+              createOptions[name] = propertyValue;
+          }
         }
-
-        this.eachAttribute((name) => {
-          if (name in properties) {
-            this.setDirtyAttribute(name, properties[name]);
-            delete properties[name];
-          }
-        });
-
-        this.eachRelationship((name, { kind }) => {
-          if (name in properties) {
-            initializedRelationships.push(name);
-            if (kind === 'belongsTo') {
-              this.setDirtyBelongsTo(name, properties[name]);
-            } else {
-              this.setDirtyHasMany(name, properties[name]);
-            }
-            delete properties[name];
-          }
-        });
-
-        emberAssign(createOptions, properties);
       }
 
       if (setOwner) {
@@ -392,17 +394,15 @@ export default class InternalModel {
 
       this._record = this.store.modelFactoryFor(this.modelName).create(createOptions);
 
-      if (hasCreateProperties && initializedRelationships.length) {
-        for (let i = 0; i < initializedRelationships.length; i++) {
-          this._relationships.get(initializedRelationships[i]).setHasData(true);
-        }
-      }
-
       this._triggerDeferredTriggers();
       heimdall.stop(token);
     }
 
     return this._record;
+  }
+
+  getFields() {
+    return get(this.modelClass, 'fields');
   }
 
   resetRecord() {
--- a/addon/-private/system/model/internal-model.js
+++ b/addon/-private/system/model/internal-model.js
@@ -602,6 +602,42 @@ export default class InternalModel {
     }
   }
 
+  getAttributeValue(key) {
+    if (key in this._attributes) {
+      return this._attributes[key];
+    } else if (key in this._inFlightAttributes) {
+      return this._inFlightAttributes[key];
+    } else {
+      return this._data[key];
+    }
+  }
+
+  setDirtyAttribute(key, value) {
+    let oldValue = this.getAttributeValue(key);
+    let originalValue;
+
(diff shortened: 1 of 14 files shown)


runspired:
cc @dwickern if you could "sanity check" this change in your application that hit issue before. I've kept your test and added additional coverage.

runspired:
This should resolve https://github.com/emberjs/data/issues/5247 as well, although not by making the API async but by ensuring that `createRecord` joins ember-data's internal backburner loop to setup relationships.

dwickern:
I'm still on 2.x so it might be difficult to test this

addon/-private/system/model/internal-model.js
```diff
@@ -348,6 +350,38 @@ export default class InternalModel {
         isError: this.isError,
         adapterError: this.error
       };
+      let hasCreateProperties = typeof properties === 'object' && properties !== null;
```
rwjblue: I would think `properties !== undefined` would be fine here...

addon/-private/system/model/internal-model.js
```diff
@@ -348,6 +350,38 @@ export default class InternalModel {
...
+            if (kind === 'belongsTo') {
+              this.setDirtyBelongsTo(name, properties[name]);
+            } else {
+              this.setDirtyHasMany(name, properties[name]);
+            }
+            delete properties[name];
+          }
+        });
+
+        emberAssign(createOptions, properties);
```
rwjblue: I would prefer to avoid mutating the provided POJO (mutating for example, completely breaks using a shared POJO to .createRecord across many tests). A straightforward way to do this would be to invert this method a bit:

```js
if (hasCreateProperties) {
  let classFields = this.getFields();

  for (let propertyName in properties) {
    let propertyValue = properties[propertyName];
    let fieldType = classFields.get(propertyName);

    if (propertyName === 'id') {
      this.setId(properties.id);
      continue;
    }

    switch (fieldType) {
      case 'attribute':
        this.setDirtyAttribute(name, propertyValue);
        break;
      case 'belongsTo':
        this.setDirtyBelongsTo(name, propertyValue);
        initializedRelationships.push(name);
        break;
      case 'hasMany':
        this.setDirtyHasMany(name, propertyValue);
        initializedRelationships.push(name);
        break;
      default:
        createOptions[propertyName] = propertyValue;
    }
  }
}


// ...snip...

getFields(attributeName) {
  return get(this.modelClass, 'fields');
}
```

addon/-private/system/model/internal-model.js
```diff
@@ -358,6 +392,12 @@ export default class InternalModel {
 
       this._record = this.store.modelFactoryFor(this.modelName).create(createOptions);
 
+      if (hasCreateProperties && initializedRelationships.length) {
+        for (let i = 0; i < initializedRelationships.length; i++) {
+          this._relationships.get(initializedRelationships[i]).setHasData(true);
```
rwjblue: Hmm, why doesn't `setDirtyHasMany` / `setDirtyBelongsTo` mark the relationship as having data?  Seems vaguely surprising to me...

runspired: We can move this into the original iteration (I believe), but not into the methods. `setDirtyHasMany` and `setDirtyBelongsTo` are used for update code-paths as well as create-code-paths and we only mark this as hasData for creates.

addon/-private/system/relationships/state/has-many.js
```diff
@@ -109,8 +111,46 @@ export default class ManyRelationship extends Relationship {
...
+    pending.push(internalModel, idx);
+
+    if (this._willUpdateManyArray === true) {
+      return;
+    }
+
+    this._willUpdateManyArray = true;
+    let backburner = this.store._backburner;
+
+    backburner.join(() => {
```
rwjblue: IMHO, it would be nice to remove this (and rely on the auto-run). What do you think?

addon/-private/system/store.js
```diff
@@ -339,34 +339,27 @@ Store = Service.extend({
...
-    let normalizedModelName = normalizeModelName(modelName);
-    let properties = copy(inputProperties) || Object.create(null);
 
-    // If the passed properties do not include a primary key,
-    // give the adapter an opportunity to generate one. Typically,
-    // client-side ID generators will use something like uuid.js
-    // to avoid conflicts.
+    return this._backburner.join(() => {
+      let normalizedModelName = normalizeModelName(modelName);
+      let properties = copy(inputProperties) || Object.create(null);
```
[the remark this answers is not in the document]
runspired: yeah, we know. we use it a ton :/

addon/-private/system/snapshot.js
```diff
@@ -24,25 +24,13 @@ export default class Snapshot {
     this._hasManyIds = Object.create(null);
     this._internalModel = internalModel;
 
-    let record = internalModel.getRecord();
+    // TODO is there a way we can assign known attributes without
```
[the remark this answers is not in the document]
runspired: i left the skipped test for tracking :D 
i can open an issue too though

addon/-private/system/snapshot.js
```diff
@@ -24,25 +24,13 @@ export default class Snapshot {
...
+    //  using `eachAttribute`? This forces us to lookup the model-class
+    //  but for findRecord / findAll these are empty and doing so at
+    //  this point in time is unnecessary.
+    internalModel.eachAttribute((keyName) => this._attributes[keyName] = internalModel.getAttributeValue(keyName));
```
pangratz: Looks like this broke `defaultValue` not being considered anymore, when the attribute hasn't been accessed yet 😔 

So given:

```js
// app/models/foo.js
export default Model.extend({
  bar: attr({ defaultValue: () => "baz" })
});
```

then

```js
let foo = store.createRecord('foo');

// this assertion breaks after upgrading to v3.2.0-beta.2 :pensive:
assert.deepEqual(foo.toJSON(), { foo: "baz" });
```

---

Currently the whole story around `defaultValue` is not ideal (e.g. #2566) and doing this in the app code by explicitly passing default values when creating a record is better, but I would still consider this a breaking change... What do you think?

rwjblue: Seems good to open an issue for this, I _think_ it should be fixable...

addon/-private/system/snapshot.js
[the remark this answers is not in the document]
runspired: Issue is #5419

addon/-private/system/snapshot.js
runspired: Should be very fixable, surprised this broke though, re-examining my changes, will open an issue.

runspired: hrm, so I didn't pull the defaultValue logic into internalModel, nor am I sure we should. https://github.com/emberjs/data/pull/5413/files#diff-c2a90350f555423721228dfa13c137f2R130

That said, the proposal in #5419 would be a good direction to address this, as we can access each attribute via the record if it exists, and only copy lazily if it does not.

PR #5414, opened 2018-04-04
[FEAT BUGFIX] adds adapterOptions ability to model.reload()

### LINKED ISSUE #5373: model.reload should take adapterOptions
When using methods like `store.findRecord`, it’s possible to pass in adapterOptions that can be accessed on `urlForFindRecord` and used to change the url. However when calling `model.reload()` there doesn’t seem to be a way to pass in these same `adapterOptions` so `urlForFindRecord` is subsequently called but `adapterOptions` is unavailable.

runspired:
Resolves #5373

### DIFF
--- a/addon/-private/system/model/model.js
+++ b/addon/-private/system/model/model.js
@@ -816,13 +816,23 @@ const Model = EmberObject.extend(Evented, {
     ```
 
     @method reload
-    @return {Promise} a promise that will be resolved with the record when the
+    @param {Object} options optional, may include `adapterOptions` hash which will be passed to adapter request
+
+   @return {Promise} a promise that will be resolved with the record when the
     adapter returns successfully or rejected if the adapter returns
     with an error.
   */
-  reload() {
+  reload(options) {
+    let wrappedAdapterOptions;
+
+    if (typeof options === 'object' && options !== null && options.adapterOptions) {
+      wrappedAdapterOptions = {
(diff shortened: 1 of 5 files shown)


addon/-private/system/model/model.js
```diff
@@ -816,13 +816,23 @@ const Model = EmberObject.extend(Evented, {
...
   */
-  reload() {
+  reload(options) {
+    let wrappedAdapterOptions;
+
+    if (typeof options === 'object' && options !== null && options.adapterOptions) {
+      wrappedAdapterOptions = {
+        adapterOptions: options.adapterOptions
+      };
+    }
```
bmac: Do we want to support this here or should we always require users to set `adapterOptions`?

bmac: Nevermind this looks good. 👍

addon/-private/system/model/model.js
```diff
@@ -816,13 +816,23 @@ const Model = EmberObject.extend(Evented, {
...
+
+   @return {Promise} a promise that will be resolved with the record when the
     adapter returns successfully or rejected if the adapter returns
     with an error.
   */
-  reload() {
+  reload(options) {
+    let wrappedAdapterOptions;
+
+    if (typeof options === 'object' && options !== null && options.adapterOptions) {
```
runspired: @rwjblue did you have a recommended way of making this more legible?

PR #5415, opened 2018-04-04
make createRecord sync and remove unnecessary run usage in tests

### LINKED ISSUE #5247: store.createRecord should be async?

runspired:
refactor all the test usage or run-wrapping for createRecord, we really should run prettier

resolves #5247 because I <3 @rwjblue

### DIFF
--- a/addon/-private/system/store.js
+++ b/addon/-private/system/store.js
@@ -338,25 +338,32 @@ Store = Service.extend({
     assert(`You need to pass a model name to the store's createRecord method`, isPresent(modelName));
     assert(`Passing classes to store methods has been removed. Please pass a dasherized string instead of ${modelName}`, typeof modelName === 'string');
 
-    return this._backburner.join(() => {
-      let normalizedModelName = normalizeModelName(modelName);
-      let properties = copy(inputProperties) || Object.create(null);
+    // This is wrapped in a `run.join` so that in test environments users do not need to manually wrap
+    //   calls to `createRecord`. The run loop usage here is because we batch the joining and updating
+    //   of record-arrays via ember's run loop, not our own.
+    //
+    //   to remove this, we would need to move to a new `async` API.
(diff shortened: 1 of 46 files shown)


sly7-7:
Any idea about the failing test against ember 2-12 ?

runspired:
seems I have some async test leaks to find

PR #5422, opened 2018-04-05
[FEAT] Prevent async test leakage (feat. adds test waiters)

runspired:
- [x] add waiter for custom backburner
- [x] enable test isolation available in ember-qunit
- [x] add waiter for ajax
- [x] bump ember-qunit and fix blueprints to account

cc @locks because this should enable the learning team to clean up some things since more stuff is waited on by-default.


rwjblue:
This looks good to me, but I do think we should work to remove a lot of this manual run-wrapping. It seems fairly archaic and likely covers some things up....

runspired:
@rwjblue this will allow me to land #5415 which does just that with much more confidence

PR #5428, opened 2018-04-10
make snapshot lazier and fix defaultValue

### LINKED ISSUE #5419: Snapshots with no record should maybe only show the attributes they know about
Basically, if a snapshot's internal-model does not have a record, then we know it is not the subject of a call to `save` and has not been used in the UI.  This means that we're likely generating it for a `findRecord` request.  We can optimize this situation by iterating attributes when generating the snapshot a bit differently: https://github.com/emberjs/data/blob/7b5a8857134d128ecc42ec2e10d1149601b68ed8/addon/-private/system/snapshot.js#L27-L31

cc @stefanpenner @hjdivad

### LINKED ISSUE #5426: Latest beta does not include `defaultValue` on Snapshots
See discussion here: https://github.com/emberjs/data/pull/5413#pullrequestreview-110426821

Fix should likely incorporate #5419

runspired:
Resolves #5426 
Resolves #5419 

Adds tests for

- 1. test that a record is not materialized (e.g. internalModel.hasRecord is false) when calling `findRecord` for the very first time
- 2. test that defaultValues provided to a model reflect on attributes on snapshots (both function based and value based)
- 3. test that non-materialized records generate attributes for snapshots lazily
- 4. test that materialized records generate attributes for snapshots greedily
- 5. unskips the test for snapshots not greedily causing model classes to be looked up

### DIFF
--- a/addon/-private/system/snapshot.js
+++ b/addon/-private/system/snapshot.js
@@ -17,18 +17,23 @@ import { get } from '@ember/object';
 */
 export default class Snapshot {
   constructor(internalModel, options = {}) {
-    this._attributes = Object.create(null);
+    this.__attributes = null;
     this._belongsToRelationships = Object.create(null);
     this._belongsToIds = Object.create(null);
     this._hasManyRelationships = Object.create(null);
     this._hasManyIds = Object.create(null);
     this._internalModel = internalModel;
 
-    // TODO is there a way we can assign known attributes without
-    //  using `eachAttribute`? This forces us to lookup the model-class
-    //  but for findRecord / findAll these are empty and doing so at
-    //  this point in time is unnecessary.
-    internalModel.eachAttribute((keyName) => this._attributes[keyName] = internalModel.getAttributeValue(keyName));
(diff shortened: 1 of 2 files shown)


runspired:
cc @pangratz @rwjblue

addon/-private/system/snapshot.js
```diff
@@ -17,18 +17,23 @@ import { get } from '@ember/object';
 */
 export default class Snapshot {
   constructor(internalModel, options = {}) {
-    this._attributes = Object.create(null);
+    this.__attributes = null;
```
rwjblue: I'd like (at some point) to start discussing moving some of these state caches into weakmaps now that we can use them indiscriminately...

tylerturdenpants: I can try and tackle that enhancement... maybe use a similar weakmap implementation found elsewhere in the code?

urbany:
Hi, thanks for this fix! Is it possible to include it in the next beta?

PR #5432, opened 2018-04-11
Remove exists-sync dependency

### LINKED ISSUE #5431: Remove deprecated exists-sync dependency
Hi,

I'm getting the following warning in my console :
`warning ember-data > exists-sync@0.0.3: Please replace with usage of fs.existsSync`

Does this dependency is still in use ?

Bartheleway:
This should close #5431

### DIFF
--- a/lib/calculate-version.js
+++ b/lib/calculate-version.js
@@ -1,5 +1,5 @@
 var path = require('path');
-var existsSync = require('exists-sync');
+var fs = require('fs');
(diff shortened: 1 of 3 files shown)


runspired:
Thank you for the contribution! I'm not sure why ember beta and canary are broken. Trying to get that resolved so this will merge :)

runspired:
@alex-ushakov-93 This just removes a deprecation warning, it isn't an error. It looks like something else went wrong.

PR #5433, opened 2018-04-11
Add failing test for hasMany unloadRecord

jmar910:
We found that unloading records in a hasMany relationship leaves null values in the relationship array. 

For example:

```
book = env.store.peekRecord('book', 1);
assert.equal(book.get('chapters.length'), 3);
chapter = env.store.peekRecord('chapter', 2);
chapter.unloadRecord();
book.get('chapters.length')  // 3, but expect to be 2
book.get('chapters')    // [Chapter, null, Chapter]
```

### DIFF
--- a/tests/integration/relationships/has-many-test.js
+++ b/tests/integration/relationships/has-many-test.js
@@ -109,6 +109,66 @@ module("integration/relationships/has_many - Has-Many Relationships", {
   }
 });
 
+test("unloadRecord should remove record from hasMany relationship - async hasMany / non-polymorphic", function(assert) {
+  run(() => {
+    env.store.push({
+      data: {
+        id: 1,
+        type: 'book',
+        relationships: {
+          chapters: {
+            data: [
+              {
+                id: 1,
+                type: 'chapter'
+              },
+              {
+                id: 2,
+                type: 'chapter'
+              },
+              {
+                id: 3,
+                type: 'chapter'
+              }
+            ]
+          }
+        }
+      }
+    });
+
+    env.store.push({
+      data: [
+        {
+          type: 'chapter',
+          id: 1
+        },
+        {
+          type: 'chapter',
+          id: 2
+        },
+        {
+          type: 'chapter',
+          id: 3
+        }
+      ]
+    });
+  });
+
+  env.adapter.findRecord = function(store, type, id, snapshots) {
+    return { data: { id, type: 'chapter', attributes: { title: 'whatever' } }};
+  }
+
+  let book, chapter;
+
+  run(() => {
+    book = env.store.peekRecord('book', 1);
+    assert.equal(book.get('chapters.length'), 3);
+    chapter = env.store.peekRecord('chapter', 2);
+    chapter.unloadRecord();
+    assert.equal(book.get('chapters.length'), 2);
+  });
+});
+
 test("When a hasMany relationship is accessed, the adapter's findMany method should not be called if all the records in the relationship are already loaded", function(assert) {
   assert.expect(0);
 


tests/integration/relationships/has-many-test.js
```diff
@@ -109,6 +109,66 @@ module("integration/relationships/has_many - Has-Many Relationships", {
...
+  }
+
+  let book, chapter;
+
+  run(() => {
+    book = env.store.peekRecord('book', 1);
+    assert.equal(book.get('chapters.length'), 3);
+    chapter = env.store.peekRecord('chapter', 2);
+    chapter.unloadRecord();
+    assert.equal(book.get('chapters.length'), 2);
```
runspired: you would likely need to wrap this in a separate run

runspired: (not to say that is a good thing)

runspired: I would expect this assertion to be for `3` still. I think the issue is not the length (which is still correct I believe) but rather that you have to wait for the relationship to re-fetch to be in a non-null state again..

Let's add an assertion for a `sync access` of the chapters here to show that if we immediately access after unloading we have a whole.

hjdivad: This is an async relationship, so unloading invalidates the relationship and a get will refetch.  However, the record is removed from retained many arrays.

See the test https://github.com/emberjs/data/blob/c438a8a5e053724eb5b5257a451259a761a06d98/tests/integration/records/unload-test.js#L1351-L1428

We assert the unloaded record is removed from a retained `ManyArray`  at https://github.com/emberjs/data/blob/master/tests/integration/records/unload-test.js#L1405

The relationship is invalidated so a get should refetch.  It may be that we're not invalidating `_relationshipPromiseCache` correctly.

fivetanley: How is the length still correct? With a normal JavaScript array, the length updates immediately when removing elements.

fivetanley: @hjdivad it also seems very surprising that calling `.get` refetches from the server, and also that you can't opt out of it. has it always been doing that?

hjdivad: @runspired so in this test we correctly invalidate & refetch on `book.get('chapters')`.

length 3 is correct, as you note.

The reason for `null` is that the record is dematerializing so `getRecord` from `ManyArray.objectAt` does not actually materialize an empty record.  I believe the only reason we can't `cancelDestroy` in `objectAt` is for older embers that have the `firstObject`/`lastObject` issue.

hjdivad: > How is the length still correct? With a normal JavaScript array, the length updates immediately when removing elements.

The chapter is not actually removed from the book though.  That could be done via a client-side delete or by updating the relationship data (either via `removeObject` or by pushing a new `relationships` payload).

> @hjdivad it also seems very surprising that calling .get refetches from the server, and also that you can't opt out of it. has it always been doing that?

The relationship is no longer loaded.  Imagine if you'd done

```
store.push({
  data: {
    id: 1,
    type: 'book',
    relationships: {
      chapters: [{ id: 1, type: 'chapter'}, { id: 2, type: 'chatper'}]
    }
  },

  included: [{
    id: 1,
    type: 'chapter'
  }]
});

store.peekRecord('book', 1).get('chapters');
```

You would expect there to be a fetch for an async relationship no?

jmar910: Not sure if it's helpful, but here's our use case in where we are running into this issue...

We have a  model `foo`, this `foo` hasMany `bars`, `bar` belongsTo `foo`. We fetch `foo` from the server, and then asynchronously we query `bars` extracting out the foreign key for `foo` to set up the relationships.

Some point later down the chain, we get a WS event that tells us that a `bar` has been deleted, so we call `bar.unloadRecord()` so the UI can reflect the deleted `bar`.

If i'm understanding correctly, this is working as it should? I get that the relationship should be invalidated if you unload one of the `bars`,  but unloading the record will keep the `null` value in the array? So, if we didn't want to do a refetch, but wanted the array to stay up to date, we would unload the record and manually remove the null value from the array?

I think this is less about the refetch, and more about the fact that we have a null value in an array that we depend on in the UI.

Still trying to wrap my head around all of this, so apologies if I misunderstood something somewhere.

fivetanley: > You would expect there to be a fetch for an async relationship no?

No, because the records were already loaded in `included` ? What's the point of `included` if not for pre-loading records?

Previous Ember Data semantics seemed to be that fetches were performed only if the relationship had never been loaded before.

fivetanley:
here's another scenario that triggers the refetch:

```javascript
const IdentityProvider = DS.Model.extend({
  owner: belongsTo('identity-provider-owner', {
    polymorphic: true,
    inverse: 'identityProvider'
  })
})
const Account = DS.Model.extend({
  identityProvider: belongsTo({ inverse: 'owner' })
});

const provider = store.createRecord('identity-provider', { owner: owner });

try {
  await provider.save();
} catch (e) {
  provider.unloadRecord(); // causes refetch of account.get('identityProvider')
}
```

PR #5434, opened 2018-04-12
[DOC release] Fix documentation syntax in attr

locks:
Due to the indentation in the description of the `defaultValue` bullet point, our markdown renderer is considering it to be a code block. I did a quick patch, but haven't verified how it renders yet.

Repro: https://www.emberjs.com/api/ember-data/3.0/classes/DS/methods/attr?anchor=attr


runspired:
The failing test runs are due to https://github.com/emberjs/ember.js/issues/16511

PR #5436, opened 2018-04-16
[CHORE] remove all usage of Ember.copy

runspired:
Adds a test utility for `deepCopy` where we needed it, uses `Object.assign` everywhere else.

### DIFF
--- a/addon/-private/system/model/internal-model.js
+++ b/addon/-private/system/model/internal-model.js
@@ -1,7 +1,6 @@
-import { assign, merge } from '@ember/polyfills';
 import { A } from '@ember/array';
 import { set, get } from '@ember/object';
-import { copy } from '@ember/object/internals';
+import { assign } from '@ember/polyfills';
 import EmberError from '@ember/error';
 import { isEqual } from '@ember/utils';
 import { setOwner } from '@ember/application';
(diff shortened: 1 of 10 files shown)


addon/-private/system/model/internal-model.js
```diff
@@ -636,7 +632,7 @@ export default class InternalModel {
       changedKeys = this._changedKeys(data.attributes);
     }
 
-    emberAssign(this._data, data.attributes);
+    Object.assign(this._data, data.attributes);
```
rwjblue: Can't rely on `Object.assign`. Using `Ember.assign` instead (IE11 does not have `Object.assign`)...

package.json
```diff
@@ -96,7 +96,7 @@
     "ember-publisher": "0.0.7",
     "ember-qunit-assert-helpers": "^0.2.1",
     "ember-resolver": "^4.1.0",
-    "ember-source": "~3.0.0",
+    "ember-source": "https://s3.amazonaws.com/builds.emberjs.com/release/shas/50db6dc8001c127e5aeb2d6ddbacf9f9205d1d7b.tgz",
```
runspired: o.OI think I have an uncleaned up ember-try state

PR #5437, opened 2018-04-17
[CHORE] removes deprecated Store.filter feature

runspired:
See https://github.com/ember-data/ember-data-filter/pull/12 for path forward for folks still using this long-since gated feature.

This approach relies on the fact that we still notify `record-array-manager` on any change to an `internal-model`, even if it isn't an addition or removal from a `LiveRecordArray`.  Once the deprecation cycle is complete (`3.5` target), we can do an additional cleanup phase in which we no longer will be required to noisily notify the `record-array-manager` of every change.

Associated deprecation RFC: https://github.com/emberjs/rfcs/pull/326


addon/-private/system/record-array-manager.js
```diff
@@ -514,6 +358,8 @@ function updateLiveRecordArray(array, internalModels) {
 
   if (modelsToAdd.length > 0)    { array._pushInternalModels(modelsToAdd); }
   if (modelsToRemove.length > 0) { array._removeInternalModels(modelsToRemove); }
+
+  return modelsToAdd || modelsToRemove;
```
igorT: Why do we return here now?

igorT:
Are there tests that assert the record array manager behavior the addon relies on?

addon/-private/system/record-array-manager.js
runspired: A minor perf optimization for the addon-based filter implementation. This essentially lets us know when none of the internal model changes were an addition or a remove, which tells the addon that it definitely needs to signal a recalculation.  Without this we blindly signal, leading to the occasional double notification.

runspired:
@igorT I thought there was one (but without a description), but looking back it seems not.I'll add one with a note as to purpose.

addon/-private/system/record-array-manager.js
runspired: I'll leave a note as to why and also that it can be removed in the follow up refactor to eliminate the unnecessary record-array-manager notifications post 3.5

runspired:
@igorT added tests and code comments

PR #5439, opened 2018-04-19
[FEAT] FASTBOOT SHOEBOX - Use actions queue so rehydration works

runspired:
Fastboot Shoebox rehydration works at the adapter level, waiting until `afterRender` to flush fetches means that we can't rehydrate properly on the initial render. This fixes that for the common case of routes, but likely does not fix anything for components that fetch data.

cc @krisselden


runspired:
Test Failure is known `ENV` flag issue against ember canary.

PR #4999, opened 2017-05-26
Adding test unregister model

Leooo:
- [x] adding failing test for https://github.com/emberjs/ember.js/issues/15268

(uncomment `//delete owner.__container__.factoryCache['model:baz'];` and the test passes)

Issue #5000, opened 2017-05-26
unregistering then re-registering a model doesn't clear the factoryCache


Leooo:
copy of https://github.com/emberjs/ember.js/issues/15268

failing test: https://github.com/emberjs/data/pull/4999

rwjblue:
Ember Data is doing its own caching around `factoryFor` and therefore has another layer that needs to be busted (which cannot be done by `owner.unregister`).  See `store._modelFactoryFor` implementation [here](https://github.com/emberjs/data/blob/58bd3960071784304a5bc5b6c8dac2a50358951b/addon/-private/system/store.js#L2092).

There are a few different possibilities here that I can think of:

1. Expose a `store.unregisterModelFactory` method or something to bust the store's internal model factory cache.
2. Remove the secondary caching layer and rely on `owner.factoryFor`'s internal caching.
3. Do not support changing a registered model after any instances have been created.

I personally am in the camp of both 2 and 3 😸, but there are likely good reasons for the extra layer of caching (e.g. to support `.modelName` system).

rwjblue:
@runspired may have more insight round item 2 above (e.g. if we do need the secondary caching level)...

runspired:
I don't believe we should allow cache-busting in ember-data for `modelFor`.  I left a few comments here: https://github.com/emberjs/data/pull/5443

TL;DR In the case described in the original Ember issue above, the following refactor would be better.

**before**
```
import MyMixin from '<app-name>/mixins/my-mixin';

moduleForModel('appointment', {
  integration: true,
  beforeEach() {
    let store = this.store();
    let MyModel = store.modelFor('my-model');
    let MyModel2 = MyModel.extend(MyMixin);
    getOwner(store).unregister('model:my-model');
    getOwner(store).register('model:my-model', MyModel2);
    let record = store.createRecord('my-model'); // BUG not extended
    record.get('aMixinProp')); // 'aMixinProp' is a property of the mixin, but is not available.
  }
});
```


**after**
```
import MyMixin from '<app-name>/mixins/my-mixin';

moduleForModel('appointment', {
  integration: true,
  beforeEach() {
    let store = this.store();
    let owner = getOwner(store);
   
    // we lookup the factory via the owner, not via modelFor
    let MyModelFactory = owner.factoryFor('model:my-model');
    let MyModel2 = MyModelFactory.klass.extend(MyMixin);
    
    owner.unregister('model:my-model');
    owner.register('model:my-model', MyModel2);
 
[...]

PR #5443, opened 2018-04-19
Add test for unregistering a model

Leooo:
replaces https://github.com/emberjs/data/pull/4999

### DIFF
--- a/tests/unit/unregister-model-test.js
+++ b/tests/unit/unregister-model-test.js
@@ -1,0 +1,42 @@
+import { run } from '@ember/runloop';
+import setupStore from 'dummy/tests/helpers/store';
+import { module, test } from 'qunit';
+import DS from 'ember-data';
+import { getOwner } from 'ember-data/-private';
+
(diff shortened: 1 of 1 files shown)


runspired:
@Leooo I have a couple of questions:

1 - I agree that unregistering / re-registering ought to be possible at times (like for instance, via an initializer). It's not a great pattern, but I don't think there's a strong reason to prevent it.  However, I wouldn't expect this to be possible once you've called `modelFor`. At that point, it seems the only approach capable of working would be `reopen` given we'd likely have already created instances.  Even the reopen approach would be error-prone as we cache attributes and relationships fairly heavily.

Would you agree that altering a model after `modelFor` has been called is not a reasonable thing to allow?

2 - Why not do this via a module level export and extend?

runspired:
Closing as we don't wish to support this form of dynamic re-definition of models. See discussion in #5000 for alternatives and reasoning.

Thank you very much @Leooo for taking the time on this one.

PR #5445, opened 2018-04-22
[FEAT] improved assertions for invalid relationship payloads

runspired:
Previously pushing an invalid relationship payload from a `has-many` would not error while a `belongs-to` would but was untested. Added tests for both.

### DIFF
--- a/addon/-private/system/store.js
+++ b/addon/-private/system/store.js
@@ -2677,11 +2677,7 @@ Store = Service.extend({
     if (isNone(resourceIdentifier)) {
       return;
     }
-    assert(`A ${relationship.internalModel.modelName} record was pushed into the store with the value of ${relationship.key} being '${JSON.stringify(resourceIdentifier)}', but ${relationship.key} is a belongsTo relationship so the value must not be an array. You should probably check your data payload or serializer.`, !Array.isArray(resourceIdentifier));
-    assert(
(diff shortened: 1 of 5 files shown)


addon/-private/system/store.js
```diff
@@ -2713,10 +2713,12 @@ Store = Service.extend({
     if (isNone(resourceIdentifier)) {
       return;
     }
+    assert(`A ${relationship.internalModel.modelName} record was pushed into the store with the value of ${relationship.key} being '${JSON.stringify(resourceIdentifier)}', but ${relationship.key} is a belongsTo relationship so the value must not be an array. You should probably check your data payload or serializer.`, !Array.isArray(resourceIdentifier));
+    assert(
+      `Ember Data expected the data for the ${relationship.key} relationship on a ${relationship.internalModel.toString()} to be in a JSON API format and include an \`id\` and \`type\` property but it found '${JSON.stringify(resourceIdentifier)}'. Please check your serializer and make sure it is serializing the relationship payload into a JSON API format.`,
```
igorT: Is this after us normalizing the id? I'm a bit worried we could get trolled here by id being 0

runspired: I believe so but I’ll double check.

runspired: Good catch, addressed.

PR #5446, opened 2018-04-22
[CLEANUP BUGFIX] fix model and factory lookup

### LINKED ISSUE #4711: We should deprecate the (private) modelFactoryFor
We largely have it so that in a few places we can warn instead of error when an unknown `type` is encountered by a serializer when normalizing a payload, however I'm fairly certain this will still hard error shortly thereafter in the actual push.

runspired:
Resolves #4711
Resolves an issue with `createRecord` failing when being called for a non-existent model
Fixes an issue found with `toString` implementation in #5117 

Example Fail case prior to this PR:

```
let person = run(() => store.push({
    data: {
      type: 'person',
      id: '1'
    }
}));

person.constructor.modelName; // will be unset
```

Previously, we had overly complicated the process of looking up a model with a mixture of public and private and public-looking but private methods.  This complexity resulted in there being multiple code-paths by which a `ModelClass` could be looked up without a `modelName` or in which no factory for the `ModelClass` would be found but we would still attempt to `create` an instance from a missing factory.

Much of this complexity stemmed from the desire to check whether a `ModelClass` existed. I have strengthened our private `_hasModelFor` implementation, and trimmed us down to a single public `modelFor` hook and a single private `_modelFactoryFor` hook to eliminate confusion and prevent these sorts of errors from occurring.

Deprecates

- `_modelForMixin`
- `_modelFor`
- `modelFactoryFor`

### DIFF
--- a/addon/-private/system/model/internal-model.js
+++ b/addon/-private/system/model/internal-model.js
@@ -172,7 +172,7 @@ export default class InternalModel {
   }
 
   get modelClass() {
-    return this._modelClass || (this._modelClass = this.store._modelFor(this.modelName));
+    return this._modelClass || (this._modelClass = this.store.modelFor(this.modelName));
   }
 
   get type() {
@@ -391,7 +391,7 @@ export default class InternalModel {
         createOptions.container = this.store.container;
       }
 
-      this._record = this.store.modelFactoryFor(this.modelName).create(createOptions);
+      this._record = this.store._modelFactoryFor(this.modelName).create(createOptions);
 
       this._triggerDeferredTriggers();
       heimdall.stop(token);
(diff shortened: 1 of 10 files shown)


runspired:
cc @samselikoff @dingoeatingfuzz

runspired:
cc @exelord for `ember-custom-actions`

This: https://github.com/Exelord/ember-custom-actions/blob/9c3f89723bb54a51a6bf480b866e8b458bf7ba54/addon/serializers/rest.js#L20

Will become this: `store._hasModelFor(modelName)`.

This is the only usage found in the wild by ember-observer code-search for the now deprecated version of `modelFactoryFor` (besides ember-m3 which I've opened an issue for).

runspired:
@exelord I also meant to tell you that `_hasModelFor` is backwards compatible and is what should already have been in use there, it's safe to preemptively refactor to that now :)

Exelord:
Ok! Thanks a lot Chris! I already migrated to it in the local branch. I will update the package soon :)

PR #5453, opened 2018-04-25
remove unused import

stefanpenner:


stefanpenner:
This may want to be backported, as it adds several hundred MS to all ember-cli app boots.

runspired:
I'm guessing we forgot to remove this when we stopped doing the globals build? If so, backporting it wouldn't do much as we only removed that in 3.2 beta.

runspired:
Seems we grab babel here though: https://github.com/emberjs/data/pull/5453/files#diff-168726dbe96b3ce427e7fedce31bb0bcL143

Perhaps we forgot to remove this even further back.

stefanpenner:
@runspired ya that backport would be different, it would be to on-demand require the module.

PR #5455, opened 2018-04-26
[BUGFIX] destroyRecord should also unload

fivetanley:
It seems common for APIs to return a paylaod when deleting a record.
Although there is a test for when `undefined` is returned from
`deleteRecord` in the adapter here: https://github.com/emberjs/data/blob/master/tests/integration/records/delete-record-test.js#L87

There are currently no tests showing what happens when you return a
payload. It seems like the correct behavior for destroyRecord should be
`deleteRecord` + `save` + `unloadRecord`.

### DIFF
--- a/tests/integration/records/delete-record-test.js
+++ b/tests/integration/records/delete-record-test.js
@@ -139,6 +139,77 @@ test('deleting a record that is part of a hasMany removes it from the hasMany re
   assert.equal(group.get('people.length'), 1, 'expected 1 related records after delete');
 });
 
+test('deleting a record that is part of a hasMany removes it from the hasMany recordArray when adapter deleteRecord function returns a payload', function(assert) {
+  let group;
+  let person;
+  const Group = DS.Model.extend({
+    people: DS.hasMany('person', { inverse: null, async: false })
+  });
+  Group.toString = () => { return 'Group'; }
+
+  env.adapter.deleteRecord = function() {
+    return EmberPromise.resolve({
+      data: {
+        id: '1',
+        type: 'group',
+        attributes: {}
+      }
+    });
+  };
+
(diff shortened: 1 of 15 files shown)


tests/integration/records/delete-record-test.js
```diff
@@ -136,6 +136,77 @@ test('deleting a record that is part of a hasMany removes it from the hasMany re
...
+  let group;
+  let person;
+  const Group = DS.Model.extend({
+    people: DS.hasMany('person', { inverse: null, async: false })
+  });
+  Group.toString = () => { return 'Group'; }
+
+  env.adapter.deleteRecord = function() {
+    return EmberPromise.resolve({
+      data: {
```
fivetanley: it's important to note that no matter what this value is, as long as it is not `undefined`, the test seems to still fail.

tests/integration/records/delete-record-test.js
```diff
@@ -165,140 +216,171 @@ test('records can be deleted during record array enumeration', function(assert)
...
-    'The new person should be removed from the store'
-  );
-});
+    await record.destroyRecord();
 
-test('Destroying an invalid newly created record should remove it from the store', function(assert) {
-  var record;
-  var store = env.store;
+    assert.equal(
+      internalModel.currentState.stateName,
```
runspired: Cc @rwjblue

btecu:
Any progress here?

snewcomer:
@btecu We have a few PRs here (one of which is a direct port of this PR) and are quite close.  Just need to put in the time to review.

#7226 will unblock #7258 

Hopefully going to review the native model refactor today.

runspired:
Closing in favor of #7258

Issue #5452, opened 2018-04-25
Clearing a belongsTo relationship using the JSON-API seems to be broken in 3.2.0-beta.2


eodb:
We have several belongsTo relationships that can be cleared by the server when pushing new updates to the client using WebSockets and the JSON-API.
Starting from 3.2.0-beta1, this seems to have no effect anymore and after the update the belongsTo records are still reachable.

I've written a small QUnit test that fits within the ED test framework in order to reproduce it. This works fine with Ember Data 3.1.1 but fails on the last assert when using Ember Data 3.2.0-beta.2.

- Eddy

```javascript
import { run } from '@ember/runloop';
import DS from "ember-data";
import { createStore } from '../store';
import { module, test } from "ember-qunit";

let store;
let User, Company;

module('integration/relationship/Clear async relationship', {
    beforeEach() {
        User = DS.Model.extend({
            name: DS.attr('string'),
            companyId: DS.belongsTo('company', {async: true})
        });

        Company = DS.Model.extend({
            name: DS.attr('string')
        });

        store = createStore({
            user: User,
            company: Company
        });

        Company = store.modelFor('company');
        User = store.modelFor('user');
    }
});

test('clear async belongsTo relationship', function (assert) {
    let companies, users;

    run(() => {
        companies = store.push({
            data: [{
                id: 'c1',
                type: 'company',
                attributes: {
                    name: "My Company"
                }
            }]
        });

        users = store.push({
            data: [{
                id: 'u1',
                type: 'user',
                attributes: {
                    name: "Peeters"
                },
                relationships: {
                    companyId: {
                        data: {
                            id: 'c1',
                            type: 'company'
                        }
                    }
                }
            }]
        });
    });

    assert.equal(companies.get('firstObject.name'), 'My Company');
    assert.equal(users.get('firstObject.companyId.name'), 'My Company');

    run(() => {
        store.push({
            "data": [{
                "id": 'u1',
                "type": 'user',
                "attributes": {
                    "name": 'Jef'
                },
                "relationships": {
                    companyId: {
                        data: null
                    }
                }
            }]
        });
    });

    run(() => {
        assert.equal(users.get('firstObject.name'), 'Jef');
        assert.equal(users.get('firstObject.companyId.name'), undefined);
    });
});
```

runspired:
@eodb could you PR this test? :)

runspired:
(and if it fails, could you PR it as a skipped test)

PR #5457, opened 2018-04-27
async belongs to test: failing QUnit test for issue #5452

eodb:
added failing QUnit test for issue #5452
Seems that my PR contains too many commits. I'm still new when it concerns the use of git...
Only relevant commit is the adding of a test 'clear async belongsTo relationship' to the belongs-to-test...


runspired:
@eodb thank you so much for the test PR! It turned out to be the same issue that @fivetanley reported recently as well. There is a fix available in #5461 which I will try to get backported to 3.2 beta :)

PR #5460, opened 2018-04-29
Try out travis's allow_failures option

bmac:


runspired:
@rwjblue @hjdivad what do you folks think on this? I'm mixed myself. We've been catching a ton of regressions against ember canary lately.

The motivation for this PR is that it's difficult to backport fixes release patch releases of ember-data with green builds against canary and beta of Ember.

However, I don't want us to stop paying attention to the early warnings of Ember canary having broken us.

Thoughts?

hjdivad:
I'm not keen on it.  Risk of missing regressions seems high.

> However, I don't want us to stop paying attention to the early warnings of Ember canary having broken us.

I agree.

runspired:
@rwjblue felt similarly when discussed in person. Seems better to merge a red backport than to have unnoticed regressions until a late hour.

PR #5462, opened 2018-05-01
[BUGFIX] ensure ManyArray state is in-sync with relationship state

runspired:
This PR extracts the non-diff changes from #5269. In general this is just a nice cleanup PR.

It ideally would include a test to show the issue with ManyArray losing sync that it addresses; however, because we do not currently correctly diff when flushing canonical state (thereby losing local changes), this out-of-sync issue is very hard to manifest as in most scenarios we would be in the same incorrect state even though we were drawing our state from the incorrect source.  To be out-of-sync required the currentState to be different from canonicalState post flushCanonical for more changes than just `isNew` records.

I've taken the opportunity to port the tests from #5269 and converted them to `todo`. While fixing the ManyArray state won't resolve those tests, it will move our ability to fix "currentState" in ManyArray forward, and this fix was needed for the full solution in that PR.

### DIFF
--- a/addon/-legacy-private/system/many-array.js
+++ b/addon/-legacy-private/system/many-array.js
@@ -136,9 +136,7 @@ export default EmberObject.extend(MutableArray, Evented, {
   },
 
   objectAt(index) {
-    if (this.relationship._willUpdateManyArray) {
-      this.relationship._flushPendingManyArrayUpdates();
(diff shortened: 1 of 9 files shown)


runspired:
cc @fivetanley this may help you, but it's just a guess

Issue #5465, opened 2018-05-04
Tests instantiating models with relationships fail on `ember-canary`


bendemboski:
[Here](https://github.com/bendemboski/destroyed-container-bug) is an addon demonstrating the issue. It's just a vanilla addon with `ember-data` installed and a couple of models generated, `user` which `belongsTo` `organization`.

If you run `ember try:one ember-canary` (as of the filing time of this issue), it will fail because it hits [this](https://github.com/emberjs/ember.js/commit/5e74fc385aa005329888c3f8eaa362a1ba7be18e#diff-23fc6627b1cb66883dbaa7b89c287c26R143) newly-added assertion:

```
not ok 12 Chrome 66.0 - Unit | Model | user: it exists
    ---
        actual: >
            null
        stack: >
            Error: Assertion Failed: expected container not to be destroyed
                at new EmberError (http://localhost:7357/assets/vendor.js:18994:25)
                at assert (http://localhost:7357/assets/vendor.js:16297:15)
                at Container.factoryFor (http://localhost:7357/assets/vendor.js:37787:58)
                at Class.factoryFor (http://localhost:7357/assets/vendor.js:59270:33)
                at Class.modelFactoryFor (http://localhost:7357/assets/vendor.js:86413:20)
                at Class._modelFactoryFor (http://localhost:7357/assets/vendor.js:86376:22)
                at Class._modelFor (http://localhost:7357/assets/vendor.js:86367:29)
                at Class.modelFor (http://localhost:7357/assets/vendor.js:86360:17)
                at Function.typeForRelationship (http://localhost:7357/assets/vendor.js:77253:34)
                at Function._findInverseFor (http://localhost:7357/assets/vendor.js:77311:28)
        message: >
            Promise rejected after "it exists": Assertion Failed: expected container not to be destroyed
        Log: |
    ...
```

This is happening because when the store is unloading the `organization` record (resulting from tearing down the test context), it looks at its relationships, which ultimately results in a `modelFor` call to look up the `user` model at the other end of the relationship, which looks it up in the container, which is already destroyed. It looks like this has been happening for a long time, but is just now failing because of the newly-added assertion. I'm not sure if `ember-data`'s behavior is incorrect, or if the assertion is overly aggressive.

rwjblue:
> I'm not sure if ember-data's behavior is incorrect, or if the assertion is overly aggressive.

Probably a bit of both. 😆

[...]

PR #5466, opened 2018-05-05
Serialize empty hasMany relationships

ryanto:
This is a fix for the issue [reported here](https://github.com/emberjs/data/pull/5324#issuecomment-386028562). We will now serialize an empty list when the relationship has no records.

This PR removes the `if (nonNewHasMany.length > 0) {` guard and restructures some of the code.


ryanto:
Tests on canary are failing due to the issue reported here: https://github.com/emberjs/data/issues/5465

runspired:
@ryanto thanks! I have a fix pending for #5465 in #5467

ryanto:
Nice! I'll pull that into my branch as soon as it's merged.

PR #5477, opened 2018-05-25
[BUGFIX] normalize model name for belongs to relationships

agrobbin:
Given these models:

```ruby
const User = DS.Model.extend({
  userProfile: DS.belongsTo(),

  streamItems: DS.hasMany()
});

const UserProfile = DS.Model.extend({
  user: DS.belongsTo()
});

const StreamItem = DS.Model.extend({
  user: DS.belongsTo()
});
```

The inferred model type of `User.streamItems` will be `stream-item` while the type of `User.userProfile` will be `userProfile`.

This goes against the concept of all model types being dasherized.

With this change, `User.userProfile` will be converted into `user-profile`.

I'm not exactly sure where this kind of thing should be tested. It seems like it belongs in same place that the `hasMany` model type normalization tests reside, but I'm having trouble finding that place!

### DIFF
--- a/tests/unit/model/relationships-test.js
+++ b/tests/unit/model/relationships-test.js
@@ -66,3 +66,53 @@ test('eachRelatedType() iterates over relations without duplication', function(a
 
   assert.deepEqual(relations, ['occupation', 'person']);
 });
+
+test('normalizing belongsTo relationship names', function(assert) {
+  const UserProfile = DS.Model.extend({
+    user: DS.belongsTo(),
+  });
+
+  let User = DS.Model.extend({
+    userProfile: DS.belongsTo(),
+  });
+
+  store = createStore({
+    user: User,
+    userProfile: UserProfile,
+  });
+
+  User = store.modelFor('user');
+
+  const relationships = get(User, 'relationships');
+
+  assert.ok(relationships.has('user-profile'), 'relationship key has been normalized');
+
+  const relationship = relationships.get('user-profile')[0];
+
+  assert.equal(relationship.meta.name, 'userProfile', 'relationship name has not been changed');
+});
+
+test('normalizing hasMany relationship names', function(assert) {
+  const StreamItem = DS.Model.extend({
+    user: DS.belongsTo(),
+  });
+
+  let User = DS.Model.extend({
+    streamItems: DS.hasMany(),
+  });
+
+  store = createStore({
+    user: User,
+    streamItem: StreamItem,
+  });
+
+  User = store.modelFor('user');
+
+  const relationships = get(User, 'relationships');
+
+  assert.ok(relationships.has('stream-item'), 'relationship key has been normalized');
+
+  const relationship = relationships.get('stream-item')[0];
+
+  assert.equal(relationship.meta.name, 'streamItems', 'relationship name has not been changed');
+});
--- a/addon/-private/system/relationship-meta.js
+++ b/addon/-private/system/relationship-meta.js
@@ -6,9 +6,12 @@ export function typeForRelationshipMeta(meta) {
   let modelName;
 
   modelName = meta.type || meta.key;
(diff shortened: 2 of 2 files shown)


runspired:
@agrobbin if you have time to add tests, I can help you locate the right spot and draft them :)

agrobbin:
@runspired absolutely, just point me in the right direction and I'll add them!

agrobbin:
@runspired were you able to locate where tests should go for this change?

runspired:
@agrobbin haven't looked yet, sorry. I should get back to you this afternoon though :)

bmac:
Unless @runspired has a better suggestion I think it would be ok to add a test for this change to [tests/unit/model/relationships-test.js](https://github.com/emberjs/data/blob/master/tests/unit/model/relationships-test.js).

tests/unit/model/relationships-test.js
```diff
@@ -66,3 +66,43 @@ test('eachRelatedType() iterates over relations without duplication', function(a
 
   assert.deepEqual(relations, ['occupation', 'person']);
 });
+
+test('normalizing belongsTo relationship names', function(assert) {
+  const UserProfile = DS.Model.extend();
+
+  let User = DS.Model.extend({
+    userProfile: DS.belongsTo(),
```
runspired: same comment as below, this should assert and I'm surprised it does not, UserProfile needs the inverse defined.

tests/unit/model/relationships-test.js
```diff
@@ -66,3 +66,43 @@ test('eachRelatedType() iterates over relations without duplication', function(a
...
+  const relationship = relationships.get('user-profile')[0];
+
+  assert.equal(relationship.meta.name, 'userProfile');
+});
+
+test('normalizing hasMany relationship names', function(assert) {
+  const StreamItem = DS.Model.extend();
+
+  let User = DS.Model.extend({
+    streamItems: DS.hasMany(),
```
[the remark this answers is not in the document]
agrobbin: Not sure why it didn't assert, but I added the inverse relationship to `StreamItem` above.

tests/unit/model/relationships-test.js
agrobbin: Ditto here, I added the inverse relationship definition to `UserProfile` above.

tests/unit/model/relationships-test.js
```diff
@@ -66,3 +66,47 @@ test('eachRelatedType() iterates over relations without duplication', function(a
...
+    user: User,
+    streamItem: StreamItem,
+  });
+
+  User = store.modelFor('user');
+
+  const relationships = get(User, 'relationships');
+  const relationship = relationships.get('stream-item')[0];
+
+  assert.equal(relationship.meta.name, 'streamItems');
```
runspired: give asserts a meaningful description as the last arg (same for the other test).

tests/unit/model/relationships-test.js
```diff
@@ -66,3 +66,47 @@ test('eachRelatedType() iterates over relations without duplication', function(a
...
+
+  store = createStore({
+    user: User,
+    streamItem: StreamItem,
+  });
+
+  User = store.modelFor('user');
+
+  const relationships = get(User, 'relationships');
+  const relationship = relationships.get('stream-item')[0];
```
runspired: we likely want an assertion that we did in fact normalize to `stream-item` from `streamItems` here, instead of just relying on it implicitly being tested due to how we access. Same with the test above.

runspired:
@agrobbin saw you resolved a few comments, left a few more. Resolve those and  we should be good :)

agrobbin:
@runspired just pushed some more updates!

agrobbin:
Thanks for all of the help @runspired and @bmac! Looking forward to seeing this included in the next Ember Data release.

Issue #5472, opened 2018-05-09
unloadRecord is not under as much test coverage as we think it is


runspired:
I recently dove into a few scenarios where unloadRecord was behaving badly but which seemed as though they should have failed tests. I discovered:

- our tests do not cover `deleteRecord` and `destroyRecord` scenarios
  - as a subset of this, it turns out that `destroyRecord` + `unloadRecord` moves the record to an empty non-deleted, non-destroyed (simply non-loaded) state 😱 
- our tests do not cover `createRecord` scenarios where relationships are involved
- our unloadRecord tests generally do not test that in the scenarios were we expect an unload to fully purge the record and internalModel (e.g. disentangled + unloaded) that this actually happens, only that the appearance of it happens.

runspired:
@fivetanley I suspect this is why our `destroyRecord` also unloading PR produced failures for you in #5455 

Also related: #5006

Gabbyjose:
Hello, I was wondering if this issue is still open? I'm a beginning and would be interested in learning more!

Vincz:
Hi! I tried the latest ember-data version (3.2.0-beta.2) and the problem described here https://github.com/emberjs/data/issues/5328 is still here. I'm stuck in 2.12 because of this.
The scenario is as follow:
1) Receive a bunch of model to delete from a websocket
2) Peek the record with store.peekRecord
3) Unload the record with object.unloadRecord()

After that, I get a bunch of 404, ember data is trying to get the record again.

runspired:
@Gabbyjose we're still working on this one, I've added some tests but we're not quite done. Sorry for not getting back to you sooner, last month was a bit... [out there](https://runspired.com/2018/05/17/the-week-that-time-stood-still/). If you'd still like to help I'd love to pair with you!

runspired:
@Vincz this issue has nothing to do with #5328. unloadRecord isn't for "forgetting" records, it is for marking them as stale (e.g. needs to be reloaded). However, there is a way to delete records which we will codify soon (been working on an RFC). The gist of it is [here](https://gist.github.com/runspired/96618af26fb1c687a74eb30bf15e58b6/)

Gabbyjose:
@runspired no worries at all - crazy what happened! i'm definitely still happy to take a stab at this and work with you. please let me know what works best for you!

runspired:
[...]

PR #5480, opened 2018-05-30
Add tests for destroyRecord in hasMany/belongsTo relationships

jmar910:
Added tests for using `destroyRecord` in relationships. We found that in a hasMany relationship that uses `links` to load the relationships, when calling `destroyRecord`, it does not remove the record in the hasMany array. See failing test. 

Related #5472


runspired:
closed with #7258

Issue #5486, opened 2018-06-05
DNS Resolution Error is handled as a generic error not TimeoutError


BryanCrotaz:
`findRecord` always returns the same error regardless of why.

Possible reasons:
* DNS did not resolve
* Timeout
* Server 5xx error
* Server 4xx error
* Malformed response

The application may want to react to these in different ways. We should give visibility in the exception response of why the failure happened.

BryanCrotaz:
Whoops - I was using typeof and not instanceof.

But I did find a case that's missed, when DNS fails to resolve on Chrome on Windows. PR supplied.

runspired:
Now that we've moved to RequestManager I think we can close this

PR #5487, opened 2018-06-05
handle DNS resolution failure in chrome as TimeoutError

BryanCrotaz:
As referenced in #5486

### DIFF
--- a/addon/adapters/rest.js
+++ b/addon/adapters/rest.js
@@ -1518,6 +1518,9 @@ function ajaxError(adapter, jqXHR, requestData, responseData) {
 
   if (responseData.errorThrown instanceof Error) {
     error = responseData.errorThrown;
+  } else if (responseData.errorThrown === "" && responseData.textStatus === "error") {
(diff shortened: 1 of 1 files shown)


BryanCrotaz:
Don't know how to go about writing tests for this

BryanCrotaz:
Not sure this is correct. Will close and reopen a new PR later

runspired:
@BryanCrotaz I think this answer has enough info for us to get it right: https://stackoverflow.com/questions/3825581/does-an-http-status-code-of-0-have-any-meaning

BryanCrotaz:
unfortunately I couldn't find a way to differentiate between an DNS resolution error and all cross domain errors. Cross domain 404 or 500 comes back as status=0 :(

runspired:
:'( maybe we need to add a "generic connection error"

BryanCrotaz:
can we spot from the url and document.location that this is a cross domain request and give a different error?

PR #5489, opened 2018-06-14
[BUGFIX] Lowercase header names in parseResponseHeaders

### LINKED ISSUE #4952: If HTTP response headers are expected to be capitalized then parseResponseHeaders should capitalize them
Hey guys,
When I get an error from my API this is what I see:
```
Ember Data Request PATCH /api/professionals/1044 returned a 403
Payload (Empty Content-Type)
[object Object]
```

Although my API responds with `content-type: "application/json; charset=utf-8"` this info is ignored because [only the capitalized header is searched](https://github.com/emberjs/data/blob/24f41209b7ab9978bcdf029e50e3e62564d0b50f/addon/adapters/rest.js#L1187) (`headers["Content-Type"]`)

Maybe [parseResponseHeaders](https://github.com/emberjs/data/blob/2edaf160a8a437cc989b2e6c980cc9c7d525c2c2/addon/-private/utils/parse-response-headers.js) can capitalize the header key names?

Here is a screenshot of my specific case:
screen shot 2017-04-28 at 15 01 17

jmar910:
Closes #4952.

Pretty sure I got all the places we rely on uppercased header names, like in `generatedDetailedMessage` 🤞

### DIFF
--- a/addon/-private/utils/parse-response-headers.js
+++ b/addon/-private/utils/parse-response-headers.js
@@ -29,6 +29,9 @@ export default function parseResponseHeaders(headersString) {
     let value = header.substring(j + 1, header.length).trim();
 
     if (value) {
+      let lowerCasedField = field.toLowerCase();
+
+      headers[lowerCasedField] = value;
       headers[field] = value;
     }
   }
--- a/tests/unit/utils/parse-response-headers-test.js
+++ b/tests/unit/utils/parse-response-headers-test.js
@@ -21,7 +21,7 @@ test('header parsing', function(assert) {
 
   let headers = parseResponseHeaders(headersString);
 
-  assert.equal(headers['Content-Encoding'], 'gzip', 'parses basic header pair');
+  assert.equal(headers['content-encoding'], 'gzip', 'parses basic header pair');
   assert.equal(
     headers['content-type'],
     'application/json; charset=utf-8',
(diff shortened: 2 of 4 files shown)


addon/-private/utils/parse-response-headers.js
```diff
@@ -25,7 +25,10 @@ export default function parseResponseHeaders(headersString) {
       continue;
     }
 
-    let field = header.substring(0, j).trim();
+    let field = header
+      .substring(0, j)
+      .trim()
+      .toLowerCase();
```
bmac: Would you also mind adding the original "unlowercased" field name to the headers object so anyone's app that is relying the the current behavior continues to work?

jmar910:
@bmac feedback addressed!

runspired:
@bmac I feel per-spec we should deprecate the non-lower-cased version in a follow-up, unsure how to implement and probably needs an RFC, otherwise mega +1 on this

tests/unit/utils/parse-response-headers-test.js
```diff
@@ -92,7 +94,7 @@ test('ignores headers that do not contain a colon', function(assert) {
 
   let headers = parseResponseHeaders(headersString);
 
-  assert.deepEqual(headers['Content-Encoding'], 'gzip', 'parses basic header pair');
+  assert.deepEqual(headers['content-encoding'], 'gzip', 'parses basic header pair');
   assert.deepEqual(headers['apple'], 'pie', 'parses basic header pair');
   assert.equal(Object.keys(headers).length, 2, 'only has the one valid header');
```
runspired: This assert's description is wrong (we had two and now we have three valid headers lol). The valid headers will be `content-encoding` `Content-Encoding` and `apple`

jmar910:
Oh whoops yep, I glossed over the fact that this changes the tests I
updated too 👌🏼

Will address.

jmar910:
Anddddd we're back, should be good now @runspired

PR #5490, opened 2018-06-14
Added check for implicitRelationships and relationships present  in m…

dnalagatla:
…odelData

### DIFF
--- a/addon/-record-data-private/system/relationships/state/relationship.js
+++ b/addon/-record-data-private/system/relationships/state/relationship.js
@@ -179,6 +179,16 @@ export default class Relationship {
     return this.inverseKey && !this.inverseIsAsync;
   }
 
+  _hasSupportForImplicitRelationships(modelData) {
+    return (
+      modelData._implicitRelationships !== undefined && modelData._implicitRelationships !== null
+    );
+  }
+
+  _hasSupportForRelationships(modelData) {
+    return modelData._relationships !== undefined && modelData._relationships !== null;
+  }
+
   get _inverseMeta() {
     if (this.__inverseMeta === undefined) {
       let inverseMeta = null;
@@ -205,6 +215,9 @@ export default class Relationship {
     // we actually want a union of members and canonicalMembers
     // they should be disjoint but currently are not due to a bug
     this.forAllMembers(inverseModelData => {
+      if (!this._hasSupportForRelationships(inverseModelData)) {
+        return;
+      }
       let relationship = inverseModelData._relationships.get(this.inverseKey);
       relationship.inverseDidDematerialize(this.modelData);
     });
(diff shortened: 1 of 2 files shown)


tests/integration/relationships/belongs-to-test.js
```diff
@@ -1725,3 +1732,75 @@ test("belongsTo relationship with links doesn't trigger extra change notificatio
...
+          },
+        },
+        included: [{ type: 'book1', id: '1' }],
+      });
+    });
+
+    let chapter = env.store.peekRecord('chapter', '1');
+    let book1 = env.store.peekRecord('book1', '1');
+
+    assert.ok(chapter.get('book1'));
```
runspired: We should make this a more meaningful assertion that we get what we expect here.

tests/integration/relationships/belongs-to-test.js
```diff
@@ -1725,3 +1732,75 @@ test("belongsTo relationship with links doesn't trigger extra change notificatio
...
+
+    // No inverse setup created for book1
+    // as Model-Data of book1 doesn't support this
+    // functionality.
+    assert.notOk(book1.get('chapter'));
```
runspired: Since `chapter` is an implicit relationship, this would not work regardless. We should instead perform some actions that would create error if we have not guarded the feature correctly as well as assert that `_implicitRelationships` on the book1 modelData is indeed undefined.

tests/integration/relationships/belongs-to-test.js
```diff
@@ -46,6 +47,11 @@ module('integration/relationship/belongs_to Belongs-To Relationships', {
       chapters: hasMany('chapters', { async: false, inverse: 'book' }),
     });
 
+    Book1 = DS.Model.extend({
+      name: attr('string'),
+      author: belongsTo('author', { async: false }),
```
runspired: Given that in our test the `ModelData` for `book1` does not support relationships, we should consider removing `author` here. We should also add an assertion guarding for the case where `Author` defines  a relationship with an inverse to `book1` but `book1` does not support relationships. It is likely this would error out early due to invalid inverse configuration but being able to error out that "relationships aren't even supported" might be more useful in this scenario.

addon/-record-data-private/system/relationships/state/relationship.js
```diff
@@ -205,6 +213,9 @@ export default class Relationship {
     // we actually want a union of members and canonicalMembers
     // they should be disjoint but currently are not due to a bug
     this.forAllMembers(inverseModelData => {
+      if (!this._hasSupportForRelationships(inverseModelData)) {
```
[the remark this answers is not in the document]
runspired: although hrm, I see why you did not. A polymorphic case where some ModelDatas support and some do not support relationships would be bad. I'm unsure what the best way to evolve this check is to ensure we don't mix them.

addon/-record-data-private/system/relationships/state/relationship.js
```diff
@@ -354,7 +371,10 @@ export default class Relationship {
       if (this.inverseKey) {
         this.removeCanonicalModelDataFromInverse(modelData);
```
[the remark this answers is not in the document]
dnalagatla: good catch, I added the check here. Thank you.

runspired:
Test failure is known ember-canary failure.

PR #5492, opened 2018-06-15
[WIP BUGFIX] Prevent failed relationship fetches from causing infinite re-render and re-fetch

### LINKED ISSUE #5479: [bug][regression] infinite find record triggered from view
Failed record fetch, initiated from view (ex: `record.relationship.relAttr` in `.hbs`), will retry request infinite number of times.

Repo with [example](https://github.com/vadim-p/emberjs-test).
In example, in index route, there are infinite loop of request to `/preset-filters/:id`.
Same for `belongsTo` relationship type.

This issue occurs after this #5376 merge.
With ember-data@3.0.2 it breaks on first retry because of TypeError:
```
GET http://localhost:4200/preset-filters/1 404 (Not Found)
Uncaught TypeError: Cannot read property 'eachAttribute' of null 
    at new Snapshot (-private.js:5041)
    at InternalModel.createSnapshot (-private.js:6940)
    at _find (-private.js:8546)
    at Class._fetchRecord (-private.js:10378)
    at _fetchRecord (-private.js:10441)
    at Class._flushPendingFetchForType (-private.js:10539)
    at cb (ember-metal.js:5356)
    at OrderedSet.forEach (ember-metal.js:5169)
    at MapWithDefault.forEach (ember-metal.js:5364)
    at Class.flushAllPendingFetches (-private.js:10421)
```

```
ember-cli: 3.1.4
ember-source: 3.1.2
ember-data: 3.1.1
```

runspired:
Resolves #5479 

Fix here is only for legacy build, I still need to port changes to the record-data build.
This will also need ported to `3.2` prior to release; however, doing so is likely non-trivial as a cherry-pick, so I will open a PR specific to `3.2` once this PR is complete.

cc @HeroicEric @fivetanley @rwjblue

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -4,6 +4,7 @@ module.exports = {
     ecmaVersion: 2017,
     sourceType: 'module',
   },
+  parser: 'babel-eslint',
   extends: ['eslint:recommended', 'prettier'],
   plugins: ['prettier'],
   rules: {
--- a/addon/-legacy-private/system/relationships/state/belongs-to.js
+++ b/addon/-legacy-private/system/relationships/state/belongs-to.js
@@ -1,7 +1,7 @@
-import { Promise as EmberPromise } from 'rsvp';
+import { resolve } from 'rsvp';
 import { assert } from '@ember/debug';
 import { assertPolymorphicType } from 'ember-data/-debug';
-import { PromiseBelongsTo } from '../../promise-proxies';
+import { PromiseBelongsTo, PromiseObject } from '../../promise-proxies';
 import Relationship from './relationship';
 
 export default class BelongsToRelationship extends Relationship {
@@ -9,7 +9,23 @@ export default class BelongsToRelationship extends Relationship {
     super(store, internalModel, inverseKey, relationshipMeta);
     this.inverseInternalModel = null;
     this.canonicalState = null;
-    this._loadingPromise = null;
+    this._promiseProxy = null;
+  }
+
+  /**
+   * Flag indicating whether all inverse records are available
+   *
+   * true if the inverse exists and is loaded (not empty)
+   * true if there is no inverse
+   * false if the inverse exists and is not loaded (empty)
+   *
+   * @returns {boolean}
+   */
+  get allInverseRecordsAreLoaded() {
+    let internalModel = this.inverseInternalModel;
+    let isEmpty = internalModel !== null && internalModel.isEmpty();
+
+    return !isEmpty;
   }
 
   setInternalModel(internalModel) {
@@ -21,7 +37,6 @@ export default class BelongsToRelationship extends Relationship {
     this.setHasAnyRelationshipData(true);
     this.setRelationshipIsStale(false);
     this.setRelationshipIsEmpty(false);
-    this.setHasRelatedResources(!this.localStateIsEmpty());
   }
 
   setCanonicalInternalModel(internalModel) {
@@ -92,6 +107,7 @@ export default class BelongsToRelationship extends Relationship {
     }
     if (this.inverseInternalModel !== this.canonicalState) {
       this.inverseInternalModel = this.canonicalState;
+      this._promiseProxy = null;
       this.notifyBelongsToChanged();
     }
 
@@ -114,13 +130,17 @@ export default class BelongsToRelationship extends Relationship {
     this.notifyBelongsToChanged();
   }
 
-  setRecordPromise(newPromise) {
-    let content = newPromise.get && newPromise.get('content');
+  setRecordPromise(belongsToPromise) {
     assert(
       'You passed in a promise that did not originate from an EmberData relationship. You can only pass promises that come from a belongsTo or hasMany relationship to the get call.',
-      content !== undefined
+      belongsToPromise instanceof PromiseObject
     );
+
+    let content = belongsToPromise.get('content');
+    let promise = belongsToPromise.get('promise');
+
     this.setInternalModel(content ? content._internalModel : content);
+    this._updateLoadingPromise(promise, content);
   }
 
   removeInternalModelFromOwn(internalModel) {
@@ -128,6 +148,7 @@ export default class BelongsToRelationship extends Relationship {
       return;
     }
     this.inverseInternalModel = null;
+    this._promiseProxy = null;
     super.removeInternalModelFromOwn(internalModel);
     this.notifyBelongsToChanged();
   }
@@ -135,6 +156,7 @@ export default class BelongsToRelationship extends Relationship {
   removeAllInternalModelsFromOwn() {
     super.removeAllInternalModelsFromOwn();
     this.inverseInternalModel = null;
+    this._promiseProxy = null;
     this.notifyBelongsToChanged();
   }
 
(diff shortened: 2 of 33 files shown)


rwjblue:
How is this looking? I’d really like to get this landed and release 3.2..

runspired:
@rwjblue need to add the fix you suggested for LTS, otherwise I think it's ready

.eslintrc.js
```diff
@@ -1,5 +1,6 @@
 module.exports = {
   root: true,
+  parser: 'babel-eslint',
```
runspired: I did not add this manually, unsure how this came in. @rwjblue any ideas?

rwjblue: Rebase issue maybe? I don’t think we should need a special parser should be able to remove...

addon/-legacy-private/system/relationships/state/belongs-to.js
```diff
@@ -156,11 +169,50 @@ export default class BelongsToRelationship extends Relationship {
...
+        promise = inverseInternalModel.getRecord().reload();
+      } else {
+        promise = this.store._findByInternalModel(inverseInternalModel);
+      }
+      promise = promise.then(
+        () => {
+          return this._handleCompletedFind();
+        },
+        e => {
+          return this._handleCompletedFind(e);
```
igorT: splitting this into two methods might be much nicer

addon/-legacy-private/system/relationships/state/belongs-to.js
```diff
@@ -156,11 +169,50 @@ export default class BelongsToRelationship extends Relationship {
...
   findRecord() {
-    if (this.inverseInternalModel) {
-      return this.store._findByInternalModel(this.inverseInternalModel);
+    let promise;
+    let { inverseInternalModel, shouldForceReload } = this;
+
+    if (inverseInternalModel) {
+      if (shouldForceReload && inverseInternalModel.hasRecord) {
+        // reload record, if it is already loaded
+        promise = inverseInternalModel.getRecord().reload();
```
igorT: if we have a link, shouldn't we go fetch the link instead of reloading the record?

addon/-legacy-private/system/relationships/state/has-many.js
```diff
@@ -16,25 +17,36 @@ export default class ManyRelationship extends Relationship {
...
-    } else {
-      this._loadingPromise = PromiseManyArray.create({
-        promise,
-        content,
-      });
+  get currentState() {
+    return this.members.list;
+  }
+
+  get hasRelatedResources() {
```
igorT: Can you please write a comment describing the intended behavior of this method

addon/-legacy-private/system/relationships/state/has-many.js
```diff
@@ -16,25 +17,36 @@ export default class ManyRelationship extends Relationship {
...
+    // check currentState for unloaded records
+    let hasEmptyRecords = this.currentState.reduce((hasEmptyModel, i) => {
```
igorT: Would this be cleaner to write as a find/filter?

addon/-legacy-private/system/relationships/state/has-many.js
igorT: This seems to be maybe equivalent to the `hasLoaded` property? If i'm reading this correctly, it should return true if you have an empty array, in which case the naming is pretty misleading

addon/-legacy-private/system/relationships/state/has-many.js
```diff
@@ -249,13 +264,14 @@ export default class ManyRelationship extends Relationship {
...
-        return this._loadingPromise;
+    if (this._promiseProxy) {
+      if (this._promiseProxy.get('isPending')) {
+        return this._promiseProxy;
       }
     }
 
-    this.setRelationshipIsStale(true);
+    this.setHasFailedLoadAttempt(false);
+    this.setShouldForceReload(true);
```
igorT: Whats the thinking behind setting this on the object/interaction with the scheduleFetchMany?

addon/-legacy-private/system/relationships/state/has-many.js
```diff
@@ -318,60 +334,113 @@ export default class ManyRelationship extends Relationship {
...
     this.internalModel.notifyHasManyAdded(this.key);
   }
 
-  getRecords() {
+  fetchData() {
     //TODO(Igor) sync server here, once our syncing is not stupid
     let manyArray = this.manyArray;
 
-    if (this.isAsync) {
+    if (this.shouldMakeRequest()) {
```
igorT: :heart:

addon/-legacy-private/system/relationships/state/relationship.js
```diff
@@ -87,10 +94,9 @@ export default class Relationship {
       true when
         => !internalModel.isNew() on initial setup
         => an inverse has been unloaded
-        => relationship.reload() has been called
         => we get a new link for the relationship
      */
-    this.relationshipIsStale = !internalModel.isNew();
+    this.relationshipIsStale = !this.isNew;
```
igorT: Can you add a comment as to what the difference between forcing a reload and a stale relationship is

addon/-legacy-private/system/relationships/state/relationship.js
```diff
@@ -515,31 +530,98 @@ export default class Relationship {
...
+      isNew,
+      fetchPromise,
+    } = this;
+
+    // never make a request if this record doesn't exist server side yet
+    if (isNew === true) {
+      return false;
+    }
+
+    // Always make a request when forced
```
igorT: ❤️

addon/-legacy-private/system/store.js
```diff
@@ -1871,7 +1871,7 @@ Store = Service.extend({
       snapshot: snapshot,
       resolver: resolver,
     });
-    emberRun.once(this, this.flushPendingSave);
+    emberRun.scheduleOnce('actions', this, this.flushPendingSave);
```
igorT: What's this about

tests/acceptance/relationships/has-many-test.js
```diff
@@ -0,0 +1,653 @@
+import { module } from 'qunit';
```
igorT: Many ❤️

addon/-legacy-private/system/store.js
runspired: Now that we have setup waiters for our backburner instance, we no longer need the autorun assertion. This allows folks (including ourselves) to use `async await` without wrapping everything in run loops.

addon/-legacy-private/system/relationships/state/relationship.js
runspired: 👍

addon/-legacy-private/system/relationships/state/belongs-to.js
runspired: 👍

addon/-legacy-private/system/relationships/state/has-many.js
runspired: changed to `allInverseRecordsAreLoaded`

addon/-legacy-private/system/relationships/state/has-many.js
runspired: unsure, I find I have to similarly pause and reason for any of the three approaches. Happy to embrace another pattern if there's one you prefer,

PR #5500, opened 2018-06-22
Refactor init setup into record-data

runspired:
Properties passed to createRecord need to be handed to `RecordData` for potential initial state setup.

There is one messy piece remaining that prevents us from seamlessly moving to only using `ModelData` or `References` to communicate relationship membership with `ModelData`. The issue is that users can set the state of a relationship to a promise-proxy (record being requested or left over proxy from previous request, as that is what users often have).

If `createRecord` were async or if we only worked by reference in this area of the public API, we could wait for these promises to resolve and grab either the `modelData` or the `reference` as desired, but because it is not async we are forced to pass through the record promise to our modelData instance which then sets it on the relationship as the promise record for the relationship to deal with.


addon/-record-data-private/system/model/internal-model.js
```diff
@@ -1240,3 +1223,29 @@ export default class InternalModel {
...
+      return A(records).every(record => record.hasOwnProperty('_internalModel') === true);
+    })()
+  );
+}
+
+function extractRecordDatasFromRecords(records) {
+  return records.map(extractRecordDataFromRecord);
+}
+
+function extractRecordDataFromRecord(record) {
```
hjdivad: I'd recommend renaming `record` given that it's a union type.

something like `recordOrPromiseProxy`

```js
function extractRecordDataFromValue(recordOrPromiseProxy) {
  if (/* detect promise proxy case */) {
    // promise proxy
    return recordOrPromiseProxy;
  }

  // record case
  return record._internalModel._modelData;
}
```

runspired:
Test failure is known issue against ember-canary, will try to look into why that hasn't been resolved later tomorrow or Wednesday.

PR #5502, opened 2018-06-24
Update parseResponseHeaders to work in fastboot.

ryanto:
On nix systems the headers will be delimited with only a newline (`\n`) character.

This PR changes the split to `\r?\n`, which will work for all fastboot environments as well as browsers.


ryanto:
hmm. not sure what to do about this failure: https://travis-ci.org/emberjs/data/jobs/396162028

runspired:
@ryanto if you rebase I pinned the offending package and your test run will continue :)

PR #5528, opened 2018-07-16
Add tests for createRecord+unloadRecord

Gabbyjose:
Added tests to cover creating and unloading an owner with associated pet, and creating and unloading a pet with associated owner

see #5472 for motivation

### DIFF
--- a/addon/-record-data-private/system/model/internal-model.js
+++ b/addon/-record-data-private/system/model/internal-model.js
@@ -827,7 +827,7 @@ export default class InternalModel {
     }
   }
 
-  notifyHasManyChange(key, record, idx) {
+  notifyHasManyChanged(key, record, idx) {
     if (this.hasRecord) {
       let manyArray = this._manyArrayCache[key];
       if (manyArray) {
@@ -849,7 +849,7 @@ export default class InternalModel {
     }
   }
 
-  notifyBelongsToChange(key, record) {
+  notifyBelongsToChanged(key, record) {
     if (this.hasRecord) {
       this._record.notifyBelongsToChanged(key, record);
       this.updateRecordArrays();
--- a/addon/-record-data-private/system/model/internal-model.js
+++ b/addon/-record-data-private/system/model/internal-model.js
@@ -319,30 +319,30 @@ export default class InternalModel {
 
     if (this._record) {
       this._record.destroy();
+
+      Object.keys(this._relationshipPromisesCache).forEach(key => {
+        // TODO Igor cleanup the guard
+        if (this._relationshipPromisesCache[key].destroy) {
+          this._relationshipPromisesCache[key].destroy();
+        }
+        delete this._relationshipPromisesCache[key];
+      });
+      Object.keys(this._manyArrayCache).forEach(key => {
+        let manyArray = (this._retainedManyArrayCache[key] = this._manyArrayCache[key]);
+        delete this._manyArrayCache[key];
+
+        if (manyArray && !manyArray._inverseIsAsync) {
+          /*
+            If the manyArray is for a sync relationship, we should clear it
+              to preserve the semantics of client-side delete.
+
+            It is likely in this case instead of retaining we should destroy
+              - @runspired
+          */
+          manyArray.clear();
+        }
+      });
     }
-
-    Object.keys(this._relationshipPromisesCache).forEach(key => {
-      // TODO Igor cleanup the guard
-      if (this._relationshipPromisesCache[key].destroy) {
-        this._relationshipPromisesCache[key].destroy();
-      }
-      delete this._relationshipPromisesCache[key];
-    });
-    // if the manyArray is for a sync relationship
-    //   we should clear it
-    Object.keys(this._manyArrayCache).forEach(key => {
-      let manyArray = (this._retainedManyArrayCache[key] = this._manyArrayCache[key]);
-      delete this._manyArrayCache[key];
-
-      /*
-          It is likely in this case instead of retaining we should destroy
-
-          - @runspired
-        */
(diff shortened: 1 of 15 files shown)


addon/-record-data-private/system/model/internal-model.js
```diff
@@ -315,24 +318,35 @@ export default class InternalModel {
...
+      /*
+          It is likely in this case instead of retaining we should destroy
+
+          - @runspired
+        */
+      if (manyArray && !manyArray._inverseIsAsync) {
+        manyArray.clear();
+      }
+    });
+
```
hjdivad: Under what circumstances do we have entries in `_relationshipPromisesCache` or `_manyArrayCache` but no record?

addon/-record-data-private/system/model/internal-model.js
```diff
@@ -315,24 +318,35 @@ export default class InternalModel {
...
+    //   we should clear it
+    Object.keys(this._manyArrayCache).forEach(key => {
+      let manyArray = (this._retainedManyArrayCache[key] = this._manyArrayCache[key]);
+      delete this._manyArrayCache[key];
+
+      /*
+          It is likely in this case instead of retaining we should destroy
+
+          - @runspired
+        */
```
hjdivad: this comment can probably be one line

addon/-record-data-private/system/model/internal-model.js
```diff
@@ -315,24 +318,35 @@ export default class InternalModel {
...
+      if (manyArray && !manyArray._inverseIsAsync) {
+        manyArray.clear();
```
hjdivad: What's the motivation for this?

addon/-record-data-private/system/relationships/state/belongs-to.js
```diff
@@ -155,6 +157,17 @@ export default class BelongsToRelationship extends Relationship {
       return;
     }
     this.canonicalState = null;
+    /*
+      This isn't exactly correct because another record's payload
+      may tell us that this relationship is no longer correct
+      but that is not enough to tell us that this relationship is
+      now empty for sure. Likely we should be stale here but
+      that is probably a breaking change.
```
hjdivad: Can you rephrase this?  what's an example payload that illustrates this?

tests/integration/records/create-record-test.js
```diff
@@ -0,0 +1,114 @@
...
+
+    chris.unloadRecord();
+
+    assert.ok(pet.get('owner') === null, 'Shen no longer has an owner');
+
+    // check that the relationship has been dissolved
+    pets = chris
+      .get('pets')
+      .toArray()
+      .map(pet => pet.get('name'));
```
hjdivad: is updating the state of unloaded records a new feature here or have we verified this was 2.12 behaviour?

addon/-record-data-private/system/model/internal-model.js
runspired: I suspect that should be an invalid state as of now but wouldn't necessarily be so in the future, moved back within the guard for `hasRecord`.

addon/-record-data-private/system/model/internal-model.js
runspired: Cleaned up this comment a bit by moving a necessary clarifying bit from further up into the comment body.

addon/-record-data-private/system/model/internal-model.js
runspired: Sync relationships act as client-side delete. Previously, in this case the manyArray would be cleared by the relationship in this scenario, but with the refactoring for `RecordData` the manyArray is now managed by `InternalModel` and thus `InternalModel` must clear it in this case.

tests/integration/records/create-record-test.js
runspired: confirmed, and this behavior is true through 3.3 stable: https://ember-twiddle.com/f022afd32f5c9d6769e66007a0004497?openFiles=templates.application.hbs%2C

addon/-record-data-private/system/store/model-data-wrapper.js
```diff
@@ -65,13 +65,13 @@ export default class ModelDataWrapper {
     internalModel.notifyPropertyChange(key);
   }
 
-  notifyHasManyChange(modelName, id, clientId, key) {
+  notifyHasManyChanged(modelName, id, clientId, key) {
```
igorT: This is public api so we shouldn't change this

addon/-record-data-private/system/relationships/state/belongs-to.js
hjdivad: > However, while we are aware that <foo:1>.bar is no longer <bar:1>, we do not know with certainty that the canonical state of the relationship is now null and not some other bar.

What does this have to do with the second payload?  It is generally true that when the relationship of some other record changes this tells us nothing about any other relationship besides its inverse.

hjdivad: Put another way, once we sever the connection between `<foo:1>` and `<bar:1>` any other information about `<bar:1>.foo` tells us nothing about `<foo:1>.bar` unless the value is `<foo:1>`

Of course something might have changed on the server and our client state could be wrong, but that's always the case.  The two flags being set represent client state; in the case you describe client state is still correct, no?  Which flag do you think is not correct in this scenario?

tests/integration/records/create-record-test.js
hjdivad: ok seems fine although i presume we do not intend to make guarantees about the state of relationships on unloaded records and adding this here is solely a matter of pragmatism.

Issue #5525, opened 2018-07-13
3.2 sync relationships going reload after related parent creation


lifeart:
Given:

JSONAPI adapter/serializer.
Prev version 2.18
Works ok in 2.18.3

`shouldBackgroundReloadRecord() => false`

```
user {
  item: belongsTo('item', { async: false }),
  book: belongsTo('book', { async: false }),
  toy: belongsTo('toy', { async: false });
}
```

going to create user with already loaded relationships, but, after save, ember-data trying to reload each relationship

runspired:
@lifeart is this an issue in 3.1?

lifeart:
@runspired, nope, can't reproduce it in 3.1.1

runspired:
@lifeart I can't reproduce this in a test, could you give me more information about how you created the record, established the relationship, and what the response from the API looked like?

lifeart:
follow up from slack

model looks like 
```js

export default DS.Model.extend({
    transplantationCenter: belongsTo('transplantation-center', {
        async: false
    }),
    clinicalData: belongsTo('clinical-info', {
        inverse: 'patient',
        async: false
    }),
    gender: belongsTo('gender', { async: false }),
    document: belongsTo('document', { async: false }),
    country: belongsTo('meta-place', { async: false, polymorphic: true }),
    status: belongsTo('patient-status', { async: false }),
    city: belongsTo('meta-place', { async: false, polymorphic: true }),
    consiliumList: belongsTo('patient/consilium', { async: false }),
    consiliums: hasMany('consilium', { async: false, inverse: 'patient' }),
    users: hasMany('basic-user', { async: false, polymorphic: true }),
    pickedDoctors: alias('users'),
    refusalList: hasMany('refuse'),
    phonesList: hasMany('patient/phone', { async: false, inverse: 'patient' }),
});
```


Issue exists only for new models (isNew = true)

for example, i do `Patient.setProperties({city: CityModel(42) })`, and do Patient.save(), I get one more request to `/city/42`


`POST (http 201)`
```js
[...]

Issue #5524, opened 2018-07-11
UnloadRecord fails to cleanup hasMany relationships and causes computed properties to explode


Kilowhisky:
This is related to all the other UnloadRecord bugs floating around (including the ones i've opened). This one in particular is present in 3.2 and has to do with unloading a record that is being watched through a hasMany relationship by a computed property. 

What happens is that when records are unloaded the hasMany array contains null / undefined entries where the unloaded record previously resided. This causes anything watching the array to explode if it doesn't have null/undefined checking which happens to be all the computed properties ember offers.

First of all here is a twiddle showing the error:
https://ember-twiddle.com/5bee61b69a3813524b1d83b319c4fc30?openFiles=routes.application.js%2C

Here is the timeline for what is happening:

1) A main record is loaded (in my case this is an "asset"). It contains a hasMany relationship to a model called "position" and also contains a computed property that sorts the "positions".
2) The "positions" are loaded on demand and reference the "asset" that they belongTo. 
3) The UI shows the sorted positions via Asset > PositionsSorted
4) Some of the "positions" are then unloaded on the fly.
5) The hasMany still contains null / undefined references where the records used to be in the array. 
6) The computed property for sorting explodes because it can't sort undefined / null records.

```
"Error: Assertion Failed: Cannot call get with 'utc' on an undefined object.
    at new EmberError (https://cdn.jsdelivr.net/npm/ember-source@3.2.2/dist/ember.debug.js:14719:25)
    at Object.assert (https://cdn.jsdelivr.net/npm/ember-source@3.2.2/dist/ember.debug.js:14270:15)
    at get (https://cdn.jsdelivr.net/npm/ember-source@3.2.2/dist/ember.debug.js:26810:64)
    at https://cdn.jsdelivr.net/npm/ember-source@3.2.2/dist/ember.debug.js:37177:64
    at Array.sort (native)
    at sortByNormalizedSortProperties (https://cdn.jsdelivr.net/npm/ember-source@3.2.2/dist/ember.debug.js:37171:40)
    at Class._emberMetal.ComputedProperty.dependentKeys (https://cdn.jsdelivr.net/npm/ember-source@3.2.2/dist/ember.debug.js:37149:16)
    at ComputedProperty.get (https://cdn.jsdelivr.net/npm/ember-source@3.2.2/dist/ember.debug.js:27503:30)
    at get (https://cdn.jsdelivr.net/npm/ember-source@3.2.2/dist/ember.debug.js:26832:27)
[...]

PR #5537, opened 2018-07-23
[BUGFIX] fix timing of ui notification to avoid trolling observers

runspired:
While our `relationship.push` method is sync, we do not wait for the end of the method to `notifyPropertyChange`. This causes issues with observers which then trigger in a broken intermediate state. Here, I've fixed this for notifying that a relationship is stale after receiving a new `link`; however, it is still possible for this to go wrong for folks when new data is pushed in. In a happier future, we ought to have a better setup to ensure we always buffer ui notifications and flush after state is settled.

### DIFF
--- a/addon/-legacy-private/system/relationships/state/relationship.js
+++ b/addon/-legacy-private/system/relationships/state/relationship.js
@@ -522,7 +522,7 @@ export default class Relationship {
     this.store._updateRelationshipState(this);
   }
 
-  updateLink(link, initial) {
+  updateLink(link) {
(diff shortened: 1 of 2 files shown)


lifeart:
related issue: https://github.com/emberjs/data/issues/5525

runspired:
In adding test coverage for this I'll try to investigate #5524 as well.

Sidenote: incorrect missing `inverse:null` should have errored for this user and did not. Was not root-cause but did affect being able to properly debug.

PR #5544, opened 2018-07-26
remove test waiter

runspired:
The test waiter is prone to pwn-ing folks with the current implementation. New implementation to follow.

### DIFF
--- a/addon/-legacy-private/system/store.js
+++ b/addon/-legacy-private/system/store.js
@@ -226,10 +226,9 @@ Store = Service.extend({
 
     if (DEBUG) {
       this.__asyncRequestCount = 0;
+      this.__asyncWaiter = () => { return this.__asyncRequestCount === 0 };
 
-      Ember.Test.registerWaiter(() => {
-        return this.__asyncRequestCount === 0;
-      });
+      Ember.Test.registerWaiter(this.__asyncWaiter);
     }
   },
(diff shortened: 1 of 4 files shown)


addon/-legacy-private/system/store.js
```diff
@@ -225,6 +223,14 @@ Store = Service.extend({
 
     this._adapterCache = Object.create(null);
     this._serializerCache = Object.create(null);
+
+    if (DEBUG) {
+      this.__asyncRequestCount = 0;
+
+      Ember.Test.registerWaiter(() => {
```
stefanpenner: This requires a sibling `Ember.Test.unregisterWaiter(fn)` or it will leak

addon/-private/system/store/common.js
```diff
@@ -34,23 +33,12 @@ export function _objectIsAlive(object) {
...
-}
-
-if (DEBUG) {
-  Ember.Test.registerWaiter(() => {
-    return ASYNC_REQUEST_COUNT === 0;
-  });
-}
-
 export function guardDestroyedStore(promise, store, label) {
   let wrapperPromise = resolve(promise, label).then(v => promise);
```
stefanpenner: can we put the incrementer here? If so, that would likely bit-rot related tester waiter imbalances.

PR #5545, opened 2018-07-27
[FEAT] TrackableRequests for when async leakage is detected

runspired:
This feature enables app developers to better use `async ... await` while simultaneously detecting asynchronous test leaks in their data layer.  It is best paired with [ember-qunit's `setupTestIsolationValidation` feature](https://github.com/emberjs/ember-qunit/blob/master/addon-test-support/ember-qunit/index.js#L276).

When running an `ember-cli` application in non-production environments, this feature adds a test-waiter per-store instance that works with the [`settled()` test-helper](https://github.com/emberjs/ember-test-helpers/blob/master/API.md#settled). This test-waiter does not cause waiting by default, in order to prevent immediately breaking any apps that upgrade to this version of ember-data that do have test leaks without warning.

This feature comes with two feature flags, only configurable in `non-production` environments, that live on the `store` instance.  Both of these flags are currently `false` by default.

1. **`store.shouldTrackAsyncRequests`**

When `true`, requests initiated via the store will be assigned a token, and while any tokens have not resolved the test-waiter (and thus `await settled()`) will wait.

When `true`, any detected leaks will throw an error when the store is destroyed after a test, causing the test to fail. The leaks will be logged, with a descriptive `label` for the request that was initiated, and potentially a `trace` (see below) for where the request initiated from.

When `false`, any detected leaks will be emitted as warnings when the store is destroyed after a test.

2. **`store.generateStackTracesForTrackedRequests`**

When `true`, a stack trace will be captured at the point that a request is initiated and saved onto the token used to track the request. For requests other than `findRecord`, these traces will include the point in `app` code that caused this request to be initiated. For `findRecord` these traces currently point back to `flushPendingFetches`; however, we also capture a trace at the point that we schedule these fetches, and will soon thread those traces through so that where the request initiated in `app` code is clear.

### DIFF
--- a/addon/-legacy-private/system/store.js
+++ b/addon/-legacy-private/system/store.js
@@ -220,9 +220,52 @@ Store = Service.extend({
     this._serializerCache = Object.create(null);
 
     if (DEBUG) {
-      this.__asyncRequestCount = 0;
+      if (this.shouldTrackAsyncRequests === undefined) {
+        this.shouldTrackAsyncRequests = false;
+      }
+      if (this.generateStackTracesForTrackedRequests === undefined) {
+        this.generateStackTracesForTrackedRequests = false;
+      }
+
+      this._trackedAsyncRequests = [];
+      this._trackAsyncRequestStart = label => {
+        let trace =
+          'set `store.generateStackTracesForTrackedRequests = true;` to get a detailed trace for where this request originated';
+
+        if (this.generateStackTracesForTrackedRequests) {
+          try {
+            throw new Error(`EmberData TrackedRequest: ${label}`);
+          } catch (e) {
+            trace = e;
+          }
+        }
+
+        let token = Object.freeze({
+          label,
+          trace,
+        });
+
+        this._trackedAsyncRequests.push(token);
+        return token;
+      };
+      this._trackAsyncRequestEnd = token => {
+        let index = this._trackedAsyncRequests.indexOf(token);
+
+        if (index === -1) {
+          throw new Error(
+            `Attempted to end tracking for the following request but it was not being tracked:\n${token}`
+          );
+        }
+
+        this._trackedAsyncRequests.splice(index, 1);
+      };
+
       this.__asyncWaiter = () => {
-        return this.__asyncRequestCount === 0;
+        let shouldTrack = this.shouldTrackAsyncRequests;
+        let tracked = this._trackedAsyncRequests;
+        let isSettled = tracked.length === 0;
+
+        return shouldTrack !== true || isSettled;
       };
 
       Ember.Test.registerWaiter(this.__asyncWaiter);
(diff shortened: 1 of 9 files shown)


addon/-legacy-private/system/store.js
```diff
@@ -220,9 +220,48 @@ Store = Service.extend({
...
+
+        this._trackedAsyncRequests.push(token);
+        return token;
+      };
+      this._trackAsyncRequestEnd = token => {
+        let index = this._trackedAsyncRequests.indexOf(token);
+
+        if (index !== -1) {
+          this._trackedAsyncRequests.splice(index, 1);
+        }
```
stefanpenner: If their is no index, something seems wrong. We should explore/consider the impact of throwing.

addon/-legacy-private/system/store.js
```diff
@@ -2840,6 +2897,27 @@ Store = Service.extend({
...
+      let isSettled = tracked.length === 0;
+
+      if (!isSettled) {
+        if (shouldTrack) {
+          throw new Error(
+            'Async Request leaks detected. Add a breakpoint here and set `store._generateStackTracesForTrackedRequests = true;`to inspect traces for leak origins:\n\t - ' +
+              tracked.map(o => o.label).join('\n\t - ')
+          );
+        } else {
+          warn(
```
stefanpenner: by default, we may want to warn here. That way apps who violate this, have some grace period to cut over.

Erroring hard, make simply make it hard to upgrade with little upside. That being said, in the future we should strongly consider this being on by default.

addon/-legacy-private/system/store.js
```diff
@@ -833,6 +872,24 @@ Store = Service.extend({
...
+          throw new Error(`Trace Origin for scheduled fetch for ${modelName}:${id}.`);
+        } catch (e) {
+          trace = e;
+        }
+
+        // enable folks to discover the origin of this findRecord call when
+        // debugging. Ideally we would have a tracked queue for requests with
+        // labels or local IDs that could be used to merge this trace with
+        // the trace made available when we detect an async leak
+        pendingFetchItem.trace = trace;
```
stefanpenner: eventually we should thread this through, and it super supersede the `_trackAsyncRequestStart` trace

tests/unit/store/async-leak-test.js
```diff
@@ -0,0 +1,359 @@
...
+          return jsonApiPayload;
+        },
+      })
+    );
+    store = owner.lookup('service:store');
+  });
+
+  test('the waiter properly waits for pending requests', async function(assert) {
+    let stepResolve;
+    let stepPromise = new Promise(resolveStep => {
```
stefanpenner: `findRecordWasInvoked`

tests/unit/store/async-leak-test.js
```diff
@@ -0,0 +1,359 @@
...
+      waiter(),
+      true,
+      'We return true when no requests have been initiated yet (pending queue flush is async)'
+    );
+
+    await stepPromise;
+
+    assert.equal(waiter(), false, 'We return false to keep waiting while requests are pending');
+
+    await request.catch(e => {
```
stefanpenner: `assert.rejects` if present in this version of QUnit, could be used to provide better test failure messages.

more context: https://github.com/qunitjs/qunit/pull/1238/files#diff-78b865bceaf70054fdd9917848e7ddd8R43

PR #5551, opened 2018-07-31
Fix `floating dependencies` test suite

dcyriller:
With prettier v1.14 release, the way decorators should be formated has changed a bit. Let's update ember-data's Prettier version + apply the new fomatting.

As a result [this kind of failure](https://travis-ci.org/emberjs/data/jobs/410439133) won't happen anymore.

You may want to have a look to this [prettier release note](https://prettier.io/blog/2018/07/29/1.14.0.html#javascript).


bmac:
Thanks @dcyriller.

PR #5534, opened 2018-07-20
Added failing tests for belongsTo relationship

selvaa89:

PR #5554, opened 2018-08-03
[BUGFIX] Destroy proxy object in BelongsTo relation when no longer needed.

pieter-v:
The `_promiseProxy` object that is used by the belongsTo relationship is not destroyed when no longer needed.

This `_promiseProxy` keeps a reference to the internalModel. This especially a problem if the record is deleted while the other side of the relationship is still alive.

This results for us in a large memory leak: our application has a small set of long living records and during the lifetime it receives a large number of records (those records will be deleted after a while) that has a belongsTo reference to the records in the first set.

There is also a memory leak when proxies are destroyed when the proxied object is still alive. (reported as a separate issue in ember.js)

### DIFF
--- a/addon/-legacy-private/system/relationships/state/belongs-to.js
+++ b/addon/-legacy-private/system/relationships/state/belongs-to.js
@@ -110,7 +110,11 @@ export default class BelongsToRelationship extends Relationship {
     }
     if (this.inverseInternalModel !== this.canonicalState) {
       this.inverseInternalModel = this.canonicalState;
-      this._promiseProxy = null;
+      const proxy = this._promiseProxy;
+      if (proxy) {
+        proxy.destroy();
+        this._promiseProxy = null;
+      }
       this.notifyBelongsToChange();
     }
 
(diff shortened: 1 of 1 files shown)


runspired:
Investigated and pushed a fix, but I'm skeptical that the fix is safe to do without landing some of the other fixes and tests that are pending for async belongsTo relationships. I'm going to hold off on this one until those land.

runspired:
This PR will depend on

- #5562
- #5534 
- #5480 
- #5455 

#5534 is especially important, and coverage within it should be expanded to ensure that we test various scenarios in which the belongs-to state changes.

pieter-v:
@runspired Any chance, this gets merged?

runspired:
@pieter-v yes! would love to clean this up and see if it still helps us. Seems likely that we've already fixed this with some other recent work, but would be good to double confirm!

pieter-v:
@runspired Sorry for the delay
Indeed, the proxy leak seems to be fixed. Thanks
But there stills seems to be another leak. 
See #6271 for more info.

runspired:
Closing this as it was resolved by other PRs.

Issue #5556, opened 2018-08-06
@ember/ordered-set interop with ember-source@3.5


rwjblue:
When ember-data (emberjs/data#4e0bc45) is used with ember-source 3.5 (current canary  emberjs/ember.js#04dec4f9) the following error occurs:

![image 2](https://user-images.githubusercontent.com/12637/43720317-2a7e43b4-995e-11e8-8c95-f34fe4a7ceee.png)

Reported in slack by @NullVoxPopuli.

---

When ember-source hits 3.5 we no longer are leveraging `@ember/ordered-set` from Ember itself ([see here](https://github.com/emberjs/ember-ordered-set/blob/70d5f6eb55d4de912bf552e922eba4d9df00a47c/addon/index.js#L6)), and instead we are using the version in the addon.

The error occurs because we are mixing native classes (because the implementation in the addon does not get transpiled) and ES5 style classes.

PR #5557, opened 2018-08-07
Fixup transpilation issues with @ember/ordered-set.

rwjblue:
The first commit here, demonstrates the failure reported in #5556 by running the test suite in "Chrome only" mode.

Follow up commits clean things up, and ensure we test both transpiled maximally (down to IE11) and minimally.


runspired:
Thanks @rwjblue ! I've been removing those sorts of `run+set` situations for a while now, glad to have something that caught the rest and got us off of them :)

PR #5562, opened 2018-08-08
[BUGFIX] use internalModel promise if already loading

runspired:
Will need backported to 3.2/3.3 in addition to landing in a new 3.4 beta.

This resolves an issue that @wagenet and I debugged wherein async belongsTo relationships would not wait for `internalModels` that are loading. The eager resolve would then leave the relationship in a state that appeared to be empty.

[WIP] Working on adding a test still.


jrjohnson:
Great fix! Thanks. Is it difficult to backport this? It clears up some issues preventing us from getting to 3.2 or 3.3. I can easily wait for 3.4, but just wanted to mention it.

PR #5564, opened 2018-08-13
Updated broken link to Ember guides

kennethlarsen:
It seems that the README contained a broken link to the Ember guides.


bmac:
Thanks @kennethlarsen.

PR #5567, opened 2018-08-14
[FEAT] enable external partner testing

runspired:
Creates a script that enables us to run the tests of external apps and addons against commits in ember-data.


lib/scripts/test-external.js
```diff
@@ -0,0 +1,67 @@
...
+
+const fs = require('fs');
+const path = require('path');
+const { execSync } = require('child_process');
+const debug = require('debug')('test-external');
+
+const projectRoot = path.resolve(__dirname, '../../');
+const externalProjectName = process.argv[2];
+const gitUrl = process.argv[3];
+const tempDir = path.join(projectRoot, '../__external-test-cache');
```
runspired: I intend to set this up such that we can cache the repositories we test on travis and use `git pull` to get latest most of the time vs needing to blow away and start over all the time.

lib/scripts/test-external.js
```diff
@@ -0,0 +1,67 @@
...
+try {
+  execSync(`${useYarn ? 'yarn link' : 'npm link'} && cd ../__external-test-cache/${externalProjectName} && ${useYarn ? 'yarn link ember-data && yarn' : 'npm link ember-data && npm install'}`);
+} catch (e) {
+  debug(e);
+  throw new Error(`Unable to complete install of dependencies for external project ${externalProjectName}`);
+}
+
+// run project tests
+console.log(`Running tests for ${externalProjectName}`);
+execWithLog(`cd ../__external-test-cache/${externalProjectName} && ember test`, true);
```
runspired: I'm unsure if an unsuccessful test run would cause us to throw an error and fail, likely need to do more here.

runspired:
cc @backspace :)

kategengler:
Ah, this may be painful/brittle if the external apps fail for reasons unrelated to ember-data. My approach was going to be to add ember-try config to the various apps to test against canary ember-data + create a dashboard with the status of the various CI results. That doesn't get PR testing, though.

runspired:
@kategengler that would still be immensely useful; however, due to certain constraints around 3.4 LTS and internal branching for `RecordData`, @hjdivad and I decided it would be immensely useful to have commit-over-commit testing that would allow us to prevent release if needed.

Obviously we would need to check if the tests failed due to our changes or those of the external app. One thought I have for reducing false failures is to run the external apps tests twice, once against it's own desired version and once against the new commit, and only fail if the first passes but the second doesn't.

lib/scripts/test-external.js
kategengler: Alternatively, you could shallow clone for some speed improvement and not have to worry about caching mechanisms.

runspired: @kategengler how does that work?

kategengler: `git clone --depth=1 <repo>` will get you just the HEAD of the repo, so you don't get all history.

backspace:
> run the external apps tests twice

This seems like a good plan. Maybe I’ll finally figure our the intermittently-failing race-condition-y test 😳 

Nice use of stages for this!

I recently updated `travis-web` to use Node 10, hopefully that‘s not eventually a problem with this using 6. The only thing I tend to have problems with across Node versions is `ember-cli-sass`.

.travis.yml
```diff
@@ -70,6 +72,21 @@ jobs:
...
+      script: yarn test-external:model-fragments
+    - env: NAME=ember-observer
+      script: yarn test-external:ember-observer
+    - env: NAME=travis-web
+      script: yarn test-external:travis-web
+    - env: NAME=emberaddons.com
+      script: yarn test-external:emberaddons.com
+    - env: NAME=ember-data-change-tracker
+      script: yarn test-external:ember-data-change-tracker
+    - env: NAME=ember-m3
```
backspace: re all these `NAME=` workarounds, you can now explicitly name jobs using `name: "Whatever"`, as with the “Max Transpilation” one earlier in the file.

runspired: 🗡 🔥 🔋

lib/scripts/test-external.js
```diff
@@ -0,0 +1,118 @@
+'use strict';
+/* eslint-disable no-console, node/no-extraneous-require, node/no-unpublished-require */
+const fs = require('fs');
+const path = require('path');
+const { execSync } = require('child_process');
```
stefanpenner: lets use `execa` as it cleans up leaked children, if the script is exited early.

runspired: @stefanpenner this? https://github.com/sindresorhus/execa

.travis.yml
```diff
@@ -54,24 +54,46 @@ jobs:
...
+    - name: "Ember Release"
+      env: EMBER_TRY_SCENARIO=ember-release
+    - name: "Ember Beta"
+      env: EMBER_TRY_SCENARIO=ember-beta
+    - name: "Ember Canary"
+      env: EMBER_TRY_SCENARIO=ember-canary
+
+    # runs tests against various open-source projects for early-warning regression analysis
+    - stage: external partner tests
+      name: "Model Fragments"
```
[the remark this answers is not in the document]
runspired: added

PR #5571, opened 2018-08-17
Add failing test when changing parent and unloading original parent

lsg-braymon:
Bug wherein changing the parent of a child then unloading the original parent sets the child association to `null`.
Setup:
```js
  // This is the association set up
  class Child extends DS.Model {
    @belongsTo('parent')
  }

  class Parent extends DS.Model {
    @hasMany('child')
  }
```

```js
  // Assume here that `child` is already associated to Parent 1
  const parent1 = this.store.peekRecord('parent', 1)
  const child = this.store.peekRecord('child', 1)

  const parent2 = this.store.peekRecord('parent', 2)
  set(child, 'parent', parent2) // We are not commiting to the database yet at this point

  // Delete Parent 1
  parent1.unloadRecord()
  child.parent // Returns null. Parent 2 is expected
```
Current work around:
```js
// Assume here that `child` is already associated to Parent 1
  const parent1 = this.store.peekRecord('parent', 1)
  const child = this.store.peekRecord('child', 1)

  const parent2 = this.store.peekRecord('parent', 2)
  const payload = {
    data: {
      type: 'child',
      id: child.id,
      relationships: {
        parent: {
          data: { type: 'parent', id: parent2.id }
        }
      }
    }
  }
  this.store.pushPayload(payload) // Instead of set(child, 'parent', parent2)

  // Delete Parent 1
  parent1.unloadRecord()
  child.parent // Returns Parent 2
```

### DIFF
--- a/packages/store/addon/-private/system/relationships/state/relationship.ts
+++ b/packages/store/addon/-private/system/relationships/state/relationship.ts
@@ -214,7 +214,8 @@ export default class Relationship {
   }
 
   recordDataDidDematerialize() {
-    if (!this.inverseKey) {
+    const inverseKey = this.inverseKey
+    if (!inverseKey) {
       return;
     }
     // TODO @runspired fairly sure we need to become stale here
@@ -226,8 +227,18 @@ export default class Relationship {
       if (!this._hasSupportForRelationships(inverseRecordData)) {
         return;
       }
-      let relationship = relationshipStateFor(inverseRecordData, this.inverseKey);
-      relationship.inverseDidDematerialize(this.recordData);
+      let relationship = relationshipStateFor(inverseRecordData, inverseKey);
+      let belongsToRelationship = inverseRecordData.getBelongsTo(inverseKey)._relationship;
+
+      // For canonical members, it is possible that inverseRecordData has already been associated to
+      // to another record. For such cases, do not dematerialize the inverseRecordData
+      if (
+        !belongsToRelationship ||
+        !belongsToRelationship.inverseRecordData ||
+        this.recordData === belongsToRelationship.inverseRecordData
+      ) {
+        relationship.inverseDidDematerialize(this.recordData);
+      }
     });
   }
 
--- a/packages/-ember-data/tests/integration/records/edit-record-test.js
+++ b/packages/-ember-data/tests/integration/records/edit-record-test.js
@@ -195,6 +195,89 @@ module('Editing a Record', function(hooks) {
           .map(pet => pet.get('name'));
         assert.deepEqual(pets, ['Shen'], 'Chris has Shen as a pet');
       });
+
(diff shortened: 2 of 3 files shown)


runspired:
At first glance this looks good, will want to poke around at it locally before merging to gain more confidence that this is the correct fix.

lsg-braymon:
@runspired Got it. Thank you.

lsg-braymon:
Hi @runspired just wanted to ask if there are updates on this bug. Thank you!

runspired:
@lsg-braymon sadly I didn't spend the time to review this yet, thanks for the reminder! Will attach some labels so I don't forget :P

richard-viney:
I work with @lsg-braymon and have just rebased this PR against the latest master and added some more detail into the commit message.

Is there anything we can do to help with getting this fix included as part of an upcoming release?

Thanks!

packages/store/addon/-private/system/relationships/state/relationship.ts
```diff
@@ -227,7 +227,14 @@ export default class Relationship {
         return;
       }
       let relationship = relationshipStateFor(inverseRecordData, this.inverseKey);
-      relationship.inverseDidDematerialize(this.recordData);
+      let inverseRecordDataRelationship = inverseRecordData.getBelongsTo(this.inverseKey!)
+        ._relationship!.inverseRecordData;
```
runspired: What are the ! Characters here for?

richard-viney: The non-null assertion on `this.inverseKey` was because the TypeScript compiler hasn't propagated the null check on it from the start of the function. I've refactored the code slightly to avoid the need for the `!` here.

Have also added a check that the `_relationship` member here is actually set, in place of the non-null assertion.

runspired: Thanks!

richard-viney:
Thanks heaps, is this change a candidate for inclusion in the current 3.10 betas?

runspired:
@richard-coney yes, but I haven’t gotten publishing to work correctly with the monorepo just yet. Need to tweak the tsconfig some more so that prepublish works. If you have a moment to help on that this weekend I’d love it cause I don’t have much time ❤️

TL;DR npm prebublish command needs to work correctly in each individual package and is broken in a few atm.

richard-viney:
Sure thing, I'll take a look later today

runspired:
@richard-viney the beta branch has some work towards it to bring back to master. Trying to release last night was an adventure 😂

runspired:
@richard-viney ended up with a little time to poke at it and think I resolved the issues :) Thank you for the offer, if you do have some time and want to contribute I left some notes in the fix-pr of some things we need to still do to the prepublish infra prior to our next stable release.

https://github.com/emberjs/data/pull/6050

PR #5572, opened 2018-08-17
Failing test for unloadAll scenario with hasMany relationship that ends up wrong

lukemelia:
This is a simplified failing test for an issue we experienced in our app.

The test passes when an optimization is commented out, specifically the conditional here: https://github.com/emberjs/data/blob/ece04949ecfd84c491490dd9a110e02d2ff0afa4/addon/-legacy-private/system/relationships/state/relationship.js#L304

We're happy to work on a fix but are not sure where to go from here.

- @lukemelia & @jasonbekolay

### DIFF
--- a/tests/integration/records/rematerialize-test.js
+++ b/tests/integration/records/rematerialize-test.js
@@ -49,6 +49,18 @@ Boat.toString = function() {
   return 'Boat';
 };
 
+let Listable = Model.extend({
+  listItems: hasMany('list-item', { inverse: 'listable' }),
+});
+let List = Model.extend({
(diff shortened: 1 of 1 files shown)


runspired:
@lukemelia thanks! I'll investigate what the next step is

lukemelia:
Cool @runspired. FWIW, we decided to temporarily run with a fork that has the optimization line I referenced commented out.

runspired:
@lukemelia any chance you got stuck here at some point? I doubt this is still a bug but if it is happy to get the test caught up and a fix investigated. Because of the way the relationship layer and unload works today this test probably isn't a meaningful addition anymore unless there's still a known issue it manages to catch

Issue #5578, opened 2018-08-21
Backport #5467 (work during destroy) to 3.1/2.18


runspired:
The fix in #5467 helps to prevent issues with `ember.js` version `3.3`, it is likely we should backport it at least that far.

jherdman:
I'm going to take this on.

runspired:
Turns out this was backported to 3.2/3.3 already. We should backport to 3.1/2.18 to catch more folks that haven't upgraded and are still re-reporting the issue.

PR #5585, opened 2018-08-22
[BUGFIX Backport to 3.1] dont cause unnecessary work during destroy

jherdman:
Per #5578 we're backporting #5467 to 3.1.


jherdman:
@runspired how much do we care about the older version fails here?

runspired:
@jherdman sometimes we don't care, but the test that is failing is unexpected. I will investigate locally and determine if something more is needed, thanks!

PR #5602, opened 2018-08-28
Publish docs to npm

sivakumar-kailasam:
This is required to be on par with ember-source so that api docs can rely on using npm for fetching the yuidocs for >= 3.4.0 releases.


sivakumar-kailasam:
@rwjblue @runspired  looks like 3.4.0 docs didn't get published to s3. Can we merge this & get a patch version?

.npmignore
```diff
@@ -1,13 +1,17 @@
 /bower_components
 /config/ember-try.js
-/dist
 /node-tests
 /tests
 /tmp
 /bin
-/docs
```
runspired: is this still required if docs are now in `dist/` ?

runspired:
So for what it does, this seems ok; however, I'm a bit confused as to why we now publish our docs on npm in this way (and why the emberjs website requires it). Does the S3 publishing no longer accomplish what the learning team needs? If so we should remove the S3 publishing.

sivakumar-kailasam:
> why we now publish our docs on npm in this way 
- robert mentioned during 3.3 release that we should use npm as the source for the docs instead of s3
- we've had other issues too over time when we had to look around for folks with s3 access & it seemed like publishing patch releases with any doc fixes was a better way to go about it

> why the emberjs website requires it
This is for building the API docs' search index.

>  If so we should remove the S3 publishing.
I've modified the PR to remove the scripts for publishing yuidoc to s3 & to publish to npm instead

PR #5616, opened 2018-08-30
[FEAT RecordData] turn on the build-flag by default

runspired:
Lands https://github.com/emberjs/rfcs/pull/293 in the beta channel. To be released as `3.5.0-beta.2`

Note: We will land this and release  `35.0-beta.2` once we cut a new `3.4` from master to pick up that last few fixes that we want in that release.


NullVoxPopuli:
is there anything blocking this?

PR #5617, opened 2018-08-30
[FEAT] IE11 tests via SauceLabs

runspired:
- We've been approved for an OSS account but await upgrade to access keys.
- The SauceLabs logo link in our ReadMe is a requirement of the free OSS account


runspired:
Going to look into using browserstack instead as that is what is used for ember.js

PR #5620, opened 2018-09-01
[DOC release] Change misleading documentation for AdapterPopulatedRecordArray

happycollision:
In my experimentation, I've discovered that though the promise returned from `store.query()` does have array-like methods on it, the actual `AdapterPopulatedRecordArray` is what the promise *resolves to*. The promise does not behave like a proxy (or what I think Ember Proxies do...). The documentation for `store.query` states this plainly, but the documentation here misrepresented that.

For example, the original code actually would behave like this:
```javascript
  // GET /users?isAdmin=true
  var admins = store.query('user', { isAdmin: true });

  admins.then(function() {
    console.log(admins.get("length")); // Throws: "admins.get" is not a function
  });
```

### DIFF
--- a/addon/-private/system/record-arrays/adapter-populated-record-array.js
+++ b/addon/-private/system/record-arrays/adapter-populated-record-array.js
@@ -19,22 +19,23 @@ import cloneNull from '../clone-null';
 
   ```javascript
   // GET /users?isAdmin=true
-  var admins = store.query('user', { isAdmin: true });
+  store.query('user', { isAdmin: true }).then(function(admins) {
 
-  admins.then(function() {
-    console.log(admins.get("length")); // 42
-  });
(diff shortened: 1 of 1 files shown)


happycollision:
And once more to remove things inside the commit message where I am speaking a little bit about things I don't know.

I realize that when I read commit messages that are merged, I take them to be the actual truth. In my original commit message, I made a comment referring to Ember's ProxyObjects, but I am really not 100% sure how those work. And that comment is actually not pertinent to this change anyhow.

runspired:
I'm in favor of this documentation change; however, historically speaking `query` did return a `PromiseArray`. we may have broken something over time without realizing it. In today's world of `async await` and better promise knowledge, just working with the promise is preferred anyway.

PR #5622, opened 2018-09-04
remove deletes

kturney:
I was mucking about with google-closure-compiler and it tripped up on these deletes that are not for properties. Looking at the blame, looks like they may have been left over from some find/replacing.


bmac:
Thanks @kturney.

runspired:
@kturney I noticed that before and meant to go fix and forgot :( Thanks for fixing :)

Issue #3700, opened 2015-08-25
Support query params when fetching hasMany relationship


SeyZ:
I use the JSONAPI adapter/serializer. I would like to add queryParams to my url when fetching a hasMany relationships. Example:

When I do something like that `gallery.get('illustrations', { sort: 'name' })`, it triggers this (links related href works well):

GET `/api/galleries/34/relationships/illustrations`

But I would like to have:

GET `/api/galleries/34/relationships/illustrations?sort=name`

How can I add the sort queryParam to the url ?

arenoir:
As it stands now links have to come from the back end (or use a callback in the serializer). I have a [pull-request](https://github.com/emberjs/data/pull/2435) which implements a `buildUrlForHasMany` function that could handle your scenario.  It is a bit out of date but you may find it useful.

It has always irked me that I had to include a link or an array of ids to do a hasMany relationship when all the information is present in the parent to do the lookup.

Finhas:
Have you tried:

this.store.query('illustrations', { sort: 'name' });

http://guides.emberjs.com/v2.0.0/models/finding-records/#toc_querying-for-multiple-records

SeyZ:
@Finhas If I do what you describe, the url won't be `/api/galleries/34/relationships/illustrations?sort=name` but something like `/api/illustrations?sort=name`.

My request needs to consider the (links related href value)[http://jsonapi.org/format/#document-links]

Finhas:
@SeyZ Since you're requesting the relationship data in your controller, couldn't you sort the return in ember before it's displayed or have default sorting set on your server?

The server MAY apply default sorting rules to top-level data if request parameter sort is not specified.

Honestly, I was thinking your request on the router would return `/api/galleries/34?includes=illustrations' but I guess it depends on how many 'illustration' categories you need to fetch from the gallery.

mdehoog:
I have created an addon for this that may be useful: https://github.com/mdehoog/ember-data-has-many-query. After configuring, simply:

`gallery.query('illustrations', { sort: 'name' });`

SeyZ:
@mdehoog You rock! Thank you very much!

Exelord:
Please Ember Data's authors, implement it this way :D  `gallery.get('illustrations', { sort: 'name' })`
This is so intuitive!

SeyZ:
One year later, is it something you're working on? It would be awesome guys :)

gnapse:
[...]

PR #5627, opened 2018-09-10
[FEAT adapterOptions] Ensure adapterOptions use is possible throughout

nathanhammond:
Addressing this comment: https://github.com/emberjs/data/issues/3700#issuecomment-379246701

- [x] Tests (@runspired asked me to split those from the actual code to ease possible backporting.)
- [x] Docs

### DIFF
--- a/addon/-legacy-private/system/references/belongs-to.js
+++ b/addon/-legacy-private/system/references/belongs-to.js
@@ -346,13 +346,35 @@ export default class BelongsToReference extends Reference {
       });
       ```
 
+      You may also pass in an options object whose properties will be
+      fed forward. This enables you to pass `adapterOptions` into a
+      a reference.
+
+      Example
+
+      ```javascript
+      userRef.load({ adapterOptions: { isPrivate: true } }).then(function(user) {
+        userRef.value() === user;
+      });
+      ```
+
+      ```app/adapters/user.js
+      export default ApplicationAdapter.extend({
+        findRecord(store, type, id, snapshot) {
+          // In the adapter you will have access to adapterOptions.
+          let adapterOptions = snapshot.adapterOptions;
+        }
+      });
+      ```
+
      @method load
+     @param {Object} options the options to pass in.
      @return {Promise} a promise that resolves with the record in this belongs-to relationship.
   */
-  load() {
+  load(options) {
     let rel = this.belongsToRelationship;
 
-    rel.getData();
+    rel.getData(options);
 
     if (rel.fetchPromise !== null) {
       return rel.fetchPromise.then(() => {
@@ -395,11 +417,22 @@ export default class BelongsToReference extends Reference {
       });
       ```
 
+      You may also pass in an options object whose properties will be
+      fed forward. This enables you to pass `adapterOptions` into a
+      a reference. A full example can be found in the `load` method.
+
+      Example
+
+      ```javascript
+      userRef.reload({ adapterOptions: { isPrivate: true } })
+      ```
+
      @method reload
+     @param {Object} options the options to pass in.
      @return {Promise} a promise that resolves with the record in this belongs-to relationship after the reload has completed.
   */
-  reload() {
-    return this.belongsToRelationship.reload().then(internalModel => {
+  reload(options) {
+    return this.belongsToRelationship.reload(options).then(internalModel => {
       return this.value();
     });
   }
--- a/addon/-legacy-private/system/references/belongs-to.js
+++ b/addon/-legacy-private/system/references/belongs-to.js
@@ -347,8 +347,8 @@ export default class BelongsToReference extends Reference {
       ```
 
       You may also pass in an options object whose properties will be
-      fed forward. This enables you to pass `adapterOptions` into a
-      a reference.
+      fed forward. This enables you to pass `adapterOptions` into the
+      request given to the adapter via the reference.
 
       Example
 
@@ -418,8 +418,9 @@ export default class BelongsToReference extends Reference {
       ```
 
       You may also pass in an options object whose properties will be
-      fed forward. This enables you to pass `adapterOptions` into a
-      a reference. A full example can be found in the `load` method.
+      fed forward. This enables you to pass `adapterOptions` into the
+      request given to the adapter via the reference. A full example
+      can be found in the `load` method.
 
       Example
 
--- a/addon/-legacy-private/system/references/has-many.js
+++ b/addon/-legacy-private/system/references/has-many.js
@@ -355,8 +355,8 @@ export default class HasManyReference extends Reference {
      ```
 
      You may also pass in an options object whose properties will be
-     fed forward. This enables you to pass `adapterOptions` into a
-     a reference.
+     fed forward. This enables you to pass `adapterOptions` into the
+     request given to the adapter via the reference.
 
      Example
 
@@ -422,8 +422,9 @@ export default class HasManyReference extends Reference {
      ```
 
      You may also pass in an options object whose properties will be
-     fed forward. This enables you to pass `adapterOptions` into a
-     a reference. A full example can be found in the `load` method.
+     fed forward. This enables you to pass `adapterOptions` into the
+     request given to the adapter via the reference. A full example
+     can be found in the `load` method.
 
      Example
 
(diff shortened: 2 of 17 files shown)


nathanhammond:
An external demo of this appearing to work can be found here:
https://github.com/nathanhammond/adapter-options/commit/8decfd78c986d898f3b0d9a3a554206771bbed21

nathanhammond:
I've added a new optional commit that would enable this pattern:

```js
import Model from 'ember-data/model';
import attr from 'ember-data/attr';
import { hasMany } from 'ember-data/relationships';

export default Model.extend({
  name: attr(),
  posts: hasMany('posts', { adapterOptions: { ... } })
});
```

This is also implementable in user-space if the user isn't afraid of tweaking the `computed` returned from `hasMany` or `belongsTo`.

It's an ergonomics-only change, but significantly exposes the plumbing of the internals.

runspired:
> This is also implementable in user-space if the user isn't afraid of tweaking the computed returned from hasMany or belongsTo.

The entire original `meta` the user defines on the model is accessible via accessing the `meta` via `relationshipsByName`.

addon/-legacy-private/system/relationships/state/belongs-to.js
```diff
@@ -231,17 +231,17 @@ export default class BelongsToRelationship extends Relationship {
     Other calls must conform to the typical expectations, for instance, sync relationships
     expect that their data is already loaded.
    */
-  getData(isForcedReload = false) {
+  getData(options = {}) {
```
runspired: let's leave `isForcedReload` as the second argument here, as it should not be leaked to end users.

addon/-legacy-private/system/relationships/state/belongs-to.js
```diff
@@ -270,7 +270,7 @@ export default class BelongsToRelationship extends Relationship {
           "' with id " +
           this.internalModel.id +
           ' but some of the associated records were not loaded. Either make sure they are all loaded together with the parent record, or specify that the relationship is async (`DS.belongsTo({ async: true })`)',
-        record === null || !record.get('isEmpty') || isForcedReload
+        record === null || !record.get('isEmpty') || options.reload === true
```
runspired: let's keep `isForcedReload`

addon/-legacy-private/system/relationships/state/has-many.js
```diff
@@ -344,17 +344,17 @@ export default class ManyRelationship extends Relationship {
       });
   }
 
-  getData(isForcedReload = false) {
+  getData(options = {}) {
```
runspired: same as for `belongsTo`, make `isForcedReload` the second arg.

addon/-legacy-private/system/relationships/state/has-many.js
```diff
@@ -379,7 +379,7 @@ export default class ManyRelationship extends Relationship {
         }' with id ${
           this.internalModel.id
         } but some of the associated records were not loaded. Either make sure they are all loaded together with the parent record, or specify that the relationship is async (\`DS.hasMany({ async: true })\`)`,
-        this.allInverseRecordsAreLoaded || isForcedReload
+        this.allInverseRecordsAreLoaded || options.reload === true
```
runspired: same as `belongsTo`, let's keep `isForcedReload` here

addon/-legacy-private/system/relationships/state/relationship.js
```diff
@@ -554,7 +554,8 @@ export default class Relationship {
 
     this.setHasFailedLoadAttempt(false);
     this.setShouldForceReload(true);
-    this.getData(true);
+    options.reload = true;
+    this.getData(options);
```
runspired: let's make this `this.getData(options, true)`. We don't want to leak the ability to add `reload` as an option to calls to `load`, and the reason for this flag is to enforce that only this specific call to `getData` is considered a `reload`.

addon/-record-data-private/system/references/belongs-to.js
```diff
@@ -291,27 +291,29 @@ export default class BelongsToReference extends Reference {
    @return {Promise} a promise that resolves with the record in this belongs-to relationship after the reload has completed.
    */
   // TODO IGOR CHECK FOR OBJECT PROXIES
-  reload() {
+  reload(options = {}) {
+    options.reload = true;
```
runspired: we can delete this

addon/-record-data-private/system/references/has-many.js
```diff
@@ -328,7 +328,8 @@ export default class HasManyReference extends Reference {
    @method reload
    @return {Promise} a promise that resolves with the ManyArray in this has-many relationship.
    */
-  reload() {
-    return this.internalModel.reloadHasMany(this.key);
+  reload(options = {}) {
+    options.reload = true;
```
runspired: we can delete this assignment

addon/-record-data-private/system/store.js
```diff
@@ -49,6 +50,8 @@ const badIdFormatAssertion = '`id` passed to `findRecord()` has to be non-empty
 const emberRun = emberRunLoop.backburner;
 
 const { ENV } = Ember;
+const WeakMapImplementation = WeakMap || Ember.WeakMap;
```
runspired: `Ember.WeakMap` already uses the global if available, and this would throw an error if it were not available.

addon/-legacy-private/system/references/belongs-to.js
```diff
@@ -398,8 +398,9 @@ export default class BelongsToReference extends Reference {
      @method reload
      @return {Promise} a promise that resolves with the record in this belongs-to relationship after the reload has completed.
   */
-  reload() {
-    return this.belongsToRelationship.reload().then(internalModel => {
+  reload(options = {}) {
+    options.reload = true;
```
runspired: we can delete this assignment

runspired: We can also delete the `options = {}` assignment since we no longer care about ensuring options is defined.

addon/-legacy-private/system/references/has-many.js
```diff
@@ -358,10 +358,10 @@ export default class HasManyReference extends Reference {
      @return {Promise} a promise that resolves with the ManyArray in
      this has-many relationship.
   */
-  load() {
+  load(options = {}) {
```
runspired: we can delete the default value here

addon/-legacy-private/system/references/has-many.js
```diff
@@ -401,7 +401,8 @@ export default class HasManyReference extends Reference {
      @method reload
      @return {Promise} a promise that resolves with the ManyArray in this has-many relationship.
   */
-  reload() {
-    return this.hasManyRelationship.reload();
+  reload(options = {}) {
+    options.reload = true;
```
runspired: we can delete the default value assignment and the `options.reload = true` assignment

addon/-legacy-private/system/relationships/state/has-many.js
runspired: also, `options` does not need to have a default value

addon/-legacy-private/system/relationships/state/belongs-to.js
runspired: options does not need a defaultValue

addon/-legacy-private/system/relationships/state/relationship.js
```diff
@@ -545,7 +545,7 @@ export default class Relationship {
     this.setRelationshipIsStale(true);
   }
 
-  reload() {
+  reload(options = {}) {
```
runspired: options does not need a default value here

addon/-legacy-private/system/store.js
```diff
@@ -48,6 +49,8 @@ const badIdFormatAssertion = '`id` passed to `findRecord()` has to be non-empty
 const emberRun = emberRunLoop.backburner;
 const { ENV } = Ember;
 
+const WeakMapImplementation = WeakMap || Ember.WeakMap;
```
runspired: just use `Ember.WeakMap`. `WeakMap` would error here if undefined and the Ember polyfill uses native if available.

addon/-legacy-private/system/store.js
```diff
@@ -1,3 +1,4 @@
+/* global WeakMap */
```
runspired: delete once no longer needed (see below)

addon/-record-data-private/system/store.js
```diff
@@ -1,3 +1,4 @@
+/* global WeakMap */
```
runspired: delete once no longer needed (see below)

PR #5640, opened 2018-09-15
[BUGFIX] fix deprecation for store entry methods

runspired:


Gaurav0:
@runspired Is it possible to backport this to Ember Data 3.4 LTS?

runspired:
@Gaurav0 this already was.

ELepolt:
Sorry, I came across this thread trying to track this down. I might not be following correctly, but isn't this the line that is currently in 3.4 LTS: https://github.com/emberjs/data/blob/lts-3-4/addon/-private/system/store.js#L3428

Gaurav0:
I'm still seeing the deprecation warning in 3.4.4

runspired:
@Gaurav0 you should see this deprecation in 3.4.4 if you are hitting it for correct reasons, this fix was to ensure that folks didn't hit it for incorrect reasons.

Gaurav0:
@runspired I am hitting it during teardown in application tests and in rendering tests in which data is added to the store.

PR #5642, opened 2018-09-18
Add a couple of ED-related add-ons as external partner tests

balinterdi:
This PR adds ember-resource-metadata and ember-data-relationship-tracker as external tests.

The former calls `_internalModel` on DS.Model record instances:
https://github.com/ef4/ember-resource-metadata/blob/master/addon/private-api.js#L16-L18

, while the latter relies on having the `relationshipsByName` available on model classes:
https://github.com/ef4/ember-data-relationship-tracker/blob/master/addon/index.js#L66


runspired:
Thanks @balinterdi !

PR #5644, opened 2018-09-19
[BUGFIX] ensure external-partner tests correctly symlink

runspired:
Turns out we weren't symlink-ing from within the external project directory and so we haven't been testing external partners correctly. We have work to do.


runspired:
I'm going to merge this even though the following addons and apps are red.

```
- ember-data-factory-guy
- ember-data-model-fragments
- ember-data-change-tracker
- ilios
```

The above addons/apps should have been failing for `record-data` all along, and we need to investigate why.

runspired:
cc @danielspaniel @jrjohnson @jakesjews

PR #5647, opened 2018-09-20
WIP: Typescript Support!

NullVoxPopuli:
closes: https://github.com/emberjs/data/pull/5473

Thanks to ember-cli-typescript :)

 - also merges private directories, eliminating the `-legacy-private` directory

Blocked: waiting on babel 7 support from the ember-cli-typescript project


NullVoxPopuli:
currently: `yarn test` never completes. 
```
cleaning up...
Built project successfully. Stored in "/home/lprestonsegoiii/Development/NullVoxPopuli/ember-data/tmp/class-tests_dist-0CYMhEok.tmp".

  FF^C
$ F

  3 tests complete (667460 ms)

  3) [null] Error
     Received SIGINT signal


Received SIGINT signal

```

ember-cli-build.js
```diff
@@ -5,7 +5,7 @@ const EmberAddon = require('ember-cli/lib/broccoli/ember-addon');
 module.exports = function(defaults) {
   let app = new EmberAddon(defaults, {
     'ember-cli-babel': {
-      throwUnlessParallelizable: true,
```
runspired: is ember-cli-typescript preventing this? 😱

runspired: cc @chriskrycho I'd recommend setting up ember-cli-typescript's tests to ensure that we don't need to disable this prior to stable release.

types/dummy/index.d.ts
```diff
@@ -0,0 +1 @@
+
```
runspired: is this file supposed to be empty?

NullVoxPopuli: it's added by default, yeah

ember-cli-build.js
NullVoxPopuli: unsure, I need to re-rebase

addon/transforms/string.ts
```diff
@@ -24,10 +24,10 @@ import Transform from './transform';
   @namespace DS
  */
 export default Transform.extend({
-  deserialize(serialized) {
+  deserialize(serialized: any) {
```
mike-north: Are `null`, `undefined` and non-serializable objects (i.e., circular references) valid things to pass here?

package.json
```diff
@@ -61,6 +63,13 @@
   "devDependencies": {
     "@ember-decorators/babel-transforms": "^2.1.2",
     "@ember-decorators/data": "^3.0.0",
+    "@types/ember": "^3.0.25",
+    "@types/ember-qunit": "^3.4.3",
+    "@types/ember-test-helpers": "^1.0.4",
+    "@types/ember-testing-helpers": "^0.0.3",
+    "@types/ember__test-helpers": "^0.7.6",
```
mike-north: all of the `@types/ember*` should be `~` dependencies, not `^`. Read more about the versioning strategy here https://github.com/mike-north/typever

package.json
```diff
@@ -83,6 +92,8 @@
     "ember-cli-shims": "^1.2.0",
     "ember-cli-sri": "^2.1.1",
     "ember-cli-test-loader": "^2.2.0",
+    "ember-cli-typescript": "^2.0.0-beta.1",
```
mike-north: We are very serious about this being a pre-release version. Please assume that it could break at any time, and take on acceptable risk accordingly

package.json
```diff
@@ -108,7 +119,8 @@
     "mocha-only-detector": "1.0.0",
     "prettier": "^1.14.3",
     "rimraf": "^2.6.2",
-    "rsvp": "^4.8.4"
+    "rsvp": "^4.8.4",
+    "typescript": "^3.1.3"
```
mike-north: TypeScript does not follow SemVer, and [includes breaking changes](https://github.com/Microsoft/TypeScript/wiki/API-Breaking-Changes) in just about every middle-number release. You'll want this to be a `~` target as well, to avoid unknowningly taking in breaking changes.

tsconfig.json
```diff
@@ -0,0 +1,55 @@
+{
+  "compilerOptions": {
+    "target": "es2017",
+    "allowJs": true,
+    "moduleResolution": "node",
+    "allowSyntheticDefaultImports": true,
+    "noImplicitAny": true,
```
mike-north: I always approach a large TS conversion by going through the following process:

1) Making a first pass w/ `"noImplicitAny": false` across the whole codebase. The main goal is to convert each file to `.ts`. This involves dealing with the subset of JS that TS is unhappy with (i.e., ES6 classes that don't state their member field types up front, extra stuff pinned onto functions, etc...)
2) Set  `"noImplicitAny": true` and make all the `any`s explicit instead of implicit
3) Get rid of any remaining `any`s as possible

Starting out with strict TS config will result in a lot of thrash, as new type-checking issues surface in waves (i.e., a commit that improves types in file B reveals type-checking failures in file A -- over and over and over)

ember-cli-build.js
```diff
@@ -5,7 +5,7 @@ const EmberAddon = require('ember-cli/lib/broccoli/ember-addon');
...
+      throwUnlessParallelizable: false,
```
mike-north: You'll want to straighten this out before considering a merge. The benefits of TS shouldn't have to come at the cost of serializing previously parallel parts of the build.

ember-cli-build.js
dfreeman: I'm not aware of anything we're doing that would prevent this, but agreed we should update our test suite to ensure it.

dfreeman: Just published 2.0.0-beta.2 with a fix

package.json
NullVoxPopuli: playing with potentially breaking things is my life ;)

addon/transforms/string.ts
NullVoxPopuli: unsure, I started this file as the first TS file just to see if I could get a build/tests going.

One of the things I'm excited about with the TS conversion, is that anybody will be able to answer that question :)

NullVoxPopuli:
saw this in the logs:
```
Circular dependency: ../../../../../tmp/broccoli-1217ZjAY9xcHWO0v/cache-089-rollup/build/-private/system/model/model.js -> ../../../../../tmp/broccoli-1217ZjAY9xcHWO0v/cache-089-rollup/build/-private/system/model/internal-model.js -> ../../../../../tmp/broccoli-1217ZjAY9xcHWO0v/cache-089-rollup/build/-private/system/references.js -> ../../../../../tmp/broccoli-1217ZjAY9xcHWO0v/cache-089-rollup/build/-private/system/references/belongs-to.js -> ../../../../../tmp/broccoli-1217ZjAY9xcHWO0v/cache-089-rollup/build/-private/system/model/model.js
```

runspired:
@NullVoxPopuli that's just the case in general on master atm and not an issue. It is one that I have a plan for resolving.

runspired:
superseded by #5719, but an incredible amount of thanks for keeping the exploration alive for this long and helping us know what steps we needed to take. <3

NullVoxPopuli:
Woo woo

PR #5657, opened 2018-10-01
fix(blueprints): remove second string argument from transform tests

### LINKED ISSUE #5656: transform-test blueprint uses qunit module wrong
Hi,
the current transform-test generates a qunit setup that works but isn't actually expected to work.

`module` [is defined as](https://api.qunitjs.com/QUnit/module) `QUnit.module( name [, hooks] [, nested ] )`

The generated module call looks like this:
`module('transform:<%= dasherizedModuleName %>', '<%= friendlyTestDescription %>', function(hooks) {`

This causes qunit to parse the friendlyTestDescription into its test environment:

Another indicator that this is probably a bug is that all other blueprints don't set the second argument as a string:

I'll open a PR soon to fix this

makepanic:
fixes #5656

### DIFF
--- a/blueprints/transform-test/qunit-files/__root__/__path__/__test__.js
+++ b/blueprints/transform-test/qunit-files/__root__/__path__/__test__.js
@@ -1,7 +1,7 @@
 import { module, test } from 'qunit';
 import { setupTest } from 'ember-qunit';
 
-module('transform:<%= dasherizedModuleName %>', '<%= friendlyTestDescription %>', function(hooks) {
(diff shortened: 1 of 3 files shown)


makepanic:
I just noticed that yarn updated the yarn.lock with `integrity`. Should i remove that from the PR?

runspired:
@makepanic yeah, please remove the yarn.lock change. Otherwise LGTM

makepanic:
I've removed the yarn.lock changes

runspired:
Thanks @makepanic !

PR #5663, opened 2018-10-02
[DEPRECATION errors] deprecates use of Ember.Evented with DS.Errors

runspired:
This one is tricky. We inserted warnings that we would eliminate mutating record-state from DS.Errors a very long time ago, but due to the mechanics of it all could not truly "deprecate". This "finalizes" that deprecation.

It also introduces the `Evented` deprecation from https://github.com/emberjs/rfcs/pull/329 for `DS.Errors` in a way that other classes that will need to deprecate events can also utilize.


runspired:
RFC has been opened here: https://github.com/emberjs/rfcs/pull/384

PR #5668, opened 2018-10-04
[CHORE record-data] Remove legacy private

runspired:
This removes all plumbing and files from the legacy internals. Going forward there is only `RecordData`. This helps to unblock the work for typescript and to separate ember-data into multiple-packages.

Of note: we still have a number of skipped tests that must be unskipped before we can ship this version of `ember-data`


NullVoxPopuli:
image
This is my favorite thing. :D

Issue #5574, opened 2018-08-17
Ember Data is using the links in relationships, not using data in relationships


fengshuo:
Hi,

I have an jsonapi response (`/contents`) with this format:

```
attributes: {},
id: 'XXX',
relationships: {
  provider: {
    data: { type: "providers", id: "99"},
    links: { self: "linkA", related: "linkB"}
  }
}
```

one provider may have many contents, the models has the `hasMany` and `belongsTo` relationship setup. The issue is that when requesting providers information, ember data is not using the `data` inside `relationships`, but using the `links` in `relationships`. 

I read in several places that ember data will use the `data` inside relationship if it exists, such as [here](https://discuss.emberjs.com/t/ember-jsonapi-how-to-correctly-handle-relationships/14009/2) and [here](http://www.amielmartin.com/blog/2017/05/17/how-ember-data-loads-async-relationships-part-2/), I wonder if ember data 3.3 is still working like this?

Thanks.

runspired:
@fengshuo if you give us a link, we will always use the link when fetching.

> I read in several places that ember data will use the data inside relationship if it exists

This is still true; however, this probably doesn't mean what you think. It refers to when you have previously loaded a record with that `type` and `id`, which includes sideloading the resource via `included: []` in the same API response.

mydea:
I do think this behavior has changed after 3.1.

In our app, with ED 3.1, for a hasMany relationship with included IDs and a link, it would load the relationship via the IDs initially. Later, if I'd call e.g. `model.hasMany('myRelationship').reload()`, it would use the link to re-fetch.

For us, this is a very useful functionality, as we can reduce the number of requests initially, while still being able to later reload the relationship.

As far as I see, this is not really possible (easily) anymore in 3.4. Now, it will _always_ use the link to fetch the relationship - which works, but leads to many more API requests.

A concrete example:

On a page, we load the hasMany relationship for a model, which returns n items.
Now, each of these items in turn can have X subItems, which we also load:

```js
let items = await get(model, 'items');
await RSVP.all(items.mapBy('subItems'));
```

Previously, this would result in 2 API requests: 

* `/api/models/XX/items` to fetch the first relationship - here, we include the IDs of the subitems as well as the link
* `/api/subItems?ids=1,2,3,4,5` then it would concatenate all the IDs from the subitems and get them in one big request.

After upgrading to ED 3.4, it results in 1 + n requests, where n is the number of items - as it will fetch the included link, one by one.

[...]

PR #5671, opened 2018-10-08
Use _scheduleFetch instead of _fetchRecord for belongsTo relationship

mydea:
In order to ensure coalesceFindRequests works for this case.

Without this, it would trigger one find-request per belongsTo relationship, instead of correctly coalescing them together.

I have stumbled over this while trying to debug something else, and was pretty puzzled as to why it wasn't working as expected. Changing this fixed the issue for me - not sure if there is something I'm not seeing there, though.

### DIFF
--- a/tests/unit/model/relationships/belongs-to-test.js
+++ b/tests/unit/model/relationships/belongs-to-test.js
@@ -1,6 +1,7 @@
 import { get } from '@ember/object';
 import { run } from '@ember/runloop';
 import setupStore from 'dummy/tests/helpers/store';
+import { Promise } from 'rsvp';
 
 import testInDebug from 'dummy/tests/helpers/test-in-debug';
 import { module, test } from 'qunit';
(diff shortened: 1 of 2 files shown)


runspired:
@mydea could you add some tests for coalescing? Both in the default and for relationships that don't have links? (we can consider the links case from our conversation in #5574 once [the RFC](https://github.com/emberjs/rfcs/pull/387) is complete). Thank you for doing all this work (including the RFC)!

mydea:
I added tests for this case - I hope in the right place, I wasn't entirely sure where to put this.
The tests cover fetching 2 belongsTo relationships concurrently, once with `coalesceFindRequests` and once without, and ensure that they run through either `findMany` or `findRecord`.

tests/unit/model/relationships/belongs-to-test.js
```diff
@@ -198,6 +199,176 @@ test('async belongsTo relationships work when the data hash has not been loaded'
...
+
+  env.adapter.findMany = function() {
+    throw new Error('findMany should not be called');
+  };
+
+  env.adapter.findRecord = function(store, type, id) {
+    assert.equal(type.modelName, 'tag', 'modelName is tag');
+
+    if (id === '3') {
+      return {
```
runspired: `return resolve(...data...)` will make this appropriately return a promise.

tests/unit/model/relationships/belongs-to-test.js
```diff
@@ -198,6 +199,176 @@ test('async belongsTo relationships work when the data hash has not been loaded'
...
+        data: {
+          id: '3',
+          type: 'tag',
+          attributes: { name: 'friendly' },
+        },
+      };
+    } else if (id === '4') {
+      return {
```
runspired: same as above, return a resolved promise

tests/unit/model/relationships/belongs-to-test.js
```diff
@@ -198,6 +199,176 @@ test('async belongsTo relationships work when the data hash has not been loaded'
...
+        data: {
+          id: '4',
+          type: 'tag',
+          attributes: { name: 'nice' },
+        },
+      };
+    }
+  };
+
+  return run(() => {
```
runspired: using `async ... await` instead of lots of run loops would clean this up nicely

tests/unit/model/relationships/belongs-to-test.js
```diff
@@ -198,6 +199,176 @@ test('async belongsTo relationships work when the data hash has not been loaded'
...
+    name: DS.attr('string'),
+    tag: DS.belongsTo('tag', { async: true }),
+  });
+
+  let env = setupStore({ tag: Tag, person: Person });
+  let { store } = env;
+
+  env.adapter.coalesceFindRequests = false;
+
+  run(() => {
```
runspired: if we make this an async test there is no need to wrap `store.push` in `run()`

mydea:
I made the requested changes to the test (to be honest, I only used to non-async-await syntax because I tried to stick to the same test style as the existing tests in the file. async-await is soo much nicer)!

PR #5676, opened 2018-10-12
update/remove rails/backend references in documentation

### LINKED ISSUE #5672: docs should not reference rails backend (or any backend)
The docs for `store` methods references Rails backend, with example, in a couple places.  We should remove references to specific backend implementations.

`addon/-private/system/store.js`

> The call to the server, using a Rails backend, will look something like this:

and 

> The call made to the server, using a Rails backend, will look something like this:

ctcpip:
resolves #5672

### DIFF
--- a/addon/-private/system/store.js
+++ b/addon/-private/system/store.js
@@ -342,7 +342,7 @@ Store = Service.extend({
 
     ```js
     store.createRecord('post', {
-      title: 'Rails is omakase'
+      title: 'Ember is awesome!'
     });
     ```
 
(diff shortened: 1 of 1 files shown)


kennethlarsen:
This looks good to me but for some reason, the following External Partner Tests fails:
Ember Data Factory Guy: 
```
Missing symlinked yarn packages: 
Package: ember-data
  * Specified: ~3.4
  * Symlinked: 3.5.0
```

Ember Data Change Tracker:
```
Missing symlinked yarn packages: 
Package: ember-data
  * Specified: ~3.3.0
  * Symlinked: 3.5.0
```

ember-m3:
Not really clear why.

@runspired Can you give a pointer on how to fix these tests?

runspired:
@kennethlarsen the addons (other than m3) fail because model-fragments does not support record-data yet. ember-m3 will fail until my PR to update it to use `recordData` instead of `modelData` is complete.

PR #5639, opened 2018-09-14
Include request/response info with Abort error

pixelhandler:
Currently the `AbortError ` does not have any context about the request/response.

This PR adds the request method, url and error if available...

```
error.message
"The adapter operation was aborted"
error.errors[0].title
"Adapter Error"
error.errors[0].detail
"Request failed: GET http://localhost:4200/api/v1/reports"
error.errors[0].status
0
```

The problem is that when you have this error there is no related info captured to identify why the adapter aborted. For example, using a service like bugsnag or raygun, the error may not have any information to help reproduce the problem in your application.

PR #5678, opened 2018-10-15
Cherry-pick Include request/response info with Abort error

pixelhandler:
`#5639 Include request/response info with Abort error` was released with v3.5, would be great to have this on the LTS version as well.

See PR #5639


pixelhandler:
@runspired thanks!

PR #5701, opened 2018-10-17
[WIP FEAT RecordIdentifier] provide a stable reference key for client side records

runspired:
Now that `RecordData` has landed we find ourselves looking for the path to:

- eliminate dependence on `InternalModel` by `store`, `Model`, and `record-data-wrapper` methods
- eliminate `InternalModel`
- clean up the relationship graph
- add the ability to provide multiple indexes for records

Much of our internals currently use `internalModel` as a "stable reference" due to the potential lack of an `id`. `RecordData` introduced `clientId` but only makes use of it for newly created records, leading to brittle contracts and redundant code checks and infra to handle the `clientId` vs `id` cases.  

Our relationship layer is now entangled with `RecordData` (which is responsible for membership), `InternalModel` (which is responsible for ui presentation), and `store` (which is responsible for converting the membership to records and which may or may not require network requests).

`unloadAll`, `unloadRecord`, and `peekAll` depend on `InternalModelMap`, which we now have two of and which doesn't fit properly in the post-InternalModel world.

This PR is a first step to addressing these problems and achieving our goals by moving to a system in which all records are uniformly referred to by a `recordIdentifier`, which will always have a `clientId`, now called `lid` in keeping with the `json-api` proposal.  This reduces method signatures, reduces risk of data misalignment, eliminates conversion into-out-of other ways of referencing records by unifying on a single standard, and improves our ability to better isolate our various primitives by providing a standard through which they can communicate.

### DIFF
--- a/addon/-private/system/cache/identifier-index.js
+++ b/addon/-private/system/cache/identifier-index.js
@@ -1,0 +1,218 @@
+// TODO kill this evil awful coerce thing
+import coerceId from '../coerce-id';
+import { DEBUG } from '@glimmer/env';
+
+let CLIENT_ID = 0;
+let IDENTIFIER = null;
+
+export function createIdentifierIndex(config) {
+  if (DEBUG) {
+    if (IDENTIFIER !== null) {
+      debugger;
+      throw new Error('createIdentifierIndex called multiple times without a matching call to clearIdentifierIndex.');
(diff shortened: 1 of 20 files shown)


runspired:
Closing in favor of #5750

PR #5703, opened 2018-10-18
[BUGFIX unloadRecord] bfs compatibility for custom RecordData

runspired:


tests/integration/records/record-data-test.js
```diff
@@ -0,0 +1,228 @@
...
+
+    try {
+      run(() => chris.unloadRecord());
+      assert.ok(true, 'expected `unloadRecord()` not to throw');
+    } catch (e) {
+      assert.ok(false, 'expected `unloadRecord()` not to throw');
+    }
+  });
+
+  test(`store.unloadRecord on a record with custom RecordData with relationship to a record with default RecordData does not error`, async function(assert) {
```
hjdivad: Let's add a test for DS.a -> custom -> DS.b; `DS.a.unloadRecord()` and confirm that this leaves `DS.b` around (unlike the non-custom link case)

tests/integration/records/record-data-test.js
```diff
@@ -0,0 +1,228 @@
...
+      this.relationships = null;
+    }
+    isNew() {
+      return this.id === null;
+    }
+  }
+
+  test(`store.unloadRecord on a record with default RecordData with relationship to a record with custom RecordData does not error`, async function(assert) {
+    const originalCreateRecordDataFor = store.createRecordDataFor;
+    store.createRecordDataFor = function provideCustomRecordData(
```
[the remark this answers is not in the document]
runspired: confirm

tests/integration/records/record-data-test.js
runspired: discussed oob that this test would be overly complicated for uncertain value

Issue #329, opened 2012-07-14
Association updates not saving


camshaft:
I opened a question on [stackoverflow](http://stackoverflow.com/questions/11480252/ember-js-data-associations-not-saving) but I realized this was more of an issue than a personal question.
### tl;dr

http://jsfiddle.net/CamShaft/HBm4j/

Let's say I have 2 models:

```
App.Address = DS.Model.extend({
   street: DS.attr('string'),
   person: DS.belongsTo('App.Person')
})

App.Person = DS.Model.extend({
  name: DS.attr('string'),
  addresses: DS.hasMany('App.Address')
})
```

Now I create a person

```
App.person = App.Person.createRecord({name: 'Bill'});
App.store.commit();
```

If I try to add an address to the person like this

```
address = App.Address.createRecord({street: '123 Fake Street'});
App.person.get('addresses').pushObject(address);
```

and commit the transaction

```
App.store.commit();
```

The new address will be saved however the person object will not be recognized as changed;
even though the list of ids has gone from

```
{
  ...
  "addresses": []
}
```

to 

```
{
  ...
  "addresses": [3]
}
```

PR #5704, opened 2018-10-19
[WIP DEPRECATION events] Implements RFC 329 evented deprecation

runspired:
[RFC #329](https://github.com/emberjs/rfcs/blob/5a1e8cddd3c1a210c2d60c7a9ac021dbfbedc370/text/1234-deprecated-ember-evented-in-ember-data.md)

Remaining Tasks

1) Revisit with @hjdivad plan to make `DeprecatedEvented` not be unnecessarily noisy on models and ensure end users may extend `Model` with `Evented` on their own.

2) Figure out if `AdapterPopulatedRecordArray` event `didLoad` is listened to and only fire deprecation then vs every time.

### DIFF
--- a/addon/-private/deprecated-evented.js
+++ b/addon/-private/deprecated-evented.js
@@ -1,0 +1,45 @@
+import Mixin from '@ember/object/mixin';
+import Evented from '@ember/object/evented';
+import { deprecate } from '@ember/debug';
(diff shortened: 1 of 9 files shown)


runspired:
Closing in favor of a new attempt

PR #5708, opened 2018-10-24
[BUGFIX CHORE tests] modernize unit/model-test (and some fixes that it uncovered)

### LINKED ISSUE #5679: Unable to peek created record from store by assigned id
Reproduction:
https://ember-twiddle.com/4520b9d2443bb9204d375b06d04ad530?openFiles=controllers.application.js%2C

Case:
I need to create totally new records, not yet persisted somewhere else, but whose ID is assigned by the client.

Working part:
```js
    let item2 = this.get('store').createRecord('user', { id: 43 });
    console.log('user:43', this.get('store').peekRecord('user', '43')); // --> item2 model
```
Broken part:
```js
    let item = this.get('store').createRecord('user');
    item.set('id', 42);
    console.log('user:42', this.get('store').peekRecord('user', '42')); //--> null
```

runspired:
resolves #5679

### DIFF
--- a/addon/-private/system/model/model.js
+++ b/addon/-private/system/model/model.js
@@ -1,13 +1,11 @@
-import ComputedProperty from '@ember/object/computed';
 import { isNone } from '@ember/utils';
 import EmberError from '@ember/error';
 import Evented from '@ember/object/evented';
 import EmberObject, { computed, get } from '@ember/object';
 import { DEBUG } from '@glimmer/env';
-import { assert, warn } from '@ember/debug';
+import { assert, warn, deprecate } from '@ember/debug';
 import { PromiseObject } from '../promise-proxies';
 import Errors from '../model/errors';
-import RootState from '../model/states';
 import {
   relationshipsByNameDescriptor,
   relationshipsObjectDescriptor,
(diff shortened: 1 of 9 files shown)


addon/-private/system/model/model.js
```diff
@@ -456,7 +459,7 @@ const Model = EmberObject.extend(Evented, {
   */
   toJSON(options) {
     // container is for lazy transform lookups
-    let serializer = this.store.serializerFor('-default');
+    let serializer = this._internalModel.store.serializerFor('-default');
```
igorT: Why do we not lookup store directly anymore?

addon/-private/system/relationship-meta.js
```diff
@@ -26,8 +26,7 @@ class RelationshipDefinition {
     this._type = '';
     this.__inverseKey = '';
     this.__inverseIsAsync = null;
-    this.modelClass = meta.parentType;
-    this.store = null;
```
igorT: Is this never set?

runspired: `store` was never set.

addon/-private/system/model/model.js
runspired: We only accessed `store` two places from `Model` ourselves. Moving these two places to `_internalModel` limits our exposure. We now no longer care whether `store` is properly assigned or not. We give it to the model but end users may override as it is not necessary to stay there.

PR #5709, opened 2018-10-25
[CHORE deps] upgrades all dependencies to latest (babel7!)

runspired:
cc @NullVoxPopuli


NullVoxPopuli:
Excellent. This should unblock me using the ects beta :D

runspired:
Not sure why the travis build didn't link here, but here it will be: https://travis-ci.org/emberjs/data/builds/446415823

NullVoxPopuli:
Looks like the build needs to be re ran?

NullVoxPopuli:
what's going on with the node tests? a quick skim doesn't show anything other than the generated files missing. _hmm_

runspired:
@nullvoxpopuli  I fixed the  finger printing issue that affects 3.4 and 3.5  and forgot that 3.5 breaks the tests for blueprints  because it doesn’t work with temp directories correctly.  I had to run to a train but I’ll push up a commit pinning to 3.4 in a moment

PR #5715, opened 2018-10-26
[CHORE tests] modernize integration/inverse-test

runspired:
Ported from #5708

### DIFF
--- a/tests/integration/inverse-test.js
+++ b/tests/integration/inverse-test.js
@@ -1,169 +1,392 @@
-import { run } from '@ember/runloop';
-import setupStore from 'dummy/tests/helpers/store';
-
-import testInDebug from 'dummy/tests/helpers/test-in-debug';
+import { setupTest } from 'ember-qunit';
 import { module, test } from 'qunit';
+import testInDebug from 'dummy/tests/helpers/test-in-debug';
+import Model from 'ember-data/model';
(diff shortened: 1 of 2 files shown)


tests/integration/inverse-test.js
```diff
@@ -1,169 +1,392 @@
...
+  test('Finds the inverse when only one side has defined it manually', function(assert) {
+    let { owner } = this;
+    owner.unregister('model:job');
+    owner.unregister('model:user');
 
-    ReflexiveModel = DS.Model.extend({
-      reflexiveProp: belongsTo('reflexive-model', { async: false }),
-    });
+    class Job extends Model {
+      @belongsTo('user', { inverse: 'previousJob', async: false })
```
igorT: I think this is missing the user belongs to which was there before. The reopen just added an owner, it didn't clobber the user

runspired: updated and restored and re-ordered the assertions to ensure we were capturing that.

I noted during the refactor that I was being "lossy" on this test from the original, but felt it didn't matter because the test was looking up the relationship via the `job` model first, which often avoids any issues of there being multiple potential inverses since it was the side with multiple.

I've updated the test to instead lookup the relationship via the `User` model first, and restored the extra `user` relationship on `Job`.

PR #5717, opened 2018-10-26
[CHORE tests] modernize unit/model/lifecycle-callbacks-test

runspired:
ported from #5708

### DIFF
--- a/tests/unit/model/lifecycle-callbacks-test.js
+++ b/tests/unit/model/lifecycle-callbacks-test.js
@@ -1,299 +1,1118 @@
-import { resolve, reject } from 'rsvp';
-import { get } from '@ember/object';
-import { run } from '@ember/runloop';
-import { createStore } from 'dummy/tests/helpers/store';
-
+import { setupTest } from 'ember-qunit';
 import { module, test } from 'qunit';
+import { settled } from '@ember/test-helpers';
+import { resolve, reject } from 'rsvp';
+import { attr } from '@ember-decorators/data';
+import Model from 'ember-data/model';
+import Adapter from 'ember-data/adapter';
+import JSONAPISerializer from 'ember-data/serializers/json-api';
+import { InvalidError } from 'ember-data/adapters/errors';
+import { run } from '@ember/runloop';
 
-import DS from 'ember-data';
+module('unit/model - Model Lifecycle Callbacks', function(hooks) {
+  setupTest(hooks);
+  let store;
 
-module('unit/model/lifecycle_callbacks - Lifecycle Callbacks');
-
-test('a record receives a didLoad callback when it has finished loading', function(assert) {
-  assert.expect(3);
-
-  const Person = DS.Model.extend({
-    name: DS.attr(),
-    didLoad() {
-      assert.ok('The didLoad callback was called');
-    },
+  hooks.beforeEach(function() {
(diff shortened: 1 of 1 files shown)


tests/unit/model/lifecycle-callbacks-test.js
```diff
@@ -1,299 +1,1118 @@
...
+    lifecycleEventMethodCalls = 0;
 
-  const Person = DS.Model.extend({
-    bar: DS.attr('string'),
-    name: DS.attr('string'),
+    await settled();
+    assert.strictEqual(
+      lifecycleEventMethodCalls,
+      1,
+      'We do not trigger didDelete when we first call record.deleteRecord'
```
igorT: We *do* trigger didDelete

tests/unit/model/lifecycle-callbacks-test.js
```diff
@@ -1,299 +1,1118 @@
...
+      assert.strictEqual(lifecycleEventMethodCalls, 1, 'We trigger becameInvalid when we reject');
 
-          assert.equal(reason.errors.length, 1, 'reason should have one error');
-          assert.equal(reason.errors[0].title, 'Invalid Attribute');
-          assert.equal(callCount, 1, 'becameInvalid called after invalidating');
-        });
-      });
+      assert.ok(reason.isAdapterError, 'reason should have been an adapter error');
+      assert.equal(reason.errors.length, 1, 'reason should have one error');
+      assert.equal(reason.errors[0].title, 'Invalid Attribute');
```
igorT: We're missing an assertion here for triggering didUpdate

tests/unit/model/lifecycle-callbacks-test.js
```diff
@@ -1,299 +1,1118 @@
...
-  // Make sure that the error handler has a chance to attach before
-  // save fails.
-  return run(() => {
-    return asyncPerson.then(person => {
-      return run(() => {
-        person.set('bar', 'Bar');
-        return person.save().catch(reason => {
-          assert.ok(reason.isAdapterError, 'reason should have been an adapter error');
+    await person.save().catch(reason => {
+      assert.strictEqual(lifecycleEventMethodCalls, 1, 'We trigger becameInvalid when we reject');
```
igorT: Just curious, what are the reasons for preferring `strictEqual` over `equal` in some cases?

tests/unit/model/lifecycle-callbacks-test.js
igorT: Fixed

tests/unit/model/lifecycle-callbacks-test.js
runspired: I use `strictEqual` when we do expect strict referential equality. The advantage is two-fold: (1) it avoids a set of bugs that happen when qunit attempts to do deep recursive "looks like the same thing" comparisons when strict equality is not matched (on records this is a mega perf hit in addition to being risky), and it is an improvement over the alternative pattern of `assert.ok(foo === bar, '')` wherein if the assertion fails there is no additional information printed about what the incorrect value is without stepping into the test.

tests/unit/model/lifecycle-callbacks-test.js
runspired: thanks!

PR #5722, opened 2018-10-27
Convert -private/utils to TypeScript

NullVoxPopuli:
this directory is just one file. figured I'd start small :)

### DIFF
--- a/addon/-private/types.ts
+++ b/addon/-private/types.ts
@@ -1,0 +1,3 @@
+type Dict<T = unknown> = {
+  [k: string]: T | undefined;
+}
--- a/addon/-private/types.ts
+++ b/addon/-private/types.ts
@@ -1,3 +1,3 @@
-type Dict<T = unknown> = {
+export type Dict<T = unknown> = {
   [k: string]: T | undefined;
 }
--- a/addon/-private/utils/parse-response-headers.ts
+++ b/addon/-private/utils/parse-response-headers.ts
@@ -1,8 +1,8 @@
 const newline = /\r?\n/;
 
-export type ResponseHeaders =
-| {}
-| { [key: string]: string };
+interface ResponseHeaders {
+  [key: string]: string
+};
 
 export default function parseResponseHeaders(headersString: string): ResponseHeaders {
   let headers: ResponseHeaders = Object.create(null);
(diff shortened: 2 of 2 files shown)


addon/-private/utils/parse-response-headers.ts
```diff
@@ -1,7 +1,11 @@
 const newline = /\r?\n/;
 
-export default function parseResponseHeaders(headersString) {
-  let headers = Object.create(null);
+export type ResponseHeaders =
+| {}
+| { [key: string]: string };
```
mike-north: Why have you chosen this type instead of ` { [key: string]: string } `?

NullVoxPopuli: good question, I don't remember. everything seems fine without the extra `{}`

addon/-private/utils/parse-response-headers.ts
```diff
@@ -1,7 +1,11 @@
 const newline = /\r?\n/;
 
-export default function parseResponseHeaders(headersString) {
-  let headers = Object.create(null);
+interface ResponseHeaders {
+  [key: string]: string
+};
```
mike-north: This type is way more foundational than just "response headers". 

Let's create a `..../-private/types.ts` and have a

```ts
export interface Dict<T = unknown> {
   [k: string]: T | undefined;
}
```

and then you can import it and

```ts
export default function parseResponseHeaders(headersString: Dict<string>)
```

NullVoxPopuli: would that not be something you'd want in the global index.d.ts?

mike-north: Nope. Consumers of the ember-data won't get any of those types.

mike-north: I'm not sure you understand. Everything in your `/types` folder is for _your project only_. If you put `Dict` there, and used it anywhere in parts of the library that users of ember-data touch, it'll cause them problems.

NullVoxPopuli: right, in my latest commit, I moved the Dict type to -private/types.ts, and it didn't need to be imported. :-\

addon/-private/types.ts
```diff
@@ -0,0 +1,3 @@
+type Dict<T = unknown> = {
```
mike-north: you still need the export, and the corresponding import

runspired:
Thanks for the effort! I think this PR is stale enough likely we should reboot from where we are now.

We have a `Dict` in our `ts-interfaces` currently but the upgrades listed here would be welcome.

PR #5723, opened 2018-10-27
Convert some system utilities to ts

NullVoxPopuli:
This is just some of the smaller utilities


addon/-private/system/is-array-like.ts
```diff
@@ -11,7 +16,7 @@ import EmberArray from '@ember/array';
   https://github.com/emberjs/ember.js/pull/11463 `Ember.isArray` is an alias of
   `Array.isArray` hence removing the "array-like" part.
  */
-export default function isArrayLike(obj) {
+export default function isArrayLike(obj?: MaybeArrayLike) {
```
NullVoxPopuli: @chriskrycho / @mike-north / @dfreeman not really sure how to type this one.

Also, side note - the check for setInterval is strange to me. Are we passing window at some point to this function??

addon/-private/system/normalize-link.ts
```diff
@@ -10,7 +18,7 @@
   @return {Object|null}
```
NullVoxPopuli: the documentation says that link should be a string, but the code says it can be an object. 
She these docs be updated?

types/dummy/index.d.ts
```diff
@@ -3,3 +3,15 @@
  * ember-data. They will not be shipped to consumers. Ember-data still relies
  * on some private Ember APIs -- those should be defined here as we encounter them.
  */
+
```
NullVoxPopuli: these do not exist in the ember.js repo / or the mixin system is throwing TS off (which I think is the case for detect?)

addon/-private/system/is-array-like.ts
chriskrycho: Correct: the point of `unknown` is that you have no idea what is passed in. You can get safer internally by way of [type narrowing](http://www.typescriptlang.org/docs/handbook/advanced-types.html#type-guards-and-differentiating-types).

types/dummy/index.d.ts
dfreeman: I don't see a `detect` function being exported from `@ember/array` either in [the API docs](https://emberjs.com/api/ember/release/modules/@ember%2Farray) or in the [RFC 176 data](https://github.com/ember-cli/ember-rfc176-data); does importing that actually work?

Re: `@ember/ordered-set`, that's [a standalone package](https://github.com/emberjs/ember-ordered-set), not a part of core Ember.

rwjblue: FWIW, `detect` is on `Mixin` [see here](https://github.com/emberjs/ember.js/blob/19f5c8919052d1acea9147b341dbbc8009c34f59/packages/%40ember/-internals/metal/lib/mixin.ts#L606) (and therefore `Ember.Array` mixin does have a `.detect` method).  Not sure if that was the specific question though...

dfreeman: Right, but this declaration is saying you should be able to `import { detect } from '@ember/array';` (which afaik you can't, since `detect` is a method on the mixin, not a module export, right? 🙂)

rwjblue: Yep, I totally missed that this was an import. I'll put my history lessons away, sorry for the noise 😸

addon/-private/system/diff-array.ts
```diff
@@ -10,7 +16,7 @@
       removedCount: <integer>       // 0 if no change
     }
 */
-export default function diffArray(oldArray, newArray) {
+export default function diffArray<T = any>(oldArray: T[], newArray: T[]): ArrayDiffResult {
```
mike-north: in what scenario will `T` fail to be inferred?

addon/-private/system/normalize-link.ts
```diff
@@ -1,3 +1,11 @@
+export interface Link {
+  href: string;
+}
+
+export type NormalizedLink =
+| Link
+| null;
```
mike-north: Should use some centrally defined Maybe<T>, rather than defining and exporting a new type alias

addon/-private/system/coerce-id.ts
```diff
@@ -1,10 +1,16 @@
+export type DirtyId =
+| null
+| undefined
+| string
+| number;
```
mike-north: I see a lot of named & exported interfaces in situations where we need these types _once_. We should decide whether this will cause needless clutter in e-d's internal API surface, compared to something like 
```ts
type Maybe<T> = T | null; // defined somewhere else

export default function coerceId(id?: Maybe<string|number>): Maybe<string>;
```

addon/-private/system/normalize-link.ts
```diff
@@ -10,7 +18,7 @@
   @return {Object|null}
   @for DS
 */
-export default function _normalizeLink(link) {
+export default function _normalizeLink(link: Link | string): NormalizedLink {
   switch (typeof link) {
     case 'object':
       return link;
```
mike-north: an exhaustive switch should be used here http://ideasintosoftware.com/exhaustive-switch-in-typescript/

`NormalizedLink` has the potential to be null -- what value for `Link|string` could be passed as an argument, which would result in `null` being returned?

types/dummy/index.d.ts
```diff
@@ -3,3 +3,15 @@
...
+ declare module '@ember/array' {
+   export function detect(obj: any): boolean;
```
mike-north: this is probably best expressed as a user-defined type guard

addon/-private/system/is-array-like.ts
mike-north: this is probably best expressed as a user-defined type guard
```ts
function isArrayLike(obj: unknown): obj is ArrayLike
```

addon/-private/system/coerce-id.ts
NullVoxPopuli: I made a `Nullable<T>`, as `Maybe` made me feel like `undefined` would _maybe_ be included. :)

types/dummy/index.d.ts
NullVoxPopuli: do you mean `unknown` instead of `any`? :confused:

NullVoxPopuli:
image
not sure why I can't reply to this one,
but I saw that @runspired is removing array-like, so, I'll wait for that and then rebase :)

addon/-private/system/normalize-link.ts
NullVoxPopuli: done, and yes, null => null, but only because `typeof null === 'object'`

addon/-private/system/is-array-like.ts
igorT: Luckily this is a moot issue given https://github.com/emberjs/data/pull/5724

addon/-private/system/normalize-link.ts
mike-north: I'm less concerned about `typeof` than I am about type equivalence in a `"strictNullChecks": true` situation. 

```ts
const myObj1: object = null; // ❌
const myObj2: {} = null; // ❌
```

types/dummy/index.d.ts
mike-north: no, I mean
```ts
export function detect(obj: any): boolean is ArrayMixin;
```

addon/-private/system/diff-array.ts
```diff
@@ -1,3 +1,9 @@
+export interface ArrayDiffResult {
+  firstChangeIndex: number | null;
```
mike-north: This should use `Maybe<T>` or `Nullable<T>` whatever you are calling it

addon/-private/system/normalize-link.ts
NullVoxPopuli: I'm confused by your link about exhaustive switch, as it uses an enum, and has nothing to do with types.

I can explicitly check for null, which would be more intuitive than what is going on now

chriskrycho: The link is getting at the fact that you can make TS give you a *type* error if you enumerate the cases you expect, so that if you expand that number of cases later and fail to cover it, you'll *know* about it. The easiest (though not only) way to do that is with a string-valued enum; you could also do it simply with a type alias enumerating the cases.

types/dummy/index.d.ts
chriskrycho: I think you mean this:

```ts
export function detect(obj: any): obj is ArrayMixin;
```

NullVoxPopuli: but https://github.com/emberjs/ember.js/blob/9bdc0075762ac24bf18629d562548b140d0309cd/packages/%40ember/-internals/metal/lib/mixin.ts#L606

chriskrycho: `<item> is <Type>` is a specific kind of `boolean`. In this case, as I understand it, it's specifically detecting that this is an `ArrayMixin`, correct? If so, then you want the type guard form, not just `boolean`, because it lets TypeScript narrow the type so that when you call `.detect` on the other side it's confirmed to be an `ArrayMixin` from the type system's POV.

chriskrycho: See the [User-Defined Type Guards](https://www.typescriptlang.org/docs/handbook/advanced-types.html#user-defined-type-guards) section of the handbook for more.

NullVoxPopuli: oh! I understand now, the `is` syntax is a like the `where` clause in C#.
Legit

addon/-private/system/diff-array.ts
mike-north: Please remove the `= any`. There's no need for a type parameter default here.

addon/-private/system/normalize-link.ts
mike-north: we need a subject matter expert to comment on this. @runspired @igorT

addon/-private/system/diff-array.ts
mike-north: There's no need for the `= any` here.

addon/-private/system/normalize-link.ts
NullVoxPopuli: I don't know how to apply that here.
sorry for not understanding :(

addon/-private/system/diff-array.ts
NullVoxPopuli: it is removed

addon/-private/system/normalize-link.ts
chriskrycho: No worries! It's complicated and not *super* well-explained in the docs if you don't already have a frame for what's going on. I will try to write up a good explanation as well as a specific recommendation for how to implement it in this PR tomorrow!

addon/-private/system/normalize-link.ts
runspired: The docs should be updated

runspired:
Thanks for the effort! I think this PR is stale enough likely we should reboot from where we are now.

PR #5724, opened 2018-10-27
[CHORE is-array-like] make isArrayLike die a terrible death

runspired:
we should look into removing the use of `Ember.Array.detect` as well.

cc @nullvoxpopuli for pointing out this thing in #5723


igorT:
Should we keep supporting objects that have a `length` on them? Technically, this might break things, but not sure if people are passing things like that in.

runspired:
@igorT I considered this but after reviewing the two locations we use this I am:

1) fairly convinced that using something other than an array or ArrayProxy does not actually work because of how we cast to `Ember.A` and use `.every`  and `.map`

https://ember-twiddle.com/96d305cbdad14d392d2c057029127f8b?openFiles=controllers.application.js%2C

2) Suspect that support here of `ArrayProxy` is also not required and should be deprecated. Receiving an ArrayProxy would likely indicate someone trying to set the `ManyArray` or `PromiseProxyArray` from one record directly onto another record. In the past we have indeed done things to explicitly support this, but I feel we should lean towards deprecating this support and encouraging use of native `array`. To transfer from one relationship to another, users may use `toArray`. In addition, I'm not very convinced that passing a `ManyArray` or `PromiseProxyArray` wouldn't error or cause bugs.

igorT:
:+1: on the length issue
I think that something like query results might be a valid/supported use case for setting many array values to a non native array

PR #5727, opened 2018-10-31
[CHORE ts] A couple of dabblings to start us down the road

runspired:
// ... please delete this default text after reading

If this is your first PR to `ember-data`, you may want to read our [Contributor Guide](https://github.com/emberjs/data/blob/master/CONTRIBUTING.md).

### DIFF
--- a/addon/-private/system/internal-model-map.ts
+++ b/addon/-private/system/internal-model-map.ts
@@ -2,10 +2,10 @@ import { assert } from '@ember/debug';
 import InternalModel from './model/internal-model';
 
 interface IDDict {
-  [id: string]: InternalModel
+  [id: string]: InternalModel;
 }
 interface MetaDict {
-  [key: string]: any
+  [key: string]: any;
 }
 
 /**
@@ -21,7 +21,7 @@ interface MetaDict {
 export default class InternalModelMap {
   private _idToModel: IDDict;
   private _models: InternalModel[];
-  private _metadata: MetaDict|null;
+  private _metadata: MetaDict | null;
 
   modelName: string;
 
@@ -37,7 +37,7 @@ export default class InternalModelMap {
    * @param id {String}
    * @return {InternalModel}
    */
-  get(id: string): InternalModel|undefined {
+  get(id: string): InternalModel | undefined {
     return this._idToModel[id];
   }
 
@@ -50,7 +50,10 @@ export default class InternalModelMap {
   }
 
   set(id: string, internalModel: InternalModel): void {
-    assert(`You cannot index an internalModel by an empty id'`, id);
+    assert(
+      `You cannot index an internalModel by an empty id'`,
+      typeof id === 'string' && id.length > 0
+    );
     assert(
       `You cannot set an index for an internalModel to something other than an internalModel`,
       internalModel instanceof InternalModel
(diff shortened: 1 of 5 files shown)


addon/-private/system/internal-model-map.ts
```diff
@@ -24,19 +37,19 @@ export default class InternalModelMap {
...
     return this._idToModel[id];
   }
 
-  has(id) {
+  has(id: string): boolean {
     return !!this._idToModel[id];
   }
 
-  get length() {
+  get length(): number {
```
mike-north: If you want to use tslint to enforce explicit return types, the rule looks like

```json
"rules": {
  "typedef": [true, "call-signature"]
}
```

types/dummy/index.d.ts
```diff
@@ -3,3 +3,25 @@
  * ember-data. They will not be shipped to consumers. Ember-data still relies
  * on some private Ember APIs -- those should be defined here as we encounter them.
  */
+declare module '@ember/debug' {
+  export function assert(msg: string, test: any): boolean;
```
igorT: should test be `Function | boolean`

mike-north: @igorT I usually rely on the API docs as the source of truth, and they don't indicate that a predicate can be passed in.
https://www.emberjs.com/api/ember/3.4/functions/@ember%2Fdebug/assert

I could see where passing a predicate could be confusing
```ts
assert('Problem with toString', obj.toString);
```
Does this ☝️ mean "the value of `.toString` was falsy", or "the value returned by `obj.toString()` was falsy"?


RE: whether the types should be imported vs. explicitly stated w/in ember-data, `assert` should be provided by the `@types/ember__debug` package. No need for it to be defined in this project.

https://github.com/DefinitelyTyped/DefinitelyTyped/blob/549360d30e796ac5635e019cb276e8af9933ff71/types/ember__debug/index.d.ts#L10

runspired: I'll look into whether I can import this type correctly, it wasn't resolving with the installed ember types but perhaps I just need to add the debug module.

addon/-private/system/internal-model-map.ts
mike-north: anything in the "rules" object applies only to TS, and there's a "jsRules" for JS

runspired: ooooooh :D then yes, I'll turn that on :D

types/dummy/index.d.ts
mike-north: @types/ember has @types/ember__debug as a dependency. Should work.

addon/-private/system/internal-model-map.ts
```diff
@@ -12,7 +19,13 @@ import InternalModel from './model/internal-model';
  @private
  */
 export default class InternalModelMap {
-  constructor(modelName) {
+  private _idToModel: IDDict;
+  private _models: InternalModel[];
+  private _metadata: MetaDict | null;
```
mike-north: It makes sense to move the initializations in the constructor below up here.

```ts
  private _idToModel: IDDict = Object.create(null);
  private _models: InternalModel[] = [];
  private _metadata: MetaDict | null = null;

  constructor(public modelName: string) {}
```

PR #5729, opened 2018-10-31
Fixed small Docs and Code mismatches

spruce:
As I was working on ember-api-docs I noticed some small missmatches between ED Api-docs and the actual implementation. Also EmberObject should also be referenced without a dot. I did fix these. 

- [ ] Also I notices that there is no Documentation for DS.Reference.
- [ ] There is no Reference of all the uses of AdapterError in `addon/-private/adapters/errors.js`. Should I add them?


igorT:
Thanks for the doc fixes! I think adding the `AdapterError` uses would be great.

PR #5714, opened 2018-10-26
[CHORE tests] modernize integration/record-array-test

runspired:
extracted from #5708

PR #5733, opened 2018-11-01
[BUGFIX inspector] Fix columns names in debug-adapter

thorsteinsson:
Fixed formatting of column desc to support many words. It used to replace only the first underscore with space and leave the rest.

### DIFF
--- a/addon/-private/system/debug/debug-adapter.js
+++ b/addon/-private/system/debug/debug-adapter.js
@@ -78,9 +78,7 @@ export default DataAdapter.extend({
 
   getRecordColumnValues(record) {
     let count = 0;
-    let columnValues = {
-      id: get(record, record.primaryKey || 'id'),
-    };
+    let columnValues = { id: get(record, 'id') };
 
     record.eachAttribute(key => {
       if (count++ > this.attributeLimit) {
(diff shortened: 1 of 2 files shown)


runspired:
@thorsteinsson the tests file is `tests/integration/debug-adapter-test.js`, if you have a moment, if you could modernize it while working on it that would be awesome.

Example modernization PRs:

- #5712 
- #5713 
- #5714 
- #5715
- #5716
- #5717

runspired:
@thorsteinsson so after reviewing this I'm fairly sure that whatever issue is being seen in Ember inspector is not related to `ember-data`. Ember Data does not and never has allowed users to use a property other than `id` as the `primaryKey` field on `Model`. The `primaryKey` modification shown in the associated PR is a pre-normalization or post-serialization property name unrelated to the store's internal format or the interface available on `Model`.

When we interface with the `data-adapter` we are always dealing with normalized data, and thus will only ever be using `id`.  Either a different data-adapter for a different data-layer is interfacing with the inspector in an incorrect way or there's some other bug involved.

thorsteinsson:
@runspired Makes sense, I wanted to use the same column name as defined in the model to provide better info in the inspector. I can change that back to `id` and only fix how the data is returned based on `primaryKey`.

I've tested the changes with the Inspector and it works. I'll work on modernizing the tests.

Since the `DebugAdapter` is only used to get debug information (I guess), should be only included in dev builds or be in a separate repo? There's a comment at https://github.com/emberjs/data/blob/master/addon/-private/index.js#L45

PR #5736, opened 2018-11-03
[DOC] Add Reference doc and AdapterError uses

spruce:
As mentioned in a previous [PR](https://github.com/emberjs/data/pull/5729) this fixes some missing Docs.

Should the Reference Doc somehow be rewritten?


runspired:
> Should the Reference Doc somehow be rewritten?

I don't believe so. `Reference` is essentially an abstract base class that users shouldn't ever have access to or need to know about. The documentation should live in the sub-classes. We recently did a Typescript based cleanup of Reference that should help with this :) See #5730

PR #5737, opened 2018-11-05
[CHORE tests] modernize integration/adapter/find-all-test

thorsteinsson:


tests/integration/adapter/find-all-test.js
```diff
@@ -164,78 +143,63 @@ test('When all records for a type are requested, records that are already loaded
...
-        return findAllDeferred.promise;
-      },
+  testInDebug('When all records are requested, assert the payload is not blank', async function(
+    assert
+  ) {
+    let adapter = store.adapterFor('person');
+    adapter.findAll = () => resolve({});
+
+    assert.expectAssertion(() => {
+      run(() => store.findAll('person'));
```
runspired: we can clean this `run` up in another pass but I believe either `expectAssertion` itself or a similar util is `promise` aware that we can use instead.

tests/integration/adapter/find-all-test.js
```diff
@@ -244,44 +208,33 @@ test('isUpdating is true while records are fetched', function(assert) {
...
 
-  let wait = run(() => {
-    return store.findAll('person').then(persons => {
-      assert.equal(persons.get('isUpdating'), false);
-      assert.equal(persons.get('length'), 2);
+    let wait = run(() => {
+      return store.findAll('person').then(persons => {
+        assert.equal(persons.get('isUpdating'), false);
+        assert.equal(persons.get('length'), 2);
+      });
```
runspired: A better pattern than:

```
let wait = run(() => return new Promise())
// ... assert intermediate state
await wait;
```

is

```
let promise = new Promise()
// ... assert intermediate state
await promise;
```

mostly this just means you can remove the `run` here. 👍

tests/integration/adapter/find-all-test.js
```diff
@@ -338,16 +280,16 @@ test('isUpdating is false if records are not fetched in the background', functio
...
-  });
 
-  let persons = store.peekAll('person');
-  assert.equal(persons.get('length'), 1);
+    let persons = store.peekAll('person');
+    assert.equal(persons.get('length'), 1);
 
-  return run(() => {
-    return store.findAll('person').then(persons => {
+    return run(() => {
```
runspired: we should be able to eliminate this use of `run`

tests/integration/adapter/find-all-test.js
thorsteinsson: I'm not sure what's the best approach here, more info at https://github.com/workmanw/ember-qunit-assert-helpers/issues/18.

igorT:
Thanks a bunch for the PR! Can you please rebase it against master, there are some merge conflicts with master right now.

PR #5738, opened 2018-11-05
[CHORE tests,owner,factoryFor] post-LTS and test cleanup

runspired:
- eliminates unnecessary fallbacks for factoryFor and getOwner
- almost eliminates use of `__container__` and related private `Container` and `Registry` API usage (we will need to kill `createStore` to kill that)
- improves test isolation
- eliminates `_attributes` and `_relationships` from `InternalModel` which were left for tests and are no longer required.


addon/-private/system/store.js
```diff
@@ -3411,11 +3410,7 @@ function getModelFactory(store, cache, normalizedModelName) {
 function _lookupModelFactory(store, normalizedModelName) {
   let owner = getOwner(store);
 
-  if (owner.factoryFor) {
-    return owner.factoryFor(`model:${normalizedModelName}`);
-  } else {
-    return owner._lookupFactory(`model:${normalizedModelName}`);
-  }
+  return owner.factoryFor(`model:${normalizedModelName}`);
```
rwjblue: Just to confirm, this drops support for Ember < 2.12 (unless someone brings in the factoryFor polyfill): what is the current minimum Ember version supported by Ember Data?

addon/-private/system/store.js
```diff
@@ -3438,14 +3433,8 @@ function _modelForMixin(store, normalizedModelName) {
   // container._registry = 1.11 - 2.0
```
rwjblue: These comments seem suspect

tests/acceptance/relationships/belongs-to-test.js
```diff
@@ -1,4 +1,4 @@
-import { module, skip as test } from 'qunit';
```
rwjblue: Waaat, this whole file was skipped?!

tests/test-helper.js
```diff
@@ -1,17 +1,13 @@
...
 import resolver from './helpers/resolver';
 import { setResolver } from '@ember/test-helpers';
 import { start } from 'ember-qunit';
 
 import QUnit from 'qunit';
 import DS from 'ember-data';
 import { wait, asyncEqual, invokeAsync } from 'dummy/tests/helpers/async';
-import loadInitializers from 'ember-load-initializers';
 
 setResolver(resolver);
```
rwjblue: We should change this to setApplication (not blocking)

addon/-private/system/model/internal-model.js
```diff
@@ -221,16 +220,6 @@ export default class InternalModel {
     return this.currentState.dirtyType;
   }
 
-  // DO NOT USE : purely to ease the transition in tests
-  get _attributes() {
```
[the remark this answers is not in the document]
runspired: If they are already doing this *bad thing* then they have both a good thing or a slightly different *bad thing* now available to them.

addon/-private/system/store.js
runspired: We have a "last two LTS" policy, meaning we support Ember 2.18+ currently

addon/-private/system/store.js
runspired: removed :)

tests/acceptance/relationships/belongs-to-test.js
runspired: yes, so was another one. Now only a few tests are skipped :)

tests/test-helper.js
runspired: Got a link handy for what that looks like?

.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
```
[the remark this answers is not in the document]
runspired: removed

addon/-private/system/model/internal-model.js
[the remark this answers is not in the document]
runspired: @rwjblue also, I think we are fairly sure in regard to `_attributes` (although not `_relationships`) that people ARE doing bad things. We could leave a nice assert here vs allowing it to fail with a less clear error.  Not totally sure how I feel about a deprecation for a private API on a Private (even if intimate) class. I suspect it would only encourage them to do the other bad thing (`internalModel._recordData._attributes`) instead of the good thing (implementing a custom `RecordData`).

tests/test-helper.js
rwjblue: This is what a new addons looks like: https://github.com/ember-cli/ember-addon-output/blob/master/tests/test-helper.js

runspired: Ok, looked into this and it appears that until we've killed the use of `createStore` that this seems to cause us too many issues (specifically we have tests without containers and tests for which this causes us to create multiple application instances). I will leave a TODO with this note.

runspired: Was able to get us further and moved the resolver setup into `createStore`'s domain. We have ~1100 of 1500 tests still setting up the store using this helper, so the burndown on it is likely to take a while.

addon/-private/system/model/internal-model.js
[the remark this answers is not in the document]
rwjblue: Ya, I’m not saying it’s blocking per-se I just wanted to call it out. It doesn’t seem directly related to the testing changes in the rest of the PR...

igorT:
Broken apart into PRs, except for `more nice things for @rwjblue` which I'll put in as soon as couple of PRs are merged due to merge conflicts

PR #5742, opened 2018-11-06
[CHORE tests] does a partial refactor of the unload tests to eliminate setupStore

igorT:
depends on #5741 and #5739


tests/integration/records/unload-test.js
```diff
@@ -136,2827 +138,2873 @@ module('integration/unload - Unloading Records', {
       book: Book,
       spoon: Spoon,
       show: Show,
-    });
-  },
+    };
+    Object.keys(models).forEach(key => owner.register(`model:${key}`, models[key]));
```
rwjblue: IMHO the `models` iteration here is a tad "cute", I think it would be easier to read as multiple lines like `owner.register(`model:foo`, Foo)` (and remove the `models` POJO just above)...

rwjblue:
In order to make these sorts of things easier to review, making the "actual" changes separate from whitespace / re-indenting would make review much much simpler. As it stands with this PR it was very difficult to review...

PR #5743, opened 2018-11-06
drop use of MODEL_FACTORY_FOR

igorT:
from #5738


tests/integration/injection-test.js
```diff
@@ -1,118 +1,71 @@
...
+    let { owner } = this;
+    Model = {
+      isModel: true,
+    };
+    owner.register('model:super-villain', Model);
+    store = owner.lookup('service:store');
+
+    let originalFactoryFor = owner.factoryFor;
+
+    owner.factoryFor = function interceptFactoryFor() {
```
rwjblue: This seems really bad... Why do we need to do this?

rwjblue: Unless I'm missing something, we can remove this whole thing (and the `factory` variable)...

runspired: converted to using `owner.factoryFor` to get the factory in the test.

PR #5744, opened 2018-11-06
[CHORE tests,owner,factoryFor] eliminate unnecessary fallbacks for fa…

igorT:
from #5738


.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
```
rwjblue: Should be removed (and added to global git ignore)

addon/-private/system/store.js
```diff
@@ -3438,14 +3433,8 @@ function _modelForMixin(store, normalizedModelName) {
   // container._registry = 1.11 - 2.0
```
rwjblue: These comments should not be needed, as we are explicitly dropping support for Ember < 2.12 here.

Issue #5548, opened 2018-07-29
[3.2.1] Error when doing HasManyReference.reload() for a non-async relationship


YoranBrondsema:
In 3.2.1, I'm seeing an error

`Error: Assertion Failed: You looked up the 'comments' relationship on a 'post' with id 1 but some of the associated records were not loaded. Either make sure they are all loaded together with the parent record, or specify that the relationship is async (`DS.hasMany({ async: true })`)`

I wasn't seeing this in 3.1. I here have a [twiddle failing in 3.2.1](https://ember-twiddle.com/7661280bc7ca1713e17801f0ac9a8975?openFiles=routes.application.js%2C) and the same code [working in 3.1.1](https://ember-twiddle.com/379973ac804a73a27a988249266ec389?openFiles=routes.application.js%2C). It looks like it's caused by the changes in [this commit](https://github.com/emberjs/data/commit/d725553f4f70d00b8bb96ca3112811664b338709#diff-96eb5e2389293df72cc41de71500176f).

All the relationships in my app are async and I use the `value()`/`load()`/`reload()` functions in the `ds-references` API to manually load data when necessary.

urbany:
HI @YoranBrondsema I'm not sure if this is a bug. @runspired is better suited to answer this but here are my thoughts:

Your model is saying `post.comments` is sync. When you retrieve the post it says it has a comment with the id 1, but that comment is not in Ember-Data yet, but if it is sync it has to be there when you reference it. Some possible solutions to apply to your code:
1. it should either be async
2. you should have the comments already in ED or get them at the same time you get the post.
3. your posts API should not give you the IDs of the comments relationship, send only the link. You can do something like this in your serializer `delete data.relationships.comment.data`.

YoranBrondsema:
Hi @urbany, thanks for your reply. Assuming that what you say is correct and that this is intended behavior and not a bug, here are some arguments to support my case of maintaining the behavior of 3.1 and earlier versions.

## Supported by the JSON API
The JSON API specification allows relationship IDs to coexist with links (see http://jsonapi.org/format/#document-resource-object-linkage). Therefore, the JSON that I provide in the twiddles is valid according to the JSON API spec. Now is the question how the two possible values of `async` interpret the JSON.

## `async` is very ill-defined
[...]

PR #5748, opened 2018-11-09
[DOC beta] Add doc for `async: false`

ppcano:
Initially, I could not find the documentation about how `async: false` relationship work. 

I found a few posts about the topic [embermap](https://embermap.com/notes/83-the-case-against-async-relationships) and [discuss](https://discuss.emberjs.com/t/what-is-an-async-relationship-async-true-vs-async-false/4107) and finally #5548 that provides the explanation about the expected behavior.

I have written some initial docs based on the mentioned github issue.

### DIFF
--- a/addon/-private/system/relationships/belongs-to.js
+++ b/addon/-private/system/relationships/belongs-to.js
@@ -85,8 +85,8 @@ import { DEBUG } from '@glimmer/env';
   ```
 
   In contrast to async relationship, accessing a sync relationship
-  will never trigger a request to fetch the resource,
-  and it will always return the existing local resource.
+  will always return the record (Model instance) for the existing
+  local resource, or null.
 
   ```
   let post = comment.get('post');
--- a/addon/-private/system/relationships/belongs-to.js
+++ b/addon/-private/system/relationships/belongs-to.js
@@ -87,7 +87,7 @@ import { DEBUG } from '@glimmer/env';
   In contrast to async relationship, accessing a sync relationship
   will always return the record (Model instance) for the existing
   local resource, or null. But it will error on access when
-  a related resource is known to exist but has not been loaded.
+  a related resource is known to exist and it has not been loaded.
 
   ```
   let post = comment.get('post');
--- a/addon/-private/system/relationships/belongs-to.js
+++ b/addon/-private/system/relationships/belongs-to.js
@@ -11,7 +11,7 @@ import { DEBUG } from '@glimmer/env';
   `DS.belongsTo` takes an optional hash as a second parameter, currently
   supported options are:
 
-  - `async`: A boolean value used to explicitly declare this to be an async relationship.
+  - `async`: A boolean value used to explicitly declare this to be an async relationship. The default is true.
   - `inverse`: A string used to identify the inverse property on a
     related model in a One-To-Many relationship. See [Explicit Inverses](#explicit-inverses)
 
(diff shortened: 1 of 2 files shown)


addon/-private/system/relationships/belongs-to.js
```diff
@@ -68,6 +68,31 @@ import { DEBUG } from '@glimmer/env';
...
+
+  export default DS.Model.extend({
+    post: DS.belongsTo('post', {
+      async: false
+    })
+  });
+  ```
+
+  In contrast to async relationship, accessing a sync relationship
+  will never trigger a request to fetch the resource,
```
runspired: This isn't strictly 100% true, but it is "true enough". There are situations involving `reload` in which we won't error and will fire a request. 

Perhaps a better wording would be `will error on access when a related resource is known to exist but has not been loaded`.

addon/-private/system/relationships/has-many.js
```diff
@@ -105,6 +105,36 @@ import { DEBUG } from '@glimmer/env';
...
+
+  export default DS.Model.extend({
+    comments: DS.hasMany('comment', {
+      async: false
+    })
+  });
+  ```
+
+  In contrast to async relationship, accessing a sync relationship
+  will never trigger a request to fetch the resources,
```
runspired: Similar to the comment on `belongsTo`, this isn't strictly 100% true, but it is "true enough". There are situations involving reload in which we won't error and will fire a request.

Perhaps a better wording would be `will error on access when any of the known related resources have not been loaded.`

addon/-private/system/relationships/has-many.js
```diff
@@ -105,6 +105,36 @@ import { DEBUG } from '@glimmer/env';
...
+  and it will always return an array object with the existing local resources.
```
runspired: It will return a [ManyArray](https://emberjs.com/api/ember-data/3.5/classes/DS.ManyArray), which is not a true array.

addon/-private/system/relationships/belongs-to.js
```diff
@@ -68,6 +68,31 @@ import { DEBUG } from '@glimmer/env';
...
+  and it will always return the existing local resource.
```
runspired: let's be more specific: `it will always return the record (Model instance) for the existing local resource, or null`

ppcano:
@runspired https://github.com/emberjs/data/pull/5748/commits/6fe41b4b0c56d3dcf841f918551eecfce2072372 has some changes based on your review. Let me know if there are something else that should be modified (English is not my strength). 

Thank you.

addon/-private/system/relationships/has-many.js
```diff
@@ -122,8 +122,8 @@ import { DEBUG } from '@glimmer/env';
   ```
 
   In contrast to async relationship, accessing a sync relationship
-  will never trigger a request to fetch the resources,
-  and it will always return an array object with the existing local resources.
+  will always return a [DS.ManyArray](/api/data/classes/DS.ManyArray.html) instance
+  containing the existing local resources, or null.
```
runspired: I don't think this returns `null` if no resources are available locally, but rather the array will be empty.

Similarly to `belongsTo`, I think we should say something about the fact that we error here if you access the relationship prior to the related resources being available (see my original feedback for suggested text).

addon/-private/system/relationships/belongs-to.js
```diff
@@ -85,8 +85,8 @@ import { DEBUG } from '@glimmer/env';
   ```
 
   In contrast to async relationship, accessing a sync relationship
-  will never trigger a request to fetch the resource,
-  and it will always return the existing local resource.
+  will always return the record (Model instance) for the existing
```
runspired: I think we should say something about the fact that we error here if you access the relationship prior to the related resources being available (see my original feedback for suggested text).

igorT:
Thank you! ❤️

PR #5750, opened 2018-11-12
[FEAT RecordIdentifier] Refactors internals to use immutable identifiers

runspired:
**Awesomeness:**

This introduces immutable `RecordIdentifiers` that allow us to

* Use `lid` as a consistent key throughout. Previously we used `type` + `id` for existing records, `type` + `clientId` for new records, `InternalModel` as a unique reference key for store methods and relationships, and `InternalModel[GUID_KEY]` for relationship membership tracking.
* Eliminate a bunch of now redundant store methods
* Eliminate how the store is coupled to `InternalModel` for coalescing / fetching / updating
* Cleanup lots of call signatures where we previously needed to pass `id` `modelName` and `clientId`
* Key things off of these identifiers (like `InternalModel` and eventually `RecordData`) for decoupling
* Drop the double identityMap we were using for `new` vs `canonical`
* [Further discussion on motivations](https://github.com/emberjs/data/pull/5750#discussion_r234894664)
* [Discussion of lookup perf](https://github.com/emberjs/data/pull/5750#discussion_r234901098)

**Next Steps:**
This works is a prerequisite for:
* [RFC level work](https://github.com/emberjs/rfcs/pull/403) to make `RecordIdentifier` a public interface
* RFC level work to allow use of additional index strategies other than `id` and `lid` (by providing us with a consistent `lid`
* Refactoring work to improve the internals of the `relationship-layer` explored [here](https://github.com/emberjs/data/tree/bugfix-rel-errors), much of which we can now land immediately without need for RFC
* RFC level work to make the `relationship-layer` capable of interop between alternative `RecordData` implementations


addon/-private/system/model/record-data.js
```diff
@@ -25,12 +27,13 @@ export default class RecordData {
...
 
   // PUBLIC API
-
+  // TODO IDENTIFIER RFC - RecordData should not be responsible for it's own identifier
   getResourceIdentifier() {
     return {
       id: this.id,
       type: this.modelName,
       clientId: this.clientId,
+      lid: this.clientId,
```
igorT: I don't think we should merge this until an RFC discussion

addon/-private/system/store/record-data-wrapper.js
```diff
@@ -54,44 +65,68 @@ export default class RecordDataWrapper {
...
-  // TODO Igor David cleanup
   inverseIsAsyncForRelationship(modelName, key) {
     let modelClass = this.store.modelFor(modelName);
     return this.relationshipsDefinitionFor(modelName)[key]._inverseIsAsync(this.store, modelClass);
   }
 
-  notifyPropertyChange(modelName, id, clientId, key) {
-    let internalModel = this.store._getInternalModelForId(modelName, id, clientId);
+  // TODO IDENTIFIER RFC - arg should be identifier
+  notifyPropertyChange(type, id, lid, key) {
```
igorT: Not sure if it's worth the churn and the temp object creation

addon/-private/system/store/finders.js
```diff
@@ -52,7 +53,7 @@ export function _find(adapter, store, modelClass, id, internalModel, options) {
         `You requested a record of type '${modelName}' with id '${id}' but the adapter returned a payload with primary data having an id of '${
           payload.data.id
         }'. Use 'store.findRecord()' when the requested id is the same as the one returned by the adapter. In other cases use 'store.queryRecord()' instead https://emberjs.com/api/data/classes/DS.Store.html#method_queryRecord`,
-        payload.data.id === id,
+        coerceId(payload.data.id) === coerceId(id),
```
igorT: We should probably add a test for this

addon/-private/system/store.js
```diff
@@ -1249,25 +1271,23 @@ const Store = Service.extend({
...
+
   // directly get an internal model from ID map if it is there, without doing any
   // processing
-  _getInternalModelForId(modelName, id, clientId) {
-    let internalModel;
-    if (clientId) {
-      internalModel = this._newlyCreatedModelsFor(modelName).get(clientId);
-    }
+  _getInternalModelForId(type, id, lid) {
+    let identifier = recordIdentifierFor(this, { type, id, lid });
```
igorT: We are now doing two lookups per IM vs one, that seems a bit unnecessary to me. I would be excited to talk about what the benefits of the cache per identifier thats stable vs keying off of type/lid are

addon/-private/system/relationships/state/belongs-to.js
```diff
@@ -213,7 +214,7 @@ export default class BelongsToRelationship extends Relationship {
     );
 
     if (recordData !== null) {
-      recordData = this.recordData.storeWrapper.recordDataFor(data.type, data.id);
+      recordData = this.recordData.storeWrapper.recordDataFor(data.type, coerceId(data.id));
```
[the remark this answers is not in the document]
runspired: currently this is very close to public API, something I discovered when I added a `throw` for anytime we received a non-normalized ID.

addon/-private/system/store.js
```diff
@@ -1258,29 +1261,6 @@ const Store = Service.extend({
     return !!internalModel && internalModel.isLoaded();
   },
```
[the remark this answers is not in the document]
runspired: I checked that, the only real usage is from an addon that hasn't been updated in 3 years and I know to not work with master. https://emberobserver.com/code-search?codeQuery=recordForId

addon/-private/system/record-identifier.ts
```diff
@@ -38,20 +45,25 @@ function makeRecordIdentifier(resourceIdentifier: IResourceIdentifier): IRecordI
...
   if (DEBUG) {
     // we enforce immutability in dev
     //  but preserve our ability to do controlled updates to the reference
-    let wrapper: IRecordIdentifier = {
+    let wrapper: IDeprecatedResourceIdentifier = {
       get lid() {
         return recordIdentifier.lid;
       },
+      // TODO deprecate this prop
+      get clientId() {
```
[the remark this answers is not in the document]
runspired: There aren't many benefits to `clientId` given that `lid` will move forward in the spec, but also this has very minimal impact given that an RFC for `storeWrapper` / `relationship` using `identifiers` would necessarily remove the need for this specific arg and 99% of usage today was constrained to variable names in method signatures.

addon/-private/system/store/finders.js
runspired: 👍

addon/-private/system/record-identifier.ts
```diff
@@ -200,6 +194,31 @@ function isNonEmptyString(str?: string | null): str is string {
...
+export function createRecordIdentifier(
+  store: TStore,
+  resourceIdentifier: IResourceIdentifier
+): IRecordIdentifier {
+  let keyOptions = getLookupBucket(store, resourceIdentifier.type);
+  let identifier = makeRecordIdentifier(resourceIdentifier);
+
+  if (identifier.id !== null) {
+    if (DEBUG) {
+      let eid = keyOptions.id[identifier.id];
```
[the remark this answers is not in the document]
runspired: Renamed.

addon/-private/system/store/record-data-wrapper.js
runspired: Follow Up:
---

_Today it:_

* Reduces the complexity of method signatures in the rest of the internals
* Ensures we have normalized `type` `id` `lid`
* Ensures that `type` `id` and `lid` are a valid combo

The following things are valid regardless of whether we make use of the _stable reference_ or move to a less constrained system. Since this is entirely non-user exposed though this shouldn't matter at the moment, and switching storage strategies with the `recordIdentifierFor` API is trivial.

_Tomorrow it:_

* Allows us to RFC `RecordDataWrapper` APIs that accept identifiers instead, which is closer to the `json-api` that `RecordData` users will be working with, especially for relationships.
* Allows us to reduce cognitive overhead by reducing method signatures (`method(identifier, propertyName) {}` is much easier than `method(type, id, lid, propertyName)`)

addon/-private/system/store.js
runspired: previously, to lookup an internalModel, we needed to:

* access `store._identityMap._map[type]._idForModel[id]` (path length 5, two dictionaries and two classed involved)

If we missed, we would fallback to:

* `store._newlyCreated._map[type]._idForModel[clientId]` (path length 5, two dictionaries and two classes involved)

We would then use `internalModel` as a key for many things, including:

* record
* recordData
* network state

----

In the new system, there is never a fallback for newly created records. All lookups take the following journey:

* `<store>.typeMap[type].lid[lid]` (path length 4, dictionaries 2, no classes involved)

In the case wherein this is our first time ever seeing this identifier and/or no `lid` yet exists or was provided, we must fallback to `id` and then generate an `lid` if necessary. This takes an additional lookup:

* `typeCache.id[id]` (path length 2, dictionaries 1, no classes)

There is an optimization here that has not been made yet if we RFC a design choice (if not already part of RecordData RFC?): if `lid` MUST be unique across types, then we can have a single level cache with an optional `type + id` fallback cache.

*  `lidCache[lid]` (path length 1, dictionaries 1, no classes)
* `<store>.typeMap[type].id[id]` (path length 4, dictionaries 2, no classes involved)

Regardless, from here it is a single lookup (path length 1, 1 dictionary) to access `_internalModel`.

**Takeaways**

* The implemented strategy is roughly equivalent for accessing `_internalModel`, but in a common case does *slightly* regress.
* There exists an optimization for this strategy that leads us to doing *slightly* better instead.

So where, other than improving method signatures and a clearer separation of concerns internally (identifier now stores identity info instead of internalModel), is the big win / reason to do this?

**It's in the very next lookup!**

Right now, `InternalModel` is the key for `RecordData` and `Record`, but this cannot continue to be with the landing of the `modelFactoryFor` RFC, in which `recordData` is handed directly to the `record` and no `InternalModel` is. While we could continue to spin up `InternalModel`s just to serve as the cache for these things, it seems better to instead create some clear boundaries. Using `Identifier` as the key instead, we would then create the following mappings, reducing overall lookups and _reducing the memory footprint by moving from deeply nested object graphs to WeakMaps [1]:_

* `Identifier => RecordData`
* `Record  => RecordData`
* `Identifier => Record`

In `DEBUG` mode, we would symbolicate these object relationships and provide access to the `Identifier` map via a symbol on `Store` so that the entire data store would be discoverable and debuggable in the console without leaking the APIs for abuse.

[1] _keying by `lid` or by `type + id` creates higher memory stress due to the way memory blocks are allocated for the objects involved. As the cache grows, especially as a 2+ level cache grows, this becomes slower and much less ideal. Keying via a WeakMap has roughly identical lookup speed to a dictionary but without this memory overhead_

addon/-private/system/model/record-data.js
runspired: note that `lid` and `clientId` are both present here, sans deprecation. Users that have already implemented custom `RecordData`s returning only `clientId` will work fully as expected, but our default will now provide both. I'll be providing an RFC for identifiers for you to review over Thanksgiving break.

addon/-private/system/record-identifier.ts
```diff
@@ -0,0 +1,238 @@
...
+
+let CLIENT_ID = 0;
+let DEBUG_MAP;
+
+if (DEBUG) {
+  DEBUG_MAP = new WeakMap<IRecordIdentifier, IRecordIdentifier>();
+}
+
+function generateLid(): string {
+  return `@ember-data:lid-${CLIENT_ID++}`;
```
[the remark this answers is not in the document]
dgeb: An auto-incrementing integer is not very compatible with offline strategies because it's prone to collisions. If these IDs are all that's used to identify client-side records, then they will need to be persisted to browser storage, and reified along with those records. A new browser instance could not safely restart with `CLIENT_ID = 0`. 

I think a UUID is preferable. A common prefix/namespace is fine though.

runspired:
Closing in favor of #5914

PR #5754, opened 2018-11-20
Ui layer cleanup

runspired:


addon/-private/attr.ts
```diff
@@ -1,12 +1,13 @@
...
 import { DEBUG } from '@glimmer/env';
-import recordDataFor from './system/record-data-for';
+import Model from './system/model/model';
+import { getRecordDataFor } from './system/model/record-data-map';
 /**
   @module ember-data
 */
 
-function getDefaultValue(record, options, key) {
+function getDefaultValue(options) {
```
runspired: I thought I could drop `record` and `key` from these `arguments` but I realize now we use them when we call `options.defaultValue.apply(null, arguments)`. This fails no test though so gonna need to add one.

runspired: It actually did fail a test, I just didn't notice. :(

addon/-private/system/model/record-data-map.ts
```diff
@@ -0,0 +1,54 @@
+import { DEBUG } from '@glimmer/env';
+
+export const RECORD_DATA_KEY = `${Date.now()}-record-data`;
```
igorT: Do we want to merge this map before we get rid of the internal model mapping? Is it easy to measure the perf impact?

addon/-private/system/model/record-data-map.ts
```diff
@@ -0,0 +1,54 @@
...
+  getAttr(key);
+  setDirtyAttribute(key, value);
+  isAttrDirty(key);
+};
+
+type Mappable = InternalModel | CreateOptions | Record;
+
+const RecordDataMap = new WeakMap<Mappable, RecordData>();
+
+export function getRecordDataFor(instance: Mappable): RecordData {
```
igorT: Can you add some comments to the file please, what the purpose is and how to use it

addon/-private/system/model/record-data-map.ts
igorT: What the thinking behind making `RECORD_DATA_KEY` time dependent?

addon/-private/system/model/model.js
```diff
@@ -1195,8 +1195,8 @@ if (DEBUG) {
   };
 
   Model.reopen({
-    init() {
-      this._super(...arguments);
+    init(createArgs) {
+      this._super(createArgs);
```
igorT: `...arguments` should be more robust to refactorings, given that we are reopening

addon/-private/attr.ts
```diff
@@ -135,22 +132,51 @@ export default function attr(type?: string | AttrOptions, options?: AttrOptions)
...
+            `Attempted to set '${key}' to '${value}' on the deleted record ${internalModel}`
+          );
+        }
+
+        let currentValue = recordData.getAttr(key);
+        if (currentValue !== value) {
+          recordData.setDirtyAttribute(key, value);
+          let isDirty = recordData.isAttrDirty(key);
+
+          internalModel.send('didSetProperty', {
```
igorT: Wonder how hard it would be to get rid of this.

addon/-private/attr.ts
```diff
@@ -135,22 +132,51 @@ export default function attr(type?: string | AttrOptions, options?: AttrOptions)
...
+        ) {
           throw new Error(
             `'${key}' is a reserved property name on instances of classes extending Model. Please choose a different property name for your attr on ${this.constructor.toString()}`
           );
         }
       }
-      return this._internalModel.setDirtyAttribute(key, value);
+      let recordData = getRecordDataFor(this);
+
+      if (this instanceof Model) {
```
igorT: This branch seems long/complex enough to extract into a function

addon/-private/system/model/internal-model.js
```diff
@@ -284,7 +286,16 @@ export default class InternalModel {
         createOptions.container = store.container;
       }
 
+      let recordData = this._recordData;
+      createOptions[RECORD_DATA_KEY] = recordData;
       this._record = store._modelFactoryFor(this.modelName).create(createOptions);
+      // TODO should we unmap createOptions after record construction?
+      // mapping of record to recordData is handled by `Model` itself.
+      // once the `modelFactoryFor` RFC implementation lands, it will be done
```
igorT: This comment is no longer correct

addon/-private/system/model/record-data-map.ts
```diff
@@ -0,0 +1,53 @@
+import { DEBUG } from '@glimmer/env';
+
+export const RECORD_DATA_KEY = `${Date.now()}-record-data`;
```
igorT: Can you please write a comment of why/how we need the `RECORD_DATA_KEY` for

addon/-private/system/model/record-data-map.ts
```diff
@@ -0,0 +1,53 @@
...
+
+// TODO is there a way to import classes without causing cyclical imports for rollup?
+//   maybe strip "typing" imports prior to rollup?
```
igorT: I'm not quite sure what the TODO here is?

addon/-private/system/model/model.js
```diff
@@ -1213,6 +1213,9 @@ if (DEBUG) {
         );
       }
 
+      let recordData = getRecordDataFor(createArgs);
+      setRecordDataFor(this, recordData);
```
igorT: Can you write a comment as to why we can't do this in IM please

addon/-private/system/model/internal-model.js
```diff
@@ -284,7 +286,16 @@ export default class InternalModel {
...
+      //   here for records that do not extend `DS.Model`.
+      if (!this._record instanceof Model) {
```
igorT: This divergent path still sketches me out :(

addon/-private/system/model/record-data-map.ts
runspired: This is for @mike-north, basically in order to use classes as "types" we import them, but rollup doesn't like cyclical imports and sometimes this causes cyclical imports.

addon/-private/system/model/record-data-map.ts
runspired: 👍

addon/-private/attr.ts
runspired: 👍

addon/-private/attr.ts
runspired: We deprecated the event, so once we land the deprecation work the answer is "not hard"

addon/-private/system/model/internal-model.js
```diff
@@ -284,7 +286,15 @@ export default class InternalModel {
         createOptions.container = store.container;
       }
 
+      let recordData = this._recordData;
```
mike-north: I don't see anywhere that `recordData` has the potential to be reassigned. Does ember-data have a policy on `let` vs `const`? 

There are advantages in TS to using const, in that guards can effectively narrow types, since no branch of the code could alter the type of the variable through reassignment.

addon/-private/system/model/model.js
mike-north: I agree w/ @igorT here -- this is a case where you don't want to mix a TS conversion with potential behavioral changes to the code. If it's desirable to move away from `_super(...arguments)`, I'd advise making that change in another PR

addon/-private/system/model/record-data-map.ts
mike-north: > There isn’t one

@runspired how do we know this?

addon/-private/system/model/record-data-map.ts
```diff
@@ -0,0 +1,63 @@
...
+ * as the first arg to `Model.init()` (in fact, they are not!).
+ *
+ * For this reason, we use this randomly generated string key to store
+ * the mapping for CreateOptions. This maintains isolation without
+ * resulting in a performance hit.
+ */
+export const RECORD_DATA_KEY = `${Date.now()}-record-data`;
+
+type CreateOptions = object;
+type Record = object;
```
mike-north: There's not much value in creating types like these. Some people have the urge to do this because it makes the code more descriptive, but that's what variable/parameter names are for.

runspired: It was more to make refactoring easier once we have true interfaces for these things.

addon/-private/system/model/internal-model.js
runspired: @igort thoughts? Overall we’ve preferred let to const until now, and for JS code we’ve definitely felt there was more signal when doing so.

igorT: Yeah, we use `const` only for constants.

addon/-private/system/model/record-data-map.ts
mike-north: I would add a `// TODO: create a specific interface` to each site then. In general, if you just create aliases to `object` all over the place, you're introducing unnecessary complexity and the _illusion_ of type-safety.

mike-north:
@runspired 	

> @mike-north RE isolating TS changes do you feel that way for new modules / edits to already converted TS files?

Particularly as contributors are ramping up on TS, it's a good idea to either change the behavior of code _or_ typings in a single PR. I feel much more strongly about this when applied to the initial conversion, but it applies to refinements as well.

PR #5755, opened 2018-11-21
[BUGFIX Model] assert when 'content' is used as a property on a record

### LINKED ISSUE #5732: "content" field doesn't work through async relationships
(This is based off a [forum question](https://discuss.emberjs.com/t/ember-handlebar-view-for-foreignkey-values/15755).)

Given these fields:

```js
// app/models/answer.js
question: belongsTo('question'),

// app/models/question.js
content: attr('string'),
```

If you try to render the `content` field through the async `question` relationship, it doesn't work.

```hbs
{{someAnswer.question.content}}
```

The expected output is the value of the `content` field. The actual output is `<frontend@model:question::ember276:5>`.

The PromiseProxy's own `content` property takes precedence.

I don't think we can easily fix this without breaking other things, but we should consider at least linting for the `content` field name in models.

runspired:
resolves #5732

### DIFF
--- a/tests/unit/model-test.js
+++ b/tests/unit/model-test.js
@@ -688,8 +688,15 @@ module('unit/model - Model', function(hooks) {
       @attr('string')
       name;
     }
+    class NativePostWithContent extends Model {
+      @attr('string')
+      content;
+      @attr('string')
+      name;
+    }
     const PROP_MAP = {
       _internalModel: NativePostWithInternalModel,
+      content: NativePostWithContent,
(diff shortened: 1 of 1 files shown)


jelhan:
There was a StackOverflow question caused by running into strange behavior due to using `content` as attribute name: https://stackoverflow.com/questions/54812019/emberjs-json-api-errors-for-one-field-are-included-in-other-fields-errors Would be good to get this one merged in.

PR #5757, opened 2018-11-22
Add *.ts to `.npmignore`

igorT:
Don't publish .ts files in order to not make them public api until
we are confident in our typings.


igorT:
This broke `ember-cli-addon-tests` https://github.com/tomdale/ember-cli-addon-tests/issues/199

Issue #5752, opened 2018-11-16
Class constructor EmberError cannot be invoked without 'new'


ssutar:
When used with `ember-native-class-polyfill` for ES6 classes, following error is thrown: 
```
Uncaught TypeError: Class constructor EmberError cannot be invoked without 'new'
    at new AdapterError (vendor.js:132445)
    at Class.handleResponse (vendor.js:144384)
    at Class.handleResponse (vendor.js:122856)
    at Class.superWrapper (vendor-static.js:51234)
    at Class.handleResponse (vendor.js:89895)
    at Class.superWrapper [as handleResponse] (vendor-static.js:51234)
    at Class.handleResponse (vendor.js:119158)
    at Class.superWrapper [as handleResponse] (vendor-static.js:51234)
    at ajaxError (vendor.js:144655)
    at ajaxErrorHandler (vendor.js:144682)
```
It is originated from https://github.com/emberjs/data/blob/master/addon/-private/adapters/errors.js#L79
```
export function AdapterError(errors, message = 'Adapter operation failed') {
  this.isAdapterError = true;
  EmberError.call(this, message);

  this.errors = errors || [
    {
      title: 'Adapter Error',
      detail: message,
    },
  ];
}
```
As `EmberError` is a class with `ember-native-class-polyfill`, it cannot be called without `new`.

ssutar:
I have a fix at https://github.com/ssutar/data/blob/v_3_5_beta/addon/-private/adapters/errors.js#L79 but its not same as the current code. 

Its something like this:

```
export function AdapterError(errors, message = 'Adapter operation failed') {
  this.isAdapterError = true;
  this.message = message;

  this.errors = errors || [
    {
      title: 'Adapter Error',
      detail: message,
    },
  ];
}
```

Its creating a new property `message` on the AdapterError instead of using one from `EmberError`. It works correctly due to prototype property lookup, but not sure if it is the correct fix.

cc: @rwjblue @krisselden

pzuraq:
This depends on usage, but I think this is meant to be a class, and if so the accurate update would be:

```js
export class AdapterError extends EmberError {
  constructor(errors, message = 'Adapter operation failed') {
    super(message);
    this.isAdapterError = true;

    this.errors = errors || [
      {
        title: 'Adapter Error',
        detail: message,
      },
    ];
  }
}

ssutar:
[...]

PR #5759, opened 2018-11-27
[BUGFIX CHORE] Change the subclassing in adapter error

ssutar:
Fix for: https://github.com/emberjs/data/issues/5752


runspired:
We resolved this by reverting the change from function to class based definition of `EmberError` upstream.

PR #5761, opened 2018-11-28
Revert "Add *.ts to `.npmignore`"

igorT:
Reverts emberjs/data#5757

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -33,9 +33,3 @@ yarn.lock
 **/*.rb
 node-tests/
 lib/version-replace.js
-
-# typescript
-#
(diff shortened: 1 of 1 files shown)


igorT:
The original PR seems to have just not shipped the `.ts` files which broke the build.

runspired:
@rwjblue we have to ship `.ts` files or folks wouldn't be able to build, I think what we want is to not ship the `.d.ts` files.

rwjblue:
@runspired - I am aware, this is why I suggested (somewhere?) that we need to decide if we support directly using the GH branch or not. ember-source (for example) publishes tarballs to ensure things are assembled properly...

PR #5767, opened 2018-11-30
Make _recordData lazy

hjdivad:
In particular, do not instantiate them for `findRecord`

### DIFF
--- a/addon/-private/system/model/internal-model.js
+++ b/addon/-private/system/model/internal-model.js
@@ -93,7 +93,7 @@ export default class InternalModel {
     this.modelName = modelName;
     this.clientId = clientId;
 
-    this._recordData = store._createRecordData(modelName, id, clientId, this);
+    this.__recordData = null;
 
     // this ensure ordered set can quickly identify this as unique
     this[Ember.GUID_KEY] = InternalModelReferenceId++ + 'internal-model';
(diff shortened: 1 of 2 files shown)


addon/-private/system/snapshot.js
```diff
@@ -65,7 +65,12 @@ export default class Snapshot {
      */
     this.modelName = internalModel.modelName;
 
-    this._changedAttributes = internalModel.changedAttributes();
+    if (internalModel.isLoading()) {
```
igorT: It seems like this logic belongs in `internalModel`.

hjdivad: @igorT seems fine; you'd prefer to push this into `internalModel.changedAttributes()`?

runspired: I am skeptical. What about `reload` ?

hjdivad: @runspired what's your preferred branch?  the state here is `root.loading`

igorT:
Empty state should work

hjdivad:
@igorT no, `root.loading` is not empty

hjdivad:
we could change `root.loading` to `isEmpty: true` but i am far more wary of that than not eagerly calculating changed attributes during record reload

hjdivad:
@igorT never mind, @runspired pointed me to the `isReloading` flag within internal model.

addon/-private/system/snapshot.js
hjdivad: I don't think it is though --- if you have an internal model for a `findRecord` (ie where `isLoading() && !isReloading`), what does it even mean conceptually to ask for changed attributes?

runspired: If I reorder that you may see what I mean a little bit better: "logic specific to requests"

PR #5768, opened 2018-11-30
Bug: findRecord does not always properly honor reload: true

hjdivad:
See failing test in this pr.

---

The issue is that when requesting a record with some id (eg `A`) and getting a returned id (eg `B`), the store will use the internal model for `A`, whose `_promiseProxy` will fulfill to a different internal model for `B`.


hjdivad:
cc @runspired @igorT

runspired:
closing in favor of #6171

PR #5773, opened 2018-12-05
Implicit null inverse test.

### LINKED ISSUE #5538: Invalid relationship inverse configuration not asserted (implicit null inverse)
The following inverse configuration should error because there is no inverse relationship definition, but does not:

```
class User extends Model {
  @attr name;
  @belongsTo('pet')
  pet;
}

class Pet extends Model {
  @attr name;
}
```

Similar issue that is equally good for new contributors https://github.com/emberjs/data/issues/5435

michielboekhoff:
This PR adds a test to verify that a missing `inverse` on a one-sided association results in the same situation as having an `inverse: null`. Closes #5538.

Two questions; in the original issue, it was mentioned this behaviour ought to be deprecated. If so, where in the docs would you like to indicate this? Then I'll make that change, too. Finally, is there any reason `"experimentalDecorators": true` isn't set in `tsconfig.json`? The decorators are causing VSCode to report code everywhere. If this is something you'd like to see, I'd be more than willing to add it.

### DIFF
--- a/tests/integration/relationships/inverse-relationships-test.js
+++ b/tests/integration/relationships/inverse-relationships-test.js
@@ -1,155 +1,172 @@
-import { run } from '@ember/runloop';
-import { createStore } from 'dummy/tests/helpers/store';
-import setupStore from 'dummy/tests/helpers/store';
-import testInDebug from 'dummy/tests/helpers/test-in-debug';
 import { module, test } from 'qunit';
+import { belongsTo, hasMany, attr } from '@ember-decorators/data';
+import { setupTest } from 'ember-qunit';
+import testInDebug from 'dummy/tests/helpers/test-in-debug';
+import Model from 'ember-data/model';
 
-import DS from 'ember-data';
+module('integration/relationships/inverse_relationships - Inverse Relationships', function(hooks) {
+  setupTest(hooks);
 
-let { Model, hasMany, belongsTo } = DS;
+  let store;
+  let register;
 
-var Post, Comment, Message, User;
+  hooks.beforeEach(function() {
+    const { owner } = this;
 
-module('integration/relationships/inverse_relationships - Inverse Relationships');
-
-test('When a record is added to a has-many relationship, the inverse belongsTo is determined automatically', function(assert) {
-  Post = DS.Model.extend({
-    comments: DS.hasMany('comment', { async: false }),
+    store = owner.lookup('service:store');
+    register = owner.register.bind(owner);
   });
 
-  Comment = DS.Model.extend({
-    post: DS.belongsTo('post', { async: false }),
-  });
+  test('When a record is added to a has-many relationship, the inverse belongsTo is determined automatically', async function(assert) {
+    class Post extends Model {
+      @hasMany('comment', { async: false })
+      comments;
+    }
 
-  var env = setupStore({ post: Post, comment: Comment });
-  var store = env.store;
+    class Comment extends Model {
+      @belongsTo('post', { async: false })
+      post;
+    }
 
-  let comment = store.createRecord('comment');
-  let post = store.createRecord('post');
+    register('model:Post', Post);
+    register('model:Comment', Comment);
 
-  assert.equal(comment.get('post'), null, 'no post has been set on the comment');
+    const comment = store.createRecord('comment');
+    const post = store.createRecord('post');
 
-  run(function() {
+    assert.equal(comment.get('post'), null, 'no post has been set on the comment');
+
     post.get('comments').pushObject(comment);
-  });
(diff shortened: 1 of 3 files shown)


michielboekhoff:
@runspired - any chance you could have a look at this?

tests/integration/relationships/inverse-relationships-test.js
```diff
@@ -1,641 +1,747 @@
...
-  Post = DS.Model.extend({
-    comments: DS.hasMany('comment', { async: false }),
+    store = owner.lookup('service:store');
+    register = owner.register.bind(owner);
   });
 
-  Comment = DS.Model.extend({
-    post: DS.belongsTo('post', { async: false }),
-  });
+  test('09, the inverse belongsTo is determined automatically', function(assert) {
```
runspired: what is the `09, ` here for?

tests/integration/relationships/inverse-relationships-test.js
```diff
@@ -1,641 +1,747 @@
...
 
-  run(function() {
-    comment = store.createRecord('comment');
-    post = store.createRecord('post');
+    register('model:Post', Post);
+    register('model:Comment', Comment);
+
+    let comment, post;
+
+    run(function() {
```
runspired: if we make this test `async` we will not need this `run`

tests/integration/relationships/inverse-relationships-test.js
```diff
@@ -1,641 +1,747 @@
...
+    register('model:Comment', Comment);
 
-  Comment = DS.Model.extend({
-    post: DS.belongsTo('post', { async: false }),
-  });
+    const comment = store.createRecord('comment');
+    const post = store.createRecord('post');
+    const post2 = store.createRecord('post');
+
+    run(function() {
```
runspired: if we make this test async we will not need this run

tests/integration/relationships/inverse-relationships-test.js
```diff
@@ -1,641 +1,747 @@
...
-  var env = setupStore({ post: Post, comment: Comment });
-  var store = env.store;
+    assert.equal(comment.get('post'), post);
+    assert.equal(post.get('bestComment'), comment);
+    assert.strictEqual(post2.get('bestComment'), null);
 
-  let comment = store.createRecord('comment');
-  let post = store.createRecord('post');
-  let post2 = store.createRecord('post');
+    run(function() {
```
runspired: make async test = no need for run

tests/integration/relationships/inverse-relationships-test.js
```diff
@@ -1,641 +1,747 @@
...
-    bestComment: DS.belongsTo('comment', { async: false }),
-  });
+    const comment = store.createRecord('comment');
+    const post = store.createRecord('post');
+    const post2 = store.createRecord('post');
 
-  Comment = DS.Model.extend({
-    post: DS.belongsTo('post', { async: false }),
-  });
+    run(function() {
```
runspired: make async test = no need for run

tests/integration/relationships/inverse-relationships-test.js
```diff
@@ -1,641 +1,747 @@
...
-    comment: Comment,
-  });
+    assert.equal(comment.get('post'), post);
+    assert.equal(post.get('bestComment'), comment);
+    assert.strictEqual(post2.get('bestComment'), null);
 
-  let comment = store.createRecord('comment');
-  let post = store.createRecord('post');
-  let post2 = store.createRecord('post');
+    run(function() {
```
runspired: make async test = no need for run

tests/integration/relationships/inverse-relationships-test.js
```diff
@@ -1,641 +1,747 @@
...
-    bestComment: DS.belongsTo('comment', { async: false }),
-  });
+    const post = store.createRecord('post');
+    const comment = store.createRecord('comment');
+    const comment2 = store.createRecord('comment');
 
-  Comment = DS.Model.extend({
-    post: DS.belongsTo('post', { async: false }),
-  });
+    run(function() {
```
runspired: make async test = no need for run

michielboekhoff:
Thanks for the code review @runspired! Any feedback on my other two points?

runspired:
@michielboekhoff rebased this for us, I gave it a review and it seems close to landing but possibly introduced a regression in `2.18` which I will check out.

RE the other two points:

> in the original issue, it was mentioned this behaviour ought to be deprecated. If so, where in the docs would you like to indicate this? Then I'll make that change, too.

This would require an RFC.

> Finally, is there any reason "experimentalDecorators": true isn't set in tsconfig.json?

I don't experience this, so I am not sure why you do. I'll investigate that further.

Issue #3803, opened 2015-09-28
No way to specify a Model attribute as 'Read Only'?


christophermlne:
It would be useful to be able to do something like this to prevent an attribute from being saved back to the server.

```'DS.attr('string', { readOnly: true })```

related Stack Overflow question: http://stackoverflow.com/questions/32829966/how-to-prevent-ember-data-from-saving-attribute-ie-read-only-attribute

wecc:
You should be able to use the `attrs` property in your serializer and set `serialize: false` to prevent data from being serialized back to your server.

See http://emberjs.com/api/data/classes/DS.JSONSerializer.html#property_attrs

Would that work?

bmac:
Hi @christophermlne. I'm going to close this issue. Setting `serialize: false` on the `attrs` object in the serializer for your model is the supported way to prevent an attribute from being saved back to the server.

PR #5801, opened 2018-12-13
Allow to define readOnly attributes and relationships in models definition

Exelord:
# The Issue
Many times I encounter bugs when developers forgot to mark the model's attributes/relationships in serializer as `serialize: false` (readOnly) during models creation based and schema definition. They even very often put comments like:
```js
name: attr('string') // readOnly
```
but forgot to disable serialization what caused API errors.

Also, it happened that developers were setting such properties eg in a form while they were excluded from serialization, which caused a question "Why this attribute was not sent to API?" or "Why this attribute changed after save?"

# Solution
I found it much simpler and convenient way to mark attrs/rels as `readOnly` in model than creating a new serializer to just mark one property as `serialize: false` in `attrs` definition. What's more, I disabled setters and raise a proper error for all `readOnly` properties, so the user will know that this property is not meant to be changed.

# Usage
## Before
```js
export default Model.extend({
  title: attr('string'), // readOnly
  author: belongsTo('user'), // readOnly
  comments: hasMany('comment') // readOnly
});

export default Serializer.extend({
  attr: {
    title: { serialize: false }
    author: { serialize: false }
    tasks: { serialize: false }
  }
});
```
**However it still didn't prevent the `set` on property**

## After
```js
export default Model.extend({
  title: attr('string', { readOnly: true })
  author: belongsTo('user', { readOnly: true })
  comments: hasMany('comment', { readOnly: true })
});
```

And in case the user will set the `readOnly` attribute, it will raise the following error:
`Assertion Failed: Cannot set read-only attribute "title" on instance of model "post"`

Related issue: https://github.com/emberjs/data/issues/3803

### DIFF
--- a/addon/-private/attr.ts
+++ b/addon/-private/attr.ts
@@ -25,6 +25,7 @@ function hasValue(internalModel, key) {
 
 interface AttrOptions {
   defaultValue?: string | null | (() => any);
+  readOnly?: boolean;
 }
 
 /**
@@ -69,6 +70,19 @@ interface AttrOptions {
         return {};
       }
     })
+  });
+  ```
+
+  - `readOnly`: A boolean value that determines if the attribute should be read only and serialized in a payload.
+
+  Example
+
+  ```app/models/user.js
+  import DS from 'ember-data';
+
+  export default DS.Model.extend({
+    username: DS.attr('string'),
+    isAdmin: DS.attr('string', { readOnly: true }),
   });
   ```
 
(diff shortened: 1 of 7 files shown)


runspired:
@Exelord thanks for exploring this feature!

adding this would require an RFC; however, we are also working on an RFC for custom record classes which would allow you to define these properties yourself quite easily. You could even do the exact syntax you have here with `DS.Model` today but implementing a custom `RecordData`.

I'm going to close this because as I mentioned adding this feature would require an RFC and there are non-RFC non-feature-add alternatives, that said, we are working towards having more of this sort of validation as a default, and both `immutable` and `read-only API props` are things that we want to make it simpler to use.

PR #5805, opened 2018-12-22
Preserve hasmany order

BryanCrotaz:
In my app I have a websocket receiving jsonapi updates from the server. I use `store.push` to put these into ember-data.

If I push a model which is part of an ordered hasMany, it gets bumped to the end of the list.

I can't reproduce this in a test, although I've tried to in this PR. My fix, however, works in the app. It is also a nice optimisation.


BryanCrotaz:
resubmitting with a clean repo

PR #5806, opened 2018-12-22
Preserve hasmany order when pushing child

BryanCrotaz:
In my app I have a websocket receiving jsonapi updates from the server. I use store.push to put these into ember-data.

If I push a model which is part of an ordered hasMany, it gets bumped to the end of the list.

I can't reproduce this in a test, although I've tried to in this PR. My fix, however, works in the app. It is also a nice optimisation when pushing a hasMany with very few changes. Previously every item was removed and then added again. I skip that if the item in the old and new array matches at this index.

### DIFF
--- a/addon/-private/system/relationships/state/belongs-to.js
+++ b/addon/-private/system/relationships/state/belongs-to.js
@@ -24,10 +24,12 @@ export default class BelongsToRelationship extends Relationship {
   }
 
   setCanonicalRecordData(recordData) {
-    if (recordData) {
-      this.addCanonicalRecordData(recordData);
-    } else if (this.canonicalState) {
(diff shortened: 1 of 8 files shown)


BryanCrotaz:
@runspired do you know why this PR is failing with the good old runloop autorun in testing error but only in LTS 2.18?

BryanCrotaz:
@runspired this is an important fix for us and has left us marooned on 3.6. Do you have a moment to see why it's only failing on the old LTS?

BryanCrotaz:
closed in favour of #7402

runspired:
Closing since stale (and rebooted in #7402), likely mostly addressed with #7493

Issue #5817, opened 2019-01-15
use `assign` instead of `merge` here ?


bartocc:
I might be wrong, but shouldn't we use `Ember.assign` instead of `Ember.merge` here to avoid a 3.6 deprecation warning ?

https://github.com/emberjs/data/blob/4c539f466a21e91499ff18ba0318827db1f27252/addon/-private/system/store/finders.js#L11

mhankus:
Deprecation warnings are also caused by (in 3.7)
[data/addon/-private/system/model/errors.js](https://github.com/emberjs/data/blob/v3.7.0/addon/-private/system/model/errors.js)
```
errorsFor(attribute) {
    let map = get(this, 'errorsByAttributeName');

    if (!map.has(attribute)) {
      map.set(attribute, new A());
    }

    return map.get(attribute);
  },
```
https://deprecations-app-prod.herokuapp.com/deprecations/v3.x/#toc_array-new-array-wrapper

rwjblue:
Looks like https://github.com/emberjs/data/pull/5812/files is already open for fixing the `Ember.merge` deprecation.

igorT:
This was fixed by #5812, thx @BryanCrotaz

PR #5818, opened 2019-01-19
[BUGFIX deprecation] stop calling A as constructor - partial fix for #5817

mhankus:
Issue  #5817 mentions two cases of deprecation warnings, changing usage of merge is being fixed in #5812, but the second case needs fixing, so here is a quick fix for it


mhankus:
I'm not sure if I should do something about failed checks (same checks fail at master), if so please let me know

rwjblue:
I haven't really poked at what is failing on master at the moment.

Perhaps @runspired has a better idea/

raido:
I just ran into this and nice to see PR already open for this.

Issue #5821, opened 2019-01-25
belongsTo resolves to undefined in 3.6.0


BryanCrotaz:
Comment adapter returns:
```
{
  data: {
    type: 'comment',
    id: 2,
    relationships: {
      post: {
        data: {
          type: 'post',
          id: '1',
        },
      },
    },
  }
}
```

```
store.findRecord('comment', 2)
.then(comment => comment.get('post'));
```
should return a promise resolving to the post.
Instead it returns a promise resolving to `undefined`.

BryanCrotaz:
@runspired this seems like a very serious regression. Is my test in the PR valid? It seems so to me.

runspired:
The test in the PR wasn't correct, though I'd love to see a test for this as I suspect it relates to/duplicates #5506

runspired:
did some more digging and I think the bug here was an issue with proxy and relationship changes made as we converted to RecordData internally. These issues have since been resolved, and since the reported issue and test cases if broken would be breaking basically every app in existence, I suspect this was a transient issue at the time that we fixed but failed to recognize was associated with this PR.

PR #5822, opened 2019-01-25
belongsTo resolves to undefined [wip]

BryanCrotaz:
Demonstrates #5821 

belongsTo resolves to undefined

### DIFF
--- a/tests/integration/relationships/belongs-to-test.js
+++ b/tests/integration/relationships/belongs-to-test.js
@@ -587,6 +587,53 @@ test('The store can load a polymorphic belongsTo association', function(assert) 
   });
 });
 
+test('The store can load a polymorphic belongsTo association via json-api', function(assert) {
+  env.registry.register(
+    'adapter:post',
+    DS.JSONAPIAdapter.extend({
+      findRecord(store, type, id) {
+        return resolve({
+          data: {
+            type: 'post',
+            id,
+          },
+        });
+      },
+    })
+  );
+
+  env.registry.register(
+    'adapter:comment',
+    DS.JSONAPIAdapter.extend({
+      findRecord(store, type, id) {
+        return resolve({
+          data: {
+            type: 'comment',
+            id,
+            relationships: {
+              post: {
+                data: {
+                  type: 'post',
+                  id: '1',
+                },
+              },
+            },
+          },
+        });
+      },
+    })
+  );
+  return run(() => {
+    return store
+      .findRecord('comment', 1)
+      .then(comment => comment.get('post'))
+      .then(post => {
+        assert.ok(post);
+        assert.equal(post.id, 1);
+      });
+  });
+});
+
 test('The store can serialize a polymorphic belongsTo association', function(assert) {
   env.adapter.shouldBackgroundReloadRecord = () => false;
   let serializerInstance = store.serializerFor('comment');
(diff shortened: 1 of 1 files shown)


BryanCrotaz:
@runspired I can't see why one test fails but the others all pass

tests/integration/relationships/belongs-to-test.js
```diff
@@ -211,8 +211,8 @@ module('integration/relationship/belongs_to Belongs-To Relationships', {
     });
 
     Comment = Message.extend({
-      body: DS.attr('string'),
-      message: DS.belongsTo('message', { polymorphic: true, async: false, inverse: null }),
+      body: DSattr('string'),
+      message: DSbelongsTo('message', { polymorphic: true, async: false, inverse: null }),
```
makepanic: shouldn't this be `DS.attr` and `DS.belongsTo`?

mikeelemuel: Agree.

BryanCrotaz: They're aliased at the top of the file - I've copied the style of the other tests

makepanic: You're right. Sorry I didn't see that.

tests/integration/relationships/belongs-to-test.js
```diff
@@ -587,6 +587,524 @@ test('The store can load a polymorphic belongsTo association', function(assert)
...
+  }
+
+  env.registry.register('model:person', Person);
+  env.registry.register('model:vetenarian', Vetenarian);
+  env.registry.register('model:pet', Pet);
+
+  env.registry.register(
+    'adapter:vetenarian',
+    DS.JSONAPIAdapter.extend({
+      findRecord(store, type, id) {
```
runspired: when links are fetched for a `belongsTo` relationship, `findBelongsTo` is used instead of `findRecord`

tests/integration/relationships/belongs-to-test.js
```diff
@@ -587,6 +587,524 @@ test('The store can load a polymorphic belongsTo association', function(assert)
...
+    'adapter:pet',
+    DS.JSONAPIAdapter.extend({
+      findRecord(store, type, id) {
+        return resolve({
+          data: {
+            type: 'pet',
+            id,
+            relationships: {
+              human: {
+                link: {
```
runspired: this is not the correct format for links, which will lead this relationship to being considered empty

tests/integration/relationships/belongs-to-test.js
```diff
@@ -587,6 +587,524 @@ test('The store can load a polymorphic belongsTo association', function(assert)
...
+    DS.JSONAPIAdapter.extend({
+      findRecord(store, type, id) {
+        return resolve({
+          data: {
+            type: 'comment',
+            id: '2',
+            relationships: {
+              message: {
+                data: {
+                  type: 'post',
```
sly7-7, line 1086: @runspired @BryanCrotaz I don't know if those tests are still somewhat valid, but here, I think there is a mistake (I've not read the whole file), the type seems completely different from the relationship name. Or maybe it's something intented, and defined as this in the model

runspired:
Closing as I'm fairly certain all these cases are covered by existing tests. I suspect the underlying issue here was a bug with some proxy changes in 3.6 that was long-since fixed. Happy to revisit if anyone has a similar issue.

Issue #3084, opened 2015-05-20
unloadingRecord w/ relationships fail in flushCanonical


sdahlbac:
The following models:

service-report.coffee
  serviceReportLines: DS.hasMany('serviceReportLine', async: true)

service-report-line.coffee
  task: DS.belongsTo('task', async: true)
  serviceReport: DS.belongsTo('serviceReport', async: true)
  usedSpareParts: DS.hasMany('usedSparePart', async: true)
  timeEntries: DS.hasMany('timeEntry', async: true)

I have the following serializers:

    app.ServiceReportSerializer = DS.RESTSerializer.extend(
      DS.EmbeddedRecordsMixin,
      attrs:
        serviceReportLines: embedded: "always"
    )

    app.ServiceReportLineSerializer = DS.RESTSerializer.extend(
      DS.EmbeddedRecordsMixin,
      attrs:
        usedSpareParts: embedded: "always"
        timeEntries: embedded: "always"
    )

calling store.unloadRecord with a service-record gives the following error

Error: Assertion Failed: calling set on destroyed object
    at new Error (native)
    at Error.EmberError (http://localhost:42000/fieldops/ui/vendor/ember-1.12.0.debug.js:12153:21)
    at Object.Ember.default.assert (http://localhost:42000/fieldops/ui/vendor/ember-1.12.0.debug.js:4854:13)
    at Object.set (http://localhost:42000/fieldops/ui/vendor/ember-1.12.0.debug.js:15722:22)
    at exports.default.mixin.Mixin.create.set (http://localhost:42000/fieldops/ui/vendor/ember-1.12.0.debug.js:29322:20)
    at Ember.Object.extend.flushCanonical (http://localhost:42000/fieldops/ui/vendor/ember-data-1.0.0-beta.17.js:6882:14)
    at ember$data$lib$system$relationships$state$has$many$$ManyRelationship.flushCanonical (http://localhost:42000/fieldops/ui/vendor/ember-data-1.0.0-beta.17.js:7125:22)
    at Queue.invoke (http://localhost:42000/fieldops/ui/vendor/ember-1.12.0.debug.js:878:18)
    at Object.Queue.flush (http://localhost:42000/fieldops/ui/vendor/ember-1.12.0.debug.js:943:13)
[...]

PR #5823, opened 2019-01-26
convert some has-many tests to async [wip]

BryanCrotaz:
converting tests to async-await

### DIFF
--- a/tests/integration/relationships/has-many-test.js
+++ b/tests/integration/relationships/has-many-test.js
@@ -95,8 +95,7 @@ module('integration/relationships/has_many - Has-Many Relationships', {
   },
 });
 
-testInDebug('Invalid hasMany relationship identifiers throw errors', function(assert) {
-  assert.expect(2);
+testInDebug('Invalid hasMany relationship identifiers throw errors', function (assert) {
   let { store } = env;
 
   // test null id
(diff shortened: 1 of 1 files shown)


BryanCrotaz:
@runspired one test is failing (`integration/relationships/has_many - Has-Many Relationships: unloading and reloading a record with hasMany relationship using async - #3084`)

It passes in the runloop version and fails in the async version - could you take a look at my translation and confirm it's correct? It may be that this has turned up a new bug.

tests/integration/relationships/has-many-test.js
```diff
@@ -3707,88 +3239,136 @@ test('unloading and reloading a record with hasMany relationship - #3084', funct
...
+  });
+
+  let user = env.store.peekRecord('user', 'user-1');
+  let message = env.store.peekRecord('message', 'message-1');
+
+  assert.equal(await get(user, 'messages.firstObject.id'), 'message-1');
+  assert.equal(await get(message, 'user.id'), 'user-1');
+
+  env.store.unloadRecord(user);
+
```
runspired: `await settled()` here

tests/integration/relationships/has-many-test.js
```diff
@@ -3707,88 +3239,136 @@ test('unloading and reloading a record with hasMany relationship - #3084', funct
...
+            data: [{ type: 'message', id: 'message-1' }],
+          },
+        },
+      },
+    ],
+  });
+
+  user = env.store.peekRecord('user', 'user-1');
+
+  await settled();
```
runspired: shouldn't need this `await settled()` call

Issue #5849, opened 2019-02-11
Broken in IE 11


btecu:
See https://github.com/emberjs/data/blob/release/addon/-private/system/many-array.js#L139

`.find` on an array is not supported in IE.

![image](https://user-images.githubusercontent.com/7517745/52573718-bd1d7e00-2de8-11e9-9242-60e3f685b32e.png)

runspired:
@btecu unless you are testing an addon in IE11 with prototype extensions turned off, arrays always have prototype extensions in which case `.find` will be present: https://github.com/emberjs/ember.js/blob/v2.14.1/packages/ember-runtime/lib/mixins/enumerable.js#L474

btecu:
@runspired Actually the application has prototype extensions turned off because of some conflicts with a third party library. Ember is also moving away from prototype extensions, since they are not good practice anyway.

btecu:
cc @rwjblue

runspired:
> Actually the application has prototype extensions turned off because of some conflicts with a third party library. Ember is also moving away from prototype extensions, since they are not good practice anyway.

You absolutely need prototype extensions to use Ember with arrays. There may come a time when Ember no longer requires this, but that day is not here and until it is then the "best practice" is to use them as it is also the only practice.

You may not notice the sometimes-subtle bugs that occur in all browsers (array mutations won't trigger re-renders) or the major performance regressions in older browsers (like IE11) that come with arrays when you turn off prototype extensions, but they are there.

If you wish to go down the dangerous road of turning off prototype extensions for an application, then I think it comes with the territory that you should also bring your own polyfill for `Array.find` for IE11.

btecu:
@runspired Actually `ember-data` runs without the prototype extensions fine. See here:
https://github.com/emberjs/data/blob/master/package.json#L103

You absolutely *do not* need prototype extensions to use Ember with arrays.

You can use the KVN compliant methods (`pushObject`, `pushObjects`...) and they work fine without having to enable prototype extensions.

Before upgrading `ember-data` from `3.4.3` to `3.7.0` it worked fine. My guess is that the issue was introduced with `RecordData`.

rwjblue:
I tend to agree with @btecu here. `this.canonicalState` should be declared as `A()` here:

https://github.com/emberjs/data/blob/c43f9f9c85806d4e9748af318fcab3720736030d/addon/-private/system/many-array.js#L133

And when we flushCanonical here:

[...]

PR #5856, opened 2019-02-15
[BUGFIX beta] Wrap `currentState` in `A()`. Fixes IE11.

btecu:
See #5849.

Thanks @rwjblue & @runspired.


runspired:
Closing this as wrapping internal arrays as ember arrays is not something we wish to do. An appropriate fix would install a polyfill for `find` in the case that the polyfill provided by `prototype extensions` has been removed, or use an alternative method for searching the array.

PR #5857, opened 2019-02-15
[BUGFIX beta] Use `filter` instead of `find` due to IE11

btecu:
Related to #5849.

cc @rwjblue & @runspired


runspired:
There's no need to PR to the `beta` branch, beta will be updated from `master`, but thank you for taking the time! <3

PR #5858, opened 2019-02-15
[BUGFIX canary] Use `filter` instead of `find` due to IE11

btecu:
Related to #5849.

cc @rwjblue & @runspired


rwjblue:
Thank you @btecu and @runspired!

PR #5859, opened 2019-02-16
Update blueprints to use native JS classes

ppcano:
Part of the Octane Tracking issue. 

Atm, the blueprints will only generate Native Classes if the env `EMBER_VERSION` is equal to `OCTANE`; the [octane-blueprints](https://github.com/ember-cli/ember-octane-blueprint) will set the variable.

It follows the same implementation than the Ember repo https://github.com/emberjs/ember.js/pull/17621

### DIFF
--- a/blueprints/model/index.js
+++ b/blueprints/model/index.js
@@ -31,6 +31,10 @@ module.exports = useEditionDetector({
     let attrs = [];
     let needs = [];
     let entityOptions = options.entity.options;
+    let includeHasMany = false;
+    let includeBelongsTo = false;
+    let includeAttr = false;
+    let importedModules = null;
 
     for (let name in entityOptions) {
       let type = entityOptions[name] || '';
@@ -47,6 +51,7 @@ module.exports = useEditionDetector({
 
       let attr;
       if (/has-many/.test(dasherizedType)) {
+        includeHasMany = true;
         let camelizedNamePlural = inflection.pluralize(camelizedName);
         attr = {
           name: dasherizedForeignModelSingular,
@@ -54,12 +59,14 @@ module.exports = useEditionDetector({
           propertyName: camelizedNamePlural,
         };
       } else if (/belongs-to/.test(dasherizedType)) {
+        includeBelongsTo = true;
         attr = {
           name: dasherizedForeignModel,
           type: dasherizedType,
           propertyName: camelizedName,
         };
       } else {
+        includeAttr = true;
         attr = {
           name: dasherizedName,
           type: dasherizedType,
@@ -73,9 +80,9 @@ module.exports = useEditionDetector({
       }
     }
 
-    if (attrs.length) {
-      let isOctane = process.env.EMBER_VERSION === 'OCTANE';
+    let isOctane = process.env.EMBER_VERSION === 'OCTANE';
 
+    if (attrs.length) {
       let attrTransformer, attrSeparator;
       if (isOctane) {
         attrTransformer = nativeAttr;
@@ -97,7 +104,22 @@ module.exports = useEditionDetector({
     });
     needs = '  needs: [' + needsDeduplicated.join(', ') + ']';
 
+    if (isOctane) {
+      importedModules = ['Model'];
+      if (includeAttr) {
+        importedModules.push('attr');
+      }
+      if (includeBelongsTo) {
+        importedModules.push('belongsTo');
+      }
+      if (includeHasMany) {
+        importedModules.push('hasMany');
+      }
+      importedModules = importedModules.join(', ');
+    }
+
     return {
+      importedModules: importedModules,
       attrs: attrs,
       needs: needs,
     };
@@ -112,20 +134,20 @@ function nativeAttr(attr) {
 
   if (type === 'belongs-to') {
     if (name === propertyName) {
-      result = '@DS.belongsTo';
+      result = '@belongsTo';
     } else {
-      result = "@DS.belongsTo('" + name + "')";
+      result = "@belongsTo('" + name + "')";
     }
   } else if (type === 'has-many') {
     if (inflection.pluralize(name) === propertyName) {
-      result = '@DS.hasMany';
+      result = '@hasMany';
     } else {
-      result = "@DS.hasMany('" + name + "')";
+      result = "@hasMany('" + name + "')";
     }
   } else if (type === '') {
-    result = '@DS.attr';
+    result = '@attr';
   } else {
-    result = "@DS.attr('" + type + "')";
+    result = "@attr('" + type + "')";
   }
   return result + ' ' + propertyName;
 }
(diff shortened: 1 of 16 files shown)


ppcano:
@rwjblue 

I would like to get rid of the param `{ isModuleUnification: true }`. This is not needed in the Ember blueprint. I think, we have to update the ember-test-helpers dependency and will work. 

What are other changes came to your mind?

rwjblue:
Not sure, I kinda want to get this in and start testing in apps to see what crops up...

runspired:
Isn't this mixing syntaxes or am I misreading these in some way?

For instance, it *looks* like this might generate the following in some cases:

```
class Foo extends DS.Model {
  myAttr: DS.attr(),
  myBars: DS.hasMany('bar')
}
```

ppcano:
@runspired 

> Isn't this mixing syntaxes or am I misreading these in some way?

😱 , Yes, you are right. I should have thought about the [ED decorators](https://ember-decorators.github.io/ember-decorators/docs/api/modules/@ember-decorators/data) case.

I have seen the `ember-decorators` is imported in the project. What is the current plan? A few options comes to my mind:

1) The Native Model Blueprint does not inject the `attrs` configuration, atm.

2) Implement `attrs` on the Native Model Blueprint using the `@ember-decorators/data` package.

3) Start the ED decorators for the `attrs/haMany/belongsTo` to be used later on this PR.

runspired:
@pzuraq probably has the best insights for whether these blueprints should

1) be held off
2) use ember-decorators
3) if our own belongsTo/hasMany will "just work" with the right ember versions and we gate the blueprints on a min ember version

pzuraq:
Happy to give context where I can!

1. We'll be creating a way to enable Octane via some sort of flag (currently it's an environment variable, but I believe they want to change that) and I believe Ember will be changing it's blueprints based on that flag (native classes if Octane, standard if not), so _ideally_ Ember Data would also probably match that behavior. Unsure if this PR does that, if not it should hold off.
2. It should not need to use `ember-decorators`
3. The computed properties exported by Ember Data for `attr`, `belongsTo`, and `hasMany` _should_ "just work" as decorators, no need for any updates I believe, though we should double check

blueprints/model/index.js
```diff
@@ -47,46 +48,96 @@ module.exports = {
...
       }
+      attrs.push(attr);
 
       if (/has-many|belongs-to/.test(dasherizedType)) {
         needs.push("'model:" + dasherizedForeignModelSingular + "'");
       }
     }
 
+    let attrTransformer, attrSeparator;
+    if (process.env.EMBER_VERSION === 'OCTANE') {
```
ppcano: We could create a util function under `lib/utilities` to detect `OCTANE`.  This function could be in the same module than `lib/utilities/module-unification` and this one could be renamed it to `env-detector`.

I would prefer to do this change in a follow-up PR and based on the reviewer instructions.

ppcano:
@runspired 

I think this PR is ready now; the Model blueprint can generate attr decorators and this behavior is validated on the tests.  

   `pzuraq` said:

> _should_ "just work" as decorators, no need for any updates I believe, though we should double check   

I tested one case of the `DS.attr` decorator in a simple app and it worked.  I thought that perhaps the easiest way to test the usage of the Ember Data `hasMany`, `attr` and `belongsTo` as decorators could be to change the existing tests like this [one](https://github.com/emberjs/data/blob/master/tests/integration/records/create-record-test.js#L8).

efx:
Tested this on a new app and found it generated the following:
```js
import DS from 'ember-data';
const { Model } = DS;

export default class UserModel extends Model {
@DS.attr('string') firstName;
  @DS.attr('string') lastName;
  @DS.attr('number') age
}
```
Shouldn't the last attribute have a semicolon?

ppcano:
@efx Thank you for testing and reporting. I have included the fix and now, it'll generate:

```js
import DS from 'ember-data';
const { Model } = DS;

export default class UserModel extends Model {
  @DS.attr('string') firstName;
  @DS.attr('string') lastName;
  @DS.attr('number') age;
}
```

runspired:
@ppcano in keeping with `prettier` and to ease the syntax a bit, could we make the output the following?

```js
import DS from 'ember-data';
const { Model, attr } = DS;

export default class UserModel extends Model {
  @attr('string')
  firstName;
  @attr('string')
  lastName;
  @attr('number')
  age;
}
```

ppcano:
@runspired Yes, I can work on these changes.

Just to double check:

```
@attr('string')
 firstName;
```

instead of 

```
@attr('string') firstName;
```

I lean toward the second.

runspired:
@ppcano while I prefer the second, `prettier` formats the first. It doesn't make sense to me to give folks something that prettier would immediately invalidate.

rwjblue:
> prettier formats the first

Prettier has changed its stance on this, checkout [this demo in the playground](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEBLAZgAgBTohASk2EwHoAqTAKwFcBnGTGCTABwCcIA3OJgC1R1MgzOzh52AWwCGMGKigBzTOVKYAvgB0oYADbS6QgGL5i2zJgACAI319MAEzjtWAbnNWnLzLel93UBYelngQHopwjADucKjsDthEwB5BgRraWlDaGDihSWSUtAxMLBzcvDACQiJiEjJyCsqqGiAANCAQrPLQdMig0uycUQAKAwi9KNJcEKgObSDW7NJgANaRAMqsy43IMOw0cO18MJK6AOoC8HRbYHDr46jyXI8AnsjghvMKdM4ww0uKGTIdDSXQ-dpUOgADwAQktVhtpJI4AAZBRwYGg8EgSFQ9aNXRwACKNAg8ExYMOIC27B+7He1mk1heumg8w4ChgZ1mlWQAA4AAztMo-M5LVjvDhwOk8eZiACONDicH+0kB0gp2J+klQu32VLoBOJpPJSBBlPaMCZ3IcvKQACZLUtULpGgBhCCSIEoKDQDHtehwAAqTIm5p+6nUQA).

runspired:
@rwjblue @ppcano awesome!

ppcano:
@runspired The Model blueprint syntax has been improved.  Now, it will generate:

```js
// Octane apps
import DS from 'ember-data';
const { Model, attr } = DS;

export default class UserModel extends Model {
  @attr('string') firstName;
  @attr('string') lastName;
  @attr('number') age;
}
```

```js
// Classic apps
import DS from 'ember-data';
const { Model, attr } = DS;

export default Model.extend({
  firstName: attr('string'),
  lastName: attr('string'),
  age: attr('number')
});
```

Let me know if Classic apps should continue using the `DS.attr/DS.belongsTo/DS.hasMany/DS.Model` syntax to remove that particular commit.

runspired:
@ppcano this looks amazing! Is it in good shape for `hasMany` and `belongsTo` as well? If so, I think we are likely able to merge.

ppcano:
@runspired Yes, `hasMany` and `belongsTo` are implemented and the tests also validate this behavior.

Issue #5866, opened 2019-02-22
Sync relationships clearing when links are present


ryanto:
I think I'm running into some unexpected behavior with sync relationships that have links.

When I push a sync relationship with links into the store the relationship is emptied with a `known-to-be-empty` warning. If my model has related data, then that related data is being unset.

### Reproduction

Here's a GH repo that shows a simplified version of what I'm seeing: https://github.com/ryanto/ember-data-sync-relationship-clear

I created a code sandbox for this repo: https://codesandbox.io/s/k9n7qz1k5r

### Description

This behavior looks like it was changed in 3.2, on 3.1 the relationship is not cleared. 

I think I would expect the `known-to-be-empty` warning to not be shown if Ember Data does know about the model having a relationship.

### Versions

```cli
 ~/p/ember-data-sync-relationship-clear$ npm ls ember-source; npm ls ember-cli; npm ls ember-data                        

ember-data-sync-relationship-clear@0.0.0 /Users/ryan/projects/ember-data-sync-relationship-clear
└── ember-source@3.7.3

ember-data-sync-relationship-clear@0.0.0 /Users/ryan/projects/ember-data-sync-relationship-clear
└── ember-cli@3.7.1

ember-data-sync-relationship-clear@0.0.0 /Users/ryan/projects/ember-data-sync-relationship-clear
└── ember-data@3.7.0
```

Thanks!

runspired:
We restored a behavior true prior to 2.18 that we had unknowingly eliminated (hadn't been a guarantee and had no test coverage, basically wasn't a known case, just one that happened to exist a long time).

Basically, if a relationship is `sync` and you push a payload that has no data member at all, we end up needing to assume that it is empty.

I think what you are finding here is that this is at odds with being able to "update the links" associated with a relationship, although I am a bit surprised that a relationship might be established without a link and later receive one...

ryanto:
Thanks for getting back to so quickly, appreciate that!

Both of the payloads have links, but that payload that made a request with jsonapi includes also had data. The backend application I work on uses a json:api library that always includes links, for requests made with and without includes.

Assuming the relationship is empty makes sense to me, but I'm wondering if we could expand the logic a little. Would you accept a PR that did...

1) If a sync relationship is not loaded and gets a payload with no data, it's assumed to be empty.
1) If a sync relationship is empty and gets a payload with no data, it's assumed to be empty.
1) If a sync relationship is not empty and gets a payload with no data, it's left unchanged.

[...]

PR #5880, opened 2019-03-01
Don't reset sync relationship

ryanto:
This PR changes the known to be empty behavior. It allows a second payload push without included data to keep any relationship data that the model knows about.

Related issue: https://github.com/emberjs/data/issues/5866

### DIFF
--- a/tests/unit/store/push-test.js
+++ b/tests/unit/store/push-test.js
@@ -606,6 +606,69 @@ testInDebug(
   }
 );
 
+testInDebug(
+  'Calling push with a link for a non async relationship should not reset an existing relationship',
+  function(assert) {
+    Person.reopen({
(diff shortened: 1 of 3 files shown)


tests/unit/store/push-test.js
```diff
@@ -606,6 +606,69 @@ testInDebug(
   }
 );
 
+testInDebug(
+  'Calling push with a link for a non async relationship should not reset an existing relationship',
+  function(assert) {
+    Person.reopen({
```
runspired: could you write a test that

- does not use `reopen`
- does not use `run`

To do this you may need to make the test `async`.  In some cases you may also want to `await settled()` to ensure relationship state is fully settled post-push, although in this case it should not matter.

PR #5901, opened 2019-03-09
fix: ensure missing resource linkage isn't normalized incorrectly

### LINKED ISSUE #5706: Loading belongsTo removes relationship for other instances
Hi,
after upgrading to the most recent data version (3.1.1 -> 3.5.0), I'm seeing different behavior when retrieving a inverse relationship.
That is as soon as I call `instance.get('relationship')` then the relationship model can't be used from any other instance that has the same relationship.

### Setup

Given 2 models:

```js 
    class Person extends Model {
      @hasMany('dog', { async: false, })
      dogs;
    }
```

```js
    class Dog extends Model {
      @belongsTo('person', { async: true, })
      person;

      @attr('string')
      name;
    }
```

Where:

* api returns a relationship without resource linkage for Person -> dogs
* api returns relationship with resource linkage for Dog -> person

### Reproduction

```js
assert.equal(dog1.belongsTo('person').id(), '1');
assert.equal(dog2.belongsTo('person').id(), '1');

await dog1.get('person');

assert.equal(dog1.belongsTo('person').id(), '1');
assert.equal(dog2.belongsTo('person').id(), '1'); // 💥
```

- https://github.com/makepanic/data/commit/3598ee45e70c263269923393d38875f86d6d3c62#diff-6d44095bc68c4170bdd69687536c84f4R4896

The issue sounds related to what https://github.com/emberjs/data/pull/5608 already fixed.

### Versions

Run the following command and paste the output below: `npm ls ember-source && npm ls ember-cli && npm ls ember-data`.

```cli
ember-source: 3.5.0
ember-cli: 3.3.0
```

makepanic:
fixes #5706

### DIFF
--- a/addon/-private/system/store/finders.js
+++ b/addon/-private/system/store/finders.js
@@ -202,12 +202,15 @@ function ensureRelationshipIsSetToParent(
         until: '3.8',
       });
     }
-    relationships[inverseKey] = relationships[inverseKey] || {};
-    relationships[inverseKey].data = fixRelationshipData(
-      relationshipData,
-      kind,
-      parentInternalModel
-    );
+
+    if (kind !== 'hasMany' || typeof relationshipData !== 'undefined') {
+      relationships[inverseKey] = relationships[inverseKey] || {};
+      relationships[inverseKey].data = fixRelationshipData(
+        relationshipData,
+        kind,
+        parentInternalModel
+      );
+    }
   }
 }
 
--- a/tests/integration/relationships/inverse-relationship-load-test.js
+++ b/tests/integration/relationships/inverse-relationship-load-test.js
@@ -4892,4 +4892,98 @@ module('inverse relationship load test', function(hooks) {
     );
     assert.equal(pal2Dogs.get('firstObject.id'), '2', 'hasMany relationship has correct records');
   });
+
+  test("loading belongsTo doesn't remove inverse relationship for other instances", async function(assert) {
+    let { owner } = this;
+
+    const scooby = {
+      id: '1',
+      type: 'dog',
+      attributes: {
+        name: 'Scooby',
+      },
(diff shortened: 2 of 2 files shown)


makepanic:
Should I also remove the warning from other tests in inverse-relationship-load-test.js ?

There's tons of `WARNING: You requested a record of type 'dog' with id '2' but the adapter returned a payload with primary data having an id of '2'.` which could be fixed by using strings for all ids.

runspired:
> There's tons of WARNING: You requested a record of type 'dog' with id '2' but the adapter returned a payload with primary data having an id of '2'. which could be fixed by using strings for all ids.

Yes, if you could make sure that we are comparing `id`s only after we have called `coerceId` on both that would eliminate the false positives here.

addon/-private/system/store/finders.js
```diff
@@ -202,12 +202,15 @@ function ensureRelationshipIsSetToParent(
...
       });
     }
-    relationships[inverseKey] = relationships[inverseKey] || {};
-    relationships[inverseKey].data = fixRelationshipData(
-      relationshipData,
-      kind,
-      parentInternalModel
-    );
+
+    if (typeof relationshipData !== 'undefined') {
```
runspired: I suspect we can be a little bit more careful here by checking if it's a `hasMany` relationship, as a `belongsTo` would never exhibit this bug behavior and should be fixed.

makepanic: Changed it to only check for undefined on hasMany relationships

runspired:
@makepanic want to open a new PR for cleaning up that warning?

makepanic:
> @makepanic want to open a new PR for cleaning up that warning?

Yes I think it makes sense as it's unrelated to this fix.

PR #5903, opened 2019-03-09
🚧 Enable CI failure for the external partners stage (and ember-canary job)

dcyriller:
While the external partners stage produces valuable information, it is subject to random failure. Let's allow it to fail in CI.

Let's do the same for ember-canary.

This would be the first step in making the CI green again.

The second step would be to fix `EMBER LTS 2.18` job.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -84,9 +84,17 @@ jobs:
       if: NOT (branch ~= /^(release|lts).*/)
       env: EMBER_TRY_SCENARIO=ember-canary
 
-    # runs tests against various open-source projects for early-warning regression analysis
-    # We typically have 4 concurrent jobs, these jobs below are ordered to optimize total completion time
-    # By running longer jobs first, we allow the shorter jobs to complete within the same time block in parallel
+    - stage: deploy
+      name: "Publish"
+      install: yarn install
+      script:
+        - node_modules/.bin/ember try:reset
+        - yarn build:production
+
+  # runs tests against various open-source projects for early-warning regression analysis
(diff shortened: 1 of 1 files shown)


runspired:
Thanks for working on this! Some notes (as the team has previously discussed what we want to do here)

- fix 2.18 tests (typically it's a case where we need to tweak how/where we have a `run` in a test. We can mostly avoid using `run` but there's a few that we can't that occasionally sneak in.

2.18 is what has been primarily keeping us red and keeping us from noticing failures in canary / external partners that we need to monitor.

- keep (don't allow failure) of ember canary test (yes, it fails decently regularly, but this is an early warning we monitor for when things have happened in ember that we either need to plan for or address in ember itself quickly, and has proved invaluable)
- keep (don't allow failure) of all external partner tests. These should be kept green at all times and now that model-fragments has done work to be compatible with more recent releases the three that were consistently failing (fragments and two others that also test against fragments) should be kept green. If this is not the case let's figure out if we can address that before we move to allow_failures for them.

runspired:
#5905 addressed the tests for `2.18` :D 

Next step is to investigate the `ember-canary` failure. It began happening around the same time as some changes to native classes and decorators landed in ember last week. Either Ember introduced a regression or we are doing something bad in the way we initialize models, either is possible but I suspect the former.

dcyriller:
Thank you for the detailed explanations. It look like we can close this PR then.

https://github.com/emberjs/data/pull/5905 does a good job fixing the tests for `2.18`

PR #5904, opened 2019-03-09
fix: coerce ids before warning for mismatch

makepanic:
see https://github.com/emberjs/data/pull/5901 for details


makepanic:
Do we need a test for this changed behavior?

runspired:
@makepanic a test is welcome

makepanic:
Added a test that fails without coerceId

PR #5905, opened 2019-03-09
Fix 2.18 tests for Travis/CI

ryanto:
I noticed there were a handful of tests, one written by me!, that were failing in 2.18. This is causing the travis badge to say "build failing".

This PR updates those tests to pass in 2.18, which should get the travis to green.

Here's what I did:

* Wrapped `application.boot` in `run` for initializer tests.
* Wrapped `unloadRecord` in `run`
* Used `.get` instead of ES getter for computeds

### DIFF
--- a/tests/integration/adapter/client-side-delete-test.js
+++ b/tests/integration/adapter/client-side-delete-test.js
@@ -75,7 +75,7 @@ test('client-side deleted records can be added back from an inverse', async func
 
   await book2.destroyRecord({ adapterOptions: { clientSideDelete: true } });
 
-  book2.unloadRecord();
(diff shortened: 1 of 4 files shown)


runspired:
A few of these I am surprised by as they were working in 2.18 previously 🤔 (mainly the `application.boot()` call). It might be something else in those tests that are causing issues; however, we drop support for 2.18 in 6 weeks I believe so it won't be long until all these `run` calls can be removed again! :D 

(cc @rwjblue)

ryanto:
Ok great! Ping me when we drop support and I'll revert the `run` calls from this PR, that way we can keep the test suite nice a tidy.

PR #5914, opened 2019-03-18
[FEAT identifiers] Implements RFC #403

runspired:
[Identifiers RFC](https://github.com/emberjs/rfcs/pull/403)


addon/-private/identifiers/cache.ts
```diff
@@ -0,0 +1,423 @@
...
+}
+
+export default class IdentifierCache {
+  // Typescript still leaks private properties in the final
+  // compiled class, so we may want to move these from _underscore
+  // to a WeakMap to avoid leaking
+  private _cache = {
+    lids: Object.create(null) as IdentifierMap,
+    types: Object.create(null) as TypeMap
+  };
```
mike-north: you may want to favor use of a type annotation over casting

addon/-private/identifiers/cache.ts
```diff
@@ -0,0 +1,423 @@
...
+  private _forget: ForgetMethod;
+  private _reset: ResetMethod;
+
+  constructor() {
+    // we cache the user configuredGenerationMethod at init because it must
+    // be configured prior and is not allowed to be changed
+    this._generate = configuredGenerationMethod || defaultGenerationMethod;
+    this._update = configuredUpdateMethod || defaultUpdateMethod;
+    this._forget = configuredForgetMethod || defaultForgetMethod;
+    this._reset = configuredResetMethod || defaultResetMethod;
```
mike-north: Is there a reason why class field initializers aren't used here?

addon/-private/identifiers/cache.ts
```diff
@@ -0,0 +1,423 @@
...
+    let index = keyOptions._allIdentifiers.indexOf(identifier);
+    keyOptions._allIdentifiers.splice(index, 1);
+
+    this._forget(identifier, 'record');
+  }
+}
+
+function isNonEmptyString(str?: string | null): str is string {
+  return typeof str === 'string' && str.length > 0;
+}
```
mike-north: You can make this a more useful check through using a generic.
```ts
function isNonEmptyString<T extends string>(str: T | undefined | null | ''): str is T {
  return typeof str === 'string' && str.length > 0;
}
```

This will allow it to work correctly with unioned string literal types

```ts
let x = 'bar' as '' | 'foo' | 'bar';

if (isNonEmptyString(x)) {
    x // 'foo' | 'bar'  (yours would have included '')
} else {
    x // '' (yours would have been never)
}

```

addon/-private/identifiers/cache.ts
```diff
@@ -0,0 +1,423 @@
...
+
+function getTypeIndex(typeMap: TypeMap, type: string): KeyOptions {
+  let typeIndex: KeyOptions = typeMap[type];
+  
+  if (typeIndex === undefined) {
+    typeIndex = {
+      lid: Object.create(null),
+      id: Object.create(null),
+      _allIdentifiers: [],
+    } as KeyOptions;
```
mike-north: prefer type annotation over casting

runspired:
Replaced by #6247

Issue #395, opened 2012-09-24
Refreshing a page on a dynamic segment, and then modifying the record causes a crash at updateId


sly7-7:
Let's say I'm in a page like projects/3/overview. When I reload the page, then want to modify a property of this project, I encouter the following error when at DS.Store#updateId:

`Uncaught Error: assertion failed: An adapter cannot assign a new id to a record that already has an id.
<Sesame.Project:ember847> had id: 3 and you tried to update it with 3. This likely happened because
 your server returned a data hash in response to a find or update that had a different id than the one you 
sent.`

I suspect that when reloading the page, the router deserialization push the id "3" (string) into the store, and when editing and receiving the server response, on didSaveRecord, the store tries to update the id with 3 (number).

Then, in the assert, there is a strict equal test. I'm not sure that beeing more permissive on the assert is the right solution

PR #5919, opened 2019-03-19
[FEAT packages] Convert to monorepo for RFC #395

HeroicEric:
https://github.com/emberjs/rfcs/blob/master/text/0395-ember-data-packages.md

all your disk space are belong to us


igorT:
It seems like addons which use ED work, but apps are not building/including ED properly, at least when linked locally

igorT:
@runspired @HeroicEric when run against Ember Observer, it seems like `ember-data/model` is not built/exported properly.

igorT:
```
not ok 35 Chrome 73.0 - [120171 ms] - Acceptance: Addons: addon not found
    ---
        actual: >
            null
        stack: >
            TypeError: Cannot set property default of #<Object> which has only a getter
```
is a bit of a misleading error, but it's caused by the import from `ember-data/model` not working. Many thanks to @rwjblue

igorT:
EmberAddons still have a babel problem, so I marked them as `allow-fail` in travis, and we can figure it out later to not block the PR

PR #5933, opened 2019-03-24
Set the json:api content-type only when needed

dcyriller:
## bug description
This would fix a bug where `contentType` was set on undefined `options`. You can see a failing test [here](https://travis-ci.org/emberjs/data/jobs/510449385) or [here](https://travis-ci.org/emberjs/data/jobs/510449392).

I noticed the error was occuring on some `DELETE` requests.

## bug fix
When sending an empty request (DELETE for instance), the Content-Type header wouldn't be needed. Here [the json:api spec regarding the requests headers](https://github.com/json-api/json-api/blob/ff35bb226ab96260cf0e80aedd87fe84ecb9cab4/_format/1.1/index.md#-universal-responsibilities)

## alternative
An alternative bug fix would have been:
```javascript
if (options) {
  options.contentType = ...
}
```

## CI jobs
This would fix the CI for 2 (or 3) external partners jobs. To get the CI green again, a last fix would be needed on `emberobserver` and `travisci` jobs.

From Discord:
> `emberobserver` unknown N failures, same content-type issue, seems to cause tests to leak and a cascade failure after that
> `travisci` unknown N failures, but the first failure is cannot read property 'any' of undefined and seems to be while looking at some env variables in a request, so possibly related to fetch


dcyriller:
This sounds like a duplicate of https://github.com/emberjs/data/pull/5935

@runspired you may close it safely

PR #5934, opened 2019-03-24
Remove dead code in rest adapter

dcyriller:
In `_ajax` method, `this._fetchRequest(options)` never gets called.


runspired:
I believe we want to keep `_fetchRequest` around to mirror `_ajax`. As awful as this sounds, `_ajax` became private API and users that had need of overriding it will likely continue to want to have a location to do so while still allowing `ajax` to setup the appropriate request context. Yes, it's private and intimate, but until we have introduced a better request paradigm there's not much we can do.

PR #5935, opened 2019-03-24
[BUGFIX canary] Fix options assignment in json-api adapter

mwpastore:
This should fix many of the partner addon issues.


runspired:
Fixes everything except for TravisWeb :D

PR #5938, opened 2019-03-25
[BUGFIX canary] Fix this.stack assignement in AdapterError

dcyriller:
When consuming ember@<3.8, an app would run into `error is undefined`.

This should fix the last failing job.


dcyriller:
Alternatively, we could check Ember version instead of checking `error` property. But I'm not sure how to do this.

runspired:
Digging at Ember Source (per @dcyriller's pointing me to some PRs) it seems some versions of Ember do not return from the constructor while others do. This safeguard then seems 👍

dcyriller:
If it can help: I noticed the failure on an app running `ember-source@3.7`.

PR #5941, opened 2019-03-28
Convert Record Data to Typescript

igorT:
Converts the RecordData class to TypeScript and partially converts InternalModel and Relationship layer to TS. In this PR I did not change any of the behavior, or the confusion around `null`/`undefined` but just typed it as is, leaving the improvements for followup work.


addon/-private/system/model/record-data.ts
```diff
@@ -5,10 +5,33 @@ import { assert, warn, inspect } from '@ember/debug';
...
+  storeWrapper: RecordDataStoreWrapper;
+  isDestroyed: boolean;
+  _isNew: boolean;
+  _bfsId: number;
+  __attributes: any;
+  __inFlightAttributes: any;
+  __data: any;
+  _scheduledDestroy: any;
+
+  constructor(modelName: string, id: string, clientId: string, storeWrapper: RecordDataStoreWrapper, store:any) {
```
runspired: I am surprised that this did not yell at us from within `InternalModel` but `id` here is `null` in the new record case fairly often.

igorT: It's called from inside store, which isn't in ts yet

igorT: Fixed

PR #5943, opened 2019-03-28
[chore] Drop references to Bower

dcyriller:
Bower is still referenced in the test suite (for the external partners tests). But all the projects against which ember-data is tested have dropped Bower. 🎉


runspired:
Thank you!

PR #5952, opened 2019-03-30
[CHORE in-progress-flags] improves infra for developing features.

runspired:
- finishes removal of dependence on `EmberEnv.FEATURES`
- updates CONTRIBUTING.md to provide guidance on how to use
  - brands these flags as `in-progress-features` to be more-clear about intent than "optional"
  - treats `true` flags as "releasable"
  - makes both `null` and `false` flags be stripped at build time
  - makes `null` flags have their own CI test that must pass
  - `null` and `false` flags are only capable of being enabled in local builds (in-project use only)

### Motivation

`ember-data` has not made use of feature-flags in a while, and our infra went stale. In addition, we were piggy-backing (poorly) on some of the infra provided by `emberjs` that has since been locked down.

Given incoming work that would be easier to parallelize with the presence of this infra, as well as recent discussions around improving the release model and infra across the ecosystem, I updated our infra with various needs and concerns in mind.

I've requested reviews here from members of other teams that have expressed high levels of interest in improving shared infra for visibility into our own needs.


runspired:
(Also I apologize for all the `lint autofix` changes that came in with this, we recently updated our lint config and these files hadn't been touched yet)

PR #5955, opened 2019-03-31
adds decorator tests for Octane

runspired:
Not functional yet. Probably needs the right Ember Version but also needs some Octane-only test helper wrapping to avoid running these modules at all (including the class definitions) any other time.


runspired:
closing in favor of #6339

PR #5957, opened 2019-04-01
Show transform usage after example

raulb:
Reading the [API information about transforms](https://api.emberjs.com/ember-data/release/classes/DS.Transform), the order seemed a little off. First, we indicate a basic example for transforms, and the usage that makes reference (`app/transforms/temperature`) is at the end of this section, after an option slightly more advanced.

This change moves the usage snippet right after the example.

### DIFF
--- a/addon/serializers/embedded-records-mixin.js
+++ b/addon/serializers/embedded-records-mixin.js
@@ -261,7 +261,7 @@ export default Mixin.create({
     Use a custom (type) serializer for the post model to configure embedded comments
 
     ```app/serializers/post.js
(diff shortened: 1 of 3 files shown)


runspired:
@raulb Thanks! I tried pointing this at `master` vs `release` as that is where PRs should be pointed but it's not a clean switch. Mind opening a new PR or fixing this one to be based off of `master` with just the necessary commit?

raulb:
@runspired Opened another [one](https://github.com/emberjs/data/pull/5961)!

PR #5966, opened 2019-04-05
poc of decorators cleaning up our common patterns

runspired:


runspired:
No longer relevant with private properties having landed. Cached getters that don't utilize the tracking system may still be something we want though at another time.

PR #5967, opened 2019-04-05
add `@ember/string` dependency in preparation of deprecation

locks:


runspired:
@locks any ideas why adding this breaks partner tests? For instance the TravisWeb job:

```
Global error: Uncaught Error: Could not find module `@ember/string` imported from `ember-resolver/resolvers/classic/index` at http://localhost:7357/assets/vendor.js, line 252
```

.yarnrc
```diff
@@ -3,5 +3,5 @@
 
 
 registry "https://registry.npmjs.org/"
-lastUpdateCheck 1575256805000
-yarn-path ".yarn/releases/yarn-1.19.2.js"
+lastUpdateCheck 1584964403046
```
locks: Used 1.18 because I was running into this problem: https://github.com/yarnpkg/yarn/issues/7807.

PR #5968, opened 2019-04-05
add explicit `@ember/string` dependency ahead of deprecation

locks:
- Update broccoli-stew so it doesn't try to fetch version of dependency that was yanked.
- Add `@ember/string` in preparation of incoming deprecation.
- Update `ember-cli-babel` to version that properly detects `@ember/string`.


locks:
I'm not sure what happened with the PR appveyor check.

PR #5973, opened 2019-04-06
[chore] Adjust external tests cache path

dcyriller:
To create the `__external-test-cache` directory out of this repository
scope, the file hierarchy has to be climbed two steps further.


igorT:
@dcyriller This broke running external tests for me locally. What was the goal behind this change?

runspired:
@igort it’s meant to be up a level to avoid polluting the project and I think helps us keep it around in Travis. Your file permissions may be off.

runspired:
@igort also note that this merely restored the location used previously prior to monorepo work

dcyriller:
> it’s meant to be up a level to avoid polluting the project and I think helps us keep it around in Travis

> this merely restored the location used previously prior to monorepo work

yes, exactly :)

igorT:
@runspired I don't think it had, because it depends on where you invoke it from. If you invoke it from repo root I think it tries to go much deeper

PR #5975, opened 2019-04-06
Add a CI job for ember-release with jQuery

dcyriller:
With the recent switch from `jQuery.ajax` to `fetch` in the adapters code and the `jQuery` removal in Ember, this job aims at adding a small safety net.

Note: in the PR diff, you may notice a ember-3.4 removal in `config/ember-try` file: this config was unused in CI jobs. And I think it doesn't make much sense to activate it. Indeed, `jQuery` was not yet removed in `ember-source@3.4`.


runspired:
I am 👍 for adding a scenario to test with jquery-removed, but after updating the LTS targets etc. for releases this now needs rebased (and potentially a slight touchup)

rwjblue:
OK, I've pushed a few changes here:

* Rebased
* Add `@ember/optional-features` -- without this the other changes have no effect
* Updated tests to run without jQuery by default (adding an ember-try scenario to test `default-with-jquery`)
* Add Azure pipelines jobs for testing with jQuery

Now that jQuery is being properly removed, this has a few test failures (19). Most seem directly related to testing jQuery functionality and can be fixed by adding a test helper to guard those tests when jQuery is missing, but others will need to be looked into to see whats going on.

dcyriller:
I've reworked unit and integration tests for the rest adapter.

Previously, there were two different files: `fetch` and `ajax` to test the 2 different scenarios. As the CI will run the 2 scenarios, I've found out that they may be merged. It has the benefit to increase the fetch scenario coverage (that previously had less tests than the ajax one).

packages/-ember-data/tests/integration/adapter/rest-adapter-test.js
```diff
@@ -2809,22 +2858,24 @@ test('createRecord - sideloaded records are pushed to the store', function(asser
   });
 });
 
-testInDebug(
-  'warns when an empty response is returned, though a valid stringified JSON is expected',
-  function(assert) {
-    server.post('/posts', function() {
-      return [201, { 'Content-Type': 'application/json' }, ''];
-    });
+if (hasjQuery) {
```
dcyriller: I've shamefully opted out of this last rest adapter integration test for fetch scenario. I've not been able to make it pass. The network call is not catched by pretender. I suspect an issue in Pretender.

rwjblue: Seems fine to me, but would you mind making an issue to dig into it?

dcyriller: Alright 👍 
Opened here: https://github.com/emberjs/data/issues/6083

runspired:
Needs a quick rebase but otherwise 👍 LGTM and would like to ship :)

dcyriller:
I've had more trouble than expected with the rebase. But it should be good now 👍

PR #5986, opened 2019-04-08
Guard against this.options.babel.plugins

dcyriller:
This would fix the emberaddons.com job. 

I've not been able to precisely track down the root cause of the failure. My hypothesis is that a slight change in the dependencies resolution has occured (with ember-cli-babel?).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,8 +31,6 @@ stages:
 
 jobs:
   fail_fast: true
-  allow_failures:
-    - name: "emberaddons.com"
 
   include:
     # runs tests with current locked deps and linting
--- a/packages/-ember-data/index.js
+++ b/packages/-ember-data/index.js
@@ -143,9 +143,10 @@ module.exports = {
   },
 
   buildBabelOptions() {
-    let existing = this.options.babel;
+    let babelOptions = this.options.babel || {};
+    let existingPlugins = babelOptions.plugins || [];
     let customPlugins = require('./lib/stripped-build-plugins')(process.env.EMBER_ENV);
-    let plugins = existing.plugins.map(plugin => {
+    let plugins = existingPlugins.map(plugin => {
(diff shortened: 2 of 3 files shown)


runspired:
@dcyriller seems to still happen once I made the packages hoist their dev-dependencies which I think reduces the chances that it's due to a weird resolution. This guard seems fine although we're still failing, so it's more likely an issue with babel resolution in general. I wonder if it is related to the failures in #5967

dcyriller:
Error in emberaddons job is:
```
Invalid Option: The plugins/built-ins 'transform-block-scoping,transform-typeof-symbol' passed to the 'exclude' option are not
    valid. Please check data/[plugin-features|built-in-features].js in babel-preset-env
```

This seems to indicate that the ember-data babel transpilation involves an old version of `@babel/preset-env`.

emberaddons.com consumes `babel-preset-env@1.6.1` though `ember-cli-babel@6.8.1`. Two babel plugins used in ember-data babel build (specifically `transform-block-scoping`, `transform-typeof-symbol`) are not supported yet in this `babel-preset-env` version.

So, ember-data would be processed in emberaddons build with emberaddons' {ember-cli,}babel versions. But I would expect `ember-data` transpilation to be done with its own {ember-cli,}babel versions (as in ember-cli every addon controls its own transpilation 🤔).

Also, I can't reproduce this failure locally (same fresh clone with the same Node and friends versions).

I will close this PR, as it doesn't fix the issue. Maybe the results of my investigation will help someone debugging this!

dcyriller:
I've pushed a commit to use a fork of emberaddons. In this fork, ember-cli-babel is upgraded to latest. Let's see what happens.

dcyriller:
Interestingly, the build error is fixed but an other error pops up: 
```
Error: Could not find module `@ember-data/model` imported from `ember-data/-private`
```

I'll close the PR as it doesn't fix the issue.

PR #5987, opened 2019-04-08
[CHORE packages] hoist dev-deps and linting

runspired:


.eslintrc.js
```diff
@@ -46,18 +46,18 @@ module.exports = {
...
+        'packages/*/bin/**',
+        'packages/*/blueprints/*/index.js',
+        'packages/*/blueprints/*.js',
+        'packages/*/config/**/*.js',
+        'packages/*/lib/**/*.js',
+        'packages/*/node-tests/**',
+        'packages/*/tests/dummy/config/**/*.js',
       ],
-      excludedFiles: ['addon/**/index.js'],
+      excludedFiles: ['packages/*/addon/**/index.js'],
```
rwjblue: Don’t need this any more (because we no longer have `index.js` without slashes or globs in the files array).

packages/-ember-data/package.json
```diff
@@ -68,67 +66,61 @@
...
-    "glob": "^7.1.3",
-    "json-typescript": "^1.1.0",
-    "loader.js": "^4.7.0",
-    "mocha": "^6.1.1",
-    "mocha-only-detector": "1.0.0",
-    "prettier": "^1.15.3",
-    "rimraf": "^2.6.2",
-    "rsvp": "^4.8.4",
-    "typescript": "~3.4.2"
+    "@babel/plugin-transform-typescript": "*",
```
rwjblue: I’m not 100% sure what yarn will do here (with all of these *’s). This seems fine for now, but keep an eye out for issues that present like the wrong version of a devDep.

PR #5993, opened 2019-04-09
add RecordData test

igorT:


packages/-ember-data/tests/integration/record-data/record-data-test.ts
```diff
@@ -0,0 +1,310 @@
...
+  let store;
+
+  hooks.beforeEach(function () {
+    let { owner } = this;
+
+    owner.register('model:person', Person);
+    owner.register('service:store', CustomStore);
+  });
+
+  test("A noop Record Data implementation that follows the spec should not error out", async function (assert) {
```
runspired: it is unclear here what a noop RecordData is. I'm assuming it's because the `TestRecordData` defined above doesn't do much. `noop` is also just a vague and difficult to grok term ime. Maybe just `"A RecordData implementation that has the required spec methods should not error out"`?

packages/-ember-data/tests/integration/record-data/record-data-test.ts
```diff
@@ -0,0 +1,310 @@
...
+        },
+      ],
+    });
+
+    await settled();
+
+    assert.equal(get(all, 'length'), 3);
+  });
+
+  test("Record Data push and save lifecycle", async function (assert) {
```
runspired: a similar test for `create and save` would be nice.

packages/-ember-data/tests/integration/record-data/record-data-test.ts
```diff
@@ -0,0 +1,310 @@
+import { get } from '@ember/object';
+import { setupTest } from 'ember-qunit';
+import Model from 'ember-data/model';
```
runspired: why are we using the legacy imports instead of the packages imports?

packages/-ember-data/tests/integration/record-data/record-data-test.ts
igorT: It's in there, should rename the test

packages/-ember-data/tests/integration/record-data/record-data-test.ts
igorT: Copied/consistency with the other tests, we can run a codemod once ready

packages/-ember-data/tests/integration/record-data/record-data-test.ts
```diff
@@ -0,0 +1,310 @@
...
+    clientPerson.save();
+    await settled();
+    clientPerson.unloadRecord();
+    await settled();
+    assert.equal(calledPush, 0, 'Called pushData');
+    assert.equal(calledWillCommit, 2, 'Called willCommit');
+    assert.equal(calledWasRejected, 1, 'Called commitWasRejected');
+    assert.equal(calledUnloadRecord, 1, 'Called unloadRecord');
+    assert.equal(calledDidCommit, 1, 'Called didCommit');
+    assert.equal(calledClientDidCreate, 1, 'Did not called clientDidCreate');
```
runspired: this should be `did call clientDidCreate`

runspired: I would add something to check the ordering of these calls over just asserting counts, because as written `clientDidCreate` could be called after the first successful save (or later) and we'd never know if that was the right intent.

Typically asserting just after the code that invokes the counter is a safer way to ensure the timing.

packages/-ember-data/tests/integration/record-data/record-data-test.ts
```diff
@@ -0,0 +1,310 @@
...
+    calledDidCommit = 0;
+
+    let clientPerson = store.createRecord('person', { id: 2 });
+    clientPerson.save();
+    await settled();
+    clientPerson.save();
+    await settled();
+    clientPerson.unloadRecord();
+    await settled();
+    assert.equal(calledPush, 0, 'Called pushData');
```
runspired: should be `did not call pushData`

packages/-ember-data/tests/integration/record-data/record-data-test.ts
runspired: I see it now, hidden in the latter portion of that test. I would break that out into two tests.

igorT:
I did originally but found this to be cleaner due to counter and class
setup. If I add even more, I might break them apart again

packages/-ember-data/tests/integration/record-data/record-data-test.ts
igorT: Great point

PR #5995, opened 2019-04-09
[CHORE packages] refactors shared infra into `-build-infra` package

runspired:


packages/-build-infra/.npmignore
```diff
@@ -0,0 +1,31 @@
+# compiled output
+/dist/
+/tmp/
+
+# dependencies
+/bower_components/
```
igorT: Don't need to add this I think

igorT: Also the other bower things

runspired: opened a different PR to separate out the blueprint changes, also dropped these for you :)

PR #5999, opened 2019-04-09
[fix] Add json:api header on every requests

dcyriller:
Please, find additional info here: https://github.com/emberjs/data/commit/f26d07b62e4240f723a822b73d5a196ebee8de22

PR #6000, opened 2019-04-09
[fix] Add json:api header on every requests

dcyriller:
This is a port of https://github.com/emberjs/data/pull/5999 to 3.9

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,7 +27,7 @@ stages:
   - additional tests
   - ember version tests
   - name: external partner tests
-    if: NOT (branch ~= /^(release|lts).*/)
+    if: NOT (branch ~= /^(emberjs:release|emberjs:lts|release|lts).*/)
   - name: deploy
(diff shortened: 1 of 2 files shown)


runspired:
hrm, unsure why appveyor and travis are being bad here. I will have to dig in.

dcyriller:
I've added a commit which is a port of https://github.com/emberjs/data/pull/6006 to release branch. It activates CI's `test` and `additional tests` for release and lts branches.

dcyriller:
I've revert the last commit as it would put too much pressure on the CI 🙂

bryanhickerson:
@dcyriller what release of ember 3.9 is this in? I can't tell from the CHANGELOG/commit history, so forgive me if it's obvious.

runspired:
@bryanhickerson probably 3.9.1 but definitely 3.9.2. The PR wasn't labelled correctly to be picked up by the changelog generation unfortunately.

runspired:
Confirmed it is in 3.9.1

PR #6003, opened 2019-04-10
remove _relationshipState from BelongsTo PromiseProxy

igorT:


packages/store/addon/-private/system/model/internal-model.ts
```diff
@@ -508,7 +508,10 @@ export default class InternalModel {
       let internalModel =
         resource && resource.data ? store._internalModelForResource(resource.data) : null;
       return PromiseBelongsTo.create({
-        _belongsToState: resource._relationship,
+        key,
+        modelName: relationshipMeta.type,
+        store,
+        originatingInternalModel: this,
```
runspired: discussed oob but adding here so we don't forget:

I don’t like leaking this many properties and there are other ways to improve the entanglement.

Let’s at least push these props somewhere they won’t conflict with the proxy.

@igorT suggested padding them onto an object replacing the _belongsToState key we already use before eventually moving to WeakMap or symbol. I am 👍 on that.

igorT: fixed

Issue #5998, opened 2019-04-09
Could not find module `fetch` imported from `ember-data/adapters/rest` on a newly generated app with ember-cli-fastboot


mansona:
### Reproduction

Steps to reproduce: 

- ember new test-app
- cd test-app
- ember install ember-cli-fastboot
- npm start
- visit http://localhost:4200 
- 💥 

here is a simple reproduction (because it was asked for) https://github.com/mansona/data-fastboot-test

### Description

I understand that there are things required for enabling dependencies in fastboot but this flow used to work and it should still work i.e. we shouldn't expect people to do extra steps to get fastboot working for something that is included in the default blueprint

### Versions

```cli
data-fastboot-test@0.0.0 /Users/mansona/temp/data-fastboot-test
└── ember-source@3.9.1

data-fastboot-test@0.0.0 /Users/mansona/temp/data-fastboot-test
└── ember-cli@3.9.0

data-fastboot-test@0.0.0 /Users/mansona/temp/data-fastboot-test
└── ember-data@3.9.0
```

runspired:
@mansona I think this is an issue with one of the fastboot dependencies vs `ember-data`. We do ensure that `fetch` is present: https://github.com/emberjs/data/blob/v3.9.0/package.json#L57

I think someone noticed once before though (but having trouble finding the ticket) that fastboot is bringing it's own and clobbering something in a bad way.

mansona:
@runspired does it have anything to do with https://github.com/ember-fastboot/ember-cli-fastboot/issues/588?

rwjblue:
The issue is that ember-cli-fastboot only calls `updateFastbootManifest` on top level addons, ember-fetch needs to use that hook to provide the correct node `fetch` (vs the browser one).

@kiwiupover has been working on this for a bit but has hit a few yak shaves as he goes (the ember-cli-fastboot test suite itself is borked due to ember-data monorepo changes for example).

runspired:
This was resolved in 3.9.1 and again in 3.9.3 (after 3.9.2 regressed it).

PR #6011, opened 2019-04-10
ember-fetch fastboot passthrough

mansona:
This is an attempt to fix the issue with fastboot on newly generated ember apps on version 3.9.0

We have discussed this solution at length in the #fastboot channel on the Ember Community Discord and this was one of the quicker ways to fix the issue and the one that required the least amount of up-front work.

Essentially the problem arises from the fact that ember-cli-fastboot should be walking the tree of addons and finding all addons that add things to the manifest using `updateFastBootManifest()` but it currently does not 😞 

As ember-data has now added ember-fetch as a dependency, we need to include the manifest information from ember-fetch for any app that uses ember-data to be able to use fastboot.

If my explanation of this doesn't seem right please feel free to correct me 😂 

Fixes https://github.com/emberjs/data/issues/5998
Fixes https://github.com/ember-fastboot/ember-cli-fastboot/issues/682

### DIFF
--- a/index.js
+++ b/index.js
@@ -8,6 +8,7 @@ const version = require('./lib/version');
 const { isInstrumentedBuild } = require('./lib/cli-flags');
 const BroccoliDebug = require('broccoli-debug');
 const calculateCacheKeyForTree = require('calculate-cache-key-for-tree');
+const resolve = require('resolve');
 
 function isProductionEnv() {
   let isProd = /production/.test(process.env.EMBER_ENV);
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -69,6 +70,19 @@ module.exports = {
     return path.join(__dirname, 'blueprints');
   },
 
+  updateFastBootManifest(manifest) {
+    manifest.vendorFiles.push('ember-data/fetch-fastboot.js');
```
runspired: I think we could lookup the ember-fetch addon and manually call `updateFastBootManifest` ourselves to be safer

rwjblue: Ya, but this is slightly different. This PR is intentionally avoiding/working around the guard added by https://github.com/ember-cli/ember-fetch/pull/167 (which prevented the `treeForPublic` from returning anything when `ember-fetch` was not a top level addon).

mansona: @rwjblue and I've also named this `ember-data/fetch-fastboot.js` instead of `ember-fetch/fetch-fastboot.js` like it would have been 🤔 I guess I could have done that but I was trying to avoid any name clashes. I don't know which way is better, I just went with my gut

runspired:
This will need ported to `master` as well.

PR #6018, opened 2019-04-11
Fix travis release

runspired:


runspired:
Shows that the issue is not present on all release targeting PRs. Traced this down to the need to add `emberjs:release` and `emberjs:lts` to the contitionals. @dcyriller will add that to #6000 and prepare a similar fix targeting release.

Issue #6002, opened 2019-04-10
Tests throw decorator deprecations


igorT:
Running the test suite, there are deprecations regarding the old decorator syntax

runspired:
I'm not sure we can eliminate these yet because we don't have a way to differentiate when `octane` decorators are available vs not. cc @pzuraq because it's a bit annoying that ember-decorators chose to deprecate prior to there being a path forward for apps and addons.

NullVoxPopuli:
There is an environment variable we can check

https://github.com/ember-cli/ember-octane-blueprint/blob/master/packages/%40ember/octane-app-blueprint/files/.ember-cli.js

pzuraq:
@NullVoxPopuli that environment variable is likely not available here, since they aren't going to be updating the Ember Data addon to use Canary (and it has to test against older versions anyways).

@runspired you should be able to use `ember-compatibility-helpers` to test against `gte('3.10.0')` since we have a target release version now, and it's in beta. I wouldn't add production code using this just yet, maybe when we're closer to launch, but it should be fine for tests.

Alternatively, you should also be able to install the polyfill, which will allow you to use the native APIs with all of the ember versions in your try matrix it looks like.

igorT:
This seems related to our build failing https://travis-ci.org/emberjs/data/jobs/519091240
getting a bunch of  `(0 , _decorator.decorator) is not a function`

NullVoxPopuli:
PR: https://github.com/emberjs/data/pull/6026

runspired:
Closed by #6026

PR #6026, opened 2019-04-13
address the syntax error with decorators in tests

NullVoxPopuli:
Addresses: https://github.com/emberjs/data/issues/6002
- Fixes decorator syntax error


igorT:
@NullVoxPopuli thanks for this! It fails with ember 3.4 though.

runspired:
@pzuraq does the polyfill support 3.4?

NullVoxPopuli:
looks like it's an issue with some classic syntax. 

in the failing test if I convert the classic model from

```ts
Model.extend({ 
  name: attr() 
});
```
to
```ts
class extends Model {
  @attr() name;
}
```
it passes

runspired:
@NullVoxPopuli weird that only one test fails for that reason, there's likely an extenuating circumstance. We should investigate so that if there is an issue with the polyfill it can be improved.

NullVoxPopuli:
Just starting with some tests here: https://github.com/pzuraq/ember-decorators-polyfill/pull/12

NullVoxPopuli:
Now that the C.I. has finished, and it's a new day, I can confirm that the issue is _likely_ with the polyfill. So, excellent. I don't have time today to keep investigating this, but I can keep poking here and there as I have little time pockets.

runspired:
Looks like the PR to ember-decorators-polyfill was merged so bumping the version should unblock this :) 

Sidenote there seems to be an issue with CI right now, someone will need to investigate.

NullVoxPopuli:
re: failing C.I.:
ember-m3 has a linting error
factory-guy broke entirely

PR #6032, opened 2019-04-15
[CHORE belongs-to] expand coverage for async resolution

runspired:


packages/-ember-data/tests/integration/relationships/belongs-to-test.js
```diff
@@ -119,6 +119,99 @@ module('integration/relationship/belongs-to BelongsTo Relationships (new-style)'
...
+    );
+
+    this.owner.register(
+      'adapter:person',
+      JSONAPIAdapter.extend({
+        shouldBackgroundReloadRecord() {
+          return false;
+        },
+        findRecord() {
+          assert.equal(++personFindRecordCalls, 1, 'We call findRecord only once for our person');
```
igorT: Can we separate the ++ into the next line please

packages/-ember-data/tests/integration/relationships/belongs-to-test.js
```diff
@@ -119,6 +119,99 @@ module('integration/relationship/belongs-to BelongsTo Relationships (new-style)'
...
+          return resolve({
+            data: {
+              type: 'person',
+              id: '1',
+              attributes: { name: 'Chris' },
+              relationships: {
+                bestDog: {
+                  data: { type: 'pet', id: '1' },
+                  links: {
+                    related: './pet/1',
```
igorT: Didn't know we supported this format

packages/-ember-data/tests/integration/relationships/belongs-to-test.js
```diff
@@ -119,6 +119,99 @@ module('integration/relationship/belongs-to BelongsTo Relationships (new-style)'
...
+
+    const person = await store.findRecord('person', '1');
+    const pet = store.peekRecord('pet', '1');
+    const personPetRequest = person.get('bestDog');
+    const personPet = await personPetRequest;
+
+    assert.ok(personPet === pet, 'We ended up in the correct state via .get');
+
+    let promiseDog;
+    await personPetRequest.then(dog => {
```
igorT: Can you write a comment about what the goal of this test is, I'm pretty confused when reading it. Doesn't await return the resolved value anyhow, so we have the `personPet` from before?

packages/-ember-data/tests/integration/relationships/belongs-to-test.js
```diff
@@ -119,6 +119,99 @@ module('integration/relationship/belongs-to BelongsTo Relationships (new-style)'
...
+      promiseDog = dog;
+      return dog;
+    });
+
+    assert.ok(promiseDog === pet, 'We ended up in the correct state via .get.then');
+
+    const foundPet = await store.findRecord('pet', '1');
+
+    assert.ok(foundPet === pet, 'We ended up in the correct state via findRecord');
```
igorT: This is asserting that there wasn't a request made right, there isn't really a failure scenario for peek to return something else than a find, right?

runspired: no, it is asserting that the result of `findRecord` matches the result of `peekRecord`

packages/-ember-data/tests/integration/relationships/belongs-to-test.js
runspired: The test description is clear... `async belongsTo resolves to the record when sideloaded`

This test checks that various mechanisms of accessing an async belongsTo correctly resolve to the record instance (vs a promiseProxy) when the record was sideloaded. It additionally guards against accidental network requests to ensure that we are actually accessing the sideloaded version.

packages/-ember-data/tests/integration/relationships/belongs-to-test.js
runspired: Yes... We support jsonapi links for relationships and this is the more common format: https://jsonapi.org/format/#document-links

packages/-ember-data/tests/integration/relationships/belongs-to-test.js
runspired: 👍

packages/-ember-data/tests/integration/relationships/belongs-to-test.js
runspired: I added some additional comments and specified in the title that we are resolving to the record, not a proxy to make it even more clear :)

packages/-ember-data/tests/integration/relationships/belongs-to-test.js
```diff
@@ -119,6 +119,114 @@ module('integration/relationship/belongs-to BelongsTo Relationships (new-style)'
...
+    const pet = store.peekRecord('pet', '1');
+
+    // the promise returned by accessing the property should
+    //   resolve to the record instance
+    const personPetRequest = person.get('bestDog');
+    const personPet = await personPetRequest;
+
+    assert.ok(person !== null, 'Precond: We loaded the person');
+    assert.ok(pet !== null, 'Precond: We loaded the pet');
+    assert.ok(personPet === pet, 'We ended up in the correct state via .get');
```
igorT: Probably should use `assert.equal` for this

packages/-ember-data/tests/integration/relationships/belongs-to-test.js
igorT: Thanks that helped a lot

packages/-ember-data/tests/integration/relationships/belongs-to-test.js
```diff
@@ -119,6 +119,114 @@ module('integration/relationship/belongs-to BelongsTo Relationships (new-style)'
...
+
+    // the promise returned by accessing the property should
+    //   resolve to the record instance even inside the promise
+    //   chain
+    let promiseDog;
+    await personPetRequest.then(dog => {
```
igorT: isn't this the same as `let promiseDog = awaitPersonPetRequest`? Maybe I'm misreading it

PR #6033, opened 2019-04-16
Pass options in finders#_query even if createRecordArray is falsy

velrest:
Currently if `adapterOptions` are passed to store.query they are ignored if `createRecordArray` is falsy.
    
To prevent this we also pass the options to `adapter.query` if `createRecordArray` is falsy.


runspired:
@velrest the check for `createRecordArray` is based on the function length of `adapter.query`, if you are not receiving the options it is because your adapter does not implement a custom `query` method expecting more than 3 arguments.

velrest:
TIL about function.length. Do you know the design choice behind why the default rest serializer ignores the adapterOptions?

Thanks a lot!

runspired:
@velrest it is expensive to materialize the recordArray for the query when you don’t need it, so if the app isn’t using the arguments we don’t materialize.

PR #6035, opened 2019-04-16
[DOC contributing.md] | Clarifying that yarn should be used to run tests

pete-the-pete:
Contributing currently says to use `ember test`, but that results in `No ember-cli-build.js found.`.
The correct command is `yarn test`.


runspired:
Thanks! :)

PR #6042, opened 2019-04-19
[BUGFIX jQuery] avoid jQuery deprecation notice for intentional use

runspired:
Uses global `jQuery` when available to avoid the deprecation on `Ember.$`. This may result in some apps using a different version of `jQuery` than expected if global `jQuery` and `Ember.$` are different versions of `jquery` (e.g. `no-conflict` mode).

cc @wagenet

### DIFF
--- a/packages/adapter/addon/rest.js
+++ b/packages/adapter/addon/rest.js
@@ -1,9 +1,8 @@
-/* globals najax */
+/* globals najax jQuery */
 /**
   @module ember-data
 */
 
-import Ember from 'ember';
 import fetch from 'fetch';
 import serializeQueryParams from 'ember-fetch/utils/serialize-query-params';
 import determineBodyPromise from 'ember-fetch/utils/determine-body-promise';
@@ -30,6 +29,7 @@ import { warn } from '@ember/debug';
 import { DEBUG } from '@glimmer/env';
 
 const Promise = EmberPromise;
+const jQ = typeof jQuery !== 'undefined' ? jQuery : undefined;
 
 /**
   The REST adapter allows your store to communicate with an HTTP server by
@@ -301,7 +301,7 @@ const RESTAdapter = Adapter.extend(BuildURLMixin, {
 
   useFetch: computed(function() {
     let ENV = getOwner(this).resolveRegistration('config:environment');
-    return (ENV && ENV._JQUERY_INTEGRATION) === false || Ember.$ === undefined;
+    return (ENV && ENV._JQUERY_INTEGRATION) === false || jQ === undefined;
   }),
 
   /**
@@ -1026,7 +1026,7 @@ const RESTAdapter = Adapter.extend(BuildURLMixin, {
     @param {Object} options jQuery ajax options to be used for the ajax request
   */
   _ajaxRequest(options) {
-    Ember.$.ajax(options);
+    jQ.ajax(options);
   },
 
   /**


packages/adapter/addon/rest.js
```diff
@@ -30,6 +30,8 @@ import { warn } from '@ember/debug';
 import { DEBUG } from '@glimmer/env';
 
 const Promise = EmberPromise;
+const jQueryDescriptor = Object.getOwnPropertyDescriptor(Ember, '$');
+const jQuery = jQueryDescriptor ? jQueryDescriptor.value || jQueryDescriptor.get() : undefined;
```
wagenet: Hmm... I don't think is right. Calling `jQueryDescriptor.get` will trigger the deprecation warning.

runspired: @wagenet I confirmed that it does not, and also it's the only way to get the value in 3.9+

wagenet: For the `useFetch` check you just want to check if it would exist right? Then if it actually gets used, triggering the deprecation warning would be correct.

wagenet: image

wagenet:
```suggestion
const jQueryDefined = jQueryDescriptor ? jQueryDescriptor.value || jQueryDescriptor.get : false;
```

packages/adapter/addon/rest.js
```diff
@@ -301,7 +303,7 @@ const RESTAdapter = Adapter.extend(BuildURLMixin, {
 
   useFetch: computed(function() {
     let ENV = getOwner(this).resolveRegistration('config:environment');
-    return (ENV && ENV._JQUERY_INTEGRATION) === false || Ember.$ === undefined;
+    return (ENV && ENV._JQUERY_INTEGRATION) === false || jQuery === undefined;
```
wagenet:
```suggestion
    return (ENV && ENV._JQUERY_INTEGRATION) === false || !jQueryDefined;
```

packages/adapter/addon/rest.js
```diff
@@ -1026,7 +1028,7 @@ const RESTAdapter = Adapter.extend(BuildURLMixin, {
     @param {Object} options jQuery ajax options to be used for the ajax request
   */
   _ajaxRequest(options) {
-    Ember.$.ajax(options);
+    jQuery.ajax(options);
```
wagenet:
```suggestion
    Ember.$.ajax(options);
```

packages/adapter/addon/rest.js
runspired: @wagenet the deprecation for `ajax` would be incorrect because you may have actually installed `@ember/jquery`.

I am surprised you receive that assertion there as testing locally I do not get that. What version of Ember are you using?

wagenet: That's 3.9.1, in the test console.

If I'm understanding correctly, if ember-jquery is installed, using `Ember.$` still isn't correct right? So either way, we'd need a different code path to avoid the deprecation.

wagenet: The deprecation warning is in the `get` for `Ember.$` so I don't see any way to avoid the deprecation warning there.

runspired: > If I'm understanding correctly, if ember-jquery is installed, using Ember.$ still isn't correct right? So either way, we'd need a different code path to avoid the deprecation.

Correct.

runspired: @rwjblue do you know why ember changed the descriptor in 3.9 to not have `value` and use `get` instead?

runspired: @wagenet I'm confused still why 3.9 for me did not print a deprecation when I tested `descriptor.get()` (maybe it was 3.9.0?, maybe we had a release with the deprecation broken? maybe I trolled myself and it doesn't print more than once and it already had?)

wagenet: https://github.com/emberjs/ember.js/commit/3c00987af16c2e80c81895f6e18cfe10db37387f

wagenet: Looks the same in [3.9.0](https://github.com/emberjs/ember.js/blob/v3.9.0/packages/ember/index.js#L564) and [3.9.1](https://github.com/emberjs/ember.js/blob/v3.9.1/packages/ember/index.js#L564)

runspired: I moved to using global jQuery

PR #6043, opened 2019-04-19
[BUGFIX jQuery] avoid jQuery deprecation notice for intentional use

runspired:


rwjblue:
I am not sure this is the correct fix 🤔

runspired:
@rwjblue yeah I reverted it once we decided we wanted to go a different direction.

PR #6044, opened 2019-04-19
[WIP][BUGFIX] createRecord can be async

pzuraq:
`createRecord` currently uses `run.join` to wrap some code that will schedule
async tasks, but is meant to complete _synchronously_. If a scheduled runloop 
already exists, however, `run.join` will schedule the task asynchronously on that 
runloop, breaking this assumption.


runspired:
After discussing more we concluded that tests that check for `RecordArray` changes synchronously following `createRecord` are wrong and should `await settled()` in between.

igorT:
@runspired whats the next step for this PR? Close?

runspired:
Confirm, closing.

PR #6045, opened 2019-04-19
[CHORE blueprints] move blueprints into their appropriate packages

runspired:
This also sets us up with a command we can use to begin migrating tests to their appropriate packages.


igorT:
@runspired I think most of this work landed as a separate PR, so gonna close for now. Please reopen if I screwed it up

runspired:
I rebased this on the plane. Literally zero of it  has landed 😂

packages/-build-infra/package.json
```diff
@@ -21,7 +23,20 @@
     "ember-cli-string-utils": "^1.1.0",
     "git-repo-info": "^2.0.0",
     "npm-git-info": "^1.0.3",
-    "silent-error": "^1.1.1"
+    "silent-error": "^1.1.1",
+    "glob": "^7.1.3",
+    "mocha": "^6.1.2",
+    "mocha-only-detector": "1.0.1",
```
rwjblue: We should make an issue to remove this (in favor of using an eslint plugin that fails for `.only`).

packages/-build-infra/package.json
```diff
@@ -21,7 +23,20 @@
...
+    "rimraf": "^2.6.2",
+    "rsvp": "^4.8.4",
+    "ember-cli-version-checker": "^3.1.2",
+    "ember-cli-blueprint-test-helpers": "^0.19.1"
+  },
+  "devDependencies": {
+    "broccoli-test-helper": "*",
+    "co": "*",
```
rwjblue: Is this still needed? Since we are Node 8+ now we should make an issue to refactor our broccoli tests to async/await...

packages/-build-infra/src/node-test-helpers/nodetest-runner.js
```diff
@@ -0,0 +1,81 @@
+'use strict';
```
rwjblue: I don't think we _need_ this file any more, we should remove (seems fine to do as a follow up).

packages/-build-infra/src/node-test-helpers/setup-test-environment.js
```diff
@@ -1,3 +1,4 @@
+/* global beforeEach afterEach */
```
rwjblue: Shouldn't need to use `globals` here, points to an issue with eslint config

packages/-build-infra/package.json
runspired: https://github.com/emberjs/data/issues/6109

packages/-build-infra/package.json
runspired: https://github.com/emberjs/data/issues/6110

packages/-build-infra/src/node-test-helpers/nodetest-runner.js
runspired: @rwjblue why wouldn't we need the test runner anymore?

rwjblue: What does it do that the `mocha` script itself doesn't? We originally (waaaaaaaaayyyyyy back) needed it because `mocha` didn't support globs properly, I think we can remove it.

packages/-build-infra/src/node-test-helpers/setup-test-environment.js
runspired: Fixed

packages/-build-infra/src/node-test-helpers/nodetest-runner.js
runspired: if we shift to a lint rule for no `.only` and fix our windows support we *might* be able to remove it? In addition to those things and glob it seems to be ensuring some form of a temporary node_modules directory, unsure if that directory is even in use anymore. Will open an issue.

runspired: https://github.com/emberjs/data/issues/6111

PR #6050, opened 2019-04-20
[BUGFIX ts] ensure all packages have tsconfig & prepublish

runspired:
This allows `npm publish --dry-run` to complete in each package which should unblock us from releasing a beta.

A few follow up notes:

- we aren't using rollup on most of the code anymore, we should restore this as a pre-publish step which would play nicer with consumption and embroider anyway
- we should make our rollup code something available in build-infra so that all our packages can use it
- we need to do a bit more work to ensure we aren't shipping types, I forget the details of the strategy that @mike-north had concocted but each package will need to be configured for that and this is something we must fix prior to this becoming a stable release.


igorT:
Can you say more about parts of the code using rollup and parts not? I'm a bit confused by what you mean

packages/model/tsconfig.json
```diff
@@ -0,0 +1,40 @@
+{
```
igorT: It might be nice to have a default ts config file that we use to generate/copy these from

packages/store/addon/-private/system/relationships/state/has-many.ts
```diff
@@ -2,17 +2,26 @@ import { assertPolymorphicType } from 'ember-data/-debug';
 import Relationship from './relationship';
 import OrderedSet from '../../ordered-set';
 import { isNone } from '@ember/utils';
-import { RelationshipRecordData } from "../../../ts-interfaces/relationship-record-data";
-import { JsonApiHasManyRelationship, JsonApiResourceIdentity } from "../../../ts-interfaces/record-data-json-api";
-import { RelationshipSchema } from "../../../ts-interfaces/record-data-schemas";
+import { RelationshipRecordData } from '../../../ts-interfaces/relationship-record-data';
```
igorT: What are these changes about?

packages/model/tsconfig.json
runspired: Yes, that was part of the work I started to do with the default blueprints for packages, it would also let us "re-init" packages to bring them up to speed as a syncing mechanism

packages/store/addon/-private/system/relationships/state/has-many.ts
runspired: Prettier doing it's thing

runspired:
@igorT 

> Can you say more about parts of the code using rollup and parts not? I'm a bit confused by what you mean

Currently we only use `rollup` on the `-private` directory of `ember-data` package, but most of the code is in the `@ember-data/store` package. This means we have lost most of the benefits of rolling up and are now producing a lot more modules than before.

We should make all of the packages rollup their private directories, and ideally we should move this to being a prepublish step such that consuming applications get a single `-private` module when they install one of our packages and don't pay the cost of rollup at all.

This will also make things play nicer with `embroider` because embroider will not need to run our build step before being able to construct the graph correctly, as our modules will be in the correct form by the time embroider sees them.

igorT:
Does that increase our build size? Lets open an issue with this context

runspired:
@igorT we should absolutely open an issue with context because yes, it will increase our build size and parse-time if not fixed.

Issue #5937, opened 2019-03-25
Model generated with blueprint doesn't work (with Octane preview)


pgengler:
### Reproduction
https://github.com/pgengler/octane-preview-app/tree/attr-decorators-for-ember-data-models

(Start `ember s` and visit http://localhost:4200/tests to see this in action.)

### Description
Declaring an attribute on a model using the `@attr` decorator doesn't actually set up that attribute on the model if used as a bare decorator; using `@attr()` (or `@attr('string')`, e.g.) seems to be ok.

On a model generated using `ember g model item value`:
```js
import DS from 'ember-data';
const { Model, attr } = DS;

export default class ItemModel extends Model {
  @attr value;
}
```

The error I get from doing `this.store.createRecord('item', { value: 'abc' })` is:
```
TypeError: Cannot assign to read only property 'value' of object '[object Object]'
    at defineProperty (http://localhost:4200/assets/vendor.js:14812:22)
    at initialize (http://localhost:4200/assets/vendor.js:29517:41)
    at Function.create (http://localhost:4200/assets/vendor.js:30060:9)
    at FactoryManager.create (http://localhost:4200/assets/vendor.js:1215:33)
    at InternalModel.getRecord (http://localhost:4200/assets/vendor.js:69232:63)
    at http://localhost:4200/assets/vendor.js:76149:32
    at Backburner._run (http://localhost:4200/assets/vendor.js:53686:25)
    at Backburner._join (http://localhost:4200/assets/vendor.js:53662:21)
    at Backburner.join (http://localhost:4200/assets/vendor.js:53379:19)
    at http://localhost:4200/assets/vendor.js:76130:33
```

### Versions

```cli
$ npm ls ember-source && npm ls ember-cli && npm ls ember-data
octane-app@0.0.0 /Users/pgengler/octane-app
└── ember-source@3.10.0-canary

octane-app@0.0.0 /Users/pgengler/octane-app
└── ember-cli@3.9.0-beta.2  (github:ember-cli/ember-cli#ed51787b3e18efee706f163d57836a30cd5286b7)

octane-app@0.0.0 /Users/pgengler/octane-app
└── ember-data@3.9.0-canary  (github:emberjs/data#435ad9d8b92998fcbe2cee1746612148648280e1)
```

runspired:
cc @pzuraq

pzuraq:
[...]

PR #6054, opened 2019-04-21
Add paranthesis to the Octane attr decorator when there is no data type

JonForest:
Addresses issue https://github.com/emberjs/data/issues/5937

I suspect this may not be how you want to fix this problem - as otherwise you'd have done it by now. However, it just cost me an hour or so stepping through the code wondering what was going on, which not everyone is going to do.  Seems better to 'fix' the blueprint, and then go back to remove the parentheses if ED is updated to detect the `@attr` without them.


Frozenfire92:
This makes sense to me, but I suspect you will also need to update the tests to reflect the new parenthesis

runspired:
Thank you!

PR #6055, opened 2019-04-23
[BUGFIX defaultValue] return null instead of undefined

lsg-braymon:
I have a model with the following definition:
```
class Person extends DS.Model {
  @DS.attr("string")
  name
}
```

When fetching a record via API, the `name` attribute is set to `null` if it doesn't have any value.
However when creating a new record, `const person = store.createRecord("person")`, the `name` attribute is `undefined` making it inconsistent with type checking.


runspired:
@lsg-braymon I was going to suggest that this was a deviation in how we treat this scenario that would require discussion but I notice you've gone ahead and closed it.

lsg-braymon:
@runspired Posted a discussion in discord about this one and it would most likely be a fix in the definitely typed repo rather than in here.

PR #6059, opened 2019-04-23
[CHORE] RFC-0329 - adding deprecation about using evented in ember-data

pete-the-pete:
Deprecating Evented for various ember-data classes.

### DIFF
--- a/packages/-ember-data/tests/helpers/deprecated-test.js
+++ b/packages/-ember-data/tests/helpers/deprecated-test.js
@@ -1,0 +1,44 @@
+import { test } from 'qunit';
+import VERSION from 'ember-data/version';
+
+// small comparison function for major and minor semver values
+function gte(EDVersion, DeprecationVersion) {
+  let _edv = EDVersion.split('.');
+  let _depv = DeprecationVersion.split('.');
+  // compare major
+  let major = +_edv[0] >= +_depv[0];
+  // compare minor
+  let minor = +_edv[1] >= +_depv[1];
+  return major || minor;
+}
+
+export function deprecatedTest(testName, deprecation, testCallback) {
+  // '4.0'
+  if (typeof deprecation.until !== 'string' || deprecation.until.length < 3) {
+    throw new Error(`deprecatedTest expects { until } to be a version.`);
+  }
+  // 'ds.<some-name>'
+  if (typeof deprecation.id !== 'string' || deprecation.id.length < 8) {
+    throw new Error(`deprecatedTest expects { id } to be a meaningful string`);
+  }
+
+  if (gte(VERSION, deprecation.until)) {
+    test(`DEPRECATION ${deprecation.id} until ${deprecation.until} | ${testName}`, testCallback);
+  } else {
+    test(`DEPRECATION ${deprecation.id} until ${
+      deprecation.until
+    } | ${testName}`, function(assert) {
+      if (deprecation.refactor === true) {
+        assert.ok(
+          false,
+          'This test includes use of a deprecated feature that should now be refactored.'
+        );
+      } else {
+        assert.ok(
+          false,
+          'This test is for a deprecated feature whose time has come and should be removed'
+        );
+      }
+    });
+  }
+}
--- a/packages/-ember-data/tests/helpers/deprecated-test.js
+++ b/packages/-ember-data/tests/helpers/deprecated-test.js
@@ -25,9 +25,7 @@ export function deprecatedTest(testName, deprecation, testCallback) {
   if (gte(VERSION, deprecation.until)) {
     test(`DEPRECATION ${deprecation.id} until ${deprecation.until} | ${testName}`, testCallback);
   } else {
-    test(`DEPRECATION ${deprecation.id} until ${
-      deprecation.until
-    } | ${testName}`, function(assert) {
+    test(`DEPRECATION ${deprecation.id} until ${deprecation.until} | ${testName}`, function(assert) {
       if (deprecation.refactor === true) {
         assert.ok(
           false,
--- a/packages/-ember-data/tests/unit/model/errors-test.js
+++ b/packages/-ember-data/tests/unit/model/errors-test.js
@@ -1,6 +1,6 @@
 import DS from 'ember-data';
-import QUnit, { module } from 'qunit';
-import testInDebug from 'dummy/tests/helpers/test-in-debug';
+import QUnit, { module, skip as testInDebug } from 'qunit';
+// import testInDebug from 'dummy/tests/helpers/test-in-debug';
 
 const AssertPrototype = QUnit.assert;
 
--- a/packages/-ember-data/tests/unit/model/errors-test.js
+++ b/packages/-ember-data/tests/unit/model/errors-test.js
@@ -1,6 +1,6 @@
 import DS from 'ember-data';
-import QUnit, { module, skip as testInDebug } from 'qunit';
-// import testInDebug from 'dummy/tests/helpers/test-in-debug';
+import QUnit, { module } from 'qunit';
+import testInDebug from 'dummy/tests/helpers/test-in-debug';
 
 const AssertPrototype = QUnit.assert;
 
--- a/packages/store/addon/-private/system/deprecated-evented.js
+++ b/packages/store/addon/-private/system/deprecated-evented.js
@@ -43,8 +43,8 @@ if (DEBUG) {
         ? `Called ${eventName} ${_deprecationData}`
         : eventName;
       deprecate(deprecationMessage, deprecations[eventName], {
-        id: 'ember-evented',
-        until: '3.12',
+        id: 'ember-data:evented-api-usage',
+        until: '4.0',
       });
       deprecations[eventName] = true;
     },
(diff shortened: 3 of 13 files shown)


packages/store/addon/-private/system/model/model.js
```diff
@@ -1193,6 +1194,10 @@ if (DEBUG) {
     init() {
       this._super(...arguments);
 
+      if (DEBUG) {
+        this._getDeprecatedEventedInfo = () => `${this.name}#${this.id}`;
```
runspired: `name` is probably the wrong thing here, `_internalModel.modelName` is likely what was intended.

packages/store/addon/-private/system/deprecated-evented.js
```diff
@@ -0,0 +1,72 @@
...
+
+    return deprecations;
+  };
+
+  DeprecatedEvented = Mixin.create(Evented, {
+    _has(name) {
+      return Evented.mixins[0].properties.has.call(this, name);
+    },
+
+    deprecrateEvented(eventName) {
```
runspired: the methodName here is misspelled (`deprecate` not `deprecrate`). Later we also attempt to invoke this as `deprecatedEvented()`. We should probably make it `_deprecateEvent(eventName)` as the signature.

packages/store/addon/-private/system/record-arrays/adapter-populated-record-array.js
```diff
@@ -84,6 +90,9 @@ export default RecordArray.extend({
 
     this.manager._associateWithRecordArray(internalModels, this);
 
+    if (DEBUG && this._has('didLoad')) {
+      this.deprecatedEvented('didLoad');
```
runspired: the method name is incorrect

pete-the-pete:
@runspired updated `deprecatedTest` but just using a barebones comparison function. I saw that we are using semver, but that's only available in node...and I didn't know if it would be worth adding another dependency. I think I got all the relevant tests, but I'll take another look tomorrow.

pete-the-pete:
🤦‍♂ finally realized that this is failing production tests...looking into that now. Also 👍 to having tests that run on a production build.

pete-the-pete:
@hjdivad I keep getting a failed test with ember-m3:
`not ok 140 Chrome 74.0 - [23 ms] - unit/model: .save errors getting updated via the store and removed upon setting a new value`

Running the tests locally with `yarn test-external:ember-m3` I don't get any test failures, any ideas?

hjdivad:
@pete-the-pete have you tried clearing out `../__external-test-cache` and `../__tarball-cache`?

FWIW I do get failures when i `yarn test-external:ember-m3` after `node ./bin/packages-for-commit.js'

pete-the-pete:
I'll try, thanks.

pete-the-pete:
@hjdivad tried again, no _test_ failures, but I do see this:
Screen Shot 2019-05-15 at 9 15 14 PM

runspired:
@pete-the-pete usually I find this means the packaging step produced bad packages. If you run that command with DEBUG=test-external as the env you will get a detailed log with info about what might have gone wrong.

pete-the-pete:
😢 still beyond my understanding...
Screen Shot 2019-05-16 at 2 23 04 PM

pete-the-pete:
thanks @runspired! i learned™, and now i do have test failures!

runspired:
I looked at what the test failures in M3 were and I suspect the issue is that we now access some of the "Event" properties to check for presence of methods in a way we did not before, which is trapped by M3's proxy handler. We may want to guard our checks to only apply to instances of `DS.Model`.

packages/store/addon/-private/system/model/internal-model.ts
```diff
@@ -301,6 +320,31 @@ export default class InternalModel {
...
+          'becameError',
+          'becameInvalid',
+          'didCreate',
+          'didDelete',
+          'didLoad',
+          'didUpdate',
+          'ready',
+          'rolledBack',
+        ].forEach(methodName => {
+          if (typeof this._record[methodName] === 'function') {
```
runspired: I believe this is where we need to guard this check behind an `instanceof DS.Model` check

runspired: Although @hjdivad may have other thoughts? The TL;DR here is this deprecation check causes non DS.Model instances (like M3's) to trap this access and treat the methodName as an attribute.

pete-the-pete: with that change i only get 1 failure:
```
not ok 140 Chrome 74.0 - [22 ms] - unit/model: .save errors getting updated via the store and removed upon setting a new value
    ---
        actual: >
            true
        expected: >
            false
        stack: >
                at http://localhost:7357/assets/tests.js:3952:18
                at tryCatcher (http://localhost:7357/assets/vendor.js:77288:21)
                at invokeCallback (http://localhost:7357/assets/vendor.js:77461:33)
                at publish (http://localhost:7357/assets/vendor.js:77447:9)
                at publishRejection (http://localhost:7357/assets/vendor.js:77383:5)
                at http://localhost:7357/assets/vendor.js:71821:16
        message: >
            model done saving
        negative: >
            false
        browser log: |
```

hjdivad: > The TL;DR here is this deprecation check causes non DS.Model instances (like M3's) to trap this access and treat the methodName as an attribute.

Checking `instanceof Model` seems ok here.

FWIW M3 models are not `Proxy`s.  The reason the property access is trapped is because Ember is adding the proxy wrapper in dev to catch es5 property access across Ember proxies (via `unknownProperty`).

pete-the-pete: @hjdivad can you explain this more? especially the 'adding the proxy wrapper'?

> > 
> FWIW M3 models are not `Proxy`s. The reason the property access is trapped is because Ember is adding the proxy wrapper in dev to catch es5 property access across Ember proxies (via `unknownProperty`).

hjdivad: @pete-the-pete In dev mod Ember will give you back native proxies (and since you're mostly looking in dev you'll mostly see these proxies).

The reason for this is to throw an error if you try to access a property via a native property accessor instead of `Ember.get` when doing so cannot work in production.  I forget the exact rules, but broadly it's if you access a property from an object that has no descriptor, but still has a value computed via `unknownProperty`.

The reason this cannot work in production is twofold.  In the first place, native proxies aren't supported in IE11 and there's no way to polyfill them.  The only reason Ember can support something like a proxy is by requiring an opt-in at every call site (ie `Ember.get`).

The second reason is that there are (potentially?) issues with using `Proxy` in production for this purpose anyway. Here I'm less hazy on the details but a) performance concerns and b) it's not clear that using Proxies in this way is intended usage and it's not clear that doing so won't cause breakage in the future, especially with the introduction of private fields.  This is discussed somewhat in the [tracked properties rfc](https://github.com/emberjs/rfcs/blob/master/text/0410-tracked-properties.md#we-could-wait-on-private-fields-and-proxy-developments) which has links to further discussion.  A word of warning: those links lead to a very deep rabbit hole.

pete-the-pete: 🙇 thanks for the good explanation.

I don't mind rabbit holes. 🔦 ⛑

packages/-ember-data/tests/unit/model/errors-test.js
```diff
@@ -1,6 +1,6 @@
 import DS from 'ember-data';
-import QUnit, { module } from 'qunit';
-import testInDebug from 'dummy/tests/helpers/test-in-debug';
+import QUnit, { module, skip as testInDebug } from 'qunit';
+// import testInDebug from 'dummy/tests/helpers/test-in-debug';
```
igorT: This seems like a leftover

packages/store/addon/-private/system/model/errors.js
```diff
@@ -225,7 +228,10 @@ export default ArrayProxy.extend(Evented, {
     this._add(attribute, messages);
 
     if (wasEmpty && !get(this, 'isEmpty')) {
-      this.trigger('becameInvalid');
+      this._registeredHandlers && this._registeredHandlers.becameInvalid();
+      if (DEBUG && this._has('becameInvalid')) {
```
igorT: Why do we only trigger this in debug mode?

packages/store/addon/-private/system/record-arrays/record-array.js
```diff
@@ -2,11 +2,13 @@
   @module ember-data
 */
 
-import Evented from '@ember/object/evented';
+//import Evented from '@ember/object/evented';
```
igorT: Seems like a leftover

packages/store/addon/-private/system/model/errors.js
runspired: I believe there was a reason around it wouldn't trigger in prod or triggers via a different mechanism in prod, will double check

packages/-ember-data/tests/helpers/deprecated-test.js
```diff
@@ -0,0 +1,42 @@
+import { test } from 'qunit';
```
igorT: ❤️

Issue #6058, opened 2019-04-23
Ember-Data 3.9.1 adapter "headers" seems to be not getting sent.


nikhilsane:
# Ember-Data 3.9.1 JSONAPIAdapter does not seem to send headers.

Hi Ember-Data team,

I have following Adapter in my application:
```
export default DS.JSONAPIAdapter.extend({
    namespace: '/foo/resources/v2.0',
    init() {
        this._super(...arguments);

        this.set('headers', {
            'Accept': 'application/json',
            'ApplicationQuery': '1'
        })
    }
});
```
When using **v3.9.1**, the "findAll" query sends the following header in the request:

> Request headers:
> Host: localhost:9000
> User-Agent: Mozilla/5.0 (Windows NT 10.0; Win64; x64; rv:66.0) Gecko/20100101 Firefox/66.0
> **_Accept: application/vnd.api+json_**
> Accept-Language: en-US,en;q=0.5
> Accept-Encoding: gzip, deflate
> Content-Type: application/vnd.api+json
> ApplicationQuery: 1
> X-Requested-With: XMLHttpRequest
> Connection: keep-alive
> Cookie: JSESSIONID=02E588F5129EFD7540B86F6F526D2371
> 

When using **v3.5.1**, the "findAll" query that is sent to the back-end API looks like the following:

> Request headers:
> Host: localhost:9000
> User-Agent: Mozilla/5.0 (Windows NT 10.0; Win64; x64; rv:66.0) Gecko/20100101 Firefox/66.0
> _**Accept: application/json**_
> Accept-Language: en-US,en;q=0.5
> Accept-Encoding: gzip, deflate
> X-Requested-With: XMLHttpRequest
> Connection: keep-alive
> Cookie: JSESSIONID=487615003A26F1141D0F22B9354F409B
> 
> 

So, as can be seen from the above outputs, the same code does not seem to work in v 3.9.1. Is this expected behavior? Am I missing something here?

### Versions

Following are the versions of the packages: ember-source, ember-cli, ember-data

```
`-- ember-source@3.7.0
`-- ember-cli@3.1.4
`-- ember-data@3.9.1
```

--------------------------------------------------------------
Thanks,

Nikhil.

runspired:
Hi Nikhil,

This does appear to be a regression.[ This line here](https://github.com/emberjs/data/blob/master/packages/adapter/addon/json-api.js#L156) needs to be changed to the following

```js
[...]

PR #6062, opened 2019-04-24
[BUGFIX release] Fixing adapter "Accept" header overwriting issue.

nikhilsane:
Fixing the bug that caused overwriting of "Accept" header in "ajaxOptions" method of the adapter. This is for addressing the issue #6058 

Change description:

1. Updated the code in "packages\adapter\addon\json-api.js". If the input hash contains value for "Accept" header, it should not be overwritten by the default "application/vnd.api+json" string.

2. Added unit tests for checking this flow.

Fixes https://github.com/emberjs/data/issues/6058

### DIFF
--- a/packages/adapter/addon/json-api.js
+++ b/packages/adapter/addon/json-api.js
@@ -153,10 +153,11 @@ const JSONAPIAdapter = RESTAdapter.extend({
     @return {Object}
   */
   ajaxOptions(url, type, options = {}) {
-    options.contentType = 'application/vnd.api+json';
+    options.contentType = options.contentType || 'application/vnd.api+json';
 
     let hash = this._super(url, type, options);
-    hash.headers['Accept'] = 'application/vnd.api+json';
+    hash.headers['Accept'] = hash.headers['Accept'] || 'application/vnd.api+json';
+
     return hash;
   },
 
--- a/packages/-ember-data/tests/unit/adapters/json-api-adapter/ajax-test.js
+++ b/packages/-ember-data/tests/unit/adapters/json-api-adapter/ajax-test.js
@@ -84,3 +84,22 @@ test('ajaxOptions() adds Accept header to existing computed properties headers',
     'headers assigned'
   );
 });
+
+test('ajaxOptions() does not overwrite Accept header if it is provided', function(assert) {
+  let url = 'example.com';
+  let type = 'GET';
+  adapter.headers = { Accept: 'application/json' };
+  let ajaxOptions = adapter.ajaxOptions(url, type, {});
+  let receivedHeaders = [];
+  let fakeXHR = {
+    setRequestHeader(key, value) {
+      receivedHeaders.push([key, value]);
+    },
+  };
+  ajaxOptions.beforeSend(fakeXHR);
+  assert.deepEqual(
+    alphabetize(receivedHeaders),
+    [['Accept', 'application/json']],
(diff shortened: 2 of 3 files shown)


package.json
```diff
@@ -89,5 +89,8 @@
     "rimraf": "^2.6.2",
     "rsvp": "^4.8.4",
     "typescript": "~3.4.3"
+  },
+  "dependencies": {
+    "yarn": "^1.15.2"
```
makepanic: Looks like you've accidentally added yarn as a package dependency. Can you revert the package.json and yarn.lock change?

packages/adapter/addon/json-api.js
```diff
@@ -156,7 +156,7 @@ const JSONAPIAdapter = RESTAdapter.extend({
     options.contentType = 'application/vnd.api+json';
 
     let hash = this._super(url, type, options);
-    hash.headers['Accept'] = 'application/vnd.api+json';
+    hash.headers['Accept'] = options.Accept || 'application/vnd.api+json';
```
makepanic: Imo what runspired proposed would solve this better, something like:

```
    options.contentType = options.contentType || 'application/vnd.api+json';

    let hash = this._super(url, type, options);
    hash.headers['Accept'] = options.contentType;
```

package.json
runspired: confirm, we do not want to change yarn deps here

packages/adapter/addon/json-api.js
runspired: We don't want to mix `accept` and `contentType`. I made a mistake in the issue when linking to `contentType` because I failed to notice the issue was for `accept`.

That said, we should fix both:

```js
options.contentType = option.contentType || 'application/vnd.api+json';

let hash = this._super(url, type, options);
hash.headers['Accept'] = hash.headers['Accept'] || 'application/vnd.api+json';
```

package.json
nikhilsane: Oh, yes. that's right. I will remove these changes.

packages/adapter/addon/json-api.js
nikhilsane: Thanks @makepanic for the review. 

 I see. However, I am a little confused here - may be I am missing something, but using the value of "options.contentType" for "hash.headers['Accept'] may be confusing for the calling function.

Consider the following invocation of this method:

`adapter.ajaxOptions('example.com', 'GET', {'Accept': 'application/json'});`

In this case, only the 'Accept' header is being sent - 'contentType' is not being sent. So, the following code will set the hash.headers['Accept'] to 'application/vnd.api+json', even when the caller has specified 'application/json' as the value for the Accept header.
 
```
options.contentType = options.contentType || 'application/vnd.api+json';
let hash = this._super(url, type, options);
hash.headers['Accept'] = options.contentType;

```

runspired:
Appears there is a linting error, but otherwise lgtm 👍

nikhilsane:
@runspired : Fixed the linting issues and pushed the code.
I am not aware of the process - what happens next?

Thanks,
nikhil.

rwjblue:
FYI - I'm labeling this as `[BUGFIX release]` as this is impacting folks in the 3.9 release.

rwjblue:
I rebased and fixed the remaining linting issue.

nikhilsane:
Thanks!

-nikhil.

Issue #6049, opened 2019-04-20
`release` branch is not ignoring tests it should ignore


runspired:
external partner tests, `release|beta|canary` tests of ember should not run for CI runs for release branch.

mbinet:
Hello @runspired!

I've had a look at this issue and I'm not sure to understand.

Indeed, see the last build on the release branch: https://travis-ci.org/emberjs/data/builds/523223535
`external partner tests`, `Ember Release`, `Ember Beta` and `Ember Canary` do not run.

Could you give me additional informations please?

runspired:
@mbinet I noticed this issue here https://travis-ci.org/emberjs/data/builds/522353527

It may be specific to the tag vs the branch.

mbinet:
Oh ok I see!

Yes it does seem to be related to the tags.

So to clarify, we do **not** want to run these tests on tags, right?

runspired:
@mbinet confirm

runspired:
@mbinet in fact I see no reason to run tests for pushes of tags at all, since they always reflect a branch at the time they are created (beta or release)

PR #6063, opened 2019-04-24
Ignore unneeded jobs on tag push

mbinet:
* Disable `external partner tests`, `Linting`, `Enabled In-Progress Features`, `Ember Release`, `Ember Beta` & `Ember Canary` tests on tag push.

Fixes https://github.com/emberjs/data/issues/6049

You can see in my fork that all the mentioned tests do run on master: https://travis-ci.com/mbinet/data/builds/109345304
but not when I push a tag: https://travis-ci.com/mbinet/data/builds/109345324

@runspired You don't ask for it in the issue but I removed `Linting` & `Enabled In-Progress Features` as well, what do you think about that?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,7 +27,7 @@ stages:
   - additional tests
   - ember version tests
   - name: external partner tests
-    if: NOT (branch ~= /^(emberjs:release|emberjs:lts|release|lts).*/)
+    if: NOT tag IS present AND NOT (branch ~= /^(emberjs:release|emberjs:lts|release|lts).*/)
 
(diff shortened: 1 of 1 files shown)


runspired:
@mbinet this looks great! Thank you :) I'm slightly conflicted on whether we want to allow tags to build at-all. But seems ok for now to at least let them run the same tests as the release/beta branch does when they are created.

Reason I am conflicted is that whenever a tag is created and generated there will be a commit for the version to the respective branch that runs the identical tests, seems wasteful to run twice for the same code.

PR #6064, opened 2019-04-24
[BUGFIX partner-tests] fix directory resolution

runspired:


runspired:
replaced with #6065 so that test runs will be faster

packages/-build-infra/bin/test-external-partner-project.js
```diff
@@ -8,12 +8,13 @@ const { shellSync } = require('execa');
 // apparently violates no-extraneous require? /shrug
 const debug = require('debug')('test-external');
 const rimraf = require('rimraf');
+const chalk = require('chalk');
 
-const projectRoot = path.resolve(__dirname, '../');
+const projectRoot = path.resolve(__dirname, '../../../');
```
igorT: This is true even when run from project root?

runspired: yes

packages/-build-infra/bin/test-external-partner-project.js
```diff
@@ -32,31 +33,90 @@ console.log(
...
 
   return shellSync(command);
 }
 
+function generatePackageConnectionCommand(useYarn, shouldLink = true) {
+  const packageManager = useYarn ? 'yarn' : 'npm';
+  const command = shouldLink ? `${packageManager} link` : `${packageManager} unlink`;
+  const packageConnectionCommand = `
+cd ${projectRoot};
+${shouldLink ? (useYarn ? 'yarn install --force;' : 'npm install;') : ''}
```
igorT: Why do we need force for yarn?

runspired: If you've just unlinked a package yarn doesn't understand it needs to replace it.

runspired: (This is in yarn's own guidance for what to do following unlink)

bin/-tarball-info.js
```diff
@@ -0,0 +1,52 @@
+'use strict';
+/* eslint-disable no-console, node/no-extraneous-require, node/no-unpublished-require */
+const fs = require('fs');
+const path = require('path');
+
+const projectRoot = path.resolve(__dirname, '../');
```
igorT: Can you write a short comment explaining what the overall purpose of this file is

azure-pipelines.yml
```diff
@@ -4,246 +4,241 @@
 # https://aka.ms/yaml
 
 trigger:
-- master
```
igorT: Are these just formatting?

.travis.yml
```diff
@@ -118,6 +118,7 @@ before_install:
 
 install:
   - yarn install
+  - node ./bin/packages-for-commit.js
```
igorT: How come we don't use yarn for running this script?

bin/packages-for-commit.js
```diff
@@ -0,0 +1,47 @@
+#!/usr/bin/env node
```
igorT: Also for this file

bin/test-external-partner-project.js
```diff
@@ -32,11 +35,12 @@ console.log(
 );
 
 function execExternal(command, force) {
-  command = `cd ${cachePath}/${externalProjectName} && ${command}`;
+  command = `cd ${projectTempDir} && ${command}`;
   return execWithLog(command, force);
 }
 
 function execWithLog(command, force) {
+  debug(chalk.cyan('Executing: ') + chalk.yellow(command));
```
igorT: :heart:

bin/test-external-partner-project.js
```diff
@@ -47,16 +51,20 @@ function execWithLog(command, force) {
...
   fs.mkdirSync(tempDir);
+} else {
+  debug(`Cache Root Exists at: ${tempDir}`);
 }
 
 if (fs.existsSync(projectTempDir)) {
   debug(`Cleaning Cache at: ${projectTempDir}`);
   rimraf.sync(projectTempDir);
+} else {
+  debug(`No pre-existing cache present at: ${projectTempDir}`);
```
igorT: :+1:

bin/test-external-partner-project.js
```diff
@@ -113,19 +86,32 @@ try {
     debug('Skipping Smoke Test');
   } else {
     debug('Running Smoke Test');
+    try {
+      execExternal(`${useYarn ? 'yarn install' : 'npm install'}`);
```
igorT: How come we first install and then modify the package. Json?

bin/test-external-partner-project.js
```diff
@@ -113,19 +86,32 @@ try {
...
 try {
-  execExternal(`${useYarn ? 'yarn link ember-data' : 'npm link ember-data'}`);
+  debug('Preparing Package To Run Tests Against Commit');
+  insertTarballsToPackageJson(packageJsonLocation);
+  // we must use npm because yarn fails to install
+  // the nested tarballs correctly (just as it fails to pack them correctly)
+  // we rimraf node_modules first because it was previously
+  // installed using yarn's resolution algorithm
+  execExternal(`rm -rf node_modules`);
+  execExternal(`npm install`);
```
igorT: If we do this does the initial install make a difference? Can we rm rf the ed folders instead of the whole node modules?

bin/test-external-partner-project.js
runspired: For the smoke-test we use their setup first

bin/test-external-partner-project.js
runspired: sadly it has to be the whole of node_modules because otherwise the `npm install` which is required to install from tarball correctly will screw up some of the resolutions of some of the other dependencies

.travis.yml
runspired: yarn workspaces has a bug with bin files in the root package not getting put into the path correctly. I hoisted these files to root because they are run on all packages and seemed to be messing with tarball generation when inside a package we were packing (I'm not sure they actually were, but architecturally they made more sense hoisted anyway so I went ahead and pulled them out just in case, ultimately `yarn pack` just wan't working and `npm pack` is)

runspired: I suspect I can move it to yarn by deleting my local install entirely and doing a fresh checkout to force the bin files to regenerate, as I think the issue I encountered with bit files in root of a workspace is only for updates and not fresh checkouts.

azure-pipelines.yml
runspired: Seem to be, I didn't notice this change. Prettier must have reformatted things when I added the package generation command.

bin/-tarball-info.js
runspired: Added docs

bin/packages-for-commit.js
runspired: Added docs

.travis.yml
runspired: Tried that and bin files still seem to be borked so going to leave as is for now

azure-pipelines.yml
runspired: I had to rebase and in the process had to redo this change and can confirm that it's because prettier does not like what we have on master currently. I'd recommend configuring prettier on your machine to run on file save.

runspired:
Travis failure isn't due to this.

igorT:
I'm not editing the file locally remember. Fixing later seems fine

PR #6065, opened 2019-04-24
[BUGFIX partner-tests] fix directory resolution

runspired:


runspired:
apparently PRing from a fork with TravisCI activated still runs the tests under emberjs's account :'(

PR #6069, opened 2019-04-26
Set up CI with Azure Pipelines

igorT:


runspired:
Can we setup IE and Windows jobs on azure as well to kill off our appveyor usage?

igorT:
@runspired Given how slow Travis is, don't think there is a need for waiting, as if the extra tests fail(which they shouldn't) we can ignore them for now, and we don't publish beta/release versions often enough for that to block this PR IMO

igorT:
@runspired yeah, hopefully, I'll look into how appveyor is setup sometime later

igorT:
I'll look into not running the external partner tests and in progress features as a followup task

runspired:
> Given how slow Travis is, don't think there is a need for waiting, as if the extra tests fail(which they shouldn't) we can ignore them for now

This is a separate point from that in my review which I thought you were going to include a fix for here in which we fail fast if `lint` and `basic tests` fail?

igorT:
Investigating couple different solutions for that, will come up in a follow on PR

PR #6070, opened 2019-04-26
azure tests fail fast

igorT:


igorT:
verified it fails fast

PR #6072, opened 2019-04-26
Run windows tests on azure

igorT:


runspired:
We should have both a mac and a windows set of jobs, yes? Or do we just want to run on windows? cc @rwjblue

igorT:
This is still a WIP, currently appyveyor runs 3/4 tests on windows, so running those

igorT:
ready for review @runspired

igorT:
given that we develop on macs, I think we should run most of the tests on OS X

runspired:
So stoked to kill appveyor

PR #6073, opened 2019-04-29
[DOC] fix JSONAPISerializer response

hedrovich:
Remove useless link namespace in documentation

### DIFF
--- a/addon/serializers/json.js
+++ b/addon/serializers/json.js
@@ -44,9 +44,7 @@ const emberAssign = assign || merge;
     id: 1,
     name: 'Sebastian',
     friends: [3, 4],
-    links: {
-      house: '/houses/lefkada'
-    }
+    house: '/houses/lefkada'
   }
   ```
 


bmac:
Hi @hedrovich. The docs you edited are for the `JSONSerializer` which confusingly accepts a different format from the `JSONAPISerializer` you mentioned in your commit message. I believe the current documentation describing the `links` object is correct if someone is using the `JSONSerializer` to extract their API responses.

`JSONSerializer` is some legacy code that predates the JSON API spec. I wonder if there is some additional disclaimer we could add here to point out that it does not conform to the JSON API spec.

runspired:
@hedrovich @bmac I believe this change may actually be correct but agree the title is misleading.

The change was to the `JSONSerializer` while the title suggests them being for the `JSONAPISerializer`.

The `JSONSerializer` expects a flat structure and produces a flat structure. It is indeed legacy and does indeed predate the json-api spec and not conform to it.  I would have to investigate how it handles links, but I doubt it expects `json-api` formatted links which is what the documentation currently suggests.

PR #6074, opened 2019-05-01
[BUGFIX release] build packages and handle OTP timeout

runspired:
* Ensures that `yarn build:production` is run for each "addon" package
* Separates out versioning from packaging and publishing
* Updates the RELEASE documentation
    
    We require 2fa to publish ember-data, but 2fa tokens expire after 30s.
    `lerna publish` which handles versioning, packaging, and publishing
    does not itself accept an `OTP` for 2fa and when passed as an environment
    variable it is likely to have long expired prior to reaching the publish
    step.
    
    Our script separates out these steps so that we can recover from publishing
    errors more gracefully, and surfaces the OTP prompt (and reprompt) at
    the correct point in time for the publish.
    
    The opportunity was taken here to add guard rails around many of the finer
    points of our release process. Outside of branching that occurs at the
    beginning of the release cycle and changelog updates our release process
    is not mostly automated.

Inspiration from https://github.com/glimmerjs/glimmer-vm/commit/01e68d7dddf28ac3200f183bffb7d520a3c71249#diff-19fef6f3236e72e3b5af7c884eef67a0

### DIFF
--- a/bin/publish.js
+++ b/bin/publish.js
@@ -1,0 +1,344 @@
+#!/usr/bin/env node
+
+'use strict';
+/* eslint-disable node/no-unsupported-features/es-syntax, no-console, no-process-exit, node/no-extraneous-require, node/no-unpublished-require */
+
+/*
+Usage
+
+publish lts|release|beta|canary
+
+Flags
+
+--distTag=latest|lts|beta|canary|release-<major>-<minor>
+--bumpMajor
+--bumpMinor
+--skipVersion
+--skipPack
+--skipPublish 
+--skipSmokeTest
+
+Inspiration from https://github.com/glimmerjs/glimmer-vm/commit/01e68d7dddf28ac3200f183bffb7d520a3c71249#diff-19fef6f3236e72e3b5af7c884eef67a0
+*/
+
+const debug = require('debug')('publish-packages');
+const chalk = require('chalk');
+const fs = require('fs');
+const path = require('path');
+const { shellSync } = require('execa');
+const cliArgs = require('command-line-args');
+const readline = require('readline');
+const semver = require('semver');
+const projectRoot = path.resolve(__dirname, '../');
+const packagesDir = path.join(projectRoot, './packages');
+const packages = fs.readdirSync(packagesDir);
+
+function cleanProject() {
+  execWithLog(
+    `cd ${projectRoot} && rm -rf packages/*/dist packages/*/tmp packages/*/node_modules node_modules`
+  );
+  execWithLog(`cd ${projectRoot} && yarn install`);
+}
+
+/**
+ *
+ * @param {*} command The command to execute
+ * @param {*} proxyIO whether to proxy stdio from the main process for this command
+ *
+ * proxyIO=true is useful when you want to see the output log or respond to prompts
+ */
+function execWithLog(command, proxyIO = false) {
+  debug(chalk.cyan('Executing: ') + chalk.yellow(command));
+  if (proxyIO) {
+    return shellSync(command, { stdio: [0, 1, 2] });
+  }
+
+  return shellSync(command).stdout;
+}
+
+function getConfig() {
+  const mainOptionsDefinitions = [{ name: 'channel', defaultOption: true }];
+  const mainOptions = cliArgs(mainOptionsDefinitions, { stopAtFirstUnknown: true });
+  const argv = mainOptions._unknown || [];
+
+  if (!mainOptions.channel) {
+    throw new Error(`Incorrect usage of publish:\n\tpublish <channel>\n\nNo channel was specified`);
+  }
+  if (!['release', 'beta', 'canary', 'lts'].includes(mainOptions.channel)) {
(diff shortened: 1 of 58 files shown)


igorT:
What does build packages refer to? What is the OTP timeout?

package.json
```diff
@@ -81,17 +81,24 @@
...
+    "semver": "^6.0.0"
   },
   "bin": {
     "test-partner-project": "./bin/test-external-partner-project.js",
-    "prepare-local-packages": "./bin/packages-for-commit.js"
+    "prepare-local-packages": "./bin/packages-for-commit.js",
+    "publish": "./bin/publish.js"
+  },
+  "dependencies": {
+    "command-line-args": "^5.1.1"
```
igorT: Should this be a dev-dependency instead?

runspired: likely, I intended it to be but it snuck up here. I don't think it matters since this is root but better to be safe.

bin/publish.js
```diff
@@ -0,0 +1,327 @@
...
+const packages = fs.readdirSync(packagesDir);
+
+function cleanProject() {
+  execWithLog(
+    `cd ${projectRoot} && rm -rf packages/*/dist packages/*/tmp packages/*/node_modules node_modules`
+  );
+  execWithLog(`cd ${projectRoot} && yarn install`);
+}
+
+function execWithLog(command, proxyIO = false) {
```
igorT: Can you leave a comment explaining what proxyIO is used for

bin/publish.js
```diff
@@ -0,0 +1,327 @@
...
+  return shellSync(command).stdout;
+}
+
+function getConfig() {
+  const mainOptionsDefinitions = [{ name: 'channel', defaultOption: true }];
+  const mainOptions = cliArgs(mainOptionsDefinitions, { stopAtFirstUnknown: true });
+  const argv = mainOptions._unknown || [];
+
+  if (!mainOptions.channel) {
+    throw new Error(`Incorrect usage of publish:\n\tpublish <channel>\n\nNo channel was specified`);
```
igorT: 👍

bin/publish.js
```diff
@@ -0,0 +1,327 @@
...
+  options.channel = mainOptions.channel;
+  options.currentVersion = currentProjectVersion;
+
+  return options;
+}
+
+const options = getConfig();
+
+function assertGitIsClean() {
+  let status = execWithLog('git status');
```
igorT: ❤️

bin/publish.js
```diff
@@ -0,0 +1,327 @@
...
+/* eslint-disable node/no-unsupported-features/es-syntax, no-console, no-process-exit, node/no-extraneous-require, node/no-unpublished-require */
+
+/*
+Usage
+
+publish release|beta|canary
+
+Flags
+
+--distTag=latest|lts|beta|canary|release-<major>-<minor>
```
igorT: Whats latest?

bin/publish.js
```diff
@@ -0,0 +1,327 @@
...
+function retrieveNextVersion() {
+  /* 
+  #master lerna.json 3.11.0-canary.x
+    releases with `canary`
+  #beta lerna.json 3.10.0-beta.x
+    cuts from last 3.10.0-canary.x master with `beta`
+  #release lerna.json 3.9.0
+    cuts from last 3.9.0-beta.x
+  #lts lerna.json 3.8.x
+     custs from last 3.8.x on release
```
igorT: cuts

bin/publish.js
```diff
@@ -0,0 +1,327 @@
...
+          chalk.yellow('⚠️  Publishing from an incorrect branch may result in a broken release ⚠️')
+      );
+      process.exit(1);
+    }
+  }
+}
+
+function retrieveNextVersion() {
+  /* 
+  #master lerna.json 3.11.0-canary.x
```
igorT: I am having a bit of trouble understanding what this comment means

igorT:
Seems good to me, but would love @rwjblue to take a look as well given that we don't have tests for this

bin/publish.js
```diff
@@ -0,0 +1,327 @@
+#!/usr/bin/env node
```
igorT: Should we link the glimmer file in a comment as the source for this?

bin/publish.js
runspired: `latest` is the latest stable release. It is what get's installed by `yarn` or `npm` if you do not otherwise specify a version or tag. It is also the tag that gets published to npm if we do not specify one.

bin/publish.js
runspired: It's a brief rundown of how version updates flow through the branches. Basically we only ever bump the major or minor version on `master`, and all other branches pick it up as those changes flow through the release cycle.

bin/publish.js
runspired: I did take some inspiration, so why not :)

bin/publish.js
runspired: done

PR #6077, opened 2019-05-01
[BUGFIX release] Remove ember-fetch dependency.

rwjblue:
This change removes ember-fetch as a direct dependency of ember-data, and allows for this dependency to be provided by the host application or for users to directly use native fetch.

Having `ember-fetch` as a direct dependency of ember-data has proven troublesome. There are issues using ember-fetch as a nested dependency (these are documented in the ember-fetch README), but it is also quite difficult for ember-data to ensure that the versions of `ember-fetch` provided will actually include the correct version (duplicated addon merging in the ember-cli addon space is complicated and error prone).

This change moves the `determineBodyPromise` and `serializeQueryParams` helper functions directly into `@ember-data/adapter`. A future version of ember-fetch will drop these utilities (as well as the ember-data adapter mixin). In addition, this also enables usage of the global `fetch` if a `fetch` module is not provided.

https://github.com/ember-cli/ember-fetch/issues/289 has been created to track removing the ember-data support code from ember-fetch.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -80,6 +80,10 @@ jobs:
           yarn test:node
         displayName: 'Node Tests'
 
+      - script: |
+          yarn test:try-one with-ember-fetch
+        displayName: 'Basic Tests with ember-fetch'
+
   - job: Windows_tests
     dependsOn: Basic_Ember_Data_tests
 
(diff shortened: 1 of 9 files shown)


Turbo87:
this change is causing build warnings:
```
'require' is imported by ../../../../../var/folders/6y/_8qpkt114jzb442txt_9w_2c0000gn/T/broccoli-27019nKpnRWsSqjWT/cache-596-rollup/build/-private/utils/fetch.js, but could not be resolved – treating it as an external dependency
```

PR #6078, opened 2019-05-01
Use HTML comments in PR template.

rwjblue:
Using HTML comments here mean they don't need to delete anything
and simplifies the process a tad bit...


runspired:
@rwjblue this does not seem to have had the intended effect:

```
<!--

If this is your first PR to `ember-data`, you may want to read our [Contributor Guide](https://github.com/emberjs/data/blob/master/CONTRIBUTING.md).

-->
```

^ this is what now gets manually deleted

runspired:
(That or I misunderstood something about what this PR aimed to achieve)

PR #6086, opened 2019-05-07
[FEAT serializers] clean up packages organization

runspired:
**Default Serializers**

Previously several `-json` and `-defaults` serializers corresponding to the `JSONSerializer` and `JSONAPISerializer` were added to applications via `inject`. These are now added via a re-export in the `app/` directory, making them overrideable and easier to drop.

**Transforms**

The RFC for packages intentionally did not provide import paths for the transforms we provide (`date`, `number` `boolean` `string`). A more ideal location for these transforms would be for them to exist in the `app/transforms` directory as overridable defaults; however, in order to continue supporting the legacy import locations in the `ember-data` package they continue to live in`@ember-data/serializer/-private` from whence they are re-exported.

Once the deprecation lifecycle is complete for pre-package import locations this code could be moved into the `app/` directory.

### DIFF
--- a/packages/-ember-data/addon/setup-container.js
+++ b/packages/-ember-data/addon/setup-container.js
@@ -1,22 +1,18 @@
 import { DebugAdapter } from './-private';
 import Store from '@ember-data/store';
-import JSONAPISerializer from '@ember-data/serializer/json-api';
-import JSONSerializer from '@ember-data/serializer/json';
-import RESTSerializer from '@ember-data/serializer/rest';
 import JSONAPIAdapter from '@ember-data/adapter/json-api';
 
-import { BooleanTransform, DateTransform, NumberTransform, StringTransform } from '@ember-data/serializer/-private';
-
-function has(applicationOrRegistry, fullName) {
-  if (applicationOrRegistry.has) {
-    // < 2.1.0
-    return applicationOrRegistry.has(fullName);
-  } else {
-    // 2.1.0+
-    return applicationOrRegistry.hasRegistration(fullName);
+function hasRegistration(application, registrationName) {
+  // fallback our ember-data tests necessary
+  // until we kill-off createStore/setupStore
+  // or @ember/test-helpers kills off it's
+  // legacy support that calls our initializer with registry
+  // instead of application
+  if (!application.hasRegistration) {
+    return application.has(registrationName);
   }
+  return application.hasRegistration(registrationName);
 }
-
 /*
  Configures a registry for use with an Ember-Data
  store. Accepts an optional namespace argument.
@@ -24,18 +20,15 @@ function has(applicationOrRegistry, fullName) {
  @method initializeStore
  @param {Ember.Registry} registry
  */
-function initializeStore(registry) {
-  let registerOptionsForType = registry.registerOptionsForType || registry.optionsForType;
-  registerOptionsForType.call(registry, 'serializer', { singleton: false });
-  registerOptionsForType.call(registry, 'adapter', { singleton: false });
-  registry.register('serializer:-default', JSONSerializer);
-  registry.register('serializer:-rest', RESTSerializer);
+function initializeStore(application) {
+  let registerOptionsForType = application.registerOptionsForType || application.optionsForType;
+  registerOptionsForType.call(application, 'serializer', { singleton: false });
+  registerOptionsForType.call(application, 'adapter', { singleton: false });
 
-  registry.register('adapter:-json-api', JSONAPIAdapter);
-  registry.register('serializer:-json-api', JSONAPISerializer);
+  application.register('adapter:-json-api', JSONAPIAdapter);
 
-  if (!has(registry, 'service:store')) {
-    registry.register('service:store', Store);
+  if (!hasRegistration(application, 'service:store')) {
+    application.register('service:store', Store);
   }
 }
 
@@ -66,23 +59,8 @@ function initializeStoreInjections(registry) {
   inject.call(registry, 'data-adapter', 'store', 'service:store');
 }
 
-/*
- Configures a registry for use with Ember-Data
- transforms.
-
- @method initializeTransforms
- @param {Ember.Registry} registry
- */
-function initializeTransforms(registry) {
-  registry.register('transform:boolean', BooleanTransform);
-  registry.register('transform:date', DateTransform);
-  registry.register('transform:number', NumberTransform);
-  registry.register('transform:string', StringTransform);
-}
-
 export default function setupContainer(application) {
   initializeDataAdapter(application);
-  initializeTransforms(application);
   initializeStoreInjections(application);
   initializeStore(application);
 }
--- a/packages/-ember-data/addon/setup-container.js
+++ b/packages/-ember-data/addon/setup-container.js
@@ -22,6 +22,8 @@ function hasRegistration(application, registrationName) {
  @param {Ember.Registry} registry
  */
 function initializeStore(application) {
+  // we can just use registerOptionsForType when setupStore is killed
+  // see https://github.com/emberjs/data/issues/6357
   let registerOptionsForType = application.registerOptionsForType || application.optionsForType;
   registerOptionsForType.call(application, 'serializer', { singleton: false });
   registerOptionsForType.call(application, 'adapter', { singleton: false });
--- a/packages/-ember-data/addon/setup-container.js
+++ b/packages/-ember-data/addon/setup-container.js
@@ -4,7 +4,8 @@ import JSONAPIAdapter from '@ember-data/adapter/json-api';
 
 function hasRegistration(application, registrationName) {
   // fallback our ember-data tests necessary
-  // until we kill-off createStore/setupStore
+  // until we kill-off setupStore
+  // see https://github.com/emberjs/data/issues/6357
   // or @ember/test-helpers kills off it's
   // legacy support that calls our initializer with registry
   // instead of application
--- a/packages/-ember-data/addon/setup-container.js
+++ b/packages/-ember-data/addon/setup-container.js
@@ -8,7 +8,7 @@ function hasRegistration(application, registrationName) {
   // or @ember/test-helpers kills off it's
   // legacy support that calls our initializer with registry
   // instead of application
-  if (!application.hasRegistration) {
+  if (typeof application.hasRegistration !== 'function') {
     return application.has(registrationName);
   }
   return application.hasRegistration(registrationName);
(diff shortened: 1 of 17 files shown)


packages/-ember-data/tests/integration/store/serializer-for-test.js
```diff
@@ -38,14 +38,20 @@ module('integration/store - serializerFor', function(hooks) {
     store = owner.lookup('service:store');
   });
 
-  test('when no serializer is available we throw an error', async function(assert) {
+  testInProduction('when no serializer is available we throw an error', async function(assert) {
```
runspired: don't need this anymore given chagnes

packages/-ember-data/tests/unit/store/adapter-interop-test.js
```diff
@@ -40,9 +42,14 @@ module('unit/store/adapter-interop - DS.Store working with a DS.Adapter', functi
   testInDebug('Adapter can not be set as an instance', function(assert) {
     assert.expect(1);
 
-    store = DS.Store.create({
+    const BadStore = DS.Store.extend({
```
runspired: lets keep the left side and see if it passes.

packages/-ember-data/tests/helpers/test-in-production.js
```diff
@@ -0,0 +1,10 @@
+import { DEBUG } from '@glimmer/env';
+import { test, skip } from 'qunit';
+
+export default function testInProduction() {
```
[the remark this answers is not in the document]
runspired: in fact, we wouldn't :)

packages/-ember-data/tests/unit/store/adapter-interop-test.js
runspired: @igorT I confirmed that using `Store.create` is not currently supported. The reason this test changed is we now do `getOwner(this)` in the constructor, but we also use getOwner in adapterFor and serializerFor.

rwjblue:
FYI - conflict in `packages/-ember-data/addon/setup-container.js` (from a recent merge)

packages/-ember-data/addon/setup-container.js
```diff
@@ -1,43 +1,36 @@
...
-function has(applicationOrRegistry, fullName) {
-  if (applicationOrRegistry.has) {
-    // < 2.1.0
-    return applicationOrRegistry.has(fullName);
-  } else {
-    // 2.1.0+
-    return applicationOrRegistry.hasRegistration(fullName);
+function hasRegistration(application, registrationName) {
+  // fallback our ember-data tests necessary
+  // until we kill-off createStore/setupStore
```
rwjblue: Not a big deal, but might be good to cross link to an issue here (so it's clearer when we can remove the work around)?

packages/-ember-data/addon/setup-container.js
```diff
@@ -1,43 +1,36 @@
...
+  // or @ember/test-helpers kills off it's
+  // legacy support that calls our initializer with registry
+  // instead of application
+  if (!application.hasRegistration) {
```
rwjblue: No need to do boolean coercion here...

```suggestion
  if (typeof application.hasRegistration !== 'function') {
```

packages/-ember-data/addon/setup-container.js
```diff
@@ -1,41 +1,34 @@
...
  @param {Ember.Registry} registry
  */
-function initializeStore(registry) {
-  let registerOptionsForType = registry.registerOptionsForType || registry.optionsForType;
-  registerOptionsForType.call(registry, 'serializer', { singleton: false });
-  registerOptionsForType.call(registry, 'adapter', { singleton: false });
-  registry.register('serializer:-default', JSONSerializer);
-  registry.register('serializer:-rest', RESTSerializer);
+function initializeStore(application) {
+  let registerOptionsForType = application.registerOptionsForType || application.optionsForType;
```
rwjblue: I think the `application.optionsForType` fallback is also only supporting the `createStore` infra, and we can remove this indirection when we drop it.

packages/-ember-data/addon/setup-container.js
```diff
@@ -1,41 +1,34 @@
...
+  registerOptionsForType.call(application, 'serializer', { singleton: false });
+  registerOptionsForType.call(application, 'adapter', { singleton: false });
 
-  registry.register('adapter:-json-api', JSONAPIAdapter);
-  registry.register('serializer:-json-api', JSONAPISerializer);
+  application.register('adapter:-json-api', JSONAPIAdapter);
```
rwjblue: Why do we need this one? Seems like it should be the same thing as the others...

packages/-ember-data/addon/setup-container.js
```diff
@@ -1,41 +1,34 @@
...
 
-  if (!has(registry, 'service:store')) {
-    registry.register('service:store', Store);
+  if (!hasRegistration(application, 'service:store')) {
+    application.register('service:store', Store);
```
rwjblue: Can we drop this one as well? I think having the reexport actually does what we need here too (custom registrations win, userland app/services/store.js wins, etc)...

packages/-ember-data/tests/integration/store/serializer-for-test.js
```diff
@@ -42,10 +42,16 @@ module('integration/store - serializerFor', function(hooks) {
...
     /*
       serializer:-default is the "last chance" fallback and is
-      registered automatically as the json-api serializer.
-      unregistering it will cause serializerFor to return `undefined`.
+      the json-api serializer which is re-exported as app/serializers/-default.
+      here we override to ensure serializerFor will return `undefined`.
      */
-    owner.unregister('serializer:-default');
+    const lookup = owner.lookup;
+    owner.lookup = registrationName => {
```
rwjblue: This test seems fundamentally impossible and invalid. Given how we are reexporting (or even before with the registration) we can **never** have `serializerFor` return undefined and therefore the assertion being tested can _basically_ never actually happen.

packages/store/addon/-private/system/core-store.ts
```diff
@@ -285,6 +286,29 @@ abstract class CoreStore extends Service {
     }
 
     if (DEBUG) {
+      // support for moduleFor style unit tests
```
rwjblue: 👍 to this comment, I might tweak it slightly to make it clearer that it's only supporting `moduleFor` with a _restricted resolver_ (because when using `moduleFor('asdf', { integration: true })` it would find our reexports).

packages/store/addon/-private/system/core-store.ts
```diff
@@ -3267,12 +3309,23 @@ abstract class CoreStore extends Service {
...
+    }
+
     assert(
-      `No serializer was found for '${modelName}' and no 'application', Adapter.defaultSerializer, or '-default' serializer were found as fallbacks.`,
+      `No serializer was found for '${modelName}' and no 'application' serializer was found as a fallback`,
       serializer !== undefined
     );
+    deprecate(`Usage of the '-default' serializer is deprecated.`, false, {
+      id: 'ember-data:default-serializer-fallback',
+      until: '3.12.0',
```
runspired: the `until` here should be `4.0` but more importantly this should probably be in a different commit since it implements https://github.com/emberjs/rfcs/pull/522 cc @pete-the-pete

packages/store/addon/-private/system/core-store.ts
```diff
@@ -285,6 +286,29 @@ abstract class CoreStore extends Service {
...
+        boolean: 'BooleanTransform',
+        number: 'NumberTransform',
+        string: 'StringTransform',
+      };
+
+      Object.keys(Mapping).forEach(attributeType => {
+        const transform = getOwner(this).lookup(`transform:${attributeType}`);
+
+        if (!transform) {
+          // we don't deprecate this because the moduleFor style tests with the closed
```
rwjblue: IMHO, we should probably emit a deprecation in this path still. `moduleFor` itself is deprecated already but who knows how long that will last. We could do it as a private API deprecation for removal in ~ 3.17.0 (I'd hate to have to support this forever...). The work around for folks still wanting `moduleFor` is to add `integration: true` and remove `needs`, so its overall much smaller work than migrating the whole suite.

packages/store/addon/-private/system/core-store.ts
```diff
@@ -3237,6 +3261,15 @@ abstract class CoreStore extends Service {
     let owner = getOwner(this);
 
     serializer = owner.lookup(`serializer:${normalizedModelName}`);
+
+    // getting rid of this case will be handed by the deprecation of moduleFor in ember test helpers
+    // in production this is handled by the re-export
+    if (DEBUG && serializer === undefined && normalizedModelName === '-json-api') {
```
rwjblue: Same basic concern as above 🤔

packages/-ember-data/addon/setup-container.js
runspired: This PR only restructures serializers, a follow up will do the same for adapters

packages/-ember-data/tests/integration/store/serializer-for-test.js
runspired: It will happen in the future once `-default` is deprecated, and it could happen if we move to "install via blueprint" instead of "re-export in app dir" which is on our radar

packages/store/addon/-private/system/core-store.ts
runspired: When we discussed this a while back we felt deprecating here and deprecating in ember-test-helpers would create unnecessary noise. Has your thinking changed?

runspired: discussed oob, will add deprecations

packages/-ember-data/addon/setup-container.js
runspired: cross linked this issue https://github.com/emberjs/data/issues/6357

packages/-ember-data/addon/setup-container.js
runspired: not sure yet, will investigate in a future PR for cleaning up Store things

packages/-ember-data/addon/setup-container.js
runspired: left a note

packages/store/addon/-private/system/core-store.ts
runspired: clarified

packages/store/addon/-private/system/core-store.ts
runspired: deprecated

packages/store/addon/-private/system/core-store.ts
runspired: added

runspired:
Only failure is known flakey test in Travis

PR #6088, opened 2019-05-07
[CHORE] Convert unit/adapters/build-url-mixin/build-url test…

pete-the-pete:
[CHORE] Convert unit/adapters/build-url-mixin/build-url test to new qunit style.

Also removes use of setupStore since this is only testing the adapter itself and shouldn't need all the setup.

### DIFF
--- a/packages/-ember-data/tests/unit/adapters/build-url-mixin/build-url-test.js
+++ b/packages/-ember-data/tests/unit/adapters/build-url-mixin/build-url-test.js
@@ -1,12 +1,12 @@
-import setupStore from 'dummy/tests/helpers/store';
+import { setupTest } from 'ember-qunit';
 import DS from 'ember-data';
 
 import { module, test } from 'qunit';
 
-let adapter, env;
-
-module('unit/adapters/build-url-mixin/build-url - DS.BuildURLMixin#buildURL', {
-  beforeEach() {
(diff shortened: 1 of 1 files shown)


packages/-ember-data/tests/unit/adapters/build-url-mixin/build-url-test.js
```diff
@@ -18,144 +18,143 @@ module('unit/adapters/build-url-mixin/build-url - DS.BuildURLMixin#buildURL', {
...
+  });
+
+  test('buildURL - findBelongsTo requestType delegates to urlForFindBelongsTo', function(assert) {
+    assert.expect(4);
+    let originalMethod = adapter.urlForFindBelongsTo;
+    let snapshotStub = { snapshot: true };
+    adapter.urlForFindBelongsTo = function(id, type, snapshot) {
+      assert.equal(id, 1);
+      assert.equal(type, 'super-user');
+      assert.equal(snapshot, snapshotStub);
```
runspired: @pete-the-pete want to look into whether there exists a lint rule we can add for `qunit` asserts to enforce that the asserts have labels?

eg. not `assert.ok(bool)` but `assert.ok(bool, 'some-label')` and not `assert.equal('a', 'b')` but `assert.equal('a', 'b', 'some-description')`

runspired: Opened an issue to track: https://github.com/emberjs/data/issues/6090

Issue #6009, opened 2019-04-10
Migrate to env variables from command flags


runspired:
Currently our build utilizes command flags for determining when to keep in-progress-features and instrumentation in the build output. We should migrate these to environment variables to avoid the warning (and potential future assertion) that ember-cli throws for these flags.

See the comment here some additional context: https://github.com/emberjs/data/pull/5997#discussion_r274039678

rwjblue:
Thanks for opening the issue for tracking @runspired.

PR #6091, opened 2019-05-08
[CHORE] Use env variables instead of command flags

zinyando:
Works on https://github.com/emberjs/data/issues/6009

Moves away from using command flags when building eg

`yarn build --instrument` to `INSTRUMENT=true yarn build` and `ENABLE_IN_PROGRESS=true yarn build`

### DIFF
--- a/packages/-build-infra/src/cli-flags.js
+++ b/packages/-build-infra/src/cli-flags.js
@@ -9,19 +9,14 @@ function wantsEnabledFeatures() {
 }
 
 function getManuallyEnabledFeatures() {
-  let args = process.argv;
   let enabled = {};
-  let ARG = '--enable-in-progress-flag';
+  let ARGS = process.env.ENABLE_IN_PROGRESS;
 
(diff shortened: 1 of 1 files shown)


packages/-build-infra/src/cli-flags.js
```diff
@@ -1,27 +1,22 @@
...
+  let ARGS = process.env.ENABLE_IN_PROGRESS;
 
-  for (let i = 0; i < args.length; i++) {
-    if (args[i].indexOf(ARG) === 0) {
-      let toEnable = args[i].substr(ARG.length + 1).split(',');
-      toEnable.forEach(function(flag) {
-        enabled[flag] = true;
-      });
-      break;
+  ARGS.split(',').forEach(function(flag) {
```
runspired: since the environment variable may not be present, we need to guard this behind a check that we actually have a value here

runspired:
Thanks! 🎉

PR #6092, opened 2019-05-08
[BUGFIX release] Fix fetch detection in FastBoot.

rwjblue:
The prior detection of when to use fetch vs when to use jQuery was not properly accounting for the `fastboot` case when `ember-fetch` is not present and a `fetch` global is not present. In that scenario, `najax` would be present (provided by `fastboot` itself).

We already had the correct logic in `_ajax`, however the `useFetch` computed was returning `true` which prevented the `najax` supported branch from being used.

### DIFF
--- a/addon/adapters/rest.js
+++ b/addon/adapters/rest.js
@@ -31,7 +31,8 @@ import { warn } from '@ember/debug';
 import { DEBUG } from '@glimmer/env';
 
 const Promise = EmberPromise;
-const jQ = typeof jQuery !== 'undefined' ? jQuery : undefined;
+const hasJQuery = typeof jQuery !== 'undefined';
+const hasNajax = typeof najax !== 'undefined';
 
 /**
   The REST adapter allows your store to communicate with an HTTP server by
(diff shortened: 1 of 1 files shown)


addon/adapters/rest.js
```diff
@@ -303,9 +304,17 @@ const RESTAdapter = Adapter.extend(BuildURLMixin, {
 
   useFetch: computed(function() {
     let ENV = getOwner(this).resolveRegistration('config:environment');
-    let shouldUseFetch = (ENV && ENV._JQUERY_INTEGRATION) === false || jQ === undefined;
-
-    return shouldUseFetch;
+    // TODO: migrate to static build time detection ala https://github.com/embroider-build/embroider/issues/162
```
igorT: Can you please file an issue

rwjblue: https://github.com/emberjs/data/issues/6093

rwjblue:
Tested this linked against the reproduction app by @mansona and it fixes the latest regression. Test out in https://github.com/mansona/ember-data-fastboot-bug/pull/1.

mansona:
I've tested this locally and I'm 👍 that it has fixed this particular bug

Issue #6090, opened 2019-05-08
[FEAT tests] lint assertions to ensure labels/descriptions


runspired:
This might be all we need https://github.com/platinumazure/eslint-plugin-qunit

TL;DR add for qunit asserts to enforce that the asserts have labels.

eg. not assert.ok(bool) but assert.ok(bool, 'some-label') and not assert.equal('a', 'b') but assert.equal('a', 'b', 'some-description')

pete-the-pete:
Looking into it.

runspired:
Completed by #6096

PR #6096, opened 2019-05-08
[FEAT] Adding eslint-plugin-qunit

pete-the-pete:
This addresses #6090 by adding the eslint plugin. Event though the actual lint rule doesn't enforce having a label for assertions (https://github.com/platinumazure/eslint-plugin-qunit/blob/master/docs/rules/assert-args.md) I feel like this could still be worthwhile to have.

### DIFF
--- a/bin/lint-js
+++ b/bin/lint-js
@@ -1,0 +1,39 @@
+#!/usr/bin/env node
+const ESLINT = '.eslintrc.js';
+const fs = require('fs');
+const path = require('path');
+const { shellSync } = require('execa');
+const projectRoot = path.resolve(__dirname, '../');
+const eslintPath = path.join(projectRoot, ESLINT);
+const eslintInfo = require(eslintPath);
+const tmpEslint = path.join(projectRoot, 'tmp', ESLINT);
+
+function unlinkTemp() {
+  if (fs.existsSync(tmpEslint)) {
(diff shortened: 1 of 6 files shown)


igorT:
Looks good to me. Seems like it ran for existing files though?

pete-the-pete:
oops, not quite there...I'll update.

bin/lint-js
```diff
@@ -0,0 +1,39 @@
...
+
+if (LIST) {
+  LIST = LIST.split(/\r?\n/);
+  // filter for javascript
+  LIST = LIST.filter(item => item.match(/^[^.].*\.js$/)).join();
+  console.log("LIST LENGTH", LIST.length)
+  if (LIST) {
+    try {
+      //update eslint config to include qunit
+      eslintInfo.extends.push('plugin:qunit/recommended');
```
hjdivad: What's the reason for adding the config here and not in `.eslintrc.js`?

pete-the-pete:
As it currently stands, running qunit eslint raises a lot of issues. Adding
it to the main .eslintrc.js means the qunit rules will in charles/azure,
and things start failing. I couldn't find a way to have only 1 plugin run
at a time, so manipulating the config at runtime was the workaround i came
up with.

If/When the qunit eslint issues have been dealt with we can remove this and
add it to the main eslintrc.

PR #6098, opened 2019-05-09
Update Azure Pipelines config to match Travis config

HeroicEric:
Updates the Azure Pipelines config to be more similar to the Travis CI config.

| Type               | Master | Beta | Release | LTS  | Tags |
| ---                | ---    | ---  | ---     | ---  | ---  |
| Lint               | yes    | yes  | no      | no   | no   |
| Basic Tests        | yes    | yes  | yes     | yes  | no   |
| Build  Tests       | yes    | yes  | yes*    | yes* | no   |
| Windows Tests      | yes    | yes  | yes*    | yes* | no   |
| Floating Deps      | yes    | yes  | yes     | yes  | no   |
| Ember LTS Branches | yes    | yes  | yes     | yes  | no   |
| Release Channels   | yes    | yes  | no      | no   | no   |
| External Partners  | yes    | yes  | no      | no   | no   |

* Runs but skips tests for in-progress-features

- Runs Basic Tests for Mac, Linux, and Windows
- Runs the rest of the tests using Linux because it is much faster. Mostly because we get to skip installing Chrome and Firefox, which adds ~= 2 minutes for each job, because they're already available from the Linux VM

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -2,337 +2,181 @@
 # Start with a minimal pipeline that you can customize to build and deploy your code.
 # Add steps that build, run tests, deploy, and more:
 # https://aka.ms/yaml
+#
+# | Type               | Master | Beta | Release | LTS  | Tags |
+# | ---                | ---    | ---  | ---     | ---  | ---  |
+# | Lint               | yes    | yes  | no      | no   | no   |
+# | Basic Tests        | yes    | yes  | yes     | yes  | no   |
+# | Build  Tests       | yes    | yes  | yes*    | yes* | no   |
+# | Windows Tests      | yes    | yes  | yes*    | yes* | no   |
+# | Floating Deps      | yes    | yes  | yes     | yes  | no   |
+# | Ember LTS Branches | yes    | yes  | yes     | yes  | no   |
+# | Release Channels   | yes    | yes  | no      | no   | no   |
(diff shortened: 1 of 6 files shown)


azure-pipelines.yml
```diff
@@ -2,11 +2,26 @@
...
+# | External Partners  | yes    | yes  | no      | no   | no   |
+#
+# * Runs but skips tests for in-progress-features
 
 trigger:
   - master
   - beta
   - release
+  - release-*
+  - lts-*
```
runspired: unsure if Azure has the same issue but in Travis for some PRs we had to also have the org name for the repo specified e.g. `emberjs:release-*`

mike-north:
This could be DRYed templates https://docs.microsoft.com/en-us/azure/devops/pipelines/process/templates?view=azure-devops

You also might want to set up build artifacts (particularly for the external partner tests) so that there's an archive you can pull down and easily examine in the event of a failure. 

Example setup for an ember app
https://github.com/mike-north/ember-octane-workshop/blob/master/.azure-pipelines/ci.yml

rwjblue:
Just merged #6316, so I think 4c4be5f can be rebased out.

rwjblue:
FYI - Landed https://github.com/hjdivad/ember-m3/pull/351 which should fix the long running failure for ember-m3 with ember-data@3.12 and higher.

igorT:
Rebased against master

Issue #4963, opened 2017-05-02
TypeError: Cannot read property 'eachAttribute' of null (store.unloadAll bug?)


shamcode:
Hi! I had error for this code:
```js
this.store.findRecord( 'animal', 1 )
  .then( () => this.store.unloadAll( 'animal' ) )
  .then( () => this.store.findRecord( 'animal', 1 ) ) ;
}
```
Error:
```
vendor.js:30401 TypeError: Cannot read property 'eachAttribute' of null
    at new Snapshot (vendor.js:73205)
    at InternalModel.createSnapshot (vendor.js:67158)
    at _find (vendor.js:76062)
    at Class._fetchRecord (vendor.js:74068)
    at _fetchRecord (vendor.js:74129)
    at Class._flushPendingFetchForType (vendor.js:74230)
    at cb (vendor.js:31972)
    at OrderedSet.forEach (vendor.js:31777)
    at MapWithDefault.forEach (vendor.js:31980)
    at Class.flushAllPendingFetches (vendor.js:74110)
```
[ember-data LOC ](https://github.com/emberjs/data/blob/master/addon/-private/system/snapshot.js#L43)

[Repo for reproduce](https://github.com/shamcode/ember-data-each-attributtes-bug) (ember-twiddle down?) 

I can be wrong, but it seems that `this._isDematerializing` in `InternalModel` should be equal to false after call `store.unloadAll` method.

nagits:
I'm having the same problem.
Workaround: unloadRecord -> schedule('destroy') => _isDematerializing: false -> findRecord

shamcode:
@nagits thank you, but is workaround :)

bichotll:
Out of curiosity. What version are you using? Is it the newest 2.13?

nagits:
Yes, ember-cli/ember/ember-data 2.13. 
2.12 hasn't got this bug.

`unloadRecord` was reworked in 2.13 (commit: https://github.com/emberjs/data/commit/2f18405a4785486f3bd7c865ca1096ea44e2c184, location: https://github.com/emberjs/data/blob/2fabe2e86dfa6561d9d0038d4726e5fcc4658320/addon/-private/system/model/internal-model.js#L489). 

A similar bug was fixed in https://github.com/emberjs/data/pull/2870 (https://github.com/emberjs/data/issues/3296#issuecomment-154502116) 2 years ago.

omh:
[...]

PR #6112, opened 2019-05-15
[BUGFIX relationships] fix infinite retry bug for failed relationship fetches

### LINKED ISSUE #5814: Infinite belongsTo fetch after 404
### Summary

Models keep fetching belongsTo relation object even when it returns a 404 (when they have an inverse relation). 

### Description 

Since we've upgraded ember-data from 3.5.2 to 3.6.0, we have some problems with getting belongsTo relation objects when they don't exist. For example, we have an object:
```js
{
  id:   ...,
  type: ...,
  relationships: {
    activity: 
      { data: { id: 58340, type: 'activity' }
  }
}
```
The back-end can't find the activity with id 58340 (or the user is not allowed to see it) and returns a 404. No problem with previous versions, but now it keeps trying to get it again and again, even when it already has received a 404. It only occurs when the relationship has an inverse. When I add `{ inverse: null }` to the belongsTo, it will just try twice (no answer to that not being once), and then it is resolved. 

#### Current behaviour
It tries to get the belongsTo object from the api, but on retrieving a 404 it keeps trying (infinitely) when an inverse is defined. 
#### Expected behaviour
try it once, and on retrieving a 404 it should stop (I expect it wouldn't exist after trying it 500 times either).

Note: I know our backend shouldn't return a relation with an id that doesn't exist anymore or the user may not retrieve, but I guess this infinite loop isn't expected behaviour either. 

### Reproduction
Remove the `{ inverse: null }` on the parent belongsTo relation in [this Twiddle](https://ember-twiddle.com/99cdb735d0451f093418157935955b8c?openFiles=models.person.js%2C).

### Versions
Latest working version: 3.5.2

Caused by this change:
https://github.com/emberjs/data/pull/5671/files#diff-fd4c34d9a34dfde73cd3413128a3c973R1563
(I changed that line back in a local linked ember-data project)

```cli
yarn list v1.12.3
├─ ember-cli@3.6.0
├─ ember-data-filter@1.13.0
├─ ember-data@3.6.0
├─ ember-source-channel-url@1.1.0
└─ ember-source@3.5.1
✨  Done in 1.06s.
```

runspired:
This PR unskips tests skipped for landing RecordData and fixes regressions introduced around relationship fetching in 3.5

cc @code0100fun please confirm this resolves #5814 et al.

### DIFF
--- a/packages/-ember-data/tests/acceptance/relationships/belongs-to-test.js
+++ b/packages/-ember-data/tests/acceptance/relationships/belongs-to-test.js
@@ -1,4 +1,4 @@
-import { module, test, skip } from 'qunit';
+import { module, test } from 'qunit';
 import { setupRenderingTest } from 'ember-qunit';
 import JSONAPIAdapter from '@ember-data/adapter/json-api';
 import Model from '@ember-data/model';
@@ -374,8 +374,8 @@ module('async belongs-to rendering tests', function(hooks) {
       assert.equal(this.element.textContent.trim(), 'Kevin has two children and one parent');
     });
 
-    skip('Rendering an async belongs-to whose fetch fails does not trigger a new request', async function(assert) {
-      assert.expect(15);
+    test('Rendering an async belongs-to whose fetch fails does not trigger a new request', async function(assert) {
+      assert.expect(14);
       let people = makePeopleWithRelationshipData();
       let sedona = store.push({
         data: people.dict['5:has-parent-no-children'],
@@ -412,10 +412,16 @@ module('async belongs-to rendering tests', function(hooks) {
       assert.equal(this.element.textContent.trim(), '', 'we have no parent');
 
       let relationshipState = sedona.belongsTo('parent').belongsToRelationship;
+      let RelationshipPromiseCache = sedona._internalModel._relationshipPromisesCache;
+      let RelationshipProxyCache = sedona._internalModel._relationshipProxyCache;
 
       assert.equal(relationshipState.isAsync, true, 'The relationship is async');
       assert.equal(relationshipState.relationshipIsEmpty, false, 'The relationship is not empty');
-      assert.equal(relationshipState.relationshipIsStale, true, 'The relationship is still stale');
+      assert.equal(
+        relationshipState.hasDematerializedInverse,
+        true,
+        'The relationship inverse is dematerialized'
+      );
       assert.equal(
         relationshipState.allInverseRecordsAreLoaded,
         false,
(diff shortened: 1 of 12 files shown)


code0100fun:
:tada:  Thanks for fixing this! It will be awesome to get back on the latest release :smile: 

This did fix my errors with infinite 404's. However, I noticed a few more errors in my test suite when running with this branch (they do not happen on current master branch). They all center around ember-data relationships and mostly when unloading a new record.

One bug seems to happen when pushing directly into a relationship (ex: `aModel.hasMany('child-model').push(data)` ) and immediately after calling `newChildModel.unloadRecord()`. This causes computeds with cache keys like `'aModel.childModels.[]'` to re-evaluate but there is a `null` item in the relationship collection. This one is tricky and I'm not sure how to reproduce in a test.

Another bug I am still working through is that calling `newChildModel.unloadRecord()` causes the internal model to be in an `isEmpty` state. Afterwards, calling `aModel.childModels` calls `InternalModel#getHasMany` which eventually gets down to  `PrivateStore#_findEmptyInternalModel` and calls `_scheduleFetch` for a model with a null id. This one is relativly straight forward so I think I can put together a failing test for it.

I still have one other test error to investigate but I'm hoping it is just another manifestation of one of the above issues.

runspired:
Test failures are the usual suspects (flakey Travis test, odd linting issue when running M3 with newer Data)

mansona:
Is this going to be backported? I'm just wondering because of the large impact of the bug?

runspired:
@mansona not likely given it wasn’t reported much despite being an issue for 6 releases and its a harder fix to port. It’ll be in 3.12.

jakebixbyavalara:
@runspired we're actually experiencing this issue with v3.1.2 (which we upgraded to for the fix for https://github.com/emberjs/data/issues/4963). We're aiming to upgrade rapidly soon, but have a few blocking factors that we have to tackle first.
Do you have any suggestions on patching to at least short circuit the retry loop?

Leooo:
@runspired any Idea if we can consume Ember data 3.12 with Ember 3.4 ? Trying ED 3.5 upgrade atm and the bug above would be a blocker for us (upgrading to higher Ember versions across our mani app + its engines would take up months). There was another blocker bug in ED 3.4

Thanks

Mifrill:
@Leooo as for "have any ways around", please try this way: https://github.com/Mifrill/ember-data-stop-infinite-retry-for-failed-relationship-fetches

PR #6113, opened 2019-05-15
[FEAT] - add a public Typescript interface for the minimal required serializer interface

pete-the-pete:
@runspired is this even close? Looking at some other TS interface definitions, is seems like there are a few layers of definitions, in this case DS.Store, DS.Model. Should those be defined as well? 

How should this be tested?

### DIFF
--- a/packages/store/addon/-private/ts-interfaces/serializer.ts
+++ b/packages/store/addon/-private/ts-interfaces/serializer.ts
@@ -1,0 +1,8 @@
+import {
+  JsonApiResource
+} from './record-data-json-api';
+export default interface Serializer {
+  store: any;
+  normalizeResponse(store: any, primaryModelClass: any, payload: JsonApiResource, id: string | number, requestType: string): any;
+  serialize(typeClass: any, hash: object): object;
+}


packages/store/addon/-private/ts-interfaces/serializer.ts
```diff
@@ -0,0 +1,9 @@
+import {
+  JsonApiResource
+} from './record-data-json-api';
+export default interface Serializer {
+  store: any;
+  normalize(typeClass: any, hash: any): any;
+  normalizeResponse(store: any, primaryModelClass: any, payload: JsonApiResource, id: string | number, requestType: string): any;
```
hjdivad: Why would the minimal serializer need to have both `normalize` and `normalizeResponse`?

pete-the-pete:
I don't know. I went off the interface of
https://api.emberjs.com/ember-data/3.10/classes/DS.Serializer

packages/store/addon/-private/ts-interfaces/serializer.ts
hjdivad: @runspired @igorT seems like a good opportunity to nix `normalize`?

runspired: `Store.normalize` is our only non-serializer code that calls `Serializer.normalize`, agreed that we should nix it, would involve deprecating `Store.normalize`.

runspired:
We may also have to include `serializeIntoHash` due to their usage in in `JSONAPIAdapter` and `RESTAdapter` for `updateRecord` and `createRecord`. However given these are user configurable hooks it may be viable to simply make a note that using these adapters out-of-the-box with a custom serializer will require a tiny amount of customization as well.

packages/store/addon/-private/ts-interfaces/serializer.ts
pete-the-pete: Does that mean I should remove it from this? or add the deprecation to Store.normalize?

hjdivad: My view is that we should nix this; the serializer interface should have only `normalizeResponse` and `serialize`

@runspired is quite right that we should separately deprecate the other hooks.

We should add tests to verify that you *can* write a userland serializer that implements only those two hooks.

runspired:
Closing in favor of #6451

PR #6124, opened 2019-05-20
chore: remove co usage

### LINKED ISSUE #6110: [CHORE] replace usage of `co` with async/await
When we needed to support older node versions we used the `co` package to manage some async patterns. Now that we only support Node 8+ we should be able to replace these patterns with async/await and remove this dependency.

makepanic:
fixes #6110


rwjblue:
Thank you!

Issue #6119, opened 2019-05-16
[CHORE packages] enhance `packages-for-commit` for shareable tarballs


runspired:
Currently the tarballs produced by `node ./bin/packages-for-commit.js` are absolute paths, making it difficult to check those tarballs into CI or share them with someone for a local test run. We should make them use relative paths such that as long as they are within the same directory as each other they work.

Steps today for testing a given commit for instance

**Generate Tarballs**
```
git clone git@github.com:emberjs/data.git;
cd data;
git checkout <sha>;
yarn install;
node ./bin/packages-for-commit.js
```

Then  in the consuming project's `package.json`

```
"ember-data": "file:${pathToGeneratedEmberDataTarball}"
```

Because the tarballs use absolute paths they effectively only work well on the same machine as the project that generated them. If we make them relative then either project contributors or a PR author needing to test a PR could produce the tarballs easily and then check them in with a project to allow CI to use them.

**Bonus Feature**

A bonus feature is to have our CI publish any builds for PR commits to a registry we own such that someone could download the tarballs for that build for other test runs (probably hosted on azure/heroku using verdaccio)

pete-the-pete:
Small note on this since I missed it, `node ./bin/packages-for-commit.js` needs to be run _before_ running external tests so that the tarballs exist (they aren't created automatically).

runspired:
Right, this ticket is for improving the existing process, the description above is documenting what the process is today with a description of some improvements to make to it.

runspired:
#6128 upgraded our script for being able to generate tarballs for various situations (local file path, remote/hosted file path, custom registry)

However, we should still do more to ensure that folks submitting PRs have a pleasant experience when attempting to test their PR against their application, will update the ticket above with the requirements for that.

runspired:
Resolved by #6128

PR #6127, opened 2019-05-20
[CHORE packages] enhance `packages-for-commit` for shareable tarballs

pete-the-pete:
#6119

### DIFF
--- a/bin/-tarball-info.js
+++ b/bin/-tarball-info.js
@@ -9,7 +9,7 @@
   that can be used to insert the locations of our tarballs
   into any package.json
 
-  Example Package Meta: 
+  Example Package Meta:
 
   Given a project named `data` located in the folder `projects`
    containing a package named `@ember-data/-example`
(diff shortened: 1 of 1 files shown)


bin/-tarball-info.js
```diff
@@ -77,7 +77,7 @@ const path = require('path');
 
 const projectRoot = path.resolve(__dirname, '../');
 // we share this for the build
-const tarballDir = path.join(projectRoot, '../__tarball-cache');
+const tarballDir = '../../__tarball-cache';
```
bmac: Not sure if windows support matters but I think we can also use `path.relative(__dirname, path.join(projectRoot, '../__tarball-cache'));` to get a platform friendly path. 

Unfortunately, published packages wont work on windows since a unix path would likely be generated a build time and encoded into the published `package.json` tarballs, but thats an issue for the future since these tarballs arn't published anywhere yet.

PR #6128, opened 2019-05-21
[CHORE packaging] enable tarballs to be installed elsewhere

runspired:
Resolves the "portability" requirements of #6119 

The tarballs now include the commit `SHA` in their version number, and they are placed together in a directory named via that same `SHA` (this both makes it a bit easier to quickly grab them but also helps us avoid accidentally using stale tarballs).

The script also now helpfully outputs information that can be used for quickly grabbing / making use of these tarballs, and comes with two flags that allow configuration of the generated tarballs for `local|hosted|registry` usage. These flags alter how the produced tarballs refer to the other produced tarballs in their `dependencies`.

### Example output:

**Building for Local Usage**

`node ./bin/packages-for-commit.js`

Screen Shot 2019-05-21 at 5 07 49 PM

**Building for Hosted Tarballs**

`node ./bin/packages-for-commit.js --hostPath="https://example.com/tarball-cache/ember-data/"`

Screen Shot 2019-05-21 at 5 07 21 PM

**Building for Custom Registry**

` node ./bin/packages-for-commit.js --referenceViaVersion`

Screen Shot 2019-05-21 at 5 04 34 PM

**Users wishing to test a specific version of `ember-data` need all of the produced tarballs**

### DIFF
--- a/bin/-tarball-info.js
+++ b/bin/-tarball-info.js
@@ -77,15 +77,7 @@ const path = require('path');
 const { shellSync } = require('execa');
 const debug = require('debug')('tarball-info');
 const chalk = require('chalk');
-
-function execWithLog(command, force) {
-  debug(chalk.cyan('Executing: ') + chalk.yellow(command));
-  if (debug.enabled || force) {
-    return shellSync(command, { stdio: [0, 1, 2] });
-  }
-
-  return shellSync(command).stdout;
-}
+const cliArgs = require('command-line-args');
 
 const projectRoot = path.resolve(__dirname, '../');
 // we share this for the build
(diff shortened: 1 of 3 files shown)


runspired:
We *may* be able to build upon this PR to move back to using `yarn` to install the dependencies for partner tests that test projects with a yarn lockfile.

Previously, we moved to using `npm` entirely because `yarn pack` produced broken tarballs and `yarn add <tarball>` was broken when those tarballs used file links to refer to other tarballs packed with `npm`.

Since the produced tarballs no longer refer to other tarballs but to a sha-based version it's likely we could use `yarn global add <tarball>`  to populate yarn's cache and thereafter installing the sha-based version might `"Just Work"™`.

runspired:
Opened an issue on `npm` for `npm cache add <tarball>` not making the added version available for install even when in `offline` mode: https://npm.community/t/npm-cache-add-fails-to-make-package-available-for-install/7649

runspired:
Shifted to a strategy of publishing file based tarballs by default, with the option (via cli flags) of prepping them for being hosted on a registry or at a url to improve share-ability. Added some checks and guards for each case, will update the description with new screenshots.

This allow us maximum flexibility so that consuming users can solve the "how do I use this locally" problem the way that best works for them. Going forward @fivetanley and @code0100fun are checking into creating place for us to host temporary artifacts for PRs to make this even more seamless.

PR #6131, opened 2019-05-22
chore: use mocha instead of a custom node test runner

### LINKED ISSUE #6111: [CHORE] eliminate nodetest-runner in favor of default mocha test runner
Currently we provide our own test runner for our mocha tests that test our node and blueprint code. If we no replace mocha-only-detector with an eslint rule and address windows support we should be able to eliminate our custom test runner.

makepanic:
fixes #6111

Open questions:
- [x] - is `argv[2]` ever `all`? I haven't found it in the project
- [x] - if `argv[2]` is `all` then [`'node-tests/**/*-test.js'` is added](https://github.com/emberjs/data/blob/master/packages/-build-infra/src/node-test-helpers/nodetest-runner.js#L41) to the mocha tests. This would create `'node-tests/{blueprints,acceptance,unit}node-tests/**/*-test.js'`  as a mocha glob. This looks like a bug. Can we remove it?
- [x] - the runner is disabled for windows. I'll try to find out if the mocha runner will work on windows
- [x] - previously `node_modules-tmp` was removed. I can't seem to find it used anywhere in the project. Can the removal be removed?


makepanic:
I've also removed run-each, as mentioned on discord.
It's now using `lerna run test:node`.

It'll run the command in all the packages and tries to parallelize tests based on their dependencies.
It seems like it's first executing `@ember-data/-build-infra` tests and then running all the other ones in parallel.



I've also removed the argv handling as it seems like it was previously broken.
Any of the run-each argv weren't passed to execa:

runspired:
`agv[2]` and `node_modules-tmp` are safe to remove. The reason the runner was disabled for windows has more to do with other parts of our infra not having windows support yet. We would very much like to ensure windows support so if that is something you want to look into that would be hugely appreciated <3

makepanic:
I just booted a [modern.ie vm](https://developer.microsoft.com/en-us/microsoft-edge/tools/vms/) and it seems to work fine with `test:node`. (I only had to run it as root because windows :heart: symlinks)

.mocharc.js
```diff
@@ -0,0 +1,13 @@
...
+  const base = 'node-tests/';
+  const dirs = ['blueprints', 'acceptance', 'unit'];
+
+  return dirs.map(dir => `${base}${dir}${suffix}`);
+}
+
+module.exports = {
+  timeout: 5000,
+  reporter: 'spec',
+  spec: testGlob('/**/*-test.js'),
```
makepanic: I think we could make this much simpler by copy pasting the three globs

runspired: e.g. 
```
spec: [
  'node-tests/blueprints/**/*-test.js',
  'node-tests/acceptance/**/*-test.js',
  'node-tests/unit/**/*-test.js'
]
```

makepanic: Yes that's what I was thinking about.
I've updated the PR with it because it's much easier to understand now.

PR #6144, opened 2019-05-31
Use the @ember-data packages in the store API docs

bmac:


runspired:
cc @sivakumar-kailasam

Issue #5100, opened 2017-07-27
Calling serialize on model with embedded relationship causes a warning


josemarluedke:
I have a model which I create a record (`createRecord`) passing only a few fields and not passing the relationship. When I call `serialize()` (or `save()`) it will issue the following warning: 

`WARNING: The embedded relationship 'tags' is undefined for 'post' with id 'null'. Please include it in your original payload.`

The warning feels weird because it is complaining about the original payload, which in this case there is no original payload.

I have a reproduction here: https://github.com/josemarluedke/ember-data-embedded-relationship-warning

This is happening in the latest Ember Data (2.14.x), but was not back in the 2.12.x version.

sly7-7:
I think that's is happening since https://github.com/emberjs/data/pull/5048. The workaround I think about for now is explicitely passing an empty array to the relationship.
Though I wonder if we can reintroduce the old behavior.

cc/ @stefanpenner

amk221:
This is affecting me also.

amk221:
Any news, although they're only warnings, it's putting me off upgrading.

atomkirk:
yeah this seems like a bug since its on create and there is no original payload

amk221:
:(

runspired:
Closing in favor of #5173 which gets at the underlying issue a bit more clearly.

amk221:
Sorry to being up an old issue...

Given a model like:

```
export default Model.extend({
  name: attr('string'),
  bar: belongsTo('bar'),
  bazs: hasMany('baz')
});
```

...and a test endpoint like:

```
this.server.get('/api/foo/1', () => {
  return [
    200,
    {},
    {
      foo: {
        id: 1,
        name: 'Test foo'
      }
    }
  ];
});
```

I often leave out the embedded relationships simply because they are not relevant to the test case and/or there are too many 'blank' relationships to make.

So I ended up with A LOT of:

```
WARNING: The embedded relationship 'bar' is undefined for 'foo' with id '1'. Please include it in your original payload.
```

...do I _really_ have to include it _all the time_ ? 

I expect the answer is yes, but thought I'd ask anyway.

PR #6145, opened 2019-06-03
Disable warning for newly created models with embedded hasMany relations

Exelord:
There is an issue which has been closed without a fix:
- https://github.com/emberjs/data/issues/5100

In short, when creating a model which has embedded hasMany relation in his definition I was getting a warning:
> The embedded relationship 'projects' is undefined for 'account' with id 'null'. Please include it in your original payload.

It's possible to get rid of it by explicitly passing an empty array as relation value to createRecord data, but I guess this shouldn't be necessary as described in the issue.

Here is a fix which turns off this warning for new records.

### DIFF
--- a/packages/serializer/addon/-private/embedded-records-mixin.js
+++ b/packages/serializer/addon/-private/embedded-records-mixin.js
@@ -424,14 +424,16 @@ export default Mixin.create({
     if (serializedKey === relationship.key && this.keyForRelationship) {
       serializedKey = this.keyForRelationship(relationship.key, relationship.kind, 'serialize');
     }
-
-    warn(
-      `The embedded relationship '${serializedKey}' is undefined for '${
-        snapshot.modelName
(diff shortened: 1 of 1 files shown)


packages/serializer/addon/-private/embedded-records-mixin.js
```diff
@@ -424,14 +424,16 @@ export default Mixin.create({
...
-
-    warn(
-      `The embedded relationship '${serializedKey}' is undefined for '${
-        snapshot.modelName
-      }' with id '${snapshot.id}'. Please include it in your original payload.`,
-      typeOf(snapshot.hasMany(relationship.key)) !== 'undefined',
-      { id: 'ds.serializer.embedded-relationship-undefined' }
-    );
+    
+    if (!snapshot._internalModel.currentState.isNew) {
```
runspired: `snapshot.record.isNew` would be better

igorT:
We should check why the value is invalid for newly created records, instead of being an empty array

Issue #6146, opened 2019-06-04
Changing a parent then unload a child record clears the hasMany relationship current state of the parent


lsg-braymon:
### Reproduction
* Pull Request Link: https://github.com/emberjs/data/pull/6147

### Description
1. Model setup:
    ```js
    class Person extends Model {
      @hasMany('pet', { inverse: 'owner', async: false })
      pets;
      @attr()
      name;
    }

    class Pet extends Model {
      @belongsTo('person', { inverse: 'pets', async: false })
      owner;
      @attr()
      name;
    }
    ```
1. Data:
    ```
    personA:
      pets: petA
  
    personB:
      pets: petB
    ```
1. Update the owner of `petB` so that it is now owned by `personA`:
    ```js
    petB.set("owner", personA)

    // It will result in the following relatioships:
    personA.pets // [petA, petB]
    personB.pets // []
    ```
1. Unload the original pet of `personA` which is `petA`: `petA.unloadRecord()`.
1. Inspect the relationships:
    ```js
    petB.owner // personA
    personA.pets // BUG: this is now empty. The expected result is [petB]
    ```

### Versions

Run the following command and paste the output below: `npm ls ember-source && npm ls ember-cli && npm ls ember-data`.

```cli
ember-source@3.10.0
ember-source@3.10.0
ember-data@3.11.0-beta.1
```
We were also using ember-data@3.10.0 before and the issue is also present in that version.

runspired:
This was fixed at some point with all of the relationship layer cleanup :)

PR #6147, opened 2019-06-04
[TEST] - Unload record clears the current state of a has many relationship

lsg-braymon:
This is a test case for issue: https://github.com/emberjs/data/issues/6146

### DIFF
--- a/packages/-ember-data/tests/integration/records/edit-record-test.js
+++ b/packages/-ember-data/tests/integration/records/edit-record-test.js
@@ -3,6 +3,7 @@ import { setupTest } from 'ember-qunit';
 import Store from '@ember-data/store';
 import Model from '@ember-data/model';
 import { attr, hasMany, belongsTo } from '@ember-data/model';
+import todo from '../../helpers/todo';
 
 class Person extends Model {
   @hasMany('pet', { inverse: 'owner', async: false })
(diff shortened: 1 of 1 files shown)


runspired:
Roughly speaking this is an interesting artifact of a very long standing issue that I have a (reasonable) amount of hope that I'll be able to fix in the next few months combined with an expected behavior.

I'll start with a short explanation of the long-standing issue: there's some additional info about that in this long twitter thread: https://twitter.com/Runspired/status/1058536977597775872

Roughly the issue is that how we currently cache and operate on the state of the graph is too lossy, and leads to situations where we are both unable to apply patches correctly and unable to preserve local state when updating information from a remote source is required. This is something we can *largely* fix with a good bit of refactoring, and for the cases we can't eventually we will expose capabilities to users that allow them to signal to use enough information to get it right 100% of the time.

Now, for the expected behavior:

`unloadRecord` moves a record into a "never-loaded" state. It does not purge references to that record from the cache unless there are no retainers (relationships, especially async relationships, are retainers).

When a record is unloaded, for async relationships this causes us to trigger a re-fetch to re-load if the record is still needed and the relationships is accessed.

For sync relationships we do a (bit of an odd) thing by treating it as a signal that we can fully purge the record. We do this not because of correctness (it's not a correct thing to do) but because for a long time buggy behavior in unload effectively allowed users to treat it as such and it became widely depended upon.  When we apply this "canonical delete", we end up doing a re-sync of our remaining "known good remote state" to our current state. This is where the buggy behavior mentioned above causes this issue for sync relationships. For async relationships the buggy behavior would manifest only once the re-fetch completes.

If you convert this test to a `TODO` (there are other examples in the code-base to pull from) I'll merge it as the test looks solid. I don't think we'll be able to address this for a matter of months however.

lsg-braymon:
@runspired The test has been converted to a `TODO`. Thanks!

runspired:
I suspect all the refactoring I did in the relationship layer may have resolved this, I'll make sure to get this test merged next week to check.

runspired:
I didn't have permissions to rebase the original so I've reopened this in a new PR #8160

Issue #6016, opened 2019-04-11
Needed for 3.10 or 3.11 - Verify that Ember Data's API docs work correctly for the monorepo, and fix issues


jenweber:
### Description

The API docs site relies on a particular structure of YUIdoc code comments in order to work, and they have some peculiarities. We need to make sure that when the monorepo lands in a stable release, the API docs are working and no docs have been dropped.

### To-do

- [ ] run `yuidoc . --server` against master, and against 3.9. Diff the resulting `data.json` to see if anything is missing
- [ ]  fix any missing docs (note that all files must include a class declaration, some methods need an `@static`, and the file path must be included in the yuidoc config for something to show up)
- [ ] Once the yuidoc json is in good shape, try running it in the API docs app. See https://github.com/ember-learn/ember-jsonapi-docs and https://github.com/ember-learn/ember-api-docs
- [ ] Coordinate with the Ember Data team to make sure that the fixes to docs get included in the release (might have to be added to an existing beta)

sivakumar-kailasam:
With #6153 we can view the latest docs on the API app. An open issue that we're aware of is the file's location shown for the `defined in` field on the api app. It needs to be handled by the api app.

sivakumar-kailasam:
Also unsure if this is intentional but the api docs' content doesn't reflect the new npm packages. They still seem to be namespaced under `DS`.

runspired:
@sivakumar-kailasam if we merge #6153 and release a beta, are we comfortable with packages becoming available in the 3.11 release?

sivakumar-kailasam:
Without that PR, the docs won't be generated for 3.11 so it has to be merged. 

Our concern was that the docs still don't reflect the new package restructuring that has happened. So it's possible that users maybe confused since they don't see any of the new packages in the docs. 

If that's intentional, then we're probably okay.

sivakumar-kailasam:
We've published 3.11 docs but there are couple of outstanding issues:

1. Links to source code is broken (this needs to be handled by the api app team. I'll be tracking it via https://github.com/ember-learn/ember-api-docs/issues/625)
2. ember data API docs don't reflect the new package structure. This needs to be handled by the data team. If the existing classes are decorated with the modules that they belong to & if the references to `DS.` is replaced, things should show up as expected. `ember-source` repo is an excellent reference for this.

Gaurav0:
@sivakumar-kailasam Is this fixed on the data side with the patch releases?

[...]

PR #6153, opened 2019-06-07
Fix yuidoc setup to support the split packages

sivakumar-kailasam:
When trying to view the latest doc changes on the API docs app, I noticed that the setup is borked.

With these changes, you can view the latest docs from this branch by following the instructions in https://github.com/ember-learn/ember-jsonapi-docs#generating-api-documentation-and-testing-api-docs-locally.

This is for https://github.com/emberjs/data/issues/6016

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -31,7 +31,7 @@ benchmarks/results/*.json
 
 .github-upload-token
 *.swp
-docs/build/
+docs/
 .metadata_never_index
 
 DEBUG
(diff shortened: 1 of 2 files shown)


rwjblue:
I think this will need to be cherry-picked into `beta` as well, @runspired can you confirm?

runspired:
@rwjblue confirm, needs to be cherry-picked

PR #6155, opened 2019-06-10
Improve fetch

wycats:
We're not sure how to test this change. We'd welcome any advice 😄 (@rwjblue?)

We also added some errors for the situation where `fetch` could not be found as a module **or** on the global. We weren't sure where people would prefer to put this error (and whether it would be a problem to throw during module evaluation). We're open to any feedback on this 😄

### DIFF
--- a/packages/adapter/addon/-private/utils/fetch.ts
+++ b/packages/adapter/addon/-private/utils/fetch.ts
@@ -1,17 +1,20 @@
 import require, { has } from 'require';
 
-type MaybeFetch = {
-  (input: RequestInfo, init?: RequestInit | undefined): Promise<Response>;
-} | null;
+type FetchFunction = (input: RequestInfo, init?: RequestInit | undefined) => Promise<Response>;
 
-let _fetch: MaybeFetch = null;
+let _fetch: (() => FetchFunction) | null = null;
 
 if (has('fetch')) {
   // use `fetch` module by default, this is commonly provided by ember-fetch
-  _fetch = require('fetch').default;
+  let foundFetch = require('fetch').default;
+  _fetch = () => foundFetch;
 } else if (typeof fetch === 'function') {
   // fallback to using global fetch
-  _fetch = fetch;
+  _fetch = () => fetch;
+} else {
+  throw new Error(
+    'cannot find the `fetch` module or the `fetch` global. Did you mean to install the `ember-fetch` addon?'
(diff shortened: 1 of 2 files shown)


packages/-ember-data/tests/helpers/todo.js
```diff
@@ -1,95 +0,0 @@
-/* global Proxy */
```
rwjblue: why?

wycats:
Done.

packages/adapter/addon/-private/utils/fetch.ts
```diff
@@ -1,17 +1,20 @@
...
-  _fetch = require('fetch').default;
+  let foundFetch = require('fetch').default;
+  _fetch = () => foundFetch;
 } else if (typeof fetch === 'function') {
   // fallback to using global fetch
-  _fetch = fetch;
+  _fetch = () => fetch;
+} else {
+  throw new Error(
+    'cannot find the `fetch` module or the `fetch` global. Did you mean to install the `ember-fetch` addon?'
```
rwjblue: Nice touch, thank you for adding this!

packages/adapter/addon/rest.js
```diff
@@ -1057,7 +1057,15 @@ const RESTAdapter = Adapter.extend(BuildURLMixin, {
...
 
   _fetchRequest(options) {
-    return fetch(options.url, options);
+    let fetchFunction = fetch();
+
+    if (fetchFunction) {
+      return fetchFunction(options.url, options);
+    } else {
+      throw new Error(
+        'cannot find the `fetch` module or the `fetch` global. Did you mean to install the `ember-fetch` addon?'
```
rwjblue: Do we need the same error in both places? What does that buy us?

wycats: I wasn't sure which of the two we should do. At first glance, I'd prefer it to happen at the top level.

runspired:
I could be misremembering but isn't stubbing ES modules in this way not possible in native and only possible due to an artifact of transpilation? If so, are we sure we want to encourage this pattern vs finding another way to provide the ability to stub?

rwjblue:
@runspired AFAIK, no one is stubbing an ES module. Pretender does `self.fetch = fakeFetch`, which isn't related to stubbing an ES module (its stubbing a global though).

PR #6156, opened 2019-06-10
Switch to todo

wycats:
I'm not sure why we don't have support for `todo` in our test suite, but it would be good to switch to QUnit.todo so we don't have to maintain our own 😄


rwjblue:
testem doesn’t support QUnit.todo, I did a spike PR a long time ago but haven’t had time to try to push it forward

runspired:
@wycats what Rob said. We have our own because it gives us a richer feature set (regression coverage within incomplete TODOs) and works with testem.

runspired:
(It would be nice to upstream the concepts though)

runspired:
Going to close this, once there is an upstream solution we can revisit

PR #6158, opened 2019-06-14
[BUGFIX reload] fix sync belongsTo reload error w/no proxy

runspired:


code0100fun:
:+1: 

This PR branch fixed a test failure for us. This was the last failure keeping us from being on the latest Ember Data :tada: 

For reference this was the error I was seeing:
```
message: >
            Cannot read property 'content' of undefined
        stack: >
            TypeError: Cannot read property 'content' of undefined
                at handleCompletedRelationshipRequest (http://localhost:7357/assets/addon-tree-output/@ember-data/store/-private/system/model/internal-model.js:1347:1)
```

PR #6160, opened 2019-06-17
chore: delete create.js rebase remnant

runspired:
this file was left as a duplicate alongside `create.ts`


igorT:
Do they match upon manual review?

runspired:
@igorT they appear to match, but also I confirmed that the `ts` file would have clobbered the `js` file in the build process

PR #6161, opened 2019-06-17
chore: refactor promiseRecord

runspired:
tiny bit of cleanup to the `store.js` file as we prepare to refactor it more

### DIFF
--- a/packages/store/addon/-private/utils/promise-record.ts
+++ b/packages/store/addon/-private/utils/promise-record.ts
@@ -1,0 +1,26 @@
+import InternalModel from '../system/model/internal-model';
+import { promiseObject } from '../system/promise-proxies';
+import { Record } from '../ts-interfaces/record';
+
+// shim type until we can properly type
+// these proxies
+export type PromiseProxy<T> = Promise<T>;
+
+/**
+ * Get the materialized model from the internalModel/promise
+ * that returns an internal model and return it in a promiseObject.
+ *
+ * Useful for returning from find methods
+ *
+ * @private
+ * @param internalModelPromise
+ * @param label
+ */
+export default function promiseRecord(
+  internalModelPromise: Promise<InternalModel>,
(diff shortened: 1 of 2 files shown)


packages/store/addon/-private/utils/promise-record.ts
```diff
@@ -0,0 +1,18 @@
...
+ * Get the materialized model from the internalModel/promise
+ * that returns an internal model and return it in a promiseObject.
+ *
+ * Useful for returning from find methods
+ *
+ * @private
+ * @param internalModelPromise
+ * @param label
+ */
+export default function promiseRecord(internalModelPromise: Promise<InternalModel>, label: string) {
```
runspired: `InternalModel` is the wrong type here, we need the `Record` type (which I've created elsewhere, will find)

igorT: Can you consult @mike-north on this

runspired: I can//will but it shouldn't block this as I don't add a return type in this PR

packages/store/addon/-private/utils/promise-record.ts
```diff
@@ -0,0 +1,26 @@
...
+// these proxies
+export type PromiseProxy<T> = Promise<T>;
+
+/**
+ * Get the materialized model from the internalModel/promise
+ * that returns an internal model and return it in a promiseObject.
+ *
+ * Useful for returning from find methods
+ *
+ * @private
```
runspired: This should be @internal

packages/store/addon/-private/utils/promise-record.ts
```diff
@@ -0,0 +1,26 @@
...
+ * @param internalModelPromise
+ * @param label
```
runspired: don't need these params unless they have more docs

PR #6162, opened 2019-06-17
chore: eliminate usage of Store in DefaultRecordData

runspired:


packages/store/addon/-private/system/relationships/state/create.ts
```diff
@@ -24,14 +32,14 @@ function createRelationshipFor(relationshipMeta: RelationshipSchema, store: any,
...
 export default class Relationships {
-  recordData: RelationshipRecordData
-  initializedRelationships: {[key: string]: BelongsToRelationship | ManyRelationship}
-  constructor(recordData) {
-    this.recordData = recordData;
+  _store: Store;
+  initializedRelationships: { [key: string]: BelongsToRelationship | ManyRelationship };
+  constructor(public recordData: RelationshipRecordData) {
     this.initializedRelationships = Object.create(null);
+    this._store = recordData.storeWrapper.store;
```
igorT: This seems more suspect tbh then us passing a store to our own record data. Maybe we should remove it from the storeWrapper instead? I'm not quite sure it's supposed to be there

packages/store/addon/-private/system/store/record-data-store-wrapper.ts
```diff
@@ -1,16 +1,27 @@
...
     this.store = store;
     this._willUpdateManyArrays = false;
     this._pendingManyArrayUpdates = [];
   }
 
+  // TODO this exists just to the default recordData can
+  //  check this in debug for relationships
+  //  we can do away with this with a bit of refactoring
+  // of the relationship layer
+  _hasModelFor(modelName: string) {
```
igorT: Seems like this is addressed by the CustomModelClasses rfc

packages/store/addon/-private/ts-interfaces/record-data-store-wrapper.ts
```diff
@@ -1,15 +1,35 @@
-import { RelationshipsSchema, AttributesSchema } from "./record-data-schemas";
+import { RelationshipsSchema, AttributesSchema } from './record-data-schemas';
+import Store from '../system/store';
+
+type Store = InstanceType<typeof Store>;
+
 export interface RecordDataStoreWrapper {
+  store: Store;
```
igorT: if we do this shouldn't this be underscored?

packages/store/addon/-private/system/store/record-data-store-wrapper.ts
runspired: I think we'll leave it for that RFC implementation to make public in the correct way

packages/store/addon/-private/system/relationships/state/create.ts
runspired: discussed oob, this is a step towards eliminating in both locations. moving `store` to `_store` to help reduce the leakage from store-wrapper as well until the relationship layer is refactored to not require store.

packages/store/addon/-private/ts-interfaces/record-data-store-wrapper.ts
runspired: it was already present, this just updated the interface, but after oob discussion moved to underscore

packages/store/addon/-private/system/store/record-data-store-wrapper.ts
```diff
@@ -47,59 +91,124 @@ export default class RecordDataStoreWrapper implements IRecordDataStoreWrapper {
...
+    const modelClass = this._store.modelFor(modelName);
+    const definition = upgradeForInternal(this.relationshipsDefinitionFor(modelName)[key]);
+
+    return definition._inverseIsAsync(this._store, modelClass);
   }
 
-  notifyPropertyChange(modelName, id, clientId, key) {
-    let internalModel = this.store._getInternalModelForId(modelName, id, clientId);
-    internalModel.notifyPropertyChange(key);
+  notifyPropertyChange(modelName: string, id: string | null, clientId: string, key: string): void;
```
igorT: Why do we need multiple signatures here. Fwiw, the new prettier line length thing, is making this much noisier for me to read

runspired: I've fixed up prettier to ensure longer line lengths

packages/store/addon/-private/system/ts-upgrade-map.ts
```diff
@@ -0,0 +1,25 @@
...
+export interface UpgradeMap {
+  RecordDataStoreWrapper: RecordDataStoreWrapper;
+  RelationshipSchema: RelationshipDefinition;
+}
+
+/**
+ * Casts a public interface to the matching internal class implementation
+ *
+ * @internal
+ * @param external
```
mike-north: we don't need `@param` unless you actually describe the argument

packages/store/addon/-private/system/store/record-data-store-wrapper.ts
```diff
@@ -52,66 +81,101 @@ export default class RecordDataStoreWrapper implements IRecordDataStoreWrapper {
...
+    this._store.setRecordId(modelName, id, clientId);
   }
 
-  isRecordInUse(modelName, id, clientId) {
-    let internalModel = this.store._getInternalModelForId(modelName, id, clientId);
-    if (!internalModel) {
-      return false;
+  isRecordInUse(modelName: string, id: string | null, clientId: string): boolean;
+  isRecordInUse(modelName: string, id: string, clientId?: string | null): boolean;
+  isRecordInUse(modelName: string, id: string | null, clientId?: string | null) {
```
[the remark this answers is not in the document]
runspired: refactored to a pattern of

```ts
if (!hasValidId(id, clientId)) {
      throw new Error(MISSING_ID_ARG_ERROR_MESSAGE);
    }
```

which will properly narrow without causing there to be a potential branch with a missing return. Annoying to redeclare this check so many places but such is Typescript :D

PR #6163, opened 2019-06-17
chore: remove dead sinceToken code

runspired:
Since an unknown time many many (many) years ago, `findAll` has not supported use of a `sinceToken`, even though the adapter API still has a slot for it in the arguments passed to `Adapter.findAll`.

This PR removes a bit of infra that attempted to pull an always missing sinceToken from an always empty cache attached to the InternalModel cache where we were never stashing one to begin with.

If this is a feature we wish to support, we should design a mechanism through which consuming applications can retrieve and update the sinceToken.


igorT:
Can you double check the docs please

runspired:
@igorT I added a ticket to track that https://github.com/emberjs/data/issues/6164

Issue #6164, opened 2019-06-17
Update documentation of `sinceToken`


runspired:
Currently the various adapters document the third argument to `findAll` being a `sinceToken`, however this variable is never populated. We should update documentation and examples to reflect this.

efx:
I'll grab this one.

efx:
@runspired I think we can close this issue.

PR #6168, opened 2019-06-18
remove remaining sinceToken and update docs

efx:
See https://github.com/emberjs/data/pull/6163/files
and #6164.

### DIFF
--- a/packages/adapter/addon/adapter.js
+++ b/packages/adapter/addon/adapter.js
@@ -130,11 +130,10 @@ export default EmberObject.extend({
     import $ from 'jquery';
 
     export default DS.Adapter.extend({
-      findAll(store, type, sinceToken) {
-        let query = { since: sinceToken };
+      findAll(store, type) {
 
         return new RSVP.Promise(function(resolve, reject) {
-          $.getJSON(`/${type.modelName}`, query).then(function(data) {
+          $.getJSON(`/${type.modelName}`).then(function(data) {
             resolve(data);
           }, function(jqXHR) {
             reject(jqXHR);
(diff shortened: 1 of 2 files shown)


efx:
@runspired I was a bit confused as a handful of tests fail on master now, but figured I'd open the PR and we can sort out the failures as part of review.

packages/adapter/addon/rest.js
```diff
@@ -516,18 +516,13 @@ const RESTAdapter = Adapter.extend(BuildURLMixin, {
     @method findAll
     @param {DS.Store} store
     @param {DS.Model} type
-    @param {String} sinceToken
     @param {DS.SnapshotRecordArray} snapshotRecordArray
     @return {Promise} promise
   */
-  findAll(store, type, sinceToken, snapshotRecordArray) {
+  findAll(store, type, snapshotRecordArray) {
```
efx: I noticed this also used `sinceToken`, so removed it as well.

packages/adapter/addon/adapter.js
```diff
@@ -147,7 +146,6 @@ export default EmberObject.extend({
     @method findAll
     @param {DS.Store} store
     @param {DS.Model} type
-    @param {String} sinceToken
```
runspired: @param {undefined} neverSet a value is never provided to this argument

packages/adapter/addon/rest.js
```diff
@@ -516,7 +516,6 @@ const RESTAdapter = Adapter.extend(BuildURLMixin, {
     @method findAll
     @param {DS.Store} store
     @param {DS.Model} type
-    @param {String} sinceToken
```
runspired: The argument slot continues to exist as the SnapshotArray is the fourth argument passed in.

```suggestion
    @param {undefined} neverSet a value is never provided to this argument
```

PR #6169, opened 2019-06-18
[BUGFIX headers] fix precedence of Content-Type headers

runspired:


packages/adapter/addon/rest.js
```diff
@@ -1104,7 +1104,11 @@ const RESTAdapter = Adapter.extend(BuildURLMixin, {
     }
 
     if (options.data && options.type !== 'GET') {
-      let contentType = options.contentType || 'application/json; charset=utf-8';
+      let contentType =
+        options.contentType ||
```
runspired: after more consideration, this isn't quite right either, as the `json-api` adapter subclasses this adapter and overrides the default `contentType` by setting `options.contentType`. I think we need a lookup for the default.

Benjy1979:
Hi, is there a workaround with the current release version 3.10?

PR #6180, opened 2019-06-22
chore: cleanup rollup implementation for packages

runspired:
- ensures we run rollup for all `-private` directories
- fixes various dependency declarations that weren't quite right along the way
- does NOT do this as a pre-publish step (follow up PR needed)
- gets rid of all those pesky circular dependency warnings


packages/-build-infra/src/addon-build-config-for-data-package.js
```diff
@@ -1,13 +1,72 @@
...
+
+function addonBuildConfigForDataPackage(PackageName) {
   return {
-    name,
+    name: PackageName,
+
+    init() {
+      this._super.init && this._super.init.apply(this, arguments);
+      this._prodLikeWarning();
+      this.debugTree = BroccoliDebug.buildDebugCallback(PackageName);
```
rwjblue:
```suggestion
      this.debugTree = BroccoliDebug.buildDebugCallback(`ember-data:${PackageName}`);
```

runspired: I did this elsewhere: https://github.com/emberjs/data/pull/6180/commits/498834f71507a4603861abe22bf82b3693e32c63#diff-2658ff1b67b663873e5835e7394a394aR14

and it should be done here too. The packages will have `@ember-data/` in their name but they wouldn't otherwise share a debug namespace and for niceness they should

PR #6181, opened 2019-06-23
Docs update for unloadRecord

ef4:
The way this was phrase implies the exactly wrong thing. `unloadRecord` is *definitely not* about "destroying" anything.

### DIFF
--- a/packages/store/addon/-private/system/model/model.js
+++ b/packages/store/addon/-private/system/model/model.js
@@ -618,7 +618,8 @@ const Model = EmberObject.extend(Evented, {
   },
 
   /**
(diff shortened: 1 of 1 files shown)


runspired:
Just noticed this. `unloadRecord` does destroy the record instance. You are correct that it does not send a `DELETE` request to the server, but it does (literally) call `record.destroy()`. The phrasing here is unfortunate because there are multiple levels of concepts and language semantics involved.

runspired:
We should either revert this or put in a new fix as the new documentation isn't accurate: `unloadRecord` does not unload any "requests" from memory, it clears as much data as it can from the cache and destroys the record instance. Any pending requests will still continue and once they resolve could resurrect the record and repopulate the cache.

PR #6186, opened 2019-06-25
chore: avoid publishing types

### LINKED ISSUE #6052: Prevent packages from publishing typings
Prior to the monorepo conversion @mike-north had concocted  a strategy for preventing us from shipping typings.

This is beneficial in a few ways:

- most of our typings concern private APIs
- our types may eventually become public API but they are too unstable currently to be considered public API
- Typescript doesn't play the best with semver historically and this is a safeguard to help us ensure we remain semver

I forget the details of the strategy we had agreed on, but effectively each package will need to be configured to prevent shipping the types and this is something we must fix prior to shipping the multi-package architecture in a stable release.

runspired:
resolves #6052

### DIFF
--- a/packages/-ember-data/package.json
+++ b/packages/-ember-data/package.json
@@ -99,6 +99,7 @@
   "keywords": [
     "ember-addon"
   ],
+  "types": "./stripped-types.d.ts",
   "ember-addon": {
     "configPath": "tests/dummy/config"
   }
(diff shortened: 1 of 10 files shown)


mike-north:
Also [here](https://www.typescriptlang.org/play/index.html#code/CYUwxgNghgTiAEBbA9sArhBByAZs5WA3AFCiSwIrqby75GnjRxKobZ4ElnOVs10uQA) is my sanity test that multiple empty module declarations with identical names merge neatly

```ts
declare module 'foo';
declare module 'foo';
declare module 'foo';
declare module 'foo';
```

PR #6194, opened 2019-06-26
chore: cleanup rollup implementation for packages (#6180)

runspired:
backports #6180 
backports #6096
backports #6045


runspired:
hrm, appears we may need to backport some other package cleanup PRs (especially around linting). Will investigate

PR #6197, opened 2019-06-26
RecordData errors

igorT:
Implements https://github.com/emberjs/rfcs/pull/465

Also cleans up some of the feature flagging infra thanks to @rwjblue


packages/store/addon/-private/system/model/internal-model.ts
```diff
@@ -1240,20 +1258,48 @@ export default class InternalModel {
...
+        this.send('becameError');
+        this._recordData.commitWasRejected({});
       }
     }
+  }
 
-    this.send('becameInvalid');
+  notifyErrorsChange() {
+    let jsonApiErrors = this._recordData.getErrors() || [];
+    // TODO NOW tighten this check
```
igorT: Needs cleanup

packages/store/addon/-private/system/model/record-data.ts
```diff
@@ -31,7 +32,7 @@ export default class RecordDataDefault implements RelationshipRecordData {
   __data: any;
   _scheduledDestroy: any;
 
-  constructor(modelName: string, id: string | null, clientId: string, storeWrapper: RecordDataStoreWrapper, store:any) {
+  constructor(modelName: string, id: string | null, clientId: string, storeWrapper: RecordDataStoreWrapper, store: any) {
```
mike-north: param properties
```ts
constructor(public modelName: string, etc...) {}
```

igorT:
@mike-north reviewed in person

packages/store/addon/-private/system/model/internal-model.ts
```diff
@@ -14,13 +14,15 @@ import OrderedSet from '../ordered-set';
...
+import { errorsHashToArray, errorsArrayToHash } from '../errors-utils';
 
 import { RecordReference, BelongsToReference, HasManyReference } from '../references';
 import { default as recordDataFor, relationshipStateFor } from '../record-data-for';
 import RecordData from '../../ts-interfaces/record-data';
-import { JsonApiResource } from '../../ts-interfaces/record-data-json-api';
+import { JsonApiResource, JsonApiValidationError } from '../../ts-interfaces/record-data-json-api';
 import { Record } from '../../ts-interfaces/record';
 import { Dict } from '../../types';
+import { parse } from 'url';
```
runspired: seems like an accidental autocomplete import

packages/store/addon/-private/system/model/internal-model.ts
```diff
@@ -245,7 +247,12 @@ export default class InternalModel {
   }
 
   isValid() {
-    return this.currentState.isValid;
+    if (false) {
+      return this.currentState.isValid;
+    } else {
+      // assert here
```
runspired: what is this code branch for? would we assert because we no longer access this state from internalModel?

packages/store/addon/-private/system/model/internal-model.ts
```diff
@@ -895,13 +902,20 @@ export default class InternalModel {
...
-  send(name, context?) {
+  send(name, context?, fromModel?) {
     let currentState = this.currentState;
 
     if (!currentState[name]) {
       this._unhandledEvent(currentState, name, context);
     }
 
+    if (true) {
+      if (fromModel && name === 'becameInvalid' && this._recordData.getErrors().length === 0) {
```
runspired: if this lands before the Evented deprecation (seems likely even though that PR is 99% ready) then we'll need to address this there

cc #6059

packages/store/addon/-private/system/model/internal-model.ts
```diff
@@ -1229,9 +1243,13 @@ export default class InternalModel {
   }
 
   hasErrors() {
-    let errors = get(this.getRecord(), 'errors');
+    if (false) {
+      let errors = get(this.getRecord(), 'errors');
+      return errors.get('length') > 0;
+    } else {
+      return this._recordData.getErrors().length > 0;
```
runspired: why would we assert for `isValid` but not for `hasErrors` ? I am now confused by the abandoned code branch above

packages/store/addon/-private/system/model/internal-model.ts
```diff
@@ -1240,20 +1258,48 @@ export default class InternalModel {
     @method adapterDidInvalidate
     @private
   */
-  adapterDidInvalidate(errors) {
-    let attribute;
+  adapterDidInvalidate(parsedErrors, error) {
+    if (false) {
```
runspired: should be `if (true)` to preserve original behavior for now until FF infra

packages/store/addon/-private/system/model/internal-model.ts
```diff
@@ -245,7 +247,12 @@ export default class InternalModel {
   }
 
   isValid() {
-    return this.currentState.isValid;
+    if (false) {
```
runspired: should be `if (true)`

packages/store/addon/-private/system/model/internal-model.ts
```diff
@@ -1229,9 +1243,13 @@ export default class InternalModel {
   }
 
   hasErrors() {
-    let errors = get(this.getRecord(), 'errors');
+    if (false) {
```
runspired: should be `if (true)`

packages/store/addon/-private/system/model/internal-model.ts
```diff
@@ -1240,20 +1258,48 @@ export default class InternalModel {
...
       }
     }
+  }
 
-    this.send('becameInvalid');
+  notifyErrorsChange() {
+    let jsonApiErrors = this._recordData.getErrors() || [];
+    // TODO NOW tighten this check
+    // don't think we need this
+    let invalidErrors: JsonApiValidationError[] = jsonApiErrors.filter((err) => err.source !== undefined && err.title !== undefined) as JsonApiValidationError[];
```
runspired: why are we forced to cast here? why would we need this filter? ( is this where userland errors enter the system?)

packages/store/addon/-private/system/model/internal-model.ts
```diff
@@ -1240,20 +1258,48 @@ export default class InternalModel {
...
+    this.notifyInvalidErrorsChange(invalidErrors);
+  }
 
-    this._recordData.commitWasRejected();
+  notifyInvalidErrorsChange(jsonApiErrors: JsonApiValidationError[]) {
+    this.getRecord().invalidErrorsChanged(jsonApiErrors);
```
runspired: do we have to go back to the record for this? Is there a behavior we can't replicate by having the record subscribe to the public channel for this?

igorT:
> Outside of these review comments I have the additional concern that we seem to have put a lot of new private methods on `Model`. If this logic truly belongs living with the record can it be encapsulated as util functions within the module instead of leaking methods to consumers?

It can't before Custom Model classes land

packages/store/addon/-private/system/model/internal-model.ts
igorT: c, because we shouldn't be calling this from inside IM

packages/store/addon/-private/system/model/internal-model.ts
igorT: Not quite sure what you mean?

packages/store/addon/-private/system/model/internal-model.ts
igorT: Nah, I'll replace `if (true)` with `if (feature_flag)` before merging. FF infra for this already landed

packages/store/addon/-private/system/model/internal-model.ts
igorT: We can't have the record subscribe before custom record class lands because there is no subscription mechanism

packages/store/addon/-private/system/model/internal-model.ts
igorT: Left a note fore myself above ^^ looking into this right now

igorT:
> module

Added two methods, which after we have a subscription mechanism with custom model classes can be isolated in a callback/not exposed

packages/store/addon/-private/system/model/internal-model.ts
runspired: just leaving a note for that PR because it is currently code complete just needs a name cleanup on the deprecations it introduces

packages/store/addon/-private/system/model/internal-model.ts
runspired: ah, didn't realize those were done separately, makes sense

packages/store/addon/-private/system/model/internal-model.ts
igorT: This was leftover code before adapter errors were handled by request state service. removed

packages/store/addon/-private/system/model/record-data.ts
igorT: fixed

PR #6203, opened 2019-06-27
Failing Test for Accessing Store During Teardown

wagenet:
This is based on a situation in our own app where we were seeing the deprecation warning about accessing a destroyed store. After investigating the stack trace, it became clear that this was not application code accessing the store. Further investigation determined that it was due to interactions between a Proxy (which has observers), view rendering (tracked properties), and the order that records were loaded into the store. This failing test reproduces that scenario.

### DIFF
--- a/packages/-ember-data/tests/acceptance/relationships/has-many-test.js
+++ b/packages/-ember-data/tests/acceptance/relationships/has-many-test.js
@@ -26,6 +26,13 @@ class Person extends Model {
   parent;
 }
 
+class Company extends Model {
+  @attr()
+  name;
+  @hasMany('person', { async: true })
+  employees;
+}
+
 class TestAdapter extends JSONAPIAdapter {
   setupPayloads(assert, arr) {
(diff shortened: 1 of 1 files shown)


runspired:
@wagenet mind rebasing this for me? I suspect we fixed it with array-observer changes but would like to make sure

wagenet:
@runspired done

runspired:
@wagenet thanks! I'm surprised it still fails but I have a suspicion it's because we destroy the store in the test but not the app rendering the records. Is that true to the scenario you hit in your app? Or should we try to update the test to destroy the app rendering the data as well.

wagenet:
@runspired I don't fully remember the context around this, but I did just check and I'm no longer seeing the deprecation warning for this in my app.

runspired:
The test here wasn't destroying the app but only the service while still trying to render objects provided by that service, the failure is expected in that case. Closing since no longer an issue for you, but thanks for the investigative work!

PR #6204, opened 2019-06-28
Record Data State

igorT:

### DIFF
--- a/packages/-ember-data/tests/integration/record-data/record-data-state-test.ts
+++ b/packages/-ember-data/tests/integration/record-data/record-data-state-test.ts
@@ -1,0 +1,286 @@
+import { get } from '@ember/object';
+import { setupTest } from 'ember-qunit';
+import Model from 'ember-data/model';
+import Store from 'ember-data/store';
+import { module, test } from 'qunit';
+import { settled } from '@ember/test-helpers';
+import EmberObject from '@ember/object';
+import { attr, hasMany, belongsTo } from '@ember-data/model';
+import Ember from 'ember';
+import RecordData from '@ember-data/store/-private/ts-interfaces/record-data';
+import { RECORD_DATA_STATE } from '@ember-data/canary-features';
+
+class Person extends Model {
+  // TODO fix the typing for naked attrs
+  @attr('string', {})
+  name;
+
+  @attr('string', {})
+  lastName;
+}
+
+class TestRecordData implements RecordData {
+  commitWasRejected(): void {}
+
+  // Use correct interface once imports have been fix
+  _storeWrapper: any;
+
+  pushData(data, calculateChange?: boolean) {}
+  clientDidCreate() {}
+
+  willCommit() {}
+
+  isRecordInUse() {
+    return false;
+  }
+  unloadRecord() {}
+  rollbackAttributes() {
+    return [];
+  }
+  changedAttributes(): any {}
+
+  hasChangedAttributes(): boolean {
+    return false;
+  }
+
+  setDirtyAttribute(key: string, value: any) {}
+
+  getAttr(key: string): string {
+    return 'test';
+  }
+
+  hasAttr(key: string): boolean {
+    return false;
+  }
+
+  getHasMany(key: string) {
+    return {};
+  }
+
+  isNew() {
+    return false;
+  }
+
+  isDeleted() {
+    return false;
+  }
+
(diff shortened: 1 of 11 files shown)


igorT:
Has a few failing tests I'm addressing

igorT:
All tests should be passing now

packages/store/addon/-private/system/store.js
```diff
@@ -2126,15 +2133,26 @@ const Store = Service.extend({
...
 
-      if (internalModel.currentState.stateName === 'root.deleted.saved') {
-        resolver.resolve();
-        continue;
-      } else if (internalModel.isNew()) {
-        operation = 'createRecord';
-      } else if (internalModel.isDeleted()) {
-        operation = 'deleteRecord';
+      if (!ENABLE_RECORD_DATA_STATE) {
+        if (internalModel.currentState.stateName === 'root.deleted.saved') {
```
runspired: where does the handling of this state move?

packages/store/addon/-private/system/model/internal-model.ts
```diff
@@ -202,15 +204,45 @@ export default class InternalModel {
     // `lastObject` have changed.  When this happens we don't want those
     // models to rematerialize their records.
 
+    // eager checks to avoid instantiating record data if we are empty or loading
+    if (this.isEmpty()) {
+      return true;
+    }
+
+    if (this.isLoading()) {
```
runspired: this seems to be a new addition to our logic?

packages/store/addon/-private/system/model/internal-model.ts
```diff
@@ -202,15 +204,45 @@ export default class InternalModel {
...
+    let isRecordFullyDeleted;
+    if (ENABLE_RECORD_DATA_STATE) {
+      isRecordFullyDeleted = this._isRecordFullyDeleted();
+    } else {
+      isRecordFullyDeleted = this.currentState.stateName === 'root.deleted.saved';
+    }
     return (
-      this._isDematerializing ||
       this.hasScheduledDestroy() ||
+      this._isDematerializing ||
```
runspired: why the ordering switch?

packages/-ember-data/tests/integration/record-data/record-data-state-test.ts
```diff
@@ -0,0 +1,277 @@
...
+
+  isNew() {
+    return false;
+  }
+
+  isDeleted() {
+    return false;
+  }
+
+  addToHasMany(key: string, recordDatas: this[], idx?: number) { }
```
mike-north: replace `this` with `RecordData`

packages/-ember-data/tests/integration/record-data/record-data-state-test.ts
```diff
@@ -0,0 +1,277 @@
...
+    owner.register('service:store', CustomStore);
+  });
+
+  test("Record Data state saving", async function (assert) {
+    assert.expect(3);
+    let isDeleted, isNew, isDeletionCommitted;
+    let calledDelete = false;
+    let calledUpdate = false;
+    let calledCreate = false;
+    let storeWrapper;
```
mike-north: with no initializer, will be an `any` unless you provide a type

packages/store/addon/-private/system/model/internal-model.ts
igorT: I flagged it, but the addition is due to `_isRecordFullyDeleted` accessing recordData, which if you are `isLoading` means that you will eagerly instantiate the RD, even though we know none of the other properties should be true. Great call on flagging though, as it turns out right now all of the tests pass both when returning false and when returning true. I will make sure to add a test on master that asserts the correct behavior

packages/store/addon/-private/system/model/internal-model.ts
igorT: debugging leftover, reverting to original

igorT:
@runspired great catch on model fragments. The test fails because they are using our record data, creating a fragment using a path that makes it think it's been pushed, and then rewiring the state machine to show that the fragment is `isNew`. So it should be pretty straightforward to fix and make their code nicer when this lands

packages/store/addon/-private/system/store.js
igorT: https://github.com/emberjs/data/pull/6204/files#diff-1080751ea355f7ab0c457d1785c50695R2103

Issue #6166, opened 2019-06-18
[QUEST] 🌲Project Trim 🌲


runspired:
# 🌲 Project Trim🌲

As the Ember Community approaches [delivery of long awaited features enabling tree-shaking](https://github.com/embroider-build/embroider), ember-data needs to take some steps to enable this feature to shed the weight of features that consuming applications do not use.

The first step is to ensure that the packages `ember-data` provides are properly encapsulated. In many cases, this will both enable support for tree-shaking once available to the broader community and enable manually opting for a subset of packages.

Later steps will ensure that `ember-data` relies on explicit imports (whether automatically generated at build-time or declared by consuming applications) where currently we rely on resolver lookups.

---

## 🌲 Use `ember-data` without `@ember-data/serializer`

The serializer package provides reference serializer implementations and a base serializer class that are not required for applications that choose to write a custom serializer implementing the minimal serializer interface. This includes the `EmbeddedRecordsMixin` and the base `Transform` class.

Transforms, while specified via `attr` available from `@ember-data/model` are a concept specific to these reference serializers and not used anywhere else within `ember-data`. Custom serializers thus may or may not choose to use them via the same mechanism (`owner.lookup('transform:<name>')`). Lastly the `Transform` base class merely shows the interface required for `transforms` but is not required to be used as a base class.

**Action Items**

- [x]  🔒 ~#6113~ #6451  add a public Typescript interface for the minimal required serializer interface
- [x]  🔒 #6086 eliminate registration of default transforms (use app re-exports instead)
- [x]  🔒 #6086 eliminate injection of `-default` and `-json-api` serializer (use app re-exports instead)
- [x]  🔒https://github.com/emberjs/rfcs/pull/522 deprecate `-default` serializer fallback in `Store.serializerFor`
- [x]  🔒https://github.com/emberjs/rfcs/pull/522 deprecate `Adapter.serializer` and `Adapter.defaultSerializer` fallbacks (which currently default to `-json-api`.
- [x]  🔒https://github.com/emberjs/rfcs/pull/522 deprecate `Model.toJSON` and support the scenario for quick debug-mode inspection via `debugData()` method which in `DEBUG` would print a similar object to the current `toJSON` method.
- [x] implement serializer deprecations from https://github.com/emberjs/rfcs/pull/522
- [x] 🔒 #6368 remove `-rest` serializer fallback usage in tests
- [x] 🔒 #6369 #6370 remove `-json-api` serializer fallback usage in tests
[...]

Issue #4695, opened 2016-12-05
Ember Data Deprecation Guide


HeroicEric:
# We need your help 😄

The Ember guides have [a great section with details about deprecated features](http://emberjs.com/deprecations/) for Ember itself and we ~~should definitely~~ now have [the same for Ember Data](http://emberjs.com/deprecations/ember-data/v2.x/).

As you can see below, there is quite a bit of work to do in order to make sure we cover all of the deprecations that have occurred throughout all of the Ember Data 2.x releases.

## What We Need From You

1. Pick a deprecation from the list below.
2. Post a comment in here letting everyone know what you're planning to work on
   so we don't accidentally duplicate effort.
3. Make a PR to the emberjs/website repo adding your deprecation to [the Ember
   Data Deprecation Guide](https://github.com/emberjs/website/blob/master/source/deprecations/ember-data/v2.x.html.md).
4. Make sure your PR links back to this issue so we can track it.

## Deprecation Format

```markdown
### Deprecations Added in Ember Data <VERSION>

#### <TITLE>

###### until: <DEPRECATED UNTIL VERSION>
###### id: <DEPRECATION ID>
###### feature: <feature flag (if applicable)>

<DETAILS>
```

## Tips

- Try to provide an explanation for why the change was made. If you're not sure, try asking in #ember-data on the [Ember Community Slack](https://ember-community-slackin.herokuapp.com/).
- Provide links to the docs where relevant.
- Provide a before and after example when possible.

### Deprecations added in 1.13

- [x] `DS.Store#filter` should probably be moved out of codebase into the addon
  [link](https://github.com/emberjs/data/blob/bc9a659998267b7a5b0f6d1a0f5f8be673819368/addon/-private/system/store.js#L1740-L1771)

### Deprecations added in 2.3

- [ ] [`ember-application.injected-container`](https://github.com/emberjs/data/blob/bc9a659998267b7a5b0f6d1a0f5f8be673819368/addon/-private/system/model/model.js#L1037-L1039)
- [x] [`ds.defaultValue.complex-object`](https://github.com/emberjs/data/blob/master/addon/attr.js#L13-L17) (:lock: @heroiceric) https://github.com/emberjs/website/pull/2741
[...]

PR #6205, opened 2019-06-28
[WIP] deprecate a few methods

efx:
notify of deprecated methods

Noticed these still annotated with TODO and were not completed as part
of #4695. Fixing this to understand how to deprecate functionality
related to #6166.


runspired:
do the other methods exist?

efx:
@runspired They do not exist! I should've checked before opening this. I'll WIP and see about adding them.

efx:
Closing until we can implement those methods.

PR #6206, opened 2019-06-28
Topic feature flag env var

igorT:

### DIFF
--- a/packages/canary-features/addon/environment/index.js
+++ b/packages/canary-features/addon/environment/index.js
@@ -1,3 +1,1 @@
-// export * from './lib/context';
-// export * from './lib/env';
 export * from './global';
--- a/packages/canary-features/addon/environment/index.js
+++ b/packages/canary-features/addon/environment/index.js
@@ -1,3 +1,3 @@
 // export * from './lib/context';
 // export * from './lib/env';
-export { default as global } from './global';
+export * from './global';
--- a/packages/canary-features/addon/environment/index.js
+++ b/packages/canary-features/addon/environment/index.js
@@ -1,0 +1,3 @@
+// export * from './lib/context';
+// export * from './lib/env';
+export { default as global } from './global';
--- a/packages/canary-features/addon/index.js
+++ b/packages/canary-features/addon/index.js
@@ -1,10 +1,8 @@
 import { assign } from '@ember/polyfills';
 import global from '@ember-data/canary-features/environment';
-// import { ENV } from '@ember-data/canary-features/
 export const ENV = {
   FEATURES: {},
 };
-console.log(global.EMBER_ENV, global.ENV, global.FEATURES);
 (EmberENV => {
   if (typeof EmberENV !== 'object' || EmberENV === null) return;
 
@@ -28,5 +26,5 @@ export const DEFAULT_FEATURES = {
   sample_feature_flag: null,
 };
 
-export const FEATURES = assign(DEFAULT_FEATURES, ENV.FEATURES);
-export const sample_feature_flag = DEFAULT_FEATURES.sample_feature_flag;
+export const FEATURES = assign({}, DEFAULT_FEATURES, ENV.FEATURES);
+export const sample_feature_flag = FEATURES.sample_feature_flag;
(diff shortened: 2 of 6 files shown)


packages/canary-features/addon/index.js
```diff
@@ -1 +1,32 @@
...
+    )
+      continue;
+    let defaultValue = ENV[flag];
+    if (defaultValue === true) {
+      ENV[flag] = EmberENV[flag] !== false;
+    } else if (defaultValue === false) {
+      ENV[flag] = EmberENV[flag] === true;
+    }
+  }
+})(global.EmberENV || global.ENV);
```
rwjblue: We should use EmberDataENV or something (not the same global that Ember uses).

packages/canary-features/addon/environment/index.js
```diff
@@ -0,0 +1,3 @@
+// export * from './lib/context';
```
rwjblue: Probably can kill these commented things

packages/canary-features/addon/index.js
```diff
@@ -1 +1,32 @@
-// export const sample_feature_flag = true;
+import { assign } from '@ember/polyfills';
+import global from '@ember-data/canary-features/environment';
+// import { ENV } from '@ember-data/canary-features/
```
rwjblue: Can cut this

packages/canary-features/addon/index.js
```diff
@@ -1 +1,32 @@
-// export const sample_feature_flag = true;
```
rwjblue: Same here

packages/canary-features/addon/index.js
```diff
@@ -1 +1,32 @@
...
+    if (defaultValue === true) {
+      ENV[flag] = EmberENV[flag] !== false;
+    } else if (defaultValue === false) {
+      ENV[flag] = EmberENV[flag] === true;
+    }
+  }
+})(global.EmberENV || global.ENV);
+
+export const DEFAULT_FEATURES = {
+  sample_feature_flag: null,
```
rwjblue: I _vaguely_ prefer the SCREAMING_SNAKE_CASE here, though I know this isn’t really new (or the point of the PR)

packages/canary-features/addon/index.js
```diff
@@ -1 +1,32 @@
...
+};
+
+export const FEATURES = assign(DEFAULT_FEATURES, ENV.FEATURES);
```
rwjblue: You don’t want to mutate DEFAULT_FEATURES here, probably want `assign({}, DEFAULT_FEATURES, ENV.FEATURES)`

packages/canary-features/addon/index.js
```diff
@@ -1 +1,32 @@
...
+export const sample_feature_flag = DEFAULT_FEATURES.sample_feature_flag;
```
rwjblue: Should be `= FEATURES.sample_feature_flag`

packages/canary-features/addon/index.js
efx: I'll get this and the remaining feedback addressed tomorrow.

packages/canary-features/addon/index.js
efx: Given Ember uses that same convention, I agree to make it the same. I don't see warrant for introducing a new convention.

efx:
Addressed most of the feedback, but this still does not work when testing enabling a flag within an ember data file. I tested with the following in internal model's constructor:
```js
import { SAMPLE_FEATURE_FLAG } from '@ember-data/canary-features';
// ...
 if (SAMPLE_FEATURE_FLAG) {
   console.log('hi')
}                                                                       

I think I'm missing how exactly debug macros correctly serializes the environment / variables from features.js into a module the browser can understand.

igorT:
@efx I tested it and it actually worked for me

rwjblue:
FYI - Igor and I did some more work on the feature flagging setup over in #6197. AFAICT the main thing left is parsing out the default features and values from `@ember-data/canary-features`'s `addon/index.js` instead of hard coding the list in both places, there is a pending TODO in the file for that though...

PR #6208, opened 2019-06-29
lock prettier version

igorT:
We seem to be getting lots of churn with prettier changes, so we should be deliberate about when and how we upgrade. I set the value to what is currently in the lock file
@runspired @rwjblue


rwjblue:
Just rebased this (to get it green), I'd like to land it...

PR #6212, opened 2019-07-01
scaffold mocha unit testing for feature flags

efx:
I setup some mocha tests. I think once we migrate the `features.js` logic into this package, we could fill out those tests. Until then, this just sets up running it like the other packages.

### DIFF
--- a/packages/canary-features/node-tests/unit/feature-merging-test.js
+++ b/packages/canary-features/node-tests/unit/feature-merging-test.js
@@ -1,0 +1,5 @@
+/* eslint-env mocha */
+describe('Unit: canary-features', function() {
+  it('features.js is the single source of truth for flags');
+  it('provides correct default values for feature flags');
+});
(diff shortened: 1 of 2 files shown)


runspired:
Ignore the failures (from elsewhere). Are you thinking add the scaffold and then come back and add more testing? Or do you want to build on this PR?

efx:
Roger. I thought we could add the actual tests once we centralize the features into that addon. I think that was the remaining TODO per [Robert's comment](https://github.com/emberjs/data/pull/6206#issuecomment-506927403). So just adding the scaffolding first in this PR.

efx:
So I'm thinking these probably should live in the debug macros package; I'm going to try addressing the [TODO](https://github.com/emberjs/data/blob/master/packages/canary-features/addon/index.js#L7-L8). After that I'll add to this PR or close it if it is necessary.

PR #6221, opened 2019-07-04
[CHORE] Modernize integration/ tests and remove run loop calls

pete-the-pete:
Updating a few tests at the top level of the integration directory...more to come but wanted to get something up to make sure i'm on the right track.

@runspired

### DIFF
--- a/packages/-ember-data/tests/integration/inverse-test.js
+++ b/packages/-ember-data/tests/integration/inverse-test.js
@@ -1,14 +1,9 @@
-import { run } from '@ember/runloop';
-import setupStore from 'dummy/tests/helpers/store';
-
+import { setupTest } from 'ember-qunit';
 import testInDebug from 'dummy/tests/helpers/test-in-debug';
 import { module, test } from 'qunit';
 
-import DS from 'ember-data';
-
-let env, store, User, Job, ReflexiveModel;
-
-const { attr, belongsTo } = DS;
+import Model from '@ember-data/model';
+import { attr, belongsTo } from '@ember-data/model';
 
 function stringify(string) {
   return function() {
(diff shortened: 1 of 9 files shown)


pete-the-pete:
@runspired the `integration/record_array_manager` failing tests are passing locally...any ideas?

runspired:
I'll pull down and see if I can spot why they seem to pass in one place and not the other, could be a good pairing session. @pete-the-pete

packages/-ember-data/tests/integration/inverse-test.js
```diff
@@ -1,68 +1,77 @@
...
+import Model from '@ember-data/model';
+import { attr, belongsTo } from '@ember-data/model';
 
 function stringify(string) {
   return function() {
     return string;
   };
 }
 
+// @pete_the_pete refer to https://github.com/emberjs/data/commit/e34cb5a37391ce2e5d25401ca189a9bd19a29340
```
runspired: forgotten?

packages/-ember-data/tests/integration/inverse-test.js
```diff
@@ -111,41 +120,41 @@ module('integration/inverse_test - inverseFor', function(hooks) {
   });
 
   test('Returns null if inverse relationship it is manually set with a different relationship key', function(assert) {
-    Job.reopen({
+    job.reopen({
```
runspired: we should avoid reopens of models. Either register a new class or extend the original class. It might be that these classes aren't as "dry" as they seem at first and should just be created and registered uniquely in each test.

packages/-ember-data/tests/integration/inverse-test.js
```diff
@@ -111,41 +120,41 @@ module('integration/inverse_test - inverseFor', function(hooks) {
...
       user: belongsTo('user', { inverse: 'previousJob', async: false }),
     });
 
-    User.reopen({
+    user.reopen({
```
runspired: We should preserve the pattern of capital letters for classes. We should be able to jsut reference the original non-registered class in these locations.

packages/-ember-data/tests/integration/inverse-test.js
pete-the-pete: 💯

packages/-ember-data/package.json
```diff
@@ -73,13 +73,14 @@
     "ember-qunit": "*",
     "ember-qunit-assert-helpers": "*",
     "ember-resolver": "*",
-    "ember-source": "*",
```
runspired: this shouldn't be necessary

packages/-ember-data/package.json
```diff
@@ -73,13 +73,14 @@
...
+    "ember-source": "~3.4.0",
     "ember-source-channel-url": "*",
     "ember-try": "*",
     "github": "*",
     "json-typescript": "*",
     "loader.js": "*",
-    "typescript": "*"
+    "typescript": "*",
+    "@ember/jquery": "^0.6.0"
```
runspired: same with this, if we need to specify it with a `*` that's fine but shouldn't be a specific version. for dev-deps the specific version should live in the root package.json

packages/-ember-data/tests/integration/inverse-test.js
runspired: shows as resolved but still shows up in the latest diff

packages/-ember-data/package.json
pete-the-pete: my bad, this was left over from running tests on a specific version.

packages/-ember-data/package.json
pete-the-pete: same as above.

snewcomer:
@pete-the-pete Any chance we can revisit this PR (perhaps the test run is cleaned up and will pass)?   I'm about to start work on the `Adapter` changes from #6166 and want to make sure we both are complementing each other's work!

pete-the-pete:
@snewcomer i'll rebase and see.

pete-the-pete:
Looks like maybe some failures with partner tests...i can take a look in the new couple days.

rwjblue:
I think you should rebase, @igor and @herioiceric (amongst others) recently got CI passing again (including tweaks / mitigations for partner tests). Rebasing should get you using those changes, so we can see if there are still any failures...

packages/-ember-data/tests/integration/inverse-test.js
runspired: Still needs removed

runspired:
We can ignore the Travis partner test error, it is a known flakey test and more recent CI runs have passed.

packages/-ember-data/tests/integration/inverse-test.js
pete-the-pete: i think i got it this time.

pete-the-pete:
welp....lots more failures 😢 , i'll look.

PR #6228, opened 2019-07-08
ensure single source of truth for canary feature flags

efx:
Extract the definitive list of flags from the `DEFAULT_FEATURES` export in canary-features/addon/index.js.
Thanks to @lifeart for the AST extraction tips.
I'll probably close https://github.com/emberjs/data/pull/6212, because the node logic lives mostly in `-build-infra/src/features`, so adding tests there seems more appropriate.

### DIFF
--- a/packages/-build-infra/package.json
+++ b/packages/-build-infra/package.json
@@ -14,6 +14,7 @@
     "test:node": "mocha"
   },
   "dependencies": {
+    "@babel/parser": "^7.5.0",
     "babel-plugin-debug-macros": "^0.3.2",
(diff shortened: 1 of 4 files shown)


igorT:
@rwjblue you mind also taking a look here?

efx:
So I went with @ef4's suggestion and used esm to source the default features directly. I think this is a better approach. You can see the working changes here: https://github.com/emberjs/data/compare/master...efx:topic-feature-flag-truth-esm?expand=1. Thumbs up and I'll close this PR and open a new one with that branch.

rwjblue:
@efx - Ya, let's go with that approach.

PR #6231, opened 2019-07-09
Use @ember-data/canary-features/default-features module as single source of truth for feature flags

efx:
See discussion in https://github.com/emberjs/data/pull/6228#issuecomment-509637960.

### DIFF
--- a/packages/-build-infra/src/features.js
+++ b/packages/-build-infra/src/features.js
@@ -1,22 +1,61 @@
 'use strict';
 
+function castValue({ type, name, value }) {
+  if (type === 'NullLiteral') {
+    return null;
+  }
+  if (type === 'BooleanLiteral') {
+    return value;
+  }
+  if (type === 'NumericLiteral') {
+    return value;
+  }
+  if (type === 'StringLiteral') {
+    return value;
+  }
+  if (type === 'Identifier' && name === 'undefined') {
+    return undefined;
+  }
+}
 const { parse } = require('@babel/parser');
 const fs = require('fs');
-
 // See ember-source for other implementation to parse typescript file
 function extractFeaturesHash() {
   let fileName = require.resolve('@ember-data/canary-features/addon/index.js');
   let fileContents = fs.readFileSync(fileName, { encoding: 'utf8' }).toString();
-  let parsed = parse(fileContents, { sourceType: 'module' })
-  return parsed.program.body;
+  let parsed = parse(fileContents, {
+    sourceType: 'module',
+    estree: true,
+  });
+  // TODO manual extracting, but should considering using a babel plugin
+  // Passing the plugins configuration option did not seem to work
+  let exportedHashDeclaration = parsed.program.body.find(node => {
+    return (
+      node.type === 'ExportNamedDeclaration' &&
+      node.declaration.type === 'VariableDeclaration' &&
+      node.declaration.declarations.length === 1 &&
+      node.declaration.declarations[0].id &&
+      node.declaration.declarations[0].id.type === 'Identifier' &&
+      node.declaration.declarations[0].id.name === 'DEFAULT_FEATURES'
+    );
+  });
+
+  // grab the fields we need
+  const {
+    declaration: {
+      declarations: [node],
+    },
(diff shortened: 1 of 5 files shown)


packages/-build-infra/src/features.js
```diff
@@ -1,11 +1,10 @@
 'use strict';
 
+const requireEsm = require('esm')(module);
 function getFeatures() {
-  const features = {
-    SAMPLE_FEATURE_FLAG: null,
-    RECORD_DATA_ERRORS: null,
-    RECORD_DATA_STATE: null,
-  };
+  const { DEFAULT_FEATURES: features } = requireEsm(
```
efx: NOTE: It was unclear reading esm's documentation whether this would affect later `require` uses.

efx:
@rwjblue here it is!

packages/canary-features/addon/default-features.js
```diff
@@ -0,0 +1,9 @@
+// -build-infra/src/features consumes this variable to
```
rwjblue: Lets tweak this comment a bit, perhaps:

```js
/*
  This list of features is used both at build time (by `@ember-data/-build-infra`)
  and at runtime (by `@ember-data/canary-features`).

  The valid values are:

  - true - The feature is enabled at all times, and cannot be disabled.
  - false - The feature is disabled at all times, and cannot be enabled.
  - null - The feature is disabled by default, but can be enabled at runtime via `EmberDataENV`.
*/
```

packages/canary-features/addon/default-features.js
```diff
@@ -0,0 +1,9 @@
...
+// populate the default features
+// we export it from its own file to avoid
+// requiring @ember/polyfills for the esm require
+export const DEFAULT_FEATURES = {
```
rwjblue: Lets use a default export here:

```suggestion
export default {
```

packages/canary-features/addon/index.js
```diff
@@ -1,17 +1,10 @@
 /* globals EmberDataENV */
 
 import { assign } from '@ember/polyfills';
+import { DEFAULT_FEATURES } from './default-features';
```
rwjblue: If we migrate to a default export over in `./default-features`:

```suggestion
import DEFAULT_FEATURES from './default-features';
```

packages/-build-infra/src/features.js
rwjblue: This needs tweaking right? Since the module now exports default, I would think you'd need:

```suggestion
  const { default: features } = requireEsm(
```

Right?

packages/-build-infra/src/features.js
```diff
@@ -1,11 +1,10 @@
 'use strict';
 
+const requireEsm = require('esm')(module);
 function getFeatures() {
-  const features = {
-    SAMPLE_FEATURE_FLAG: null,
-    RECORD_DATA_ERRORS: null,
-    RECORD_DATA_STATE: null,
-  };
+  const { default: features } = requireEsm(
```
rwjblue:
```suggestion
  const { default: features } = requireEsm('@ember-data/canary-features/addon/default-features.js')
```

packages/-build-infra/src/features.js
```diff
@@ -1,11 +1,10 @@
...
+    '@ember-data/canary-features/addon/default-features.js'
```
rwjblue:
```suggestion
```

packages/-build-infra/src/features.js
```diff
@@ -1,11 +1,10 @@
...
+  );
```
rwjblue:
```suggestion
```

efx:
@rwjblue apologies on the back and forth; I was rushing :(

rwjblue:
No worries!

rwjblue:
The remaining CI failure is due to https://github.com/emberjs/data/issues/6235 (which we should fix separately).

rwjblue:
@efx - I'd definitely love to have similar changes over in emberjs/ember.js if you have the time. Using `esm` to require the file directly is **much** simpler than what we are doing over there at the moment...

PR #6246, opened 2019-07-11
chore: encapsulate InternalModel usage

runspired:
Depends on #6239


runspired:
The ember-m3 failure is because M3 reaches into the InternalModelMap. We could provide a shim to preserve this but once identifiers work enters the cleanup phase that map will go away entirely anyway.

runspired:
Opened an issue in M3 to track the failure there: https://github.com/hjdivad/ember-m3/issues/301

It's a test-only failure.

packages/store/addon/-private/system/model/internal-model.ts
```diff
@@ -1014,9 +1016,9 @@ export default class InternalModel {
     }
   }
 
-  notifyBelongsToChange(key, record) {
+  notifyBelongsToChange(key: string) {
     if (this.hasRecord) {
-      this._record.notifyBelongsToChange(key, record);
+      this._record.notifyBelongsToChange(key, this._record);
```
igorT: I don't think we need to pass in the second argument here, and in the case that we did, it would probably not be `this._record` but rather the value being set. Though that code path seems to have been removed, so we can just not pass the second arg

packages/store/addon/-private/system/store.ts
```diff
@@ -703,12 +705,12 @@ const Store = Service.extend({
     );
     assert(badIdFormatAssertion, (typeof id === 'string' && id.length > 0) || (typeof id === 'number' && !isNaN(id)));
 
-    let normalizedModelName = normalizeModelName(modelName);
-
-    let internalModel = this._internalModelForId(normalizedModelName, id);
+    const normalizedModelName = normalizeModelName(modelName);
```
igorT: I recall a discussion in which we decided to only use `const` for constants, and not inside methods.

packages/store/addon/-private/system/store/internal-model-factory.ts
```diff
@@ -0,0 +1,228 @@
...
+}
+
+/**
+ * The InternalModelFactory handles the lifecyle of
+ * instantiating, caching, and destroying InternalModel
+ * instances.
+ *
+ * @internal
+ */
+export default class InternalModelFactory {
```
igorT: Fwiw, it seems a bit odd to me that something named a factory would also keep a cache of things.

packages/store/addon/-private/system/store.ts
```diff
@@ -2745,27 +2663,29 @@ const Store = Service.extend({
...
+    return internalModelFactoryFor(this).getByResource(resource);
+  },
+
+  /**
+   * TODO Only needed temporarily for test support
+   *
+   * @internal
+   */
+  _internalModelForId(modelName: string, id: string | null, lid: string | null): InternalModel {
+    return internalModelFactoryFor(this).lookup(modelName, id, lid);
```
igorT: If this is still the old `clientId` concept it might be misleading to call it `lid`

packages/store/addon/-private/system/store/internal-model-factory.ts
```diff
@@ -0,0 +1,228 @@
...
+      isNone(existingInternalModel) || existingInternalModel === internalModel
+    );
+
+    this.modelMapFor(internalModel.modelName).set(id, internalModel);
+    this._newlyCreatedModelsFor(internalModel.modelName).remove(internalModel, lid);
+
+    internalModel.setId(id);
+  }
+
+  peekIdOnly(modelName: string, id: string): InternalModel | null {
```
igorT: `peekById` maybe?

packages/store/addon/-private/system/store/internal-model-factory.ts
```diff
@@ -0,0 +1,228 @@
...
+      //   which will call `cancelDestroy` instead for this unload + then
+      //   sync push scenario. Once we have true client-side
+      //   delete signaling, we should never call destroySync
+      internalModel.destroySync();
+      internalModel = null;
+    }
+    return internalModel;
+  }
+
+  build(modelName: string, id: string | null, data?: any, clientId?: string | null) {
```
igorT: should we underscore the private methods like `build`?

packages/store/addon/-private/system/store/internal-model-factory.ts
```diff
@@ -0,0 +1,228 @@
...
+    if (id) {
+      recordMap.remove(internalModel, id);
+    }
+
+    if (clientId) {
+      this._newlyCreatedModelsFor(internalModel.modelName).remove(internalModel, clientId);
+    }
+  }
+
+  modelMapFor(modelName: string): InternalModelMap {
```
igorT: can we now refactor this to use this factory/map interface rather than having two things?

packages/store/addon/-private/system/store/record-data-store-wrapper.ts
```diff
@@ -53,8 +54,11 @@ export default class RecordDataStoreWrapper implements IRecordDataStoreWrapper {
   }
 
   notifyErrorsChange(modelName: string, id: string | null, clientId: string | null) {
-    let internalModel = this._store._getInternalModelForId(modelName, id, clientId);
-    internalModel.notifyErrorsChange();
+    let internalModel = internalModelFactoryFor(this._store).peekId(modelName, id, clientId);
+
+    if (internalModel) {
```
igorT: We should probably assert if no IM here, that doesn't seem like a valid code path

packages/store/addon/-private/system/store/record-data-store-wrapper.ts
```diff
@@ -105,8 +112,11 @@ export default class RecordDataStoreWrapper implements IRecordDataStoreWrapper {
     if (!hasValidId(id, clientId)) {
       throw new Error(MISSING_ID_ARG_ERROR_MESSAGE);
     }
-    let internalModel = this._store._getInternalModelForId(modelName, id, clientId);
-    internalModel.notifyPropertyChange(key);
+    let internalModel = internalModelFactoryFor(this._store).peekId(modelName, id, clientId);
+
+    if (internalModel) {
+      internalModel.notifyPropertyChange(key);
```
igorT: Should we assert or warn here if no IM

packages/store/addon/-private/ts-interfaces/record-data-json-api.ts
```diff
@@ -15,11 +15,21 @@ export interface JsonApiResource {
...
   meta?: any;
 }
-export interface JsonApiResourceIdentity {
-  id?: string | null;
+
+export interface ExistingResourceIdentifierObject {
   type: string;
-  clientId?: string;
+  id: string;
+  clientId?: string | null;
```
igorT: Can this be `clientId?: string` Or does that error?

PR #6247, opened 2019-07-11
[RFC 403] Identifiers Infra

runspired:
Implements the minimum infrastructure for RFC 403 (Identifiers)

~Depends on #6246 #6245 #6244 #6239 #6161~

This implementation does not provide public API access to the feature, but merely ensures that an identifier is created and correctly associated with every `InternalModel`.

A future PR will expose these APIs via feature flag in the locations specified by the RFC.


runspired:
The `ember-m3` failure is because M3 reaches into the InternalModelMap. We could provide a shim to preserve this but once identifiers work enters the cleanup phase that map will go away entirely anyway.

packages/store/addon/-private/identifiers/cache.ts
```diff
@@ -0,0 +1,426 @@
...
+import coerceId from '../system/coerce-id';
+import uuidv4 from './utils/uuid-v4';
+import normalizeModelName from '../system/normalize-model-name';
+import isStableIdentifier from './is-stable-identifier';
+import isNonEmptyString from '../utils/is-non-empty-string';
+
+type IdentifierMap = Dict<string, StableRecordIdentifier>;
+type TypeMap = Dict<string, KeyOptions>;
+
+interface KeyOptions {
```
igorT: Maybe pull this above line 22 where its used

packages/store/addon/-private/identifiers/cache.ts
```diff
@@ -0,0 +1,426 @@
...
+  configuredResetMethod = method;
+}
+
+function defaultGenerationMethod(data: ResourceIdentifierObject, bucket: string): string {
+  if (isNonEmptyString(data.lid)) {
+    return data.lid;
+  }
+  let { type, id } = data;
+  if (isNonEmptyString(id)) {
+    return `@ember-data:lid-${normalizeModelName(type)}-${id}`;
```
igorT: Does this work out with the edge cases of repeat usage and similar?

packages/store/addon/-private/identifiers/cache.ts
```diff
@@ -0,0 +1,426 @@
...
+  }
+  return uuidv4();
+}
+
+const IdentifierCaches = new WeakMap<object, IdentifierCache>();
```
igorT: Can we scope this a bit more narrow than an object?

packages/store/addon/-private/identifiers/cache.ts
```diff
@@ -0,0 +1,426 @@
...
+  id: IdentifierMap;
+  _allIdentifiers: StableRecordIdentifier[];
+}
+
+let configuredForgetMethod: ForgetMethod;
+let configuredGenerationMethod: GenerationMethod;
+let configuredResetMethod: ResetMethod;
+let configuredUpdateMethod: UpdateMethod;
+
+export function setIdentifierGenerationMethod(method: GenerationMethod): void {
```
igorT: How does this work with multiple apps/fastboot etc?

packages/store/addon/-private/identifiers/cache.ts
```diff
@@ -0,0 +1,426 @@
...
+  private _generate: GenerationMethod;
+  private _update: UpdateMethod;
+  private _forget: ForgetMethod;
+  private _reset: ResetMethod;
+
+  constructor() {
+    // we cache the user configuredGenerationMethod at init because it must
+    // be configured prior and is not allowed to be changed
+    this._generate = configuredGenerationMethod || defaultGenerationMethod;
+    this._update = configuredUpdateMethod || defaultEmptyCallback;
```
igorT: Why do we have to set an empty callback?

packages/store/addon/-private/identifiers/cache.ts
```diff
@@ -0,0 +1,426 @@
...
+    this._forget = configuredForgetMethod || defaultEmptyCallback;
+    this._reset = configuredResetMethod || defaultEmptyCallback;
+  }
+
+  // allows us to peek without generating when needed
+  // useful for the "create" case when we need to see if
+  // we are accidentally overwritting something
+  peekRecordIdentifier(resource: ResourceIdentifierObject, shouldGenerate: true): StableRecordIdentifier;
+  peekRecordIdentifier(resource: ResourceIdentifierObject, shouldGenerate: false): StableRecordIdentifier | undefined;
```
igorT: In other places we use `peek` to mean look up but don't generate. Maybe we should keep that distinction here?

igorT: Nice typing

packages/store/addon/-private/identifiers/cache.ts
```diff
@@ -0,0 +1,426 @@
...
+        if (!(resource.lid in this._cache.lids) || this._cache.lids[resource.lid] !== resource) {
+          throw new Error(`The supplied identifier ${resource} does not belong to this store instance`);
+        }
+      }
+      return resource;
+    }
+
+    // `type` must always be present
+    if (DEBUG) {
+      if (!isNonEmptyString(resource.type)) {
```
igorT: Should we use assert for this instead of throw new Eerror?

packages/store/addon/-private/identifiers/cache.ts
```diff
@@ -0,0 +1,426 @@
...
+    if (identifier === undefined) {
+      // we have definitely not seen this resource before
+      // so we allow the user configured `GenerationMethod` to tell us
+      let newLid = this._generate(resource, 'record');
+
+      // we do this _even_ when `lid` is present because secondary lookups
+      // may need to be populated, but we enforce not giving us something
+      // different than expected
+      if (lid !== null && newLid !== lid) {
+        throw new Error(`You should not change the <lid> of a RecordIdentifier`);
```
igorT: Maybe an assert?

packages/store/addon/-private/identifiers/cache.ts
```diff
@@ -0,0 +1,426 @@
...
+      } else if (lid === null) {
+        // allow configuration to tell us that we have
+        // seen this `lid` before. E.g. a secondary lookup
+        // connects this resource to a previously seen
+        // resource.
+        identifier = keyOptions.lid[newLid];
+      }
+
+      if (shouldGenerate === true) {
```
igorT: maybe `if (shouldGenerate)` ?

packages/store/addon/-private/identifiers/is-stable-identifier.ts
```diff
@@ -0,0 +1,7 @@
+import { StableRecordIdentifier, IS_IDENTIFIER } from '../ts-interfaces/identifier';
+
+export default function isStableIdentifier(
+  identifier: StableRecordIdentifier | any
+): identifier is StableRecordIdentifier {
```
igorT: nice typing

packages/store/addon/-private/identifiers/utils/uuid-v4.ts
```diff
@@ -0,0 +1,67 @@
...
+const CRYPTO =
+  typeof window !== 'undefined' &&
+  window.msCrypto &&
+  typeof window.msCrypto.getRandomValues === 'function'
+    ? window.msCrypto
+    : crypto;
+
+// we might be able to optimize this by requesting more bytes than we need at a time
+function rng() {
+  // WHATWG crypto RNG - http://wiki.whatwg.org/wiki/Crypto
```
igorT: Is the crypto one just as fast? We don't need crypto strength

packages/store/addon/-private/identifiers/utils/uuid-v4.ts
```diff
@@ -0,0 +1,67 @@
...
+/**
+ * Convert array of 16 byte values to UUID string format of the form:
+ * XXXXXXXX-XXXX-XXXX-XXXX-XXXXXXXXXXXX
+ */
+const byteToHex: string[] = [];
+for (let i = 0; i < 256; ++i) {
+  byteToHex[i] = (i + 0x100).toString(16).substr(1);
+}
+
+function bytesToUuid(buf) {
```
igorT: Is there a small library we could use for this

packages/store/addon/-private/system/model/internal-model.ts
```diff
@@ -129,16 +132,15 @@ export default class InternalModel {
...
-  constructor(
-    public modelName: string,
-    public id: string | null,
-    public store: Store,
-    public clientId?: string | null
-  ) {
+  constructor(public store: Store, private identifier: RecordIdentifier) {
+    this.id = identifier.id;
+    this.modelName = identifier.type;
+    this.clientId = identifier.lid;
```
igorT: we might want to feature flag not setting a store

packages/store/addon/-private/system/model/internal-model.ts
```diff
@@ -1332,7 +1336,7 @@ export default class InternalModel {
   hasErrors() {
     if (RECORD_DATA_ERRORS) {
       if (this._recordData.getErrors) {
-        return this._recordData.getErrors({}).length > 0;
+        return this._recordData.getErrors(this.identifier).length > 0;
```
igorT: Probably want to feature flag this

packages/store/addon/-private/system/model/record-data.ts
```diff
@@ -37,15 +40,12 @@ export default class RecordDataDefault implements RelationshipRecordData {
   _isDeleted: boolean;
   _isDeletionCommited: boolean;
 
-  constructor(
-    public modelName: string,
-    public id: string | null,
-    public clientId: string,
-    public storeWrapper: RecordDataStoreWrapper
-  ) {
+  constructor(private identifier: RecordIdentifier, public storeWrapper: RecordDataStoreWrapper) {
```
igorT: Need to feature flag this change

packages/store/addon/-private/system/store.ts
```diff
@@ -1270,7 +1271,7 @@ const Store = Service.extend({
     return _findHasMany(adapter, this, internalModel, link, relationship, options);
   },
 
-  _findHasManyByJsonApiResource(resource, parentInternalModel, relationshipMeta, options) {
+  _findHasManyByJsonApiResource(resource, parentInternalModel, relationshipMeta, options): Promise<unknown> {
```
igorT: Sometimes we have a `Promise<any>` and sometimes `Promise<unknown>`, I remember running into this issue as well, also related to RSVP vs native Promise or smth. Any ideas/maybe consult @mike-north

packages/store/addon/-private/identifiers/cache.ts
runspired: it's that or doing a `typeof fn === 'function'` check every time

packages/store/addon/-private/identifiers/utils/uuid-v4.ts
runspired: not without pulling in way more than we need, and the ones I investigated had largely not done the more basic optimizations. I benchmarked this implementation and made it faster but it's still notably slow due to random byte generation needs, hence discussion with @littlecalculist previously around needing this API baked in. There are a few potential ways of optimizing this further, one of which is to allocate more bytes than we need so that if generating multiple in a row later ones are faster, the other way is to increment off of a randomly generated seed per app instance. This latter strategy may have some integrity issues, would need to consider the math on it.

packages/store/addon/-private/system/store.ts
runspired: Mike and I investigated this and discovered the issue is RSVP typings, he PR'd a fix yesterday.

runspired: (We want to use `Promise<unknown>`)

packages/store/addon/-private/identifiers/cache.ts
runspired: yes, should be scoped to Store, when I first wrote `cache` we didn't have the ability to do that yet.

packages/store/addon/-private/identifiers/cache.ts
runspired: it works out for most cases, individuals who need even stronger guarantees of uniqueness or whom already return an `id` unique across types can override the generation to a strategy that suits them better

runspired: (this strategy being the default was discussed in the RFC fwiw)

packages/store/addon/-private/identifiers/cache.ts
runspired: generally speaking the generation method should be constant across instances, set once. The mechanics of this was discussed during the RFC process.

packages/store/addon/-private/identifiers/cache.ts
runspired: throwing vs assert has advantages for

- typescript tooling
- test harness
- debugging

packages/store/addon/-private/identifiers/cache.ts
runspired: see earlier response

packages/store/addon/-private/identifiers/utils/uuid-v4.ts
runspired: if we generate a uuid then we must use the crypto lib as it is the only thing that gives us access to generation of random bytes, which is necessary for generating true uuid's

packages/store/addon/-private/identifiers/cache.ts
runspired: I generally prefer stricter checks (no type casts) with TS both will narrow but one advantage of the stricter runtime check is that folks can't get away with passing in "truthy" and "falsey" things in practice. We've historically had issues with "accepting anything" and then trying to either case or weed out all the bad inputs in EmberData, with a few instances where it turned out we were converting things from objects to null (such as during createRecord) that caused us issues refactoring later.

packages/store/addon/-private/identifiers/cache.ts
runspired: I'll give this some thought

packages/store/addon/-private/identifiers/is-stable-identifier.ts
```diff
@@ -0,0 +1,7 @@
+import { StableRecordIdentifier, IS_IDENTIFIER } from '../ts-interfaces/identifier';
+
+export default function isStableIdentifier(
+  identifier: StableRecordIdentifier | any
```
mike-north: anything ` | any` just becomes an `any`, so we should just type it as such
Screen Shot 2019-07-17 at 1 00 15 PM

packages/store/addon/-private/utils/is-non-empty-string.ts
```diff
@@ -0,0 +1,3 @@
+export default function isNonEmptyString(str?: any): str is string {
```
mike-north: I'm not sure it's appropriate for this to be an optional arg. Invoking a user-defined type guard without passing an argument doesn't make much sense to me

packages/store/addon/-private/system/model/record-data.ts
```diff
@@ -57,7 +76,9 @@ export default class RecordDataDefault implements RelationshipRecordData {
   // PUBLIC API
 
   getResourceIdentifier(): RecordIdentifier {
-    return this.identifier;
+    return IDENTIFIERS
```
igorT: Did we use to have `lid` in the non feature flag case?

packages/store/addon/-private/ts-interfaces/record-data.ts
```diff
@@ -15,7 +15,13 @@ export default interface RecordData {
   pushData(data: JsonApiResource, calculateChange?: boolean): void;
   clientDidCreate(): void;
   willCommit(): void;
+
   commitWasRejected(recordIdentifier?: RecordIdentifier, errors?: JsonApiValidationError[]): void;
+  /**
+   * @deprecated
+   */
+  commitWasRejected(recordIdentifier?: {}, errors?: JsonApiValidationError[]): void;
+  commitWasRejected(recordIdentifier?: RecordIdentifier | {}, errors?: JsonApiValidationError[]): void;
```
igorT: Could this still be empty object?

packages/store/addon/-private/system/model/record-data.ts
runspired: it get's impossibly hard to type the identifiers if we have a branch that doesn't have `lid` or which makes `lid` optional.

packages/store/addon/-private/ts-interfaces/record-data.ts
runspired: is that what this allows?

packages/store/addon/-private/identifiers/cache.ts
igorT: Should we stop using `Ember.assert` completely and switch? cc @mike-north @rwjblue

PR #6252, opened 2019-07-16
[BUGFIX] Includes yuidoc's output data.json to published package

sivakumar-kailasam:
The `.npmignore` file entry that includes the `data.json` file generated by yuidoc doesn't seem to work after the package movement. I used `yarn pack` to verify that this includes the file as a part of the published package.

We need a patch release of 3.11 with this commit.

### DIFF
--- a/packages/-ember-data/.npmignore
+++ b/packages/-ember-data/.npmignore
@@ -34,6 +34,9 @@
 /bin
 /docs
 /node-tests
-!/dist/docs # ignore all assets except /dist/docs folder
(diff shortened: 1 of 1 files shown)


rwjblue:
Thank you for digging into this!

rwjblue:
CI is failing for canary at the moment, will dig into it separately.

sivakumar-kailasam:
@rwjblue @runspired can we get a 3.11 patch release with this commit?

PR #6256, opened 2019-07-18
[BUGFIX release] fix type overrides

runspired:
Updates how we guard against stripping types by:

- not running `ts:precompile`
- ignoring any `.d.ts` files in our `.npmignore`
- ignoring the `/types/` directory in each package where overrides live

See https://typed-ember.github.io/ember-cli-typescript/versions/master/docs/ts-guide/with-addons#publishing

### DIFF
--- a/packages/-build-infra/.npmignore
+++ b/packages/-build-infra/.npmignore
@@ -1,6 +1,7 @@
 # compiled output
 /dist/
 /tmp/
+/types/
 
 # misc
 /.bowerrc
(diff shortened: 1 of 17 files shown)


mike-north:
Given that this is a fix for a regression, can we write a test (maybe just a shell script we can `&&` with the existing tests) to ensure this covers all of our bases?

runspired:
Opened an issue for adding a test: https://github.com/emberjs/data/issues/6261

Likely what we want is something that uses the latest ember-cli, ember-cli-typescript, ember-data to create an app, scaffold a few things (adapters / serializers / models) and ensure it can build without erring

PR #6257, opened 2019-07-18
[BUGFIX docs] ensure we publish docs

runspired:
The previous fix in #6252 didn't work properly with `npm` pack which does not run `prepublishOnly`. Incidentally this means we've not been running our `prepublishOnly` script with previous releases, and `ts:precompile` which it contained does not currently work. #6256 removes this command anyway so I've also gone ahead and removed it here to ensure we build properly.

I added the new `.npmignore` rules and made sure we do a production build for each individual package. At a glance, it appears that the `data.json` file produced in the `-ember-data` package is the only such file and contains the docs from the other packages.

cc @sivakumar-kailasam

### DIFF
--- a/bin/publish.js
+++ b/bin/publish.js
@@ -248,6 +248,25 @@ function packAllPackages() {
     const pkgInfo = require(pkgPath);
     if (pkgInfo.private !== true) {
       // will pack into the project root directory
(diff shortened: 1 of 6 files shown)


igorT:
👍

PR #6268, opened 2019-07-20
[BUGFIX][DOCS]Fix yuidoc generation across packages

sivakumar-kailasam:
Last time I made [a PR like this](#6252) I tested the changes with `yarn pack` without realizing that we were using `npm` for publishes 😅. This time I tested the changes by running,

`node bin/publish.js release --skipPublish --skipSmokeTest --skipVersion --force`


runspired:
The issue was the original wasn't fully backported, not the preceding slash

PR #6272, opened 2019-07-22
[FEAT identifiers] adds flagged public API and integration tests 

### LINKED ISSUE #1726: Extra sideloaded records can cause duplicates in DS.Store
Our server returns extra sideloaded records that aren't entirely related to the request, which should work, but it's causing duplicate records in the store and weird UI behavior. (http://jsfiddle.net/jgillick/agYMQ/)

To simplify it, imagine two models: child and toys. The child hasMany toys. When you add a toy to the child, the XHR responds with not only the new toy record, but also sideloads the child object and list of toys. This way more of the UI can update with one request. The JSON ends up looking like this:

``` javascript
{
    "toy": {
        "id": 1,
        "name": "doll",
        "quantity": 1
    },
    "children": [
        {
            "id": 0,
            "name": "Ben",
            "quantity": 1,
            "toys": [0, 1]
        }
    ],
    "toys": [
        {
            "id": 0,
            "name": "car",
            "quantity": 1
        },
        {
            "id": 1,
            "name": "doll",
            "quantity": 1
        }
    ]
}
```

For some reason this causes the ember store to have 3 toys in it's record array. This shouldn't happen because these items have IDs. So loading records with duplicate IDs should cause it to overwrite the previous one. Instead, when I look at the store's record array, the records are duplicated -- IDs and all.

The other weird effect is when you add the first toy, the list item is created for it in the template, but all values are undefined. After you add the second toy the UI updates as expected. Try the jsfiddle: http://jsfiddle.net/jgillick/agYMQ/
## Why it's happening

When a record is created, it's ID is set to null. When saved, the returned sideloaded records are pushed to the store first and then the target record is handled: [_commit](https://github.com/emberjs/data/blob/master/packages/ember-data/lib/system/store.js#L1675) calls [didSaveRecord](https://github.com/emberjs/data/blob/master/packages/ember-data/lib/system/store.js#L943) which calls [updateId](https://github.com/emberjs/data/blob/master/packages/ember-data/lib/system/store.js#L991) which simply changes the ID on the record from null to what was returned form the server, without checking if exists in the store's record array at that moment.
## Proposed Solution

In updateId it should check if another record of that type and ID exists. If it does, replace that record instead of creating a duplicate.
## Why Support This?

Actively sideloading  extra records can reduce the number of XHR requests by passing important state information in each request. This can lead to a more responsive UI.

For example, adding an product entry to a shopping cart would update the cost values on the cart itself. So you'd create an entry, submit it to the server and then would want to refresh the cart. This can all be done with one request by sideloading the cart to the entry creation response. So you'd have the server returning these properties in the payload: cart, entry, entries (all the entries in the cart). And as shown, it would cause duplicate entries in the store's record array.

### LINKED ISSUE #1829: Ember Data creates duplicate records

### LINKED ISSUE #2407: Add support for STI
Having support to mark relations as having a Single Table Inheritance model would be nice.

We have STI in our app, and see performance hits because we have to fetch the base class (Person) _and_ the implementing class (Client) for ED to understand what's going on.

Any thoughts / ideas on this?

### LINKED ISSUE #4262: pushPayload/save can create duplicate records in the store
I'm working on an app that uses websockets to push out changes from a server to multiple clients, keeping them in sync.  (Specifically the server uses Rails Action Cable, and the client subscribes to channels it wants to receive updates on.)  Data updates are sent down the socket/channel in JSON-API format, thus the clients simply call `this.store.pushPayload(data);` to update themselves.

On the whole this works great, but occasionally we'll see duplicate records in the store of a client that has created a record.  Some investigation shows that this happens when the socket received it's copy of the record before the `save()` call had completed.

Obviously having two copies of the same record (with the same ID) in the store is less than ideal.

The reason for the duplicate is because the original record on which the `save` call was made gets updated with the attributes returned back from the server (`adapterDidCommit` in `ember-data/-private/system/model/internal-model`), filling in the ID of the record object being saved, whilst the `pushPayload` call from the socket has already pushed a new copy with the same ID.

It seems to me that in an ideal world, `adapterDidCommit` could notice that the `id` has been set/updated, and if it can find a duplicate record in the store it could remove that dupe.

A simple workaround for this problem is to slightly delay the call to `pushPayload`, but that will only make the problem less likely to occur.

### LINKED ISSUE #4377: Polymorphic relationships with unknown model types
This issue follows the discussion with @pangratz on Slack regarding the use of a polymorphic `belongsTo` relationship with model types that might not be known to the app.

My use case is roughly like this:

``` js
// models/object.js
  description: DS.attr('string'),
  related: DS.belongsTo('base-object', { polymorphic: true })
});
```

``` js
// models/base-object.js
  description: DS.attr('string')
});
```

``` js
// models/sensor.js
export default BaseObject.extend({
  values: DS.attr()
});
```

``` js
// models/fax.js
export default BaseObject.extend({
  pages: DS.attr('number')
});
```

From the API I'm getting multiple `object` objects, each with a `type` property describing the type of object. Unfortunately the list of types is not fixed and new types can be added by the API authors at any time. Currently a new type will cause an "Unknown model" exception though.

One possible solution to this is checking for the existence of the related model factory in the `object` serializer and setting `related` to `null` if the model can't be found. The only ways to check if a model factory exists right now are a try-catch around `modelFor()` or manually checking the container for `model:...`. It might be a good idea to expose some public method for this use case instead.

Another use case of this is running `peekRecord()` with a type parameter coming from a WebSocket event. In this case I have to handle a missing record via `if` but then also handle the missing model type by wrapping it in a try-catch.

/cc @pangratz

### LINKED ISSUE #4552: Saving with EmbeddedRecordsMixin does not update related models dirty attributes
#2487 is still reproducible on "ember-data": "^2.4.0", "ember-cli": "2.7.0",

and it is big issue for performance while saving ember data.

Copied issue discription from #2487 

No matter what way the serializer is configured in Ember dirty attributes of the relationships records are not updated with those from the response, non dirty attributes are updated though.

The only workaround I have found is to manually go through and mark them individually as willCommit and then didCommit by overriding save on the model - I tried doing this in the serializer for better encapsulation but couldn't get it to work.

If I've specified serialize: 'records' for my serializer when using EmbeddedRecordsMixin I would expect those records to be automatically set as saving/saved etc. along with the parent according to the normal save flow.

### LINKED ISSUE #4972: Saving a new record that returns a existing record now fails.

### LINKED ISSUE #5083: hasMany relationship created the record with id: null
We've two models Contact and phones. In contact model, we got  `phones: hasMany('phone'),` . When we try to create a contact record with multiple phones, the records are created with id:null. Once the response from the server is came back the id is not still resolved on contrary it created duplicate records.
Phone Model: 
`number: attr('string'),
  type: attr('string')`
For Example: In store, contact contains two phones [{id: null, number: +543210, type: 'main'}, {id: null, number: +543211, type: 'other'}]. Once the we got the response from backend, phone records will look like this  
`[{id: null, number: +543210, type: 'main'}, {id: null, number: +543211, type: 'other'},{id: 1, number: +543210, type: 'main'}, {id: 2, number: +543211, type: 'other'}]`
**Issue:** Duplicate records with id and without ids.
**Question:** How ember will map the response correctly with already created record with id:null? Or how to remove the null records to avoid duplication?

### LINKED ISSUE #5237: Support for natural key's

### LINKED ISSUE #6170: findRecord requests do not honor `reload: true` if the returned id does not match the requested id
### Reproduction

Failing test in #6171 
[Failing test](https://github.com/emberjs/data/blob/efd2a8106500b5a8bf1a69346f0fae33a4b36dad/packages/-ember-data/tests/integration/adapter/find-test.js#L115-L161)

### Description

```js
let record = await store.findRecord('x', 1);

// This will not fire a request, despite `reload: true` 
// if the prior request responded with any id besides
// something that coerces to '1'
await store.findRecord('x', 1, { reload: true });
```
### Root Cause

In the first request, we create an `internalModel` for the requested id.  When the request fulfills, we create a new `internalModel` for the id of the JSONAPI resource and abandon the first `internalModel`.

In the second request, we find the original `internalModel` which thinks it's still loading when we check [_findEmptyInternalModel](https://github.com/emberjs/data/blob/8d621d9c3f3b224916c05e946ba2786faf6db8db/packages/store/addon/-private/system/store.js#L782-L793).  However, its `_promiseProxy` result is the new `internalModel`, resulting in the following:

```js
(internalModel._promiseProxy._result === internalModel) === false;
```

This fundamentally happens because `internalModel._promiseProxy` is fulfilled with the return value of `store._push`, which is the `internalModel` of the primary resource in the payload (for a payload with 1 resource).

### Versions

This is an issue in master.

runspired:
resolves #6170 
closes #6171 

Duplicate records on save (mostly already closed in favor of 1829)
resolves #1726 
resolves #1829 
resolves #4262 
resolves #5083 
resolves #5237 
resolves  #4972 
resolves https://github.com/lytics/ember-data-model-fragments/issues/221
resolves emberjs/rfcs#173
resolves #4552

STI / Polymorphism
resolves #2407 
resolves #4377 

Forum
resolves https://discuss.emberjs.com/t/proposal-fix-saving-new-embedded-records-creates-duplicates/3677/17

Once relationships are refactored such that they are managed by identifier as well there's another long list to close :)


packages/store/addon/-private/ts-interfaces/ember-data-json-api.ts
```diff
@@ -31,7 +31,7 @@ export interface ExistingResourceIdentifierObject {
    *
    * [Identifiers RFC](https://github.com/emberjs/rfcs/blob/master/text/0403-ember-data-identifiers.md#ember-data--identifiers)
    */
-  lid?: string;
+  lid?: string | null;
```
runspired: lets not do the | null, and check if clientId is there

packages/store/addon/-private/ts-interfaces/identifier.ts
```diff
@@ -105,3 +100,11 @@ export type ForgetMethod = (identifier: StableIdentifier, bucket: string) => voi
  to the application instance, you may not need to implement the `resetMethod`.
 */
 export type ResetMethod = () => void;
+
+interface Foo {
```
runspired: delete this

packages/store/addon/-private/utils/has-valid-id.ts
```diff
@@ -0,0 +1,7 @@
+export default function hasValidId(id?: string | null, clientId?: string | null): id is string {
```
runspired: should delete the previous version and please add comment here

packages/-ember-data/tsconfig.json
```diff
@@ -24,16 +24,10 @@
...
   },
-  "include": [
-    "app/**/*",
-    "addon/**/*",
-    "tests/**/*",
-    "types/**/*",
-    "test-support/**/*",
-    "addon-test-support/**/*"
-  ],
+  "include": ["app/**/*", "addon/**/*", "tests/**/*", "types/**/*", "test-support/**/*", "addon-test-support/**/*"],
```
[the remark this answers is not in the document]
runspired: it seems our prettier line length does indeed want this

packages/store/addon/-private/identifiers/cache.ts
```diff
@@ -27,25 +27,30 @@ interface KeyOptions {
...
+) => StableRecordIdentifier;
 
-let configuredForgetMethod: ForgetMethod;
-let configuredGenerationMethod: GenerationMethod;
-let configuredResetMethod: ResetMethod;
-let configuredUpdateMethod: UpdateMethod;
+let configuredForgetMethod: ForgetMethod | null;
+let configuredGenerationMethod: GenerationMethod | null;
+let configuredResetMethod: ResetMethod | null;
+let configuredUpdateMethod: UpdateMethod | null;
```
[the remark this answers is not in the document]
runspired: actually it isn't, these are necessary to allow removing the method on teardown.

packages/store/addon/-private/system/model/internal-model.ts
```diff
@@ -166,6 +166,13 @@ export default class InternalModel {
     this._recordReference = null;
   }
 
+  get id() {
+    if (IDENTIFIERS) {
+      return this.identifier.id || this._id;
```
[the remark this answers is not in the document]
runspired: There are tests for setting then peeking during create and post-create pre-save. There was not one for mutating after save that I could see. Have added.

packages/store/addon/-private/identifiers/cache.ts
```diff
@@ -275,38 +298,52 @@ export class IdentifierCache {
...
-      );
 
-      let id = identifier.id;
-      let newId = coerceId(data.id);
+    If a merge occurs, it is possible the returned identifier does not match the originally
+    provided identifier. In this case the abandoned identifier will go through the usual
+    `forgetRecordIdentifier` codepaths.
+  */
+  updateRecordIdentifier(
+    identifier: StableRecordIdentifier,
```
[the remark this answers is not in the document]
runspired: shifted to RecordIdentifier to match rfc

packages/store/addon/-private/system/model/internal-model.ts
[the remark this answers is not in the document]
runspired: I verified that pre-any-identifiers work that mutating the id after it had been set once by any means would have erred. Have added a test for such and preserved that behavior. The id can update by means other than mutation (server communicated) but not via a set.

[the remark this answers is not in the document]
runspired: I would note though that in both cases (old and new) the error is not what we'd expect as it hits a codepath by which we alert the error under the assumption that the change originated from a save returning a different id, but this codepath is also taken by local mutations.

mike-north:
still working on reviewing this

packages/-ember-data/tests/integration/identifiers/configuration-test.ts
```diff
@@ -0,0 +1,560 @@
...
+                age: 31,
+              },
+            },
+          });
+        }
+      }
+      this.owner.register('adapter:application', TestAdapter);
+      this.owner.register('serializer:application', TestSerializer);
+
+      let updateMethodCalls = 0 as number;
```
mike-north: remove unnecessary cast

packages/-ember-data/tests/integration/identifiers/configuration-test.ts
```diff
@@ -0,0 +1,560 @@
...
+                age: 23,
+              },
+            },
+          });
+        }
+      }
+      this.owner.register('adapter:application', TestAdapter);
+      this.owner.register('serializer:application', TestSerializer);
+
+      let updateMethodCalls = 0 as number;
```
mike-north: remove unnecessary cast

packages/-ember-data/tests/integration/identifiers/lid-reflection-test.ts
```diff
@@ -0,0 +1,138 @@
...
+      class User extends Model {
+        @attr() name;
+        @attr() age;
+      }
+
+      owner.register('model:user', User);
+      store = owner.lookup('service:store');
+    });
+
+    test(`We can access the lid when serializing a record`, async function(assert) {
```
mike-north: doesn't need to be an async function

packages/-ember-data/tests/integration/identifiers/scenarios-test.ts
```diff
@@ -0,0 +1,678 @@
...
+
+if (IDENTIFIERS) {
+  module('Integration | Identifiers - scenarios', function(hooks) {
+    setupTest(hooks);
+
+    module('Secondary Cache based on an attribute', function(hooks) {
+      let store;
+      let calls;
+      let secondaryCache: {
+        id: Dict<string, string>;
```
mike-north: When did the `Dict` type get changed to allow two typeparams? (`{ [KK in K]: V; }`). It's now not really a dictionary (arbitrary keys) and matches nearly exactly with the built-in `Record<K, V>` utility type.

packages/store/addon/-private/system/coerce-id.ts
```diff
@@ -25,4 +24,19 @@ function coerceId(id: Coercable): string | null {
...
+  } else if (typeof id === 'number' && !isNaN(id)) {
+    normalized = '' + id;
+  }
+
+  if (DEBUG && normalized === null) {
+    throw new Error(`Expected id to be a string or number, recieved ${String(id)}`);
+  }
+
+  return normalized as string;
+}
```
mike-north: the non-null assertion operator is a safer cast for this kind of thing.
```ts
return normalized!;
```

to see the difference, look at the types of x and y below

```ts
function foo(): string | number | boolean | null {
  return '';
}
const x = foo()!;
const y = foo() as string;
```

packages/store/addon/-private/system/store/record-data-store-wrapper.ts
```diff
@@ -188,17 +208,9 @@ export default class RecordDataStoreWrapper implements IRecordDataStoreWrapper {
...
 
-    let internalModel = internalModelFactoryFor(this._store).peekId(modelName, id, clientId);
+    let internalModel = internalModelFactoryFor(this._store).peek(modelName, id, clientId);
     if (internalModel) {
       internalModel.destroyFromRecordData();
     }
   }
 }
-
-function hasValidId(id?: string | null, clientId?: string | null): id is string {
```
mike-north: This is a questionable type guard, and if misused could result in some "lying" in the type-checker. There's a significant difference between the static analysis behavior

`"true if id is a string"`

and the runtime behavior

`"true unless both id and clientId are truthy"`

We should take a close look at this before merging the PR

packages/-ember-data/tests/integration/identifiers/configuration-test.ts
```diff
@@ -0,0 +1,560 @@
...
+    });
+
+    hooks.afterEach(function() {
+      setIdentifierGenerationMethod(null);
+      setIdentifierResetMethod(null);
+      setIdentifierUpdateMethod(null);
+      setIdentifierForgetMethod(null);
+    });
+
+    test(`The configured generation method is used for pushed records`, async function(assert) {
```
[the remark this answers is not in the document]
runspired: All data tests need to be to avoid run loop wrapping

packages/-ember-data/tests/integration/identifiers/configuration-test.ts
```diff
@@ -0,0 +1,560 @@
...
+                age: 31,
+              },
+            },
+          });
+        }
+      }
+      this.owner.register('adapter:application', TestAdapter);
+      this.owner.register('serializer:application', TestSerializer);
+
+      let updateMethodCalls = 0 as number;
```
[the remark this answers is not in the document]
runspired: The type gets set to “0” even with the let due to the side-effect nature of how this counter is incremented, requiring a cast.

packages/-ember-data/tests/integration/identifiers/lid-reflection-test.ts
```diff
@@ -0,0 +1,138 @@
...
+if (IDENTIFIERS) {
+  module('Integration | Identifiers - lid reflection', function(hooks) {
+    setupTest(hooks);
+    let store;
+
+    hooks.beforeEach(function() {
+      const { owner } = this;
+      owner.register(`service:store`, Store);
+
+      class User extends Model {
```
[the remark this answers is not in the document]
runspired: Unfortunately, the usage of models as "static schema" results in class definitions leaking state between tests. This pattern is intentional to ensure that our tests do not fall victim to this leakage.  The pattern is prevalent within our tests and common enough a comment would likely not add much value.

packages/-ember-data/tests/integration/identifiers/configuration-test.ts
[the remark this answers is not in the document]
runspired: (this is also a good practice for tests more generally to help guard against leaks between tests)

packages/-ember-data/tests/integration/identifiers/configuration-test.ts
```diff
@@ -0,0 +1,560 @@
...
+      };
+
+      set(record, 'age', 23);
+
+      await record.save();
+
+      assert.ok(updateMethodCalls === 1, 'We made a single call to our update method after save');
+    });
+
+    test(`The reset method is called when the application is destroyed`, async function(assert) {
```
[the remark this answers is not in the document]
runspired: see previous reply

packages/-ember-data/tests/integration/identifiers/configuration-test.ts
runspired: see previous reply

packages/-ember-data/tests/integration/identifiers/lid-reflection-test.ts
runspired: see previous reply

packages/-ember-data/tests/integration/identifiers/scenarios-test.ts
runspired: Great catch for two reasons:

1) I was not aware of `Record<K, V>` being a built-in type and that explains why I've sometimes found that definition trying to trump our own defined `Record` type.

2) As far as I can recall this was the `Dict` we introduced together a while back.  I'll make a cleanup pass to refactor to `Dict<ValueType>`

runspired: (said cleanup pass will be separate from this PR as `Dict` has this <K, V>` usage more broadly)

packages/store/addon/-private/system/coerce-id.ts
runspired: I see the value more generally, in this case it should only ever be string but will make the update

packages/store/addon/-private/system/store/record-data-store-wrapper.ts
runspired: We've looked at this together before. It's basically a "caught between a rock and a hard place" situation where method signature overloads don't "fall through" to calls to other methods with similar signatures.  Here's an example https://typescript-play.js.org/#code/GYVwdgxgLglg9mABAEzgdQBQEMAqALGMAcwC5EBnKAJ0KIBpEsw4o8BTK-Wsym4xAD6IwIADaiAlGQBucGMgDcAKFCRYCFOmxdiParUHCxohkxbtOBXRX3EpiWfOWro8JKky4rpG3yKGRcVNmVg4dH14DIUDJGTlkRABvAF8lFXBXDVQADW1vPT9g8zD83yijcXtHRXS1N01cr24y-mjjItDLZsi7OKdazPc4RvCC8piOi1GW-zbKvoTEpUQVxAB6NcQOKjgqckQAIzYILBByNkZwxBh9gAMe2YrRW+XVqjYoECohz3DJktoEmUqQG6iGAE08t1bPRGCEpqUHgFjFV4s4MmDNJCmtYkXMTHDil1cTDUf0XJjUNjpninv9iREYcj5g54klXqsNogAEQAd12AGtyNyOYh3p9vppft5GPsHvTwkClCCKfVUAAtKEkwqEzo0pn4sk1VVZOCanGMvzMglmPWI0kLdF1U3m-VW-EK+3up5G9mrdabPmC4Wi8VfH5a2G2hEGLBy0nAtJcoNUIXcw4gKDCOC8xiIUQsRBwYCICBULAALwAnqX2BAhYx3tnc+8AI4gGDvBKERBQBD7YuIAC2HzwcGQ5CUXN57CQ3IQF0HoXOve8+yOsZm6d5cYcWFE8iwUDYCSw+YQ-lgI8ObGAuzYoLVcAAwuXq8+60LI2NiJ7od6YiNJ1Bk0V9KyrD9ji-C0f0eCZdRjbVAUdR9TTA99P3Ib8ZmtP8kNaH0Fj9VYYBLbB4QBfgAEIAF4aJ9Yj-TDSUPEjPCiCVf1ki2UQV1IxBI0QWj6MAxi3g+cMpTYhDKI45QuJ4lcln9FYuTANhpA4UUVlYHZc3U3MAFEqB2KgMFuAA5DSOFuTjVlSf1lSAA

packages/-ember-data/tests/integration/identifiers/configuration-test.ts
runspired: see previous reply

packages/-ember-data/tests/integration/identifiers/configuration-test.ts
```diff
@@ -0,0 +1,560 @@
...
+
+if (IDENTIFIERS) {
+  module('Integration | Identifiers - configuration', function(hooks) {
+    setupTest(hooks);
+    let store: Store;
+
+    hooks.beforeEach(function() {
+      const { owner } = this;
+
+      class User extends Model {
```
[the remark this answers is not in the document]
runspired: it is intentional

packages/-ember-data/tests/integration/identifiers/configuration-test.ts
```diff
@@ -0,0 +1,560 @@
...
+      const record = store.createRecord('user', { firstName: 'Chris', username: '@runspired', age: 31 });
+      const identifier = recordIdentifierFor(record);
+      assert.strictEqual(
+        identifier.lid,
+        'local:user:9000',
+        'Precond: We receive the expected identifier for a new record'
+      );
+      assert.strictEqual(identifier.id, null, 'Precond: We have no id yet');
+      assert.ok(updateMethodCalls === 0, 'Precond: We have not updated the identifier yet');
+      updateCallback = (updatedIdentifier, resource) => {
```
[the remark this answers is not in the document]
runspired: The point of this check is to ensure that it is not called after createRecord prior to save, so we don't expect it to be used but the test is ordered specifically to ensure that it is not used.

packages/store/addon/-private/system/store/record-data-store-wrapper.ts
runspired: I would also note that the near-term solution for not needing this craziness is to continue refactoring towards use of `identifiers`. Because identifiers are typed as a union of these two cases (one or the other must be a string) we reduce the number of method overloads and pass the "overloaded" type signature around with the value instead.

runspired:
2 test failures to investigate, only one blocking

- blocking: need to flag one new test as being dev-time only
- investigate: when all feature-flags are one, one identifiers test fails. unsure which other flag is causing this, but shouldn't block.

sandstrom:
Fantastic @runspired! ❤️

PR #6274, opened 2019-07-23
[BUGFIX RELEASE] Fix generating from blueprints

### LINKED ISSUE #6263: CI unskip node:test for blueprints
We skipped them here: https://github.com/emberjs/data/pull/6241

TL;DR the blueprint test helper does not have a way for us to configure the app environment and so both `ember-data` and the specific package (such as `@ember-data/model`) will be top-level dependencies. The blueprint re-export from `ember-data` is winning out and does not work in this configuration environment. We need to improve both ember-cli-bluepriint-test-helpers and ember-cli to enable the environment to be configured to either ensure the correct blueprint is used or remove `ember-data` from it.

cc @Turbo87 @rwjblue for CLI needs.

Gaurav0:
This fixes unable to generate models, adapters, serializers, and transforms.

Note: This is my first attempt to commit to ember-data.

Also fixes #6263 

Please note that we could not just override `path` or `filesPath` because `path` gets overwritten on `init` and `filesPath` is patched from outside the blueprints object.

Update: Rebased against master

### DIFF
--- a/packages/-build-infra/src/utilities/edition-detector.js
+++ b/packages/-build-infra/src/utilities/edition-detector.js
@@ -5,7 +5,7 @@ const path = require('path');
 module.exports = function(blueprint) {
   blueprint.filesPath = function() {
     let rootPath = process.env.EMBER_VERSION === 'OCTANE' ? 'native-files' : 'files';
-    return path.join(blueprint.blueprintPath, rootPath);
+    return path.join(blueprint.root, rootPath);
   };
 
   return blueprint;
--- a/packages/-build-infra/src/utilities/edition-detector.js
+++ b/packages/-build-infra/src/utilities/edition-detector.js
@@ -5,7 +5,7 @@ const path = require('path');
 module.exports = function(blueprint) {
   blueprint.filesPath = function() {
     let rootPath = process.env.EMBER_VERSION === 'OCTANE' ? 'native-files' : 'files';
-    return path.join(this.path, rootPath);
+    return path.join(blueprint.blueprintPath, rootPath);
   };
 
   return blueprint;
--- a/packages/-build-infra/src/utilities/test-framework-detector.js
+++ b/packages/-build-infra/src/utilities/test-framework-detector.js
@@ -15,7 +15,7 @@ module.exports = function(blueprint) {
     let dependencies = this.project.dependencies();
 
     if ('ember-qunit' in dependencies) {
-      if (fs.existsSync(blueprint.blueprintPath + '/qunit-rfc-232-files')) {
+      if (fs.existsSync(blueprint.root + '/qunit-rfc-232-files')) {
         type = 'qunit-rfc-232';
       } else {
         type = 'qunit';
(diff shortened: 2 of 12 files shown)


packages/-build-infra/src/utilities/test-framework-detector.js
```diff
@@ -15,21 +15,21 @@ module.exports = function(blueprint) {
...
-      if (fs.existsSync(this.path + '/qunit-rfc-232-files')) {
+      if (fs.existsSync(blueprint.blueprintPath + '/qunit-rfc-232-files')) {
         type = 'qunit-rfc-232';
       } else {
         type = 'qunit';
       }
     } else if ('ember-cli-qunit' in dependencies) {
       let checker = new VersionChecker(this.project);
-      if (fs.existsSync(this.path + '/qunit-rfc-232-files') && checker.for('ember-cli-qunit', 'npm').gte('4.2.0')) {
+      if (fs.existsSync(blueprint.blueprintPath + '/qunit-rfc-232-files') && checker.for('ember-cli-qunit', 'npm').gte('4.2.0')) {
```
efx: I think these need to be reformatted per the [linting](https://dev.azure.com/ember-data/emberjs/_build/results?buildId=772&view=logs&jobId=3a559e2a-952e-58d2-b8db-2e604a9266d7) errors.
You can easily do this by invoking:
```js
./node_modules/.bin/prettier --write packages/-build-infra/src/utilities/test-framework-detector.js
```

Gaurav0: I fixed just now using `eslint --fix --plugin prettier .`

packages/-build-infra/src/utilities/edition-detector.js
```diff
@@ -5,7 +5,7 @@ const path = require('path');
 module.exports = function(blueprint) {
   blueprint.filesPath = function() {
     let rootPath = process.env.EMBER_VERSION === 'OCTANE' ? 'native-files' : 'files';
-    return path.join(this.path, rootPath);
+    return path.join(blueprint.blueprintPath, rootPath);
```
rwjblue: I’d vaguely prefer to use `root` vs `blueprintPath`. On addons `root` has the same meaning as this property (it’s the directory that the addon itself is in), and `blueprintPath` is actually a hook (aka method) that returns the parent directory of all the addons’ blueprints. 

Tldr; I’d like to avoid the confusion of using a common name (`blueprintPath`) in a different / confusing way.

Gaurav0: It's not the directory the addon itself is in, it's where the blueprint is, i.e. root/blueprints/<blueprint_name>

rwjblue: Yep, I know 😃. My point is that I’d like `this.root` to always mean “this things root path” whether that thing is an addon, a blueprint, a command, etc. 

Using the name `blueprintPath` here is confusing because the same name on an addon means a **completely** different thing.

Gaurav0: Ok, changing it.

Issue #6255, opened 2019-07-18
Move unnecessary dependencies to devDependencies


andreyfel:
I've updated to Ember Data 3.11 and noticed a bunch of new dependencies in the lock file.
Some of them look like they shouldn't be there.
For example `@ember-data/-build-infra` has `mocha` and `chai` (through `ember-cli-blueprint-test-helpers`).
Can they be moved to devDependencies?

runspired:
I don't believe so in the current state as `-build-infra` is providing them as dependencies for the other packages. We could potentially separate `-build-infra` into two packages: one for dev-time things and the other for shared infra that each package needs to ship with.

rwjblue:
Ya, you are correct. Would you mind sending a PR to fix?

snewcomer:
Anyway I can help here?  What shared infra is needed to ship with each package?

runspired:
Resolved by #6291 

@snewcomer if you're interested in helping here's a good quick hitter: https://github.com/emberjs/data/issues/6267

andreyfel:
Thanks guys for that fix! Do you plan to make another patch release of 3.11 with this change in place?

PR #6291, opened 2019-07-29
[BUGFIX] Move unnecessary dependencies to devDependencies

Gaurav0:
For https://github.com/emberjs/data/issues/6255

Created a new package, -test-infra, that should be able to be in devdependencies only.

Also hoists and syncs all devdependencies and dependencies.


rwjblue:
> I noticed that -build-infra only contains node js files, so made it a devDependency wherever used.

I think we actually do need `-build-infra` as a dependency in each of the packages. Specifically, they are used by the other packages in the `index.js` for building the addon itself (which is definitely needed when consumed).

runspired:
Confirm what Rob said. The test-helper bits should probably be in a different package from the build-utils bits. I'm thinking `@ember-data/-build-infra` for these built utils and moving the stuff that can be in dev-deps to a new `@ember-data/-dev-infra` package.

packages/adapter/node-tests/blueprints/adapter-test.js
```diff
@@ -12,10 +12,10 @@ const expect = chai.expect;
 
 const SilentError = require('silent-error');
 
-const generateFakePackageManifest = require('@ember-data/-build-infra/src/node-test-helpers/generate-fake-package-manifest');
-const fixture = require('@ember-data/-build-infra/src/node-test-helpers/fixture');
+const generateFakePackageManifest = require('../../../-test-infra/src/node-test-helpers/generate-fake-package-manifest');
```
runspired: relative path should be absolute

packages/adapter/node-tests/blueprints/adapter-test.js
```diff
@@ -12,10 +12,10 @@ const expect = chai.expect;
...
+const fixture = require('@ember-data/-test-infra/src/node-test-helpers/fixture');
 
-const setupTestEnvironment = require('@ember-data/-build-infra/src/node-test-helpers/setup-test-environment');
+const setupTestEnvironment = require('../../../-test-infra/src/node-test-helpers/setup-test-environment');
```
runspired: relative path should be absolute

packages/model/node-tests/blueprints/model-test.js
```diff
@@ -9,10 +9,10 @@ const modifyPackages = blueprintHelpers.modifyPackages;
 const chai = require('ember-cli-blueprint-test-helpers/chai');
 const expect = chai.expect;
 
-const generateFakePackageManifest = require('@ember-data/-build-infra/src/node-test-helpers/generate-fake-package-manifest');
-const fixture = require('@ember-data/-build-infra/src/node-test-helpers/fixture');
+const generateFakePackageManifest = require('../../../-test-infra/src/node-test-helpers/generate-fake-package-manifest');
```
runspired: relative path should be absolute

packages/model/node-tests/blueprints/model-test.js
```diff
@@ -9,10 +9,10 @@ const modifyPackages = blueprintHelpers.modifyPackages;
...
+const fixture = require('@ember-data/-test-infra/src/node-test-helpers/fixture');
 
-const setupTestEnvironment = require('@ember-data/-build-infra/src/node-test-helpers/setup-test-environment');
+const setupTestEnvironment = require('../../../-test-infra/src/node-test-helpers/setup-test-environment');
```
runspired: relative path should be absolute

packages/serializer/node-tests/blueprints/serializer-test.js
```diff
@@ -12,10 +12,10 @@ const expect = chai.expect;
 
 const SilentError = require('silent-error');
 
-const generateFakePackageManifest = require('@ember-data/-build-infra/src/node-test-helpers/generate-fake-package-manifest');
-const fixture = require('@ember-data/-build-infra/src/node-test-helpers/fixture');
+const generateFakePackageManifest = require('../../../-test-infra/src/node-test-helpers/generate-fake-package-manifest');
```
runspired: relative path should be absolute

packages/serializer/node-tests/blueprints/serializer-test.js
```diff
@@ -12,10 +12,10 @@ const expect = chai.expect;
...
+const fixture = require('@ember-data/-test-infra/src/node-test-helpers/fixture');
 
-const setupTestEnvironment = require('@ember-data/-build-infra/src/node-test-helpers/setup-test-environment');
+const setupTestEnvironment = require('../../../-test-infra/src/node-test-helpers/setup-test-environment');
```
runspired: relative path should be absolute

packages/serializer/node-tests/blueprints/transform-test.js
```diff
@@ -9,10 +9,10 @@ const modifyPackages = blueprintHelpers.modifyPackages;
 const chai = require('ember-cli-blueprint-test-helpers/chai');
 const expect = chai.expect;
 
-const generateFakePackageManifest = require('@ember-data/-build-infra/src/node-test-helpers/generate-fake-package-manifest');
-const fixture = require('@ember-data/-build-infra/src/node-test-helpers/fixture');
+const generateFakePackageManifest = require('../../../-test-infra/src/node-test-helpers/generate-fake-package-manifest');
```
runspired: relative path should be absolute

packages/serializer/node-tests/blueprints/transform-test.js
```diff
@@ -9,10 +9,10 @@ const modifyPackages = blueprintHelpers.modifyPackages;
...
+const fixture = require('@ember-data/-test-infra/src/node-test-helpers/fixture');
 
-const setupTestEnvironment = require('@ember-data/-build-infra/src/node-test-helpers/setup-test-environment');
+const setupTestEnvironment = require('../../../-test-infra/src/node-test-helpers/setup-test-environment');
```
runspired: relative path should be absolute

package.json
```diff
@@ -36,6 +36,8 @@
   "devDependencies": {
     "@babel/plugin-transform-typescript": "^7.5.5",
     "@ember/optional-features": "^0.7.0",
+    "@ember/ordered-set": "^2.0.3",
+    "@glimmer/env": "^0.1.7",
```
runspired: glimer/env and ordered-set being added here seems suspect

packages/-test-infra/package.json
```diff
@@ -0,0 +1,23 @@
...
+  "license": "MIT",
+  "author": "",
+  "directories": {
+    "doc": "doc",
+    "test": "tests"
+  },
+  "scripts": {
+  },
+  "dependencies": {
+    "ember-cli-blueprint-test-helpers": "^0.19.1"
```
runspired: I suspect more should be here (like mocha)

package.json
```diff
@@ -45,59 +47,86 @@
     "@types/qunit": "^2.5.3",
     "@types/rsvp": "^4.0.3",
     "babel-eslint": "^10.0.2",
+    "babel-plugin-debug-macros": "^0.3.2",
```
runspired: a lot more things were added to this file than I would have thought, I feel many likely should have been placed into the test-infra package?

packages/-test-infra/package.json
Gaurav0: mocha isn't 1`require`d by any of the files.

package.json
Gaurav0: They are being required in one of the packages, and I hoisted all of the dependencies to the main package.json. See comment below.

package.json
Gaurav0: I hoisted all dependencies to the main package.json file, and synchronized versions. Putting them here as well as in the individual package.json files in each package ensures that the same version is used across all packages.

The other reason I did this is I got warnings running https://github.com/mybigday/yarn-workspace-sync-deps

runspired: there were a few things that needed fixed (I noticed a few dangling dev-deps with "*") but there's already a `sync` command within the repo that does what we want. It looks like this is hoisting `dependencies` in addition to `devDependencies` which is not something I think we want at this point.

Gaurav0: I got errors running that script.

PR #6293, opened 2019-07-30
Make rest adapter test pass for fetch

### LINKED ISSUE #6083: tests: a fetch query seems not to be catched by Pretender
https://github.com/emberjs/data/pull/5975 does the CI setup for testing with both opting in and out of jQuery. As a consequence, the adapter test suite is run with both `fetch` and `ajax`.

When using ajax, [this test](https://github.com/emberjs/data/pull/5975/files#diff-ce1f08de89610ad4a9e80807e91536cdR2862) is running fine. With fetch, the network call is not catched by Pretender.

For now, the test is scoped to jQuery. But it could enabled when jQuery is opted out.

See this conversation for details: https://github.com/emberjs/data/pull/5975/files#r280937905

Gaurav0:
Fixes #6083 

The issue wasn't pretender after all. ember-fetch overrides ajaxError so we have to move the debugging code.

### DIFF
--- a/packages/-ember-data/tests/integration/adapter/rest-adapter-test.js
+++ b/packages/-ember-data/tests/integration/adapter/rest-adapter-test.js
@@ -2665,7 +2665,6 @@ module('integration/adapter/rest_adapter - REST Adapter', function(hooks) {
     });
   });
 
-
   testInDebug('warns when an empty response is returned, though a valid stringified JSON is expected', function(
     assert
   ) {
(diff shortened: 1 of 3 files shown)


bin/publish.js
```diff
@@ -142,11 +142,7 @@ function assertGitIsClean() {
   }
 
   let expectedChannelBranch =
-    options.distTag === 'canary'
-      ? 'master'
-      : options.distTag === 'latest'
-      ? 'release'
-      : options.distTag;
+    options.distTag === 'canary' ? 'master' : options.distTag === 'latest' ? 'release' : options.distTag;
```
runspired: seems like accidental reformatting, after rebasing make sure this file isn't in your change diff still

Gaurav0: Went away upon rebase

PR #6301, opened 2019-08-01
[BUGFIX] Await potentially async operations

pzuraq:
Some operations in Ember Data flush using `run.join`. This will flush
_synchronously_ if no runloop or autorun currently exists, but will
join the existing runloop otherwise, making the scheduled task
effectively asynchronous.

With the tracked properties feature flag enabled, the way we've
scheduled autoruns has changed. A few Data tests were failing because
they were expecting an operation to flush synchronously, but it flushed
asynchronously due to an existing autorun. This PR adds appropriate
`await settled();` statements in these cases.

### DIFF
--- a/packages/-ember-data/tests/unit/model/rollback-attributes-test.js
+++ b/packages/-ember-data/tests/unit/model/rollback-attributes-test.js
@@ -12,6 +12,7 @@ import { module, test } from 'qunit';
 import DS from 'ember-data';
 import { setupTest } from 'ember-qunit';
 import { settled } from '@ember/test-helpers';
+import { gte } from 'ember-compatibility-helpers';
 
 let env, store, Person;
 
@@ -367,7 +368,11 @@ module('unit/model/rollbackAttributes - model.rollbackAttributes()', function(ho
   });
 
   test("invalid record's attributes can be rollbacked", async function(assert) {
-    assert.expect(12);
+    if (gte('3.13.0')) {
+      assert.expect(14);
+    } else {
+      assert.expect(13);
+    }
 
     class Dog extends Model {
       @attr() name;
@@ -438,7 +443,11 @@ module('unit/model/rollbackAttributes - model.rollbackAttributes()', function(ho
   });
 
   test(`invalid record's attributes rolled back to correct state after set`, async function(assert) {
-    assert.expect(14);
+    if (gte('3.13.0')) {
+      assert.expect(15);
+    } else {
+      assert.expect(14);
+    }
 
     class Dog extends Model {
       @attr() name;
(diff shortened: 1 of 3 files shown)


packages/-ember-data/tests/integration/record-array-manager-test.js
```diff
@@ -256,11 +256,11 @@ module('integration/record_array_manager', function(hooks) {
     assert.deepEqual(internalModel._recordArrays.toArray(), [recordArray]);
   });
 
-  test('liveRecordArrayFor always return the same array for a given type', function(assert) {
+  test('liveRecordArrayFor always return the same array for a given type', async function(assert) {
     assert.equal(manager.liveRecordArrayFor('foo'), manager.liveRecordArrayFor('foo'));
   });
 
-  test('liveRecordArrayFor create with content', function(assert) {
+  test('liveRecordArrayFor create with content', async function(assert) {
```
rwjblue: Why are some of these changed to `async` functions without any `await`'s in them?

runspired: it prevents leakage for unsettled things we don't care about, which is fairly common. It effectively adds `await settled()` to every test.

runspired:
I opened https://github.com/emberjs/ember.js/issues/18225 for the failing model tests, as it represents a regression in observer behavior that may have performance impacts.

packages/-ember-data/tests/integration/record-array-manager-test.js
runspired: After OOB discussion @rwjblue has opened up this PR to enable us to not add unnecessary `async` tests: https://github.com/emberjs/ember-qunit/pull/528

rwjblue:
I'd rather update to https://github.com/emberjs/ember-qunit/releases/tag/v4.5.1 than add `async` to all the tests that aren't intentionally using asynchrony. 

@runspired - This is the change we discussed in discord, basically in ember-qunit@4.5.1 we now allow a small "settling" period before triggering the test isolation errors.

rwjblue:
https://github.com/emberjs/data/pull/6325 does that ember-qunit update

runspired:
dropped commits for unnecessary async tests now that #6325 has landed

packages/-ember-data/tests/unit/model/rollback-attributes-test.js
```diff
@@ -367,7 +368,11 @@ module('unit/model/rollbackAttributes - model.rollbackAttributes()', function(ho
   });
 
   test("invalid record's attributes can be rollbacked", async function(assert) {
-    assert.expect(12);
+    if (gte('3.13.0')) {
```
rwjblue: I think these should be `3.13.0-beta.1`, right? Otherwise it will continue to fail until 3.13.0 is released as stable.

runspired: Gte treats beta as satisfying this

Issue #1540, opened 2013-12-03
isDirty based on strict comparison


Herriau:
JSFiddle: http://jsfiddle.net/U6Xc3/

It seems like to determine whether a record is dirty Ember Data uses strict comparison between the original (last committed) value and the new value. While this may arguably make sense at first, it ends up posing some challenges.

Indeed, as soon as you start playing with 2-way binding, what used to be an integer or a date may become a string, and even though the intrinsic value has not really changed (e.g. `123` probably has the same meaning as `"123"`) Ember Data will consider that value dirty. 

This becomes even more challenging as you start playing with more advanced form components that are bound to non-primitive values such as arrays (e.g. `([2,3,5] === [2,3,5]) === false`). The strict comparison operator determines whether the two references point to the same object but it certainly isn't of any help when it comes to comparing the actual values conveyed by those objects.

The faulty piece of code in Ember Data:

``` javascript
var didSetProperty = function(record, context) {
  if (context.value === context.originalValue) {
    delete record._attributes[context.name];
    record.send('propertyWasReset', context.name);
  } else if (context.value !== context.oldValue) {
    record.send('becomeDirty');
  }

  record.updateRecordArraysLater();
};
```

At the very least you should consider using a coercive comparison (`==`), or an even more tolerant comparison such as `a.valueOf() == b.valueOf()` when possible. Also, if the operands implement a comparison method with a conventional name such as `equals()` or `isEqual()` it could be used instead.

I don't see how I can properly override this default behavior beside overriding all the substates in `DS.RootState` with a more tolerant `didSetProperty`, which is far from being ideal. Another solution that I considered is to specialize some Ember components (e.g. having a text field for dates, etc...) that would cast the value to one of the primitive types, but that does not really apply to all cases, this is a bit of a pain and it certainly deviates from the way things should be elegantly written using Ember.

superlou:
I think I'm hitting this issue when I'm trying to set a DS.attr('date').  Would be interested in a baked in solution.

oriSomething:
is there is a new about that? `isDirty` is a real pain with `array`s

jpaas:
The ability to manage embedded objects/arrays and track whether they or dependent hasMany are dirty is killing me. Any solutions people have worked out in the past year are all broken on ember-data 1.19. Has anyone found anything that works on 1.19?

andrewbranch:
[...]

PR #6306, opened 2019-08-05
[BUGFIX store-push]

azhiv:
Added test to demonstrate how store.push incorrectly handles date attribute

### DIFF
--- a/packages/-ember-data/tests/unit/store/push-test.js
+++ b/packages/-ember-data/tests/unit/store/push-test.js
@@ -17,6 +17,7 @@ module('unit/store/push - DS.Store#push', function(hooks) {
     Person = DS.Model.extend({
       firstName: attr('string'),
       lastName: attr('string'),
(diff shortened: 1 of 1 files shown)


runspired:
Hi @azhiv 

This is not a bug but (persay) but rather you discovering the limitations described in https://github.com/emberjs/data/issues/1540

Because you are using a class instance (`new Date()`) the dates will not be referentially equal. I personally would not recommend using transforms to produce class instances but instead push the serialized date string into the store and instantiate a new date object for the model either via a computed property, a helper/util, or in very advanced scenarios either a custom RecordData or a custom RecordClass (coming soon, see recent RFCs).

azhiv:
Thanks for your answer, @runspired. It's a pity there's no general/easy-to-use solution yet.
In our workflow we ended up calling `rollbackAttributes()` before the push to get rid of all these side effects.

Issue #6219, opened 2019-07-03
`shouldRollupPrivate: false` produces a broken build


ef4:
I realize this may not be public API, but the `shouldRollupPrivate` flag that now exists in `@ember-data/-build-infra` doesn't appear to work.

The problem is that if you set it to `false`, you get a build that lacks the `version.js` file.

runspired:
Sounds like a good thing to fix but confirm it should not be public.

PR #6312, opened 2019-08-06
[BETA] Add environment variable to opt-out of rolling up `-private`

rwjblue:
This allows us to:

* Gather data about size per `-private` module
* Gather usage details about each `-private` module (via Chrome's coverage stats)
* Test the overall application performance impact of doing the `rollup` stages (do we still need them? how much impact does it have?)


runspired:
Resolves #6219

rwjblue:
> also there seems to be a lint error :)

Indeed, but its in a file I haven't touched 😩

rwjblue:
https://github.com/emberjs/data/pull/6315 fixes that linting issue

PR #6289, opened 2019-07-27
 [CHORE identifiers] refactor internals to use identifiers

runspired:
Some final prep (added tests / internal factorings) to prepare for identifiers being turned on

PR #6313, opened 2019-08-06
Custom Model Classes

igorT:
Implements RFC https://github.com/emberjs/rfcs/pull/487


packages/serializer/addon/json-api.js
```diff
@@ -493,7 +494,7 @@ const JSONAPISerializer = JSONSerializer.extend({
 
     if (this._canSerialize(key)) {
       let belongsTo = snapshot.belongsTo(key);
-      let belongsToIsNotNew = belongsTo && belongsTo.record && !belongsTo.record.get('isNew');
+      let belongsToIsNotNew = belongsTo && belongsTo.record && !belongsTo.isNew;
```
runspired: since we no longer check record.isNew we don't need to check belongsTo.record

packages/store/addon/-private/ts-interfaces/record.ts
```diff
@@ -1,3 +1,5 @@
+import EmberObject from '@ember/object';
```
runspired: unnecessary import

packages/store/addon/-private/ts-interfaces/record.ts
```diff
@@ -12,12 +14,14 @@
...
 type Primitive = string | number | boolean | null;
 interface Object {
   [member: string]: Value | undefined | ((...args: any[]) => any);
 }
 interface Arr extends Array<Object | Arr | undefined> {}
 
 type Value = Primitive | Object | Arr;
+*/
 
 export type Record = Object;
```
runspired: rename to `RecordInstance` in a new PR

packages/store/addon/-private/ts-interfaces/record-data-record-wrapper.ts
```diff
@@ -0,0 +1,50 @@
...
+
+import { RecordIdentifier } from './identifier';
+import { ChangedAttributesHash } from './record-data';
+
+/**
+  @module @ember-data/store
+*/
+
+export default interface RecordDataRecordWrapper {
+  rollbackAttributes(): string[];
```
runspired: once we have the proxy we need to ensure it is in the ts-upgrade-map

packages/store/addon/-private/system/store/record-data-store-wrapper.ts
```diff
@@ -19,12 +20,10 @@ const MISSING_ID_ARG_ERROR_MESSAGE = `Either an id or a clientId is required as
 
 export default class RecordDataStoreWrapper implements IRecordDataStoreWrapper {
   [BRAND_SYMBOL]: 'RecordDataStoreWrapper';
-  _store: Store;
   _willUpdateManyArrays: boolean;
   private _pendingManyArrayUpdates: StringOrNullOrUndefined[];
 
-  constructor(store: Store) {
-    this._store = store;
+  constructor(public _store: CoreStore) {
```
runspired: where are we still accessing this outside of the class?

packages/store/addon/-private/system/store/record-data-store-wrapper.ts
```diff
@@ -52,6 +51,14 @@ export default class RecordDataStoreWrapper implements IRecordDataStoreWrapper {
   _scheduleManyArrayUpdate(modelName: string, id: string | null, clientId: string, key: string): void;
   _scheduleManyArrayUpdate(modelName: string, id: string, clientId: string | null | undefined, key: string): void;
   _scheduleManyArrayUpdate(modelName: string, id: string | null, clientId: string | null | undefined, key: string) {
+    if (!hasValidId(id, clientId)) {
```
runspired: I think this goes away if we land #6289

packages/store/addon/-private/system/store/record-data-store-wrapper.ts
```diff
@@ -114,15 +123,32 @@ export default class RecordDataStoreWrapper implements IRecordDataStoreWrapper {
   inverseForRelationship(modelName: string, key: string): string {
     const modelClass = this._store.modelFor(modelName);
     const definition = upgradeForInternal(this.relationshipsDefinitionFor(modelName)[key]);
-
-    return definition._inverseKey(this._store, modelClass);
+    if (CUSTOM_MODEL_CLASS) {
+      if (definition.inverse !== undefined) {
+        return definition.inverse;
+      } else {
+        return definition._inverseKey(this._store, modelClass);
```
runspired: what if a custom model class does not provide a definition?

packages/store/addon/-private/system/store/record-data-store-wrapper.ts
```diff
@@ -114,15 +123,32 @@ export default class RecordDataStoreWrapper implements IRecordDataStoreWrapper {
...
   }
 
   inverseIsAsyncForRelationship(modelName: string, key: string): boolean {
     const modelClass = this._store.modelFor(modelName);
     const definition = upgradeForInternal(this.relationshipsDefinitionFor(modelName)[key]);
-
-    return definition._inverseIsAsync(this._store, modelClass);
+    if (CUSTOM_MODEL_CLASS) {
+      if (definition.inverse === null) {
+        return true;
```
runspired: why would this be true?

packages/store/addon/-private/system/snapshot.js
```diff
@@ -85,6 +90,29 @@ export default class Snapshot {
...
+    if (attributes === null) {
+      let record = this.record;
+      attributes = this.__attributes = Object.create(null);
+      let attrs = Object.keys(this._store._attributesDefinitionFor(this.modelName, record.id));
+      if (CUSTOM_MODEL_CLASS) {
+        attrs.forEach(keyName => {
+          if (this.type.isModel) {
+            attributes[keyName] = get(record, keyName);
+          } else {
+            attributes[keyName] = this._internalModel._recordData.getAttr(keyName);
```
runspired: lets use recordData directly instead of relying on _internalModel, either via stashing a prop or helper lookup

packages/store/addon/-private/system/snapshot-record-array.js
```diff
@@ -124,6 +124,14 @@ export default class SnapshotRecordArray {
   get type() {
     return this._type || (this._type = this._recordArray.get('type'));
   }
+  /**
+    The modelName of the underlying records for the snapshots in the array, as a DS.Model
+    @property type
+    @type {DS.Model}
+  */
+  get modelName() {
+    return this._recordArray.modelName;
```
runspired: check whether `_type` here is the class or just an alias to modelName

packages/store/addon/-private/system/references/reference.ts
```diff
@@ -30,16 +29,13 @@ export default abstract class Reference {
   public recordData: InternalModel['_recordData'];
   constructor(
     // TODO: shouldn't have to instance<factory of instance>
```
runspired: remove

packages/store/addon/-private/system/schema-definition-service.ts
```diff
@@ -0,0 +1,141 @@
...
+
+export class SchemaDefinitionService {
+  private _modelFactoryCache = Object.create(null);
+  private _relationshipsDefCache = Object.create(null);
+  private _attributesDefCache = Object.create(null);
+
+  constructor(public store: Store) {}
+
+  // Following the existing RD implementation
+  attributesDefinitonFor(identifier: RecordIdentifier | string): AttributesSchema {
```
runspired: spelling

packages/store/addon/-private/system/record-notification-manager.ts
```diff
@@ -0,0 +1,49 @@
...
+/*
+  Currently only support a single callback per identifier
+*/
+export default class NotificationManager {
+  constructor(private store: CoreStore) {}
+
+  subscribe(identifier: RecordIdentifier, callback: NotificationCallback): UnsubscribeToken {
+    let stableId = this.store._stableIdentifierFor(identifier);
+    Cache.set(stableId, callback);
+    let unsubToken = new Object();
```
runspired: weird way of generating something?

packages/store/addon/-private/system/model/shim-model-class.ts
```diff
@@ -0,0 +1,63 @@
+import { map } from '@ember/object/computed';
+import CoreStore from '../core-store';
+import { stringify } from 'querystring';
+import { RelationshipSchema, AttributeSchema } from '../../ts-interfaces/record-data-schemas';
+import { Dict } from '../../ts-interfaces/utils';
+
+// Mimics the static apis of DSModel
+export default class ShimModelClass {
+  store: CoreStore;
```
runspired: I do not believe this is a static API?

packages/store/addon/-private/system/model/shim-model-class.ts
```diff
@@ -0,0 +1,63 @@
...
+  modelName: string;
+  _id?: string;
```
runspired: we shouldn't have `_id` mixed in with schema

runspired: since we need this, we should mark private / symbolicate or some such

packages/store/addon/-private/system/model/shim-model-class.ts
```diff
@@ -0,0 +1,63 @@
...
+
+  // TODO Maybe expose the class here?
+  constructor(store: CoreStore, modelName: string, id?: string) {
+    this.store = store;
+    this.modelName = modelName;
+    this._id = id;
+  }
+
+  get fields(): Map<string, 'attribute' | 'belongsTo' | 'hasMany'> {
```
runspired: we should fast follow a deprecation, I believe as part of the RFC we were looking at the shim being a place to do. Seems good to have full API support at the beginning though.

packages/store/addon/-private/system/model/shim-model-class.ts
```diff
@@ -0,0 +1,63 @@
...
+  get attributes(): Map<string, AttributeSchema> {
+    let attrs = this.store._attributesDefinitionFor(this.modelName, this._id);
+    //TODO Talk to mnorth about needing a cast
+    return new Map(Object.entries(attrs) as [string, AttributeSchema][]);
+  }
+
+  get relationshipsByName(): Map<string, RelationshipSchema> {
+    let relationships = this.store._relationshipsDefinitionFor(this.modelName, this._id);
+    //TODO Talk to mnorth about needing a cast
+    return new Map(Object.entries(relationships) as [string, RelationshipSchema][]);
```
runspired: are we worried about these being uncached getters?

runspired: if we can't cache we should document why we can't assume cacheable and how users should approach using these APIs when they are cacheable.

packages/store/addon/-private/system/core-store.ts
```diff
@@ -341,6 +361,140 @@ class Store extends Service {
...
+      return record;
+    } else {
+      throw 'should not be here, custom model class ff error';
+    }
+  }
+
+  abstract instantiateRecord(
+    identifier: StableRecordIdentifier,
+    createRecordArgs: { [key: string]: any }, // args passed in to store.createRecord() and processed by recordData to be set on creation
+    recordDataFor: (identifier: StableRecordIdentifier) => RecordDataRecordWrapper,
```
runspired: the signature here should be `RecordIdentifier` not `StableRecordIdentifier` as this is a userland called API

packages/store/addon/-private/system/core-store.ts
```diff
@@ -2737,21 +2840,36 @@ class Store extends Service {
       // contains unknown attributes or relationships, log a warning.
 
       if (ENV.DS_WARN_ON_UNKNOWN_KEYS) {
-        let modelClass = this.modelFor(modelName);
+        let unknownAttributes, unknownRelationships;
```
runspired: this will only work for the case where the schemas for all `type` are identical, not the one where they may differ by identifier.

runspired: (this is a low priority fix)

packages/store/addon/-private/system/ds-model-store.ts
```diff
@@ -0,0 +1,373 @@
...
+  `store#peekAll()` or `store#findAll()`. This means any
+  data bindings or computed properties that depend on the RecordArray
+  will automatically be synced to include the new or updated record
+  values.
+
+  @class Store
+  @namespace DS
+  @extends Ember.Service
+*/
+function notifyChanges(
```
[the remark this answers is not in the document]
runspired: also the signature should likely be `StableRecordIdentifier` since this is fully internal

packages/store/addon/-private/system/model/internal-model.ts
```diff
@@ -449,7 +468,8 @@ export default class InternalModel {
     this._doNotDestroy = false;
 
     if (this._record) {
-      this._record.destroy();
+      this.store.teardownRecord(this._record);
+      //this._record.destroy();
```
runspired: ff removal since `teardownRecord` is sub-classable and thus this behavior could be done away with

packages/store/addon/-private/system/ds-model-store.ts
```diff
@@ -0,0 +1,383 @@
...
+    if (CUSTOM_MODEL_CLASS) {
+      return this._relationshipsDefinitionFor(modelName)[key];
+    } else {
+      let modelClass = this.modelFor(modelName);
+      let relationshipsByName = get(modelClass, 'relationshipsByName');
+      return relationshipsByName.get(key);
+    }
+  }
+
+  _attributesDefinitionFor(modelName) {
```
runspired: double-check the `identifiers` codepath for retrieving attributes does not pull from this info

runspired: (same for relationships definition)

packages/store/addon/-private/system/ds-model-store.ts
```diff
@@ -0,0 +1,383 @@
...
+  }
+
+  getSchemaDefinitionService(): SchemaDefinitionService {
+    if (CUSTOM_MODEL_CLASS) {
+      if (!this._schemaDefinitionService) {
+        this._schemaDefinitionService = new SchemaDefinitionService(this);
+      }
+      return this._schemaDefinitionService;
+    } else {
+      throw 'schema service is only available when custom model class feature flag is on';
```
runspired: new Error(

packages/store/addon/-private/system/core-store.ts
```diff
@@ -341,6 +362,140 @@ class Store extends Service {
...
+      //recordToInternalModelMap.set(record, internalModel);
+      return record;
+    } else {
+      throw 'should not be here, custom model class ff error';
+    }
+  }
+
+  abstract instantiateRecord(
+    identifier: StableRecordIdentifier,
+    createRecordArgs: { [key: string]: any }, // args passed in to store.createRecord() and processed by recordData to be set on creation
```
mike-north: values should be `unknown`

packages/store/addon/-private/system/core-store.ts
```diff
@@ -341,6 +362,140 @@ class Store extends Service {
...
+    recordDataFor: (identifier: StableRecordIdentifier) => RecordDataRecordWrapper,
+    notificationManager: NotificationManager
+  ): Record;
+
+  abstract teardownRecord(record): void;
+
+  _internalDeleteRecord(internalModel) {
```
mike-north: TODO: type this

packages/store/addon/-private/system/core-store.ts
```diff
@@ -2904,11 +2988,49 @@ class Store extends Service {
     return internalModelFactoryFor(this).lookup(modelName, id, lid);
   }
 
-  _createRecordData(modelName, id, clientId): RecordData {
+  serializeRecord(record: Record, options?: Dict<string, unknown>): unknown {
+    if (CUSTOM_MODEL_CLASS) {
+      let identifier = recordIdentifierFor(record);
+      let internalModel = internalModelFactoryFor(this).peek(identifier.type, identifier.id, identifier.lid);
+      // TODO we used to check if the record was destroyed here
+      return internalModel!.createSnapshot(options).serialize(options);
```
mike-north: leave a comment indicating why `!` is safe here

packages/store/addon/-private/system/ds-model-store.ts
```diff
@@ -0,0 +1,384 @@
...
+      currentState: internalModel.currentState,
+      container: null,
+    };
+    assign(createOptions, createRecordArgs);
+
+    if (setOwner) {
+      // ensure that `getOwner(this)` works inside a model instance
+      setOwner(createOptions, getOwner(this));
+    } else {
+      //createOptions.container = this.container;
```
mike-north: seems strange.

packages/store/addon/-private/system/ds-model-store.ts
```diff
@@ -0,0 +1,384 @@
...
+        attributes = Object.create(null);
+        attributeMap.forEach((meta, name) => (attributes[name] = meta));
+        this._attributesDefCache[modelName] = attributes;
+      }
+
+      return attributes;
+    }
+  }
+
+  _relationshipsDefinitionFor(modelName) {
```
mike-north: type this thing

packages/store/addon/-private/system/model/internal-model.ts
```diff
@@ -494,9 +514,10 @@ export default class InternalModel {
     let resolver = RSVP.defer<InternalModel>(promiseLabel);
 
     if (REQUEST_SERVICE) {
-      return this.store.scheduleSave(this, resolver, options);
+      // Casting to narrow due to the feature flag paths inside scheduleSave
+      return this.store.scheduleSave(this, resolver, options) as RSVP.Promise<void>;
     } else {
-      this.store.scheduleSave(this, resolver, options);
+      this.store.scheduleSave(this, resolver, options) as void;
```
mike-north: cast unnecessary unless you're consuming a return value

packages/store/addon/-private/system/model/internal-model.ts
```diff
@@ -1064,37 +1085,67 @@ export default class InternalModel {
 
   manyArrayRecordAdded(key) {
```
mike-north: type this

packages/store/addon/-private/system/model/shim-model-class.ts
```diff
@@ -0,0 +1,68 @@
...
+  }
+
+  get relationshipsByName(): Map<string, RelationshipSchema> {
+    let relationships = this.store._relationshipsDefinitionFor(this.modelName, this._id);
+    //TODO Talk to mnorth about needing a cast
+    return new Map(Object.entries(relationships as MakeConfident<RelationshipSchema>));
+    //return new Map(Object.entries(relationships) as [string, RelationshipSchema][]);
+  }
+
+  eachAttribute(callback: (key: string, attribute: AttributeSchema) => void, binding: unknown) {
```
mike-north: Might be overdoing it, but this is conceptually what the intent is 
```ts
eachAttribute<T>(callback: (this: T, key: string, attribute: AttributeSchema) => void, binding: T)
```

packages/store/addon/-private/system/relationships/state/relationship.ts
```diff
@@ -666,13 +667,23 @@ export default class Relationship {
...
-          // We know we are not an implicit relationship here
-          this.key as string
-        );
+        if (CUSTOM_MODEL_CLASS) {
+          storeWrapper.notifyBelongsToChange(
+            recordData.modelName,
+            recordData.id,
+            recordData.clientId,
+            // We know we are not an implicit relationship here
+            this.key as string
```
mike-north: try removing it. If you must keep it, change to the "lighter touch" `this.key!`

packages/serializer/addon/json-api.js
```diff
@@ -493,7 +493,7 @@ const JSONAPISerializer = JSONSerializer.extend({
...
```
runspired:
```suggestion
      let belongsToIsNotNew = belongsTo && !belongsTo.isNew;
```

packages/store/addon/-private/system/core-store.ts
```diff
@@ -341,6 +362,140 @@ class Store extends Service {
...
+      return record;
+    } else {
+      throw 'should not be here, custom model class ff error';
+    }
+  }
+
+  abstract instantiateRecord(
+    identifier: StableRecordIdentifier,
+    createRecordArgs: { [key: string]: any }, // args passed in to store.createRecord() and processed by recordData to be set on creation
+    recordDataFor: (identifier: StableRecordIdentifier) => RecordDataRecordWrapper,
```
runspired:
```suggestion
    recordDataFor: (identifier: RecordIdentifier) => RecordDataRecordWrapper,
```

packages/store/addon/-private/system/core-store.ts
runspired:
```suggestion
      // because a `Record` is provided there will always be a matching internalModel
      return internalModel!.createSnapshot(options).serialize(options);
```

alt: use lookup

packages/store/addon/-private/system/core-store.ts
```diff
@@ -2904,11 +2988,49 @@ class Store extends Service {
...
+      let internalModel = internalModelFactoryFor(this).peek(identifier.type, identifier.id, identifier.lid);
+      // TODO we used to check if the record was destroyed here
+      // Casting can be removed once REQUEST_SERVICE ff is turned on
+      return (internalModel!.save(options) as RSVP.Promise<void>).then(() => record);
+    } else {
+      throw 'saveRecord is only available when CUSTOM_MODEL_CLASS ff is on';
+    }
+  }
+
+  relationshipReferenceFor(identifier: RecordIdentifier, key: string): Reference {
```
runspired:
```suggestion
  relationshipReferenceFor(identifier: RecordIdentifier, key: string): BelongsToReference | HasManyReference {
```

packages/store/addon/-private/system/relationships/state/has-many.ts
```diff
@@ -18,7 +19,7 @@ export default class ManyRelationship extends Relationship {
   key: string;
   constructor(
     store: any,
-    inverseKey: string,
+    inverseKey: string | null,
```
runspired: this can never be null, only `Implicit` can be null which uses the base relationship class

packages/store/addon/-private/system/relationships/state/relationship.ts
```diff
@@ -368,7 +369,7 @@ export default class Relationship {
         relationship = relationships[this.inverseKeyForImplicit] = new Relationship(
           this.store,
           // we know we are not an implicit relationship here
-          this.key as string,
+          this.key,
```
runspired: we can remove this comment

packages/store/addon/-private/system/snapshot.js
```diff
@@ -97,6 +125,7 @@ export default class Snapshot {
 
     return attributes;
   }
+  */
```
runspired: we can delete this instead of commenting it out

packages/store/addon/-private/system/snapshot.js
```diff
@@ -356,9 +388,17 @@ export default class Snapshot {
     @param {Function} callback the callback to execute
     @param {Object} [binding] the value to which the callback's `this` should be bound
   */
+  /*
   eachAttribute(callback, binding) {
     this.record.eachAttribute(callback, binding);
   }
+  */
```
runspired: delete vs comment out

packages/store/addon/-private/system/snapshot.js
```diff
@@ -376,9 +416,17 @@ export default class Snapshot {
     @param {Function} callback the callback to execute
     @param {Object} [binding] the value to which the callback's `this` should be bound
   */
+  /*
   eachRelationship(callback, binding) {
     this.record.eachRelationship(callback, binding);
   }
+  */
```
runspired: delete vs comment out

packages/store/addon/-private/system/store/record-data-store-wrapper.ts
```diff
@@ -65,6 +64,7 @@ export default class RecordDataStoreWrapper implements IRecordDataStoreWrapper {
     backburner.join(() => {
       backburner.schedule('syncRelationships', this, this._flushPendingManyArrayUpdates);
     });
+    //}
```
runspired: forgotten, remove

packages/store/addon/-private/system/store/record-data-store-wrapper.ts
runspired: we looked into the existing behavior and existing behavior is to return false when unknowable

packages/store/addon/-private/ts-interfaces/schema-definition-service.ts
```diff
@@ -0,0 +1,14 @@
...
+*/
+import { RecordIdentifier } from './identifier';
+import { AttributesSchema, RelationshipsSchema } from './record-data-schemas';
+// we import the class not the interface because we expect
+// because we expect to use this internally with the more complete set
+// of APIs
+
+export interface SchemaDefinitionService {
+  doesTypeExist(modelName: string): boolean;
+  attributesDefinitionFor(identifier: RecordIdentifier | string): AttributesSchema;
```
runspired: minor nit, overloads would allow separate variable name and documentation for the modelName case

packages/store/addon/-private/system/model/internal-model.ts
```diff
@@ -1064,37 +1085,67 @@ export default class InternalModel {
...
     if (this.hasRecord) {
-      this._record.notifyPropertyChange(key);
+      if (CUSTOM_MODEL_CLASS) {
+        this.store._notificationManager.notify(this.identifier, 'property');
+      } else {
+        this._record.notifyPropertyChange(key);
+      }
       this.updateRecordArrays();
     }
     let manyArray = this._manyArrayCache[key] || this._retainedManyArrayCache[key];
```
runspired: is this dead code?

packages/-ember-data/tests/unit/model/relationships/has-many-test.js
```diff
@@ -2299,10 +2300,11 @@ module('unit/model/relationships - DS.hasMany', function(hooks) {
 
   test('DS.ManyArray is lazy', function(assert) {
     let peopleDidChange = 0;
+    let expectedNumberOfChanges = CUSTOM_MODEL_CLASS ? 1 : 2;
     const Tag = DS.Model.extend({
       name: DS.attr('string'),
       people: DS.hasMany('person'),
-      peopleDidChange: observer('people', function() {
```
runspired: we should be very careful with this one, possibly confirm with some apps powering their selects with hasMany's that this doesn't break them.

packages/-ember-data/tests/unit/custom-class-support/custom-class-model-test.ts
```diff
@@ -0,0 +1,410 @@
...
+      let person = store.createRecord('person', { name: 'chris' });
+      assert.equal(returnValue, person, 'createRecord returns the instantiated record');
+      assert.deepEqual(returnValue, person, 'record instantiating does not modify the returned value');
+    });
+
+    test('recordData lookup', function(assert) {
+      assert.expect(3);
+      let CreationStore = CustomStore.extend({
+        instantiateRecord(identifier, createRecordArgs, recordDataFor, notificationManager) {
+          let rd = recordDataFor(identifier);
```
runspired: we never assert this is what we expect

packages/store/addon/-private/system/store/record-data-store-wrapper.ts
igorT: https://github.com/emberjs/data/blob/master/packages/store/addon/-private/system/relationships/state/create.ts#L36

packages/store/addon/-private/system/model/shim-model-class.ts
igorT: :+1:

packages/store/addon/-private/ts-interfaces/record.ts
igorT: 👍

packages/store/addon/-private/system/model/shim-model-class.ts
igorT: this is pretty neat

packages/store/addon/-private/system/snapshot.js
```diff
@@ -141,6 +141,9 @@ export default class Snapshot {
   }
 
   get isNew() {
+    if (!CUSTOM_MODEL_CLASS) {
+      throw 'isNew is only available when custom model class ff is on';
```
runspired: use `new Error`

packages/store/addon/-private/system/ds-model-store.ts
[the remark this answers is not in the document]
igorT: I think we should probably throw a deprecation if you subclassed modelFor/factoryFor, and then extract in a future release, otherwise we'll break older versions of m3/other addons which subclassed those methods. Shouldn't be hard to extract later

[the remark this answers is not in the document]
runspired: The extract comment was on the notifyChanges method itself, not the class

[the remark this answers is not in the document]
igorT: Yup, the method itself. If we extracted it, we would need to put it on the ds.model and m3 model's wouldn't get notifyPropertyChange called on them from RD.

packages/store/addon/-private/system/snapshot-record-array.js
igorT: is class

igorT:
@rwjblue CI is green

PR #6316, opened 2019-08-10
[BUGFIX beta] Fix Rollup warning about unresolved dependency

HeroicEric:
Extracted from https://github.com/emberjs/data/pull/6098

Fixes the following warning:

```
'@ember-data/store' is imported by -private/system/model/model.js, but could not be resolved – treating it as an external dependency
```

I'm not sure I understand why but it seems like Rollup does not like the other style of import.

This also resolves an issue with the external partner tests for ember-observer. You can see an example of what happens without this at https://dev.azure.com/ember-data/emberjs/_build/results?buildId=962


fsmanuel:
@HeroicEric I'm interested in joining the external parter tests with [ember-local-storage](https://github.com/funkensturm/ember-local-storage). How can I apply?

PR #6321, opened 2019-08-13
[BUGFIX BETA CANARY] Add ember-decorators-polyfill dep to @ember-data/store

HeroicEric:
Extracted from https://github.com/emberjs/data/pull/6098

https://github.com/emberjs/data/pull/6249 refactored `Store` to a class. As part of this refactoring, `Store#defaultAdapter` was implemented using the `@computed` decorator. Using the `@computed` decorator requires Ember 3.10+.

This adds [`ember-decorators-polyfill`](https://github.com/pzuraq/ember-decorators-polyfill) as a dependency to `@ember-data/store` to add support for apps using an Ember version < 3.10.

---

Note: I don't believe this is needed for `release` because the conversion of `Store` to a class [was reverted](https://github.com/emberjs/data/commit/7dd0dd1d5540805b75eac276e838599ffc6f625c)

### DIFF
--- a/packages/store/package.json
+++ b/packages/store/package.json
@@ -26,6 +26,7 @@
     "ember-cli-babel": "^7.8.0",
     "ember-cli-path-utils": "^1.0.0",
     "ember-cli-typescript": "^2.0.2",
+    "ember-decorators-polyfill": "^1.0.6",
     "heimdalljs": "^0.3.0"
   },
   "devDependencies": {


packages/store/package.json
```diff
@@ -26,6 +26,7 @@
     "ember-cli-babel": "^7.8.0",
     "ember-cli-path-utils": "^1.0.0",
     "ember-cli-typescript": "^2.0.2",
+    "ember-decorators-polyfill": "^1.0.6",
```
runspired: I think we only need this in our ember-try config, I believe it would be on consuming apps to install the polyfill? Otherwise this lgtm.

cc @rwjblue

rwjblue: If that is true, then we have to revert #6249 (since it means that ember-data@3.13 is not compatible with its own current support policy, and requires that the host app bring the polyfill).

🤔

runspired: @rwjblue I’m more concerned with whether adding the polyfill would troll end consumers that bring their own version.

igorT: I agree with @rwjblue, but lets discuss today

runspired: fwiw it wouldn't mean we need to revert, only that we should change from using @computed to using a getter

runspired:
Replaced by #6331

PR #6329, opened 2019-08-15
[BUGFIX BETA CANARY] remove @computed usage in Store

runspired:


igorT:
I thought we decided to define it as a computed outside the class  rather
than a getter.

PR #6330, opened 2019-08-15
[BUGFIX] Fix crypto to `window.crypto`

igorT:
@HeroicEric discovered that storefront tests fail when accessing `crypto` without `window`. We are still looking into the underlying causes


runspired:
This needs backported to `beta` before we release `3.13` stable to resolve #6458 

@Gaurav0 can you grab this?

runspired:
cc @igorT

PR #6334, opened 2019-08-16
[BUGFIX BETA CANARY] usage of recordDataFor for client side create needs to be in correct cache

runspired:


igorT:
We should follow up with a test

PR #6338, opened 2019-08-16
[CHORE] Use model name instead of store in isPrimaryType

### LINKED ISSUE #5809: Rest Serializer `isPrimaryType` should not use `store.modelFor`
Currently Rest Serializer implicitly does type normalization in `store.isPrimaryType` by calling `store.modelFor`. It shouldn't do that, it should normalize the type name itself

ansmonjol:
Update rest serializer `isPrimaryType` to not use store anymore. 

Closes #5809

### DIFF
--- a/packages/serializer/addon/rest.js
+++ b/packages/serializer/addon/rest.js
@@ -273,7 +273,7 @@ const RESTSerializer = JSONSerializer.extend({
         continue;
       }
 
-      var isPrimary = !forcedSecondary && this.isPrimaryType(modelName, PrimaryModelClass);
+      var isPrimary = !forcedSecondary && this.isPrimaryType(modelName, primaryModelClass);
       var value = payload[prop];
 
       if (value === null) {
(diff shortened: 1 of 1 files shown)


packages/serializer/addon/rest.js
```diff
@@ -273,7 +273,7 @@ const RESTSerializer = JSONSerializer.extend({
         continue;
       }
 
-      var isPrimary = !forcedSecondary && this.isPrimaryType(store, typeName, primaryModelClass);
+      var isPrimary = !forcedSecondary && this.isPrimaryType(modelName, PrimaryModelClass);
```
runspired: if we capitalize the argument here we need to capitalize it elsewhere in this method as well (this is why the basic tests are failing)

ansmonjol: Of course haha 😅 my bad 

Will have a look at new failing ones

PR #6339, opened 2019-08-16
[FEAT] Adds support for parenless attr, belongsTo, and hasMany

pzuraq:
This will only be added in modern Ember versions that support decorators. I tried to figure out how to write better tests, but I couldn't find basic unit tests to emulate, will definitely take pointers if anyone has any!


chancancode:
Failures seem unrelated, the same tests seem to fail on master.

packages/-ember-data/tests/unit/model-test.js
```diff
@@ -1442,6 +1443,42 @@ module('unit/model - Model', function(hooks) {
...
+        cat = store.push({
+          data: {
+            type: 'mascot',
+            id: '1',
+            attributes: {
+              name: 'Argon',
+            },
+          },
+        });
+
```
runspired: can we check accessing via dot notation as well?

Issue #6167, opened 2019-06-18
JSONAPI adapter content-type is being blanked over by application/vnd.api+json on POST


Benjy1979:
### Reproduction

Please provide one of the following:

- a PR linking to this issue with a failing test showing the issue
A what?
- a [Twiddle](https://ember-twiddle.com/) with a simplified reproduction and instructions for how to
  observe the unexpected result.
Not familiar with it. It doesn't let me add an adapter and does not have the right version in the dropdown. 
- a github repository with a simplified reproduction and instructions for how to observe the unexpected result
 [Here is a repository](https://github.com/Benjy1979/ember-adapter-bug)

Check POST in the model adapters of CreateRecord call.

### Description

[See this for description ](https://github.com/emberjs/ember.js/issues/18112#issuecomment-503137074)

### Versions

Run the following command and paste the output below: `npm ls ember-source && npm ls ember-cli && npm ls ember-data`.

```cli
ember-roomviewer-bighomeshop@0.0.0 C:\inetpub\wwwroot\ember-roomviewer-bighomeshop
`-- ember-source@3.10.1

ember-roomviewer-bighomeshop@0.0.0 C:\inetpub\wwwroot\ember-roomviewer-bighomeshop
`-- ember-cli@3.10.1

ember-roomviewer-bighomeshop@0.0.0 C:\inetpub\wwwroot\ember-roomviewer-bighomeshop
`-- ember-data@3.10.0

```

*P.S. If any of the packages show more than one installed version, that may be the root cause of the issue!*

--------------------------------------------------------------

Thanks again!

The `ember-data` Team <3

Benjy1979:
I've put a project showing simple program to reproduce the issue:
 [Here is a repository](https://github.com/Benjy1979/ember-adapter-bug)

runspired:
@Benjy1979 if you properly case to `Content-Type` is it set?

Benjy1979:
No it stays the same.
`Accept: application/vnd.api+json
Accept-Encoding: gzip, deflate, br
Accept-Language: en-US,en;q=0.9
Cache-Control: no-cache
Connection: keep-alive
Content-Length: 66
Content-Type: application/vnd.api+json
Cookie: _ga=GA1.1.1842830690.1557486605; lc_sso10275727=1558697657640
Host: localhost:4200
Origin: http://localhost:4200
Pragma: no-cache
Referer: http://localhost:4200/
[...]

PR #6341, opened 2019-08-17
[BUGFIX]: set Content-Type for non GET requests only

snewcomer:
- [x] ensure fetch nor ajax sets content-type header if GET request
- [x] if existing Content-Type header, don't add another one
- [x] add tests

Piggybacked off of #6169 with some tests and default content type header

close #6341 #6167

### DIFF
--- a/packages/adapter/addon/json-api.js
+++ b/packages/adapter/addon/json-api.js
@@ -143,6 +143,8 @@ import { pluralize } from 'ember-inflector';
 const JSONAPIAdapter = RESTAdapter.extend({
   defaultSerializer: '-json-api',
 
+  defaultContentType: 'application/vnd.api+json',
+
   /**
     @method ajaxOptions
     @private
@@ -152,9 +154,8 @@ const JSONAPIAdapter = RESTAdapter.extend({
     @return {Object}
   */
   ajaxOptions(url, type, options = {}) {
-    options.contentType = options.contentType || 'application/vnd.api+json';
-
     let hash = this._super(url, type, options);
+
     hash.headers['Accept'] = hash.headers['Accept'] || 'application/vnd.api+json';
 
     return hash;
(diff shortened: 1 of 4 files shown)


packages/adapter/addon/json-api.js
```diff
@@ -143,6 +143,8 @@ import { pluralize } from 'ember-inflector';
 const JSONAPIAdapter = RESTAdapter.extend({
   defaultSerializer: '-json-api',
 
+  defaultContentType: 'application/vnd.api+json',
```
snewcomer: perhaps privatize with `_`?

packages/adapter/addon/rest.js
```diff
@@ -1093,14 +1095,19 @@ const RESTAdapter = Adapter.extend(BuildURLMixin, {
...
 
     if (get(this, 'useFetch')) {
+      if (options.data && options.type !== 'GET') {
+        if (!options.headers['Content-Type'] && !options.headers['content-type']) {
+          options.headers['content-type'] = contentType;
+        }
+      }
       options = fetchOptions(options, this);
     } else {
+      if (options.data && options.type !== 'GET') {
```
snewcomer: A GET request with a non-simple content-type will result in a preflight request.  One consequence of adding the `Content-Type` header on GET (where we used to not) is the backend server perhaps doesn't expect this and won't return the right a-c-a-o header.

packages/adapter/addon/json-api.js
runspired: yes please

packages/adapter/addon/rest.js
runspired: Can we leave this explanation of this conditional as a comment in the code?

brandynbennett:
@snewcomer Is this change part of any release?

runspired:
@brandynbennett it is in beta

brandynbennett:
Awesome thanks!

PR #6342, opened 2019-08-18
[INTERNAL]: replace -rest adapter usage in tests

snewcomer:
This PR is meant to eliminate `-rest` usage as described in #6166. 

- [x] drop `-rest` from usage in tests and instead register `application`

Follow up:
- drop `-rest` from `createStore` && `setupStore`
- if using RestAdapter, also register RestSerializer explicitly in test
- drop `-default` from `createStore` && `setupStore` and instead register `application`.  Don't set this adapter on store in these helpers
...

Ref https://github.com/emberjs/data/issues/6166

### DIFF
--- a/packages/-ember-data/tests/integration/store/adapter-for-test.js
+++ b/packages/-ember-data/tests/integration/store/adapter-for-test.js
@@ -154,7 +154,7 @@ module('integration/store - adapterFor', function(hooks) {
   });
 
   test('we can specify a fallback adapter by name in place of the application adapter', async function(assert) {
-    store.adapter = '-rest';
+    store.adapter = '-fallback';
     let { owner } = this;
 
     let didInstantiateRestAdapter = false;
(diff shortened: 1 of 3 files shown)


packages/-ember-data/tests/integration/store/adapter-for-test.js
```diff
@@ -223,38 +223,13 @@ module('integration/store - adapterFor', function(hooks) {
...
-    let restAdapter = store.adapterFor('-rest');
-    assert.ok(restAdapter instanceof RestAdapter, 'We found the correct adapter for -rest');
+    let restAdapter = store.adapterFor('-fallback');
+    assert.ok(restAdapter instanceof RestAdapter, 'We found the correct adapter for -fallback');
     assert.ok(!didInstantiateAppAdapter, 'We did not instantiate the application adapter again');
     assert.ok(didInstantiateRestAdapter, 'We instantiated the fallback adapter');
     assert.ok(restAdapter !== appAdapter, `We did not use the application adapter instance`);
   });
 
-  test('we can specify a fallback adapter by name in place of the application adapter', async function(assert) {
```
snewcomer: near duplicate of another test above

PR #6346, opened 2019-08-19
[BUGFIX release] Add @ember-data/canary-features dependency where used.

rwjblue:
@ember-data/-build-infra directly requires `@ember-data/canary-features` (in `packages/-build-infra/src/features.js`), but without a dependency on it directly we can't ensure that it is present.


igorT:
Release should be manually set to be the older version right?

PR #6347, opened 2019-08-19
[CHORE] Refactor away from `createStore` test helper

HeroicEric:
From https://github.com/emberjs/data/issues/6166

This refactors some tests so that they don't rely on the `createStore` test helper.

`createStore` was helpful in the past but is less so now that we have [nice APIs for dependency injection](https://guides.emberjs.com/release/applications/dependency-injection/).

---

One test was removed from `unit/store/adapter-interop - Store working with a Adapter` because I don't think it was correct.

Here was the test:

https://github.com/emberjs/data/blob/39182f324d5777e4a75c4a582f0791631e37d8de/packages/-ember-data/tests/unit/store/adapter-interop-test.js#L28-L32

This test was, at one time, testing that it was possible to pass a factory as the `adapter` property to `Store`. This feature was removed in https://github.com/emberjs/data/pull/3191 but the test continued passing because of the way the `createStore` works.

https://github.com/emberjs/data/blob/39182f324d5777e4a75c4a582f0791631e37d8de/packages/-ember-data/tests/helpers/store.js#L57-L60

### DIFF
--- a/packages/-ember-data/tests/unit/adapters/rest-adapter/group-records-for-find-many-test.js
+++ b/packages/-ember-data/tests/unit/adapters/rest-adapter/group-records-for-find-many-test.js
@@ -1,32 +1,33 @@
 import { run } from '@ember/runloop';
 import { Promise as EmberPromise } from 'rsvp';
-import { createStore } from 'dummy/tests/helpers/store';
+import { setupTest } from 'ember-qunit';
 
 import { module, test } from 'qunit';
 
-import DS from 'ember-data';
+import Model from '@ember-data/model';
+import RESTAdapter from '@ember-data/adapter/rest';
 
-let GroupsAdapter, store, requests;
+let store, requests;
 let maxLength;
 let lengths;
 
 module('unit/adapters/rest_adapter/group_records_for_find_many_test - DS.RESTAdapter#groupRecordsForFindMany', function(
   hooks
 ) {
+  setupTest(hooks);
+
   hooks.beforeEach(function() {
     maxLength = -1;
     requests = [];
     lengths = [];
 
-    GroupsAdapter = DS.RESTAdapter.extend({
+    const ApplicationAdapter = RESTAdapter.extend({
       coalesceFindRequests: true,
 
       findRecord(store, type, id, snapshot) {
         return { id };
       },
-    });
 
-    GroupsAdapter.reopen({
       ajax(url, type, options) {
         requests.push({
           url,
@@ -48,14 +49,10 @@ module('unit/adapters/rest_adapter/group_records_for_find_many_test - DS.RESTAda
       },
     });
 
-    store = createStore({
-      adapter: GroupsAdapter,
-      testRecord: DS.Model.extend(),
-    });
-  });
+    this.owner.register('adapter:application', ApplicationAdapter);
+    this.owner.register('model:test-record', Model.extend());
 
-  hooks.afterEach(function() {
-    run(store, 'destroy');
+    store = this.owner.lookup('service:store');
   });
 
   test('groupRecordsForFindMany - findMany', function(assert) {
--- a/packages/-ember-data/tests/unit/debug-test.js
+++ b/packages/-ember-data/tests/unit/debug-test.js
@@ -1,37 +1,33 @@
+import Model, { attr, belongsTo, hasMany } from '@ember-data/model';
 import { computed } from '@ember/object';
-import { createStore } from 'dummy/tests/helpers/store';
+import { setupTest } from 'ember-qunit';
 
 import { module, test } from 'qunit';
 
-import DS from 'ember-data';
+module('Debug', function(hooks) {
+  setupTest(hooks);
 
-const TestAdapter = DS.Adapter.extend();
-
-module('Debug', function() {
   test('_debugInfo groups the attributes and relationships correctly', function(assert) {
-    const MaritalStatus = DS.Model.extend({
-      name: DS.attr('string'),
+    const MaritalStatus = Model.extend({
+      name: attr('string'),
     });
 
-    const Post = DS.Model.extend({
-      title: DS.attr('string'),
+    const Post = Model.extend({
+      title: attr('string'),
     });
 
-    const User = DS.Model.extend({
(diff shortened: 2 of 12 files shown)


packages/-ember-data/tests/unit/store/adapter-interop-test.js
```diff
@@ -2,58 +2,50 @@ import { A } from '@ember/array';
...
-
-  hooks.afterEach(function() {
-    run(() => {
-      if (store) {
-        store.destroy();
-      }
-    });
-  });
-
-  test('Adapter can be set as a factory', function(assert) {
```
HeroicEric: This is the test that was removed

HeroicEric:
If this looks good I can do the same for `setupStore`

packages/-ember-data/tests/unit/debug-test.js
```diff
@@ -1,105 +1,113 @@
...
+      includeOtherProperties: true,
+      groups: [
+        {
+          name: 'Attributes',
+          properties: ['id', 'name', 'isDrugAddict'],
+          expand: true,
+        },
+        {
+          name: 'maritalStatus',
+          properties: ['maritalStatus'],
```
igorT: How do you feel about keeping things like 
```
const MaritalStatus = DS.Model.extend({
    name: DS.attr('string'),	   
  });	
```
so that searching for the name/defintion of `maritalStatus` will find results

HeroicEric: yea, I can change those. Initially I thought it was nice to save the vertical space but then I ran prettier and it is actually even longer the other way anyway

HeroicEric: @igorT i moved the model definitions back outside of the registrations

packages/-ember-data/tests/unit/adapters/rest-adapter/group-records-for-find-many-test.js
```diff
@@ -48,14 +49,10 @@ module('unit/adapters/rest_adapter/group_records_for_find_many_test - DS.RESTAda
       },
     });
 
-    store = createStore({
-      adapter: GroupsAdapter,
-      testRecord: DS.Model.extend(),
-    });
-  });
+    this.owner.register('adapter:application', ApplicationAdapter);
```
snewcomer: Are there any cases where we have to register the corresponding `RESTSerializer`? (previously the createStore was doing this)

packages/-ember-data/tests/unit/model/merge-test.js
```diff
@@ -238,7 +235,7 @@ module('unit/model/merge - Merging', function(hooks) {
     assert.equal(person.get('city'), 'Boston', 'the original data applies');
 
     run(() => {
```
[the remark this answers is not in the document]
HeroicEric: I'm not sure 🙃 There are a lot of tests involved with this (and the following `setupStore`) refactor so I didn't want to dig into anything extra for now

packages/-ember-data/tests/unit/adapters/rest-adapter/group-records-for-find-many-test.js
HeroicEric: The `createStore` helper was actually registering the `-default` serializer as the `JSONAPISerializer`:

https://github.com/emberjs/data/blob/7c7ebdfb3c0d586cb52f42e57815cbea665c9331/packages/-ember-data/tests/helpers/store.js#L70

There were a few cases where I needed to do this where the test involves mocking the response from the adapter. 
Some examples:

https://github.com/emberjs/data/pull/6347/files#diff-8aaaee520d117ac876f402e7a684ebd8R34
https://github.com/emberjs/data/pull/6347/files#diff-8aaaee520d117ac876f402e7a684ebd8R101
https://github.com/emberjs/data/pull/6347/files#diff-8aaaee520d117ac876f402e7a684ebd8R189

packages/-ember-data/tests/unit/model/merge-test.js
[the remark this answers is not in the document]
snewcomer: https://github.com/emberjs/data/pull/6221/files#diff-285316c98a233fb579604652813f294eR276

For future us, looks like we can.

packages/-ember-data/tests/unit/debug-test.js
```diff
@@ -1,108 +1,106 @@
...
+
+    const Post = Model.extend({
+      title: attr('string'),
+    });
+
+    const User = Model.extend({
+      name: attr('string'),
+      isDrugAddict: attr('boolean'),
+      maritalStatus: belongsTo('marital-status', { async: false }),
+      posts: computed(() => [1, 2, 3])
```
runspired: this got copied from the test below but should still be a normal hasMany relationship

packages/-ember-data/tests/unit/store/has-model-for-test.js
```diff
@@ -1,20 +1,17 @@
...
 
   test(`hasModelFor correctly normalizes`, function(assert) {
+    this.owner.register('model:one-foo', Model.extend({}));
+    this.owner.register('model:two-foo', Model.extend({}));
+
+    let store = this.owner.lookup('service:store');
+
     assert.equal(store._hasModelFor('oneFoo'), true);
-    assert.equal(store._hasModelFor('twoFoo').true);
+    assert.equal(store._hasModelFor('twoFoo'), true);
```
runspired: nice catch, would have been asserting `undefined === undefined` before

packages/-ember-data/tests/unit/store/unload-test.js
```diff
@@ -1,20 +1,24 @@
...
+let store, tryToFind;
 
 module('unit/store/unload - Store unloading records', function(hooks) {
+  setupTest(hooks);
+
   hooks.beforeEach(function() {
-    Record = DS.Model.extend({
-      title: DS.attr('string'),
-      wasFetched: DS.attr('boolean'),
+    let Record = Model.extend({
```
runspired: I would prefer const in this situation as it would be a mistake to reassign

PR #6349, opened 2019-08-19
Set group name to relationship.kind

rwwagner90:
This fixes a bug in inspector where the groups are named the model name instead of the relationship type.

### DIFF
--- a/packages/-ember-data/tests/unit/debug-test.js
+++ b/packages/-ember-data/tests/unit/debug-test.js
@@ -37,8 +37,11 @@ test('_debugInfo groups the attributes and relationships correctly', function(as
   let propertyInfo = record._debugInfo().propertyInfo;
 
   assert.equal(propertyInfo.groups.length, 4);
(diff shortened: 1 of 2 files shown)


igorT:
@rwwagner90 thanks for the fix? Any chance you could update a test to cover this please? We have some in https://github.com/emberjs/data/blob/master/packages/-ember-data/tests/integration/debug-adapter-test.js

rwwagner90:
@igorT I just updated the existing tests to check the correct relationship types as the names for the groups. There seem to be other failures, but I don't know if they are related to my change or not.

rwwagner90:
Thanks for merging! What release will this make it into?

igorT:
@rwwagner90 is this actively broken in the wild? We could backport as well and make sure it goes into beta

rwwagner90:
@igorT yeah, inspector has been showing the wrong names for the groups for quite awhile. Not sure when exactly it started.

PR #6350, opened 2019-08-20
[Chore] Rename Record interface to RecordInstance

igorT:


packages/store/addon/-private/ts-interfaces/record-instance.ts
```diff
@@ -14,4 +14,4 @@
 */
 
 // TODO Rename to RecordInstance
-export type Record = Object;
+export type RecordInstance = Object;
```
rwjblue: 😩 	`Instance` should never be in the name of a thing

```suggestion
export type IRecord = Object;
```

Using `IRecord` seems infinitely better (even if you hate the `I` prefixing semantic) 🤔

runspired: Isn't `IRecord` the controversial naming scheme used for something like "RecordInterface" ? Wouldn't that be more confusing?

igorT: @rwjblue in this case the problem we have is that `Record` is a inbuilt TS type that means roughly a dict. `IRecord` would mean RecordInterface which this is very much not. We could call it a Model, but usually we have tried to get Model to mean the class/schema like thing and not the actual object

rwjblue: The types here provide no insight into what `RecordInstance` actually is, can you give me a quick summary of the difference and hopefully the practical end-user concept that is represented by each:

* record
* record data
* model
* record instance

From my perspective, this is extremely unclear...

runspired: RecordInstance is a placeholder for a user provided value. As our type abilities expand it will become user defined when they extend the store to implement the instantiate record hook. Using this placeholder prepares us for enabling the user defined type to correctly flow through the internals such that other store methods which return records will return the correct type to the user

igorT: - record, and recordInstance-> object received from methods like peek, find etc. Used to mean an instance of DS.Model but that requirement is being lifted. 
- model: DS.Model class, class that exists in /models folder
- record data: not an end-user concept. Backing data structure for a record

rwjblue: > record, and recordInstance

But what is the difference between the two?

rwjblue: Also, this should really be `unknown` then, right? Because its supposed to be any user land thing (and they could use other structures than an `Object`).

igorT: > But what is the difference between the two?
We use `Record` in method names and everywhere else. However it's not great to use it as a type, `Record` is also a built Typescript type. That causes lots of confusion and also very trollish error messages, so we have to come up with a different name, herewith proposed `RecordInstance`. Open to other options though.

igorT: >Also, this should really be unknown then, right? Because its supposed to be any user land thing (and they could use other structures than an Object).

I think returning a primitive, null or undefined to represent a record seems like a bug and I am having trouble imagining the use case. Moreover we rely on sticking user defined records into weak maps, so them being records is also a practical limitation

rwjblue: I was specifically thinking of using a Symbol, which seems reasonable to me...

PR #6352, opened 2019-08-20
[DOC debug-adapter]: Add comments for each method in debug-adapter

SYU15:
This will be useful for debugging in the future when developers are stepping through both this `debug-adapter` and the `m3` `debug-adapter`.

### DIFF
--- a/packages/-ember-data/addon/-private/system/debug/debug-adapter.js
+++ b/packages/-ember-data/addon/-private/system/debug/debug-adapter.js
@@ -18,6 +18,15 @@ import Model from '@ember-data/model';
   @private
 */
 export default DataAdapter.extend({
+  /**
+    Specifies how records can be filtered based on the state of the record
+    Records returned will need to have a `filterValues`
+    property with a key for every name in the returned array
+    @public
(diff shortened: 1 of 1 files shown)


packages/-ember-data/addon/-private/system/debug/debug-adapter.js
```diff
@@ -26,10 +35,24 @@ export default DataAdapter.extend({
     ];
   },
 
+  /**
+    Detect whether a class is a DS.Model
+    @public
+    @method detect
+    @param {DS.Model} typeClass
+    @return boolean Whether the typeClass is a DS.Model class or not
```
igorT: should it be {boolean}?

packages/-ember-data/addon/-private/system/debug/debug-adapter.js
```diff
@@ -101,6 +166,14 @@ export default DataAdapter.extend({
     };
   },
 
+  /**
+    Each record can have a color that represents its state
```
igorT: maybe as a `Returns a color...`

PR #6353, opened 2019-08-20
[CHORE tests] simplify setupStore registrations

runspired:
This PR allows for more incremental migration away from `createStore` and `setupStore` to modern test syntax.

Specifically it

- ensures that we aren't registering anything unneeded for tests currently
- provides a path for registering a serializer for use with the test
- encourages use of "application" adapter/serializer

cc @pete-the-pete @snewcomer

### DIFF
--- a/packages/-ember-data/addon/setup-container.js
+++ b/packages/-ember-data/addon/setup-container.js
@@ -31,7 +31,6 @@ function initializeStore(registry) {
   registerOptionsForType.call(registry, 'adapter', { singleton: false });
   registry.register('serializer:-default', JSONSerializer);
   registry.register('serializer:-rest', RESTSerializer);
-  registry.register('adapter:-rest', RESTAdapter);
 
   registry.register('adapter:-json-api', JSONAPIAdapter);
   registry.register('serializer:-json-api', JSONAPISerializer);
--- a/packages/-ember-data/addon/setup-container.js
+++ b/packages/-ember-data/addon/setup-container.js
@@ -4,7 +4,6 @@ import JSONAPISerializer from '@ember-data/serializer/json-api';
 import JSONSerializer from '@ember-data/serializer/json';
 import RESTSerializer from '@ember-data/serializer/rest';
 import JSONAPIAdapter from '@ember-data/adapter/json-api';
-import RESTAdapter from '@ember-data/adapter/rest';
 
 import { BooleanTransform, DateTransform, NumberTransform, StringTransform } from '@ember-data/serializer/-private';
 
--- a/packages/-ember-data/tests/helpers/store.js
+++ b/packages/-ember-data/tests/helpers/store.js
@@ -3,11 +3,9 @@ import { setResolver } from '@ember/test-helpers';
 import EmberObject from '@ember/object';
 import Ember from 'ember';
 import Store from 'ember-data/store';
-import JSONAPIAdapter from '@ember-data/adapter/json-api';
 import Adapter from '@ember-data/adapter';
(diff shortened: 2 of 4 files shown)


runspired:
cc @HeroicEric

packages/-ember-data/addon/setup-container.js
```diff
@@ -31,7 +30,6 @@ function initializeStore(registry) {
   registerOptionsForType.call(registry, 'adapter', { singleton: false });
   registry.register('serializer:-default', JSONSerializer);
   registry.register('serializer:-rest', RESTSerializer);
-  registry.register('adapter:-rest', RESTAdapter);
```
snewcomer: Out of curiosity - this isn't a "breaking" change, right?  (still acclimating)

packages/-ember-data/tests/unit/store/adapter-interop-test.js
```diff
@@ -32,9 +32,12 @@ module('unit/store/adapter-interop - DS.Store working with a DS.Adapter', functi
   });
 
   test('Adapter can be set as a name', function(assert) {
-    store = createStore({ adapter: 'application' });
+    const env = setupStore({ adapter: 'custom-adapter' });
```
snewcomer: I'm sure you already saw this in #6347, but this test was deleted in that PR.

packages/-ember-data/tests/helpers/store.js
```diff
@@ -67,23 +73,31 @@ export default function setupStore(options) {
...
+    // Many tests rely on falling back to this serializer
+    // they should refactor to register this as the application serializer
+    owner.register('serializer:-default', JSONAPISerializer);
+
+    // RESTAdapter specifies a defaultSerializer of -rest
+    // Tests using this should refactor to register this as the application serializer
+    owner.register('serializer:-rest', RESTSerializer);
+
+    env.restSerializer = store.serializerFor('-rest');
+    env.serializer = store.serializerFor('-default');
```
snewcomer: 👍

packages/-ember-data/addon/setup-container.js
runspired: Correct. The existence of this adapter was for tests and private. To use it folks would have had to discover it and looked it up via -rest manually. Given the easy of importing RESTAdapter and no evidence of usage I don’t believe we need to treat this as intimate.

packages/-ember-data/tests/unit/store/adapter-interop-test.js
runspired: The removed test was the version of this that set it to an instance, not the one that set it to a string. It does refactor it to not use setupStore/createStore though :)

packages/-ember-data/tests/unit/store/adapter-interop-test.js
```diff
@@ -32,9 +32,12 @@ module('unit/store/adapter-interop - DS.Store working with a DS.Adapter', functi
...
+    const { store, owner } = env;
+    const CustomAdapter = DS.Adapter.extend({ isCustomAdapter: true });
```
igorT: do we need `isCustomAdapter`, makes it look as a special prop

packages/-ember-data/tests/unit/store/adapter-interop-test.js
igorT: :+1: on not naming it application, that seemed trolly

packages/-ember-data/tests/unit/store/adapter-interop-test.js
runspired: we don't, just trying to make it clear it's custom :D

PR #6354, opened 2019-08-20
Revert "[CHORE] Use model name instead of store in isPrimaryType"

runspired:
Reverts emberjs/data#6338


runspired:
This PR has failing tests still that weren't correctly reflecting in the Github UI when I merged it. Strangely the tests pass when run in a normal browser but fail in a headless browser. Will need some investigation.

runspired:
cc @ansmonjol for re-open + fix

PR #6356, opened 2019-08-21
[CHORE tests] Convert non-nested QUnit modules to nested

HeroicEric:
Converts the remaining QUnit modules that are using the non-nested syntax to the nested syntax.

This will help reduce noise in diffs for further work to modernize the test suite to use things like the `setupTest` helper from `ember-qunit`.

**Highly recommend hiding whitespace changes while reviewing**


HeroicEric:
cc @runspired

PR #6370, opened 2019-08-23
 [CHORE tests] eliminates remaining usage of -json-api serializer fallback in tests

runspired:
cc @HeroicEric @snewcomer 

look at the (reverted) "DELETE ME" commit to see how I ensured no fallback usage for doing similar when removing the `-default` serializer fallback usage in tests.

also note the deletion from the test-helper, this is also required at the same time to make sure we aren't registering the fallbacks anywhere

### DIFF
--- a/packages/-ember-data/tests/integration/relationships/belongs-to-test.js
+++ b/packages/-ember-data/tests/integration/relationships/belongs-to-test.js
@@ -242,9 +242,6 @@ module('integration/relationship/belongs_to Belongs-To Relationships', function(
       section: Section,
     });
 
-    env.registry.optionsForType('serializer', { singleton: false });
-    env.registry.optionsForType('adapter', { singleton: false });
-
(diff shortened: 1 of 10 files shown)


packages/-ember-data/tests/integration/relationships/belongs-to-test.js
```diff
@@ -242,9 +242,6 @@ module('integration/relationship/belongs_to Belongs-To Relationships', function(
       section: Section,
     });
 
-    env.registry.optionsForType('serializer', { singleton: false });
-    env.registry.optionsForType('adapter', { singleton: false });
```
HeroicEric: What was this actually doing?

runspired: nothing, it was redundant as the helper already did exactly this (And did it twice lol)

PR #6374, opened 2019-08-24
[CHORE tests] Replace remaining setupStore for unit tests

HeroicEric:
Part of https://github.com/emberjs/data/issues/6166

This replaces the remaining usage of the `setupStore` helper in our unit tests (`tests/unit/*`).

The one exception is https://github.com/emberjs/data/blob/a68bfacbbd473657fe293b86f066ad0bf26e2393/packages/-ember-data/tests/unit/store/model-for-test.js#L30, which uses the `replaceContainerNormalize` that comes from the `setupStore` helper. I'm not sure exactly what this is testing, if is still necessary, or how to replicate it.

cc @runspired


HeroicEric:
This does introduce one new occurrence of the deprecation about falling back to the `-json-api` serializer. This is because we have a test that is specifically testing that case

https://github.com/emberjs/data/blob/a68bfacbbd473657fe293b86f066ad0bf26e2393/packages/-ember-data/tests/unit/store/serializer-for-test.js#L49-L54

Maybe we need to remove that test?

PR #6375, opened 2019-08-25
[CHORE tests] Remove explicit -default serializer registration in tests

snewcomer:
ref #6166


snewcomer:
Now I see https://github.com/emberjs/data/pull/6374/files.  Looks like I'll revert the unit tests and wait for that to go in!

PR #6384, opened 2019-08-27
chore(eslint): Add override for bin files

dcyriller:
Add an override for bin-specific lint rules in .eslintrc config file.

Sorry for the whitespaces noise, but it's probably better to trim them down.


dcyriller:
Do you know why external partener tests travis is failing?

Issue #6385, opened 2019-08-27
3.11 tests fail against Ember's beta channel


dcyriller:
Backporting a bugfix to 3.11, we found out that this version's CI is failing against Ember's beta channel.

It fails with the following error message:
```
Build Error (broccoli-persistent-filter:Babel > [Babel: ember-data]) in ember-data/-debug/index.js

Cannot use the decorators and decorators-legacy plugin together
```

## Links

Full stack trace here: https://dev.azure.com/ember-data/emberjs/_build/results?buildId=1251

PR #6387, opened 2019-08-28
[BUGFIX 3.11] Upgrade babel-plugin-filter-imports

dcyriller:
To ease (and optimize) the upgrade, I've first deduplicated the dependencies with `npx yarn-deduplicate yarn.lock`.

This could close https://github.com/emberjs/data/issues/6385

EDIT: I wonder why 2 tests are failing (against Ember beta). But their failure has been hidden for quite a time. So it might worth it to merge that and fix them in a follow-up.


dcyriller:
Two tests fail, but at least the build succeeds.

dcyriller:
Oops closed by mistake

PR #6388, opened 2019-08-28
[DOC] Correct `RESTAdapter` namespace documentation

btecu:
Certain methods on the `RESTAdapter` are namespaced improperly in the documentation:
https://api.emberjs.com/ember-data/3.12/classes/RESTAdapter/methods/urlForFindMany?anchor=urlForFindRecord

### DIFF
--- a/packages/adapter/addon/-private/build-url-mixin.js
+++ b/packages/adapter/addon/-private/build-url-mixin.js
@@ -125,9 +125,9 @@ export default Mixin.create({
    Example:
 
    ```app/adapters/user.js
-   import JSONAPIAdapter from '@ember-data/adapter/json-api';
+   import RESTAdapter from '@ember-data/adapter/rest';
 
-   export default JSONAPIAdapter.extend({
+   export default RESTAdapter.extend({
      urlForFindRecord(id, modelName, snapshot) {
        let baseUrl = this.buildURL(modelName, id, snapshot);
        return `${baseUrl}/users/${snapshot.adapterOptions.user_id}/playlists/${id}`;
@@ -152,9 +152,9 @@ export default Mixin.create({
    Example:
 
    ```app/adapters/comment.js
-   import JSONAPIAdapter from '@ember-data/adapter/json-api';
+   import RESTAdapter from '@ember-data/adapter/rest';
 
-   export default JSONAPIAdapter.extend({
+   export default RESTAdapter.extend({
      urlForFindAll(modelName, snapshot) {
        return 'data/comments.json';
      }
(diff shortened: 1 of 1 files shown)


btecu:
Is not a huge deal, I can close this if there's no easy way to fix it.

runspired:
I think the best solution here is to overwrite the docs within the JSONAPIAdapter

btecu:
Is there an example of what that would look like?

runspired:
@btecu sure. The way docs work, while they live with the code they document they aren't actually coupled to it in anyway. You can have doc comments for methods, classes, and even full packages that don't actually exist in your code.

In our case, we want to take advantage of that to alter the docs for inherited methods. So for example:

```js
import RESTAdapter from './rest';

/**
 @module @ember-data/adapter
 @class JSONAPIAdapter
*/
export default class JSONAPIAdapter extends RESTAdapter {}
```

`@module` here tells us this documentation is part of the `@ember-data/adapter` package.
`@class` here tells us this documentation belongs to the `JSONAPIAdapter` class.

Until another `@module` or `@class` is discovered in the file, the documentation that follows will be appended to this class.

So for instance


```js
import RESTAdapter from './rest';

/**
 @module @ember-data/adapter
 @class JSONAPIAdapter
*/
export default class JSONAPIAdapter extends RESTAdapter {
  /**
    @property namespace
    @type {String} the namespace for API requests issued by this adapter
  */
  /**
    @method createRecord
    @param {Store} store the store instance asking for this record to be created
    @param {ModelSchema} schema schema information for the primaryType for this request
    @param {Snapshot} snapshot a snapshot of locally available information to save as part of this request
  */
   /**
    @method updateRecord
    @param {Store} store the store instance asking for this record to be created
    @param {ModelSchema} schema schema information for the primaryType for this request
    @param {Snapshot} snapshot a snapshot of locally available information to save as part of this request
  */
  async updateRecord(store, modelSchema, snapshot) {
    const response = await fetch('./user/1', { method: 'PATCH', data: {} });
    return response.toJSON();
  }
}
```

note how above I added docs for a property and a two methods, but only defined one of the methods.  Our API docs will include all of these for the class. So in this way we can easily "fill in" the API docs for anything inherited without actually needing to implement any methods :)

Hope this helps!

MelSumner:
@btecu @runspired what needs to happen to move this PR forward? Thank you for your work!

runspired:
@MelSumner @locks the docs need to be copied forward to RESTAdapter and their examples updated to show rest-ish examples. Probably best done as a new PR at this point.

Issue #6380, opened 2019-08-26
ember-data@3.12 does not work in IE11


btecu:
Steps to reproduce:
```
ember new test-ie
ember build --prod
```

When trying to browse in IE, the page will stay blank and throw:
```
SCRIPT5009: 'Symbol' is undefined
vendor.js (85124,5)
```

raido:
I am targeting IE11 (including Browserstack running tests with IE11). The `const` is not a that much of a problem for me. The `Symbol` issue is, so perhaps let's keep this issue focused on that.

raido:
https://github.com/emberjs/data/blob/63e13baf74ff97a3e08b6d7b1541a64695935316/packages/store/addon/-private/ts-interfaces/utils/brand.ts - basically this is the source of the issue I assume?

raido:
It seems my testing with v3.11 was actually testing v3.12 anyway - https://github.com/emberjs/data/issues/6378 - because of floating versions.

dcyriller:
I'll dig into it

btecu:
@runspired can we please get a `3.12.1` release with this fix?

runspired:
@btecu sure, but I believe there is one symbol still not being stripped.

dcyriller:
The last Symbol to fix is present on master but not in v3.12. I still have to do the fix with WeakMap to solve the whole Symbol issue. I could port the current fix to 3.12 if you think it makes sense.

runspired:
@dcyriller lets port the current fix so that we can fix 3.12 then get back to the WeakMap fix

runspired:
3.12.1 with this fix has been released

btecu:
Unfortunately, `ember-data` `3.12.1` still doesn't work with IE11.
I now get a different error:
![image](https://user-images.githubusercontent.com/7517745/64727610-38423200-d4a7-11e9-8d46-c66c3f3b457a.png)

I saw there's an issue #6326 about this, not specific to IE.
I do have `"jquery-integration": true`.

PR #6389, opened 2019-08-28
fix: Strip Symbol from production builds

dcyriller:
To ensure compatibility with IE11.

This should close https://github.com/emberjs/data/issues/6380.

This implementation is not non-enumerable (as opposed to Symbol). But making it enumerable would come with a cost. And it will only be required for [`identifiers`-related Symbols](https://github.com/emberjs/data/blob/69542df2df7cd0f57404f61f8d7ef0d2e3bae280/packages/store/addon/-private/ts-interfaces/identifier.ts#L4-L8).

These remaining Symbols are still unpublished or behind a feature flag (not sure but they are not part of the build outputs). They are not included in this PR. As they'll require a more sophisticated approach to be specifically non-enumerable. As we said, I'll leave it to you @runspired.

### DIFF
--- a/packages/store/addon/-private/system/fetch-manager.ts
+++ b/packages/store/addon/-private/system/fetch-manager.ts
@@ -15,6 +15,7 @@ import RequestCache from './request-cache';
 import { CollectionResourceDocument, SingleResourceDocument } from '../ts-interfaces/ember-data-json-api';
 import { RecordIdentifier } from '../ts-interfaces/identifier';
 import { FindRecordQuery, SaveRecordMutation, Request } from '../ts-interfaces/fetch-manager';
+import { symbol } from '../ts-interfaces/utils/symbol';
 import Store from './ds-model-store';
 import recordDataFor from './record-data-for';
 import CoreStore from './core-store';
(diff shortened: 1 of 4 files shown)


dcyriller:
I’ll update symbol implementation with WeakMap very soon (incl. for identifiers related Symbols).

dcyriller:
I've digged a bit and couldn't figure out how the `WeakMap` thing would look like. Could you give me another clue?

rwjblue:
@dcyriller 	in general, the idea would be that instead of:

https://github.com/emberjs/data/blob/63e13baf74ff97a3e08b6d7b1541a64695935316/packages/store/addon/-private/ts-interfaces/identifier.ts#L4

https://github.com/emberjs/data/blob/f472f2776076e31f7463f7b9065babc4f28a62fb/packages/store/addon/-private/identifiers/cache.ts#L418-L423

https://github.com/emberjs/data/blob/63e13baf74ff97a3e08b6d7b1541a64695935316/packages/store/addon/-private/identifiers/is-stable-identifier.ts#L7-L9

You would do something akin to:

```js
const IDENTIFIERS = new WeakMap();

export function isStableIdentifier(identifier) {
  return IDENTIFIERS.has(identifier);
}

export function markStableIdentifier(identifier) {
  IDENTIFIERS.set(identifier);
}
```

PR #6396, opened 2019-08-30
[DOC] Added documentation URL to evented deprecation warnings

yohanmishkin:
In #6059 deprecation warnings for Evented usage we added. This PR adds a URL to the deprecation documentation for the console warnings.

**before**

image

**after**

image

### DIFF
--- a/packages/-ember-data/tests/unit/model-test.js
+++ b/packages/-ember-data/tests/unit/model-test.js
@@ -608,7 +608,9 @@ module('unit/model - Model', function(hooks) {
 
         record.on('event!', F);
 
-        assert.expectDeprecation(/Called event! on person/);
+        if (DEBUG) {
+          assert.expectDeprecation(/Called event! on person/);
+        }
 
         record.trigger('event!');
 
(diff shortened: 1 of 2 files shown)


rwjblue:
@yohanmishkin - Mind rebasing? We've fixed some bugs on master that should fix CI...

packages/-ember-data/tests/unit/model-test.js
```diff
@@ -607,11 +607,15 @@ module('unit/model - Model', function(hooks) {
         let record = store.createRecord('person');
 
         record.on('event!', F);
+
+        assert.expectDeprecation(/Called event! on person/);
```
runspired: this will need to be wrapped in an `if (DEBUG)`

packages/serializer/addon/rest.js
```diff
@@ -286,8 +286,10 @@ const RESTSerializer = JSONSerializer.extend({
           'The adapter returned an array for the primary data of a `queryRecord` response. This is deprecated as `queryRecord` should return a single record.';
 
         deprecate(message, !isQueryRecordAnArray, {
-          id: 'ds.serializer.rest.queryRecord-array-response',
+          id: 'ember-data:store-queryrecord-array-response-with-restserializer',
```
runspired: dont change the ID

PR #6397, opened 2019-08-30
[DOC] Added URL to `queryRecord` deprecation warning

yohanmishkin:
Maybe this deprecation should be removed altogether (and throw some kind of error) since we've passed 3.0?

### DIFF
--- a/packages/serializer/addon/rest.js
+++ b/packages/serializer/addon/rest.js
@@ -286,7 +286,7 @@ const RESTSerializer = JSONSerializer.extend({
           'The adapter returned an array for the primary data of a `queryRecord` response. This is deprecated as `queryRecord` should return a single record.';
 
         deprecate(message, !isQueryRecordAnArray, {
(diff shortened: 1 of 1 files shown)


packages/serializer/addon/rest.js
```diff
@@ -286,8 +286,10 @@ const RESTSerializer = JSONSerializer.extend({
           'The adapter returned an array for the primary data of a `queryRecord` response. This is deprecated as `queryRecord` should return a single record.';
 
         deprecate(message, !isQueryRecordAnArray, {
-          id: 'ds.serializer.rest.queryRecord-array-response',
```
runspired: I have concerns about changing the ID of the deprecation post-release. Also I think I agree that we should just change this to an assert at this point. Thoughts @rwjblue ?

rwjblue: Ya, agree.

yohanmishkin: Updated.

Issue #6407, opened 2019-09-04
[BETA] Shows warning during build


Gaurav0:
When building an application or addon with "ember-data": "beta" included, you see the following warning:

 '@ember-data/store' is imported by -private/system/model/model.js, but could not be resolved – treating it as an external dependency

sly7-7:
I think this is resolved in https://github.com/emberjs/data/pull/6412

PR #6408, opened 2019-09-04
[BUGFIX BETA] Fix warning on build

Gaurav0:
For #6407 

Fixes "@ember-data/store' is imported by -private/system/model/model.js, but could not be resolved – treating it as an external dependency" warning on building application or addon including ember-data beta as a dependency.

Note: This is not an issue on master.


sly7-7:
I think this is already done via #6412

PR #6410, opened 2019-09-05
Remove module-unification blueprints

dcyriller:
This experiement has been disabled in Ember, meaning it can't be used anymore (starting from `ember-source@v3.12.0-beta.1`).


rwjblue:
FYI - ember-cli is working towards removal as well (see https://github.com/ember-cli/ember-cli/pull/8820)

rwjblue:
Looks like the CI failures are legit (probably blueprint tests?), you should be able to repro by running `lerna run test:node`.

dcyriller:
Thank you for the tip. I missed a few tests that need to be updated.

dcyriller:
It should be good now! Sorry for the run aounrd @rwjblue

Issue #6017, opened 2019-04-11
Add tests to check for breaking changes to the API docs


jenweber:
### Description

The ember.js repo has some tests that check to see if a PR's changes cause additions or removals of API documentation. We should add those same tests to ember-data too.

The tests help catch mistakes like when someone moves a function to its own file, but forgets to add the necessary YUIdoc code comments that make it get included in the API docs.

Here is a link to the PR that did the initial test implementation: https://github.com/emberjs/ember.js/pull/16910

To-do:
- [x] bring the tests over into ember-data
- [x] update/improve the test failure messages (if needed)
- [x] Add a note to CONTRIBUTING.md about how to resolve the test failure. For example, the tests fail on changes, such as _adding_ a new feature, in which case someone would just increment the number of API docs entries.

Part of work for https://github.com/emberjs/rfc-tracking/issues/11

Gaurav0:
First checkmark is complete.

PR #6418, opened 2019-09-10
[TEST] [DOC] Test to ensure no docs are missing.

Gaurav0:
For #6017 

expected,js was generated from the API docs for Ember Data 3.10


packages/-ember-data/package.json
```diff
@@ -4,18 +4,20 @@
...
   "directories": {
-    "doc": "doc",
+    "doc": "docs",
     "test": "tests"
   },
   "scripts": {
     "build": "ember build",
     "build:production": "ember build --environment=production",
     "start": "ember server",
+    "docs": "ember ember-cli-yuidoc",
```
runspired: what provides this command?

runspired: I see we already have it installed

Gaurav0:
It was ember-cli-yuidocs. Already installed.

PR #6419, opened 2019-09-10
[BUGFIX BETA RELEASE] ensure commit based versioning passes semver

runspired:
At some point recently `npm` installs began semver validating package versions. The versions we were producing for builds for individual commits failed validation because we required a `-` or ` +` before the sha.

Based on the precedence rules outlined here: https://github.com/semver/semver/blob/master/semver.md

I've updated our versioning from `x.x.x.<SHA>` to `x.x.x+sha.<SHA>`

for beta/canary branches the `-beta` and `-alpha` pre-release identifier will already be present

### DIFF
--- a/bin/-tarball-info.js
+++ b/bin/-tarball-info.js
@@ -136,7 +136,7 @@ packages.forEach(localName => {
   } catch (e) {
     return;
   }
-  const version = `${pkgInfo.version}.${CurrentSha}`;
+  const version = `${pkgInfo.version}-sha.${CurrentSha}`;
   const tarballName = `${convertPackageNameToTarballName(pkgInfo.name)}-${version}.tgz`;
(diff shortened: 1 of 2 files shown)


bin/-tarball-info.js
```diff
@@ -166,7 +166,7 @@ function generatePackageReference(version, tarballName) {
 function insertTarballsToPackageJson(fileLocation, options = {}) {
   const pkgInfo = require(fileLocation);
   if (options.isRelativeTarball) {
-    pkgInfo.version = `${pkgInfo.version}.${CurrentSha}`;
+    // pkgInfo.version = `${pkgInfo.version}`;
```
rwjblue: This seems odd?

runspired: checking but I think we were unnecessarily re-appending here

PR #6428, opened 2019-09-11
[BUGFIX canary]: Remove the last Symbols

dcyriller:
### Description

This is an attempt at removing the last `Symbol`s from master. The goal is to fix IE compatibility regarding Symbol support, as detailed in this issue: https://github.com/emberjs/data/issues/6380. 

It won't have to be backported. Indeed, the Identifier feature has been turned on in https://github.com/emberjs/data/pull/6366 and included in a tag for the first time in `v3.14.0-alpha.2`.

Thank you for pointing me to that solution @rwjblue (https://github.com/emberjs/data/pull/6389).

### Concerns

With the implementation I've done, I'm concerned by the fact:
- it kind of changes the `StableIdentifier` interface
- the `isStableIdentifier` signature is modified

But as it is still on canary, it might be fine?

Is the `symbol` fix good enough? It has been suggested it could be replaced by the WeakMap fix every time `symbol` is used.

### Tests

I've run the tests with the command: `yarn test`. Identifier feature flag being turned on, the test seemed to be fine. I'd be happy to add more if needed.

### DIFF
--- a/packages/-ember-data/tests/integration/identifiers/cache-test.ts
+++ b/packages/-ember-data/tests/integration/identifiers/cache-test.ts
@@ -1,0 +1,55 @@
+import { module, test } from 'qunit';
+import { setupTest } from 'ember-qunit';
+import { IDENTIFIERS } from '@ember-data/canary-features';
+import Store from '@ember-data/store';
+import { identifierCacheFor } from '@ember-data/store/-private';
+
+if (IDENTIFIERS) {
+  module('Integration | Identifiers - cache', function(hooks) {
+    setupTest(hooks);
+    let store, cache;
+
+    hooks.beforeEach(function() {
+      this.owner.register(`service:store`, Store);
+      store = this.owner.lookup('service:store');
+      cache = identifierCacheFor(store);
+    });
+
+    module('getOrCreateRecordIdentifier()', function() {
+      test('creates a new resource identifier if forgetRecordIdentifier() has been called on the existing identifier', async function(assert) {
+        const runspiredHash = {
+          type: 'person',
+          id: '1',
+          attributes: {
+            name: 'runspired',
+          },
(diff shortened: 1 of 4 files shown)


dcyriller:
External Partner Travis is failing in Azure, but not in Travis.

packages/store/addon/-private/ts-interfaces/identifier.ts
```diff
@@ -1,11 +1,13 @@
 /**
   @module @ember-data/store
 */
-export const IS_IDENTIFIER = Symbol('is-identifier');
+import { symbol } from '../ts-interfaces/utils/symbol';
+
+export const IDENTIFIERS = new WeakMap();
```
rwjblue: Nothing in here uses the WeakMap, can you move the WeakMap itself into the location of the utility methods that use it and not export it? (either moving the weakmap there, or moving those helper methods here)

dcyriller: Makes sense! Thanks, it's done.

rwjblue:
Ya, if there are for sure places where we know we can remove, we totally should.

packages/-ember-data/tests/integration/identifiers/cache-test.ts
```diff
@@ -0,0 +1,45 @@
...
+      } catch (e) {
+        assert.equal(
+          e.message,
+          'The lid generated for the new record is not unique as it matches an existing identifier'
+        );
+      }
+
+      cache.forgetRecordIdentifier(identifier);
+
+      const regeneratedIdentifier = cache.createIdentifierForNewRecord(houseHash);
```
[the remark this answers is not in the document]
dcyriller: done, I ended up splitting the test into two tests

PR #6430, opened 2019-09-11
[BUGFIX BETA RELEASE] make fetch function JIT

### LINKED ISSUE #6326: cannot find the `fetch` module or the `fetch` global when using with fastboot
Howdy folks, it's me again 👋 

This issue is essentially the same problem as https://github.com/emberjs/data/issues/5998 but with a different error message so I thought I would create a new issue.

Essentially adding ember-cli-fastboot to a newly generated Ember app doesn't work and here is the error we get this time: 

```
Error: cannot find the `fetch` module or the `fetch` global. Did you mean to install the `ember-fetch` addon?
```

### Reproduction

Steps to reproduce:

- npm init ember-app test-app (npm init ember-app always makes sure it is up-to-date)
- cd test-app
- ember install ember-cli-fastboot
- npm start
- visit http://localhost:4200
- 💥

Here is a simple reproduction that has the issue https://github.com/mansona/ember-data-fetch-test but as you can see from the steps it's pretty easy to reproduce

### Versions

```cli
test-app@0.0.0 /Users/mansona/temp/test-app
└── ember-source@3.11.1

test-app@0.0.0 /Users/mansona/temp/test-app
└── ember-cli@3.11.0

test-app@0.0.0 /Users/mansona/temp/test-app
└── ember-data@3.11.4
```

/cc @rwjblue @runspired as we were looking at this together last time

### LINKED ISSUE #6380: ember-data@3.12 does not work in IE11

runspired:
resolves #6326 
resolves #6380

### DIFF
--- a/packages/adapter/addon/-private/utils/fetch.ts
+++ b/packages/adapter/addon/-private/utils/fetch.ts
@@ -2,23 +2,24 @@ import require, { has } from 'require';
 
 type FetchFunction = (input: RequestInfo, init?: RequestInit | undefined) => Promise<Response>;
 
-let _fetch: FetchFunction | null = null;
+let _fetch: () => FetchFunction | null = null;
 
 export default function getFetchFunction(): FetchFunction {
   if (_fetch !== null) {
-    return _fetch;
+    return _fetch();
   }
 
   if (has('fetch')) {
     // use `fetch` module by default, this is commonly provided by ember-fetch
-    _fetch = require('fetch').default;
+    let fetchFn = require('fetch').default;
+    _fetch = () => fetchFn;
   } else if (typeof fetch === 'function') {
     // fallback to using global fetch
-    _fetch = fetch;
+    _fetch = () => fetch;
   } else {
     throw new Error(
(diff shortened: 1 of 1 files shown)


packages/adapter/addon/-private/utils/fetch.ts
```diff
@@ -2,19 +2,23 @@ import require, { has } from 'require';
 
 type FetchFunction = (input: RequestInfo, init?: RequestInit | undefined) => Promise<Response>;
 
-let _fetch: (() => FetchFunction) | null = null;
+let _fetch: FetchFunction | null = null;
```
rwjblue: Actually, we can't cache this here. If we do, I believe that we break the ability for pretender to patch it.

runspired: fixed

PR #6440, opened 2019-09-12
Add failing test for setting an async belongsTo relationship to null, then restoring it

amk221:
This worked in 3.10, but not 3.12.2 (or maybe earlier?)

```javascript
const bar = foo.get('bar');
foo.set('bar', null);
foo.set('bar', bar);    // 'Roll-back' belongs to relationship

// foo.get('bar') === bar      // In 3.10.0
// foo.get('bar') === null     // In 3.12.2
```

FYI this does work when using `foo.belongsTo('bar').value()`, just not when it's a proxy


amk221:
Anything I can do to move this along?

I'm not hugely familiar with ED's internals. But I do know that this is preventing us from upgrading, so I'd love to see it fixed. Willing to have a bash with the right guidance

amk221:
😩

runspired:
@amk221 sorrrrry!!!!!!!!

Yes, I'm hoping to have @code0100fun look at this as part of his work to shore up the unload APIs.

amk221:
Hellooo 👋 
It's been a couple of months, and I'm still stick on 3.10

Really would like to get this sorted. Again, I can try to help if pointed in the right direction.

amk221:
I reeaaally need help on this 🤞

amk221:
Also I npm audit is complaining about the version I am stuck on (3.10)

packages/-ember-data/tests/acceptance/relationships/belongs-to-test.js
```diff
@@ -392,6 +392,38 @@ module('async belongs-to rendering tests', function(hooks) {
...
+      // render
+      this.set('sedona', sedona);
+
+      await render(hbs`
+      <p>{{sedona.parent.name}}</p>
+      `);
+
+      assert.equal(this.element.textContent.trim(), 'Kevin has two children and one parent');
+
+      const parent = sedona.get('parent');
```
snewcomer, line 412: Since this is an `async` relationship, I think the fix for this test is: 

```js
const parent = await sedona.get('parent');
```

Otherwise it is a PromiseProxy.  I think that should fix the test.

amk221: hmm, if the test is wrong, I need to check my app. I'll get back to you

amk221: I think the point is here, that it worked in 3.10

If `sedona.parent` is a proxy in the first place, and it works.
it is then set to `null`
it is then set back to the same proxy

..then I would expect it to be the same.

snewcomer: I see!  That may make sense if we kept around a local copy of the remove "parent".  However, I'm not sure if that was publicly supported behaviour before 3.10 or not.  Might have to dig around commits in that time frame.

amk221:
In trying to investigate this further, I created a separate repo with another [failing test](https://github.com/amk221/-ember-data-relationships-test/blob/5011fb98be416f558ce213d0660760bb95ad60b3/tests/unit/models/rollback-test.js#L8-L43), which I think is the essential 'bug'.

I'm hoping somebody can explain to me why:

- The test passes in 3.10
- The test fails in ~3.17
- And more weirdly: the test passes if you comment out [this line](https://github.com/amk221/-ember-data-relationships-test/blob/5011fb98be416f558ce213d0660760bb95ad60b3/tests/unit/models/rollback-test.js#L17)! (A massive wtf to me)

Thanks for checking it out.

snewcomer:
https://github.com/emberjs/data/compare/v3.12.0-canary.0...v3.12.0

It seems the changes might be somewhere in here.  However, I haven't been able to pinpoint.

Other than a test failure, do you have a app example that shows this failing?

amk221:
Umm, so the test case code Is exactly the same as the app code

snewcomer:
This looks like a better reproduction!  Thank you.  The test fix is still the same - `await foo1.get("bar");`.  `originalBar` is a `PromiseProxy` without `await` (thus is subject to swapping out to another record at some future point in time via the #get trap....and with `await`, it is the actual `BarRecord`. Curious if the app code fixed as easily as the test.

I have a sense of all the moving parts internally.  I just need to pair with @runspired to see if this is a bug/regression/or expected.

amk221:
I kind of see what you mean, but I suppose because it worked without await, I must have created an incorrect mental model.... and thought of `const originalBar = foo.get('bar')` as `originalBar` being a proxy specifically for bar 1.

amk221:
....would love to know the answer to this before I put effort into upgrading.

amk221:
Here is an even more [succinct demonstration](https://ember-twiddle.com/38bc80958a22a29345f3f3f807f1bae0?openFiles=controllers.application%5C.js%2C).

amk221:
To add weight to my argument, consider this:

```javascript
// (Where these both point to the same record)

const fulfilledPromiseProxyObjectA = foo.get('bar');
const fulfilledPromiseProxyObjectB = store.findRecord('bar' 1);

foo.set('bar', null); 
```

In ED 3.18 this sets `fulfilledPromiseProxyObjectA` to null, even though I didn't ask it to.
but `fulfilledPromiseProxyObjectB` is _not null_.

In ED 3.10 both `fulfilledPromiseProxyObjectA` and `fulfilledPromiseProxyObjectB` remain as such - fulfilled promise proxy objects.

runspired:
This change is expected although the app code result is unfortunate in this scenario. If the target of a proxy changes, the proxy's value changes: in the past we seem to have accidentally failed to update the target whereas today we do update the target. This likely is the result of better proxy lifecycle management (previously we were leaking proxy instances).

If the desired app behavior is to retain access to the original value, that value should be resolved first. Resolving promise proxy values prior to using them in JS computations has been a best practice since the beginning of EmberData when Promise Proxies were created to handle the complexity of chaining promisified values in templates.

runspired:
Also I'm sorry I didn't pick up on the semantics here being "reset the value to a previous proxy instance" much earlier so as to help you on this journey, I was under the impression for a long time that this issue was concerning reseting to a resolved value.

amk221:
@runspired Thank you! At least I now can move on with my life :)

> Also I'm sorry I didn't pick up on the semantics here being "reset the value to a previous proxy instance" much earlier so as to help you on this journey, I was under the impression for a long time that this issue was concerning reseting to a resolved value.

Don't apologise, part of the problem was my inability to explain what was happening. Got there in the end.

amk221:
...for the record, I'll be migrating to synchronous relationships 😂

PR #6448, opened 2019-09-13
[CHORE]: eliminate initializer logic not needed after setupStore removed

### LINKED ISSUE #6357: Remove use of setupStore/createStore from tests
This is just a tracking issue for

- [x] removing `createStore` usage
- [x] removing `setupStore` usage
- [x] eliminating any `initializer` logic that remains purely for supporting `createStore` and `setupStore`

cc @HeroicEric

snewcomer:
close #6357

### DIFF
--- a/packages/-ember-data/addon/setup-container.js
+++ b/packages/-ember-data/addon/setup-container.js
@@ -2,18 +2,6 @@ import { DebugAdapter } from './-private';
 import Store from '@ember-data/store';
 import JSONAPIAdapter from '@ember-data/adapter/json-api';
 
-function hasRegistration(application, registrationName) {
-  // fallback our ember-data tests necessary
-  // until we kill-off setupStore
-  // see https://github.com/emberjs/data/issues/6357
-  // or @ember/test-helpers kills off it's
-  // legacy support that calls our initializer with registry
-  // instead of application
-  if (typeof application.hasRegistration !== 'function') {
-    return application.has(registrationName);
-  }
-  return application.hasRegistration(registrationName);
-}
 /*
  Configures a registry for use with an Ember-Data
  store. Accepts an optional namespace argument.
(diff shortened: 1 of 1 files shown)


rwjblue:
Looks like we need to dig into the CI failures here.

packages/-ember-data/addon/setup-container.js
```diff
@@ -1,62 +1,54 @@
...
-  // registry.injection for Ember < 2.1.0
-  // application.inject for Ember 2.1.0+
-  let inject = registry.inject || registry.injection;
-  inject.call(registry, 'controller', 'store', 'service:store');
-  inject.call(registry, 'route', 'store', 'service:store');
-  inject.call(registry, 'data-adapter', 'store', 'service:store');
+function initializeStoreInjections(application) {
+  application.inject('controller', 'store', 'service:store');
+  application.inject('route', 'store', 'service:store');
+  application.inject('data-adapter', 'store', 'service:store');
```
snewcomer: Looks like, based on the external partner tests, that we may need to add a similar deprecation for `registry.injection` or keep it around...What are your thoughts @runspired?

runspired: The one deprecation is enough but we need to restore the deprecated code path

packages/-ember-data/addon/setup-container.js
```diff
@@ -1,62 +1,54 @@
...
+        ` AND your tests must use \`setApplication()\` instead of \`setResolver()\` and` +
+        ` \`module()\` with \`setup*Test()\`instead of \`moduleFor*()\`.`,
+      false,
+      {
+        id: 'ember-data:legacy-test-helper-support',
+        until: '3.17',
+      }
+    );
+
+    application.optionsForType('serializer', { singleton: false });
```
snewcomer: And then lastly, do we want `let registerOptionsForType = application.registerOptionsForType || application.optionsForType;` here still?

packages/-ember-data/addon/setup-container.js
```diff
@@ -46,9 +46,24 @@ function initializeDataAdapter(application) {
 }
 
 function initializeStoreInjections(application) {
-  application.inject('controller', 'store', 'service:store');
-  application.inject('route', 'store', 'service:store');
-  application.inject('data-adapter', 'store', 'service:store');
+  if (DEBUG && !application.inject) {
+    deprecate(
```
runspired: I don’t think we need to also deprecate here as the other deprecation will always be thrown simultaneously. No sense throwing it twice.

packages/-ember-data/addon/setup-container.js
runspired: No I set it up so that in production builds the legacy test support is stripped

PR #6450, opened 2019-09-14
[CHORE]: Eliminate default injections in favor of re-export

snewcomer:
- [x] eliminate explicit registration
- [x] re-export

@runspired Do we also need [this](https://github.com/emberjs/data/blob/f0d74188bc1e4683ac817c09777454e4464b3f01/packages/store/addon/-private/system/core-store.ts#L3298)?

ref #6166

### DIFF
--- a/packages/-build-infra/package.json
+++ b/packages/-build-infra/package.json
@@ -16,6 +16,7 @@
   "dependencies": {
     "@babel/plugin-transform-block-scoping": "^7.6.0",
     "@ember-data/canary-features": "3.14.0-alpha.3",
+    "@ember/edition-utils": "^1.1.1",
     "babel-plugin-debug-macros": "^0.3.3",
     "babel-plugin-feature-flags": "^0.3.1",
     "babel-plugin-filter-imports": "^3.0.0",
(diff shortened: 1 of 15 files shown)


runspired:
> @runspired Do we also need this?

yes but for the adapter package

Gaurav0:
Is there an RFC for this or is it mentioned in another RFC?

I think we need to deprecate access to the explicit registration but keep explicit registration working until 4.0, there are many applications that depend on it.

runspired:
@Gaurav0 

> I think we need to deprecate access to the explicit registration but keep explicit registration working until 4.0, there are many applications that depend on it.

Registration of these is an intimate API at most. We'll do the same thing here we did for the deprecation of the explicit serializer registration and target 3.17 removal https://github.com/emberjs/data/blob/master/packages/store/addon/-private/system/core-store.ts#L3314

rwjblue:
FWIW, I was confused by the title here. When I saw "default injections" I _thought_ that this was talking about the default injection of `this.store` in routes / controllers / etc. After reviewing this is about removing _default registrations_ (not injections), and relying on the resolver to find these objects instead. I figured I'd comment here just in case others were equally confused (and don't take the time to review the diff).

runspired:
@rwjblue agree, but the naming is my fault I worded it this way in the project-trim task list. I have a bad habit of using registration and injection interchangeably due to how most of the time in initializers we were calling "inject" :(

rwjblue:
Oh, ya, no worries here at all, I was just confused until I read the diff (which looks great).

PR #6451, opened 2019-09-14
[DOC serializer] implements MinimumSerializerInterface

runspired:
Replaces #6113 

Ref: #6166


packages/store/addon/-private/ts-interfaces/minimum-serializer-interface.ts
```diff
@@ -0,0 +1,280 @@
...
+  Any serializer in `app/serializers` can be looked up by `name` using `store.serializerFor(name)`.
+
+  ### Default Serializers
+
+  It is recommended that apps write their own serializer to best suit the needs of their API and
+  application.
+
+  However, for applications whose APIs are *very close to* or *exactly* the `REST` or `JSON:API`
+  format the `@ember-data/serializer` package contains implementations these applications can
+  extend. It also contains a simly `JSONSerializer` for serializing to/from very simple JSON objects.
```
pete-the-pete: small typo "contains a simple..."

runspired:
```suggestion
  extend. It also contains a simple `JSONSerializer` for serializing to/from very basic JSON objects.
```

runspired: Thanks!

packages/store/addon/-private/system/fetch-manager.ts
```diff
@@ -136,7 +136,14 @@ export default class FetchManager {
       },
       function(error) {
         if (error instanceof InvalidError) {
-          let parsedErrors = serializer.extractErrors(store, modelClass, error, snapshot.id);
+          let parsedErrors = error.errors;
```
rwwagner90: Should we extract these few lines into a helper function, to avoid duplication, or not worry about it since it is just a few lines?

packages/adapter/addon/-private/utils/serialize-into-hash.js
```diff
@@ -0,0 +1,11 @@
+export default function serializeIntoHash(store, modelClass, snapshot, options = { includeId: true }) {
+  const serializer = store.serializerFor(modelClass.modelName);
+
+  if (typeof serializer.serializeIntoHash === 'function') {
```
rwjblue: We don't expect this to be present? The current code that this is replacing (e.g. in packages/adapter/addon/json-api) didn't guard...

packages/store/addon/-private/ts-interfaces/minimum-serializer-interface.ts
```diff
@@ -0,0 +1,280 @@
...
+
+  Data received from an API response is `"normalized"` into 
+  [JSON:API](https://jsonapi.org/) (the format used internally
+  by `EmberData`), while data sent to an API is `"serialized"`
+  into the format the API expects.
+
+  ### Implementing a Serializer
+
+  There are only two required serializer methods, one for
+  normalizing data in `JSON:API`, and another for serializing
```
rwjblue: Can you tweak this _slightly_ to make it clear that you mean `normalizing data from the "server format" into JSON:API`

packages/store/addon/-private/ts-interfaces/minimum-serializer-interface.ts
```diff
@@ -0,0 +1,280 @@
...
+  records via `Snapshot`s into the expected API.
```
rwjblue: Can you tweak the end of this from "expected API" (which is actually ambiguous, expected by whom?) to `expected server format` or something else that makes it clear which side we are talking about?

packages/store/addon/-private/ts-interfaces/minimum-serializer-interface.ts
```diff
@@ -0,0 +1,280 @@
...
+
+  To implement a serializer, export a class implementing the
+  [MinimumSerializerInterface](MinimumSerializerInterface) from
+  the `app/serializers/` directory. An example is below.
+
+  ```ts
+  import EmberObject from '@ember/object';
+
+  export default class ApplicationSerializer extends EmberObject {
+    normalizeResponse(_, __, rawPayload) {
```
rwjblue: Maybe it makes sense to leave the actual names for the first two args?

packages/store/addon/-private/system/fetch-manager.ts
runspired: the duplication with the logic in the finders? That's temporary while this feature flag still exists.

packages/store/addon/-private/ts-interfaces/minimum-serializer-interface.ts
```diff
@@ -0,0 +1,280 @@
...
+        attributes: snapshot.attributes()
+      };
+
+      return serializedResource;
+    }
+  }
+ ```
+
+
+  #### Serializer Resolution
```
rwjblue: Can you add an example to this section as well? I think it would really help things gel.  

Thinking something like:

```
store.serializerFor('author');

// lookup paths (in order) =>

app/serializers/author.js
app/serializers/application.js
```

packages/store/addon/-private/ts-interfaces/minimum-serializer-interface.ts
```diff
@@ -0,0 +1,280 @@
...
+  The instances of serializers defined in `app/serializers/` can be looked up
+  via `store.serializerFor(name)`.
+
+  It is recommended that applications define only a single `application` adapter and serializer
+  where possible.
+
+  `serializerFor` first attempts to find a serializer with an exact match on `name`,
+  then falls back to checking for the presence of a serializer named `application`.
+
+  Because most requests in `ember-data` occur from the perspective of a primary `type`
```
rwjblue: The ordering of this paragraph seems slightly odd juxtaposed with the one just before it. This explains the reasoning of what ember-data does, maybe it should actually be first?

packages/store/addon/-private/ts-interfaces/minimum-serializer-interface.ts
```diff
@@ -0,0 +1,280 @@
...
+  ### Using a Serializer
+
+  Any serializer in `app/serializers` can be looked up by `name` using `store.serializerFor(name)`.
+
+  ### Default Serializers
+
+  It is recommended that apps write their own serializer to best suit the needs of their API and
+  application.
+
+  However, for applications whose APIs are *very close to* or *exactly* the `REST` or `JSON:API`
```
rwjblue: Do we need to say this specifically here? If their format is very close to what we expect, it is _equally_ easy to do the minimal interface and pass through (vs us JSONAPISerializer), right?

packages/store/addon/-private/ts-interfaces/minimum-serializer-interface.ts
```diff
@@ -0,0 +1,280 @@
...
+   *
+   * @method serialize
+   * @public
+   * @param {Snapshot} snapshot - A Snapshot for the record to serialize
+   * @param {object} options
+   */
+  serialize(snapshot: Snapshot, options?: OptionsHash): JSONObject;
+
+  /**
+   * This method is intended to normalize data into a `JsonApiDocument` representing
```
rwjblue: lets add a link to the specific part of the JSON:API spec that this is talking about.

packages/store/addon/-private/ts-interfaces/minimum-serializer-interface.ts
```diff
@@ -0,0 +1,280 @@
...
+   *
+   * @method normalize [OPTIONAL]
+   * @public
+   * @optional
+   * @param {ShimModelClass} schema - An object with methods for accessing information about
+   *  the type, attributes and relationships of the primary type associated with the request.
+   * @param {JSONObject} rawPayload - Some raw JSON data to be normalized into a `JSON:API` resource.
+   * @param {string} [prop] - When called by the `EmbeddedRecordsMixin` this param will be the
+   *  property at which the object provided as rawPayload was found.
+   * @returns {SingleResourceDocument} - A `JSON:API` document containing a single `resource` as
```
rwjblue: link to the spec here?

packages/store/addon/-private/ts-interfaces/minimum-serializer-interface.ts
```diff
@@ -0,0 +1,280 @@
...
+   * ```js
+   * serializeIntoHash(hash, ModelClass, snapshot, options) {
+   *   hash[`_${snapshot.modelName}`] = this.serialize(snapshot, options).data;
+   * }
+   * ```
+   *
+   * @method serializeIntoHash [OPTIONAL]
+   * @public
+   * @optional
+   * @param hash - an top most object of the request payload onto
```
rwjblue:
```suggestion
   * @param hash - a top most object of the request payload onto
```

packages/store/addon/-private/ts-interfaces/minimum-serializer-interface.ts
```diff
@@ -0,0 +1,280 @@
...
+   * @param options
+   * @returns {void}
+   */
+  serializeIntoHash?(hash: object, schema: ShimModelClass, snapshot: Snapshot, options?: OptionsHash): void;
+
+  /**
+   * This method is called by `store.pushPayload` and should be implemented if
+   * you want to use that method.
+   *
+   * It is recommended to use `store.push` over `store.pushPayload` after normalizing
```
rwjblue: really? Why?

packages/adapter/addon/-private/utils/serialize-into-hash.js
runspired: We didn’t mean for it to be required, and if it’s not present we can safely do a different thing. What this PR does (in addition to the new docs) is ensure that things we hadn’t intended to be required in the past are truly optional.

packages/store/addon/-private/ts-interfaces/minimum-serializer-interface.ts
runspired: I can see the ordering either way “what a serializer does” vs “how/when ember-data is going to use your serializer”. I felt introducing the concept was a slightly better way to go because telling someone how a thing is used when they don’t know what that thing is seemed a bit confusing.

packages/store/addon/-private/ts-interfaces/minimum-serializer-interface.ts
runspired: Yes. I’m not sure the best way to word this for indicating that these are more optional convenience for getting started quickly than the best way to go.

packages/store/addon/-private/ts-interfaces/minimum-serializer-interface.ts
runspired: Unsure, will think on it

packages/store/addon/-private/ts-interfaces/minimum-serializer-interface.ts
runspired: There's a few layers to the motivations for this:

Brittleness

- users don't have clarity around which serializer's `pushPayload` is going to be called and the serializer is not given the context of what the "primary type" for a given pushPayload call is expected to be. This makes the normalization "blind" to what format it is normalizing from.
- the convenience REST serializer expects a format that differs from what would be expected normalizing any other payload.

Features

- users often want access to the records they have pushed, which they get from `push` but not `pushPayload`

Long Term Support

- We've eyeballed both this serializer method and the associated store method as a candidate for deprecation for a long time given it (A) is redundant (B) is not part of any primary APIs (C) encourages a non-linear flow for processing data and (D) is one of several points that force the store to be aware of the specifics of serialization and normalization which is not it's concern

packages/store/addon/-private/ts-interfaces/minimum-serializer-interface.ts
runspired: Added them

packages/store/addon/-private/ts-interfaces/minimum-serializer-interface.ts
```diff
@@ -0,0 +1,313 @@
...
+  @public
+*/
+
+import { Object as JSONObject } from 'json-typescript';
+import Store from '../system/core-store';
+import { JsonApiDocument, SingleResourceDocument } from './ember-data-json-api';
+import Snapshot from '../system/snapshot';
+import ShimModelClass from '../system/model/shim-model-class';
+
+type OptionsHash = Record<string, any>;
```
mike-north: `Record<string, T>` should probably be converted to `Dict<T>`

packages/store/addon/-private/ts-interfaces/minimum-serializer-interface.ts
```diff
@@ -0,0 +1,313 @@
...
+  format when making a request.
+
+  ### Using a Serializer
+
+  Any serializer in `app/serializers` can be looked up by `name` using `store.serializerFor(name)`.
+
+  ### Default Serializers
+
+  It is recommended that apps write their own serializer to best suit the needs of their API and
+  application.
```
mike-north: Again, I am coming across an assertion that things should be done a particular way, and am left without a rationale to explain it (until I keep reading).

If you state the rationale up front and conclude with the recommendation, there's no period of time where readers will wonder.

packages/store/addon/-private/ts-interfaces/minimum-serializer-interface.ts
```diff
@@ -0,0 +1,313 @@
...
+
+  However, for applications whose APIs are *very close to* or *exactly* the `REST` or `JSON:API`
+  format the `@ember-data/serializer` package contains implementations these applications can
+  extend. It also contains a simple `JSONSerializer` for serializing to/from very basic JSON objects.
```
mike-north: Putting myself in the shoes of someone who might take "recommendations" a bit too rigidly, I might have trouble reconciling the existence of these turnkey solutions with the above "we recommend you write your own serializer".

packages/store/addon/-private/ts-interfaces/minimum-serializer-interface.ts
```diff
@@ -0,0 +1,313 @@
...
+
+  Many applications will find writing their own serializer to be more performant and less
+  complex than extending these classes even when their API format is very close to that expected
+  by these serializers.
```
mike-north: Again, this information is what I wish I had when initially reading the above. We should really flip this around

packages/store/addon/-private/ts-interfaces/minimum-serializer-interface.ts
mike-north: I would be prepared for people to take this statement as "`pushPayload` is bad, and you shouldn't use it". AFAIK they're both useful for different things (single record where you get a reference back, vs. complex payload where there is no "primary" thing to return as a record).

This smells a little bit like an attempt to pre-deprecate, which our community has suffered from in the past (e.g., "controllers are dead").

runspired: @mike-north it is what it smells like :( We intend to deprecate this but haven't had time to work on the RFC https://github.com/emberjs/rfcs/issues/357

Now that more contributors are getting involved though maybe there are hands for this. @snewcomer @code0100fun would either of you be interested in taking on this deprecation RFC so that we don't leave this guidance hanging in a weird state?

packages/store/addon/-private/ts-interfaces/minimum-serializer-interface.ts
```diff
@@ -0,0 +1,313 @@
...
+
+      return serializedResource;
+    }
+  }
+ ```
+
+
+  #### Serializer Resolution
+
+  The instances of serializers defined in `app/serializers/` can be looked up
```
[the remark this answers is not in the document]
runspired: @mike-north tbh I'm not sure if there is confusion although within EmberData's own methods there is certainly the potential for this confusion. Contrast `serializerFor` (instance) with `modelFor` (class)

packages/store/addon/-private/ts-interfaces/minimum-serializer-interface.ts
```diff
@@ -0,0 +1,313 @@
...
+  via `store.serializerFor(name)`.
+
+  It is recommended that applications define only a single `application` adapter and serializer
+  where possible.
```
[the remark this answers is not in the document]
runspired: I'll see if I can make that flow nicely :)

packages/store/addon/-private/ts-interfaces/minimum-serializer-interface.ts
```diff
@@ -0,0 +1,313 @@
...
+  into the format the API expects.
+
+  ### Implementing a Serializer
+
+  There are only two required serializer methods, one for
+  normalizing data from the server API format into `JSON:API`, and
+  another for serializing records via `Snapshot`s into the expected
+  server API format.
+
+  To implement a serializer, export a class implementing the
```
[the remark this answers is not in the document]
runspired: I agree with keeping the mental overhead as low as possible for beginners and will look at refactoring the language here.

This raises a broader discussion we should have though about what terminology we want to use or expect folks to be able to pick up quickly as EmberData is moving towards being  heavily "interface" driven.

packages/store/addon/-private/ts-interfaces/minimum-serializer-interface.ts
mike-north: In the short term, I'd leave this out of the docs until such time that we're ready to do this the right way.

Deprecation is the tool we use to steer people away from APIs we intend to remove, and we should use it for things like this

packages/store/addon/-private/ts-interfaces/minimum-serializer-interface.ts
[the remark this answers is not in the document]
ro0gr: Maybe it's a ts user talks from me, but interface seems to be a widely accepted term in programming. I believe it's quite natural and should not cause a confusion to the most of the readers.

packages/store/addon/-private/ts-interfaces/minimum-serializer-interface.ts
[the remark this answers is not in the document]
runspired: I tried to clean this up a little bit but due to potential confusion with the workings of `modelFor` I've left the word `instance` involved

packages/store/addon/-private/ts-interfaces/minimum-serializer-interface.ts
runspired: I updated the recommendation a bit here to be less "soft-deprecation" and more "why/where to steer away"

packages/store/addon/-private/ts-interfaces/utils.ts
```diff
@@ -2,4 +2,4 @@
   @module @ember-data/store
 */
 
-export type Dict<K extends string, V> = { [KK in K]: V };
+export type Dict<V> = { [key: string]: V };
```
mike-north: This definition of `Dict` describes an object where we're 100% sure we'll find present values for any possible property key.

It turns out that this is usually a more accurate type, which puts responsibility on the consumer to check whether they got something after retrieving a property.
```suggestion
export type Dict<V> = { [key: string]: V | undefined };
```

There are some cases that I've discussed with @igorT where it may be more reasonable to assume the properties are definitely present (i.e., `Object.keys(myObj).forEach` ) but I would argue that the more common case is that you _want_ undefined here.

packages/store/addon/-private/ts-interfaces/minimum-serializer-interface.ts
```diff
@@ -0,0 +1,319 @@
...
+   * When using `Model`, this method is called by `record.serialize()`.
+   *
+   * When using `JSONAPIAdapter` or `RESTAdapter` this method is called
+   * by `updateRecord` and `createRecord` if `Serializer.serializeIntoHash`
+   * is not implemented.
+   *
+   * @method serialize
+   * @public
+   * @param {Snapshot} snapshot - A Snapshot for the record to serialize
+   * @param {object} options
```
mike-north:
```suggestion
   * @param {object} [options]
```

packages/store/addon/-private/ts-interfaces/minimum-serializer-interface.ts
```diff
@@ -0,0 +1,319 @@
...
+   *
+   * @method serializeIntoHash [OPTIONAL]
+   * @public
+   * @optional
+   * @param hash - a top most object of the request payload onto
+   *  which to append the serialized record
+   * @param {ShimModelClass} schema - An object with methods for accessing information about
+   *  the type, attributes and relationships of the primary type associated with the request.
+   * @param {Snapshot} snapshot - A Snapshot for the record to serialize
+   * @param options
```
mike-north:
```suggestion
   * @param [options]
```

packages/store/addon/-private/ts-interfaces/utils.ts
runspired: I agree for now but also agree with @igorT. I wonder if we should have a `StrictDict` for when we don't need to check? Otherwise iteration with `Object.keys` gets very annoying.

PR #6456, opened 2019-09-16
Fix/ds docs

Gaurav0:
Fixes some links to classes and reverts `model_package_named_exports` to `@ember-data/model`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1858,7 +1858,7 @@ post.constructor.typeKey => postSnapshot.typeKey
 If you need to access the underlying record of a snapshot you can do so by
 accessing `snapshot.record`.
 
-The full API reference of `DS.Snapshot` can be found [here](https://emberjs.com/ember-data/release/classes/DS.Snapshot).
+The full API reference of `DS.Snapshot` can be found [here](https://emberjs.com/ember-data/release/classes/Snapshot).
 
 #### Changes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1858,7 +1858,7 @@ post.constructor.typeKey => postSnapshot.typeKey
 If you need to access the underlying record of a snapshot you can do so by
 accessing `snapshot.record`.
 
-The full API reference of `DS.Snapshot` can be found [here](https://emberjs.com/api/data/classes/DS.Snapshot.html).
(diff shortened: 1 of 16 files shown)


CHANGELOG.md
```diff
@@ -1858,7 +1858,7 @@ post.constructor.typeKey => postSnapshot.typeKey
 If you need to access the underlying record of a snapshot you can do so by
 accessing `snapshot.record`.
 
-The full API reference of `DS.Snapshot` can be found [here](https://emberjs.com/api/data/classes/DS.Snapshot.html).
+The full API reference of `DS.Snapshot` can be found [here](https://emberjs.com/ember-data/release/classes/DS.Snapshot).
```
runspired: pretty sure `DS.Snapshot` is just `Snapshot` now

packages/model/addon/index.ts
```diff
@@ -35,7 +35,7 @@
 
   @module @ember-data/model
   @main @ember-data/model
-  @class model_package_named_exports
+  @class Model
```
runspired: I think we fully remove `class` here if we aren't using `model_package_named_exports` to create the fake class

packages/serializer/addon/-private/transforms/boolean.js
```diff
@@ -9,7 +9,7 @@ import Transform from './transform';
   The `BooleanTransform` class is used to serialize and deserialize
   boolean attributes on Ember Data record objects. This transform is
   used when `boolean` is passed as the type parameter to the
-  [attr](./model_package_named_exports?anchor=attr) function.
+  [attr](/ember-data/release/modules/@ember-data%2Fmodel/?anchor=attr) function.
```
runspired: are we sure this page exists when consumed in the docs app? it doesn't exist in the yuidocs.

Gaurav0: Yes. It works testing locally.

packages/model/addon/index.ts
Gaurav0: Keeping class here as "Model" creates a useful link to Model class.

Gaurav0:
@runspired There is no reason to "set up" AWS. Feel free to use any credentials you already have or create a free trial from any email account. The AWS files you need to read are permissioned as public.

I can confirm the pages exist.

PR #6459, opened 2019-09-17
 [TEST] ensure @ember-data/serializer package encapsulation

runspired:
This surfaces an issue that shows we have one more action item to fully encapsulate the `@ember-data/serializer` package as part of #6166 

The issue is that `@ember-data/store` has an unexpected entanglement with the ember-data package still via the `ember-data/-debug` util. Since the `ember-data` package requires the `@ember-data/serializer` package this makes us not able to drop the package yet.

Action item to move this debug util into its own package.


runspired:
Realized that only tests and `@ember-data/store` make use of this so I moved the util into the store package for now. Based on how it uses `Model` though long term it will either need to become stand-alone or move to the model package

PR #6468, opened 2019-09-19
[TEST] allInverseRecordsAreLoaded excludes new records

caselas:
Note, I've included a "fix" to simply help illustrate the problem. I don't necessarily expect this to be the final solution, as I'm a first-time contributor and don't know the source code very well.

As of `v3.12.0`, a potential bug exists where a hasMany relationship with _link-no-data_ is refetched upon creating a new record in the data store.

It seems that `_findHasManyByJsonApiResource` checks `allInverseRecordsAreLoaded` and a new record `isEmpty`, so an erroneous `findHasMany` follows.

https://github.com/emberjs/data/blob/4551d05f0bc1dbc38c619d6bd39726e0988bccd5/packages/store/addon/-private/system/core-store.ts#L1622-L1628

https://github.com/emberjs/data/blob/e0732481885abde4eb187d3a1573276374068a2f/packages/store/addon/-private/system/relationships/state/has-many.ts#L277-L281

### DIFF
--- a/packages/-ember-data/tests/acceptance/relationships/has-many-test.js
+++ b/packages/-ember-data/tests/acceptance/relationships/has-many-test.js
@@ -398,6 +398,33 @@ module('async has-many rendering tests', function(hooks) {
       assert.deepEqual(names, ['Selena has a parent', 'Sedona has a parent'], 'We rendered the names');
     });
 
+    test('Re-rendering an async hasMany with a link and a new record does not cause a new fetch', async function(assert) {
+      let people = makePeopleWithRelationshipLinks(true);
+      let parent = store.push({
+        data: people.dict['3:has-2-children-and-parent'],
+      });
+
+      adapter.setupPayloads(assert, [people.links['./person/3:has-2-children-and-parent/children']]);
+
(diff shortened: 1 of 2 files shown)


packages/-ember-data/tests/acceptance/relationships/has-many-test.js
```diff
@@ -398,6 +398,33 @@ module('async has-many rendering tests', function(hooks) {
...
+      {{/each}}
+      </ul>
+    `);
+
+      let items = this.element.querySelectorAll('li');
+      let names = domListToArray(items).map(e => e.textContent);
+
+      assert.deepEqual(names, ['Selena has a parent', 'Sedona has a parent'], 'We rendered the names');
+
+      store.createRecord('person', { parent: parent });
```
runspired: new models shouldn't be empty, something seems fishy here. iirc you mentioned this seemed to break after 3.10? could you bisect, probably starting with 3.11?

runspired: Cc @igort

PR #6455, opened 2019-09-16
Sample pr

Gaurav0:
Sample PR for #6443 

In response to 
https://github.com/emberjs/data/pull/6098#pullrequestreview-236860218

DO NOT MERGE

PR #6443, opened 2019-09-12
Update Azure Pipelines config

Gaurav0:
This rebases #6098 against master (9/18/2019)

Updates the Azure Pipelines config to be more similar to the Travis CI
config.

| Type               | Master | Beta | Release | LTS  | Tags |
| ---                | ---    | ---  | ---     | ---  | ---  |
| Lint               | yes    | yes  | no      | no   | no   |
| Basic Tests        | yes    | yes  | yes     | yes  | no   |
| Build  Tests       | yes    | yes  | yes*    | yes* | no   |
| Windows Tests      | yes    | yes  | yes*    | yes* | no   |
| Floating Deps      | yes    | yes  | yes     | yes  | no   |
| Ember LTS Branches | yes    | yes  | yes     | yes  | no   |
| Release Channels   | yes    | yes  | no      | no   | no   |
| External Partners  | yes    | yes  | no      | no   | no   |

* Runs but skips tests for in-progress-features

Co-Authored-By: @pete-the-pete  Pete <pete-the-pete@users.noreply.github.com>

Anyone with permissions is free to force push this to @HeroicEric 's branch.

Issue #6282, opened 2019-07-25
Kill off use of Travis CI


runspired:
We've been running on Azure with a ton of success for a few months now. #6098 if polished off would reach full parity with what we were doing on Travis before, although what we have now is already better.

This ticket can be closed when

- [x] ~#6098~ #6472 is landed
- [x] CI status badge in README is replaced (https://github.com/emberjs/data/pull/6517)
- [x] #6472 is backported to beta branch
- [x] #6472 is backported to release branch
- [x] #6472 is backported to LTS branch
- [x] #6547 The Azure Pipeline config is removed
- [x] The TravisCI job is removed

runspired:
cc @HeroicEric if you have some time to finish this :)

dhuang612:
I would like to work on this!

runspired:
@dhuang612 thanks :) I'll let @HeroicEric lead

runspired:
Allowed fails are now the only remaining use for TravisCI

We can replace this by using a github action plus function to run a test as "allowed to fail" and then leaving a comment on the PR if the allowed scenario did fail with details.

PR #6472, opened 2019-09-20
Use GitHub Actions for CI

runspired:
written by @HeroicEric (don't want to put words in @runspired's mouth 😄)

---

This supersedes https://github.com/emberjs/data/pull/6098 but with [GitHub actions](https://help.github.com/en/articles/about-github-actions) rather than Azure Pipelines.

Part of https://github.com/emberjs/data/issues/6282

Closes https://github.com/emberjs/data/pull/6443
Closes https://github.com/emberjs/data/pull/6455
Closes https://github.com/emberjs/data/pull/6098

I think we should continue running Travis CI alongside this for a little while to make sure everything is working correctly but we should remove the Azure Pipelines config.

## Why?

- This GitHub version runs much faster because we're given more resources to run jobs concurrently.
- The syntax is also nicer, which is my opinion of course but it seems to be shared by others.

## Stages

The jobs are configured to run in three stages, each only running once all of the jobs in the previous stage have passed:

1. lint & basic-tests
1. floating-dependencies, lts, releases, additional-scenarios
1. external-partners

This setup is intended to make sure that one build is not using all of the repo's allowed jobs. It is a trade off between speed and resource availability that we might want to tweak in the future.

## Allowed failures

GitHub actions, like Azure Pipelines, doesn't provide a feature like [Travis CI's `allow_failures`](https://docs.travis-ci.com/user/customizing-the-build/#rows-that-are-allowed-to-fail), which allows a job to fail without failing the entire build.  As a workaround, this uses the [`continue-on-error`](https://help.github.com/en/articles/workflow-syntax-for-github-actions#jobsjob_idstepscontinue-on-error) feature to allow some external partner tests to fail without failing the entire build. **The downside** is that it is difficult to know when these jobs are actually failing since they appear to have passed in GitHub's UI. We will likely want to do something to address this in the future.

## Downsides

- GitHub Actions Workflows are still in beta so it's possible that there could be breaking changes.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -9,6 +9,8 @@ on:
       - release
       - 'release-*'
       - 'lts-*'
+    tags:
+      - '*'
 
 jobs:
   lint:
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,0 +1,188 @@
+name: CI
+
+on:
+  pull_request:
+  push:
+    branches:
+      - master
+      - beta
+      - release
+      - 'release-*'
+      - 'lts-*'
+
+jobs:
+  lint:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@master
+      - uses: actions/setup-node@v1
+        with:
+          node-version: 10.x
+      - name: Yarn install
+        run: yarn
+      - name: Lint features
+        run: yarn lint:features
+      - name: Lint prettier
+        run: yarn lint:prettier
+      - name: Check for TypeScript problems
+        run: yarn problems
+
+  basic-tests:
+    strategy:
+      matrix:
+        os: [macOS-10.14, windows-2016, ubuntu-18.04]
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@master
+      - uses: actions/setup-node@v1
+        with:
+          node-version: 10.x
+      - name: Install dependencies
+        run: yarn install
+      - name: Basic tests
+        run: yarn test
+      - name: Encapsulation tests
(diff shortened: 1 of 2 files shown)


HeroicEric:
This supersedes https://github.com/emberjs/data/pull/6098 but with [GitHub actions](https://help.github.com/en/articles/about-github-actions) rather than Azure Pipelines.

I think we should continue running Travis CI alongside this for a little while to make sure everything is working correctly but we should remove the Azure Pipelines config.

## Why?

- This GitHub version runs much faster because we're given more resources to run jobs concurrently.
- The syntax is also nicer, which is my opinion of course but it seems to be shared by others.

## Stages

The jobs are configured to run in three stages, each only running once all of the jobs in the previous stage have passed:

1. lint & basic-tests
1. floating-dependencies, lts, releases, additional-scenarios
1. external-partners

This setup is intended to make sure that one build is not using all of the repo's allowed jobs. It is a trade off between speed and resource availability that we might want to tweak in the future.

## Allowed failures

GitHub actions, like Azure Pipelines, doesn't provide a feature like [Travis CI's `allow_failures`](https://docs.travis-ci.com/user/customizing-the-build/#rows-that-are-allowed-to-fail), which allows a job to fail without failing the entire build.  As a workaround, this uses the [`continue-on-error`](https://help.github.com/en/articles/workflow-syntax-for-github-actions#jobsjob_idstepscontinue-on-error) feature to allow some external partner tests to fail without failing the entire build. **The downside** is that it is difficult to know when these jobs are actually failing since they appear to have passed in GitHub's UI. We will likely want to do something to address this in the future.

## Downsides

- GitHub Actions Workflows are still in beta so it's possible that there could be breaking changes.

PR #6476, opened 2019-09-20
Enable eslint-plugin-qunit

yohanmishkin:
- `qunit/assert-args`
- `qunit/literal-compare-order`
- `qunit/no-identical-names`
- `qunit/no-ok-equality` (#6477)
- `qunit/require-expect`
- `qunit/resolve-async`
- remove `bin/lint-js`

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -6,7 +6,7 @@ module.exports = {
     sourceType: 'module',
   },
   plugins: ['prettier', 'qunit', 'mocha'],
-  extends: ['eslint:recommended', 'prettier'],
+  extends: ['eslint:recommended', 'prettier', 'plugin:qunit/recommended'],
   rules: {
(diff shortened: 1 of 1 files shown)


.eslintrc.js
```diff
@@ -37,6 +37,14 @@ module.exports = {
     // Tracked in issue https://github.com/emberjs/data/issues/6405
     'no-prototype-builtins': 'off',
     'require-atomic-updates': 'off',
+
+    // eslint-plugin-qunit
+    'qunit/assert-args': 'off',
+    'qunit/literal-compare-order': 'off',
+    'qunit/no-identical-names': 'off',
+    'qunit/no-ok-equality': 'off',
+    'qunit/require-expect': 'off',
```
rwjblue: I think we want this one to be `except-simple`

PR #6477, opened 2019-09-20
Enable qunit/no-ok-equality eslint rule

yohanmishkin:
Needs #6476 

https://github.com/platinumazure/eslint-plugin-qunit/blob/master/docs/rules/no-ok-equality.md

> Equality comparisons in assert.ok or assert.notOk calls are not valuable because if the assertion fails, QUnit cannot reveal any comparison information. Instead, developers should use assert.equal, assert.strictEqual, assert.notEqual, or assert.notStrictEqual to allow QUnit to track the expected and actual values and show a useful diff. This makes test output much easier to read.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -6,7 +6,7 @@ module.exports = {
     sourceType: 'module',
   },
   plugins: ['prettier', 'qunit', 'mocha'],
-  extends: ['eslint:recommended', 'prettier'],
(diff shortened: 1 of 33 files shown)


Gaurav0:
@runspired I think the situation can generally be avoided by calling .toJSON() when we recognize the problem.

I also think we should stop allowing all these private / intimate parts of a record to be enumerable.

runspired:
Closing as lint mostly has auto fixed this already

PR #6479, opened 2019-09-20
[CHORE] clean up object.assign patterns and remove unneeded code comment

runspired:


packages/store/addon/-private/system/core-store.ts
```diff
@@ -114,21 +114,6 @@ function deprecateTestRegistration(factoryType: 'serializer' | 'adapter', factor
   );
 }
 
-// Implementors Note:
```
rwjblue: This seems unrelated? Is it just outdated / needing cleanup?

runspired: jsut outdated

PR #6480, opened 2019-09-20
[FEATURE] attributesDefintionFor and relationshipsDefinitionFor accept identifiers

igorT:

### DIFF
--- a/packages/store/addon/-private/system/ds-model-store.ts
+++ b/packages/store/addon/-private/system/ds-model-store.ts
@@ -268,9 +268,13 @@ class Store extends CoreStore {
     }
   }
 
-  _relationshipsDefinitionFor(modelName: string): RelationshipsSchema {
+  _relationshipsDefinitionFor(modelName: string, identifier?: StableRecordIdentifier): RelationshipsSchema {
     if (CUSTOM_MODEL_CLASS) {
-      return this.getSchemaDefinitionService().relationshipsDefinitionFor(modelName);
+      if (identifier) {
+        return this.getSchemaDefinitionService().relationshipsDefinitionFor(identifier);
+      } else {
+        return this.getSchemaDefinitionService().relationshipsDefinitionFor(modelName);
+      }
     } else {
       let relationships = this._relationshipsDefCache[modelName];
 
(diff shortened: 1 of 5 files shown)


packages/store/addon/-private/system/ds-model-store.ts
```diff
@@ -245,9 +245,13 @@ class Store extends CoreStore {
     }
   }
 
-  _attributesDefinitionFor(modelName: string) {
+  _attributesDefinitionFor(modelName: string, identifier?: StableRecordIdentifier) {
     if (CUSTOM_MODEL_CLASS) {
-      return this.getSchemaDefinitionService().attributesDefinitionFor(modelName);
+      if (identifier) {
```
rwjblue: I think this should also be guarded by `IDENTIFIERS`, right? Or is that not needed because `CUSTOM_MODEL_CLASS` itself fundamentally requires `IDENTIFIERS`?

packages/store/addon/-private/system/ds-model-store.ts
```diff
@@ -245,9 +245,13 @@ class Store extends CoreStore {
     }
   }
 
-  _attributesDefinitionFor(modelName: string) {
+  _attributesDefinitionFor(modelName: string, identifier?: StableRecordIdentifier) {
```
rwjblue: Probably a silly question, but can't we use an identifier for the model name itself (e.g. when there is no `id`/`lid`)?

packages/store/addon/-private/system/ds-model-store.ts
igorT: Not needed cause CUSTOM_MODEL_CLASS requires IDENTIFIERS

packages/store/addon/-private/system/ds-model-store.ts
igorT: Yup, going to propose an amendment to the RFC to slightly change the design to make it nicer to use

rwjblue: Sounds good, we should probably try to do that before enabling the `CUSTOM_MODEL_CLASS` flag.

PR #6484, opened 2019-09-22
[FEATURE] Return pushPayload result

esbanarango:
This PR adds a `return` value (`the record(s) that was created or updated`) to `pushPayload`.

This was already done here: https://github.com/emberjs/data/pull/5265/files but apparently was mistakenly removed here https://github.com/emberjs/data/pull/5384/files

### DIFF
--- a/packages/serializer/addon/json-api.js
+++ b/packages/serializer/addon/json-api.js
@@ -211,10 +211,11 @@ const JSONAPISerializer = JSONSerializer.extend({
     @method pushPayload
     @param {DS.Store} store
     @param {Object} payload
+    @return the record(s) that was created or updated.
   */
   pushPayload(store, payload) {
     let normalizedPayload = this._normalizeDocumentHelper(payload);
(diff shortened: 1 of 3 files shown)


jonnii:
I think this was feature flagged and removed on purpose if my memory serves correctly.

runspired:
@jonnii is correct. This feature never landed and we decided not to push it forward.

esbanarango:
@runspired do we have any context on why the decision was made?

runspired:
@esbanarango many factors

- the implementation had gone stale and didn't have a champion
- the change is complicated semantically as there's no ability to determine whether primary or all records should be returned
- we want to move away from pushPayload entirely
- a public API for this already exists

esbanarango:
@runspired Thank you 🙏

runspired:
@esbanarango there is even an open PR atm that will update the docs to actively discourage use of it: https://github.com/emberjs/data/pull/6451#discussion_r325296660 :)

PR #6485, opened 2019-09-22
[CHORE] Use `ember-data` without `@ember-data/debug`

pete-the-pete:
Provides the DataAdapter for the ember-inspector

- [x] refactor to use service injection, kill initializer injection
- [x] refactor to manage own registration
- [ ] `detect` should not require `Model`
- [ ] stop automatically shipping to production (ship by config only)
- [ ] Add Encapsulation Test

### DIFF
--- a/packages/-ember-data/addon/-private/system/debug/debug-adapter.js
+++ b/packages/-ember-data/addon/-private/system/debug/debug-adapter.js
@@ -1,7 +1,6 @@
 /**
   @module ember-data
 */
-import { getOwner } from '@ember/application';
 import { addObserver, removeObserver } from '@ember/object/observers';
 import { inject as service } from '@ember/service';
 import { A } from '@ember/array';
@@ -20,14 +19,6 @@ import Model from '@ember-data/model';
 */
 export default DataAdapter.extend({
   store: service('store'),
-
-  init() {
-    this._super(...arguments);
-    let owner = getOwner(this.get('store'));
-    if (!owner.hasRegistration('data-adapter:main')) {
-      owner.register('data-adapter:main', this);
-    }
-  },
 
   /**
     Specifies how records can be filtered based on the state of the record
--- a/packages/-ember-data/addon/-private/system/debug/debug-adapter.js
+++ b/packages/-ember-data/addon/-private/system/debug/debug-adapter.js
@@ -1,8 +1,9 @@
 /**
   @module ember-data
 */
+import { getOwner } from '@ember/application';
 import { addObserver, removeObserver } from '@ember/object/observers';
-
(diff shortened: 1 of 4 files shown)


packages/-ember-data/addon/-private/system/debug/debug-adapter.js
```diff
@@ -18,6 +19,16 @@ import Model from '@ember-data/model';
   @private
 */
 export default DataAdapter.extend({
+  store: service('store'),
+
+  init() {
+    this._super(...arguments);
+    let owner = getOwner(this.get('store'));
+    if (!owner.hasRegistration('data-adapter:main')) {
```
runspired: I believe this is too late in time. I may be wrong but I think files at the top level are considered "main", so exporting this as `app/debug-adapter.js` should "work". @rwjblue can you confirm?

pete-the-pete: I believe you, but can you explain (🤞 with links) on how it "work" so I can understand?

rwjblue: yes, `app/debug-adapter.js` would be what is looked for when you do `owner.lookup('debug-adapter:main')`.

packages/-ember-data/addon/setup-container.js
```diff
@@ -41,19 +40,13 @@ function initializeStore(application) {
   }
 }
 
-function initializeDataAdapter(application) {
-  application.register('data-adapter:main', DebugAdapter);
```
runspired: with this removed there should be an `app/data-adapter.js` file now which contains that `DebugAdapter`

pete-the-pete:
Ya, I created `packages/-ember-data/app/debug-adapter.js` locally, but I
haven't updated b/c I wanted to see how the tests would behave. I should
have some time today and will update anyway if I can't figure it out.

pete-the-pete:
Ok, updated...I think I'm running the external-partner tests wrong again, b/c i _thought_ they passed locally.

I'll take another look at what's going on/wrong, but I think it is the import failing for the test.

rwjblue:
Hmm, this doesn't seem to do what the PR title suggests. I think the change is great, but it doesn't yet move the debug adapter into `@ember-data/debug` (for example). It removes even more cruft from our initializer setup (:tada:).

runspired:
Yeah it’s not the full set of things but seemed better to land these wins and then work on the move as a follow up

rwjblue:
yep, seems good. just would love to see the title/PR description updated as we change what the intention of the PR is. Makes it easier to review / understand.

PR #6488, opened 2019-09-23
[CHORE/TESTS] Use strictEqual for identifiers/cache test

dcyriller:
Instead of the more permissive `assert.equal`.

This is a follow-up of https://github.com/emberjs/data/pull/6428


rwjblue:
Restarted the failed Travis CI job (looked like vm failure?)...

PR #6502, opened 2019-09-25
[TEST BETA CANARY] Test against Ember 3.12 LTS

Gaurav0:
Question: Should we also test against Ember 3.12 LTS with jquery?


runspired:
We actually made this fix to LTS and Release already, so removed the Release tag :)

PR #6507, opened 2019-09-25
[DOCS] Fix some links

Gaurav0:
Fixes some more links

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1858,7 +1858,7 @@ post.constructor.typeKey => postSnapshot.typeKey
 If you need to access the underlying record of a snapshot you can do so by
 accessing `snapshot.record`.
 
-The full API reference of `DS.Snapshot` can be found [here](https://emberjs.com/ember-data/release/classes/Snapshot).
+The full API reference of `DS.Snapshot` can be found [here](https://api.emberjs.com/ember-data/release/classes/Snapshot).
 
 #### Changes
 
@@ -2092,7 +2092,7 @@ to set metadata.
 
 ##### `ManyArray`s are no longer `RecordArray`s
 
-[ManyArray](https://emberjs.com/ember-data/release/classes/ManyArray),
+[ManyArray](https://api.emberjs.com/ember-data/release/classes/ManyArray),
 the object Ember Data uses to represent `DS.hasMany` relationships has
 been changed so it no longer extends from `RecordArray`. This means if
 you were relying on the RecordArray's `content` property to access the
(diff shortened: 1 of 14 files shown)


CHANGELOG.md
```diff
@@ -2092,7 +2092,7 @@ to set metadata.
 
 ##### `ManyArray`s are no longer `RecordArray`s
 
-[ManyArray](https://emberjs.com/ember-data/release/classes/ManyArray),
+[ManyArray](https://api.emberjs.com/@ember-data/release/classes/ManyArray),
```
runspired:
```suggestion
[ManyArray](https://api.emberjs.com/ember-data/release/classes/ManyArray),
```

packages/model/addon/-private/belongs-to.js
```diff
@@ -149,7 +149,7 @@ function belongsTo(modelName, options) {
         }
         if (opts.hasOwnProperty('serialize')) {
           warn(
-            `You provided a serialize option on the "${key}" property in the "${this._internalModel.modelName}" class, this belongs in the serializer. See Serializer and it's implementations https://emberjs.com/ember-data/release/classes/Serializer`,
+            `You provided a serialize option on the "${key}" property in the "${this._internalModel.modelName}" class, this belongs in the serializer. See Serializer and it's implementations /ember-data/release/classes/Serializer`,
```
runspired: since this is a warning in the console I don't think we can use a relative link. We should update it to point at the serializer package overview though :)

packages/model/addon/-private/belongs-to.js
```diff
@@ -159,7 +159,7 @@ function belongsTo(modelName, options) {
 
         if (opts.hasOwnProperty('embedded')) {
           warn(
-            `You provided an embedded option on the "${key}" property in the "${this._internalModel.modelName}" class, this belongs in the serializer. See EmbeddedRecordsMixin https://emberjs.com/ember-data/release/classes/EmbeddedRecordsMixin`,
+            `You provided an embedded option on the "${key}" property in the "${this._internalModel.modelName}" class, this belongs in the serializer. See EmbeddedRecordsMixin /ember-data/release/classes/EmbeddedRecordsMixin`,
```
runspired: same here

CHANGELOG.md
Gaurav0: Updated

packages/model/addon/-private/belongs-to.js
Gaurav0: I fixed the link, but I think pointing it to the serializer package overview should be left to a separate PR, as that is not really fixing docs.

packages/model/addon/-private/belongs-to.js
Gaurav0: Link fixed.

PR #6508, opened 2019-09-26
[CHORE] scaffold debug and record-data packages

runspired:
The packages RFC introduced the `debug` and `record-data` packages but without any importable modules, so we haven't needed them before. As we prepare to move code into these packages as part of Project Trim we now need them.

cc @pete-the-pete @zinyando


runspired:
@igorT reviewed but without being signed in.

PR #6509, opened 2019-09-26
Improved guarding in _triggerDeferredTriggers()

### LINKED ISSUE #6504: Can't use 'trigger' as a table field name
Added a new table field named `trigger` to our database and it generates a type error when added to the model. The record is loaded but the error breaks rendering

```shell
Uncaught TypeError: trigger.apply is not a function
```
Digging into the source the error is generated in the private `internal-model.js` module, specifically:

```js
  _triggerDeferredTriggers() {
    if (!this.hasRecord) {
      return;
    }
    let triggers = this._deferredTriggers;
    let record = this._record;
    let trigger = record.trigger;
    for (let i = 0, l = triggers.length; i < l; i++) {
      trigger.apply(record, triggers[i]);
    }

    triggers.length = 0;
  }
```
`let trigger =` gets the field value and tries to execute it.  I noticed a guard was recently added for the loop in pr #6313. The guard works if the data field is empty but not otherwise.

The fix would also check if the trigger is a function

```js
  if (trigger && typeof trigger === 'function')
      for (let i = 0, l = triggers.length; i < l; i++) {
        trigger.apply(record, triggers[i]);
      }
```
If this looks like a reasonable fix I can open a PR.

maxwondercorn:
Verifies the trigger is a function.  I noticed a couple tests failed but they also failed without the additional guard added. Fixes #6504

### DIFF
--- a/packages/store/addon/-private/system/model/internal-model.ts
+++ b/packages/store/addon/-private/system/model/internal-model.ts
@@ -1265,7 +1265,7 @@ export default class InternalModel {
     let record = this._record;
     let trigger = record.trigger;
     // TODO Igor make nicer check
-    if (trigger) {
+    if (trigger && typeof trigger === 'function') {
       for (let i = 0, l = triggers.length; i < l; i++) {
         let eventName = triggers[i];
         trigger.apply(record, eventName);


runspired:
@maxwondercorn what you thought locally were failures were probably the skipped tests which testem+headless chrome shows as failures incorrectly

PR #6513, opened 2019-09-26
[CHORE] redivide assets to move record-data into package

runspired:
Types are hard :(

A few notes:

**TS Config Pains**

- whatever package you open first in vscode is the tsconfig.json that applies to the rest
- the `-ember-data` tsconfig.json is what is used when running tests and applied to all packages
- running `tsc` (as we do during CI and release) correctly uses the appropriate tsconfig.json for each package

This is gnarly and leads to a horrible experience. We should refactor to a single root tsconfig.json and teach ember-cli-typescript how to find it.

**Circular Types & Imports**

- some interfaces were refactored and a new technique for type import was introduced to handle issues around circular type imports between the store and record-data packages. As we further refactor away from `internal-model` we should be able to eliminate much of this weirdness.
- a few utils are now duplicated in record-data / store to avoid circular imports (`relationshipStateFor` `coerceId`)

**DefaultRecordData**

We now wire RecordData in via the `ember-data/store` for apps consuming `ember-data`. "BYOPackages" mode requires doing the same.


.eslintignore
```diff
@@ -6,6 +6,7 @@ vendor/
 /packages/-ember-data/docs/
 dist
 tmp
+.git/
```
rwjblue: O_o I've never seen eslint try to traverse the .git directory before!?!

packages/-ember-data/addon/store.ts
```diff
@@ -1 +1,21 @@
-export { default } from '@ember-data/store';
+import Store from '@ember-data/store';
+import { RecordData } from '@ember-data/record-data/-private';
+import { identifierCacheFor } from '@ember-data/store/-private';
+import { IDENTIFIERS } from '@ember-data/canary-features';
+
+type RecordDataStoreWrapper = import('@ember-data/store/-private/ts-interfaces/record-data-store-wrapper').RecordDataStoreWrapper;
```
rwjblue: Non-blocking: can we import from `@ember-data/store/-private` instead?

packages/-ember-data/tests/integration/identifiers/configuration-test.ts
```diff
@@ -533,7 +533,7 @@ if (IDENTIFIERS) {
             name: 'callback-hell',
           },
           relationships: {
-            retainer: { type: 'retainer', id: '1' },
+            retainer: { data: { type: 'retainer', id: '1' } },
```
rwjblue: Can you help me understand this test change? Doesn't immediately seem related to the other changes in this PR to me...

packages/-ember-data/tsconfig.json
```diff
@@ -25,13 +25,14 @@
       "@ember-data/model/*": ["../model/addon/*"],
       "@ember-data/adapter": ["../adapter/addon"],
       "@ember-data/adapter/*": ["../adapter/addon/*"],
+      "@ember-data/adapter/error": ["../adapter/addon/error"],
```
rwjblue: Do we make changes related to `@ember-data/adapter/error` in this PR?

packages/-serializer-encapsulation-test-app/tests/integration/create-record-test.js
```diff
@@ -27,10 +27,11 @@ class Person extends Model {
   }
 }
 
-module('inteegration/create-record - running createRecord with minimum serializer', function(hooks) {
+module('integration/create-record - running createRecord with minimum serializer', function(hooks) {
   setupTest(hooks);
 
   hooks.beforeEach(function(assert) {
+    this.owner.unregister('service:store');
```
rwjblue: Why do we need to unregister? if we are explicitly registering (just below) and only rely on resolution (e.g. there is no earlier `owner.register('service:store')` ) this change should not be needed.

packages/-serializer-encapsulation-test-app/tests/integration/errors-test.js
```diff
@@ -20,6 +20,7 @@ module(
     setupTest(hooks);
 
     hooks.beforeEach(function() {
+      this.owner.unregister('service:store');
```
rwjblue: Seems odd, I'd like to understand better why this is needed.

packages/-serializer-encapsulation-test-app/tests/integration/relationships-test.js
```diff
@@ -26,6 +26,7 @@ module('integration/relationships - running requests for async relatonships with
   setupTest(hooks);
 
   hooks.beforeEach(function() {
+    this.owner.unregister('service:store');
```
rwjblue: Same 🌮

packages/-serializer-encapsulation-test-app/tests/integration/requests-test.js
```diff
@@ -10,6 +10,7 @@ module('integration/requests - running requests with minimum serializer', functi
   setupTest(hooks);
 
   hooks.beforeEach(function(assert) {
+    this.owner.unregister('service:store');
```
rwjblue: 🥘

packages/record-data/addon/-private/normalize-link.ts
```diff
@@ -1,19 +1,18 @@
...
-
 /*
   This method normalizes a link to an "links object". If the passed link is
   already an object it's returned without any modifications.
 
   See http://jsonapi.org/format/#document-links for more information.
 
   @method _normalizeLink
-  @private
+  @internal
```
rwjblue: Is this an actual valid pragma?

packages/record-data/addon/-private/normalize-link.ts
```diff
@@ -1,19 +1,18 @@
...
   @param {String} link
   @return {Object|null}
 */
-export default function _normalizeLink(link) {
+type Link = import('@ember-data/store/-private/ts-interfaces/ember-data-json-api').Link;
```
rwjblue: Seems better to import from `-private`, no?

packages/record-data/addon/-private/record-data.ts
```diff
@@ -98,7 +99,7 @@ export default class RecordDataDefault implements RelationshipRecordData {
       changedKeys = this._changedKeys(data.attributes);
     }
 
-    assign(this._data, data.attributes);
+    assign(this._data, data.attributes || {});
```
rwjblue: Hmm, this seems odd. `Ember.assign` (and Object.assign) both allow for passed arguments to be `null` or `undefined` and skip them. I think this fallback should be removed.

```suggestion
    assign(this._data, data.attributes);
```

packages/record-data/addon/-private/record-data.ts
```diff
@@ -347,7 +348,7 @@ export default class RecordDataDefault implements RelationshipRecordData {
     }
     let changedKeys = this._changedKeys(newCanonicalAttributes);
 
-    assign(this._data, this.__inFlightAttributes, newCanonicalAttributes);
+    assign(this._data, this.__inFlightAttributes, newCanonicalAttributes || {});
```
rwjblue: See above, there should be no reason for the fallback.

packages/record-data/addon/-private/relationships/state/relationship.ts
```diff
@@ -509,11 +515,11 @@ export default class Relationship {
         const id = guidFor(inverseRecordData);
 
         if (this._hasSupportForImplicitRelationships(inverseRecordData) && seen[id] === undefined) {
-          const relationship = implicitRelationshipStateFor(inverseRecordData, this.inverseKeyForImplicit)
+          const relationship = implicitRelationshipStateFor(inverseRecordData, this.inverseKeyForImplicit);
```
rwjblue: 😱 I'm surprised we didn't have this issue lint warnings?!

packages/record-data/types/@ember/debug/index.d.ts
```diff
@@ -0,0 +1,2 @@
+export { assert, warn, inspect } from '@ember/debug';
```
rwjblue: Why do we need these specializations? The fact that this is an export from indicates that the upstream definitely typed types already have them. Looks like the only one it doesn't have is `deprecate`? Can you cross link an issue on definitelytyped to get this fixed? we shouldn't need to define this at all.

packages/record-data/types/@ember/polyfills/index.d.ts
```diff
@@ -0,0 +1,7 @@
+/**
```
rwjblue: Why do we have to define this? Doesn't DT have `@ember/polyfills`?

packages/record-data/types/@ember/runloop/index.d.ts
```diff
@@ -0,0 +1,2 @@
+
+export const run: any;
```
rwjblue: O_o seems bad?

packages/store/addon/-private/system/record-array-manager.js
```diff
@@ -351,18 +351,17 @@ function updateLiveRecordArray(array, internalModels) {
...
-      list[j]._removeInternalModels([internalModel]);
-    }
-
-    internalModel._recordArrays.clear();
+    removeFromAll(internalModels[i]);
   }
 }
 
+function removeFromAll(internalModel) {
+  internalModel._recordArrays.forEach(function(recordArray) {
```
rwjblue: Why move to `.forEach` here instead of sticking with the original looping?

packages/store/addon/-private/system/snapshot.js
```diff
@@ -5,9 +5,19 @@ import { inspect } from '@ember/debug';
 import EmberError from '@ember/error';
 import { get } from '@ember/object';
 import { assign } from '@ember/polyfills';
-import recordDataFor, { relationshipStateFor } from './record-data-for';
+import recordDataFor from './record-data-for';
 import { CUSTOM_MODEL_CLASS } from '@ember-data/canary-features';
 
+function relationshipsFor(instance) {
+  let recordData = recordDataFor(instance) || instance;
```
rwjblue: This fallback seems vaguely odd, why do we need it?

packages/-ember-data/addon/store.js
```diff
@@ -0,0 +1,19 @@
+import Store from '@ember-data/store';
```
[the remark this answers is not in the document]
runspired: Already done, I think you needed to refresh

.eslintignore
runspired: but it does :( I was digging into our eslint perf and this was one of a few areas causing weirdness.

packages/-ember-data/addon/store.ts
runspired: we would have to incur the bloat of re-exporting our interfaces there, so this seems better ?

packages/-ember-data/tests/integration/identifiers/configuration-test.ts
runspired: once we correctly configured TS and fixed the type issues for relationships on payloads which that uncovered TS alerted us that this test was using a malformed payload.

packages/-ember-data/tsconfig.json
runspired: no, but this config is used when we run tests etc. so it is forced to see all the things. This should have been captured already with the `*` but I think vscode auto updated this based on an autocomplete.

packages/record-data/addon/-private/normalize-link.ts
runspired: yes, it prevents it from showing in the docs at all which is what we want given this method has never been user importable

packages/record-data/addon/-private/normalize-link.ts
runspired: Same reasoning as above, we would have to re-export all the types which seems bad

packages/record-data/addon/-private/record-data.ts
runspired: The types for Ember.assign say otherwise :'(

runspired: We tried to overload but I think they weren't getting discovered by vscode, I think now that I know the mechanism by which vscode trolls us (basically never trust vscode) likely the overload was working and we can revert this change.

packages/record-data/addon/-private/relationships/state/relationship.ts
runspired: We have a PR from @HeroicEric in the works to fix lint, turns out we weren't linting any TS files

packages/record-data/types/@ember/debug/index.d.ts
runspired: there already is an open issue, and the way they were typed you can't just add deprecate.

packages/record-data/types/@ember/polyfills/index.d.ts
runspired: it does but doesn't have the right signature, there was an open issue

packages/record-data/types/@ember/runloop/index.d.ts
runspired: would you want to try to type backburner :D

packages/store/addon/-private/system/record-array-manager.js
runspired: good point. I moved to forEach at first trying to resolve the "creating an extra nested closure"  lint error (which didn't work), now that it is a helper function the original loop should be fine, will restore

packages/store/addon/-private/system/snapshot.js
runspired: we may have received

- a recordData
- a snapshot
- an internalModel
- one of a host of other things

some of this has improved since we originally wrote this util but we should revisit soon

packages/record-data/types/@ember/debug/index.d.ts
rwjblue: Ya, seems good can we link to that from a comment in here? Seems fine to make an issue with a checklist of all our overrides so we can audit them (and either remove them or add links to the issue)

packages/record-data/types/@ember/polyfills/index.d.ts
rwjblue: Let's make an issue for that and link here

packages/record-data/types/@ember/runloop/index.d.ts
rwjblue: haha, I did 😉

packages/store/addon/-private/system/snapshot.js
rwjblue: Gotcha, seems like we need this as typescript so that set of overloads is clearer.

packages/-serializer-encapsulation-test-app/tests/integration/create-record-test.js
runspired: wasn't required afterall, removed

packages/record-data/types/@ember/polyfills/index.d.ts
runspired: https://github.com/DefinitelyTyped/DefinitelyTyped/issues/38681

packages/record-data/types/@ember/debug/index.d.ts
runspired: https://github.com/DefinitelyTyped/DefinitelyTyped/issues/38682

packages/record-data/types/@ember/runloop/index.d.ts
runspired: We do this specifically because we're reaching in to avoid triggering the autorun assertion, left a comment

chriskrycho:
I'm happy to help tweak the experience for making sure it works well for VS Code – it'll be relatively straightforward, and it'll be useful prior art for other projects I'm working on. The two approaches that should Just Work™ are:

- Create  VS Code workspace and open *that* instead of opening individual packages.
- Create a root `tsconfig.json` and have the other `tsconfig.json` files extend from it. The bits that matter for `ember-cli-typescript` should all still work as expected, since the specific mappings it sets up are *not* used for transpilation, *only* for type resolution.

packages/store/addon/-private/system/core-store.ts
```diff
@@ -71,15 +67,18 @@ import { RequestPromise } from './request-cache';
...
 import ShimModelClass from './model/shim-model-class';
 import RecordDataRecordWrapper from '../ts-interfaces/record-data-record-wrapper';
-import Reference from './references/reference';
+import RecordData from '../ts-interfaces/record-data';
 import { Dict } from '../ts-interfaces/utils';
 
 import constructResource from '../utils/construct-resource';
 import { errorsArrayToHash } from './errors-utils';
+
+type Relationship = import('@ember-data/record-data/-private').Relationship;
```
chriskrycho: My guess is you're using this as a way of working around the fact that you want *only* a type imported and *not* the runtime class, presumably to avoid circular dependencies in the module graph?

For what it's worth, the Babel build *should* handle type-only import usage correctly (as of a relatively recent fix).

```suggestion
import { Relationship } from '@ember-data/record-data/-private';
```

At least in my copy of this PR, there's no build or test issue with using the normal import here!

packages/store/addon/-private/system/model/internal-model.ts
```diff
@@ -36,6 +32,19 @@ import { StableRecordIdentifier } from '../../ts-interfaces/identifier';
 import { internalModelFactoryFor, setRecordIdentifier } from '../store/internal-model-factory';
 import CoreStore from '../core-store';
 import coerceId from '../coerce-id';
+import recordDataFor from '../record-data-for';
+
+type DefaultRecordData = import('@ember-data/record-data/-private').RecordData;
```
chriskrycho: As elsewhere:

```suggestion
import { RecordData as DefaultRecordData } from '@ember-data/record-data/-private';
```

packages/store/addon/-private/system/model/internal-model.ts
```diff
@@ -36,6 +32,19 @@ import { StableRecordIdentifier } from '../../ts-interfaces/identifier';
...
+
+function relationshipsFor(instance) {
+  let recordData = recordDataFor(instance) || instance;
+
+  return recordData._relationships;
+}
+
+function relationshipStateFor(instance, propertyName) {
```
chriskrycho: Besides my comments above, presumably `propertyName` should be `: string` here?

packages/store/addon/-private/system/model/internal-model.ts
```diff
@@ -36,6 +32,19 @@ import { StableRecordIdentifier } from '../../ts-interfaces/identifier';
 import { internalModelFactoryFor, setRecordIdentifier } from '../store/internal-model-factory';
 import CoreStore from '../core-store';
 import coerceId from '../coerce-id';
+import recordDataFor from '../record-data-for';
+
+type DefaultRecordData = import('@ember-data/record-data/-private').RecordData;
+
+function relationshipsFor(instance) {
```
chriskrycho: I recognize the problem you have down inside `recordDataFor` at the moment (though it *may* be possible for you to work around that with something like what I'm suggesting in the main review comment)—but I wonder if we could constrain the argument and return types *here* not to be implicitly `any`, so that at least callers of *this* will be enabled to do the right thing?

packages/store/addon/-private/system/model/internal-model.ts
```diff
@@ -215,7 +224,7 @@ export default class InternalModel {
 
   get _recordArrays() {
     if (this.__recordArrays === null) {
-      this.__recordArrays = new OrderedSet();
+      this.__recordArrays = new Set();
```
chriskrycho: I strongly recommend (I know it's not new here) explicitly naming the return type of this getter.

packages/store/addon/-private/system/model/internal-model.ts
```diff
@@ -654,7 +663,7 @@ export default class InternalModel {
   }
 
   getBelongsTo(key, options) {
-    let resource = this._recordData.getBelongsTo(key);
+    let resource = (this._recordData as DefaultRecordData).getBelongsTo(key);
```
chriskrycho: Is this cast safe? If so, how do we *know* it's safe?

Note that if it *is* safe, then the `!`s on line 684 become unnecessary.

packages/store/addon/-private/system/model/internal-model.ts
```diff
@@ -707,7 +716,7 @@ export default class InternalModel {
   // TODO Igor consider getting rid of initial state
   getManyArray(key, isAsync = false) {
     let relationshipMeta = this.store._relationshipMetaFor(this.modelName, null, key);
-    let jsonApi = this._recordData.getHasMany(key);
+    let jsonApi = (this._recordData as DefaultRecordData).getHasMany(key);
```
chriskrycho: Same question here and for all similar casts below: how do we know this cast is safe?

packages/store/addon/-private/system/snapshot.js
chriskrycho: If I've followed the PR correctly, we end up with three identical copies of this function. Is that *desired* here, and if so, do you have a strategy in mind for further refactoring to get that down to a single (typed) implementation in the future?

packages/store/addon/-private/ts-interfaces/ember-data-json-api.ts
```diff
@@ -6,6 +6,18 @@ import { Dict } from './utils';
...
 
 export type Meta = Dict<JSONValue>;
+export type LinkObject = { href: string; meta?: Dict<JSONValue> };
+export type Link = string | LinkObject;
+export interface Links {
+  related?: Link;
+  self?: Link;
+}
+export interface PaginationLinks extends Links {
+  first?: Link | null;
```
chriskrycho: Just so I'm clear: this is doing optional-or-`null` for conformity with the JSON API spec, which allows these to be left off *or* explicitly set to `null`?

packages/store/types/@ember/runloop/index.d.ts
```diff
@@ -1,2 +1,4 @@
-
-export const run: any;
\ No newline at end of file
+// necessary because our "run" is run.backburner
+// which we use to avoid autorun triggering for Ember <= 3.4
+// we can drop this and use run directly ~11/1/2019
+export const run: any;
```
chriskrycho: I *think* we *should* be able to put these in a types-only `types` package in the `packages` dir instead of repeating them in each package? But cc. @mike-north who has worked with the monorepo structure more than I have.

packages/-ember-data/tests/integration/multiple-stores-test.js
```diff
@@ -2,7 +2,8 @@ import { run } from '@ember/runloop';
 import { setupTest } from 'ember-qunit';
 import { module, test } from 'qunit';
 import { get } from '@ember/object';
-import Store from '@ember-data/store';
+// we intentionally test against the ember-data version here
```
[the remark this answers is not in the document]
runspired: no it's because the `ember-data/store` uses `DefaultRecordData` while `@ember-data/store` does not

packages/record-data/addon/-private/coerce-id.ts
```diff
@@ -0,0 +1,39 @@
...
+
+export function ensureStringId(id: Coercable): string {
+  let normalized: string | null = null;
+  if (typeof id === 'string') {
+    normalized = id.length > 0 ? id : null;
+  } else if (typeof id === 'number' && !isNaN(id)) {
+    normalized = '' + id;
+  }
+
+  if (DEBUG && normalized === null) {
```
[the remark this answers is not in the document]
runspired: yep :)

packages/-ember-data/addon/store.ts
```diff
@@ -1 +1,21 @@
-export { default } from '@ember-data/store';
+import Store from '@ember-data/store';
+import { RecordData } from '@ember-data/record-data/-private';
+import { identifierCacheFor } from '@ember-data/store/-private';
+import { IDENTIFIERS } from '@ember-data/canary-features';
```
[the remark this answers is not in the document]
runspired: No, TS conversion is a slow "when it happens" thing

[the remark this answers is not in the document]
runspired: (I'm honestly surprised we've converted as much as we have)

packages/-ember-data/tests/integration/multiple-stores-test.js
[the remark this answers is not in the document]
runspired: I'll add a comment about this to the file

packages/-serializer-encapsulation-test-app/app/services/store.js
```diff
@@ -0,0 +1,19 @@
+import Store from '@ember-data/store';
```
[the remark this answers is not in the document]
runspired: We don't use the `ember-data` package in encapsulation tests, so the configuration is up to the consuming application

packages/record-data/addon/-private/relationships/state/relationship.ts
```diff
@@ -202,7 +202,7 @@ export default class Relationship {
     return recordData._implicitRelationships !== undefined && recordData._implicitRelationships !== null;
   }
 
-  _hasSupportForRelationships(recordData: RelationshipRecordData): boolean {
+  _hasSupportForRelationships(recordData: RelationshipRecordData): recordData is RecordData {
```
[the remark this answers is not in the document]
runspired: This is absolutely correct but is definitely "insider knowledge", there is no real way around this sort of magic until Singleton RecordData lands to fix the relationship layer.

packages/record-data/addon/-private/ts-interfaces/relationship-record-data.ts
```diff
@@ -0,0 +1,33 @@
...
+export interface RelationshipRecordData extends RecordData {
+  //Required by the relationship layer
+  isNew(): boolean;
+  modelName: string;
+  storeWrapper: RecordDataStoreWrapper;
+  id: string | null;
+  clientId: string | null;
+  isEmpty(): boolean;
+  getResourceIdentifier(): RecordIdentifier;
+  _relationships: Relationships;
```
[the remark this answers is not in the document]
runspired: You misunderstand, `RelationshipRecordData` is the narrowed thing :)

Not all record-data implementations will have support for relationships, indeed in their current form only our own (or those extending our own) are even possible to have support.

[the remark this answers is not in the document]
runspired: (the checks for `null` / `undefined` are checking for whether we should narrow to our own)

packages/store/addon/-private/system/core-store.ts
runspired: Making this change causes builds to throw maximum call-stack exceeded errors, and as far as we've seen babel / babel-typescript does not strip type-only imports. We may not have gotten whatever goodness you refer to yet, or it needs a config we don't have, or it just doesn't work here.

packages/store/addon/-private/system/model/internal-model.ts
runspired: yes

packages/store/addon/-private/system/model/internal-model.ts
runspired: it is safe; however, `resource` doesn't seem to receive the knowledge that the return of `getBelongsTo` has `_relationships` even with this cast, which could be due to not all files being typed yet.

it is safe for similar reasons to the `recordData as RecordData` return type you commented on earlier, basically relationship support doesn't exist for RecordDatas that don't at least extend DefaultRecordData

packages/store/addon/-private/system/model/internal-model.ts
runspired: see above

packages/store/addon/-private/ts-interfaces/ember-data-json-api.ts
runspired: correct. Typescript has many of these breakdowns when it comes to typing JSON structures. `undefined` is not actually a valid JSON value, but both `null` and `not-present` ARE

runspired: (and this is the fundamental reason why EmberData ends up with `null | undefined` throughout the system instead of just one or the other :'( )

packages/store/types/@ember/runloop/index.d.ts
runspired: I've been contemplating moving many of our private interfaces and types into a private package. Needing to know whether this would break compile is what has held me back. Publishing a package, even if technically "private" is a non-starter.

runspired: (fwiw I've figured out that it seems its ok, so this is an option to explore, but a single root tsconfig lets us put these in one place as well so there's more than one path forward)

mike-north: to be clear, I think @chriskrycho is proposing a user-defined type root (`paths: { "*": ["foo"] }`) and not a published package. Sounds like a good suggestion to me, particularly in that there are quite a few private types floating around, and a centrally-defined single place for "non-exported things necessary for ember-data to internally compile" could catch some inconsistencies between package boundaries.

packages/store/addon/-private/system/record-data-for.ts
```diff
@@ -0,0 +1,31 @@
...
+ *  "Ember.getOwner(obj)" style.
+ */
+interface InternalModel {
+  _recordData: RecordData;
+}
+
+type DSModelOrSnapshot = { _internalModel: InternalModel };
+type Reference = { internalModel: InternalModel };
+
+type Legit = InternalModel | RecordData | DSModelOrSnapshot | Reference;
```
chriskrycho: Heh, let's call it `Instance` instead of `Legit`. Otherwise, 👍

packages/store/addon/-private/system/model/internal-model.ts
```diff
@@ -37,11 +37,12 @@ import recordDataFor from '../record-data-for';
 
 type DefaultRecordData = import('@ember-data/record-data/-private').RecordData;
 type RecordArray = InstanceType<typeof RecordArray>;
+type RelationshipRecordData = import('@ember-data/record-data/-private/ts-interfaces/relationship-record-data').RelationshipRecordData;
 
 // once the presentation logic is moved into the Model package we can make
 // eliminate these lossy and redundant helpers
 function relationshipsFor(instance: any) {
-  let recordData = recordDataFor(instance) || instance;
+  let recordData = (recordDataFor(instance) || instance) as RelationshipRecordData;
```
chriskrycho: Do we have the ability to type `instance` *here*?

Also, can we add a return type annotation to constrain it (keeping it from breaking in the future)?

packages/record-data/addon/-private/ts-interfaces/relationship-record-data.ts
```diff
@@ -30,6 +30,9 @@ export interface RelationshipRecordData extends RecordData {
   _implicitRelationships: {
     [key: string]: Relationship;
   };
+  __implicitRelationships: {
```
chriskrycho: Do you have a `Dict` type you could use here? `Dict<Relationship>`?

packages/store/addon/-private/system/model/internal-model.ts
runspired: @chriskrycho I can look into it, would need to be a `import()` to get the type of a class

packages/record-data/addon/-private/ts-interfaces/relationship-record-data.ts
runspired: We do

packages/store/types/@ember/runloop/index.d.ts
chriskrycho: Yep, exactly that! Safely lets us define type-only stuff in one place with zero worry we’ll accidentally publish it as a package.

packages/-serializer-encapsulation-test-app/app/services/store.js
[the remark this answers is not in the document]
Gaurav0: This seems like a large hoop to go through in order for an application not to use the serializer package. Is there no way to detect if @ember-data/record-data is included in the app from @ember-data/store and do this automatically if it's there?

PR #6526, opened 2019-09-27
[CHORE] Configure @typescript-eslint for linting TypeScript files

HeroicEric:
This updates our ESLint configuration so that it also runs for TypeScript files using [typescript-eslint](https://github.com/typescript-eslint/typescript-eslint).

## If it is preferred, I can break this up into smaller pull requests

Details

- Adds [`@typescript-eslint/parser`](https://github.com/typescript-eslint/typescript-eslint/tree/master/packages/parser)
- Adds [`@typescript-eslint/eslint-plugin`](https://github.com/typescript-eslint/typescript-eslint/tree/master/packages/eslint-plugin)
- Disables all rules that cause linting TypeScript files to fail
- Renames `bin/lint-js` -> `bin/lint-changed-js` & `yarn lint:js` -> `yarn lint:js:changed` to make it's purpose more obvious
- Updates `bin/lint-changed-js` to include linting TypeScript files
- Replaces `yarn lint:prettier` with `yarn lint:js` which runs ESLint for JavaScript & TypeScript, including enforcing our Prettier config
- Updates all CI scripts to use `yarn lint:js` rather than `yarn lint:prettier`

Since we haven't been running eslint for TypeScript files, some files don't pass with our current eslint config. Any rules that were causing errors have been disabled temporarily so we can make any neccessary code changes enable the rules again separately.

---

The diff here is kind of large so this was broken up into many commits. The first one configures `typescript-eslint` and updates our CI scripts to use include typescript files when linting. Any rules that were failing were disabled.

The following 6 commits each enable a rule and fix any errors.

There are probably more rules that we will want to look into enabling that are provided by `typescript-eslint`. This only attempts to configure the same rules that we already use for regular JS files.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -58,7 +58,6 @@ module.exports = {
       plugins: ['@typescript-eslint'],
       extends: ['eslint:recommended', 'plugin:@typescript-eslint/eslint-recommended'],
       rules: {
-        'no-console': 'off',
         'no-empty': 'off',
         'no-extra-semi': 'off',
         'no-unused-vars': 'off',
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -48,6 +48,27 @@ module.exports = {
     node: false,
   },
   overrides: [
+    // TypeScript files
+    {
+      files: ['**/*.ts'],
+      parser: '@typescript-eslint/parser',
+      parserOptions: {
+        sourceType: 'module',
+      },
+      plugins: ['@typescript-eslint'],
+      extends: ['eslint:recommended', 'plugin:@typescript-eslint/eslint-recommended'],
+      rules: {
+        'no-console': 'off',
+        'no-empty': 'off',
+        'no-empty-pattern': 'off',
+        'no-extra-semi': 'off',
+        'no-prototype-builtins': 'off',
+        'no-unused-vars': 'off',
+        'prettier/prettier': 'off',
+        'require-atomic-updates': 'off',
+      },
+    },
+
     // node files
     {
       files: [
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -58,6 +58,7 @@ module.exports = {
       plugins: ['@typescript-eslint'],
       extends: ['eslint:recommended', 'plugin:@typescript-eslint/eslint-recommended'],
       rules: {
+        '@typescript-eslint/no-unused-vars': ['error', { args: 'none' }],
         'no-unused-vars': 'off',
         'require-atomic-updates': 'off',
       },
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -58,9 +58,7 @@ module.exports = {
       plugins: ['@typescript-eslint'],
       extends: ['eslint:recommended', 'plugin:@typescript-eslint/eslint-recommended'],
       rules: {
-        'no-extra-semi': 'off',
         'no-unused-vars': 'off',
-        'prettier/prettier': 'off',
         'require-atomic-updates': 'off',
       },
     },
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -60,7 +60,6 @@ module.exports = {
       rules: {
         'no-console': 'off',
         'no-empty': 'off',
-        'no-empty-pattern': 'off',
         'no-extra-semi': 'off',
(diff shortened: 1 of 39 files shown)


.eslintrc.js
```diff
@@ -60,15 +64,17 @@ module.exports = {
         'packages/*/ember-cli-build.js',
         'packages/*/index.js',
         'packages/*/testem.js',
-        'packages/*/bin/**',
```
HeroicEric: dont remember why this change is here. Note to self: check this again

packages/-ember-data/tests/integration/record-data/store-wrapper-test.ts
```diff
@@ -284,7 +284,7 @@ module('integration/store-wrapper - RecordData StoreWrapper tests', function(hoo
 
     class RecordDataForTest extends TestRecordData {
       id: string;
-      _isNew: boolean = false;
```
runspired: this looks like an accidental removal

HeroicEric: This is one of the rules, i don't remember which one, but I plan on adding details about each of these changes to the PR.

I think it was something about this being unnecessary

runspired: ah, it can't be inferred though :( it makes the type `false` instead of `boolean` which then errors when you set it to true

packages/-ember-data/tests/integration/record-data/record-data-state-test.ts
```diff
@@ -127,11 +125,6 @@ module('integration/record-data - Record Data State', function(hooks) {
     let { owner } = this;
 
     class LifecycleRecordData extends TestRecordData {
-      constructor(sw) {
-        super();
-        storeWrapper = sw;
```
HeroicEric: This one was a little weird. `storeWrapper` is not actually being used anywhere so this was triggering the `@typescript-eslint/no-unused-vars` rule.

packages/store/addon/-private/system/core-store.ts
```diff
@@ -3620,88 +3618,6 @@ function _commit(adapter, store, operation, snapshot) {
...
 }
 
-/**
- *
- * @param store
- * @param cache modelFactoryCache
- * @param normalizedModelName already normalized modelName
- * @return {*}
- */
-function getModelFactory(store, cache, normalizedModelName) {
```
HeroicEric: This function and the others that were deleted in this file are not used at all so they were caught by `@typescript-eslint/no-unused-vars`.

packages/store/addon/-private/ts-interfaces/record-data-record-wrapper.ts
```diff
@@ -39,7 +38,7 @@ export default interface RecordDataRecordWrapper {
   /**
    * @deprecated
    */
-  getErrors?({}): JsonApiValidationError[];
+  getErrors?({}): JsonApiValidationError[]; // eslint-disable-line no-empty-pattern
```
HeroicEric: I think this is actually a case where we're seeing false positives because the rule was originally not implemented with TypeScript in mind, so eslint is interpreting the type definition as an attempt to destructure.

I will open an issue with @typescript-eslint

packages/store/types/@ember/array/index.d.ts
```diff
@@ -1,7 +1,9 @@
 import EmberArray, { A } from '@ember/array';
+
 namespace EmberArray {
-    function detect(arr: any): boolean;
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
```
HeroicEric: I think this is maybe a bug with `@typescript-eslint` but I'm not sure how namespaces work.

packages/-ember-data/tests/integration/record-data/store-wrapper-test.ts
HeroicEric: These changes were removed

packages/store/addon/-private/system/core-store.ts
runspired: Another PR also deletes them, happy to rebase it though :D

PR #6528, opened 2019-09-27
[FIX] only include version in ember-data output

### LINKED ISSUE #6219: `shouldRollupPrivate: false` produces a broken build

runspired:
Previously we were adding version to the tree in every package (whoops!)

This also ensures that version is always included in the output regardless of whether shouldRollupPrivate is used.

Resolves #6219

### DIFF
--- a/packages/-build-infra/src/addon-build-config-for-data-package.js
+++ b/packages/-build-infra/src/addon-build-config-for-data-package.js
@@ -2,7 +2,6 @@ const calculateCacheKeyForTree = require('calculate-cache-key-for-tree');
 const Funnel = require('broccoli-funnel');
 const merge = require('broccoli-merge-trees');
 const BroccoliDebug = require('broccoli-debug');
-const version = require('./create-version-module');
 const { isInstrumentedBuild } = require('./cli-flags');
 const rollupPrivateModule = require('./utilities/rollup-private-module');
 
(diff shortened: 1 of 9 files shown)


packages/-build-infra/src/addon-build-config-for-data-package.js
```diff
@@ -111,6 +110,9 @@ function addonBuildConfigForDataPackage(PackageName) {
     },
 
     treeForAddon(tree) {
+      if (typeof this._additionalAddonFiles === 'function') {
```
rwjblue: If this is only needed by `ember-data` package, can we put a custom `treeForAddon` there instead and merge in the file after calling the original implementation here?

runspired: since we don't have class based inheritance, no :(

rwjblue: We can definitely do it

```js
module.exports = Object.assign(addonBaseConfig, {
 // other stuff here

  treeForAddon() {
    let tree = addonBaseConfig.treeForAddon.apply(this, arguments);
    return merge([tree,version()]);
  }
});
```

runspired: 👍

Issue #6460, opened 2019-09-17
[SIDE QUEST] MinimumSerializerInterface Tests


runspired:
# A Side-Quest for Project Trim!

Side Quests?!?! How fun!

## Mission

#6459 introduced test infrastructure for testing ember-data without the `@ember-data/serializer` package being present at all while #6451 introduced the MinimumSerializerInterface. In this repository we want to add integration tests covering three areas.

- requests
- mutation
- serialization

The following things should be tested:

**Requests**

- [x] `store.<findAll|findRecord|query|queryRecord>` call `serializer.normalizeResponse` with the store methodName as the `requestType` once the adapter resolves
- [x] accessing an async hasMany relationship without links results in `serializer.normalizeResponse` being called with the requestType `findMany`
- [x] accessing an async hasMany relationship with links results in `serializer.normalizeResponse` being called with the requestType `findHasMany`
- [x] accessing an async belongsTo relationship with links results in `serializer.normalizeResponse` being called with the requestType `findBelongsTo`
- [x] errors in JSON:API format "just work" with no `extractErrors` hook on the serializer

**Mutation**

- [x] normalizeResponse deleteRecord
- [x] serialize updateRecord
- [x] serializeIntoHash if present updateRecord
- [x] normalizeResponse updateRecord
- [x] serializeIntoHash if present createRecord
- [x] serialize createRecord
- [x] normalizeResponse createRecord

**Serialization/Normalization**

- [x] record.serialize
- [x] store.serialize
- [x] snapshot.serialize
- [x] store.normalize
- [x] store.pushPayload calls serializer.pushPayload

runspired:
@Gaurav0 is going to work to complete this side-quest :)

Gaurav0:
I've interpreted record.serialize as Model.serialize, but there doesn't seem to be a store.serialize?

Gaurav0:
Side quest completed solo.

runspired:
And much appreciation for it! Great work :D I'm very happy to have all the additional test coverage this introduced :)

P.S. I'll be speccing out the tests for Adapters this week

PR #6530, opened 2019-09-29
[TEST] deleteRecord test for serializer encapsulation side quest

Gaurav0:
For #6460

### DIFF
--- a/packages/-serializer-encapsulation-test-app/tests/integration/delete-record-test.js
+++ b/packages/-serializer-encapsulation-test-app/tests/integration/delete-record-test.js
@@ -1,0 +1,251 @@
+import Model, { attr } from '@ember-data/model';
+import Store from 'serializer-encapsulation-test-app/services/store';
+import JSONAPIAdapter from '@ember-data/adapter/json-api';
+import EmberObject from '@ember/object';
+import { setupTest } from 'ember-qunit';
+import { module, test } from 'qunit';
+import { resolve } from 'rsvp';
+
+class Person extends Model {
+  @attr
+  firstName;
+
+  @attr
+  lastName;
+
+  // override to not call serialize()
+  toJSON() {
+    const { id, firstName, lastName } = this;
+    return {
+      id,
+      type: this.constructor.modelName,
+      attributes: {
+        firstName,
+        lastName,
+      },
+    };
+  }
+}
+
+module('inteegration/delete-record - running deleteRecord with minimum serializer', function(hooks) {
+  setupTest(hooks);
+
+  hooks.beforeEach(function(assert) {
+    this.owner.register('service:store', Store);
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -24,6 +24,7 @@
     "test:try-one": "yarn workspace ember-data test:try-one",
     "test:docs": "yarn workspace ember-data test:docs",
     "test:encapsulation": "lerna run --scope='*-encapsulation-test-app' test",
+    "test:encapsulation:serializer": "yarn workspace serializer-encapsulation-test-app test",
```
runspired: for local dev you can run `ember test` in the directory directly. For CI the `test:encapsulation` command already runs this. Let's remove this.

packages/-serializer-encapsulation-test-app/tests/integration/delete-record-test.js
```diff
@@ -0,0 +1,187 @@
...
+    let _payloads = [
+      {
+        id: '1',
+        type: 'person',
+        attributes: {
+          firstName: 'John',
+          lastName: 'Smith',
+        },
+      },
+      '',
```
runspired: if we leave a payload here for the `deleteRecord` response we do in fact call `normalizeResponse`. Let's add a test for that as well.

I added an issue for discussing the fact we don't always call `normalizeResponse` here: https://github.com/emberjs/data/issues/6541

I was not aware of this and it seems a bit bonkers.

Gaurav0: I am confused about how it makes any sort of sense to have a response to a DELETE request and why we should do anything with it if there is. JSON-API I believe requires a 204 No Response.

package.json
Gaurav0: done.

packages/-serializer-encapsulation-test-app/tests/integration/delete-record-test.js
Gaurav0: Added additional test.

packages/-serializer-encapsulation-test-app/tests/integration/delete-record-test.js
```diff
@@ -0,0 +1,251 @@
...
+      type: this.constructor.modelName,
+      attributes: {
+        firstName,
+        lastName,
+      },
+    };
+  }
+}
+
+module('inteegration/delete-record - running deleteRecord with minimum serializer', function(hooks) {
```
runspired:
```suggestion
module('integration/delete-record - running deleteRecord with minimum serializer', function(hooks) {
```

runspired: I've fixed this spelling error in most the already merged tests, suspect you are copying from one we haven't fixed yet

PR #6533, opened 2019-09-30
[CHORE] Cleanup ESLint config

HeroicEric:
Cleans up our ESLint config to be more close to the ember-cli addon blueprints.

- Uses more consistent ignore pattern style
- Removes some unnecessary ESLint rule declarations that are already set via the latest `eslint:recommended` in 6.x.x
- Re-enables some rules for linting bin scripts that no longer need to be disabled

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,26 +1,26 @@
 # unconventional js
-packages/*/blueprints/*/*files/
-vendor/
+**/blueprints/*/*files/
+**/vendor/
 
 # compiled output
+**/dist/
+**/tmp/
 /packages/-ember-data/docs/
-dist
-tmp
 
 # dependencies
-bower_components
-node_modules
+**/bower_components/
+**/node_modules/
 
 # misc
-/coverage/
+**/coverage/
 !.*
 /.yarn/
-.git
+/.git/
 
 # ember-try
-.node_modules.ember-try
-bower.json.ember-try
-package.json.ember-try
+**/.node_modules.ember-try/
+**/bower.json.ember-try
+**/package.json.ember-try
 
 # ember-data
-packages/*/node-tests/fixtures
+**/node-tests/fixtures/
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -10,26 +10,12 @@ module.exports = {
   rules: {
     'mocha/no-exclusive-tests': 'error',
     'prettier/prettier': 'error',
-
-    'no-unused-vars': [
-      'error',
-      {
-        args: 'none',
-      },
-    ],
-
+    'no-unused-vars': ['error', { args: 'none' }],
     'no-cond-assign': ['error', 'except-parens'],
     eqeqeq: 'error',
     'no-eval': 'error',
-    'new-cap': [
-      'error',
-      {
-        capIsNew: false,
-      },
-    ],
+    'new-cap': ['error', { capIsNew: false }],
     'no-caller': 'error',
-    'no-irregular-whitespace': 'error',
-    'no-undef': 'error',
     'no-eq-null': 'error',
     'no-console': 'error', // no longer recommended in eslint v6, this restores it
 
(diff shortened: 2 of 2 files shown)


.eslintrc.js
```diff
@@ -10,26 +10,12 @@ module.exports = {
...
     'no-eval': 'error',
-    'new-cap': [
-      'error',
-      {
-        capIsNew: false,
-      },
-    ],
+    'new-cap': ['error', { capIsNew: false }],
     'no-caller': 'error',
-    'no-irregular-whitespace': 'error',
```
HeroicEric: This is already set via `eslint:recommended` in ESLint 6.x.

.eslintrc.js
```diff
@@ -10,26 +10,12 @@ module.exports = {
...
-    'no-undef': 'error',
```
HeroicEric: This is already set via `eslint:recommended` in ESLint 6.x.

.eslintrc.js
```diff
@@ -54,22 +40,26 @@ module.exports = {
...
         'bin/*',
         'packages/-build-infra/src/**/*.js',
         'packages/-test-infra/src/**/*.js',
+        'packages/*/.ember-cli.js',
+        'packages/*/.eslintrc.js',
+        'packages/*/.template-lintrc.js',
         'packages/*/ember-cli-build.js',
         'packages/*/index.js',
         'packages/*/testem.js',
-        'packages/*/bin/**',
```
HeroicEric: this directory doesn't exist. All bin scripts are in the root `bin` directory.

.eslintrc.js
```diff
@@ -54,22 +40,26 @@ module.exports = {
         '.mocharc.js',
         '.eslintrc.js',
         '.prettierrc.js',
-        '.template-lintrc.js',
```
HeroicEric: This file does not exist. All template lint files are in packages directories.

.eslintignore
```diff
@@ -1,26 +1,26 @@
...
+/packages/*/tmp/
 /packages/-ember-data/docs/
-dist
-tmp
 
 # dependencies
-bower_components
-node_modules
+/packages/*/bower_components/
+/packages/*/node_modules/
```
Gaurav0: node_modules is usually created in the root directory.

.eslintrc.js
Gaurav0: Without this, `yarn workspace <packagename> lint:js` fails

.eslintignore
HeroicEric: Nice catch!

.eslintrc.js
HeroicEric: @Gaurav0 do you mind expanding on this?

I just tested the command you mentioned with both of the workspaces that have a `lint:js` defined and it seemed to be fine:

These are the commands I ran (on this branch):

```bash
yarn workspace adapter-encapsulation-test-app lint:js
yarn workspace debug-encapsulation-test-app lint:js
```

.eslintignore
HeroicEric: Updated pattern to include `node_modules` and `bower_components` in the root directory

HeroicEric:
@Gaurav0 you mentioned:

> Please note that the root .eslintrc.js is used for all packages as well. I made that happen to speed up eslint, otherwise running it on root was processing many configuration files.

Do you mean that this is going to undo that work?

.eslintrc.js
Gaurav0: You are correct, it is working on your branch. You may have fixed something, or I may have gotten the error another way. Feel free to ignore me for now.

Gaurav0:
I don't think so. I just wanted to ensure you were testing if `yarn lint:js` succeeds on all packages it is defined in.

.eslintignore
```diff
@@ -1,26 +1,26 @@
...
+**/tmp/
 /packages/-ember-data/docs/
-dist
-tmp
 
 # dependencies
-bower_components
-node_modules
+**/bower_components/
+**/node_modules/
```
runspired: this would not allow for ignore root level

.eslintignore
```diff
@@ -1,26 +1,26 @@
...
 
 # misc
-/coverage/
+**/coverage/
```
runspired: I'm not sure where the coverage report is generated since I don't think we've done one since moving to packages. It may still be root

.eslintignore
HeroicEric: Are you sure about that? AFAICT the `.eslintignore` file follows the same semantics as `.gitignore`. This means to ignore any directories named `node_modules` anywhere in the repo, including the root.

From https://git-scm.com/docs/gitignore#_pattern_format:

> A leading `"**"` followed by a slash means match in all directories. For example, `"**/foo"` matches file or directory `"foo"` anywhere, the same as pattern `"foo"`. `"**/foo/bar"` matches file or directory `"bar"` anywhere that is directly under directory `"foo"`.

You can verify this if you make the same changes to `.gitignore` and either check to see that the root `node_modules` stays ignored or by running [`git check-ignore *`](https://git-scm.com/docs/git-check-ignore).

runspired: Yeah reading glob rules seems I am mistaken here, although I swear this has not worked for me before :/

PR #6547, opened 2019-10-02
[CHORE ci] Remove Azure Pipelines configuration

HeroicEric:
Related to https://github.com/emberjs/data/issues/6282

This removes our configuration for Azure Pipelines.

We recently added GitHub Actions for CI in https://github.com/emberjs/data/pull/6472

We decided to let Azure Pipelines run alongside the GitHub Actions workflow for a short period of time to give the new GitHub workflow some time to bake.


runspired:
Overall I am +1 on this. I've seen a few issues with gh-actions where I still trust azure a bit more, but not enough to want to keep both and the future clearly seems to be gh-actions.

PR #6549, opened 2019-10-02
[CHORE] Removing detect function (and dependence on model package) in @debug 

pete-the-pete:
Removes the use of Model and `detect` in the @debug/data-adapter.


packages/debug/addon/index.js
```diff
@@ -38,15 +38,72 @@ export default DataAdapter.extend({
...
+    const _releaseMethods = [];
+    const discoveredTypes = StoreTypesMap.get(store);
+
+    // Add any models that were added during initialization of the app, before the inspector was opened
+    discoveredTypes.forEach((_, type) => {
+      this.watchTypeIfUnseen(store, discoveredTypes, type, typesAdded, typesUpdated, _releaseMethods);
+    });
+
+    // Overwrite _createRecordData so newly added models will get added to the list
+    store._createRecordData = function(identifier) {
```
runspired: this should be an arrow function to preserve the `this` context of the data adapter
```suggestion
    store._createRecordData = (identifier) => {
```

Issue #5958, opened 2019-04-01
Calling findRecord without an id results in confusing error state in production builds


paddyobrien:
Calling `store.findRecord('foo', undefined)` fails in a dev build with a very helpful error message:

```
Error: Assertion Failed: `id` passed to `findRecord()` has to be non-empty string or number
```

However doing the same thing in a production build does not fail and instead issues a request as if you had called `store.findAll('foo')`. It then resolves the promise with a model in a weird state where all fields are undefined:

![Screenshot 2019-04-01 at 17 22 14](https://user-images.githubusercontent.com/846372/55343299-b6cf8800-54a2-11e9-960c-212b449fa78e.png)

This probably happens because the argument validation is implemented using `Ember.assert` [here](https://github.com/emberjs/data/blob/master/addon/-private/system/store.js#L752-L755) and these assertions are stripped from production builds.

This behaviour can result in an extremely confusing and hard to understand situation where a model hook resolves but the model is in a completely broken state. It would be preferable in these cases for the model hook to just fail with the same or similar exception as we see in dev.

snewcomer:
I find the debug scenario helpful but in client code overall, if I am unsure of the type of identifier will be passed, I assumed this would be left up to the client to guard.  What are your thoughts?

travisghansen:
All things considered, the divergent behavior here is extremely bad form IMO.

paddyobrien:
Looks like this got fixed in https://github.com/emberjs/data/pull/6561

PR #6561, opened 2019-10-04
[BUGFIX] findRecord should not resolve without id in production builds

Gaurav0:
For #5958


paddyobrien:
As a person who got burned by this I have a few thoughts:

1. Not everyone uses typescript.
2. Even if you do use typescript it may not save you from this problem (specifically in our case the problem came about at run time by reading a bad value out of localStorage)
3. The problem for me wasn't so much the lack of validation as the inconsistent validation. It took me way longer to understand what was happening because a dev build threw a helpful exception while production build just did completely a completely unexpected thing.

travisghansen:
I'm with @paddyobrien here. The different behavior based on environment in this context is bad form IMHO.

PR #6562, opened 2019-10-05
[BUGFIX fetchOptions] fix fetchOptions to support POST body of all valid types

nlfurniss:
Landed some fixes to `ember-fetch` in [PR #174](https://github.com/ember-cli/ember-fetch/pull/174) to correctly support POST `body`s other than objects. These fixes came after these utils were added to ED.

Also added several tests from ember-fetch to increase coverage for this util, some of which pertain to this change.

### DIFF
--- a/packages/-ember-data/tests/unit/adapters/rest-adapter/fetch-options-test.js
+++ b/packages/-ember-data/tests/unit/adapters/rest-adapter/fetch-options-test.js
@@ -1,0 +1,148 @@
+import { module, test } from 'qunit';
+import { fetchOptions } from '@ember-data/adapter/rest';
+
+module('unit/adapters/rest-adapter/fetch-options', function(hooks) {
+  test("fetchOptions removes undefined query params when method is POST and 'data' is an object", function(assert) {
+    assert.expect(1);
+
+    const dataAsObject = {
+      a: 1,
+      b: undefined,
+      c: 3,
+      d: null,
+      e: 0,
+      f: false,
+    };
+
+    const undefinedQueryStringOptions = {
+      url: 'https://emberjs.com',
+      method: 'POST',
+      data: dataAsObject,
+    };
+
+    let options = fetchOptions(undefinedQueryStringOptions);
+    assert.deepEqual(options.body, '{"a":1,"c":3,"d":null,"e":0,"f":false}');
+  });
+
+  test('fetchOptions sets the request body correctly when the method is not GET or HEAD', function(assert) {
+    assert.expect(3);
+
+    const baseOptions = {
+      url: '/',
+      method: 'POST',
+      data: { a: 1 },
+    };
+
+    // Tests POST method.
+    let options = fetchOptions(baseOptions);
+    assert.equal(options.body, JSON.stringify(baseOptions.data), 'POST request body correctly set');
+
+    // Tests PUT method.
+    baseOptions.method = 'PUT';
+    options = fetchOptions(baseOptions);
+    assert.equal(options.body, JSON.stringify(baseOptions.data), 'PUT request body correctly set');
+
+    // Tests DELETE method.
+    baseOptions.method = 'DELETE';
+    options = fetchOptions(baseOptions);
+    assert.equal(options.body, JSON.stringify(baseOptions.data), 'DELETE request has the correct body');
+  });
+
+  test("fetchOptions sets the request body correctly when the method is POST and 'data' is a string", function(assert) {
+    assert.expect(1);
+
+    // Tests stringified objects.
+    const stringifiedData = JSON.stringify({ a: 1, b: 2 });
+    const optionsWithStringData = {
+      url: 'https://emberjs.com',
+      method: 'POST',
+      data: stringifiedData,
+    };
+
+    let options = fetchOptions(optionsWithStringData);
+    assert.equal(options.body, stringifiedData);
+  });
+
+  test('fetchOptions does not set a request body when the method is GET or HEAD', function(assert) {
+    assert.expect(4);
+
+    const baseOptions = {
+      url: '/',
+      method: 'GET',
+      data: { a: 1 },
+    };
+
+    let options = fetchOptions(baseOptions);
+    assert.strictEqual(options.body, undefined, 'GET request does not have a request body');
+
+    baseOptions.method = 'HEAD';
+    options = fetchOptions(baseOptions);
+    assert.strictEqual(options.body, undefined, 'HEAD request does not have a request body');
+
+    baseOptions.data = {};
+    options = fetchOptions(baseOptions);
+    assert.strictEqual(
+      options.body,
+      undefined,
+      'HEAD request does not have a request body when `data` is an empty object'
+    );
+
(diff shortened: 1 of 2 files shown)


packages/adapter/addon/rest.js
```diff
@@ -1363,7 +1363,13 @@ export function fetchOptions(options, adapter) {
     } else {
       // NOTE: a request's body cannot be an object, so we stringify it if it is.
       // JSON.stringify removes keys with values of `undefined` (mimics jQuery.ajax).
-      options.body = JSON.stringify(options.data);
+      // If the data is not a POJO (it's a String, FormData, etc), we just set it.
+      // If the data is a string, we assume it's a stringified object.
+      if (Object.prototype.toString.call(options.data) === '[object Object]') {
```
rwjblue: @nlfurniss - I'm curious, what cases does `Object.prototype.toString` cover that a `typeof` wouldn't? Would you mind updating this comment to explain that (and confirming that some of the new tests added specifically test that difference)?

packages/-ember-data/tests/unit/adapters/rest-adapter/fetch-options-test.js
```diff
@@ -0,0 +1,143 @@
...
+    options = fetchOptions(baseOptions);
+    assert.equal(options.body, undefined, 'HEAD request does not have a request body');
+
+    baseOptions.data = {};
+    options = fetchOptions(baseOptions);
+    assert.equal(options.body, undefined, 'HEAD request does not have a request body when `data` is an empty object');
+
+    baseOptions.method = 'GET';
+    options = fetchOptions(baseOptions);
+    assert.equal(options.body, undefined, 'GET request does not have a request body when `data` is an empty object');
```
rwjblue:
```suggestion
    assert.strictEqual(options.body, undefined, 'GET request does not have a request body when `data` is an empty object');
```

packages/-ember-data/tests/unit/adapters/rest-adapter/fetch-options-test.js
```diff
@@ -0,0 +1,143 @@
...
+    let options = fetchOptions(baseOptions);
+    assert.equal(options.body, undefined, 'GET request does not have a request body');
+
+    baseOptions.method = 'HEAD';
+    options = fetchOptions(baseOptions);
+    assert.equal(options.body, undefined, 'HEAD request does not have a request body');
+
+    baseOptions.data = {};
+    options = fetchOptions(baseOptions);
+    assert.equal(options.body, undefined, 'HEAD request does not have a request body when `data` is an empty object');
```
rwjblue:
```suggestion
    assert.strictEqual(options.body, undefined, 'HEAD request does not have a request body when `data` is an empty object');
```

packages/-ember-data/tests/unit/adapters/rest-adapter/fetch-options-test.js
```diff
@@ -0,0 +1,143 @@
...
+      method: 'GET',
+      data: { a: 1 },
+    };
+
+    let options = fetchOptions(baseOptions);
+    assert.equal(options.body, undefined, 'GET request does not have a request body');
+
+    baseOptions.method = 'HEAD';
+    options = fetchOptions(baseOptions);
+    assert.equal(options.body, undefined, 'HEAD request does not have a request body');
```
rwjblue:
```suggestion
    assert.strictEqual(options.body, undefined, 'HEAD request does not have a request body');
```

packages/-ember-data/tests/unit/adapters/rest-adapter/fetch-options-test.js
```diff
@@ -0,0 +1,143 @@
...
+    assert.expect(4);
+
+    const baseOptions = {
+      url: '/',
+      method: 'GET',
+      data: { a: 1 },
+    };
+
+    let options = fetchOptions(baseOptions);
+    assert.equal(options.body, undefined, 'GET request does not have a request body');
```
rwjblue:
```suggestion
    assert.strictEqual(options.body, undefined, 'GET request does not have a request body');
```

packages/-ember-data/tests/unit/adapters/rest-adapter/fetch-options-test.js
```diff
@@ -0,0 +1,143 @@
...
+
+    const formData = new FormData();
+    const postData = {
+      url: 'https://emberjs.com',
+      method: 'POST',
+      data: formData,
+    };
+
+    const postOptions = fetchOptions(postData);
+    assert.equal(postOptions.body, formData, "'options.body' is the FormData passed in");
```
rwjblue:
```suggestion
    assert.strictEqual(postOptions.body, formData, "'options.body' is the FormData passed in");
```

packages/-ember-data/tests/unit/adapters/rest-adapter/fetch-options-test.js
```diff
@@ -0,0 +1,143 @@
...
+    // Tests stringified objects.
+    const stringifiedData = JSON.stringify({ a: 1, b: 2 });
+    const optionsWithStringData = {
+      url: 'https://emberjs.com',
+      method: 'POST',
+      data: stringifiedData,
+    };
+
+    let options = fetchOptions(optionsWithStringData);
+    assert.deepEqual(options.body, stringifiedData);
```
rwjblue: Hmm, `deepEqual` seems odd here, we are comparing a string to a string, can we use:

```suggestion
    assert.equal(options.body, stringifiedData);
```

packages/-ember-data/tests/unit/adapters/rest-adapter/fetch-options-test.js
```diff
@@ -0,0 +1,143 @@
+import { setupTest } from 'ember-qunit';
+import { module, test } from 'qunit';
+import { fetchOptions } from '@ember-data/adapter/rest';
+
+module('unit/adapters/rest-adapter/fetch-options', function(hooks) {
+  setupTest(hooks);
```
rwjblue: I _think_ we can remove this part? It seems like all the tests are using the utility function directly, and don't need any additional test setup. Did I miss something?

```suggestion
```

packages/adapter/addon/rest.js
nlfurniss: Originally it was to distinguish between an Object and an Array or other type that returns "object" from `typeof. But maybe we don't need to check that

nlfurniss: Because it doesn't catch the data structures (like FormData) that return `object` from `typeof`. Will add comment.

The test for FormData will fail if someone changes it to `typeof`

runspired: it feels like `instanceof FormData` and `typeof x === 'string'` could be used to have a better guarantee here

nlfurniss: But its not just FormData that could be passed in, and it seemed like this was a lot nicer way of doing it than a large switch statement that may expand in the future

nlfurniss:
@runspired any other changes you'd like or questions I can answer?

Issue #6573, opened 2019-10-08
[TRACKING] Relationship Refactor


runspired:
Tracking Issue for Improvements to the Performance, Reliability, and Encapsulation of the Relationship Layer.

runspired:
Parallel Efforts that assist this effort (the first three are already tasks for Project Trim)

- [ ] The Singleton RecordData Infrastructure
- [ ] RecordData Documentation
- [ ] A Refactor moving as much relationship UI logic out of internalModel/the store into the Model package as possible
- [ ] A Singleton RecordData Implementation for EmberData

runspired:
cc @sly7-7 This is an effort I'm pairing with @skaterdav85 on to improve relationships that should help you greatly :) If you have the ability to contribute to it to help resolve your performance issues with relationships that would be awesome!

skaterdav85:
👋

runspired:
I removed the Diff util in #8550 which we may end up wanting later. It was currently unused so seemed better to drop for now.

runspired:
The main change I want to see before marking the API as complete is a story around pagination. I began a spike for it in #9320.

The largest unanswered question is what the underlying page storage mechanism should be, and how (with it) to account for the potential for someone to fetch an arbitrary page of the relationship (where arbitrary means that a gap in pages exists between that page and any pages connected to the first or last page).

Pagination is so varied it is necessarily hard to support arbitrary pages (whereas it is relatively easy to support first/prev/next/last as pointers for iteration of a linked list).

Potential data structures include:

- a LinkedList
- a dictionary, holey array keyed by a signaled index
- a Map keyed by a signaled index
- something that keys by url but accounts for indexing

PR #6574, opened 2019-10-08
[CHORE] - Removes allInverseRecordsAreLoaded from has-many and belongs-to state classes

skaterdav85:
Eliminate usage of `recordData` in `has-many` and `belongs-to` state classes by moving `allInverseRecordsAreLoaded` logic to core-store.

https://github.com/emberjs/data/issues/6573

### DIFF
--- a/packages/store/addon/-private/system/core-store.ts
+++ b/packages/store/addon/-private/system/core-store.ts
@@ -3673,8 +3673,8 @@ function areAllInverseRecordsLoaded(store: CoreStore, resource: JsonApiRelations
   if (Array.isArray(resource.data)) {
     // treat as collection
     // check for unloaded records
-    let hasEmptyRecords = resource.data.reduce((hasEmptyModel, i) => {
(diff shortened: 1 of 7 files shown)


packages/store/addon/-private/system/core-store.ts
```diff
@@ -3655,3 +3657,43 @@ if (DEBUG) {
...
+ *
+ * @return {boolean}
+ */
+function areAllInverseRecordsLoaded(store: CoreStore, resource: JsonApiRelationship): boolean {
+  const cache = identifierCacheFor(store);
+
+  if (Array.isArray(resource.data)) {
+    // treat as collection
+    // check for unloaded records
+    let hasEmptyRecords = resource.data.reduce((hasEmptyModel, i) => {
```
skaterdav85: @runspired should we rename `i` to `resourceIdentifier`?

runspired: seems like a good rename

runspired:
There seems to be some instability in CI. Looking into whether we can get a more stable testem=>chrome connection for github actions.

PR #6599, opened 2019-10-16
Using a config to include ember-data/debug in the build

pete-the-pete:
Part of #6166 , the debug package is included based on a config setting in `environment.js`.

This relies on the `isEnabled` hook of the debug package's index.js. I manually ran yarn build in the following scenarios with the following outcomes:

1. in debug package -> not included
2. in the main addon -> included
3. in the main added with `ENV.includeDebug = false;` in environment.js -> not included

I used the environment.js settings to hold the potential config value, but this is mostly a POC, please let me know if there is a better location, name, etc.

### DIFF
--- a/packages/debug/index.js
+++ b/packages/debug/index.js
@@ -4,9 +4,51 @@ const name = require('./package').name;
 const addonBuildConfigForDataPackage = require('@ember-data/private-build-infra/src/addon-build-config-for-data-package');
 const addonBaseConfig = addonBuildConfigForDataPackage(name);
 
+function getApp(addon) {
+  while (addon && !addon.app) {
+    addon = addon.parent;
+  }
+  if (!addon) {
+    throw new Error(`Unable to find the parent application`);
+  }
+  return addon.app;
+}
+
 module.exports = Object.assign({}, addonBaseConfig, {
   shouldRollupPrivate: false,
+  __isEnabled: null,
   externalDependenciesForPrivateModule() {
     return [];
+  },
+  treeFor() {
+    // Nested addons don't call isEnabled automatically,
+    // So this ensures that we return empty trees whenever
+    // we are not enabled.
+    if (this.isEnabled()) {
+      return this._super.treeFor.call(this, ...arguments);
+    }
+  },
+  isEnabled() {
+    if (this.__isEnabled !== null) {
+      return this.__isEnabled;
+    }
+    const options = this.setupOptions();
+    const env = getApp(this).env;
+
+    this.__isEnabled = env !== 'production' || options.includeDataAdapterInProduction === true;
+
+    return this.__isEnabled;
+  },
+  setupOptions() {
+    const app = getApp(this);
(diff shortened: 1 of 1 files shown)


pete-the-pete:
@runspired first pass.

packages/debug/index.js
```diff
@@ -9,4 +9,19 @@ module.exports = Object.assign({}, addonBaseConfig, {
...
     return [];
   },
+  isEnabled() {
+    const { hasOwnProperty } = Object.prototype;
+    const config = this.parent.config(this.app.env);
+    const userDefinedProperty = hasOwnProperty.call(config, 'includeDebug');
+    if (userDefinedProperty) {
+      // use the user defined property from environment.js
+      return config.includeDebug;
+    } else if (config.APP.name && config.APP.name === 'ember-data') {
```
runspired: `config.App.name` I don't think is what we want to check here, what we want to check is if the parent addon is ember-data. Potentially we could have the `ember-data` package set a different default config variable and the debug package wouldn't need to know about the config switch. For this approach I think the `parent` property should have this info. cc @rwjblue for thoughts.

packages/debug/index.js
```diff
@@ -9,4 +9,19 @@ module.exports = Object.assign({}, addonBaseConfig, {
   externalDependenciesForPrivateModule() {
     return [];
   },
+  isEnabled() {
+    const { hasOwnProperty } = Object.prototype;
+    const config = this.parent.config(this.app.env);
+    const userDefinedProperty = hasOwnProperty.call(config, 'includeDebug');
+    if (userDefinedProperty) {
+      // use the user defined property from environment.js
+      return config.includeDebug;
```
runspired: I'd default to using an `option` for this, e.g. in `ember-cli-build` someone would specify:

```js
new EmberApp(defaults, {
  emberData: {
    includeDataAdapterInProduction: false
  }
})
```

Then we would check options for this in `index.js` of `debug` package:

```js
config() {
    let optionFlag =
      this.app &&
      this.app.options &&
      this.app.options.emberData &&
      this.app.options.emberData.includeDataAdapterInProduction;

    return {
      emberData: {
        enableRecordDataRFCBuild: optionFlag !== undefined ? optionFlag : false,
      },
    };
  }
```

This also gives the `ember-data` package a way to alter the default by setting this option:

```js
setupOptions() {
  if (this.app) {
    let options = this.app.options = this.app.options || {};
    options.emberData = options.emberData || {};
    if (options.emberData.includeDataAdapterInProduction === undefined) {
       options.emberData.includeDataAdapterInProduction = true;
    } 
  }
```

`ember-data` would need to call this during the included hook I believe, which will be called prior to the `isEnabled` hook in the `debug` package requesting the config.

pete-the-pete: This sounds good, but I'm confused about how the ember-cli-build options actually works. The comments in the file https://github.com/emberjs/data/blob/master/packages/-ember-data/ember-cli-build.js#L18 say it is only for the dummy app...I also tried putting code in there but never found the right command or location to run the commands to have it run...I'll keep playing around.

PR #6600, opened 2019-10-16
[TEST] - failing test to push records with async hasMany rel. after they have been destroyed on client-side

eodb:
This is a failing qunit test (named ''Records with an async hasMany can be pushed again after they were destroyed on client side'), to illustrate that the async hasMany relationships of a record are not as one would expect when the record(s) are pushed a 2nd time from the server (we are using web sockets) after they were previously destroyed on the client side.

In our application, records tend to too large so we have to destroy the child records when leaving some views in order to reduce the memory footprint of the browser. Next time when entering the same view, the records are pushed again from the server. Other records are long-lived and are reused in other views.

Starting from ember-data v3.13.1, we noticed that whenever the destroyed records where pushed again, one of the async hasMany collections is ever growing. The ids of the records within the collection are identical and accumulating on every push. 

This qunit test within the 'delete-record-test' can be used to easily reproduce/test the issue:
1 record ('Company') is a long lived record, the other two records ('Group' and 'Employee') refer to one-other using async hasMany relations. Group and Employee are destroyed and pushed once more. The Group ends up with 2 (identical) employees instead of 1.


eodb:
@runspired Thanks for the integration of #7109. This also fixes this unit test. I suggest to integrate it as a test to restore the async relationships after some child records of a long-living record have been deleted and are restored on a next push. A common scenario in our products to free up memory. I rebased the pull request with master, so I hope Integration is straight forward.

runspired:
closing in favor of #7258 or a follow up after #7475

PR #6606, opened 2019-10-20
[FEAT links] ensures full links object support for relationships

runspired:
In [RFC 332](https://github.com/emberjs/rfcs/blob/master/text/0332-ember-data-record-links-and-meta.md) we assumed that the full `links` object was supported for relationships. However, our implementation restricted access to just the `related` link and discarded all other links.

This PR

- Preserves all existing behaviors attached to the `related` link
- Adds `reference.links()` for accessing links
- Adds `HasMany.links` and `PromiseHasMany.links` for accessing links
- Ensures that we return links from the cache in whichever format the user supplied us (`{ href }` or `string`).

While the PR is non-breaking it does introduce some new APIs. I would argue given past promises in the linked RFC and around links more generally that this is "API backfill"  that does not require a new RFC, but I am open to authoring a short RFC for these things.

### DIFF
--- a/packages/-ember-data/tests/integration/adapter/record-persistence-test.js
+++ b/packages/-ember-data/tests/integration/adapter/record-persistence-test.js
@@ -28,7 +28,7 @@ module('integration/adapter/record_persistence - Persisting Records', function(h
     this.owner.register('serializer:application', JSONAPISerializer.extend());
   });
 
-  test("When a store is committed, the adapter's `commit` method should be called with records that have been changed.", function(assert) {
+  test("When a store is committed, the adapter's `updateRecord` method should be called with records that have been changed.", async function(assert) {
     assert.expect(2);
 
     let store = this.owner.lookup('service:store');
@@ -42,27 +42,18 @@ module('integration/adapter/record_persistence - Persisting Records', function(h
       return run(RSVP, 'resolve');
     };
 
-    run(() => {
-      store.push({
-        data: {
-          type: 'person',
-          id: '1',
-          attributes: {
-            name: 'Braaaahm Dale',
-          },
+    const tom = store.push({
+      data: {
+        type: 'person',
+        id: '1',
+        attributes: {
+          name: 'Braaaahm Dale',
         },
-      });
+      },
     });
 
-    let tom;
-
-    return run(() => {
-      return store.findRecord('person', 1).then(person => {
-        tom = person;
-        set(tom, 'name', 'Tom Dale');
-        return tom.save();
-      });
-    });
+    set(tom, 'name', 'Tom Dale');
+    await tom.save();
   });
 
   test("When a store is committed, the adapter's `commit` method should be called with records that have been created.", function(assert) {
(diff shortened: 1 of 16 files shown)


packages/-ember-data/tests/integration/adapter/record-persistence-test.js
```diff
@@ -28,7 +28,7 @@ module('integration/adapter/record_persistence - Persisting Records', function(h
     this.owner.register('serializer:application', JSONAPISerializer.extend());
   });
 
-  test("When a store is committed, the adapter's `commit` method should be called with records that have been changed.", function(assert) {
+  test("When a store is committed, the adapter's `updateRecord` method should be called with records that have been changed.", async function(assert) {
```
runspired: the small improvement to this test file is not related to the rest of the PR, just slipped in as a teaching example while pairing on the wider PR. happy to separate out if desired but seemed small and a happy cleanup.

igorT: ❤️ for the note

packages/-ember-data/tests/integration/adapter/record-persistence-test.js
```diff
@@ -42,27 +42,18 @@ module('integration/adapter/record_persistence - Persisting Records', function(h
...
 
-    run(() => {
-      store.push({
-        data: {
-          type: 'person',
-          id: '1',
-          attributes: {
-            name: 'Braaaahm Dale',
-          },
+    const tom = store.push({
```
igorT: Nbd but I don't think we decided to switch to `const` for normal variables

packages/store/addon/-private/system/store/finders.js
```diff
@@ -302,7 +303,8 @@ export function _findHasMany(adapter, store, internalModel, link, relationship,
 export function _findBelongsTo(adapter, store, internalModel, link, relationship, options) {
   let snapshot = internalModel.createSnapshot(options);
   let modelClass = store.modelFor(relationship.type);
-  let promise = adapter.findBelongsTo(store, snapshot, link, relationship);
+  let relatedLink = !link || typeof link === 'string' ? link : link.href;
```
igorT: if link is null doesn't this make relatedLink === true?

runspired: no, `!link` would result in the value being `link` so `!null` (true) results in the value being `null`. This is the value of the ternary operator and is equivalent to

```ts
(!link || typeof link === 'string') ? link : link.href
```

image

packages/-ember-data/tests/integration/adapter/record-persistence-test.js
runspired: We typically use `const` when it should not change which is the case here

packages/store/addon/-private/system/store/finders.js
igorT: Oh god, okay, please add some parens :)

runspired: (also fwiw I authored with the parens for clarity but sadly prettier removes)

runspired:
Added the team discussion label so we can decide whether to

- use a feature flag for the new APIs
- RFC the new APIs

igorT:
Let's rewrite it to be clearer then please

PR #6610, opened 2019-10-21
[BUGFIX] modelFor should return a stable ShimModelClass

runspired:
While working to encapsulate some Model related things I noticed that there are some unintentional changes in the custom-model-class feature flag branch.

- [x] ShimModelClass should be a stable reference per-type
- [x] Test for ShimModelClass stability
- [x] Test that we are able to call `modelFor` for a `type` handled by `instantiateRecord` that has no `ModelClass`
- [x] Fix for the above since the test failed
- [x] Fix for the custom-model-class tests since they should have failed before

### DIFF
--- a/packages/store/addon/-private/system/ds-model-store.ts
+++ b/packages/store/addon/-private/system/ds-model-store.ts
@@ -10,7 +10,7 @@ import { isPresent } from '@ember/utils';
 import { deprecate } from '@ember/application/deprecations';
 import EmberError from '@ember/error';
 import { get } from '@ember/object';
-import ShimModelClass from './model/shim-model-class';
+import ShimModelClass, { getShimClass } from './model/shim-model-class';
 import { setOwner, getOwner } from '@ember/application';
 import { DSModel } from '../ts-interfaces/ds-model';
 import NotificationManager from './record-notification-manager';
@@ -21,6 +21,7 @@ import RecordDataRecordWrapper from '../ts-interfaces/record-data-record-wrapper
 import { SchemaDefinitionService } from '../ts-interfaces/schema-definition-service';
 import { RelationshipsSchema } from '../ts-interfaces/record-data-schemas';
 import notifyChanges from './model/notify-changes';
+type DSModelClass = import('@ember-data/model').default;
 
 /**
   The store service contains all of the data for records loaded from the server.
@@ -143,7 +144,7 @@ class Store extends CoreStore {
   @param {String} modelName
   @return {Model}
     */
-  modelFor(modelName) {
+  modelFor(modelName: string): ShimModelClass | DSModelClass {
     if (DEBUG) {
       assertDestroyedStoreOnly(this, 'modelFor');
     }
@@ -156,15 +157,18 @@ class Store extends CoreStore {
     let maybeFactory = this._modelFactoryFor(modelName);
 
     // for factorFor factory/class split
-    let klass = maybeFactory.class ? maybeFactory.class : maybeFactory;
-    if (!klass.isModel) {
-      return new ShimModelClass(this, modelName);
+    let klass = maybeFactory && maybeFactory.class ? maybeFactory.class : maybeFactory;
+    if (!klass || !klass.isModel) {
+      if (!CUSTOM_MODEL_CLASS || !this.getSchemaDefinitionService().doesTypeExist(modelName)) {
+        throw new EmberError(`No model was found for '${modelName}' and no schema handles the type`);
+      }
+      return getShimClass(this, modelName);
     } else {
       return klass;
     }
   }
 
-  _modelFactoryFor(modelName) {
+  _modelFactoryFor(modelName: string): DSModelClass {
     if (DEBUG) {
       assertDestroyedStoreOnly(this, '_modelFactoryFor');
     }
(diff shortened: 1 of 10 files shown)


packages/store/addon/-private/system/ds-model-store.ts
```diff
@@ -156,15 +157,18 @@ class Store extends CoreStore {
     let maybeFactory = this._modelFactoryFor(modelName);
 
     // for factorFor factory/class split
-    let klass = maybeFactory.class ? maybeFactory.class : maybeFactory;
-    if (!klass.isModel) {
-      return new ShimModelClass(this, modelName);
+    let klass = maybeFactory && maybeFactory.class ? maybeFactory.class : maybeFactory;
+    if (!klass || !klass.isModel) {
+      if (!CUSTOM_MODEL_CLASS || !this.getSchemaDefinitionService().doesTypeExist(modelName)) {
```
igorT: Should this be in the prod build?

packages/store/addon/-private/system/ds-model-store.ts
```diff
@@ -21,6 +21,7 @@ import RecordDataRecordWrapper from '../ts-interfaces/record-data-record-wrapper
 import { SchemaDefinitionService } from '../ts-interfaces/schema-definition-service';
 import { RelationshipsSchema } from '../ts-interfaces/record-data-schemas';
 import notifyChanges from './model/notify-changes';
+type Model = import('@ember-data/model').default;
```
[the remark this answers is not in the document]
runspired: No. Here only want the type and want to ensure no entanglement with the package.

packages/store/addon/-private/system/ds-model-store.ts
```diff
@@ -143,7 +144,7 @@ class Store extends CoreStore {
   @param {String} modelName
   @return {Model}
     */
-  modelFor(modelName) {
+  modelFor(modelName: string): ShimModelClass | Model {
```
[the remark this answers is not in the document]
runspired: I’ll figure out how to make sure TS understands this is a type that extends Model

packages/store/addon/-private/system/model/shim-model-class.ts
```diff
@@ -1,5 +1,24 @@
 import CoreStore from '../core-store';
 import { RelationshipSchema, AttributeSchema } from '../../ts-interfaces/record-data-schemas';
+import { Dict } from '../../ts-interfaces/utils';
+
+const AvailableShims = new WeakMap<CoreStore, Dict<ShimModelClass>>();
+
+export function getShimClass(store: CoreStore, modelName: string): ShimModelClass {
```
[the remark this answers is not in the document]
runspired: It’s accessed a few places and must be keyed by store to support multiple stores.

packages/store/addon/-private/system/ds-model-store.ts
runspired: Yes? Why wouldn’t it be?

packages/-model-encapsulation-test-app/app/app.js
```diff
@@ -3,6 +3,13 @@ import Resolver from './resolver';
 import loadInitializers from 'ember-load-initializers';
 import config from './config/environment';
 
+window.EmberDataENV = {
```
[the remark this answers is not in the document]
runspired: Reasoning?

packages/store/addon/-private/system/ds-model-store.ts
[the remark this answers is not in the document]
runspired: http://www.typescriptlang.org/play/#code/MYGwhgzhAECiAeYC2AHEBTaBvAUNaEALmIQJbDQAW6IIA9gFwGEBOpAdgObQC80A5AEYADKOH8A3HmgB3OixAATJkTZdeAgMz8cAXxw4AZgFd2wMnXbQW6QgmRp0ACgCUTQgE8U6OobiJUDGxpG0JjFit7QPQpfRxgSyJoMA1QqMdXKTAAOmpaOikgA

If `Model` were a class and was typed then we would use `typeof`, however because it comes from JS it is an `any` and `typeof any` is unhappy. This is something we will need to fix once we can type `model`.

I spiked typing `Model`, however, until we can extend using class syntax this problem will persist.

packages/store/addon/-private/system/ds-model-store.ts
runspired: Are you asking if we should reconsider the existence of this prod validation? (It is currently there in prod but I'd be willing to have it be stripped)

packages/store/addon/-private/system/model/shim-model-class.ts
```diff
@@ -1,5 +1,24 @@
...
+  let shim = shims[modelName];
+  if (shim === undefined) {
+    shim = shims[modelName] = new ShimModelClass(store, modelName);
+  }
+
+  return shim;
+}
 
 // Mimics the static apis of DSModel
 export default class ShimModelClass {
```
chriskrycho: Given my suggestions above and specifically what (2) would entail, you could define it so that here you had `ShimModelClass implements ModelStatic` or whatever you called it. That'd guarantee that instead of just mimicking the static API of DSModel, you'd be *guaranteed* to uphold it.

runspired: I'm going to table this as an improvement we should do, and one I promise to get to soon :P

runspired: I'm going to table this as something we should do (and do soon). It'll be a key step of the continuing work to pull the Model specific code bits into the model package.

PR #6612, opened 2019-10-21
[CHORE]: Refactor integration/adapter/find-test.js

dmuneras:
## Description

Attempt to help to modernize the tests.
 
* Remove run(() => { }) when possible.
* Setup test information in the same test to ease reviews and have all the context required within the test.
* Use const when possible.

### DIFF
--- a/packages/-ember-data/tests/integration/adapter/find-test.js
+++ b/packages/-ember-data/tests/integration/adapter/find-test.js
@@ -192,7 +192,8 @@ module('integration/adapter/find - Finding Records', function(hooks) {
       await store.findRecord('person', 'the-id');
       assert.ok(false, 'We expected to throw but did not');
     } catch (e) {
-      const expectedMessageRegex = "Assertion Failed: You made a 'findRecord' request for a 'person' with id 'the-id', but the adapter's response did not have any data";
+      const expectedMessageRegex =
+        "Assertion Failed: You made a 'findRecord' request for a 'person' with id 'the-id', but the adapter's response did not have any data";
 
       assert.strictEqual(expectedMessageRegex, e.message, 'error has the correct error message');
     }
@@ -220,7 +221,8 @@ module('integration/adapter/find - Finding Records', function(hooks) {
     try {
       await all(promises);
     } catch (e) {
-      const expectedMessageRegex = "Assertion Failed: You made a 'findMany' request for 'person' records with ids '\[1,2\]', but the adapter's response did not have any data";
+      const expectedMessageRegex =
+        "Assertion Failed: You made a 'findMany' request for 'person' records with ids '[1,2]', but the adapter's response did not have any data";
 
       assert.strictEqual(expectedMessageRegex, e.message, 'error has the correct error message');
     }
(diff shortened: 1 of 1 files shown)


packages/-ember-data/tests/integration/adapter/find-test.js
```diff
@@ -59,17 +63,24 @@ module('integration/adapter/find - Finding Records', function(hooks) {
...
-
     this.owner.register('serializer:application', JSONAPISerializer.extend());
 
-    run(() => {
-      store.findRecord('person', 1);
-      store.findRecord('person', 1);
-    });
+    const store = this.owner.lookup('service:store');
+
+    store.findRecord('person', 1);
```
runspired: wherever we are currently allowing numeric ID to be passed in as an argument we should switch to string ID, unless is it an explicit test that "passing by number" works. This extends more generally to "lets use string IDs in tests unless explicitly testing a number", as it will make it that much easier to work to constrain the system in the future.

packages/-ember-data/tests/integration/adapter/find-test.js
```diff
@@ -144,32 +160,52 @@ module('integration/adapter/find - Finding Records', function(hooks) {
...
+    const store = this.owner.lookup('service:store');
 
-    assert.expectAssertion(() => {
-      run(() => store.findRecord('person', 'the-id'));
-    }, /You made a 'findRecord' request for a 'person' with id 'the-id', but the adapter's response did not have any data/);
+    try {
+      await store.findRecord('person', 'the-id');
+      assert.ok(false, 'We expected to throw but did not');
+    } catch (e) {
+      const expectedMessageRegex = /You made a 'findRecord' request for a 'person' with id 'the-id', but the adapter's response did not have any data/;
```
runspired: this particular error feels like we could do a strictEquals comparison and be cleaner

packages/-ember-data/tests/integration/adapter/find-test.js
```diff
@@ -144,32 +160,52 @@ module('integration/adapter/find - Finding Records', function(hooks) {
...
+
+      assert.ok(expectedMessageRegex.test(e.message));
```
runspired: always make sure that assertions have labels :)

packages/-ember-data/tests/integration/adapter/find-test.js
```diff
@@ -178,17 +214,26 @@ module('integration/adapter/find - Finding Records', function(hooks) {
...
-        store.findRecord('person', '1');
-        store.findRecord('person', '2');
-      });
-    }, /You made a 'findMany' request for 'person' records with ids '\[1,2\]', but the adapter's response did not have any data/);
+    try {
+      await all(promises);
+    } catch (error) {
+      const expectedMessageRegex = /You made a 'findMany' request for 'person' records with ids '\[1,2\]', but the adapter's response did not have any data/;
+
+      assert.ok(expectedMessageRegex.test(error.message));
```
runspired: same as above, label for the assertion and likely shift to strictEquals string comparison

PR #6613, opened 2019-10-21
Adds relationship-performance-test-app

skaterdav85:


packages/-relationship-performance-test-app/app/models/color.js
```diff
@@ -0,0 +1,11 @@
+import DS from 'ember-data';
+
+const { Model, attr, hasMany } = DS;
+
+export default Model.extend({
+  name: attr('string'),
+  sortOrder: attr('number'),
+
+  // this makes the opposite side of the relationship faster
```
runspired: This is something that surprised me when your team brought it up. We should add a similar test without this to demonstrate that it is slower, likely there are things we can fix.

packages/-relationship-performance-test-app/app/routes/application.js
```diff
@@ -0,0 +1,22 @@
+import Route from '@ember/routing/route';
+
+export default Route.extend({
+  model() {
```
runspired: having this be `application` route is fine for now, but we should separate out various perf scenarios into different routes going forward, for instance so that we can create a second similar scenario without the inverse relationships being defined.

packages/-relationship-performance-test-app/app/models/color.js
skaterdav85: So from what I was told, the inverse on a `belongsTo` relationship always improved the speed and the inverse on a `hasMany` relationship made it slower. I couldn't reproduce it in this app though.

runspired: We should remove the comments, especially If we can't reproduce it. I'd still like to try to recreate the scenario that performed badly at a later time though.

kjhangiani: I'll see if I can get a test scenario that reproduces the problem we had.  In the case that I was investigating, the pathological performance was in a scenario where you have ~10000 cars, but are only iterating over ~500 of them and accessing the hasMany('color').  In this scenario, the reverse relationship on color was causing a significant slowdown relative to the reverse relationship not being present.

My (unconfirmed) hunch is that even though we are only iterating over a subset of the cars, when the hasMany property is accessed, it is actually creating the reverse relationships on colors for ALL of the remaining cars as well

PR #6626, opened 2019-10-23
[CHORE] adds infra for testing calls to Ember warn|deprecate|assert

runspired:
Depends on #6625 for tsconfig fixes

This:

- Allows us to remove more run loops.
- Gives us the ability to expand how we handle managing deprecations (done in a follow up commit)

It introduces:

- [x] expected deprecations/warnings are now removed from the list of encountered deprecations/warnings so that expectNo<Deprecation|Warning> can correctly assert no other deprecations or warnings were encountered
 - [x] assert.expectDeprecation | assert.expectAssertion | assert.expectWarning allow for async
 - [x] deprecations and warnings can be matched by id

### DIFF
--- a/packages/-ember-data/tests/helpers/qunit-asserts/assert-assertion.ts
+++ b/packages/-ember-data/tests/helpers/qunit-asserts/assert-assertion.ts
@@ -1,0 +1,76 @@
+import QUnit from 'qunit';
+import { checkMatcher } from './check-matcher';
+import RSVP from 'rsvp';
+
+let HAS_REGISTERED = false;
+
+interface AssertSomeResult {
+  result: boolean;
+  actual: string;
+  expected: string;
+  message: string;
+}
+interface AssertNoneResult {
+  result: boolean;
+  actual: string;
+  expected: '';
+  message: string;
+}
+
+function expectAssertion(message: string, matcher: string | RegExp): AssertSomeResult {
+  let passed = checkMatcher(message, matcher);
+
+  return {
+    result: passed,
+    actual: message,
+    expected: String(matcher),
+    message: `Expected an assertion during the test`,
+  };
+}
+
+function expectNoAssertion(message: string | undefined): AssertNoneResult {
+  let passed = !message;
+  return {
+    result: passed,
+    actual: message || '',
+    expected: '',
+    message: `Expected no assertions during test`,
+  };
+}
+
+export function configureAssertionHandler() {
+  if (HAS_REGISTERED === true) {
+    throw new Error(`Attempting to re-register the assert-assertion handler`);
+  }
+  HAS_REGISTERED = true;
+
+  QUnit.assert.expectAssertion = async function(cb: () => unknown, matcher: string | RegExp): Promise<void> {
+    let outcome;
+    try {
+      let result = cb();
+      if (result instanceof Promise || result instanceof RSVP.Promise) {
+        await result;
+      }
+      outcome = expectAssertion('', matcher);
+    } catch (e) {
+      outcome = expectAssertion(e.message, matcher);
+    }
+
+    this.pushResult(outcome);
+  };
+
+  QUnit.assert.expectNoAssertion = async function(cb: () => unknown) {
+    let outcome;
+    try {
+      let result = cb();
+      if (result instanceof Promise || result instanceof RSVP.Promise) {
+        await result;
+      }
+      outcome = expectNoAssertion('');
+    } catch (e) {
+      outcome = expectNoAssertion(e.message);
+    }
+
+    this.pushResult(outcome);
+  };
+}
--- a/packages/-ember-data/tests/helpers/qunit-asserts/assert-assertion.ts
+++ b/packages/-ember-data/tests/helpers/qunit-asserts/assert-assertion.ts
@@ -18,24 +18,24 @@ interface AssertNoneResult {
   message: string;
 }
 
-function verifyAssertion(message: string, matcher: string | RegExp): AssertSomeResult {
+function verifyAssertion(message: string, matcher: string | RegExp, label?: string): AssertSomeResult {
   let passed = checkMatcher(message, matcher);
 
   return {
     result: passed,
     actual: message,
     expected: String(matcher),
-    message: `Expected an assertion during the test`,
+    message: label || `Expected an assertion during the test`,
   };
 }
 
-function verifyNoAssertion(message: string | undefined): AssertNoneResult {
+function verifyNoAssertion(message: string | undefined, label?: string): AssertNoneResult {
   let passed = !message;
   return {
     result: passed,
     actual: message || '',
     expected: '',
-    message: `Expected no assertions during test`,
+    message: label || `Expected no assertions during test`,
   };
 }
 
@@ -45,7 +45,11 @@ export function configureAssertionHandler() {
   }
   HAS_REGISTERED = true;
 
-  QUnit.assert.expectAssertion = async function(cb: () => unknown, matcher: string | RegExp): Promise<void> {
+  QUnit.assert.expectAssertion = async function(
+    cb: () => unknown,
+    matcher: string | RegExp,
+    label?: string
+  ): Promise<void> {
     let outcome;
     if (DEBUG) {
       try {
@@ -53,9 +57,9 @@ export function configureAssertionHandler() {
         if (isThenable(result)) {
           await result;
         }
-        outcome = verifyAssertion('', matcher);
+        outcome = verifyAssertion('', matcher, label);
       } catch (e) {
-        outcome = verifyAssertion(e.message, matcher);
+        outcome = verifyAssertion(e.message, matcher, label);
       }
     } else {
       outcome = {
@@ -69,16 +73,16 @@ export function configureAssertionHandler() {
     this.pushResult(outcome);
   };
 
-  QUnit.assert.expectNoAssertion = async function(cb: () => unknown) {
+  QUnit.assert.expectNoAssertion = async function(cb: () => unknown, label?: string) {
     let outcome;
     try {
       let result = cb();
       if (isThenable(result)) {
         await result;
       }
-      outcome = verifyNoAssertion('');
+      outcome = verifyNoAssertion('', label);
     } catch (e) {
-      outcome = verifyNoAssertion(e.message);
+      outcome = verifyNoAssertion(e.message, label);
     }
 
     this.pushResult(outcome);
--- a/packages/-ember-data/tests/helpers/qunit-asserts/assert-assertion.ts
+++ b/packages/-ember-data/tests/helpers/qunit-asserts/assert-assertion.ts
@@ -1,5 +1,6 @@
 import QUnit from 'qunit';
 import { checkMatcher } from './check-matcher';
+import { DEBUG } from '@glimmer/env';
 import RSVP from 'rsvp';
 
 let HAS_REGISTERED = false;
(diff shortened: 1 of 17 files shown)


packages/-ember-data/tests/helpers/qunit-asserts/assert-warning.ts
```diff
@@ -0,0 +1,168 @@
...
+
+/**
+ * Returns a qunit assert result object which passes if the given warning
+ * `id` was found *exactly* `count` times.
+ *
+ * Fails if not found or found more or less than `count`.
+ * Fails if `until` not specified
+ * Optionally fails if `until` has been passed.
+ */
+function expectWarning(config: WarningConfig): AssertSomeResult {
```
rwjblue: Can we tweak the name of this function a bit? I find it fairly odd that this is `expectWarning` **and** the `QUnit.assert` method is `expectWarning` but they are quite different conceptually.

I'd suggest something like `validateWarnings` or `checkWarnings`.

packages/-ember-data/tests/helpers/qunit-asserts/assert-deprecation.ts
```diff
@@ -0,0 +1,168 @@
...
+    result: passed,
+    actual: { id: config.id, count: matchedDeprecations.length },
+    expected: { id: config.id, count: expectedCount },
+    message: `Expected ${expectedCount} deprecation${expectedCount === 1 ? '' : 's'} for ${config.id} during test, ${
+      passed ? expectedCount : 'but ' + matchedDeprecations.length
+    } deprecations were found.`,
+  };
+}
+
+function expectNoDeprecation(): AssertNoneResult {
```
rwjblue: Lets call this method something different, its pretty confusing to have both `QUnit.assert.expectNoDeprecation` and this inline function with the same name but quite different.

I'm thinking something like `verifyNoDeprecationsIssued`...

packages/-ember-data/tests/helpers/qunit-asserts/assert-deprecation.ts
```diff
@@ -0,0 +1,168 @@
...
+
+/**
+ * Returns a qunit assert result object which passes if the given deprecation
+ * `id` was found *exactly* `count` times.
+ *
+ * Fails if not found or found more or less than `count`.
+ * Fails if `until` not specified
+ * Optionally fails if `until` has been passed.
+ */
+function expectDeprecation(config: DeprecationConfig): AssertSomeResult {
```
rwjblue: Let's call this method something different, its pretty confusing to have both `QUnit.assert.expectDeprecation` and this inline function with the same name but quite different.

I'm thinking something like `verifyDeprecations`...

packages/-ember-data/tests/helpers/qunit-asserts/assert-assertion.ts
```diff
@@ -0,0 +1,86 @@
...
+  message: string;
+}
+interface AssertNoneResult {
+  result: boolean;
+  actual: string;
+  expected: '';
+  message: string;
+}
+
+function expectAssertion(message: string, matcher: string | RegExp): AssertSomeResult {
```
rwjblue: Lets call this method something different, its pretty confusing to have both `QUnit.assert.expectAssertion` and this inline function with the same name but quite different.

I'm thinking something like `verifyAssertions`...

packages/-ember-data/tests/helpers/qunit-asserts/assert-assertion.ts
```diff
@@ -0,0 +1,86 @@
...
+
+  return {
+    result: passed,
+    actual: message,
+    expected: String(matcher),
+    message: `Expected an assertion during the test`,
+  };
+}
+
+function expectNoAssertion(message: string | undefined): AssertNoneResult {
```
rwjblue: Same thing RE: naming here, maybe this should be `verifyNoAssertions`?

packages/-ember-data/tests/helpers/qunit-asserts/assert-warning.ts
```diff
@@ -0,0 +1,168 @@
...
+  // TODO optionally throw if `until` is the current version or older than current version
+  let matchedWarnings = WARNINGS_FOR_TEST.filter(warning => {
+    if (!warning.options || !warning.options.id) {
+      throw new Error(`Expected warning to have an id, found: ${warning}`);
```
rwjblue: Do we actually need to do this? Doesn't Ember itself error if there is no `id`?

rwjblue: See here: 

https://github.com/emberjs/ember.js/blob/129d00b3b4f14826bd9ea4499fd0c8bc7ee5a5a5/packages/%40ember/debug/lib/warn.ts#L105-L106

packages/-ember-data/tests/helpers/qunit-asserts/assert-warning.ts
```diff
@@ -0,0 +1,168 @@
...
+    result: passed,
+    actual: { id: config.id, count: matchedWarnings.length },
+    expected: { id: config.id, count: expectedCount },
+    message: `Expected ${expectedCount} warning${expectedCount === 1 ? '' : 's'} for ${config.id} during test, ${
+      passed ? expectedCount : 'but ' + matchedWarnings.length
+    } warnings were found.`,
+  };
+}
+
+function expectNoWarning(): AssertNoneResult {
```
rwjblue: Same thing RE: naming

packages/-ember-data/tests/helpers/qunit-asserts/assert-warning.ts
```diff
@@ -0,0 +1,168 @@
...
+        until: '4.0',
+      };
+    }
+
+    if (callback) {
+      WARNINGS_FOR_TEST = [];
+      let result = callback();
+      if (result instanceof Promise || result instanceof RSVP.Promise) {
+        await result;
+      }
```
rwjblue: This should be fine here:

```
await callback();
```

packages/-ember-data/tests/helpers/qunit-asserts/assert-warning.ts
```diff
@@ -0,0 +1,168 @@
...
+
+  QUnit.assert.expectNoWarning = async function(cb) {
+    let origWarnings = WARNINGS_FOR_TEST;
+
+    if (cb) {
+      WARNINGS_FOR_TEST = [];
+      let result = cb();
+      if (result instanceof Promise || result instanceof RSVP.Promise) {
+        await result;
+      }
```
rwjblue: `await cb();`

packages/-serializer-encapsulation-test-app/tests/integration/normalize-test.js
```diff
@@ -83,7 +83,7 @@ module('integration/serializer - normalize method forwards to Serializer#normali
 
     const store = this.owner.lookup('service:store');
 
-    assert.expectAssertion(() => {
+    assert.throws(() => {
```
rwjblue: Why does this one change? Is it always an error (e.g. not an assertion)?

packages/-serializer-encapsulation-test-app/tests/integration/push-payload-test.js
```diff
@@ -90,7 +90,7 @@ module('integration/push-payload - pushPayload method forwards to Serializer#pus
 
     const store = this.owner.lookup('service:store');
 
-    assert.expectAssertion(() => {
+    assert.throws(() => {
```
rwjblue: Same question: why does this one become `assert.throws`, is the underlying thing a `throw new Error()` vs `assert(...)`?

packages/-ember-data/tests/helpers/qunit-asserts/assert-assertion.ts
```diff
@@ -0,0 +1,76 @@
...
+  if (HAS_REGISTERED === true) {
+    throw new Error(`Attempting to re-register the assert-assertion handler`);
+  }
+  HAS_REGISTERED = true;
+
+  QUnit.assert.expectAssertion = async function(cb: () => unknown, matcher: string | RegExp): Promise<void> {
+    let outcome;
+    try {
+      let result = cb();
+      if (result instanceof Promise || result instanceof RSVP.Promise) {
```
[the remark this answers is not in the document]
runspired: 🤔 does make it work with more thenables. Seems like an ok change.

packages/-ember-data/tests/helpers/qunit-asserts/assert-deprecation.ts
```diff
@@ -0,0 +1,168 @@
...
+      isMatched = checkMatcher(deprecation.message, config.message);
+    }
+    return isMatched;
+  });
+  DEPRECATIONS_FOR_TEST = DEPRECATIONS_FOR_TEST.filter(deprecation => {
+    matchedDeprecations.indexOf(deprecation) === -1;
+  });
+  HANDLED_DEPRECATIONS_FOR_TEST.push(...matchedDeprecations);
+
+  let expectedCount = typeof config.count === 'number' && config.count !== 0 ? config.count : 1;
```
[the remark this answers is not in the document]
runspired: I think you needed to refresh :P

packages/-ember-data/tests/helpers/qunit-asserts/assert-assertion.ts
```diff
@@ -0,0 +1,86 @@
...
+    throw new Error(`Attempting to re-register the assert-assertion handler`);
+  }
+  HAS_REGISTERED = true;
+
+  QUnit.assert.expectAssertion = async function(cb: () => unknown, matcher: string | RegExp): Promise<void> {
+    let outcome;
+    if (DEBUG) {
+      try {
+        let result = cb();
+        if (result instanceof Promise || result instanceof RSVP.Promise) {
```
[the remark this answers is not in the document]
runspired: it isn't backwards compatible, I'll move this to the `.then` check

packages/-ember-data/tests/helpers/qunit-asserts/assert-assertion.ts
```diff
@@ -0,0 +1,86 @@
...
+  };
+}
+
+export function configureAssertionHandler() {
+  if (HAS_REGISTERED === true) {
+    throw new Error(`Attempting to re-register the assert-assertion handler`);
+  }
+  HAS_REGISTERED = true;
+
+  QUnit.assert.expectAssertion = async function(cb: () => unknown, matcher: string | RegExp): Promise<void> {
```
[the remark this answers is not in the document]
runspired: you mean a label?

packages/-ember-data/tests/helpers/qunit-asserts/assert-deprecation.ts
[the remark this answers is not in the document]
runspired: I believe I removed this (might be stuck in a later PR), but the general reasoning was that you should never expect zero deprecations, later decided that was meh :)

packages/-ember-data/tests/helpers/qunit-asserts/assert-deprecation.ts
```diff
@@ -0,0 +1,168 @@
...
+  };
+}
+
+export function configureDeprecationHandler() {
+  if (HAS_REGISTERED === true) {
+    throw new Error(`Attempting to re-register the assert-deprecation handler`);
+  }
+  HAS_REGISTERED = true;
+
+  QUnit.testStart(function() {
```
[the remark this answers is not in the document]
runspired: `testDone` doesn't work well for this, a later PR does add this check to all of our tests alongside a filter so that we only error for unhandled deprecations that originated from EmberData.

packages/-ember-data/tests/helpers/qunit-asserts/assert-warning.ts
runspired: Odd, we've had many without ID's slip past before 🤔

packages/-serializer-encapsulation-test-app/tests/integration/normalize-test.js
runspired: its not run in production and it doesn't share the infra that ember-data's tests have

packages/-serializer-encapsulation-test-app/tests/integration/push-payload-test.js
runspired: it's an assert but it throws like a normal error

packages/-ember-data/tests/helpers/qunit-asserts/assert-deprecation.ts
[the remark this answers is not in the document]
runspired: made sure it was removed in this PR

packages/-ember-data/tests/helpers/qunit-asserts/assert-warning.ts
runspired: Went ahead and removed for now

runspired:
After discussing with Rob a follow-up to refactor away our "sync" usage of these helpers will occur so that we can move to an async only world.

PR #6627, opened 2019-10-23
[CHORE] allow asserting all tests for deprecations

runspired:
turned off until we fix tests in follow up PR, but have written all necessary test fixes locally already.

This gives us guarantees we don't regress on our own deprecations

It introduces: 

- [x] deprecatedTest now auto asserts deprecations it surrounds
- [x] all tests now assert unexpected ember-data deprecations

There's a LOT more we can do from here for things like asserting encountered deprecations have `url` and `until` attached to them for instance, or making a report that errors if we have deprecations coming from `Ember` itself.

I also think we should wrap the deprecation and warning helpers in `if (DEBUG)` and make them pass in prod automatically to reduce the number of tests that require `testInDebug`.

### DIFF
--- a/packages/-ember-data/tests/test-helper.js
+++ b/packages/-ember-data/tests/test-helper.js
@@ -8,6 +8,7 @@ import { DEBUG } from '@glimmer/env';
 import QUnit from 'qunit';
 import { wait, asyncEqual, invokeAsync } from 'dummy/tests/helpers/async';
 import configureAsserts from 'dummy/tests/helpers/qunit-asserts';
+import { SHOULD_ASSERT_ALL } from './helpers/deprecated-test';
 
 configureAsserts();
 
@@ -40,7 +41,9 @@ QUnit.begin(() => {
       const hooks = (mod.hooks.afterEach = mod.hooks.afterEach || []);
 
       if (mod.tests.length !== 0) {
-        hooks.unshift(assertAllDeprecations);
+        if (SHOULD_ASSERT_ALL) {
+          hooks.unshift(assertAllDeprecations);
+        }
       }
     });
   }
--- a/packages/-ember-data/tests/test-helper.js
+++ b/packages/-ember-data/tests/test-helper.js
@@ -3,9 +3,9 @@ import config from '../config/environment';
 import RSVP from 'rsvp';
 import { setApplication } from '@ember/test-helpers';
 import { start } from 'ember-qunit';
+import { DEBUG } from '@glimmer/env';
 
 import QUnit from 'qunit';
-import DS from 'ember-data';
 import { wait, asyncEqual, invokeAsync } from 'dummy/tests/helpers/async';
 import configureAsserts from 'dummy/tests/helpers/qunit-asserts';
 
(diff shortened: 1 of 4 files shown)


packages/-ember-data/tests/test-helper.js
```diff
@@ -3,41 +3,60 @@ import config from '../config/environment';
...
 
 QUnit.begin(() => {
+  function assertAllDeprecations(assert) {
+    if (typeof assert.test.expected === 'number') {
+      assert.test.expected += 1;
+    }
+    assert.expectNoDeprecation(undefined, undefined, deprecation => {
+      // only assert EmberData deprecations
+      const id = deprecation.options.id;
+      const isEmberDataDeprecation = id.includes('DS') || id.includes('EmberData') || id.includes('ember-data');
```
rwjblue: I assume the ones other than `ember-data` are "legacy" (meaning we would use `ember-data` or `@ember-data/foo` going forward), right?

packages/-ember-data/tests/test-helper.js
```diff
@@ -3,41 +3,60 @@ import config from '../config/environment';
...
+        console.log('Detected Non-Ember-Data Deprecation', deprecation);
+      }
+
+      return isEmberDataDeprecation;
+    });
+  }
+  // ensure we don't regress quietly
+  // this plays nicely with `expectDeprecation`
+  if (DEBUG) {
+    QUnit.config.modules.forEach(mod => {
```
rwjblue: I don't _particularly_ like this bit, I'd rather add an import that does what we want `setupDeprecationAssertions(hooks)`. I think this is fine for a starting point, but would you mind adding a issue to move to a "less magic" setup (I'm happy to work on a small codemod to make it easy, since it seems like we have a few different things we want to tweak coming up).

packages/-ember-data/tests/test-helper.js
```diff
@@ -3,41 +3,60 @@ import config from '../config/environment';
...
+        }
+      }
+    });
+  }
+
   RSVP.configure('onerror', reason => {
     // only print error messages if they're exceptions;
     // otherwise, let a future turn of the event loop
     // handle the error.
+    // TODO kill this off
```
rwjblue: Why do we even do this? Does anything fail if we remove this `RSVP.configure`?

packages/-ember-data/tests/test-helper.js
runspired: @rwjblue correct

packages/-ember-data/tests/test-helper.js
runspired: @rwjblue I don't know that I agree: manually adding this to every single test (or module) when you run it on every test seems annoying

packages/-ember-data/tests/test-helper.js
runspired: I'm not sure yet, but I left the note because I want to try removing it :)

packages/-ember-data/tests/test-helper.js
runspired: (but we should discuss)

PR #6631, opened 2019-10-23
Add TracerBench to unpublished-relationship-performance-test-app

skaterdav85:
(Edited)

This PR adds TracerBench v2.3.0 to `unpublished-relationship-performance-test-app`. The main script to do everything is `./bin/relationship-performance-tracking/src/head-vs-master.sh`. This script will create a build of `unpublished-relationship-performance-test-app` off of master (the control) and for the current PR (the experiment). Then, HAR Remix is used to serve up each build using a HAR file. `har-remix.js` will swap out `vendor.js` with the `vendor.js` file from the control and experiment. Then TracerBench is used to run 60 samples and analyzes the results.


runspired:
> What is the difference between tracerbench create-archive and tracerbench trace? Do both visit the supplied URL?

`create-archive` generates a `HAR` that can be reused to perform multiple analyses. 
`trace` performs an analysis, but kindly it generates a `HAR` if one does not already exist.

RE 2) I'll take a peek

skaterdav85:
```
server-replay ./packages/-relationship-performance-test-app/bin/localhost.har --port 4200 & yarn run tracerbench:compare
```

PR #6632, opened 2019-10-23
[CHORE] Refactor integration/multiple-stores-test.js in order to remove run loop usage

dmuneras:
## Description

* Remove run(() => { }) when possible.
* Use const when possible.

### DIFF
--- a/packages/-ember-data/tests/integration/multiple-stores-test.js
+++ b/packages/-ember-data/tests/integration/multiple-stores-test.js
@@ -1,7 +1,5 @@
-import { run } from '@ember/runloop';
 import { setupTest } from 'ember-qunit';
 import { module, test } from 'qunit';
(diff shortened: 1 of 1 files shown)


igorT:
Thanks for the PR.  I'll double check the removing of the runloop, but going over it quickly seems great. Re: the `const` changes, in the repo we try to follow the style of using consts for top level values that are semantically constants but not for local variables.

runspired:
I'll follow up with Igor because my understanding (and own practices) differ from his on const usage. I prefer let to const *except* when something MUST not be reassigned.

e.g. if reassignment would break the code or the assumptions, then use const. In the case of these tests, reassignment would break the contract of what we are testing. E.g. you wouldn't want `andromedaStore = this.owner('service:store-a')` to ever be reassigned or you would no longer be testing what you intended.

Issue #6012, opened 2019-04-10
PromiseProxy meta access error message is untested


igorT:
https://github.com/emberjs/data/blob/f94776c761d0c47b17f2a340775689cce1aa582b/packages/store/addon/-private/system/promise-proxies.js#L89

This error message seems to not be tested

Mikek2252:
Hey @igorT i'm looking at adding this test but i'm not 100% sure how i am accessing the object. I presumed it would be `DS.PromiseBelongsTo` but that doesn't seem to work, i guess because there isn't an export function like the other models. I have the test case, but could you point me to how i can create the object in the test.

PR #6640, opened 2019-10-25
ISSUE-6012: test: added test for meta property and bug fix.

Mikek2252:
#6012 Issue to add test for meta property for PromiseBelongsTo.

I think i also found a bug in the assertion because internal model is not present on the object, just originatingInternalModel, so i changed it to modelName. I need confirmation that this is the intended behaviour.

### DIFF
--- a/packages/-ember-data/tests/unit/promise-proxies-test.js
+++ b/packages/-ember-data/tests/unit/promise-proxies-test.js
@@ -1,6 +1,9 @@
 import { Promise as EmberPromise } from 'rsvp';
 import { A } from '@ember/array';
-
+import Model, { belongsTo } from '@ember-data/model';
+import Adapter from '@ember-data/adapter';
+import JSONAPISerializer from '@ember-data/serializer/json-api';
+import { setupTest } from 'ember-qunit';
 import { module, test } from 'qunit';
 
 import DS from 'ember-data';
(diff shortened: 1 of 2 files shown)


packages/-ember-data/tests/unit/promise-proxies-test.js
```diff
@@ -110,3 +113,69 @@ module('PromiseManyArray', function() {
...
+    @belongsTo('child', { async: true })
+    child;
+  }
+  class Child extends Model {
+    @belongsTo('parent', { async: false })
+    parent;
+  }
+  class ChildAdapter extends Adapter {
+    findRecord(store, type, id, snapshot) {
+      const ChildRecord = {
```
runspired: Do we know if accessing meta ends up triggering this request? We should add an assertion to codify the current behavior so we at least know if/when we change it later.

Mikek2252: Yeah, in the console it was trying to fetch the record but the test would fail because the endpoint doesn't exist so i just mocked it. 
I added some debuggers in and it appears that findRecord runs before we check the meta, so i guess the `parent.child` bit causes the request to trigger.

runspired: That's what I'd expect. We should add a test so that this behavior is explicitly covered. It's not necessarily a good behavior, but knowing that's what we do now is important to refactoring later :)

PR #6642, opened 2019-10-25
[FEAT] RequestManager (emberjs/rfcs#860)

runspired:
Implements @ember-data/request Package from https://github.com/emberjs/rfcs/pull/860

progress towards finalizing #6166 
Required for #8086 

TODO

- [x] implement RequestManager
- [x] implement `Fetch` handler
- [x] add RequestManager tests
- [x] update docs and code import paths to reflect named import/export
- [x] re-review new READMEs
- [x] add API docs
- [x] update RFC with implementation notes and final interfaces

TODO (follow up)

- add experimental typescript definitions for RequestManager / Handler / Fetch handler
- makes `store.notifications` public to give access to the notification manager
- add `cache.peek` to give access to full resource data (emberjs/rfcs#854) 
- add support for `RequestManager` to the store
- add LegacyHandler


packages/request/README.md
```diff
@@ -0,0 +1,362 @@
...
+```ts
+import Store from '@ember-data/store';
+import RequestManager from '@ember-data/request';
+
+export default class extends Store {
+  requestManager = new RequestManager();
+
+  constructor(args) {
+    super(args);
+    this.requestManager.use([LegacyHandler]);
```
sly7-7, line 355: I just can't see where this `LegacyHandler` is coming from

runspired: Oh I see, the import didn't make it into the ReadMe. It's in the rfc iirc, will add here too.

packages/store/README.md
```diff
@@ -1,31 +1,180 @@
...
+```js
+import Store from '@ember-data/store';
+import Adapter from '@ember-data/adapter/json-api';
+import { getOwner, setOwner } from '@ember/application';
+
+class extends Store {
+  #adapter = null;
+
+  adapterFor() {
+    let adapter = thsi.#adapter;
```
sly7-7, line 123: *this

packages/request/README.md
sly7-7: Sorry for not looking at the RFC, indeed I just see it there

PR #6644, opened 2019-10-25
[BUGFIX] fix single-table polymorphism when IDENTIFIERS is ON

juggy:
This is a failing test for the polymorphic association and identifiers. @runspired 

This happens in my app when using 3.14.beta0: 

A record of type `company` is a subclass of `party` and it is first loaded as a `company` within a `belongsTo` relation, then a `findRecord` is done on `party` with the same id.

This PR is a failing test that is a direct reference to the RFC#403 section on single-table polymorphism and identifier: 
```
For instance, when using single-table polymorphism (in which ferrari and bmw extend car and share a common id space) then ferrari:1 and bmw:2 are the same vehicles as car:1 and car:2.
```

### DIFF
--- a/packages/store/addon/-private/identifiers/cache.ts
+++ b/packages/store/addon/-private/identifiers/cache.ts
@@ -325,7 +325,7 @@ export class IdentifierCache {
     let newId = coerceId(data.id);
 
     const keyOptions = getTypeIndex(this._cache.types, identifier.type);
-    let existingIdentifier = detectMerge(keyOptions, identifier, newId);
+    let existingIdentifier = detectMerge(keyOptions, identifier, data, newId, this._cache.lids);
 
     if (existingIdentifier) {
       identifier = this._mergeRecordIdentifiers(keyOptions, identifier, existingIdentifier, data, newId as string);
@@ -364,6 +364,9 @@ export class IdentifierCache {
 
     // ensure a secondary cache entry for this id for the identifier we do keep
     keyOptions.id[newId] = kept;
+    // ensure a secondary cache entry for this id for the abandoned identifier's type we do keep
+    let baseKeyOptions = getTypeIndex(this._cache.types, existingIdentifier.type);
+    baseKeyOptions.id[newId] = kept;
 
     // make sure that the `lid` on the data we are processing matches the lid we kept
     data.lid = kept.lid;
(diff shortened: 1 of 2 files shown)


runspired:
@juggy just wanted to make sure my memory was correct, things work correctly for you in 3.13 stable, yes?

igorT:
@runspired I have comfirmed that it passes with the Identifiers ff off

runspired:
@igorT in that case I think we should turn off the flag for this release, but let's leave it on for canary

runspired:
Opened a tracking issue here https://github.com/emberjs/data/issues/6768

runspired:
A summary of why this worked before for future explorers:

This case is very similar to the `username` case where the response payload has a different ID:

```js
record = await store.findRecord('user', 'runspired');
record2 = await store.findRecord('user', '1');
record === record2; // true
```

In the above case, the user ends up with two RecordIdentifiers and two InternalModels in the cache, but both resolve to the same record. We added a `mergeIdentifier` operation when we detect this which removes one of the two RecordIdentifiers and results in only a single InternalModel in the end.

In this case, instead of different IDs we have the same ID but different `types`. Again we end up with two RecordIdentifiers and two InternalModels in the cache, but with both resolving to the same record.

Our solution here is the same, we run the same `mergeOperation` when we detect the case to end up in a state with only one RecordIdentifier and one InternalModel. Thus, this is both a bugfix for the IDENTIFIERS work but also an improvement/bugfix over the state of things pre-identifiers.

This said, the heuristic we are using could be problematic, and it would be better for applications using polymorphism to explicitly configure identifier generation to ensure records that can have multiple types are properly considered the same without depending on this merge.

packages/store/addon/-private/identifiers/cache.ts
```diff
@@ -509,13 +514,25 @@ function performRecordIdentifierUpdate(
...
 ): StableRecordIdentifier | false {
-  const { id } = identifier;
+  const { id, type, lid } = identifier;
   if (id !== null && id !== newId && newId !== null) {
     const existingIdentifier = keyOptions.id[newId];
 
     return existingIdentifier !== undefined ? existingIdentifier : false;
+  } else {
+    // detect unconfigured single-table polymorphism
+    // TODO deprecate in favor of users configuring lid generation
```
igorT: I think the discussion we had was that we are not gonna require people to configure lid generation but rather find  way to get the request context to the identifier merge function so that we can do the merge every time. Can you please update the todo and file an issue?

runspired: https://github.com/emberjs/data/issues/6890

PR #6668, opened 2019-11-04
[CHORE] fix location of fastboot test for new person

runspired:
Somehow these got left behind when we renamed directories in #6641


igorT:
Seems odd that travis partner test started failing, restarting ci

Issue #6666, opened 2019-11-04
[SIDE QUEST] MinimumAdapterInterface Tests


runspired:
# A Side-Quest for Project Trim!

Side Quests?!?! How fun!

## Mission

#6497 introduced test infrastructure for testing ember-data without the `@ember-data/adapter` package being present at all while #6643 introduces the MinimumAdapterInterface. In this repository we want to add integration tests covering the minimum APIs.

The following requests should be tested:

**Queries**

- [x] store.findRecord calls adapter.findRecord w/correct args
- [x] store.findAll calls adapter.findAll w/correct args
- [x] store.query calls adapter.query w/correct args
- [x] store.queryRecord calls adapter.queryRecord w/correct args
- if coalesceFindRequests is true
  - [x] if findMany is not defined we call findRecord for each request and do not call groupRecordsForFindMany
  - [x] we call groupRecordsForFindMany with the right args if multiple findRecords are triggered
  - [x] we call findMany with the right args
  - [x] we do not call findRecord
- [x] if coalesceFindRequests is false or undefined, multiple calls to `store.findRecord` are not coalesced

- if a belongsTo relationship has a link but no data
  - [x] if findBelongsTo is defined we call it and not findRecord
  - [x] if findBelongsTo is not defined we throw a clear error

- if a belongsTo relationship has data but not a link
  - [x] if findBelongsTo is defined we call findRecord not findBelongsTo
  - [x] if findBelongsTo is undefined we still call findRecord not findBelongsTo (no error)

- if a belongsTo relationship has a link and data
  - [x] if findBelongsTo is defined we call it and not findRecord
  - [x] if findBelongsTo is not defined we call findRecord (no error)

- if a hasMany relationship has a link but no data
  - [x] if findHasMany is defined we call it and not findRecord OR findMany
  - [x] if findHasMany is not defined we throw a clear error

- if a hasMany relationship has data but not a link
  - [x] if findHasMany is defined and coalescing is off we call findRecord for each record
  - [x] if findHasMany is defined and coalescing is on we call findMany
  - [x] if findHasMany is undefined and coalescing is off we call findRecord for each record (no error)
  - [x] if findHasMany is undefined and coalescing is on we call findMany (no error)

- if a hasMany relationship has a link and data
  - [x] if findHasMany is defined we call it and not findRecord or findMany
  - [x] if findHasMany is not defined and coalescing is off we call findRecord for each record (no error)
[...]

PR #6673, opened 2019-11-05
[BUGFIX TEST] reloading tests for adapter encapsulation side quest - shouldReloadAll

Gaurav0:
For #6666 

It additionally contains fixes for when adapters do not implement `shouldReloadAll` and `shouldBackgroundReloadAll`

### DIFF
--- a/packages/store/addon/-private/system/core-store.ts
+++ b/packages/store/addon/-private/system/core-store.ts
@@ -2245,7 +2245,7 @@ abstract class CoreStore extends Service {
     if (options.reload !== false) {
       if (
         (adapter.shouldReloadAll && adapter.shouldReloadAll(this, snapshotArray)) ||
-        (!adapter.shouldReloadAll && snapshotArray.length > 0)
+        (!adapter.shouldReloadAll && snapshotArray.length === 0)
       ) {
         set(array, 'isUpdating', true);
         return promiseArray(_findAll(adapter, this, modelName, options));
--- a/packages/store/addon/-private/system/core-store.ts
+++ b/packages/store/addon/-private/system/core-store.ts
@@ -2242,7 +2242,7 @@ abstract class CoreStore extends Service {
 
     let snapshotArray = array._createSnapshot(options);
 
-    if (adapter.shouldReloadAll(this, snapshotArray)) {
+    if (options.reload !== false && adapter.shouldReloadAll(this, snapshotArray)) {
       set(array, 'isUpdating', true);
       return promiseArray(_findAll(adapter, this, modelName, options));
     }
--- a/packages/store/addon/-private/system/core-store.ts
+++ b/packages/store/addon/-private/system/core-store.ts
@@ -2251,7 +2251,7 @@ abstract class CoreStore extends Service {
       return promiseArray(Promise.resolve(array));
     }
 
-    if (options.backgroundReload || adapter.shouldBackgroundReloadAll(this, snapshotArray)) {
+    if (options.backgroundReload || adapter.shouldBackgroundReloadAll && adapter.shouldBackgroundReloadAll(this, snapshotArray)) {
       set(array, 'isUpdating', true);
       _findAll(adapter, this, modelName, options);
     }
--- a/packages/store/addon/-private/system/core-store.ts
+++ b/packages/store/addon/-private/system/core-store.ts
@@ -2242,9 +2242,11 @@ abstract class CoreStore extends Service {
 
     let snapshotArray = array._createSnapshot(options);
 
-    if (options.reload !== false && adapter.shouldReloadAll(this, snapshotArray)) {
-      set(array, 'isUpdating', true);
-      return promiseArray(_findAll(adapter, this, modelName, options));
+    if (options.reload !== false) {
+      if (adapter.shouldReloadAll && adapter.shouldReloadAll(this, snapshotArray) || array.length !== 0) {
+        set(array, 'isUpdating', true);
+        return promiseArray(_findAll(adapter, this, modelName, options));
+      }
     }
 
     if (options.backgroundReload === false) {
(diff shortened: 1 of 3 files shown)


packages/store/addon/-private/system/core-store.ts
```diff
@@ -2242,16 +2242,24 @@ abstract class CoreStore extends Service {
 
     let snapshotArray = array._createSnapshot(options);
 
-    if (adapter.shouldReloadAll(this, snapshotArray)) {
-      set(array, 'isUpdating', true);
-      return promiseArray(_findAll(adapter, this, modelName, options));
+    if (options.reload !== false) {
+      if (
+        (adapter.shouldReloadAll && adapter.shouldReloadAll(this, snapshotArray)) ||
+        (!adapter.shouldReloadAll && snapshotArray.length > 0)
```
runspired: The length check here is reversed: We should default to reloading only if the length is `0`. This is what ensures we fire at least one request so long as you've never loaded a record previously by another means. This is a bit of a troll (and always has been) but is the current behavior. We could improve this in the future by tracking whether we've ever made *at least* one findAll request and using that to determine whether we need to make this initial request or not.

```suggestion
        (!adapter.shouldReloadAll && snapshotRecordArray.length === 0)
```

packages/unpublished-adapter-encapsulation-test-app/tests/integration/reload-test.js
```diff
@@ -0,0 +1,173 @@
...
+  this.store = this.owner.lookup('service:store');
+  this.adapter = this.owner.lookup('adapter:application');
+}
+
+module('integration/reload - Reloading Tests', function(hooks) {
+  setupTest(hooks);
+
+  hooks.beforeEach(function() {
+    // Needed to avoid deprecation warning even though not using any transforms.
+    this.owner.register('transform:date', class DateTransform extends Transform {});
```
runspired: we don't need to register them to avoid this deprecation. There's several nice patterns but probably the best option for this package is to remove the serializer package and register a very simple custom serializer.

packages/unpublished-adapter-encapsulation-test-app/tests/integration/reload-test.js
```diff
@@ -0,0 +1,173 @@
...
+    this.owner.register('transform:number', class NumberTransform extends Transform {});
+    this.owner.register('transform:boolean', class BooleanTransform extends Transform {});
+    this.owner.register('transform:string', class StringTransform extends Transform {});
+
+    this.owner.register('service:store', Store);
+    this.owner.register('serializer:application', JSONAPISerializer);
```
runspired:
```ts
class Serializer extends EmberObject {
  normalizeResponse(_, __, data) { return data; }
}
```
^ this should be all this test needs

packages/unpublished-adapter-encapsulation-test-app/tests/integration/reload-test.js
```diff
@@ -0,0 +1,173 @@
...
+    test('adapter.shouldReloadAll is not called when store.findAll is called with a reload: true flag', async function(assert) {
+      setupReloadTest.call(this, {
+        shouldReloadAll: false,
+        shouldBackgroundReloadAll: false,
+      });
+
+      await this.store.findAll('person', { reload: true });
+
+      assert.equal(this.adapter.shouldReloadAllCalled, 0, 'shouldReloadAll is not called');
+      assert.equal(this.adapter.requestsMade, 1, 'an ajaz request is made');
```
runspired:
```suggestion
      assert.equal(this.adapter.requestsMade, 1, 'an ajax request is made');
```

packages/store/addon/-private/system/core-store.ts
Gaurav0: This is weird, but nice catch. I'll have to fix tests also.

packages/unpublished-adapter-encapsulation-test-app/tests/integration/reload-test.js
Gaurav0: That may be best, but is complex enough that it would be better to do in a separate PR. What are the other options?

packages/unpublished-adapter-encapsulation-test-app/tests/integration/reload-test.js
Gaurav0: pushPayload is also required I think. I'll try it out.

packages/store/addon/-private/system/core-store.ts
Gaurav0: fixed.

packages/unpublished-adapter-encapsulation-test-app/tests/integration/reload-test.js
Gaurav0: fixed.

packages/unpublished-adapter-encapsulation-test-app/tests/integration/reload-test.js
runspired: pushPayload shouldn't be required as we aren't using it in the tests at all

packages/unpublished-adapter-encapsulation-test-app/tests/integration/reload-test.js
runspired: After more thought I want us to keep the serializer package in dependencies, and I feel there is a bug in how we determine to throw the transform deprecation. Will investigate separately, let's leave these for now.

packages/unpublished-adapter-encapsulation-test-app/tests/integration/reload-test.js
Gaurav0: Seems to work. I thought it was needed because we call store.push.

runspired: Nope :) only store.pushPayload uses serializer.pushPayload, it is called at no other time.

PR #6676, opened 2019-11-05
[CHORE] Improve output and fail test for Asset size Check

runspired:


.github/workflows/asset-size-check.yml
```diff
@@ -14,6 +14,16 @@ jobs:
...
           node-version: 12.x
           repo-token: ${{ secrets.GITHUB_TOKEN }}
+      - name: Check SHA
+        run: |
+          sha=$(git rev-parse --short=8 HEAD)
+          echo "HEAD sha=$sha"
+          echo "GITHUB_SHA sha=$GITHUB_SHA"
+          mkdir -p tmp
+          echo $sha > tmp/sha-for-check.txt
+          originSha=$(git rev-parse HEAD^2)
```
rwjblue: O_o

runspired: TL;DR is we can’t trust $GITHUB_SHA

runspired: To summarize my findings for future wanderers:

- we've had a large number of check_run failures where `$GITHUB_SHA` was not a valid reference.
- most failures occur when additional commits are pushed to a PR
- after logging I was able to determine that the flow used by `actions/checkout` creates a merge commit if one does not exist yet at the time you checkout the repo for the run
- in both failing and successful runs, the sha being tested is not the sha of the commit to the PR branch, but rather the sha of this merge commit.

What we do here is

- get the real sha and save it for use later when we need it
- get the sha of the commit to the PR for display purposes later (the merge commit sha is meaningless to the PR author)
- log the real sha and $GITHUB_SHA so that if we encounter this problem again from a different perspective we'll have a paper trail.

I've been explaining this fail scenario to the github actions folks on twitter, and added comments about it to an open issue where folks assumed it was due to something else that we do not do.

igorT: Can you please write this comment in the yml?

Issue #6680, opened 2019-11-05
Fastboot test run throws deprecations


igorT:
```
  DEPRECATION: The <fastboot-test-app@adapter:application::ember174>#fastboot computed property was just overriden. This removes the computed property and replaces it with a plain value, and has been deprecated. If you want this behavior, consider defining a setter which does it manually. [deprecation id: computed-property.override] See https://emberjs.com/deprecations/v3.x#toc_computed-property-override for more details.
        at logDeprecationStackTrace (/var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T/broccoli-11489a1QZBnuWWstu/out-260-append_ember_auto_import_analyzer/assets/@ember/debug/lib/deprecate.js:106:1)
        at HANDLERS.(anonymous function) (/var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T/broccoli-11489a1QZBnuWWstu/out-260-append_ember_auto_import_analyzer/assets/@ember/debug/lib/handlers.js:26:1)
        at raiseOnDeprecation (/var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T/broccoli-11489a1QZBnuWWstu/out-260-append_ember_auto_import_analyzer/assets/@ember/debug/lib/deprecate.js:133:1)
        at HANDLERS.(anonymous function) (/var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T/broccoli-11489a1QZBnuWWstu/out-260-append_ember_auto_import_analyzer/assets/@ember/debug/lib/handlers.js:26:1)
        at invoke (/var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T/broccoli-11489a1QZBnuWWstu/out-260-append_ember_auto_import_analyzer/assets/@ember/debug/lib/handlers.js:38:1)
        at deprecate (/var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T/broccoli-11489a1QZBnuWWstu/out-260-append_ember_auto_import_analyzer/assets/@ember/debug/lib/deprecate.js:171:1)
        at ComputedProperty.clobberSet (/var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T/broccoli-11489a1QZBnuWWstu/out-260-append_ember_auto_import_analyzer/assets/@ember/-internals/metal/index.js:3396:1)
[...]

Issue #6679, opened 2019-11-05
Fastboot tests throw typescript errors


igorT:
Running the fastboot tests gives 
```
[96mapp/adapters/application.ts[0m:[93m1[0m:[93m25[0m - [91merror[0m[90m TS2307: [0mCannot find module '@ember-data/adapter/rest'.

[7m1[0m import RESTAdapter from '@ember-data/adapter/rest';
[7m [0m [91m                        ~~~~~~~~~~~~~~~~~~~~~~~~~~[0m

[96mapp/serializers/application.ts[0m:[93m1[0m:[93m28[0m - [91merror[0m[90m TS2307: [0mCannot find module '@ember-data/serializer/json'.

[7m1[0m import JSONSerializer from '@ember-data/serializer/json';
[7m [0m [91m                           ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~[0m

[96mapp/serializers/application.ts[0m:[93m2[0m:[93m38[0m - [91merror[0m[90m TS2307: [0mCannot find module '@ember-data/serializer/rest'.

[7m2[0m import { EmbeddedRecordsMixin } from '@ember-data/serializer/rest';
[7m [0m [91m                                     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~[0m
```

PR #6683, opened 2019-11-06
[BUGFIX RELEASE BETA CANARY] Fastboot issues

Gaurav0:
For #6679 and partial fix for #6680


packages/unpublished-fastboot-test-app/tsconfig.json
```diff
@@ -1,11 +1,34 @@
 {
   "extends": "../../tsconfig.json",
+  "baseUrl": "../..",
   "compilerOptions": {
     "paths": {
-      "fastboot-test-app/tests/*": ["tests/*"],
-      "fastboot-test-app/*": ["app/*"],
-      "*": ["types/*"]
+      "ember-data": ["packages/-ember-data/addon"],
```
runspired: why do we need to set these? does extending clobber vs merge?

runspired: For reference it does not appear that it should clobber:

```
The extends is a top-level property in tsconfig.json (alongside compilerOptions, files, include, and exclude). extends’ value is a string containing a path to another configuration file to inherit from.

The configuration from the base file are loaded first, then overridden by those in the inheriting config file. If a circularity is encountered, we report an error.

files, include and exclude from the inheriting config file overwrite those from the base config file.

All relative paths found in the configuration file will be resolved relative to the configuration file they originated in.
```

https://www.typescriptlang.org/docs/handbook/tsconfig-json.html

Gaurav0: See https://github.com/microsoft/TypeScript/issues/14527

PR #6685, opened 2019-11-06
Backport release

igorT:
* Add FastBoot test for store.createRecord

Fails at the moment.

* Enable store.createRecord in FastBoot

Changes the V4 UUID generation to leverage `FastBoot.require` when
possible. This currently requires that the host application add the
following to their `package.json`:

```
"fastbootDependencies": [
  "crypto"
]
```

* fix types

### DIFF
--- a/packages/-fastboot-test-app/app/router.ts
+++ b/packages/-fastboot-test-app/app/router.ts
@@ -1,0 +1,15 @@
+import EmberRouter from '@ember/routing/router';
+import config from './config/environment';
+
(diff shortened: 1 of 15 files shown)


Gaurav0:
The requirement for fastboot apps with ember-data should probably be documented.

Gaurav0:
Error message isn't so bad:

`Error while processing route: person.new Unable to require module 'crypto' because it was not in the whitelist.`

Issue #6690, opened 2019-11-06
ED Fastboot mode requires node crypto package but doesn't really tell users


igorT:
We need to have a good error message and maybe is there a blueprint where we can add the fastboot dependency? cc @rwjblue

runspired:
The error message portion is above. There is not currently a blueprint used when installing fastboot but if there ever is one we should have it detect the presence of `ember-data` and add `crypto` to fastbootDependencies if present. `crypto` is a native module so it does not need installed, only specified.

runspired:
Closing as there is now tracking for a blueprint in the fastboot repo.

IgnaceMaes:
I just ran into this issue again. For anyone looking to copy paste the solution, add the following in your `package.json`:

```js
{
  // dependencies, devDependencies, ...
  "fastbootDependencies": [
    "crypto"
  ],
}
```

runspired:
fastboot should also give you a much clearer error message now since the latest release contained the change to improve the error message around this

satyam73:
> I just ran into this issue again. For anyone looking to copy paste the solution, add the following in your `package.json`:
> 
> ```js
> {
>   // dependencies, devDependencies, ...
>   "fastbootDependencies": [
>     "crypto"
>   ],
> }
> ```

Thanks @IgnaceMaes , it worked!

PR #6692, opened 2019-11-06
[BUGFIX 3.13 RELEASE BETA CANARY] Better fastboot crypto error message

Gaurav0:
For issue raised in #6685

PR #6700, opened 2019-11-07
Backport #6692 to 3.13

Gaurav0:


runspired:
The accidental branch naming here is unfortunate, but I can confirm that is the 3.13 branch

runspired:
resolves #6690 for 3.13

igorT:
The label target-release seems to have been wrongly added here, removing

Issue #6588, opened 2019-10-11
adapter.headers is given precedence over options.headers


runspired:
https://github.com/emberjs/data/blob/master/packages/adapter/addon/rest.js#L1089

See this discourse thread for more info: https://discuss.emberjs.com/t/how-to-override-the-adapter-headers-that-get-sent-in-the-ajax-request/17106

runspired:
After discussing, we've decided to resolve this by:

- [ ] Submitting a fix to LTS 3.12 to warn if headers conflict and are overridden
- [ ] Submitting a fix to 3.13+ that reverses this behavior to ensure the headers from options are given precedence.

Gaurav0:
What should the warning say?

PR #6701, opened 2019-11-07
[LTS] Warn about changing header precedence in lts

Gaurav0:
For #6588

### DIFF
--- a/packages/adapter/addon/rest.js
+++ b/packages/adapter/addon/rest.js
@@ -1084,6 +1084,21 @@ const RESTAdapter = Adapter.extend(BuildURLMixin, {
 
     let headers = get(this, 'headers');
     if (headers !== undefined) {
+      if (DEBUG) {
+        options.headers = options.headers || {};
+        let conflictingHeaders = Object.keys(headers).filter(key => options.headers[key] !== undefined);
+        if (conflictingHeaders.length > 0) {
+          conflictingHeaders.forEach(header => {
+            let value1 = headers[header];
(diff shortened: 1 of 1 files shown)


packages/adapter/addon/rest.js
```diff
@@ -1084,6 +1084,17 @@ const RESTAdapter = Adapter.extend(BuildURLMixin, {
 
     let headers = get(this, 'headers');
     if (headers !== undefined) {
+      if (DEBUG) {
+        let optionsHeaders = Object.keys(options.headers || {});
+        let conflictingHeaders = Object.keys(headers).filter(key => optionsHeaders.includes(key));
+        if (conflictingHeaders.length > 0) {
+          conflictingHeaders.forEach(header => {
+            warn(
+              `${header} defined in adapter will be overridden by ${header} defined in ajaxOptions argument options in the next version of Ember Data.`
```
runspired: Two things:

1) I think if the values match we shouldn't warn.
2) I think we should print the values.

So something like: `if (optionsHeaders[header] !== conflictingHeaders[header]) { ... }`

```suggestion
              `The request header \`${header}\` with the value \`${optionsHeaders[header]}\` defined on the options passed to \`adapter.ajaxOptions(url, method, options)\` will be overridden with the value \`${conflictingHeaders[header]}\` defined via \`adapter.headers\`.  This behavior will change in the next version of EmberData and the value provided on the \`options\` passed to \`adapter.ajaxOptions\` will be respected.
```

PR #6702, opened 2019-11-08
[BREAKING BUGFIX] eliminate infrastructure for filter record arrays

runspired:
We deprecated filters ages ago and completed killing them off after 3.4. This is the scheduled removal of some secret notification infra we had left in place to keep them working a tad longer if needed.

**⚠️ WARNING**

This infrastructure had the unintended bug of making us notify array `length` and `[]` over-eagerly.
Due to this, the change **MAY** break apps that were using observers or computed properties on the `length` or `[]` properties of `RecordArray`s when intending to observe changes to properties on records inside these arrays.

For such cases there is an easy refactor to more standard Ember computed property usage ala `allRecords.@each.{aProp,bProp}`. This risk may be lower in versions of Ember in which @tracked is in use.

### DIFF
--- a/packages/-ember-data/tests/unit/store/asserts-test.js
+++ b/packages/-ember-data/tests/unit/store/asserts-test.js
@@ -73,7 +73,6 @@ module('unit/store/asserts - DS.Store methods produce useful assertion messages'
     '_push',
     'pushPayload',
     'normalize',
-    'recordWasLoaded',
     'adapterFor',
     'serializerFor',
   ];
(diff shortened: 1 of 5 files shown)


runspired:
There are 6 travis failures, all in one module, that are not the usual travis-ci flake. Will investigate.

runspired:
Fixed travis upstream, they had implemented a custom filter that (like the deprecated and removed filter) depended on this infra. They now use computed properties observing peekAll's result with more typical `@each` chaining

- https://github.com/travis-ci/travis-web/pull/2222
- https://github.com/travis-ci/travis-web/pull/2223

runspired:
While there is some risk to apps that fixing this will cause minor breakages, the fixes to app code when this occurs are simple, and in apps >= ember-source 3.13 there may not even be this risk due to tracked and autotracking.

After considering whether this is a risk we should take, we're going to merge this, as the effort to attempt a deprecation would be large and the refactoring and improvements it hinders are big.

PR #6710, opened 2019-11-09
[CHORE] only fix deprecated registrations when using legacy ember-data

runspired:


packages/store/addon/-private/system/core-store.ts
```diff
@@ -296,7 +297,7 @@ abstract class CoreStore extends Service {
     }
 
     if (DEBUG) {
-      if (HAS_SERIALIZER_PACKAGE) {
+      if (HAS_DS_META_PACKAGE && HAS_SERIALIZER_PACKAGE) {
```
Gaurav0: This turns off autoregistration of transforms in applications without ember-data package, Is that intended?

It won't break existing apps, so ok with me as long as it is what you want.

runspired: yes this is what I intend here. Basically, if you are using the `@ember-data` packages directly you'll need to have already upgraded your tests or manually specify imports you want (either or). E.g. support the deprecated behavior only in the primary package (ember-data).

PR #6711, opened 2019-11-10
[BUGFIX TEST] reloading tests for adapter encapsulation side quest - shouldBackgroundReloadAll

Gaurav0:
For #6666

### DIFF
--- a/packages/unpublished-adapter-encapsulation-test-app/tests/integration/reload-test.js
+++ b/packages/unpublished-adapter-encapsulation-test-app/tests/integration/reload-test.js
@@ -172,7 +172,61 @@ module('integration/reload - Reloading Tests', function(hooks) {
     });
   });
 
-  module('adapter.shouldBackgroundReloadAll', function() {});
+  module('adapter.shouldBackgroundReloadAll', function() {
+    test('adapter.shouldBackgroundReloadAll is not called called when store.findAll is called with reload: true flag (but we do make request)', async function(assert) {
+      setupReloadTest.call(this, {});
+
+      await this.store.findAll('person', { reload: true });
+
+      assert.equal(this.adapter.shouldBackgroundReloadAllCalled, 0, 'shouldBackgroundReloadAll not called');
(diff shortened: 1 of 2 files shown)


packages/unpublished-adapter-encapsulation-test-app/tests/integration/reload-test.js
```diff
@@ -172,7 +172,84 @@ module('integration/reload - Reloading Tests', function(hooks) {
...
+        shouldBackgroundReloadAll: true,
+      });
+
+      await this.store.findAll('person');
+
+      assert.equal(this.adapter.shouldBackgroundReloadAllCalled, 1, 'shouldBackgroundReloadAll is called');
+      assert.equal(this.adapter.requestsMade, 1, 'an ajax request is made');
+    });
+
+    test('adapter.shouldBackgroundReloadAll is called when store.findAll is called and there is no backgroundReload flag (returns true', async function(assert) {
```
runspired:
```suggestion
    test('adapter.shouldBackgroundReloadAll is called when store.findAll is called and there is no backgroundReload flag (returns false)', async function(assert) {
```

packages/unpublished-adapter-encapsulation-test-app/tests/integration/reload-test.js
```diff
@@ -172,7 +172,84 @@ module('integration/reload - Reloading Tests', function(hooks) {
...
+      setupReloadTest.call(this, {
+        shouldReloadAll: false,
+      });
+
+      await this.store.findAll('person');
+
+      assert.equal(this.adapter.requestsMade, 1, 'an ajax request is made');
+    });
+
+    test('adapter.shouldBackgroundReloadAll is called when store.findAll is called and there is no backgroundReload flag (returns true', async function(assert) {
```
runspired:
```suggestion
    test('adapter.shouldBackgroundReloadAll is called when store.findAll is called and there is no backgroundReload flag (returns true)', async function(assert) {
```

PR #6712, opened 2019-11-10
[CHORE] refactor: Remove runloop usage in integration tests for store/ find All module

dmuneras:
## Description

- [X] Remove usage of runloop in store - findAll module tests.
- [X] Remove beforeEach loop usage in the module and setup data in each test.

### DIFF
--- a/packages/-ember-data/tests/integration/store-test.js
+++ b/packages/-ember-data/tests/integration/store-test.js
@@ -696,7 +696,9 @@ module('integration/store - findAll', function(hooks) {
 
     assert.equal(cars.length, 1, 'Store resolves with the existing records');
 
-    await resolvefindAll();
+    resolvefindAll();
+
+    await settled();
 
     cars = store.peekAll('car');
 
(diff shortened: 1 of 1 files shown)


packages/-ember-data/tests/integration/store-test.js
```diff
@@ -913,19 +913,15 @@ module('integration/store - findAll', function(hooks) {
 
     assert.equal(cars.length, 2, 'It returns all cars');
 
-    let carsInStore = store.peekAll('car');
+    await settled();
```
runspired: I think we want to structure this a bit differently:

```js
carsPromiseArray = store.findAll('car');

assert.equal(carsPromiseArray.length, 2, 'It returns all cars');
let mini = cars.findBy('id', '1');
assert.equal(mini.model, 'Mini', 'Records have not yet been updated');	    

await carsPromiseArray;
```

after this we would peek and assert the length and ensure that the model has been updated

PR #6713, opened 2019-11-10
[CHORE] refactor: Remove runloop usage in integration/store-tests module deleteRecord

dmuneras:
## Description

- [X] Remove usage of runloop in store - deleteRecord module tests.
- [X] Remove beforeEach loop usage in the module and setup data in each test.

### DIFF
--- a/packages/-ember-data/tests/integration/store-test.js
+++ b/packages/-ember-data/tests/integration/store-test.js
@@ -990,53 +990,51 @@ module('integration/store - findAll', function(hooks) {
 module('integration/store - deleteRecord', function(hooks) {
   setupTest(hooks);
 
-  hooks.beforeEach(function() {
-    this.owner.register('model:person', Person);
-    this.owner.register('model:car', Car);
-    this.owner.register('adapter:application', RESTAdapter.extend());
-    this.owner.register('serializer:application', RESTSerializer.extend());
-  });
-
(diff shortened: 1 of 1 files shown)


packages/-ember-data/tests/integration/store-test.js
```diff
@@ -990,53 +990,51 @@ module('integration/store - findAll', function(hooks) {
...
+        attributes: {
+          name: 'Tom Dale',
         },
-      });
-      person = store.peekRecord('person', 1);
+      },
     });
 
-    assert.ok(store.hasRecordForId('person', 1), 'expected the record to be in the store');
+    person = store.peekRecord('person', '1');
```
runspired: Just a note when using push for a single-resource-document we can just use the returned record vs following with a peek. e.g.

```js
const person = store.push({ data: { ... } });
```

packages/-ember-data/tests/integration/store-test.js
```diff
@@ -990,53 +990,51 @@ module('integration/store - findAll', function(hooks) {
...
     let store = this.owner.lookup('service:store');
 
-    run(() => {
-      store.push({
-        data: null,
-      });
-    });
+    try {
+      store.push({ data: null });
+    } catch (_error) {
```
runspired: I love this improvement :)

PR #6715, opened 2019-11-11
[CHORE]: Extract internalModel access to identifiers in RecordArray modules

snewcomer:
This is to reduce the need of `InternalModel` throughout the codebase.  This is possible because we have an identifier cache available to lookup records that were populated locally.  This doesn't encompass all the spots where an identifier can be swapped for an InternalModel, but is a start.

Modifications include Record-Array, Record-Array-Manager, Adapter-Populated-Record-Array and Core-Store. 

1. Method calls on RecordArray give identifiers instead of internalModels
2. Calls to Record-Array-Manager pass identifiers and expects to receive identifiers.

ref #6166


packages/-ember-data/tests/integration/record-arrays/adapter-populated-record-array-test.js
```diff
@@ -107,9 +107,8 @@ module('integration/record-arrays/adapter_populated_record_array - AdapterPopula
       },
     };
 
-    run(() => {
-      recordArray._setInternalModels(store._push(payload), payload);
-    });
+    let results = store.push(payload);
+    recordArray._setIdentifiers(results.map(r => r._internalModel.identifier), payload);
```
snewcomer: It might be worth refactoring `getRecord()` to return out an identifier as a top level property.

packages/-ember-data/tests/integration/relationships/has-many-test.js
```diff
@@ -3679,75 +3679,70 @@ module('integration/relationships/has_many - Has-Many Relationships', function(h
       user: belongsTo('user', { inverse: null, async: false }),
     });
 
-    run(() => {
```
snewcomer: just reminding to revisit this...

packages/store/addon/-private/system/model/internal-model.ts
```diff
@@ -122,7 +122,6 @@ export default class InternalModel {
   __recordData: RecordData | null;
   _isDestroyed: boolean;
   isError: boolean;
-  _pendingRecordArrayManagerFlush: boolean;
```
snewcomer: this state was only used in record-array-manager, so put it there with a `Set` of pending identifiers.

packages/store/addon/-private/system/record-array-manager.js
```diff
@@ -11,6 +11,25 @@ import { RecordArray, AdapterPopulatedRecordArray } from './record-arrays';
 import { internalModelFactoryFor } from './store/internal-model-factory';
 
 const emberRun = emberRunloop.backburner;
+const pendingForIdentifier = new Set([]);
+
+function recordArraysForIdentifier(store, identifier) {
+  let internalModel = internalModelFactoryFor(store).peek(identifier);
+  if (internalModel) {
```
snewcomer: There was only one failing test that required this guard.  `disconnectRecord` in `store-wrapper-test`.  It is pushing with `data: []` and `included: [houseHash]`.  Is this possible?

packages/store/addon/-private/system/record-arrays/record-array.js
```diff
@@ -9,6 +9,15 @@ import { Promise } from 'rsvp';
...
 import SnapshotRecordArray from '../snapshot-record-array';
+import { internalModelFactoryFor } from '../store/internal-model-factory';
+
+function recordForIdentifier(store, identifier) {
+  return internalModelForIdentifier(store, identifier).getRecord();
+}
+
+function internalModelForIdentifier(store, identifier) {
+  return internalModelFactoryFor(store).lookup(identifier);
+}
```
snewcomer: globally available utility functions or are you ok leaving them per test module?

packages/-ember-data/tests/integration/record-arrays/adapter-populated-record-array-test.js
runspired: a helper function for tests I think exists for getting the identifier from the record

packages/store/addon/-private/system/record-arrays/record-array.js
runspired: per test module seems fine for now, as I'd like to kill off these helpers as well in a follow up.

runspired:
The M3 failure is because M3 registers query-arrays (result of M3's queryURL method) with the record-array-manager. There's a few paths forward here:

- Intercept Approach
  - detect internalModels during creation and convert them
  - implement legacy _pushIM/_setIM methods and convert to identifiers within
- Refactor M3
  - for the non-custom-model-class branch, move M3 to using identifiers or off of RecordArrayManager
- Wait for M3
  - once the custom-model-class FF is on, M3 no longer uses the RecordArrayManager

packages/store/addon/-private/system/record-array-manager.js
runspired: Turns out M3 hit this case so I had to make some adjustments, roughly speaking it is possible in the current state of things to try to sync-destroy during push via this API. I'm going to bring this up in a meeting as EmberData does not do a sync-destroy here itself and we likely should not allow it going forward as the "syncLiveArray" code is undesirable complexity that remains from our need to handle observers greedily firing when we begin changes.

We may be able to refactor overtime to avoid delivering changes to the UI until a single "ui flush" which would help us avoid this problem.

packages/store/addon/-private/system/model/internal-model.ts
```diff
@@ -482,9 +470,9 @@ export default class InternalModel {
     }
 
     // move to an empty never-loaded state
+    this.updateRecordArrays();
```
igorT: Why did we have to move this? The update now might not have the correct state?

packages/store/addon/-private/system/record-array-manager.js
```diff
@@ -9,8 +9,59 @@ import { run as emberRunloop } from '@ember/runloop';
 import { assert } from '@ember/debug';
 import { RecordArray, AdapterPopulatedRecordArray } from './record-arrays';
 import { internalModelFactoryFor } from './store/internal-model-factory';
+import isStableIdentifier from '../identifiers/is-stable-identifier';
+import { CUSTOM_MODEL_CLASS } from '@ember-data/canary-features';
 
 const emberRun = emberRunloop.backburner;
+const pendingForIdentifier = new Set([]);
+const IMFallback = new WeakMap();
+const RecordArraysCache = new WeakMap(); // store StableIdentifier => Set[RecordArray[]]
```
igorT: Can we move the comment above the line please

packages/store/addon/-private/system/model/internal-model.ts
runspired: I moved it when trying to determine if it's order was causing an issue in M3 (it wasn't). But I also realized it's current order is problematic as we notify observers on the record-array that the record has changed before we actually settle into the state we are moving to.

runspired:
After reviewing with @igorT we decided we will

- introduce a feature flag for `RECORD_ARRAY_IDENTIFIERS`, when off this PR should in effect have "no change"
- introduce a feature flag for `DEPRECATED_PRIVATE_RECORD_ARRAY_SUPPORT`, which when RECORD_ARRAY_IDENTIFIERS is on and DEPRECATED_PRIVATE_RECORD_ARRAY_SUPPORT is off has the effect of removing our legacy support for private API hacks used to add custom record arrays (such as what M3 does). 
- update this PR to make use of these flags
- add some tests for private API hacks that use the record array manager
- land this PR then land a PR turning both the new flags on.
- at a later time (likely around the time that CUSTOM_MODEL_CLASS flag is activated) turn the DEPRECATED_PRIVATE_RECORD_ARRAY_SUPPORT off.

packages/-ember-data/tests/integration/record-arrays/adapter-populated-record-array-test.js
```diff
@@ -18,7 +19,7 @@ const Person = Model.extend({
   },
 });
 
-module('integration/record-arrays/adapter_populated_record_array - AdapterPopulatedRecordArray', function(hooks) {
+module('scott integration/record-arrays/adapter_populated_record_array - AdapterPopulatedRecordArray', function(hooks) {
```
runspired: don't forget the `Scott` removal

snewcomer:
@runspired I believe this PR is ready to review.  All tests passed with both feature flags on and off!!

packages/canary-features/addon/default-features.ts
```diff
@@ -177,4 +177,6 @@ export default {
   REQUEST_SERVICE: null,
   CUSTOM_MODEL_CLASS: null,
   FULL_LINKS_ON_RELATIONSHIPS: true,
+  RECORD_ARRAY_MANAGER_IDENTIFIERS: false,
```
igorT, line 180: These should be null

packages/store/addon/-private/system/record-array-manager.js
```diff
@@ -8,347 +8,804 @@ import { get, set } from '@ember/object';
...
+  };
 
-  if (modelsToAdd.length > 0) {
-    array._pushInternalModels(modelsToAdd);
-  }
-  if (modelsToRemove.length > 0) {
-    array._removeInternalModels(modelsToRemove);
-  }
-}
+  const pushIdentifiers = function(recordArray, identifiers, cache) {
```
igorT, line 775: Can we name this function please

packages/store/addon/-private/system/record-array-manager.js
```diff
@@ -8,347 +8,804 @@ import { get, set } from '@ember/object';
...
+// only used by RECORD_ARRAY_MANAGER_IDENTIFIERS
+import isStableIdentifier from '../identifiers/is-stable-identifier';
 import { AdapterPopulatedRecordArray, RecordArray } from './record-arrays';
 import { internalModelFactoryFor } from './store/internal-model-factory';
 
+const RecordArraysCache = new WeakMap();
 const emberRun = emberRunloop.backburner;
+let RecordArrayManager;
+
+export let associateWithRecordArray;
```
igorT, line 22: Lets export this after it's been assigned

packages/store/addon/-private/system/record-array-manager.js
```diff
@@ -8,347 +8,804 @@ import { get, set } from '@ember/object';
...
+      this.isDestroyed = true;
+    }
+
+    destroy() {
+      this.isDestroying = true;
+      emberRun.schedule('actions', this, this.willDestroy);
+    }
+  };
+
+  const destroy = function(entry) {
```
igorT, line 314: Lets bring these back into the named style, and see if we run into issues

packages/store/addon/-private/system/record-array-manager.js
```diff
@@ -8,347 +8,804 @@ import { get, set } from '@ember/object';
...
+    const recordArrays = internalModel._recordArrays;
+
+    recordArrays.forEach(function(recordArray) {
+      recordArray._removeInternalModels([internalModel]);
+    });
+
+    recordArrays.clear();
+  };
+
+  associateWithRecordArray = function assocWithRecordArray(internalModels, array) {
```
igorT, line 375: should we maybe name `assocWithRecordArray` something more like `legacyAssocWithRecordArray`

packages/store/addon/-private/system/record-array-manager.js
```diff
@@ -8,347 +8,804 @@ import { get, set } from '@ember/object';
...
+    for (let i = 0, l = internalModels.length; i < l; i++) {
+      let internalModel = internalModels[i];
+      internalModel._recordArrays.add(array);
+    }
+  };
+} else {
+  const emberRun = emberRunloop.backburner;
+  const pendingForIdentifier = new Set([]);
```
igorT, line 383: These should probably be typed

packages/store/addon/-private/system/record-array-manager.js
```diff
@@ -8,347 +8,804 @@ import { get, set } from '@ember/object';
...
+  const IMFallback = new WeakMap();
+  // store StableIdentifier => Set[RecordArray[]]
```
igorT, line 385: Leftover code

packages/store/addon/-private/system/record-array-manager.js
```diff
@@ -8,347 +8,804 @@ import { get, set } from '@ember/object';
...
+
+  const getIdentifier = function(identifier) {
```
igorT, line 387: lets name this argument `identfierOrInternalModel`

packages/store/addon/-private/system/record-array-manager.js
```diff
@@ -8,347 +8,804 @@ import { get, set } from '@ember/object';
...
+  };
 
-  _visibleInternalModelsByType(modelName) {
-    let all = internalModelFactoryFor(this.store).modelMapFor(modelName)._models;
-    let visible = [];
-    for (let i = 0; i < all.length; i++) {
-      let model = all[i];
-      if (model.isHiddenFromRecordArrays() === false) {
-        visible.push(model);
+  const peek = function(cache, identifier) {
```
igorT, line 400: @igorT go back to this

packages/store/addon/-private/system/record-array-manager.js
```diff
@@ -8,347 +8,804 @@ import { get, set } from '@ember/object';
...
 
-  if (modelsToAdd.length > 0) {
-    array._pushInternalModels(modelsToAdd);
-  }
-  if (modelsToRemove.length > 0) {
-    array._removeInternalModels(modelsToRemove);
-  }
-}
+  const pushIdentifiers = function(recordArray, identifiers, cache) {
+    if (RECORD_ARRAY_MANAGER_LEGACY_COMPAT && !recordArray._pushIdentifiers) {
```
snewcomer, line 776: As discussed with Igor -

r/RECORD_ARRAY_MANAGER_LEGACY_COMPAT/REMOVE_RECORD_ARRAY_MANAGER_LEGACY_COMPAT 

(or something else/shorter.  Taking recommendations for a name!)

The problem - with the current name, we would turn from `false` to `true` (when `RECORD_ARRAY_MANAGER_IDENTIFIERS` is turned on)  and then finally `false` to indicate this block of code can be removed at some point.  This doesn't follow how the other feature flags work. 

So the path forward - `RECORD_ARRAY_MANAGER_IDENTIFIERS` will be turned to `true`.  However, this flag will still be `false`, thus this block of code will still apply for some period of time.  When we turn the flag to `true`, we are indicating we want to remove this block of code.  This should be a safe operation b/c the flag in question is effectively a no-op without `RECORD_ARRAY_MANAGER_IDENTIFIERS` on.

This should look something like...

```
if (!REMOVE_RECORD_ARRAY_MANAGER_LEGACY_COMPAT && !recordArray._pushIdentifiers) {
```

packages/store/addon/-private/system/record-array-manager.js
snewcomer: Since we have this assigned in a conditional block, it only allows us to import/export at the top level.

packages/store/addon/-private/system/core-store.ts
```diff
@@ -2657,7 +2658,12 @@ abstract class CoreStore extends Service {
     internalModel.setupData(data);
 
     if (!isUpdate) {
-      this.recordArrayManager.recordDidChange(internalModel);
+      // TODO: do we want to maintain old API with this flag?
```
igorT, line 2661: This seems like a leftover comment

packages/store/addon/-private/system/core-store.ts
```diff
@@ -2657,7 +2658,12 @@ abstract class CoreStore extends Service {
...
+      if (RECORD_ARRAY_MANAGER_IDENTIFIERS) {
+        this.recordArrayManager.recordDidChange(internalModel.identifier);
```
igorT, line 2663: Given that we use the identifier on line 2644, we could extract the `let identifier`

packages/store/addon/-private/system/internal-model-map.ts
```diff
@@ -42,6 +44,10 @@ export default class InternalModelMap {
     return this._models.length;
   }
 
+  get modelIdentifiers(): StableRecordIdentifier[] {
```
igorT, line 47: we should probably call this recordIdentifiers. We also only use it in a single place, maybe it's not worth creating a method for it?

packages/store/addon/-private/system/model/internal-model.ts
```diff
@@ -1363,8 +1359,14 @@ export default class InternalModel {
     @private
   */
   updateRecordArrays() {
-    // @ts-ignore: Store is untyped and typescript does not detect instance props set in `init`
-    this.store.recordArrayManager.recordDidChange(this);
+    // TODO: do we want to maintain old API with this flag?
```
igorT, line 1362: Leftover comment

packages/store/addon/-private/system/model/internal-model.ts
```diff
@@ -1363,8 +1359,14 @@ export default class InternalModel {
...
+    if (RECORD_ARRAY_MANAGER_IDENTIFIERS) {
+      // @ts-ignore: Store is untyped and typescript does not detect instance props set in `init`
```
igorT, line 1364: This `ts-ignore` is slightly incorrect, it seems we just need to expose recordArrayManager from CoreStore

packages/store/addon/-private/system/record-array-manager.js
```diff
@@ -8,347 +8,814 @@ import { get, set } from '@ember/object';
 import { assign } from '@ember/polyfills';
 import { run as emberRunloop } from '@ember/runloop';
 
+import {
+  RECORD_ARRAY_MANAGER_IDENTIFIERS,
+  REMOVE_RECORD_ARRAY_MANAGER_LEGACY_COMPAT,
+} from '@ember-data/canary-features';
+
+// only used by RECORD_ARRAY_MANAGER_IDENTIFIERS
```
igorT, line 16: Lets remove this comment

packages/store/addon/-private/system/record-array-manager.js
```diff
@@ -8,347 +8,814 @@ import { get, set } from '@ember/object';
...
+      this.isDestroying = false;
+      this.isDestroyed = false;
+      this._liveRecordArrays = Object.create(null);
+      this._pending = Object.create(null);
+      this._adapterPopulatedRecordArrays = [];
+    }
+
+    /**
+     * @method recordDidChange
+     * @public
```
igorT, line 54: I don't think this is meant to be public, no?

packages/store/addon/-private/system/record-array-manager.js
```diff
@@ -8,347 +8,814 @@ import { get, set } from '@ember/object';
...
-
-    let internalModels = this._visibleInternalModelsByType(modelName);
+    /**
+      Create a `AdapterPopulatedRecordArray` for a modelName with given query.
+
+      @method createAdapterPopulatedRecordArray
+      @param {String} modelName
+      @param {Object} query
+      @param {Array} internalModels
+      @param {Object} payload
```
igorT, line 238: ❤️

packages/store/addon/-private/system/record-array-manager.js
```diff
@@ -8,347 +8,814 @@ import { get, set } from '@ember/object';
...
+      this.isDestroyed = true;
+    }
+
+    destroy() {
+      this.isDestroying = true;
+      emberRun.schedule('actions', this, this.willDestroy);
+    }
+  };
+
+  const destroy = function destroyEntry(entry) {
```
igorT, line 317: Don't need the const/should make the names consistent

packages/store/addon/-private/system/record-array-manager.js
```diff
@@ -8,347 +8,814 @@ import { get, set } from '@ember/object';
...
+    entry.destroy();
+  };
+
+  const remove = function removeEntry(array, item) {
```
igorT, line 321: Don't need the const/should make the names consistent

packages/store/addon/-private/system/record-arrays/adapter-populated-record-array.js
```diff
@@ -74,14 +74,46 @@ export default RecordArray.extend({
...
+});
+
+if (RECORD_ARRAY_MANAGER_IDENTIFIERS) {
+  /**
+    @method _setIdentifiers
+    @param {StableRecordIdentifier[]} identifiers
+    @param {Object} payload normalized payload
+    @internal
+  */
+  AdapterPopulatedRecordArray.prototype._setIdentifiers = function _setIdentifiers(identifiers, payload) {
```
igorT, line 86: Seems safer to extend here

packages/store/addon/-private/system/record-arrays/adapter-populated-record-array.js
```diff
@@ -74,14 +74,46 @@ export default RecordArray.extend({
...
+      isLoaded: true,
+      isUpdating: false,
+      meta: assign({}, payload.meta),
+      links: assign({}, payload.links),
+    });
 
+    this.manager._associateWithRecordArray(identifiers, this);
+
+    if (DEPRECATE_EVENTED_API_USAGE) {
+      const _hasDidLoad = DEBUG ? this._has('didLoad') : this.has('didLoad');
```
igorT, line 102: Switch with `let`

packages/store/addon/-private/system/record-arrays/adapter-populated-record-array.js
igorT, line 86: function setObjects(identifierOrInternalModel) {
    // TODO: initial load should not cause change events at all, only
    // subsequent. This requires changing the public api of adapter.query, but
    // hopefully we can do that soon.
    this.get('content').setObjects(identifiers);

    this.setProperties({
      isLoaded: true,
      isUpdating: false,
      meta: assign({}, payload.meta),
      links: assign({}, payload.links),
    });

    this.manager._associateWithRecordArray(identifiers, this);

    if (DEPRECATE_EVENTED_API_USAGE) {
      const _hasDidLoad = DEBUG ? this._has('didLoad') : this.has('didLoad');
      if (_hasDidLoad) {
        // TODO: should triggering didLoad event be the last action of the runLoop?
        once(this, 'trigger', 'didLoad');
      }
    }
  };
}

if (RECORD_ARRAY_MANAGER_IDENTIFIERS) {
  AdapterPopulatedRecordArray.prototype._setIdentifiers = function _setIdentifiers(identifiers, payload) {
    function setObjects(identifiers, payload)l
  } 
} else {
  AdapterPopulatedRecordArray.prototype._setInternalModels = function _setInternalModels(internalModels, payload) {
    function setObjects(identifiers, payload)l

}
}

packages/store/addon/-private/system/record-arrays/record-array.js
```diff
@@ -7,9 +7,20 @@ import { DEBUG } from '@glimmer/env';
...
 import DeprecatedEvented from '../deprecated-evented';
 import { PromiseArray } from '../promise-proxies';
 import SnapshotRecordArray from '../snapshot-record-array';
+import { internalModelFactoryFor } from '../store/internal-model-factory';
+
+function recordForIdentifier(store, identifier) {
+  return internalModelForIdentifier(store, identifier).getRecord();
+}
+
+function internalModelForIdentifier(store, identifier) {
```
igorT, line 21: We probably don't need this helper function

packages/store/addon/-private/system/record-arrays/record-array.js
```diff
@@ -113,8 +124,13 @@ export default ArrayProxy.extend(DeprecatedEvented, {
     @return {Model} record
   */
   objectAtContent(index) {
-    let internalModel = get(this, 'content').objectAt(index);
-    return internalModel && internalModel.getRecord();
+    let identifier = get(this, 'content').objectAt(index);
```
igorT, line 127: this might read nicer if 
```
    if (RECORD_ARRAY_MANAGER_IDENTIFIERS) {
        let identifier = get(this, 'content').objectAt(index);
      return identifier ? recordForIdentifier(this.store, identifier) : undefined;
    } else {
      return identifier ? identifier.getRecord() : undefined;
    }
```

packages/store/addon/-private/system/record-arrays/record-array.js
```diff
@@ -257,12 +238,97 @@ export default ArrayProxy.extend(DeprecatedEvented, {
     // this is private for users, but public for ember-data internals
     return new SnapshotRecordArray(this, this.get('meta'), options);
   },
+});
 
-  /*
+if (RECORD_ARRAY_MANAGER_IDENTIFIERS) {
```
igorT, line 243: Should probably be extending here

packages/unpublished-fastboot-test-app/app/templates/application.hbs
```diff
@@ -2,4 +2,4 @@
 
 {{outlet}}
 
-<a href="/tests">Tests</a>
+<a href={{"/tests"}}>Tests</a>
```
igorT, line 5: Lets try removing this

packages/store/addon/-private/system/record-array-manager.js
```diff
@@ -8,347 +8,814 @@ import { get, set } from '@ember/object';
...
+// only used by RECORD_ARRAY_MANAGER_IDENTIFIERS
+import isStableIdentifier from '../identifiers/is-stable-identifier';
 import { AdapterPopulatedRecordArray, RecordArray } from './record-arrays';
 import { internalModelFactoryFor } from './store/internal-model-factory';
 
+const RecordArraysCache = new WeakMap();
 const emberRun = emberRunloop.backburner;
+let RecordArrayManager;
+
+export let associateWithRecordArray;
```
igorT, line 25: Lets leave a comment that we should remove this after the FF land

packages/store/addon/-private/system/record-arrays/adapter-populated-record-array.js
snewcomer: Done!  I think the original intention was to work on the object directly instead of subclassing but this works too!  Aligns closer to the semantics of defining objects in this file as well.  

One question though - When it comes to switching this file over to TS though, would we modify the prototype or subclass?

PR #6727, opened 2019-11-12
polymorphic relations and native classes mixins bug test

yratanov:
Broken test example for issue 

ISSUE: #6728


yratanov:
Test output: 
`Error: Assertion Failed: The 'video' type does not implement 'message' and thus cannot be assigned to the 'messages' relationship in 'user'. Make it a descendant of 'message' or use a mixin of the same name`

runspired:
Ported to #6767 which has a fix

PR #6734, opened 2019-11-13
[CHORE] build speed improvements & docs path fix

runspired:
- Only build docs automatically for publishing vs for all production builds
- Don't uglify app.js, tests.js, and test-support.js for testing
- adds `record-data` and `debug` packages to the docs generation script

Note we weren't publishing any of the built artifacts other than one docs file already

*.npmignore*
```
/dist/
/dist/**/*
# whitelist yuidoc's data.json for api docs generation
!/dist/docs/data.json
```


packages/-ember-data/node-tests/fixtures/expected.js
```diff
@@ -252,5 +252,15 @@ module.exports = {
     'pushPayload [OPTIONAL]',
     'serializeIntoHash [OPTIONAL]',
     'normalize [OPTIONAL]',
+    'columnNameToDesc',
```
igorT: What caused this change?

packages/-ember-data/package.json
```diff
@@ -11,14 +11,14 @@
...
+    "docs": "mkdir -p dist && ember ember-cli-yuidoc",
     "test": "ember test",
     "test:all": "ember try:each",
     "test:production": "ember test -e production",
     "test:optional-features": "ember test",
     "test:try-one": "ember try:one",
-    "test:docs": "qunit node-tests/docs/test-coverage.js",
-    "prepublishOnly": "ember build --environment=production"
+    "test:docs": "mkdir -p dist && qunit node-tests/docs/test-coverage.js",
+    "prepublishOnly": "ember ember-cli-yuidoc"
```
igorT: How come docs and test docs need to create a new folder but prepublish doesnt?

packages/-ember-data/node-tests/fixtures/expected.js
runspired: fixing the inclusion of the `debug` package, there is a follow up PR to this on that will make our tests and fixtures show the module, class, package and public/private status of each piece of API documentation so that "why" something changes is more clear.

packages/-ember-data/package.json
runspired: a mistake, good catch

packages/-ember-data/package.json
```diff
@@ -11,14 +11,14 @@
     "build": "ember build",
     "build:production": "ember build --environment=production",
     "start": "ember server",
-    "docs": "ember ember-cli-yuidoc",
+    "docs": "mkdir -p dist && ember ember-cli-yuidoc",
```
igorT: Does this work correctly if you already have stuff in dist?

runspired: yes

igorT: I meant more from there being existing code inside the folder, vs just the folder itself

packages/-ember-data/node-tests/fixtures/expected.js
igorT: Can we add that to the PR description please?

runspired: added

packages/-ember-data/yuidoc.json
```diff
@@ -3,7 +3,7 @@
   "description": "The ember-data API: a data persistence library for Ember.js",
   "url": "https://github.com/emberjs/data",
   "options": {
-    "enabledEnvironments": ["production"],
+    "enabledEnvironments": [],
```
igorT: What do `enabledEnvironments` do?

packages/-ember-data/yuidoc.json
```diff
@@ -12,9 +12,10 @@
       "../model/addon",
       "../serializer/addon",
       "../store/addon",
-      "../store/addon/-private/system"
+      "../record-data/addon",
```
igorT: Did we remove system on purpose?

packages/-ember-data/yuidoc.json
runspired: if you add an environment the docs will build when a build or serve for that environment occurs. This was the root of the reason why we were doing production builds during prepublish before, now we just call the docs script directly.

packages/debug/addon/index.js
```diff
@@ -68,8 +69,9 @@ export default DataAdapter.extend({
   /**
     Fetch the model types and observe them for changes.
     Maintains the list of model types without needing the Model package for detection.
-    @public
+
     @method watchModelTypes
+    @internal
```
igorT: This is public

packages/-ember-data/yuidoc.json
runspired: yes, it was redundant

packages/debug/addon/index.js
runspired: It is public for Ember's not ours.

packages/-ember-data/yuidoc.json
igorT: :+1:

packages/-ember-data/node-tests/fixtures/expected.js
igorT: Thanks!

PR #6736, opened 2019-11-14
[CHORE] delete unused infra

runspired:
we had two mechanisms for feature flagging, this eliminates the older mechanism.


packages/private-build-infra/src/stripped-build-plugins.js
```diff
@@ -23,78 +19,22 @@ function isProduction(environment) {
...
     postTransformPlugins.push([StripClassCallCheck]);
     uniqueAdd(filteredImports, '@ember-data/store/-debug', ['assertPolymorphicType']);
   }
 
   plugins.push(
     [FilterImports, { imports: filteredImports }],
     [StripFilteredImports, filteredImports],
-    [TransformBlockScoping, { throwIfClosureRequired: true }]
+    [TransformBlockScoping, { throwIfClosureRequired: true }],
+    ...DebugMacros
```
igorT: This got added in this PR?

runspired: No, just moved to the appropriate spot. When the new infra got added before we added it somewhere else instead of here.

PR #6737, opened 2019-11-14
[CHORE] add package flag infra, strip flag infra

### LINKED ISSUE #6709: Move @ember-data/ package detection to build time
Currently for deprecations and encapsulation we rely on the detection of the presence of `@ember-data/<packageName>` via a check on `require.has` at runtime. We should switch these flags to being imports that are injected at build-time and eliminated via our conventional flag stripping infra. Ember compatibility helpers is a great example library for how to achieve this.

runspired:
resolves #6709

With this we use build time flags instead of runtime heuristics for wrapping conditional code for optional-package dependencies. This reduces the cost of supporting those packages.

We also now drop the feature-flag modules from production builds entirely, which saves us some additional bytes :)

Additionally this fixes stripping dead import statements, which apparently was silently broken due to how more recent versions of plugins in our build interop. Asset size monitoring should help to watch for that in the future.

Our asset-size checks currently do not report the size of packages like `store` when `model` `adapter` and `serializer` are removed, but this likely ought to have an additional effect between .5-1 KB currently in addition to the wins from import-stripping and module-dropping that this PR brings.


igorT:
Seems like this needs TS review, added a label

packages/private-build-infra/src/utilities/unique-add.js
```diff
@@ -0,0 +1,9 @@
+module.exports = function uniqueAdd(obj, key, values) {
```
rwjblue: This should die 😄 (we can use `Set`'s)

packages/store/addon/-private/system/core-store.ts
```diff
@@ -13,7 +13,7 @@ import { default as RSVP, all, resolve, Promise, defer } from 'rsvp';
 import Service from '@ember/service';
 import { typeOf, isPresent, isNone } from '@ember/utils';
 
-import require, { has } from 'require';
+import require from 'require';
```
rwjblue: Just curious, what are we using `require` for here now?

packages/private-build-infra/addon/index.ts
```diff
@@ -0,0 +1,22 @@
+/**
+  @module @ember-data/private-build-infra
+*/
+
+import { has } from 'require';
+import POSSIBLE_PACKAGES from './available-packages';
+
+type PackageSettings = Record<keyof typeof POSSIBLE_PACKAGES, boolean>;
+
+const PACKAGES = Object.keys(POSSIBLE_PACKAGES).reduce((obj, name) => {
```
rwjblue: `PACKAGES` isn't exported here, I'd prefer to make a simpler method and invoke it just below

Maybe something like this:

https://github.com/emberjs/ember.js/blob/892ca20980c67aaef247bdd5c69263f81522d7e1/packages/%40ember/canary-features/index.ts#L63-L77

packages/store/addon/-private/system/core-store.ts
runspired: if we do have the model package we have to import the model class for `instanceof` checks

packages/private-build-infra/src/utilities/unique-add.js
runspired: that might be messier given the format that the babel plugin expects is an array

packages/private-build-infra/addon/index.ts
runspired: cleaned up

packages/canary-features/index.js
```diff
@@ -1 +1,31 @@
-module.exports = { name: require('./package').name };
+function getApp(addon) {
```
rwjblue: I’d rather use the prexisting this._findHost

packages/canary-features/index.js
```diff
@@ -1 +1,31 @@
...
+    // we are not enabled.
+    if (this.isEnabled()) {
+      return this._super.treeFor.call(this, ...arguments);
+    }
+  },
+  isEnabled() {
+    if (this.__isEnabled !== undefined) {
+      return this.__isEnabled;
+    }
+    const env = getApp(this).env;
```
rwjblue: If we only need env, let’s use process.env.EMBER_ENV instead

packages/private-build-infra/index.js
```diff
@@ -0,0 +1,31 @@
+function getApp(addon) {
```
rwjblue: Can we use _findHost instead?

packages/private-build-infra/index.js
```diff
@@ -0,0 +1,31 @@
...
+    // we are not enabled.
+    if (this.isEnabled()) {
+      return this._super.treeFor.call(this, ...arguments);
+    }
+  },
+  isEnabled() {
+    if (this.__isEnabled !== undefined) {
+      return this.__isEnabled;
+    }
+    const env = getApp(this).env;
```
rwjblue: Same RE: process.env.EMBER_ENV

packages/private-build-infra/src/addon-build-config-for-data-package.js
```diff
@@ -11,6 +11,16 @@ function isProductionEnv() {
   return isProd && !isTest;
 }
 
+function getApp(addon) {
```
rwjblue: Can we use this_findHost?

packages/private-build-infra/src/stripped-build-plugins.js
```diff
@@ -2,37 +2,27 @@
...
-function uniqueAdd(obj, key, values) {
-  const a = (obj[key] = obj[key] || []);
-
-  for (let i = 0; i < values.length; i++) {
-    if (a.indexOf(values[i]) === -1) {
-      a.push(values[i]);
-    }
-  }
-}
+const uniqueAdd = require('./utilities/unique-add');
```
rwjblue: :P

runspired: removed

PR #6740, opened 2019-11-14
[CHORE] ISSUE-6719 - update prettier

### LINKED ISSUE #6719: Prettier bump requires running fix
#6716 fails because prettier's output changes. I suspect just upgrading the dependency locally and running `yarn lint:js --fix` will resolve this.

Mikek2252:
Updated prettier and ran `yarn lint:js --fix` should resolve #6719


runspired:
Failures are due to a temporary npm outage. Will kickstart.

PR #6746, opened 2019-11-15
[BUGFIX] ensure MandatorySetter does not trigger for modelName

### LINKED ISSUE #6143: Ember octane: tracked issue
There is issue with modelName, when I try to create polymorphic relationship
```
export class SeoRecordableModelSchema extends Model {
  @hasMany('base', { async: false, polymorphic: true })
  contents;
}
```
in console I have this message:
```
You attempted to update model:base.modelName to "text", but it is being tracked by a tracking context, such as a template, computed property, or observer. In order to make sure the context updates properly, you must invalidate the property when updating it. You can mark the property as `@tracked`, or use `@ember/object#set` to do this.
```
resolved  when add `@tracked` prop to the Base model
```
export default class Base extends Model {
  @tracked
  static modelName
}
```

runspired:
resolves #6143 

Users who extend a model from another model (not constrained to polymorphism) will encounter a potential race condition when upgrading to ember-source 3.13 or later.

- If the model being extended is looked up prior to the model doing the extending
- and if the model being extended has it's modelName accessed in a manner that would install a MandatorySetter prior to the lookup of the model doing the extending

Then setting the static modelName property of the extended class would error.

## Backporting Notes

We should attempt to backport the fix as far as possible; however, due to file re-arrangements and refactoring while the fix will remain the same the location is likely to change for 3.12 and again for 3.8.

### DIFF
--- a/packages/-ember-data/tests/integration/store/model-name-test.js
+++ b/packages/-ember-data/tests/integration/store/model-name-test.js
@@ -1,0 +1,83 @@
+import { module, test } from 'qunit';
+import { setupTest } from 'ember-qunit';
+import Model, { attr } from '@ember-data/model';
+import { computed } from '@ember/object';
+import Service, { inject } from '@ember/service';
+import Store from '@ember-data/store';
+import { P } from 'Object/_api';
+
+module('@ember-data/model klass.modelName', function(hooks) {
+  setupTest(hooks);
+
+  test('Extending a model properly sets the modelName', function(assert) {
+    const { owner } = this;
+    class Animal extends Model {
+      @attr() species;
+    }
+    class Pet extends Animal {
+      @attr() name;
+    }
+    class AnimalHelper extends Service {
+      @inject store;
+
+      @computed('animal.constructor.modelName')
+      get animalModelName() {
+        return this.animal.constructor.modelName;
+      }
+    }
+    owner.register('model:animal', Animal);
+    owner.register('model:pet', Pet);
+    owner.register('service:animal-helper', AnimalHelper);
+    owner.register('sercice:store', Store);
+    const store = owner.lookup('service:store');
+    const animalHelper = owner.lookup('service:animal-helper');
+
+    // first we push the base class
+    const snake = store.push({
+      data: {
+        type: 'animal',
+        id: '1',
+        attributes: { species: 'snake' },
+      },
+    });
+
+    // then we do a thing that could install a MandatorySetter
+    // on the modelName property on the `Animal` class.
+    animalHelper.set('animal', snake);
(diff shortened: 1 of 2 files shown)


packages/store/addon/-private/system/schema-definition-service.ts
```diff
@@ -103,7 +104,7 @@ export function getModelFactory(store: CoreStore, cache, normalizedModelName: st
     if (klass.isModel) {
       let hasOwnModelNameSet = klass.modelName && Object.prototype.hasOwnProperty.call(klass, 'modelName');
       if (!hasOwnModelNameSet) {
-        klass.modelName = normalizedModelName;
+        Object.defineProperty(klass, 'modelName', { value: normalizedModelName });
```
pzuraq: You can probably do this in DEBUG only, if it matters

pzuraq: I suppose it'll happen in DEBUG only anyways

runspired: I considered that but I feel the little bit of extra immutability safety here is nice.

packages/-ember-data/tests/integration/store/model-name-test.js
```diff
@@ -0,0 +1,82 @@
...
+    // then we do a thing that could install a MandatorySetter
+    // on the modelName property on the `Animal` class.
+    animalHelper.set('animal', snake);
+    assert.strictEqual(snake.constructor.modelName, 'animal', 'Snake has the right modelName');
+    assert.strictEqual(animalHelper.animalModelName, 'animal', 'We got the modelName');
+
+    // ensure modelName is immutable
+    try {
+      snake.constructor.modelName = 'bear';
+      assert.ok(false, 'expected modelName to be immutable');
```
Turbo87: does assert.throws() not work here?

packages/-ember-data/tests/integration/store/model-name-test.js
```diff
@@ -0,0 +1,82 @@
...
+          type: 'pet',
+          id: '1',
+          attribute: { species: 'dog', name: 'fido' },
+        },
+      });
+
+      assert.strictEqual(fido.constructor.modelName, 'pet', 'Fido has the right modelName');
+      assert.strictEqual(snake.constructor.modelName, 'animal', 'Snake has the right modelName');
+      assert.strictEqual(animalHelper.animalModelName, 'animal', 'AnimalHelper has the right modelName');
+    } catch (e) {
```
Turbo87: why the try/catch instead of letting it fail?

PR #6753, opened 2019-11-16
[CHORE] deactivate all but allowed-fail travis jobs

runspired:


igorT:
I'd leave the other partners tests in, so we can turn them on & off if need be in the future

PR #6754, opened 2019-11-17
[TEST] queries tests for adapter encapsulation side quest

Gaurav0:
For #6666 

This also exposes, but does not fix, a bug where the payload object is unexpectedly mutated if IDENTIFIERS is turned on. See issue #6758 

This also goes back and cleans up the transform code needed previously to avoid a deprecation warning.

### DIFF
--- a/packages/unpublished-adapter-encapsulation-test-app/tests/integration/queries-test.js
+++ b/packages/unpublished-adapter-encapsulation-test-app/tests/integration/queries-test.js
@@ -1,0 +1,225 @@
+import { setupTest } from 'ember-qunit';
+import { module, test } from 'qunit';
+import EmberObject from '@ember/object';
+import Store from 'adapter-encapsulation-test-app/services/store';
+import Model, { attr } from '@ember-data/model';
+import { resolve } from 'rsvp';
+
+class MinimalSerializer extends EmberObject {
+  normalizeResponse(_, __, data) {
+    return data;
+  }
+
+  serialize(snapshot) {
+    return {
+      data: {
+        id: snapshot.id,
+        type: snapshot.modelName,
+        attributes: snapshot.attributes(),
+      },
+    };
+  }
+}
+
+class Person extends Model {
+  @attr
+  firstName;
+
+  @attr
+  lastName;
+}
+
+module('integration/queries - Queries Tests', function(hooks) {
+  setupTest(hooks);
+
+  hooks.beforeEach(function() {
+    this.owner.register('service:store', Store);
+    this.owner.register('serializer:application', MinimalSerializer);
+    this.owner.register('model:person', Person);
+  });
+
+  test('store.findRecord calls adapter.findRecord w/correct args', async function(assert) {
+    let findRecordCalled = 0;
+    let expectedResult = {
+      data: {
+        id: '12',
+        type: 'person',
+        attributes: {
+          firstName: 'Gaurav',
+          lastName: 'Munjal',
+        },
+      },
+    };
+    let { owner } = this;
+    let store = owner.lookup('service:store');
+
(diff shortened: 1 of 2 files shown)


packages/store/addon/-private/system/core-store.ts
```diff
@@ -1142,7 +1142,8 @@ abstract class CoreStore extends Service {
         if (IDENTIFIERS) {
           // ensure that regardless of id returned we assign to the correct record
           if (payload.data && !Array.isArray(payload.data)) {
-            payload.data.lid = identifier.lid;
+            payload = { ...payload };
+            payload.data = { ...payload.data, lid: identifier.lid };
```
runspired: not sure how I feel about this. While it does assign to the passed in payload (a mutation) this is a hot path where making an immutable copy is likely worse.

Gaurav0: Discussed on Discord. We will open an issue and remove the bugfix from this PR.

packages/unpublished-adapter-encapsulation-test-app/tests/integration/queries-test.js
```diff
@@ -0,0 +1,222 @@
...
+
+    let record = await store.findRecord('person', '12');
+
+    // This code is a workaround for issue https://github.com/emberjs/data/issues/6758
+    // expectedResult is unexpectedly mutated during store.findRecord
+    // if IDENTIFIERS is turned on
+    delete expectedResult.data.lid;
+
+    assert.equal(findRecordCalled, 1, 'findRecord is called once');
+    assert.deepEqual(record.serialize(), expectedResult, 'findRecord returns expected result');
```
runspired: If we don't re-use expectedResult we won't hit the `lid` issue.

While it's good we'e surfaced the mutation issue for discussion, you never want the `expected` to be something you've potentially operated on.

Let's refactor this to either inline the expected result or use the deep-copy util to preserve a copy when we define it. (same for the similar things below)

packages/unpublished-adapter-encapsulation-test-app/tests/integration/queries-test.js
```diff
@@ -0,0 +1,222 @@
...
+    let { owner } = this;
+    let store = owner.lookup('service:store');
+
+    class TestQueryAdapter extends EmberObject {
+      query(passedStore, type, query) {
+        queryCalled++;
+
+        assert.equal(passedStore, store, 'instance of store is passed to query');
+        assert.equal(type, Person, 'model is passed to query');
+        assert.deepEqual(query, { firstName: 'Chris' }, 'query is passed to query');
```
runspired: While not implemented by the default adapters, the full signature of query is 5 args

```ts
query(Store: Store, schema: ModelClass, query, recordArray, options): Promise<unknown>;
```

packages/unpublished-adapter-encapsulation-test-app/tests/integration/queries-test.js
```diff
@@ -0,0 +1,222 @@
...
+    let { owner } = this;
+    let store = owner.lookup('service:store');
+
+    class TestQueryRecordAdapter extends EmberObject {
+      queryRecord(passedStore, type, query) {
+        queryRecordCalled++;
+
+        assert.equal(passedStore, store, 'instance of store is passed to queryRecord');
+        assert.equal(type, Person, 'model is passed to queryRecord');
+        assert.deepEqual(query, { firstName: 'Gaurav' }, 'query is passed to queryRecord');
```
runspired: While not implemented by the default adapters, the full signature for `queryRecord` is 4 args

```ts
  queryRecord(store: Store, schema: ModelClass, query, options): Promise<unknown>;
```

PR #6755, opened 2019-11-17
[TEST] mutations tests for adapter encapsulation side quest

Gaurav0:
For #6666

### DIFF
--- a/packages/unpublished-adapter-encapsulation-test-app/tests/integration/mutations-test.js
+++ b/packages/unpublished-adapter-encapsulation-test-app/tests/integration/mutations-test.js
@@ -1,0 +1,333 @@
+import { setupTest } from 'ember-qunit';
+import { module, test } from 'qunit';
+import EmberObject from '@ember/object';
+import Store from 'adapter-encapsulation-test-app/services/store';
+import Model, { attr } from '@ember-data/model';
+import { resolve } from 'rsvp';
+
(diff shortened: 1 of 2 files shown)


packages/unpublished-adapter-encapsulation-test-app/tests/integration/mutations-test.js
```diff
@@ -0,0 +1,333 @@
...
+
+        return resolve(expectedData);
+      }
+    }
+
+    this.owner.register('adapter:application', TestUpdateRecordAdapter);
+
+    let record = store.push(expectedData);
+
+    assert.equal(createRecordCalled, 0, 'createRecord is not called');
```
runspired: this should be after the save call

packages/unpublished-adapter-encapsulation-test-app/tests/integration/mutations-test.js
```diff
@@ -0,0 +1,333 @@
...
+    let record = store.createRecord('person', props);
+    await record.save();
+
+    assert.equal(createRecordCalled, 1, 'createRecord is called once');
+
+    record.firstName = 'Kevin';
+    expectedData.data.attributes.firstName = 'Kevin';
+    await record.save();
+
+    assert.equal(updateRecord, 1, 'updateRecord is called once');
```
runspired: let's also assert that createRecord is still only `1` here

PR #6763, opened 2019-11-19
[CHORE] rename Record to RecordInstance

runspired:
Replaces #6350

### DIFF
--- a/packages/store/addon/-private/ts-interfaces/ds-model.ts
+++ b/packages/store/addon/-private/ts-interfaces/ds-model.ts
@@ -1,11 +1,10 @@
-import { Record } from './record';
 import RSVP from 'rsvp';
 import EmberObject from '@ember/object';
 import { JsonApiValidationError } from './record-data-json-api';
 import { RelationshipSchema } from './record-data-schemas';
 
 // Placeholder until model.js is typed
-export interface DSModel extends Record, EmberObject {
+export interface DSModel extends EmberObject {
   toString(): string;
   save(): RSVP.Promise<DSModel>;
   eachRelationship(callback: (key: string, meta: RelationshipSchema) => void): void;
--- a/packages/store/addon/-private/ts-interfaces/ds-model.ts
+++ b/packages/store/addon/-private/ts-interfaces/ds-model.ts
@@ -2,9 +2,10 @@ import RSVP from 'rsvp';
 import EmberObject from '@ember/object';
 import { JsonApiValidationError } from './record-data-json-api';
 import { RelationshipSchema } from './record-data-schemas';
+import { RecordInstance } from './record-instance';
 
 // Placeholder until model.js is typed
-export interface DSModel extends EmberObject {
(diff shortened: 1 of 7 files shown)


packages/store/addon/-private/ts-interfaces/ds-model.ts
```diff
@@ -1,11 +1,10 @@
-import { Record } from './record';
 import RSVP from 'rsvp';
 import EmberObject from '@ember/object';
 import { JsonApiValidationError } from './record-data-json-api';
 import { RelationshipSchema } from './record-data-schemas';
 
 // Placeholder until model.js is typed
-export interface DSModel extends Record, EmberObject {
+export interface DSModel extends EmberObject {
```
igorT: Missing RecordInstance here

packages/store/addon/-private/ts-interfaces/record.ts
```diff
@@ -10,8 +10,7 @@
   model instances (aka `Records`) are completely user supplied and opaque to the internals, we need a type
   through which to communicate what is valid.
 
-  The type belows allows for either a class instance, or an object, but not primitive values or functions.
+  The type belows allows for anything extending object.
 */
 
-// TODO Rename to RecordInstance
-export type Record = Object;
+export type RecordInstance = object;
```
igorT: What is the difference between lowercase and capitalized object?

packages/store/addon/-private/ts-interfaces/record.ts
```diff
@@ -10,8 +10,7 @@
   model instances (aka `Records`) are completely user supplied and opaque to the internals, we need a type
   through which to communicate what is valid.
 
-  The type belows allows for either a class instance, or an object, but not primitive values or functions.
+  The type belows allows for anything extending object.
```
igorT: We should rename this file now that the type has been renamed

packages/store/addon/-private/ts-interfaces/record.ts
igorT: Should we make this the default export?

igorT:
You mentioned this replaces #6350, are there substantial changes, or is it just a rebase?

igorT:
cc @rwjblue

packages/store/addon/-private/ts-interfaces/record.ts
runspired: ack, thanks, that was an accidental change

packages/store/addon/-private/ts-interfaces/ds-model.ts
runspired: it's intentional, it should not be extending `RecordInstance`

packages/store/addon/-private/ts-interfaces/record.ts
runspired: I'd rather not as there seems to be an annoying thing where default exports warn during transpile when they were just a type

runspired:
@igorT it's a re-create since files have been re-arranged since

PR #6764, opened 2019-11-19
[CHORE] introduce static ModelSchema interface

runspired:
cc @chriskrycho this is part of the next steps we previously discussed due to issues with being able to type DS.Model

### DIFF
--- a/packages/store/addon/-private/ts-interfaces/ds-model.ts
+++ b/packages/store/addon/-private/ts-interfaces/ds-model.ts
@@ -1,8 +1,8 @@
 import RSVP from 'rsvp';
 import EmberObject from '@ember/object';
 import { JsonApiValidationError } from './record-data-json-api';
-import { RelationshipSchema } from './record-data-schemas';
 import { RecordInstance } from './record-instance';
+import { RelationshipSchema, AttributeSchema } from './record-data-schemas';
 
 // Placeholder until model.js is typed
 export interface DSModel extends RecordInstance, EmberObject {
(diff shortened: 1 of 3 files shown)


packages/store/addon/-private/ts-interfaces/ds-model.ts
```diff
@@ -13,3 +13,18 @@ export interface DSModel extends Record, EmberObject {
   invalidErrorsChanged(errors: JsonApiValidationError[]): void;
   [key: string]: unknown;
 }
+
+// When model.js is typed this should be the static methods and properties
```
igorT: :+1: for the comment

packages/store/addon/-private/ts-interfaces/minimum-serializer-interface.ts
```diff
@@ -139,7 +139,7 @@ interface Serializer {
    * @method normalizeResponse
    * @public
    * @param {Store} store The store service that initiated the request being normalized
-   * @param {ShimModelClass} schema An object with methods for accessing information about
+   * @param {ModelSchema} schema An object with methods for accessing information about
```
igorT: If we don't publish types, how does this work out for the docs?

igorT:
> cc @chriskrycho this is part of the next steps we previously discussed due to issues with being able to type DS.Model

Can you expand on the issues you are referring to please?

runspired:
@igorT we can't convert `model.js` to typescript for at least another full LTS cycle due to object model compat issues.

packages/store/addon/-private/ts-interfaces/minimum-serializer-interface.ts
runspired: the docs don't link these things

pzuraq:
I believe the only native things that aren’t supported in 3.8 are decorators, and if you install the decorators polyfill those _should_ work as well I believe. The polyfill can be included as a dep like a standard ember-addon, and will be included in the parent app. Unsure if that would work for your use case.

runspired:
@pzuraq does that mean the class fix in ember-source was backported?

pzuraq:
Ah, I forgot about the transpilation bug. Yeah, that fix likely won’t make it back unfortunately.

runspired:
After further discussion it turns out that the transpilation fix was not backported to 3.12 either, so currently 3.21 is the soonest we can refactor Model to a class unless we tell folks to use something like https://github.com/pzuraq/ember-class-constructor-fix when it comes available or install it ourselves for them.

igorT:
@runspired, to me converting to a class seems orthogonal to converting to
typescript, no?

PR #6767, opened 2019-11-19
[BUGFIX] fix detect usage for native class polymorphism

### LINKED ISSUE #6728: Polymorphic relationships and native classes
### Pull request with broken test

#6727

### Description

I use mixins to create polymorphic associations, but I've faced the problem with using it for native classes:
```
export default class MyModel extends DS.Model.extend(Polymorphable) {
 ....
}

// other model
@hasMany('polymorphable', { polymorphic: true }) polymorphable; 
```

When I try to assign new value for it I get that MyModel is not polymorphable. When I don't use native class for MyModel it works as expected.
Is it a bug? What's the proper way of using polymorphic associations with native classes without using inheritance?
### Versions

Run the following command and paste the output below:

```cli
---@0.0.0 /home/yuri/projects/---
└── ember-source@3.13.3 

---@0.0.0 /home/yuri/projects/---
└── ember-cli@3.13.1 

---@0.0.0 /home/yuri/projects/---
└── ember-data@3.13.1 

```

runspired:
resolves #6728 
Ports the test from #6727 and adds a fix

### DIFF
--- a/packages/-ember-data/tests/integration/relationships/polymorphic-mixins-has-many-test.js
+++ b/packages/-ember-data/tests/integration/relationships/polymorphic-mixins-has-many-test.js
@@ -14,13 +14,15 @@ module(
   function(hooks) {
     setupTest(hooks);
 
+    let Message;
+
     hooks.beforeEach(function() {
       const User = Model.extend({
         name: attr('string'),
         messages: hasMany('message', { async: true, polymorphic: true }),
       });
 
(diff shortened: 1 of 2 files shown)


runspired:
@pzuraq it does. If we were failing to detect mixins from things being extended before then this is an even better bugfix :) Hopefully we can get away from class based polymorphism soon :D

PR #6775, opened 2019-11-20
Fix setting ID on Record Data from DS.Model

igorT:


packages/store/addon/-private/system/model/internal-model.ts
```diff
@@ -1359,6 +1359,10 @@ export default class InternalModel {
 
     if (didChange && id !== null) {
       this.store.setRecordId(this.modelName, id, this.clientId);
+      // internal set of ID to get it to RecordData from DS.Model
+      if (this._recordData.__setId) {
+        this._recordData.__setId(id);
```
mike-north: may want to leave yourself a note to use optional chaining here when TS 3.7 is safe to adopt

packages/-ember-data/tests/unit/model-test.js
```diff
@@ -372,6 +372,13 @@ module('unit/model - Model', function(hooks) {
       assert.equal(idChange, 0);
       person._internalModel.setId('john');
       assert.equal(idChange, 1);
+      let recordData = recordDataFor(person);
+      assert.equal(
+        recordData.getResourceIdentifier().id,
```
runspired: nice catch

packages/store/addon/-private/system/model/internal-model.ts
igorT: Thanks, want to leave the check more loud for the reader/explicit in this case

PR #6792, opened 2019-11-24
[DOC] improve urlForFindAll example

mcfiredrill:
While the example provided is valid, I think a slightly more fleshed out
example of usage might be more helpful for newcomers.

### DIFF
--- a/packages/adapter/addon/-private/build-url-mixin.js
+++ b/packages/adapter/addon/-private/build-url-mixin.js
@@ -153,7 +153,8 @@ export default Mixin.create({
 
    export default JSONAPIAdapter.extend({
      urlForFindAll(modelName, snapshot) {
-       return 'data/comments.json';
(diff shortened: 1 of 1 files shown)


efx:
@mcfiredrill thanks for contributing! 
longer-term we are trying to clean up all of these embedded examples to make ember-data's API documentation coherent and polished.  (tl;dr here: https://www.notion.so/QUEST-issue-documentation-draft-94444df09d804ae7bce9665401f462be). We'll be putting together the actual QUEST issue soon.

mcfiredrill:
@efx Thanks for the link! I'd love to help out with that effort if I can.

efx:
@mcfiredrill awesome, that would be greatly appreciated. I will cc you once I create the QUEST 😄

Gaurav0:
LGTM

PR #6797, opened 2019-11-25
[BUGFIX TEST] relationships tests for adapter enapsulation side quest

Gaurav0:
For #6666

### DIFF
--- a/packages/unpublished-adapter-encapsulation-test-app/tests/integration/belongs-to-test.js
+++ b/packages/unpublished-adapter-encapsulation-test-app/tests/integration/belongs-to-test.js
@@ -105,7 +105,7 @@ module('integration/belongs-to - Belongs To Tests', function(hooks) {
         id: '3',
         type: 'comment',
         attributes: {
-          text: 'You suck',
+          text: 'You rock',
         },
         relationships: {
           post: {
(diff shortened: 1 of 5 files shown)


packages/unpublished-adapter-encapsulation-test-app/tests/integration/belongs-to-test.js
```diff
@@ -0,0 +1,526 @@
...
+  test('if a belongsTo relationship has a link but no data (findBelongsTo is defined)', async function(assert) {
+    let findRecordCalled = 0;
+    let findBelongsToCalled = 0;
+
+    let initialRecord = {
+      data: {
+        id: '3',
+        type: 'comment',
+        attributes: {
+          text: 'You suck',
```
runspired:
```suggestion
          text: 'You rock',
```

runspired: Even though this switches out after let's be more positive up front (change the others below too) ;)

packages/unpublished-adapter-encapsulation-test-app/tests/integration/has-many-test.js
```diff
@@ -0,0 +1,777 @@
...
+  post;
+}
+
+let expectedResult = {
+  data: [
+    {
+      id: '3',
+      type: 'comment',
+      attributes: {
+        text: 'You suck',
```
runspired: let's keep it positive in this file too

PR #6800, opened 2019-11-25
[CHORE] inject the Model specific _modelForMixin implementation into the store from the Model package

Gaurav0:
For #6166

### DIFF
--- a/packages/store/addon/-private/system/schema-definition-service.ts
+++ b/packages/store/addon/-private/system/schema-definition-service.ts
@@ -5,6 +5,7 @@ import { getOwner } from '@ember/application';
 import normalizeModelName from './normalize-model-name';
 import { RelationshipsSchema, AttributesSchema } from '../ts-interfaces/record-data-schemas';
 import CoreStore from './core-store';
+import { RecordInstance } from '../ts-interfaces/record-instance';
 
 export class DSModelSchemaDefinitionService {
   private _modelFactoryCache = Object.create(null);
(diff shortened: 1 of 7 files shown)


packages/store/addon/-private/system/schema-definition-service.ts
```diff
@@ -4,19 +4,8 @@ import { get } from '@ember/object';
...
-type Model = import('@ember-data/model').default;
-
-let _Model;
-function getModel() {
-  if (HAS_MODEL_PACKAGE) {
-    _Model = _Model || require('@ember-data/model').default;
-  }
-  return _Model;
-}
+import { RecordInstance } from '../ts-interfaces/record-instance';
```
runspired: `RecordInstance` is for the instantiated record instances, I believe based on usage below you are in need of `ModelSchema` or `DSModelSchema`

packages/store/addon/-private/system/schema-definition-service.ts
```diff
@@ -83,15 +72,15 @@ export class DSModelSchemaDefinitionService {
...
+export function getModelFactory(store: CoreStore, cache, normalizedModelName: string): RecordInstance | null {
   let factory = cache[normalizedModelName];
 
   if (!factory) {
     factory = _lookupModelFactory(store, normalizedModelName);
 
     if (!factory) {
       //Support looking up mixins as base types for polymorphic relationships
-      factory = _modelForMixin(store, normalizedModelName);
+      factory = store._modelForMixin(normalizedModelName);
```
runspired: instead of adding a private method to store and injecting it from the model package, we should use the `HAS_MODEL_PACKAGE`  build flag to conditionally import a private function from that package.

```ts
let _modelForMixin;
if (HAS_MODEL_PACKAGE) {
 _modelForMixin = require('@ember-data/model/-private')._modelForMixin;
}

// ....

if (HAS_MODEL_PACKAGE) {
  factory = _modelForMixin(store, normalizedModelName);
}
```

possibly with a module scope cache for the method to avoid looking it up each time.

runspired:
Closing in favor of #6802

PR #6801, opened 2019-11-26
[CHORE ts] type Snapshot

runspired:
Will be used by the minimum-adapter-interface so typing it makes that type more robust

### DIFF
--- a/packages/store/addon/-private/system/core-store.ts
+++ b/packages/store/addon/-private/system/core-store.ts
@@ -1152,7 +1152,6 @@ abstract class CoreStore extends Service {
     internalModel.loadingData();
     let identifier = internalModel.identifier;
 
-    // TODO when we upgrade to TS 3.7 we can make use of assert
     assertIdentifierIsExisting(identifier);
 
     let promise = this._fetchManager.scheduleFetch(identifier, options, generateStackTrace);
--- a/packages/store/addon/-private/system/core-store.ts
+++ b/packages/store/addon/-private/system/core-store.ts
@@ -61,7 +61,7 @@ import { ResourceIdentifierObject } from '../ts-interfaces/ember-data-json-api';
 import promiseRecord from '../utils/promise-record';
 import { identifierCacheFor, IdentifierCache } from '../identifiers/cache';
 import { internalModelFactoryFor, setRecordIdentifier, recordIdentifierFor } from './store/internal-model-factory';
-import { RecordIdentifier, StableRecordIdentifier } from '../ts-interfaces/identifier';
+import { RecordIdentifier, StableRecordIdentifier, StableExistingRecordIdentifier } from '../ts-interfaces/identifier';
 import { RecordReference, HasManyReference, BelongsToReference } from './references';
 import { Backburner } from '@ember/runloop/-private/backburner';
 import Snapshot from './snapshot';
@@ -1148,7 +1148,13 @@ abstract class CoreStore extends Service {
     // TODO  remove this once we dont rely on state machine
     internalModel.loadingData();
     let identifier = internalModel.identifier;
-    let promise = this._fetchManager.scheduleFetch(internalModel.identifier, options, generateStackTrace);
+
+    // TODO when we upgrade to TS 3.7 we can make use of assert
+    if (!identifierIsExisting(identifier)) {
+      throw new Error(`Attempted to schedule a fetch for a record without an id.`);
+    }
+
+    let promise = this._fetchManager.scheduleFetch(identifier, options, generateStackTrace);
     return promise.then(
       payload => {
         if (IDENTIFIERS) {
@@ -2481,7 +2487,9 @@ abstract class CoreStore extends Service {
       let pendingItem = pending[i];
       let snapshot = pendingItem.snapshot;
       let resolver = pendingItem.resolver;
-      let internalModel = snapshot._internalModel;
+      // TODO We have to cast due to our reliance on this private property
+      // this will be refactored away once we change our pending API to be identifier based
+      let internalModel = ((snapshot as unknown) as { _internalModel: InternalModel })._internalModel;
       let adapter = this.adapterFor(internalModel.modelName);
       let operation;
 
@@ -3034,7 +3042,7 @@ abstract class CoreStore extends Service {
     return internalModel.reloadBelongsTo(key, options);
   }
 
-  _internalModelForResource(resource: RecordIdentifier): InternalModel {
+  _internalModelForResource(resource: ResourceIdentifierObject): InternalModel {
     return internalModelFactoryFor(this).getByResource(resource);
   }
 
@@ -3724,4 +3732,8 @@ function internalModelForRelatedResource(
 ): InternalModel {
   const identifier = cache.getOrCreateRecordIdentifier(resource);
   return store._internalModelForResource(identifier);
+}
+
+function identifierIsExisting(identifier: StableRecordIdentifier): identifier is StableExistingRecordIdentifier {
+  return identifier.id !== null;
 }
--- a/packages/store/addon/-private/system/core-store.ts
+++ b/packages/store/addon/-private/system/core-store.ts
@@ -82,7 +82,6 @@ type SingleResourceDocument = import('../ts-interfaces/ember-data-json-api').Sin
 type CollectionResourceDocument = import('../ts-interfaces/ember-data-json-api').CollectionResourceDocument;
 type JsonApiDocument = import('../ts-interfaces/ember-data-json-api').JsonApiDocument;
 type ExistingResourceObject = import('../ts-interfaces/ember-data-json-api').ExistingResourceObject;
-
 type RecordIdentifier = import('../ts-interfaces/identifier').RecordIdentifier;
 type StableRecordIdentifier = import('../ts-interfaces/identifier').StableRecordIdentifier;
 type StableExistingRecordIdentifier = import('../ts-interfaces/identifier').StableExistingRecordIdentifier;
@@ -94,7 +93,7 @@ type Dict<T> = import('../ts-interfaces/utils').Dict<T>;
 type RecordDataRecordWrapper = import('../ts-interfaces/record-data-record-wrapper').RecordDataRecordWrapper;
 type AttributesSchema = import('../ts-interfaces/record-data-schemas').AttributesSchema;
 type SchemaDefinitionService = import('../ts-interfaces/schema-definition-service').SchemaDefinitionService;
-
+type PrivateSnapshot = import('./snapshot').PrivateSnapshot;
 type Relationship = import('@ember-data/record-data/-private').Relationship;
 
 const emberRun = emberRunLoop.backburner;
@@ -2489,7 +2488,7 @@ abstract class CoreStore extends Service {
       let resolver = pendingItem.resolver;
       // TODO We have to cast due to our reliance on this private property
       // this will be refactored away once we change our pending API to be identifier based
-      let internalModel = ((snapshot as unknown) as { _internalModel: InternalModel })._internalModel;
+      let internalModel = ((snapshot as unknown) as PrivateSnapshot)._internalModel;
       let adapter = this.adapterFor(internalModel.modelName);
       let operation;
 
--- a/packages/store/addon/-private/system/core-store.ts
+++ b/packages/store/addon/-private/system/core-store.ts
@@ -13,7 +13,6 @@ import { default as RSVP, all, resolve, Promise, defer } from 'rsvp';
 import Service from '@ember/service';
 import { typeOf, isPresent, isNone } from '@ember/utils';
 
-import { addSymbol } from '../ts-interfaces/utils/symbol';
 import require from 'require';
 import Ember from 'ember';
 import { assert, warn, inspect } from '@ember/debug';
@@ -54,34 +53,15 @@ import {
   HAS_SERIALIZER_PACKAGE,
 } from '@ember-data/private-build-infra';
 
-import { RecordInstance } from '../ts-interfaces/record-instance';
-import { JsonApiRelationship } from '../ts-interfaces/record-data-json-api';
-import { ResourceIdentifierObject } from '../ts-interfaces/ember-data-json-api';
-
 import promiseRecord from '../utils/promise-record';
 import { identifierCacheFor, IdentifierCache } from '../identifiers/cache';
 import { internalModelFactoryFor, setRecordIdentifier, recordIdentifierFor } from './store/internal-model-factory';
-import { RecordIdentifier, StableRecordIdentifier, StableExistingRecordIdentifier } from '../ts-interfaces/identifier';
 import { RecordReference, HasManyReference, BelongsToReference } from './references';
 import { Backburner } from '@ember/runloop/-private/backburner';
 import Snapshot from './snapshot';
-import {
-  EmptyResourceDocument,
-  SingleResourceDocument,
-  CollectionResourceDocument,
-  JsonApiDocument,
-  ExistingResourceObject,
-} from '../ts-interfaces/ember-data-json-api';
 import { RequestPromise } from './request-cache';
-import { PromiseProxy } from '../ts-interfaces/promise-proxies';
-import { DSModel } from '../ts-interfaces/ds-model';
 import NotificationManager from './record-notification-manager';
-import { AttributesSchema } from '../ts-interfaces/record-data-schemas';
-import { SchemaDefinitionService } from '../ts-interfaces/schema-definition-service';
 import ShimModelClass, { getShimClass } from './model/shim-model-class';
-import { RecordDataRecordWrapper } from '../ts-interfaces/record-data-record-wrapper';
-import { RecordData } from '../ts-interfaces/record-data';
-import { Dict } from '../ts-interfaces/utils';
 
 import constructResource from '../utils/construct-resource';
 import { errorsArrayToHash } from './errors-utils';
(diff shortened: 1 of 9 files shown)


packages/store/addon/-private/system/fetch-manager.ts
```diff
@@ -451,7 +464,9 @@ export default class FetchManager {
       // will once again convert the records to snapshots for adapter.findMany()
       let snapshots = new Array<Snapshot>(totalItems);
       for (let i = 0; i < totalItems; i++) {
-        let options = optionsMap.get(identifiers[i]);
+        // we know options is in the map due to having just set it above
+        // but TS doesn't know so we cast it
+        let options = optionsMap.get(identifiers[i]) as Dict<any>;
```
chriskrycho: Suggestion: use the same type as in the declaration, `as Object`, or use `!`. (I don't love `!` and it's easy for it to get missed when fixing these kinds of things, so `as Object` is probably better.)

packages/store/addon/-private/system/snapshot.ts
```diff
@@ -7,14 +7,51 @@ import { get } from '@ember/object';
...
+  // @mnorth has suggested that over time we will be able to use
+  // https://github.com/microsoft/rushstack/tree/master/apps/api-extractor
+  // to mark all APIs as public but separate out what is public vs private
+  // for docs and public types purposes.
+  // That said keeping things `private` helps us avoid breaking API
+  // boundaries internally so we should think about how to improve
+  // the design to avoid these sorts of casts but preserve the ability
+  // to have private helper utils for a module.
+  let i = (instance as unknown) as { _internalModel: InternalModel };
+  // TODO this case is not safe but it is the assumption of the current
```
chriskrycho:
```suggestion
  // TODO this cast is not safe but it is the assumption of the current
```

packages/store/addon/-private/system/snapshot.ts
```diff
@@ -25,14 +62,22 @@ function relationshipStateFor(instance, propertyName) {
...
 export default class Snapshot {
-  constructor(options, identifier, store) {
-    this.__attributes = null;
-    this._belongsToRelationships = Object.create(null);
-    this._belongsToIds = Object.create(null);
-    this._hasManyRelationships = Object.create(null);
-    this._hasManyIds = Object.create(null);
-    let internalModel = (this._internalModel = store._internalModelForResource(identifier));
-    this._store = store;
+  private __attributes: Dict<any> | null = null;
```
chriskrycho: Could we make these `Dict<unknown>` instead? Doing so forces the check on us in at least one place: setting up `adapterOptions`, which requires you to check that it's at least not bullish, which in turn guarantees that `public adapterOptions` not being optional is actually true. 😅 

```suggestion
  private __attributes: Dict<unknown> | null = null;
```

packages/store/addon/-private/system/snapshot.ts
```diff
@@ -98,34 +143,33 @@ export default class Snapshot {
...
+    if (CUSTOM_MODEL_CLASS) {
+      attrs.forEach(keyName => {
+        if (schemaIsDSModel(this.type)) {
+          attributes[keyName] = get(record as DSModel, keyName);
+        } else {
+          attributes[keyName] = recordDataFor(this._internalModel).getAttr(keyName);
+        }
+      });
+    } else {
+      (record as DSModel).eachAttribute(keyName => (attributes[keyName] = get(record as DSModel, keyName)));
```
chriskrycho: Is this cast safe because we're not in the `CUSTOM_MODEL_CLASS`? Can we do it with type narrowing or at a minimum add a comment?

packages/store/addon/-private/system/snapshot.ts
```diff
@@ -167,7 +209,7 @@ export default class Snapshot {
    @param {String} keyName
    @return {Object} The attribute value or undefined
    */
-  attr(keyName) {
+  attr(keyName: string): any {
```
chriskrycho:
```suggestion
  attr(keyName: string): unknown {
```

packages/store/addon/-private/system/snapshot.ts
```diff
@@ -187,7 +229,7 @@ export default class Snapshot {
    @method attributes
    @return {Object} All attributes of the current snapshot
    */
-  attributes() {
+  attributes(): Dict<any> {
```
chriskrycho:
```suggestion
  attributes(): Dict<unknown> {
```

packages/store/addon/-private/system/snapshot.ts
```diff
@@ -25,14 +62,22 @@ function relationshipStateFor(instance, propertyName) {
...
+  private _hasManyRelationships: Dict<Snapshot[]> = Object.create(null);
+  private _hasManyIds: Dict<RecordId[]> = Object.create(null);
+  private _internalModel: InternalModel;
+  private _changedAttributes: ChangedAttributesHash;
+
+  public identifier: StableRecordIdentifier;
+  public modelName: string;
+  public id: string | null;
+  public include?: string;
+  public adapterOptions: Dict<any>;
```
chriskrycho:
```suggestion
  public adapterOptions: Dict<unknown>;
```

packages/store/addon/-private/system/snapshot.ts
```diff
@@ -98,34 +143,33 @@ export default class Snapshot {
...
-          }
-        });
-      } else {
-        record.eachAttribute(keyName => (attributes[keyName] = get(record, keyName)));
-      }
+    if (CUSTOM_MODEL_CLASS) {
+      attrs = Object.keys(this._store._attributesDefinitionFor(this.modelName, this.identifier));
+    } else {
+      attrs = Object.keys(this._store._attributesDefinitionFor(this.modelName));
+    }
```
chriskrycho: Non-TS, and mostly a point of curiosity (esp. since it predates this) – any particular preference for `let` and the `if`/`else` vs. just doing this?

```ts
const attrs = CUSTOM_MODEL_CLASS
  ? Object.keys(this._store._attributesDefinitionFor(this.modelName, this.identifier))
  : Object.keys(this._store._attributesDefinitionFor(this.modelName));
```

packages/store/addon/-private/system/snapshot.ts
```diff
@@ -274,12 +316,16 @@ export default class Snapshot {
       );
     }
 
-    relationship = relationshipStateFor(this, keyName);
+    // TODO @runspired it seems this code branch would not work with CUSTOM_MODEL_CLASSes
+
+    // TODO @runspired instead of casting here either generify relationship state or
+    // provide a mechanism on relationship state by which to narrow.
```
chriskrycho: 👍 I like the proposed direction there.

packages/store/addon/-private/system/snapshot.ts
```diff
@@ -294,9 +340,9 @@ export default class Snapshot {
     }
 
     if (id) {
-      this._belongsToIds[keyName] = result;
+      this._belongsToIds[keyName] = result as RecordId;
     } else {
-      this._belongsToRelationships[keyName] = result;
+      this._belongsToRelationships[keyName] = result as Snapshot;
     }
```
chriskrycho: Can we add a comment explaining what makes these casts safe, so that if the code around it changes later we'll at least have some signal that we need to do to make sure the invariants still hold?

packages/store/addon/-private/system/snapshot.ts
```diff
@@ -362,18 +414,20 @@ export default class Snapshot {
         let internalModel = store._internalModelForResource(member);
         if (!internalModel.isDeleted()) {
           if (ids) {
-            results.push(member.id);
+            (results as RecordId[]).push(member.id);
           } else {
-            results.push(internalModel.createSnapshot());
+            (results as Snapshot[]).push(internalModel.createSnapshot());
           }
```
chriskrycho: Same request as above for a comment explaining what makes this hold.

packages/store/addon/-private/system/snapshot.ts
```diff
@@ -395,14 +449,14 @@ export default class Snapshot {
     @param {Function} callback the callback to execute
     @param {Object} [binding] the value to which the callback's `this` should be bound
   */
-  eachAttribute(callback, binding) {
+  eachAttribute(callback: (key: string, meta: AttributeSchema) => void, binding?: any): void {
```
chriskrycho:
```suggestion
  eachAttribute(callback: (key: string, meta: AttributeSchema) => void, binding?: unknown): void {
```

packages/store/addon/-private/system/snapshot.ts
```diff
@@ -422,14 +476,14 @@ export default class Snapshot {
     @param {Function} callback the callback to execute
     @param {Object} [binding] the value to which the callback's `this` should be bound
   */
-  eachRelationship(callback, binding) {
+  eachRelationship(callback: (key: string, meta: RelationshipSchema) => void, binding?: any): void {
```
chriskrycho:
```suggestion
  eachRelationship(callback: (key: string, meta: RelationshipSchema) => void, binding?: unknown): void {
```

packages/store/addon/-private/system/snapshot.ts
```diff
@@ -422,14 +476,14 @@ export default class Snapshot {
...
     if (CUSTOM_MODEL_CLASS) {
       let relationshipDefs = this._store._relationshipsDefinitionFor(this.modelName, this.identifier);
       Object.keys(relationshipDefs).forEach(key => {
         callback.call(binding, key, relationshipDefs[key]);
       });
     } else {
-      this.record.eachRelationship(callback, binding);
+      (this.record as DSModel).eachRelationship(callback, binding);
```
chriskrycho: As in the previous place where we had the same check on `CUSTOM_MODEL_CLASS`, it'd be great if we could push this into the type system or at least give it a comment explaining why it's safe so folks don't break it!

packages/store/addon/-private/system/snapshot.ts
```diff
@@ -458,7 +512,7 @@ export default class Snapshot {
     @param {Object} options
     @return {Object} an object whose values are primitive JSON values only
    */
-  serialize(options) {
-    return this.record.store.serializerFor(this.modelName).serialize(this, options);
+  serialize(options: any): unknown {
```
chriskrycho:
```suggestion
  serialize(options: unknown): unknown {
```

packages/store/addon/-private/system/snapshot.ts
runspired: It's safe because we are not in that flag branch, there isn't a good way to do it with narrowing because we can't type model yet but I can add a comment.

runspired: (to clarify, we can't type model && there isn't a good prop for us to be able to narrow by that won't be there for other known alternative models)

packages/store/addon/-private/system/snapshot.ts
runspired: no :) I prefer using a ternary here but I think we reach for the if/else out of habit.

packages/store/addon/-private/system/snapshot.ts
chriskrycho: That's what I figured. Thanks for clarifying!

packages/store/addon/-private/system/fetch-manager.ts
runspired: made them both `Dict<unknown>`

packages/store/addon/-private/system/snapshot.ts
runspired: I ended up with a slightly different approach due to how this is threaded through the whole method for both `belongsTo` and `hasMany`, basically these methods are used in one of two "modes", returning id/s or returning snapshot/s. I renamed `id` and `ids` to `returnModeIsId` and `returnModeIsIds` to clarify this.

packages/store/addon/-private/system/core-store.ts
```diff
@@ -1136,7 +1136,13 @@ abstract class CoreStore extends Service {
     // TODO  remove this once we dont rely on state machine
     internalModel.loadingData();
     let identifier = internalModel.identifier;
-    let promise = this._fetchManager.scheduleFetch(internalModel.identifier, options, generateStackTrace);
+
+    // TODO when we upgrade to TS 3.7 we can make use of assert
```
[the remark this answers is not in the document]
runspired: we already upgraded to 3.7 so we no longer ship this in prod :)

packages/store/addon/-private/system/core-store.ts
```diff
@@ -2466,7 +2472,9 @@ abstract class CoreStore extends Service {
       let pendingItem = pending[i];
       let snapshot = pendingItem.snapshot;
       let resolver = pendingItem.resolver;
-      let internalModel = snapshot._internalModel;
+      // TODO We have to cast due to our reliance on this private property
+      // this will be refactored away once we change our pending API to be identifier based
+      let internalModel = ((snapshot as unknown) as { _internalModel: InternalModel })._internalModel;
```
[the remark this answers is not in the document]
runspired: there are very few places we rely on leaking this detail of Snapshot, seems better to lock it down and leave the remnant intentional so we can work on eliminating the entanglements.

packages/store/addon/-private/system/fetch-manager.ts
```diff
@@ -318,7 +323,12 @@ export default class FetchManager {
     for (let i = 0, l = expectedSnapshots.length; i < l; i++) {
       let snapshot = expectedSnapshots[i];
 
-      if (!found[snapshot.id]) {
+      // TODO @runspired
+      // We know id is a string because you can't fetch
+      // without one.
+      // we should refactor to keying pending by identifier.lid
```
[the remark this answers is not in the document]
runspired: we can because that matching is done elsewhere, here we are dealing with Snapshots and identifiers

packages/store/addon/-private/system/fetch-manager.ts
```diff
@@ -339,14 +349,17 @@ export default class FetchManager {
 
   rejectFetchedItems(seeking: { [id: string]: PendingFetchItem }, snapshots: Snapshot[], error?) {
     for (let i = 0, l = snapshots.length; i < l; i++) {
-      let identifier = snapshots[i];
-      let pair = seeking[identifier.id];
+      let snapshot = snapshots[i];
+      // TODO refactor to identifier.lid to avoid this cast to string
+      //  we can do this case because you can only fetch an identifier
+      //  that has an ID
```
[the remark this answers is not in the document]
runspired: Same as above, we can because that matching is done elsewhere, here we are dealing with Snapshots and identifiers. Id usage is incidental and not required at this layer.

packages/store/addon/-private/system/core-store.ts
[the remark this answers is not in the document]
igorT: We discussed a new plan with another internalSnapshotInterface

packages/store/addon/-private/system/fetch-manager.ts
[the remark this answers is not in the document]
igorT: We discussed this in person and realized that the comment isnt quite correct

packages/store/addon/-private/system/fetch-manager.ts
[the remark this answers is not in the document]
igorT: SAme here

packages/store/addon/-private/system/snapshot.ts
igorT: @chriskrycho I think for feature flags specifically, if/else are easier to scan & read because you can just jump between blocks, and possibly collapse them compared to interpreting the ternary

packages/store/addon/-private/system/snapshot.ts
```diff
@@ -7,14 +7,51 @@ import { get } from '@ember/object';
...
+  // state of the code. We need to update this class to handle CUSTOM_MODEL_CLASS
+  // requirements.
+  let recordData = i._internalModel._recordData as RelationshipRecordData;
 
   return recordData._relationships;
 }
 
-function relationshipStateFor(instance, propertyName) {
+function schemaIsDSModel(schema: ModelSchema | DSModelSchema): schema is DSModelSchema {
+  return 'isModel' in schema && schema.isModel === true;
```
igorT: Why do you need the in check?

packages/store/addon/-private/system/snapshot.ts
```diff
@@ -25,14 +62,22 @@ function relationshipStateFor(instance, propertyName) {
...
+  private _belongsToIds: Dict<RecordId> = Object.create(null);
+  private _hasManyRelationships: Dict<Snapshot[]> = Object.create(null);
+  private _hasManyIds: Dict<RecordId[]> = Object.create(null);
+  private _internalModel: InternalModel;
+  private _changedAttributes: ChangedAttributesHash;
+
+  public identifier: StableRecordIdentifier;
+  public modelName: string;
+  public id: string | null;
+  public include?: string;
```
igorT: Maybe make this as unknown?

packages/store/addon/-private/system/snapshot.ts
```diff
@@ -252,18 +296,18 @@ export default class Snapshot {
    relationship or null if the relationship is known but unset. undefined
    will be returned if the contents of the relationship is unknown.
    */
-  belongsTo(keyName, options) {
-    let id = options && options.id;
-    let relationship;
-    let inverseInternalModel;
-    let result;
+  belongsTo(keyName: string, options?: { id?: boolean }): Snapshot | RecordId | undefined {
+    const returnModeIsId = !!(options && options.id);
```
igorT: I don't think this needs to be a const

packages/store/addon/-private/system/snapshot.ts
```diff
@@ -7,14 +7,51 @@ import { get } from '@ember/object';
...
+  // this private API usage.
+  // @mnorth has suggested that over time we will be able to use
+  // https://github.com/microsoft/rushstack/tree/master/apps/api-extractor
+  // to mark all APIs as public but separate out what is public vs private
+  // for docs and public types purposes.
+  // That said keeping things `private` helps us avoid breaking API
+  // boundaries internally so we should think about how to improve
+  // the design to avoid these sorts of casts but preserve the ability
+  // to have private helper utils for a module.
+  let i = (instance as unknown) as { _internalModel: InternalModel };
```
igorT: We can also use InternalSnapshot here

packages/store/addon/-private/system/snapshot.ts
```diff
@@ -331,17 +379,19 @@ export default class Snapshot {
...
    */
-  hasMany(keyName, options) {
-    let ids = options && options.ids;
-    let relationship;
-    let results;
-
-    if (ids && keyName in this._hasManyIds) {
-      return this._hasManyIds[keyName];
+  hasMany(keyName: string, options?: { ids?: boolean }): RecordId[] | Snapshot[] | undefined {
+    const returnModeIsIds = !!(options && options.ids);
```
igorT: Don't think should be using const here

packages/store/addon/-private/ts-interfaces/ds-model.ts
```diff
@@ -8,13 +8,13 @@ import { RelationshipSchema, AttributeSchema } from './record-data-schemas';
 export interface DSModel extends RecordInstance, EmberObject {
   toString(): string;
   save(): RSVP.Promise<DSModel>;
-  eachRelationship(callback: (key: string, meta: RelationshipSchema) => void): void;
-  eachAttribute(callback: (key: string) => void): void;
+  eachRelationship<T>(callback: (this: T, key: string, meta: RelationshipSchema) => void, binding?: T): void;
```
igorT: ❤️

PR #6808, opened 2019-11-26
[FIX] improve tarball insertion

runspired:
Since we wrote this script, yarn has improved and we might now be able to rely on its use of resolutions. Resolutions seems to be the only way to force partners that have addons with dependencies on ember-data to use the version we want, otherwise things like `ember-cli-addon-docs` install a stable release as well.

### DIFF
--- a/bin/-tarball-info.js
+++ b/bin/-tarball-info.js
@@ -166,8 +166,11 @@ function generatePackageReference(version, tarballName) {
 }
 
 function insertTarballsToPackageJson(fileLocation, options = {}) {
+  // in some flows we have the potential to have previously written
+  //  to the package.json already prior to calling this method.
+  //  this ensures we get the latest and not a stale module
   const location = require.resolve(fileLocation);
-  delete require[location];
+  delete require.cache[location];
   const pkgInfo = require(location);
   if (options.isRelativeTarball) {
     pkgInfo.version = `${pkgInfo.version}-sha.${CurrentSha}`;
--- a/bin/-tarball-info.js
+++ b/bin/-tarball-info.js
@@ -166,7 +166,9 @@ function generatePackageReference(version, tarballName) {
 }
 
 function insertTarballsToPackageJson(fileLocation, options = {}) {
-  const pkgInfo = require(fileLocation);
+  const location = require.resolve(fileLocation);
+  delete require[location];
+  const pkgInfo = require(location);
   if (options.isRelativeTarball) {
     pkgInfo.version = `${pkgInfo.version}-sha.${CurrentSha}`;
   }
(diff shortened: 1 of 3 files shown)


bin/test-external-partner-project.js
```diff
@@ -109,34 +120,43 @@ try {
...
+
+  // clear node_modules installed for the smoke-test
+  execExternal(`rm -rf node_modules`);
+  // we are forced to use yarn so that our resolutions will be respected
+  // in addition to the version file link we insert otherwise nested deps
+  // may bring their own ember-data
+  //
+  // For this reason we don't trust the lock file
+  // we also can't trust the cache
+  execExternal(`yarn install --no-lockfile --cache-folder=tmp/yarn-cache`);
```
rwjblue: FWIW, npm and yarn aren't drop in replacements for each other. There are quite a few bizarre inconsistencies (e.g. around using `file:` or `link:` references for deps/devDeps).

runspired: Before we were using npm regardless of whether the partner used yarn or npm because yarn was failing to extract the tarballs. This appears to have been fixed! However, I discovered a bug with npm wherein both during link and when installing from tarball nested dependencies specifying ember-data wont resolve to our tarball even if it is "in range" (I think this may be due to it being an alpha).

So we can't use npm safely, so using yarn+resolutions (to force the nested deps) always seems like the way to go.

bin/-tarball-info.js
```diff
@@ -166,7 +166,9 @@ function generatePackageReference(version, tarballName) {
 }
 
 function insertTarballsToPackageJson(fileLocation, options = {}) {
-  const pkgInfo = require(fileLocation);
+  const location = require.resolve(fileLocation);
+  delete require[location];
```
rwjblue: Why do we need to do this? If the intent is to purge the require cache, this isn't the correct way to do it.

The simplest thing would be to do `delete require.cache[location]`, but you may still leak the prior entry in the specific parent that required it.

See https://github.com/ember-cli/ember-cli-htmlbars/blob/d1b9683dc770042be91f4ac96f93d58dcaaed74f/lib/utils.js#L113-L126 for a reference.

bin/test-external-partner-project.js
rwjblue: FWIW, tarballs can never be "in range". They are always treated as "exotic" and the system cannot make assumptions about greater than / less than.

bin/-tarball-info.js
runspired: I forgot the .cache thanks for catching that

runspired: As for why, this is mostly a preventative measure since for the data packages we sometimes write a version to the file earlier in the process in some command flows.

bin/test-external-partner-project.js
runspired: @rwjblue all the more reason to not install with npm unless they’ve added support for resolutions that I don’t know about (based on testing they don’t seem to respect them)

bin/-tarball-info.js
```diff
@@ -166,7 +166,12 @@ function generatePackageReference(version, tarballName) {
 }
 
 function insertTarballsToPackageJson(fileLocation, options = {}) {
-  const pkgInfo = require(fileLocation);
+  // in some flows we have the potential to have previously written
+  //  to the package.json already prior to calling this method.
+  //  this ensures we get the latest and not a stale module
+  const location = require.resolve(fileLocation);
+  delete require.cache[location];
+  const pkgInfo = require(location);
```
rwjblue: FWIW, we can also avoid this issue via:

```js
const location = require.resolve(fileLocation);
let pkgInfo = JSON.parse(fs.readFileSync(location, 'utf-8'));
```

runspired: changed

PR #6809, opened 2019-11-26
[BUGFIX] cleanup rollup warnings

runspired:

### DIFF
--- a/packages/private-build-infra/src/addon-build-config-for-data-package.js
+++ b/packages/private-build-infra/src/addon-build-config-for-data-package.js
@@ -3,6 +3,7 @@ const Funnel = require('broccoli-funnel');
 const merge = require('broccoli-merge-trees');
 const BroccoliDebug = require('broccoli-debug');
 const rollupPrivateModule = require('./utilities/rollup-private-module');
+const chalk = require('chalk');
 
 function isProductionEnv() {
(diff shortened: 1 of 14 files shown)


packages/private-build-infra/src/addon-build-config-for-data-package.js
```diff
@@ -45,10 +46,34 @@ function addonBuildConfigForDataPackage(PackageName) {
...
+          console.log(
+            chalk.yellow(
+              `\n\n⚠️  Add ${chalk.white(
+                message.source
+              )} to the array returned by externalDependenciesForPrivateModule in index.js of ${chalk.white(
+                this.name
+              )}\n\n`
+            )
+          );
+          return;
```
rwjblue: Instead of suppressing these, we should fail the build when developing. This will still have the same result (consumers will not see warnings), but actually forces us to ensure the repo is in the correct shape.

runspired: @rwjblue I initially did make it throw but figured outputting a clear warning with specific guidance on how to fix it might be enough. I can still throw after this warning.

rwjblue: Opened #6850 to address my concern.

PR #6810, opened 2019-11-26
[CHORE] run lint with --fix on every commit

runspired:

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,6 +5,7 @@
     "packages/*"
   ],
   "pre-commit": [
+    "lint:js",
     "lint:js:changed"
   ],
   "scripts": {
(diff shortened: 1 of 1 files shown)


runspired:
@rwjblue I'll label this for discussion later then, @igorT previously expressed an interest in the lint actually being a full lint and I know my own experience is that it's very frustrating to have the lint test fail CI because I forgot to run the check myself first.

igorT:
@runspired the changes should've been caught by the changed lint pre commit no? Afai remember those were caused by only running prettier locally and not all of eslint, but going over the code it seems that now we run the full eslint so we should be good? Maybe that script in bin is buggy? Full eslint takes a bit too long locally, and I agree overall with @rwjblue points.

runspired:
A few thoughts:

- full lint takes a few milliseconds once the cache is warm, it only takes time if you haven't run lint for the project in a very long time.

- the bin script isn't buggy per-say, just more limited currently, it just only runs a few specific rules. This changes it to run the whole set.

- lint errors can be introduced in files you didn't edit if you change imports/exports/types

- we commit so regularly the odds of a prettier change being unrelated to your commit is tiny

PR #6814, opened 2019-11-27
[FEAT] Strip resolved Deprecations

runspired:
Allows build-time stripping of resolved deprecations from production builds.

- With this change, users may configure EmberData to remove support for deprecated behaviors by supplying the minimum version of EmberData with which they are fully compatible.

For instance, if a deprecation was introduced in 3.13, and the app specifies 3.13 as its minimum version compatibility, any deprecations introduced before or during 3.13 would be stripped away.

This configuration is done by an app's `ember-cli-build` file and looks like this:

```js
let app = new EmberApp(defaults, {
    emberData: {
      compatWith: '3.12',
    },
});
``` 

- Instead of tracking down the precise versions in which all of our existing deprecations were added, wherever I was unsure they have been conservatively set to at least 2 minor versions after they were introduced, and in some cases many more.

```ts
export default {
  DEPRECATE_EVENTED_API_USAGE: '3.12',
  DEPRECATE_RECORD_LIFECYCLE_EVENT_METHODS: '3.12',
  DEPRECATE_MODEL_DATA: '3.8',
  DEPRECATE_MODEL_TOJSON: '3.15',
  DEPRECATE_LEGACY_TEST_HELPER_SUPPORT: '3.15',
  DEPRECATE_LEGACY_TEST_REGISTRATIONS: '3.15',
  DEPRECATE_DEFAULT_SERIALIZER: '3.15',
  DEPRECATE_DEFAULT_ADAPTER: '3.15',
  DEPRECATE_METHOD_CALLS_ON_DESTROY_STORE: '3.15',
  DEPRECATE_MISMATCHED_INVERSE_RELATIONSHIP_DATA: '3.12',
};
```

- This PR additionally adds a github-actions test scenario in which we run our tests with all deprecations stripped.

- Deprecation flags are named imports from `@ember-data/private-build-infra/deprecations`, e.g.

```ts
import { DEPRECATE_EVENTED_API_USAGE } from '@ember-data/private-build-infra/deprecations';
```

- Deprecation flag names follow the pattern of `DEPRECATE_{deprecation-id-in-caps}`


packages/unpublished-test-infra/index.js
```diff
@@ -1,14 +1,17 @@
...
-
+module.exports = Object.assign({}, addonBaseConfig, {
   treeForAddon() {
     if (process.env.EMBER_CLI_TEST_COMMAND) {
-      let tree = version();
-      return this._super.treeForAddon.call(this, tree);
+      const options = this.getEmberDataConfig();
+      let compatVersion = options.compatWith;
+      let tree = version(compatVersion);
+      return this.debugTree(this._super.treeForAddon.call(this, tree), 'addon-output');
```
Gaurav0: Did you forget to remove this.debugTree?

runspired: no, debugTree is something we leave in permanently. It's a noop until you need it. It outputs to the file system when you use `BROCCOLI_DEBUG=` with a pattern that includes the name of the tree. In our case we namespace all of these to our packages e.g. `ember-data:@ember-data/adapter:addon-output` sorta deal

runspired:
RE tests, your suggestion of writing a test that asserts roughly:

```
let wasStripped = true;
if (DEPRECATE_SOMETHING) {
  wasStripped = false;
}

assert.ok(wasStripped);
```

looks great. We should find a way to implement this and then test several different flag levels (LTS 3.8 and 3.12 and then "everything" seem like good candidates). I'll open an issue to track that.

packages/-ember-data/ember-cli-build.js
```diff
@@ -3,13 +3,16 @@
...
+    emberData: {
+      compatWith,
+    },
     babel: {
       // this ensures that the same `@ember-data/canary-features` processing that the various
-      // ember-data addons do is done in the dummy app
-      plugins: [
-        ...require('@ember-data/private-build-infra/src/debug-macros')(null, process.env.EMBER_ENV === 'production'),
-      ],
+      // ember-data addons do is done in the dummy app and tests
```
rwjblue: don't think you added this, but the phrasing is pretty odd and I don't understand what the comment is saying

packages/debug/index.js
```diff
@@ -25,6 +25,12 @@ module.exports = Object.assign({}, addonBaseConfig, {
     const options = this.getEmberDataConfig();
     const env = process.env.EMBER_ENV;
 
+    const parentIsEmberDataAddon = this.parent.pkg.name === 'ember-data';
+
+    if (options.includeDataAdapterInProduction === undefined) {
+      options.includeDataAdapterInProduction = parentIsEmberDataAddon;
+    }
```
rwjblue: ~~This seems vaguely unrelated to deprecation stripping (possibly a bug fix on its own merits), why is it included in this PR?~~

I see, this was moved.

packages/model/addon/-private/model.js
```diff
@@ -1375,34 +1392,41 @@ if (DEBUG) {
...
-
-  let lookupDeprecations = function lookupInstanceDeprecations(instance) {
-    let deprecations = INSTANCE_DEPRECATIONS.get(instance);
-
-    if (!deprecations) {
-      deprecations = new Set();
-      INSTANCE_DEPRECATIONS.set(instance, deprecations);
-    }
+  let lookupDeprecations;
+  let DEPRECATED_LIFECYCLE_EVENT_METHODS;
```
rwjblue: Could probably leave this defined as a const and do:

```js
const DEPRECATED_LIFECYLE_EVENT_METHODS = DEPRECATED_RECORD_LIFECYCLE_EVENT_METHODS === true && [
  // ...snip...
];
```

Super minor, feel free to disregard.

rwjblue: Also, I find the naming of these two variables to be pretty confusing (one is the array of methods, the other is the deprecation flag) since they are so similarly named. Maybe you can rename the local method name variable?

packages/private-build-infra/addon/current-deprecations.ts
```diff
@@ -0,0 +1,14 @@
+export default {
+  DEPRECATE_CATCH_ALL: '99.0',
```
rwjblue: What is this one for?

packages/private-build-infra/src/create-version-module.js
```diff
@@ -28,6 +28,12 @@ function calculateVersion() {
...
 
-module.exports = function() {
-  return createFile('version.js', 'export default "' + calculateVersion() + '";');
+module.exports = function(compatVersion) {
+  return createFile(
+    'version.js',
+    'export default "' +
+      calculateVersion() +
+      '";\n' +
+      (compatVersion && compatVersion !== 'null' ? `export const COMPAT_VERSION = "${compatVersion}";\n` : '')
```
rwjblue: This is intentionally checking for the string "null"? Seems slightly odd, why is that?

packages/store/addon/-private/system/store/finders.js
```diff
@@ -169,20 +168,21 @@ function ensureRelationshipIsSetToParent(payload, parentInternalModel, store, pa
       typeof relationshipData !== 'undefined' &&
       !relationshipDataPointsToParent(relationshipData, parentInternalModel)
     ) {
-      let quotedType = Ember.inspect(type);
-      let quotedInverse = Ember.inspect(inverseKey);
-      let expected = Ember.inspect({
+      let inspect = function inspect(thing) {
+        return `'${JSON.stringify(thing)}'`;
+      };
```
rwjblue: FWIW, `Ember.inspect` does a bit more than `JSON.stringify` (deals with cycles, symbols, function properties, etc). It may not matter in this case (I can't tell what the actual type of `thing` is expected to be).

packages/unpublished-test-infra/addon-test-support/deprecated-test.js
```diff
@@ -33,10 +33,15 @@ export function deprecatedTest(testName, deprecation, testCallback) {
     }
   }
 
+  let testFn = test;
+  if (COMPAT_VERSION && gte(COMPAT_VERSION, VERSION)) {
```
rwjblue: I don't fully grok this. How can `COMPAT_VERSION` ever be higher than the current version?

rwjblue:
I also think we need to add documentation (somewhere) explaining how you opt in to this, what it does, etc.

packages/model/addon/-private/model.js
runspired: this was an already existing variable but thank you for catching the similarity, I'll rename it.

packages/private-build-infra/addon/current-deprecations.ts
runspired: It allows you to write something in tests that is only stripped if all deprecations are stripped, we don't use it yet and could remove it but I suspect we will want it for adding tests for the infra.

packages/private-build-infra/src/create-version-module.js
runspired: left over from my initial implementation of 

```
  const compatWith = process.env.EMBER_DATA_FULL_COMPAT ? '99.0' : null;
```

packages/store/addon/-private/system/store/finders.js
runspired: nothing with a cycle :)

runspired:
@rwjblue I'll open a ticket make sure we add api docs once it lands in stable.

packages/-ember-data/ember-cli-build.js
runspired: I'll see if I can figure it out and improve it

packages/unpublished-test-infra/addon-test-support/deprecated-test.js
runspired: COMPAT_VERSION can be set to any number. We use this to allow ourselves to aggressively strip everything by setting it to `99`.

PR #6815, opened 2019-11-27
[DOC] JSONAPIAdapter's findMany

sly7-7:

### DIFF
--- a/packages/adapter/addon/json-api.js
+++ b/packages/adapter/addon/json-api.js
@@ -1,5 +1,5 @@
 /**
-  @module @ember-data/adapter
+@module @ember-data/adapter
 */
 import { dasherize } from '@ember/string';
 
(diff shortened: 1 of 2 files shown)


locks:
Hi! Can you address the linting errors?

sly7-7:
@locks Is it good enough now ?

sly7-7:
@locks, I'm sorry, I don't understand the `@static`thing here, because this method is not a static one. 😕

locks:
The Ember Data tests verify that the "table of contents" for API documentation did not change.
Since you are introducing `findMany` by documenting it, you need to tell https://github.com/emberjs/data/blob/master/packages/-ember-data/node-tests/fixtures/expected.js to expect `findMany`!

packages/adapter/addon/json-api.js
```diff
@@ -218,9 +218,35 @@ const JSONAPIAdapter = RESTAdapter.extend({
...
+    ```js
+    {
+      "id": 1,
+      "title": "Rails is omakase",
+      "comments": [ 1, 2, 3 ]
+    }
+    ```
+    The IDs will be passed in this form:
+    ```
+    filter[id]=1,2,3
```
runspired, line 236: The example here appears to have very little to do with how this method operates and even less to do with JSON:API format. I'd suggest looking at the `findMany` documentation from the `MinimumAdapterInterface` for inspiration.

sly7-7:
Sorry for the noise, I think I've been too much agressive during my housekeeping...

locks:
Closing due to focus on the request manager going forward. Thanks for contributing!

PR #6817, opened 2019-11-27
[CHORE] remove unneeded configs

runspired:
These configs simply re-export our root config, which doesn't provide any value over just having the root config but does introduce the risk that we make config edits in some packages we don't intend.

tsc uses our root config and vscode will always find it there as well. ember-cli-typescript likely doesn't need it at all but if it does we can point it to the right place.

### DIFF
--- a/packages/-ember-data/tsconfig.json
+++ b/packages/-ember-data/tsconfig.json
@@ -1,3 +1,0 @@
-{
-  "extends": "../../tsconfig.json"
-}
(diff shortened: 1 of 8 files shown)


igorT:
Seems reasonable to me, but would love to check with the resident ts experts. Do you know why we put it there in the first place?

runspired:
> Do you know why we put it there in the first place?

We didn't know any better and our scaffolding generates files there for us.

PR #6818, opened 2019-11-27
[CHORE] cleanup build

runspired:
Refactor to use _findHost and provide a shared mechanism for accessing `EmberData` config.

Also removes test-infra's VERSION module from our production output if we aren't in a testing environment.

### DIFF
--- a/packages/debug/index.js
+++ b/packages/debug/index.js
@@ -4,16 +4,6 @@ const name = require('./package').name;
 const addonBuildConfigForDataPackage = require('@ember-data/private-build-infra/src/addon-build-config-for-data-package');
 const addonBaseConfig = addonBuildConfigForDataPackage(name);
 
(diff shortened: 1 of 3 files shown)


packages/debug/index.js
```diff
@@ -32,23 +22,11 @@ module.exports = Object.assign({}, addonBaseConfig, {
     if (this.__isEnabled !== null) {
       return this.__isEnabled;
     }
-    const options = this.setupOptions();
-    const env = getApp(this).env;
+    const options = this.getEmberDataConfig();
+    const env = process.env.EMBER_ENV;
```
igorT: What caused this change? Do we not need to do getApp for env?

igorT:
Seems reasonable to me, but @rwjblue would be a much better set of 👀

PR #6820, opened 2019-11-27
[CHORE] improve comment bot

runspired:
This PR

- changes the comment-bot to update existing comments instead of deleting old comments
- fixes the bug that was preventing deletion of old comments
- fixes the performance analysis script to also post a comment correctly (it's been failing to do so although trying for some time)

You can see the effects of the updates (edit history on the pr comments) within this PR. Overall it seems like a good change, although I dislike how the comment time stamp remains the time of the very original comment instead of showing the time edited. This makes it harder to judge quickly if it is the most recent analysis, although checking the commit sha that is printed will verify.

### DIFF
--- a/bin/asset-size-tracking/src/post-comment.sh
+++ b/bin/asset-size-tracking/src/post-comment.sh
@@ -33,9 +33,10 @@ API_VERSION=v3
 API_HEADER="Accept: application/vnd.github.${API_VERSION}+json; application/vnd.github.antiope-preview+json"
 AUTH_HEADER="Authorization: token ${GITHUB_TOKEN}"
 
-delete_comment_if_exists() {
+update_comment_if_exists() {
   # Get all the comments for the pull request.
   body=$(curl -sSL -H "${AUTH_HEADER}" -H "${API_HEADER}" "${URI}/repos/${GITHUB_REPOSITORY}/issues/${NUMBER}/comments")
+  FOUND_EXISTING=1
 
   for row in $(echo -E "${body}" | jq --raw-output  '.[] | @base64'); do
     comment=$(echo -E "${row}" | base64 --decode | jq --raw-output '{id: .id, body: .body, author: .user.login}')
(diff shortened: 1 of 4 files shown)


runspired:
Ilios failure seems to be due to them having merged a bad PR

runspired:
cc @jrjohnson

jrjohnson:
Thanks for the heads up @runspired looks like we have a floating dependency with a breaking change. It's our own package so entirely my fault. I'll get that cleaned up ASAP.

bin/asset-size-tracking/src/post-comment.sh
```diff
@@ -33,9 +33,10 @@ API_VERSION=v3
 API_HEADER="Accept: application/vnd.github.${API_VERSION}+json; application/vnd.github.antiope-preview+json"
 AUTH_HEADER="Authorization: token ${GITHUB_TOKEN}"
 
-delete_comment_if_exists() {
+update_comment_if_exists() {
```
igorT: Unrelated to this PR but as these grow/become more complicated should we rewrite them in js?

bin/asset-size-tracking/src/post-comment.sh
```diff
@@ -46,15 +47,19 @@ delete_comment_if_exists() {
       # We have found our comment.
       # Delete it.
```
igorT: Leftover old comment

bin/asset-size-tracking/src/post-comment.sh
```diff
@@ -46,15 +47,19 @@ delete_comment_if_exists() {
...
+      curl -sSL -H "${AUTH_HEADER}" -H "${API_HEADER}" -d "$COMMENT_TEXT" -H "Content-Type: application/json" -X PATCH $UPDATE_URL
+      FOUND_EXISTING=0
     fi
   done
+
+  return $FOUND_EXISTING;
 }
 
 post_comment() {
+  echo "Posting new comment"
```
igorT: 👍

igorT:
My gmail sends many ❤️

PR #6822, opened 2019-11-28
[CHORE] Move PromiseBelongsTo, PromiseManyArray and ManyArray to model package

Gaurav0:
For #6166 

Also fixes some missing docs


packages/model/addon/index.ts
```diff
@@ -39,4 +39,4 @@
   @public
  */
 
-export { Model as default, attr, belongsTo, hasMany } from './-private';
```
runspired: We shouldn't re-export private-only things here

packages/store/addon/-private/system/promise-proxies.js
```diff
@@ -85,73 +83,8 @@ export function promiseArray(promise, label) {
...
-
-  reload(options) {
-    assert('You are trying to reload an async belongsTo before it has been created', this.get('content') !== undefined);
-    let { key, store, originatingInternalModel } = this._belongsToState;
-
-    return store.reloadBelongsTo(this, originatingInternalModel, key, options).then(() => this);
-  },
-});
-
 export function proxyToContent(method) {
```
runspired: If we move this too the entanglement will reduce

packages/store/index.js
```diff
@@ -7,6 +7,12 @@ const addonBaseConfig = addonBuildConfigForDataPackage(name);
 module.exports = Object.assign({}, addonBaseConfig, {
   shouldRollupPrivate: true,
   externalDependenciesForPrivateModule() {
-    return ['@ember-data/canary-features', 'ember-inflector', '@ember-data/store/-debug', 'require'];
+    return [
+      '@ember-data/canary-features',
+      'ember-inflector',
+      '@ember-data/store/-debug',
+      'require',
+      '@ember-data/model',
```
runspired: if we're doing this, we're breaking encapsulation, so I'm skeptical

packages/store/addon/-private/system/model/internal-model.ts
```diff
@@ -55,14 +64,27 @@ function relationshipStateFor(instance: InternalModel, propertyName: string) {
...
+if (HAS_MODEL_PACKAGE) {
+  _getModelPackage = function() {
+    if (!_found) {
+      let modelPackage = require('@ember-data/model/-private');
+      ({ ManyArray, PromiseBelongsTo, PromiseManyArray } = modelPackage);
+      if (ManyArray && PromiseBelongsTo && PromiseManyArray) {
+        _found = true;
+
+        // Some tests require PromiseManyArray instanceof PromiseArray to be true
+        Object.setPrototypeOf(PromiseManyArray, PromiseArray);
```
runspired: if we're doing this hack here to fix tests then something unexpected has changed which we should fix

packages/model/addon/index.ts
Gaurav0: Fixed

packages/store/index.js
Gaurav0: Fixed.

PR #6824, opened 2019-11-29
[TEST] generateIdForRecord tests for adapter encapsulation side quest

Gaurav0:
For #6666

### DIFF
--- a/packages/unpublished-adapter-encapsulation-test-app/tests/integration/generate-id-test.js
+++ b/packages/unpublished-adapter-encapsulation-test-app/tests/integration/generate-id-test.js
@@ -81,7 +81,7 @@ module('integration/generate-id - GenerateIdForRecord Tests', function(hooks) {
     });
   });
 
-  test('store.createRecord does not error if adapter.generateIdForRecord is undefined.', async function (assert) {
+  test('store.createRecord does not error if adapter.generateIdForRecord is undefined.', async function(assert) {
     let store = this.owner.lookup('service:store');
     let expectedData = {
       data: {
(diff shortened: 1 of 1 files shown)


packages/unpublished-adapter-encapsulation-test-app/tests/integration/generate-id-test.js
```diff
@@ -0,0 +1,106 @@
...
+    this.owner.register('adapter:application', TestGenerateIdForRecordAdapter);
+
+    let props = expectedData.data.attributes;
+    let record = store.createRecord('person', props);
+
+    assert.equal(record.id, 'manually generated id 1', 'manually generated id used');
+
+    let recordFromPeekRecord = store.peekRecord('person', record.id);
+
+    assert.equal(record, recordFromPeekRecord, 'peekRecord returns the same record');
```
runspired: let's use strictEqual here

packages/unpublished-adapter-encapsulation-test-app/tests/integration/generate-id-test.js
```diff
@@ -0,0 +1,106 @@
...
+  });
+
+  test('store.createRecord calls adapter.generateIdForRecord if defined and we use this ID for the record', async function(assert) {
+    let generateIdForRecordCalled = 0;
+    let seq = 0;
+
+    let store = this.owner.lookup('service:store');
+    let expectedData = {
+      data: {
+        id: '12',
```
runspired: let's remove this `id` as it's not part of what we expect

packages/unpublished-adapter-encapsulation-test-app/tests/integration/generate-id-test.js
```diff
@@ -0,0 +1,106 @@
...
+        attributes: props,
+      },
+    });
+  });
+
+  test('store.createRecord does not error if adapter.generateIdForRecord is undefined.', async function(assert) {
+    let store = this.owner.lookup('service:store');
+    let expectedData = {
+      data: {
+        id: '12',
```
runspired: let's not include this id as it's not part of what we expect (we expect `null`)

PR #6828, opened 2019-11-30
[CHORE] Fix workflow and enable partner test debugging

### LINKED ISSUE #6825: Ember Observer partner test hangs indefinitely
This is only happening for the github-action based version of the check, it does not happen for azure.

runspired:
resolves #6825 

This PR

- Fixes our YML so that we actually run tests on all the Operating Systems we'd specified (they were all running ubuntu before)
- Adds timeouts to our jobs so that they don't hang for hours (the default timeout is 6 hours)
- Uploads artifacts from failed partner-test runs so that we can debug them better
- Fixes the bug it enabled debugging that was causing EmberObserver to fail

## Why EmberObserver was Timing Out in Github Actions only

This turned out to be a canary-only dev-build-only transpilation bug that only affects partners transpiling for IE11 in their tests.

The primary difference between Travis/Azure and Github Actions is that in Github Actions we correctly set `env` to include `CI=true`. This causes ember-observer to target `ie11`, which
introduces a transpilation bug when using `typeof`.

Our code in `canary-features` (this is stripped in production and non-canary builds)

```ts
const ENV = typeof EmberDataENV === 'object' && EmberDataENV !== null ? EmberDataENV : {};
```

Due to this default babel plugin https://babeljs.io/docs/en/babel-plugin-transform-typeof-symbol
this is transpiled to 

```ts
  function _typeof(obj) { if (typeof Symbol === "function" && typeof Symbol.iterator === "symbol") { _typeof = function _typeof(obj) { return typeof obj; }; } else { _typeof = function _typeof(obj) { return obj && typeof Symbol === "function" && obj.constructor === Symbol && obj !== Symbol.prototype ? "symbol" : typeof obj; }; } return _typeof(obj); }

  var ENV = _typeof(EmberDataENV) === 'object' && EmberDataENV !== null ? EmberDataENV : {};
```

The above is an error because we call `_typeof` with `EmberDataENV` when it is potentially undefined. We don't hit this in our own `max-transpilation-test` because our tests always have this variable defined via the test's `index.html` file.

Changing the check to this removes the transpilation as we no longer need to check for `symbol`

```ts
const ENV: ConfigEnv = typeof EmberDataENV !== 'undefined' && EmberDataENV !== null ? EmberDataENV : {};
```

Here we are more loosely assuming that if we are defined and not null we are an object, but given the very narrow set of things that hit this code path this seems ok.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -217,4 +217,4 @@ jobs:
         uses: actions/upload-artifact@v1
         with:
           name: ${{ matrix.partner }}-dist
-          path: ../../data/__external-test-cache/${{ matrix.partner }}/tmp/
+          path: ../../data/__external-test-cache/${{ matrix.partner }}/dist
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -14,6 +14,7 @@ on:
 
 jobs:
   lint:
+    timeout-minutes: 4
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@master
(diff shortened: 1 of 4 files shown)


igorT:
Looks fine to me, but not sure, what are we fixing here?

runspired:
@igorT updated description for you

PR #6829, opened 2019-11-30
[BUGFIX] Octane: should not need to use get with model.id

Gaurav0:
For #6143 

Edit: Now includes a WIP bugfix, but it has TypeScript errors I don't know how to fix.

### DIFF
--- a/packages/-ember-data/tests/acceptance/tracking-model-id-test.js
+++ b/packages/-ember-data/tests/acceptance/tracking-model-id-test.js
@@ -4,10 +4,13 @@ import { render, settled } from '@ember/test-helpers';
 import hbs from 'htmlbars-inline-precompile';
 import Store from '@ember-data/store';
 import Model, { attr } from '@ember-data/model';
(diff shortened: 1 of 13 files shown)


packages/-ember-data/tests/acceptance/tracking-model-id-test.js
```diff
@@ -0,0 +1,101 @@
...
+import { dependentKeyCompat } from '@ember/object/compat';
+import { gte } from 'ember-compatibility-helpers';
+import { has } from 'require';
+import { resolve } from 'rsvp';
+
+if (gte('3.14.0') && has('@glimmer/component')) {
+  class Widget extends Model {
+    @attr() name;
+
+    @dependentKeyCompat
```
pzuraq: This shouldn't be necessary it looks like, nothing is depending on this property directly. This would be necessary if you had:

```js
@computed('numericId')
```

Somewhere, or an observer that was trying to watch it. The template layer does not use observers or computeds anymore.

runspired:
Assigned Igor for backporting.

PR #6830, opened 2019-12-01
[CHORE] IE11 Testing

runspired:
We're still a ways (if ever) from being able to run the full test suite in IE11 https://github.com/emberjs/data/pull/6830/checks?check_run_id=327832618

But this lets us run the production tests.

The small number of changes to EmberData itself are mostly for dev-time support of IE11 (which I got pretty far with but ultimately the widespread lack of support for IE11 at dev time in Ember itself prevents us from getting any further).

Some of the changes will affect prod builds but only once feature flags are turned on (request-state cache for instance).

### DIFF
--- a/packages/unpublished-test-infra/addon-test-support/watch-property.js
+++ b/packages/unpublished-test-infra/addon-test-support/watch-property.js
@@ -1,6 +1,13 @@
 import { removeObserver, addObserver } from '@ember/object/observers';
 import QUnit from 'qunit';
 
+function freeze(obj) {
+  if (typeof Object.freeze === 'function') {
+    return Object.freeze(obj);
+  }
+  return obj;
+}
+
 function makeCounter() {
   let count = 0;
   const counter = Object.create(null);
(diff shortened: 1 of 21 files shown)


packages/unpublished-test-infra/addon-test-support/watch-property.js
```diff
@@ -1,6 +1,13 @@
 import { removeObserver, addObserver } from '@ember/object/observers';
 import QUnit from 'qunit';
 
+function freeze(obj) {
```
Gaurav0: Let's import this function from one freeze.ts file

runspired: We may want to do this long term but we would have a separate util for runtime vs tests. currently we only use freeze in one location in tests (here) and two locations in store. The locations in store could possible share a definition but both are DEBUG only and thus stripped away, likely we'll refactor it to a util if we hit a third case.

rwjblue:
> ultimately the widespread lack of support for IE11 at dev time in Ember itself prevents us from getting any further

I don't accept this statement, please file issues for anything you ran into.

runspired:
Will need a local version to do that but it is definitely the case that DEBUG is a minefield of external issues, mostly with Proxy and string methods

PR #6831, opened 2019-12-02
[BUGFIX TESTS] matrix tests for fastboot & ie11 w/jquery & fetch

### LINKED ISSUE #6431: [FEAT CI] Fastboot and IE11 Integration Tests
We should test the following

*IE11*

- [x] run all tests with jQuery
- [x] run all tests without jQuery

*Fastboot*

- [x] run all tests with jQuery
- [x] run all tests without jQuery

runspired:
resolves #6431 

*Bugfix: jQuery+fastboot*

The added tests for fastboot revealed that fastboot+jQuery was broken because our computed property for getting the fastboot service wasn't setting and updating it's value properly leading to it always being false. This led us down the wrong code-path and ultimately meant that the url we provided to `najax` wasn't parsed correctly and would error.


packages/adapter/addon/rest.js
```diff
@@ -301,10 +301,10 @@ const RESTAdapter = Adapter.extend(BuildURLMixin, {
       if (this._fastboot) {
         return this._fastboot;
       }
-      return getOwner(this).lookup('service:fastboot');
+      return (this._fastboot = getOwner(this).lookup('service:fastboot'));
```
igorT: This seems like a bugfix in a test/ci fix? Is this fixing a bug? Can we separate it into a different PR then?

runspired: see the updated commit, I'll change the PR to also state bugfix, the bug is uncovered by the tests

igorT: Don't see the bug in the commit, maybe you didn't push?

.github/workflows/main.yml
```diff
@@ -87,9 +85,33 @@ jobs:
...
+  fastboot-tests:
+    needs: [lint]
+    timeout-minutes: 6
+    strategy:
+      matrix:
+        os: [macos-latest, windows-latest, ubuntu-latest]
+        scenario: [fastboot-with-ember-fetch, fastboot-with-jquery]
+    runs-on: ${{ matrix.os }}
+    steps:
+      - if: matrix.os == 'macos-latest'
```
igorT: Is there a way to factor this setup set as we are repeating it in a bunch of places

packages/unpublished-fastboot-test-app/config/ember-try.js
```diff
@@ -0,0 +1,33 @@
+'use strict';
+
+module.exports = function() {
+  return {
+    useYarn: true,
+    scenarios: [
+      {
+        name: 'fastboot-with-ember-fetch',
+        env: {
+          EMBER_OPTIONAL_FEATURES: JSON.stringify({ 'jquery-integration': false }),
```
igorT: The ember-try config for -ember-data doesn't set `jquery-integration` to false. Is that difference on purpose?

runspired: ill clean that up so it’s clear what we are testing there. Main thread has 4 variants to test where fastboot has 2.

- native fetch
- ember fetch jquery true
- ember fetch jquery false
- jquery true no ember fetch

runspired: We only care about the outcome (does it work with jquery, does it work with fetch) in fastboot.

igorT: but none of the main thread variations set jquery-integration to false? https://github.com/emberjs/data/blob/master/packages/-ember-data/config/ember-try.js

packages/adapter/addon/rest.js
runspired: I meant I added `BUGFIX` to the commit, I will also update the description with the nature of the bugfix.

.github/workflows/main.yml
runspired: no, each of these is a bit different or a unique scenario and abstracting to a stand-alone action would result in just as many lines

packages/unpublished-fastboot-test-app/config/ember-try.js
runspired: @igorT we default jquery integration to false throughout the project, but we should make it explicit in the tests so that we don't accidentally get trolled.

igorT: Yeah, my comment was that it seemed that here we hard code to false, but we don't in other similar places

runspired: Made the scenarios explicit

PR #6839, opened 2019-12-03
[DEPRECATION] deprecate belongsToReference.push accepting a model

Gaurav0:
For #6166


packages/store/addon/-private/system/references/belongs-to.js
```diff
@@ -3,6 +3,7 @@ import { assertPolymorphicType } from '@ember-data/store/-debug';
 import Reference from './reference';
 import recordDataFor from '../record-data-for';
 import { peekRecordIdentifier } from '../store/internal-model-factory';
+import { deprecate } from '@ember/application/deprecations';
```
runspired: importing from `@ember/debug` works and is a better long term location now

packages/store/addon/-private/system/references/belongs-to.js
```diff
@@ -128,8 +129,11 @@ export default class BelongsToReference extends Reference {
     return resolve(objectOrPromise).then(data => {
       let record;
 
-      // TODO deprecate data as Model
       if (peekRecordIdentifier(data)) {
+        deprecate('Pushing a model into a BelongsToReference is deprecated', false, {
```
runspired:
```suggestion
        deprecate('Pushing a record into a BelongsToReference is deprecated. You should push data in as a single-resource JSON:API document.', false, {
```

runspired: Maybe the above might be more clear? cc @igorT

packages/store/addon/-private/system/references/belongs-to.js
```diff
@@ -128,10 +130,17 @@ export default class BelongsToReference extends Reference {
     return resolve(objectOrPromise).then(data => {
       let record;
 
-      // TODO deprecate data as Model
-      if (peekRecordIdentifier(data)) {
-        record = data;
-      } else {
+      if (DEPRECATE_BELONGS_TO_REFERENCE_PUSH) {
+        if (peekRecordIdentifier(data)) {
```
runspired: We can combine these to one line and keep the if/else branching to avoid the `record === undefined` check.

e.g.

```ts
if (DEPRECATE_BELONGS_TO_REFERENCE_PUSH && peekRecordIdentifier(data)) {

} else {

}
```

Issue #6687, opened 2019-11-06
Chron job to check for external deprecations


runspired:
We should leverage our new deprecation infrastructure to generate a nightly report of any deprecations that our test suite currently generates that aren't our own. Mostly these would come from ember-source but ember-qunit or test helpers etc. may also raise deprecations.

- [x] add deprecation assertion to all tests in all packages
- [x] run all tests with an ENV variable enabling asserting no deprecations of any kind
- [x] find a way to generate a list of the encountered deprecations
- [ ] if job fails we ?? (open issue?)

runspired:
cc @Gaurav0 think you could work on this?

PR #6844, opened 2019-12-03
[CHORE] Cron job to check for external deprecations

Gaurav0:
For #6687

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -1,0 +1,36 @@
+name: 'Nightly'
+
+on:
+  schedule:
+    - cron: '* 8 * * *' # 8am in UTC is midnight PST
+    branches:
+      - master
+
+jobs:
+  test-all-deprecations:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@master
+      - uses: actions/setup-node@v1
+        with:
+          node-version: 12.x
+          repo-token: ${{ secrets.GITHUB_TOKEN }}
+      - name: Check SHA
+        run: |
+          sha=$(git rev-parse --short=8 HEAD)
+          echo "HEAD sha=$sha"
+          echo "GITHUB_SHA sha=$GITHUB_SHA"
+          mkdir -p tmp
+          echo $sha > tmp/sha-for-check.txt
+          originSha=$(git rev-parse HEAD^2)
+          echo $originSha > tmp/sha-for-commit.txt
+          git show --format=short --no-patch $originSha
+      - name: Checkout master
+        run: git checkout master
+      - name: Install dependencies for master
+        run: yarn install
+      - name: Basic Tests
+        env:
+          CI: true
+          ASSERT_ALL_DEPRECATIONS: true
+        run: yarn test
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -2,12 +2,16 @@ name: 'Nightly'
 
 on:
   schedule:
-    - cron: '* 8 * * *' # 8am in UTC is midnight PST
+    - cron: '* 14 * * *' # 14:00 in UTC is 6am PST
     branches:
       - master
 
 jobs:
   test-all-deprecations:
+    strategy:
+      fail-fast: false
+      matrix:
+        scenario: [ember-beta, ember-canary]
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v1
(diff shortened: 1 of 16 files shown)


packages/unpublished-test-infra/addon-test-support/assert-all-deprecations.js
```diff
@@ -0,0 +1,62 @@
...
+          hooks.unshift(assertAllDeprecations);
+        }
+      });
+    }
+  });
+
+  QUnit.done(function() {
+    if (ASSERT_ALL_DEPRECATIONS) {
+      let deprecations = {};
+      for (let deprecation of ALL_ASSERTED_DEPRECATIONS) {
```
runspired: this is an array so we can use `for (let i = 0 ....)` instead, which is important as otherwise this won't run in IE11

packages/unpublished-test-infra/addon-test-support/assert-all-deprecations.js
```diff
@@ -0,0 +1,62 @@
...
+        if (deprecation in deprecations) {
+          deprecations[deprecation]++;
+        } else {
+          deprecations[deprecation] = 1;
+        }
+      }
+
+      QUnit.config.deprecations = deprecations;
```
runspired: was there a motivation for building this object from the array here vs incrementing properties on a dictionary as we iterate the test suite?

packages/unpublished-test-infra/addon-test-support/assert-all-deprecations.js
```diff
@@ -0,0 +1,62 @@
...
+          id.includes('ds.') ||
+          id.includes('emberdata') ||
+          id.includes('ember-data') ||
+          id.includes('mismatched-inverse-relationship-data-from-payload');
+
+        if (!ASSERT_ALL_DEPRECATIONS && !isEmberDataDeprecation) {
+          console.warn('Detected Non-Ember-Data Deprecation:', deprecation.message, deprecation.options.stacktrace);
+        }
+        if (ASSERT_ALL_DEPRECATIONS) {
+          ALL_ASSERTED_DEPRECATIONS.push((deprecation.options && deprecation.options.id) || deprecation);
```
runspired: we don't know from within this filter method whether the deprecation is handled, only that it is encountered.

I think we can just ignore things that match `isEmberDataDeprecation` since we already assert all our own in regular tests and thus can be sure they are handled.  This will reduce the noise in the report.

.github/workflows/nightly.yml
```diff
@@ -0,0 +1,24 @@
...
+        with:
+          node-version: 12.x
+          repo-token: ${{ secrets.GITHUB_TOKEN }}
+      - name: Install dependencies for master
+        run: yarn install
+      - name: Basic Tests
+        env:
+          CI: true
+          ASSERT_ALL_DEPRECATIONS: true
+        run: yarn test
```
runspired: we should run

- basic tests
- fastboot tests
- fetch tests
- encapsulation tests
- canary/beta Ember tests

runspired: In a future PR we should also consider how to monitor for deprecations in the relationship-performance-test-app, but that can be ignored for now.

.github/workflows/nightly.yml
```diff
@@ -0,0 +1,24 @@
+name: 'Nightly'
+
+on:
+  schedule:
+    - cron: '* 8 * * *' # 8am in UTC is midnight PST
```
runspired: let's do something like 8am PST so that no one gets their sleep interrupted with email notifications

.github/workflows/nightly.yml
```diff
@@ -0,0 +1,24 @@
...
+    branches:
+      - master
+
+jobs:
+  test-all-deprecations:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@master
```
runspired:
```suggestion
      - uses: actions/checkout@v1
```

.github/workflows/nightly.yml
runspired: Actually let's do 6am

packages/unpublished-test-infra/addon-test-support/assert-all-deprecations.js
Gaurav0: 👍

packages/unpublished-test-infra/addon-test-support/assert-all-deprecations.js
Gaurav0: Nope.

packages/unpublished-test-infra/addon-test-support/assert-all-deprecations.js
Gaurav0: Ok!

.github/workflows/nightly.yml
Gaurav0: Done.

.github/workflows/nightly.yml
Gaurav0: I didn't set up fastboot tests and they aren't working right now, so not adding them currently. What are the fetch tests?
Added the others.

runspired: @Gaurav0 can you clarify what you mean by:

> I didn't set up fastboot tests and they aren't working right now

I am happy to merge this and add fastboot and fetch tests in follow up

Gaurav0: Yes, I reverted updating test-helper.js in fastboot-test-app to assert on deprecations. I think the code I wrote with the custom QUnitAdapter for testem doesn't work in fastboot, that may be why it fails.

PR #6849, opened 2019-12-04
[CHORE] Cleanup imports/exports

runspired:


igorT:
Why did you want to delete the store wrapper interface?

runspired:
@igorT I'm 50:50 on that.

Roughly speaking, either I need to upgrade the interface to the class lots more places or delete it, and exporting the upgrade infra to upgrade it in other packages comes at a cost at the moment (I have a build-infra change I'm cooking up to eliminate that cost).

After thinking about why we have it more, and talking with @mike-north, I feel our path-to-shipping types is more likely to include stripping type info we don't want to ship from the classes we choose to ship types for rather than maintaining separate types.

It currently feels like a pre-mature optimization to me. We rightly wanted to try to constrain ourselves to its public API but it brings a lot of hassle.

Since we end up needing the upgraded version everywhere, the hope that it would help us constrain our private API usage seems to be unrealized. That said, we *are* close and we don't have much private API in the thing.

I'll defer to your judgement, I'm happy to restore it.

runspired:
closing as no longer relevant, most of this has since happened (if not all of it)

PR #6850, opened 2019-12-04
Make rollup warnings build errors.

rwjblue:
This will ensure that all messages that are emitted to the console by rollup trigger a build failure and therefore prevent landing any changes to Ember Data that would introduce such a warning.

Follow up to https://github.com/emberjs/data/pull/6809#discussion_r350961572

### DIFF
--- a/packages/private-build-infra/src/addon-build-config-for-data-package.js
+++ b/packages/private-build-infra/src/addon-build-config-for-data-package.js
@@ -45,35 +45,28 @@ function addonBuildConfigForDataPackage(PackageName) {
       }
     },
 
-    _suppressUneededRollupWarnings(message, next) {
+    _handleRollupWarning(message) {
       if (message.code === 'CIRCULAR_DEPENDENCY') {
         return;
-      } else if (message.code === 'NON_EXISTENT_EXPORT') {
+      } else if (message.code === 'NON_EXISTENT_EXPORT' && message.message.includes('ts-interfaces')) {
         // ignore ts-interface imports
-        if (message.message.indexOf(`/ts-interfaces/`) !== -1) {
-          return;
-        }
-      } else if (message.code === 'UNRESOLVED_IMPORT') {
-        if (!this.isDevelopingAddon()) {
-          // don't print these for consumers
-          return;
-        } else {
-          const chalk = require('chalk');
-          // make warning actionable
-          // eslint-disable-next-line no-console
-          console.log(
-            chalk.yellow(
-              `\n\n⚠️  Add ${chalk.white(
-                message.source
-              )} to the array returned by externalDependenciesForPrivateModule in index.js of ${chalk.white(
-                this.name
-              )}\n\n`
-            )
-          );
-          throw message.message;
-        }
+        return;
+      } else {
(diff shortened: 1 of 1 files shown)


packages/private-build-infra/src/addon-build-config-for-data-package.js
```diff
@@ -45,35 +45,28 @@ function addonBuildConfigForDataPackage(PackageName) {
       }
     },
 
-    _suppressUneededRollupWarnings(message, next) {
+    _handleRollupWarning(message) {
       if (message.code === 'CIRCULAR_DEPENDENCY') {
         return;
-      } else if (message.code === 'NON_EXISTENT_EXPORT') {
+      } else if (message.code === 'NON_EXISTENT_EXPORT' && message.message.includes('ts-interfaces')) {
```
runspired: the removed slashes were there intentionally although I do see how it might look like a borked regex ;)

It's unlikely for this shorter string to match anything the longer string didn't match but the slashes were insurance we were finding a warning.

On the code style change I'm mixed. Folks rushed into using `contains` and `includes` and `startsWith` in our library and test code and I recently had to revert all of that to get us back to IE11 support. I like using new things but I dislike mixing styles. Overall though this is clearly a build file so its a bit more clear that it works here.

packages/private-build-infra/src/addon-build-config-for-data-package.js
```diff
@@ -45,35 +45,28 @@ function addonBuildConfigForDataPackage(PackageName) {
...
         // ignore ts-interface imports
-        if (message.message.indexOf(`/ts-interfaces/`) !== -1) {
-          return;
-        }
-      } else if (message.code === 'UNRESOLVED_IMPORT') {
-        if (!this.isDevelopingAddon()) {
```
runspired: we shouldn't remove this check as rollup, babel, rollup plugins, broccoli, and broccoli-rollup may all be a different version in a consuming app than what we test and ship with. It's better to print an unnecessary warning than to break the build for those folks.

packages/private-build-infra/src/addon-build-config-for-data-package.js
```diff
@@ -45,35 +45,28 @@ function addonBuildConfigForDataPackage(PackageName) {
...
-            )
-          );
-          throw message.message;
-        }
+        return;
+      } else {
+        const chalk = require('chalk');
+        // make warning actionable
+        // eslint-disable-next-line no-console
+        console.log(
```
runspired: this warning (which is nicely actionable) would not pertain to any errors that don't match the removed `message.code === 'UNRESOLVED_IMPORT'`. We should bring back that condition.

packages/private-build-infra/src/addon-build-config-for-data-package.js
rwjblue: > Folks rushed into using contains and includes and startsWith in our library and test code and I recently had to revert all of that to get us back to IE11 support. I like using new things but I dislike mixing styles.

Ya, but this is Node code and has no requirement for IE11 support.

runspired:
These have been build errors on master for some time, closing as unnecessary

PR #6857, opened 2019-12-05
fix nightly.yml

Gaurav0:
I wish I could figure out how to test this.

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -9,10 +9,6 @@ on:
 
 jobs:
   test-all-deprecations:
-    strategy:
-      fail-fast: false
-      matrix:
-        scenario: [ember-beta, ember-canary]
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v1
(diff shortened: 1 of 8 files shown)


runspired:
@Gaurav0 maybe we should make it run on PRs while we get it working? now that it's on master likely it will trigger

.github/workflows/nightly.yml
```diff
@@ -3,15 +3,9 @@ name: 'Nightly'
 on:
   schedule:
     - cron: '* 14 * * *' # 14:00 in UTC is 6am PST
-    branches:
-      - master
```
runspired: let's bring this back, we still only want to run this on master we just also wanted to run it during PR while deving

runspired: NVM: 

> You can schedule a workflow to run at specific UTC times using POSIX cron syntax. Scheduled workflows run on the latest commit on the default or base branch. The shortest interval you can run scheduled workflows is once every 5 minutes.

https://help.github.com/en/actions/automating-your-workflow-with-github-actions/events-that-trigger-workflows#scheduled-events-schedule

We don't need the branch specifier

PR #6861, opened 2019-12-06
[DOCS] fix inclusion of docs for deprecation and canary features

runspired:

### DIFF
--- a/packages/-ember-data/yuidoc.json
+++ b/packages/-ember-data/yuidoc.json
@@ -13,7 +13,9 @@
       "../serializer/addon",
       "../store/addon",
       "../record-data/addon",
-      "../debug/addon"
+      "../debug/addon",
+      "../private-build-infra/addon",
+      "../canary-features/addon"
     ],
     "exclude": "vendor",
     "outdir": "dist/docs"
--- a/packages/private-build-infra/addon/index.ts
+++ b/packages/private-build-infra/addon/index.ts
@@ -1,7 +1,3 @@
-/**
-  @module @ember-data/private-build-infra
-*/
-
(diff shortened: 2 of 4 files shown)


packages/-ember-data/yuidoc.json
```diff
@@ -13,7 +13,9 @@
       "../serializer/addon",
       "../store/addon",
       "../record-data/addon",
-      "../debug/addon"
+      "../debug/addon",
+      "../private-build-infra/addon",
```
igorT: Why are we adding the private-build infra to yuidocs?

packages/private-build-infra/addon/index.ts
```diff
@@ -1,7 +1,3 @@
-/**
-  @module @ember-data/private-build-infra
-*/
-
 import { has } from 'require';
```
igorT: Why are we removing this?

runspired: we're removing it cause we don't document this package directly in the api docs so having yuidocs comments for it just means we accidentally create an empty package we wouldn't otherwise have

igorT:
The PR title says that we are fixing deprecation and canary features docs, but the non tests changes don't mention deprecations, just the canary-features?

packages/-ember-data/yuidoc.json
runspired: we need to parse to get the `@ember-data/deprecations` guide

runspired:
@igorT the guide for deprecations is a fake package in the file in which we define the current flags for deprecations called `@ember-data/deprecations`. That file is in the private-build-infra package

PR #6864, opened 2019-12-06
Revert "Bump broccoli-merge-trees from 3.0.2 to 4.0.0"

rwjblue:
Reverts emberjs/data#6863

This version drops support for Node 8, ember-data cannot upgrade until it is also ready to drop support (likely about a month from now...). CI in that PR passed because we do not actually run any CI jobs in Node 8 (maybe we should?).


rwjblue:
Closed in favor of https://github.com/emberjs/data/pull/6865

Issue #6851, opened 2019-12-04
`handleResponse` on the adapter is never called when using fetch and a response with no body


kategengler:
When using `fetch`, a response with a status and an empty body never hits `handleResponse` on the application adapter. 

If I turn `jquery-integration` back on, such that `useFetch` is true, then handleResponse is hit, for the same response code and empty body.

Jumping through the debugger, I can follow it down to `response.text()` in `determineBodyPromise`, but it never makes it to the `then` on that promise. It seems like there may be a missing `catch` in the chain of promises here https://github.com/emberjs/data/blob/master/packages/adapter/addon/rest.js#L1002  

### Versions

Run the following command and paste the output below: `yarn list ember-source && yarn list ember-cli && yarn list --pattern ember-data`.

```cli
└─ ember-source@3.14.2

└─ ember-cli@3.14.0

├─ @ember-data/-build-infra@3.14.0
├─ @ember-data/adapter@3.14.0
├─ @ember-data/canary-features@3.14.0
├─ @ember-data/model@3.14.0
├─ @ember-data/rfc395-data@0.0.4
├─ @ember-data/serializer@3.14.0
├─ @ember-data/store@3.14.0
└─ ember-data@3.14.0
```

runspired:
I think there's three changes to make and add test coverage for here:

- we need to catch and handle if determineBodyPromise errors
- we need to make sure if `fetch` itself rejects we still catch that and process correctly
- we should fix determineBodyPromise's ability to handle empty payloads

@Gaurav0 mind taking a look?

andreyfel:
I think I encountered a similar issue. When my server returns 401 with an empty body I'm getting "SyntaxError: Unexpected end of JSON input" in the application route error action instead of the 401 error.
It started happening in Ember Data 3.14. It worked as expected in ED 3.13.

A change in this file caused the issue:
packages/adapter/addon/-private/utils/determine-body-promise.ts
https://github.com/emberjs/data/pull/6293

There is a new `throw` that breaks the chain. Interesting that the throw happens only in `if (DEBUG)` check. So, in the production build the behavior would be different.

kategengler:
@andreyfel I think that you may be seeing the same bug, but I don't think the PR you linked to is the cause: In my testing, it never made it into the `then` on `response.text()` in that method, so it would never reach the code that was added.

andreyfel:
@kategengler In my case I've set a breakpoint on the code I was talking about and I could see that it was reached.

[...]

PR #6871, opened 2019-12-09
[BUGFIX] ensure handleResponse always called in rest adapter

Gaurav0:
For #6851

### DIFF
--- a/packages/-ember-data/tests/integration/adapter/handle-response-test.js
+++ b/packages/-ember-data/tests/integration/adapter/handle-response-test.js
@@ -1,0 +1,270 @@
+import { resolve, reject } from 'rsvp';
+import { setupTest } from 'ember-qunit';
+import { module, test } from 'qunit';
+import { DEBUG } from '@glimmer/env';
+import JSONAPIAdapter from '@ember-data/adapter/json-api';
+import AdapterError from '@ember-data/adapter/error';
+import JSONAPISerializer from 'ember-data/serializers/json-api';
+import Model, { attr } from '@ember-data/model';
+
+class Person extends Model {
+  @attr name;
+}
+module('integration/adapter/handle-response', function(hooks) {
+  setupTest(hooks);
+
+  hooks.beforeEach(function() {
+    this.owner.register('model:person', Person);
+    this.owner.register('serializer:application', JSONAPISerializer);
+
+    this.store = this.owner.lookup('service:store');
+  });
+  test('handleResponse is called with normal response', async function(assert) {
+    let handleResponseCalled = 0;
+
+    let samplePayload = {
+      data: [
+        {
+          id: 1,
+          type: 'person',
+          attributes: {
+            name: 'John Smith',
+          },
+        },
+        {
+          id: 2,
+          type: 'person',
+          attributes: {
+            name: 'Zhang San',
+          },
+        },
+      ],
+    };
+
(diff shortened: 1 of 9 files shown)


packages/-ember-data/tests/integration/adapter/handle-response-test.js
```diff
@@ -0,0 +1,212 @@
...
+
+    class TestAdapter extends JSONAPIAdapter {
+      handleResponse(status, headers, payload, requestData) {
+        handleResponseCalled++;
+
+        if (this.isSuccess(status, headers, payload)) {
+          return payload;
+        }
+
+        let errors = this.normalizeErrorResponse(status, headers, payload);
```
runspired: should we just super once we increment?

packages/-ember-data/tests/unit/utils/determine-body-promise-test.js
```diff
@@ -0,0 +1,75 @@
...
+  });
+
+  test('determineBodyResponse returns the body even if it is not json', function(assert) {
+    assert.expect(1);
+
+    const response = new Response('this is not json', { status: 200 });
+    const bodyPromise = determineBodyPromise(response, {});
+
+    return bodyPromise.then(body => {
+      assert.deepEqual(body, 'this is not json');
```
runspired: let's add a description here

packages/-ember-data/tests/unit/utils/determine-body-promise-test.js
```diff
@@ -0,0 +1,75 @@
...
+
+module('Unit | determineBodyPromise', function() {
+  test('determineBodyResponse returns the body when it is present', function(assert) {
+    assert.expect(1);
+
+    const response = new Response('{"data": "foo"}', { status: 200 });
+    const bodyPromise = determineBodyPromise(response, {});
+
+    return bodyPromise.then(body => {
+      assert.deepEqual(body, { data: 'foo' });
```
runspired: let's add a description

packages/-ember-data/tests/integration/adapter/rest-adapter-test.js
```diff
@@ -2576,10 +2576,6 @@ module('integration/adapter/rest_adapter - REST Adapter', function(hooks) {
         hash.error(jqXHR, jqXHR.responseText, errorThrown);
       };
 
-      adapter.handleResponse = function(status, headers, payload) {
```
runspired: why did we need to delete this? Are we now calling this method *more* often than we used to? (note how this is in a `hasJQuery` branch)

Gaurav0: Yes. We are now ensuring we are calling it always, whether fetch or jquery. Let me know if this is not correct.

packages/-ember-data/tests/integration/adapter/handle-response-test.js
Gaurav0: done.

packages/-ember-data/tests/integration/adapter/rest-adapter-test.js
Gaurav0: Update: I now no longer call it in this situation (a non JSON string; status 200)

igorT:
Do we want this on the LTS?

Issue #6867, opened 2019-12-06
Setting async belongsTo not working after previous fetch rejected


danwenzel:
### Description

Starting with ember-data version 3.12.0, setting a `belongsTo` relationship property on a model is not working after that relationship previously rejected.  As an example, if an async fetch of a post's `user` returns a 404, I'm then catching the error, creating a new user, and setting it to the post:

```js
async fetchOrCreateUser() {
  try {
    await this.post.get('user');
  } catch (error) {
    // 404 returned.  Create a new user and set it as the post's user
    const newUser = this.store.createRecord('user', {
      name: 'Larry',
    });
    this.post.set('user', newUser);
  }
},
```

#### Expected

`this.post.user` is now the new user

#### Actual

`this.post.user` continues to be the rejected promise from the first `get`

I'm pretty sure this has to do with this change: https://github.com/emberjs/data/pull/6112

This issue does not occur in version 3.11.5 and below.

### Reproduction - Failing Case (v3.12.0)

- Pull this branch: https://github.com/danwenzel/ember-data-relationship-issue/tree/3.12.0
- run `yarn && yarn start`
- visit "http://localhost:4200/
- click "Fetch or Create User"

Notice that the user name never populates, and `isRejected` is always true

### Reproduction - Non-failing Case (v3.11.5)

- Pull this branch: https://github.com/danwenzel/ember-data-relationship-issue/tree/3.11.5
- run `yarn && yarn start`
- visit "http://localhost:4200/
- click "Fetch or Create User"

Notice that the user name does populate successfully

### Versions - Failing Case

```cli
yarn list v1.19.1
warning Filtering by arguments is deprecated. Please use the pattern option instead.
└─ ember-source@3.13.4
✨  Done in 0.62s.
yarn list v1.19.1
warning Filtering by arguments is deprecated. Please use the pattern option instead.
└─ ember-cli@3.13.2
✨  Done in 0.63s.
yarn list v1.19.1
├─ @ember-data/-build-infra@3.12.0
├─ @ember-data/adapter@3.12.0
├─ @ember-data/canary-features@3.12.0
├─ @ember-data/model@3.12.0
├─ @ember-data/rfc395-data@0.0.4
├─ @ember-data/serializer@3.12.0
├─ @ember-data/store@3.12.0
└─ ember-data@3.12.0
```

### Versions - Non-failing Case

```cli
yarn list v1.19.1
warning Filtering by arguments is deprecated. Please use the pattern option instead.
[...]

PR #6874, opened 2019-12-10
[TEST] - Failing test for resetting previously failed linked async belongs-to

danwenzel:
Failing test for https://github.com/emberjs/data/issues/6867

### DIFF
--- a/packages/-ember-data/tests/acceptance/relationships/belongs-to-test.js
+++ b/packages/-ember-data/tests/acceptance/relationships/belongs-to-test.js
@@ -555,4 +555,37 @@ module('async belongs-to rendering tests', function(hooks) {
       }
     });
   });
+
+  test('Can reset a previously failed linked async belongs-to', async function(assert) {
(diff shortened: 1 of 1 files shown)


runspired:
@sly7-7 you might be able to debug why this happens, but happy to pair on it with you tomorrow morning (my tomorrow morning) if you're free and would like.

sly7-7:
@runspired I try to debug and I think I found the first "symptom". When accessing the belongsTo (even after setting it), the `relationship.state.hasFailedLoadAttempt` is true, so it returns immediatly the cached Proxy (with the null content).
So there is something wrong here, but I can't tell what. I naievely tried to reset the flag to `false` to see what happen, but in that case, it seems like there is a loop, when the relationship is fetched again and again.
So I guess something is missing when setting the belongsTo to the new record, like to update the cached proxy content and resetting it's state to something correct 🤷  But I don't know what exactly, neither where to do such things.

PR #6875, opened 2019-12-10
[CHORE] Lint enforcement of type-only imports

### LINKED ISSUE #6804: We should lint against type only imports using es style
If an import is only used for type information, typescript offers us a mechanism to ensure we only import the type. Using this style both prevents cycles in rollup and helps to ensure we have not broken package encapsulation.

We should find or write a lint rule such that an import using ES syntax being used only for type information is disallowed.

e.g.

Not

```ts
import Snapshot from './snapshot';

function withSnapshot(snapshot: Snapshot) {}
```

but

```ts
type Snapshot = import('./snapshot').default;

function withSnapshot(snapshot: Snapshot) {}
```

runspired:
resolves #6804 

Also runs the `--fix` for it.

~~This likely will reduce our asset-sizes.~~
**Edit:** looks like thanks to previous manual optimization of this + rollup we don't see a win here (previous manual optimization did result in wins). But this will help avoid future regressions :) 

Will revisit #6849 once this lands.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -43,12 +43,13 @@ module.exports = {
       parserOptions: {
         sourceType: 'module',
       },
-      plugins: ['@typescript-eslint'],
+      plugins: ['@typescript-eslint', 'ember-data'],
       extends: ['eslint:recommended', 'plugin:@typescript-eslint/eslint-recommended'],
       rules: {
         '@typescript-eslint/no-unused-vars': ['error', { args: 'none' }],
         'no-unused-vars': 'off',
         'require-atomic-updates': 'off',
+        'ember-data/prefer-type-only-import': 'error',
       },
     },
 
@@ -61,6 +62,7 @@ module.exports = {
         'bin/**',
         'packages/private-build-infra/src/**/*.js',
         'packages/unpublished-test-infra/src/**/*.js',
+        'packages/unpublished-eslint-rules/src/**/*.js',
         'packages/*/.ember-cli.js',
         'packages/*/.eslintrc.js',
         'packages/*/.template-lintrc.js',
--- a/package.json
+++ b/package.json
@@ -110,6 +110,7 @@
     "eslint-plugin-node": "^10.0.0",
     "eslint-plugin-prettier": "^3.1.1",
     "eslint-plugin-qunit": "^4.0.0",
+    "eslint-plugin-ember-data": "3.16.0-alpha.2",
     "execa": "^3.4.0",
     "fromentries": "^1.2.0",
     "fs-extra": "^8.1.0",
--- a/packages/-ember-data/tests/integration/identifiers/configuration-test.ts
+++ b/packages/-ember-data/tests/integration/identifiers/configuration-test.ts
@@ -14,12 +14,12 @@ import Model, { attr, belongsTo } from '@ember-data/model';
 import Adapter from '@ember-data/adapter';
 import Serializer from '@ember-data/serializer';
 import { resolve, all } from 'rsvp';
-import { ExistingResourceObject } from '@ember-data/store/-private/ts-interfaces/ember-data-json-api';
-import { StableRecordIdentifier } from '@ember-data/store/-private/ts-interfaces/identifier';
 import { set } from '@ember/object';
 import { run } from '@ember/runloop';
 import { settled } from '@ember/test-helpers';
 import { identifierCacheFor } from '@ember-data/store/-private';
+type StableRecordIdentifier = import('@ember-data/store/-private/ts-interfaces/identifier').StableRecordIdentifier;
+type ExistingResourceObject = import('@ember-data/store/-private/ts-interfaces/ember-data-json-api').ExistingResourceObject;
 
 if (IDENTIFIERS) {
   module('Integration | Identifiers - configuration', function(hooks) {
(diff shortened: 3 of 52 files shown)


igorT:
This lint rule seems more generically useful, should we publish it as a separate repo so other people can use it?

packages/unpublished-eslint-rules/index.js
```diff
@@ -0,0 +1,288 @@
+const TypeParamNamingSchema = [['T'], ['K', 'V'], ['T', 'K', 'V'], ['T', 'K', 'V', 'R']];
```
igorT: Is the structure of this code copied and adapter from somewhere or is this a 100% greenfield effort?

igorT:
Some things that you import as `type` have been `interfaces` while others have been classes, are there any consequences to this intermingling?

packages/unpublished-eslint-rules/index.js
runspired: It's greenfield

packages/unpublished-eslint-rules/index.js
```diff
@@ -0,0 +1,285 @@
...
+              local: node.local,
+              imported: node.imported || node.local,
+              from: node.parent.source,
+              isDefault: false,
+            };
+          },
+          Identifier(node) {
+            let type = node.parent.type;
+
+            if (type === 'ImportSpecifier' || type === 'ImportDefaultSpecifier') {
```
igorT: We can probably ignore more types here, like declarations etc.

packages/unpublished-eslint-rules/index.js
```diff
@@ -0,0 +1,285 @@
...
+              // console.log(`found ${node.name} for import type ${type}`);
+            } else if (type === 'TSTypeReference') {
+              // console.log(`found type reference ${node.name}`);
+              // do nothing, this Identifier is handled elsewhere
+            } else if (imports[node.name] !== undefined) {
+              const scope = context.getScope();
```
igorT: const?

packages/unpublished-eslint-rules/index.js
```diff
@@ -0,0 +1,285 @@
...
+              if (scope.type === 'class') {
```
igorT: Lets make the top if a switch or something like that, to make these nested ifs easier to read

packages/unpublished-eslint-rules/index.js
```diff
@@ -0,0 +1,285 @@
...
+            usedTypes = {};
+            lastImport = null;
+          },
+          'Program:exit'() {
+            // console.log('\n\n\nProgram:exitBegin\n======\n');
+            // eslint will only pass over a file 10x before bailing
+            // our approach lets us autofix more things by collapsing fixes for
+            // a single import declaration into one error+change instead of fix
+            // by individual imports within a declaration
+            for (let i = 0; i < declarations.length; i++) {
```
igorT: Maybe use something like forEach etc.

packages/unpublished-eslint-rules/index.js
```diff
@@ -0,0 +1,285 @@
+const TypeParamNamingSchema = [['T'], ['K', 'V'], ['T', 'K', 'V'], ['T', 'K', 'V', 'R']];
+const Alphabet = ['A', 'B', 'C', 'D', 'E', 'F', 'G', 'H', 'I', 'J', 'K', 'L', 'M', 'N', 'O', 'P'];
+
+function getUnusedSpecifiers(declaration, used) {
```
igorT: Can we write a comment for this

packages/unpublished-eslint-rules/index.js
```diff
@@ -0,0 +1,285 @@
...
+        let imports, usedImports, usedTypes, lastImport, declarations;
+        return {
+          Program() {
+            declarations = [];
+            imports = {};
+            usedImports = {};
+            usedTypes = {};
+            lastImport = null;
+          },
+          'Program:exit'() {
```
igorT: This method seems kinda large, maybe would be nice to break up a bit?

packages/unpublished-eslint-rules/index.js
```diff
@@ -0,0 +1,285 @@
...
+            // console.log('\n\n\nProgram:exitBegin\n======\n');
+            // eslint will only pass over a file 10x before bailing
+            // our approach lets us autofix more things by collapsing fixes for
+            // a single import declaration into one error+change instead of fix
+            // by individual imports within a declaration
+            for (let i = 0; i < declarations.length; i++) {
+              let declaration = declarations[i];
+              let unusedImports = getUnusedSpecifiers(declaration, usedImports);
+              let typeImports = getTypeSpecifiers(unusedImports, usedTypes);
```
igorT: Maybe rename this to typeOnlyImports?

packages/unpublished-eslint-rules/index.js
```diff
@@ -0,0 +1,285 @@
...
+          },
+          TSTypeReference(node) {
+            let identifier = node.typeName;
+            if (imports[identifier.name] !== undefined) {
+              // console.log(`found used type ${identifier.name}`);
+              if (node.typeParameters && node.typeParameters.params.length) {
+                // this type usage contains generics e.g. `Dict<Thing>`
+                // so we save off how many params it uses to fix up the import
+                // later
+                usedTypes[identifier.name] = node.typeParameters.params.length;
```
igorT: Maybe create another object to keep the type params so we don't have both booleans and numbers in this one

packages/unpublished-eslint-rules/index.js
```diff
@@ -0,0 +1,285 @@
...
+    }
+    for (let i = 0; i < paramValue; i++) {
+      params.push(alphabet[i]);
+    }
+    paramString = `<${params.join(', ')}>`;
+  }
+  return paramString;
+}
+
+function getRangeToRemove(exportToRemove) {
```
igorT: Can we describe this method a bit more please

PR #6879, opened 2019-12-10
[CHORE lint] configure and run import ordering rules

runspired:
Many EmberData modules have dozens of dependencies, these rules help us keep them quickly scannable and should reduce the number of merge conflicts that arise when imports change.

Most of the rules added here are autofixable. Ordering of `require` statements in `node` is not, and enforcing no-duplicates has a bug in its autofix so it is set to `warn` instead.

The order we enforce is:

- side-effect-imports
- glimmer & ember dependencies
- library dependencies
- library dependencies beginning with ember-
- @ember-data/ dependencies
- anything else we missed for libraries
- relative imports

There will be a single extra line in between these.


package.json
```diff
@@ -140,7 +140,10 @@
     "prepare-local-packages": "./bin/packages-for-commit.js",
     "publish": "./bin/publish.js"
   },
-  "dependencies": {},
+  "dependencies": {
```
rwjblue: Should not be a dep, should be a devDep

Issue #6858, opened 2019-12-06
Add tests for deprecation-infra


runspired:
See https://github.com/emberjs/data/pull/6814#issuecomment-562317877

Gaurav0:
Can be closed.

PR #6881, opened 2019-12-10
[TEST] deprecation infra

Gaurav0:
For #6858 

Attempts to test whether deprecations are correctly being stripped based on compatWith passed in.


packages/unpublished-test-infra-addon/tests/unit/deprecations-stripped-test.js
```diff
@@ -0,0 +1,23 @@
+import { module, test } from 'ember-qunit';
+import config from 'dummy/config/environment';
+import DEPRECATE_EVENTED_API_USAGE from '@ember-data/private-build-infra/deprecations';
```
runspired: This should be a named import

PR #6889, opened 2019-12-12
[DEPRECATION] deprecate intimate Reference.internalModel

Gaurav0:
For #6166 

@runspired I was able to refactor record-reference.ts to use public api but stuck on the rest here.


packages/private-build-infra/addon/current-deprecations.ts
```diff
@@ -50,4 +50,5 @@ export default {
   DEPRECATE_MISMATCHED_INVERSE_RELATIONSHIP_DATA: '3.12',
   DEPRECATE_SERIALIZER_QUERY_RECORD_ARRAY_RESPONSE: '3.4',
   DEPRECATE_BELONGS_TO_REFERENCE_PUSH: '3.16',
+  DEPRECATE_REFERENCE_INTERNAL_MODEL: '3.16',
```
runspired, line 53:
```suggestion
  DEPRECATE_REFERENCE_INTERNAL_MODEL: '3.19',
```

packages/store/addon/-private/system/references/reference.ts
```diff
@@ -28,6 +32,8 @@ function isResourceIdentiferWithRelatedLinks(
   return value && value.links && value.links.related;
 }
 
+export const INTERNAL_MODELS = new WeakMap<RecordData, InternalModel>();
```
runspired, line 35: let's key this by reference instance instead of by RecordData, and lets do the same for `recordData` as that would nicely allow us to not leak that property either.

runspired: We can deprecate accessing `recordData` as part of the same deprecation

packages/store/addon/-private/system/references/reference.ts
```diff
@@ -201,4 +208,17 @@ if (FULL_LINKS_ON_RELATIONSHIPS) {
   };
 }
 
+if (DEPRECATE_REFERENCE_INTERNAL_MODEL) {
+  Object.defineProperty(Reference.prototype, 'internalModel', {
+    get() {
+      deprecate('Accessing the internalModel property of Reference is deprecated', false, {
+        id: 'ember-data:reference-internal-model',
+        until: '3.17',
```
runspired, line 216:
```suggestion
        until: '3.21',
```

packages/store/addon/-private/system/references/record.ts
```diff
@@ -123,9 +123,9 @@ export default class RecordReference extends Reference {
      @method value
      @return {Model} the record for this RecordReference
   */
-  value() {
-    if (this.internalModel.hasRecord) {
-      return this.internalModel.getRecord();
+  value(): RecordInstance | null {
+    if (this._id !== null) {
+      return this.store.peekRecord(this.type, this._id);
```
runspired, line 128: this doesn't quite work due to the situation when a record was client side created. We can pee k by identifier once we land #7104 but until then we should do the weak map lookup + getRecord

packages/store/addon/-private/system/references/record.ts
```diff
@@ -170,11 +170,9 @@ export default class RecordReference extends Reference {
      @return {Promise<record>} the record for this RecordReference
   */
   reload() {
-    let record = this.value();
-    if (record) {
-      return record.reload();
+    if (this._id !== null) {
+      return this.store.findRecord(this.type, this._id, { reload: true });
```
runspired, line 174: since you cant reload a record you've created client side this works well (unlike value above). Once #7104 lands though we should find by identifier.

packages/store/addon/-private/system/references/record.ts
```diff
@@ -170,11 +170,9 @@ export default class RecordReference extends Reference {
...
     }
-
-    return this.load();
+    throw new Error(`Unable to fetch record of type ${this.type} without an id`);
```
runspired, line 176: should we assert instead of throw to avoid shipping to production? I'm not sure. Value both ways. I'm tempted to create a specialized "strictModeAssert" that we usually strip but which users can opt to leave in for chasing down bugs.

packages/store/addon/-private/ts-interfaces/record-data.ts
```diff
@@ -13,6 +13,11 @@ export interface ChangedAttributesHash {
 }
 
 export interface RecordData {
+  id: string | null;
+  clientId: string | null;
+  modelName: string;
```
runspired, line 18: these properties and the getResourceIdentifier method aren't part of RecordData (only our own implementation). We can get the identifier from internal model for now for places we need it, later the constructor for references should receive the identifier directly.

PR #6927, opened 2019-12-20
[FIX] groupRecordsForFindMany should be optional

runspired:


igorT:
Can we add a test for this please

packages/store/addon/-private/system/core-store.ts
```diff
@@ -1360,7 +1360,12 @@ abstract class CoreStore extends Service {
         snapshots[i] = internalModels[i].createSnapshot(optionsMap.get(internalModel));
       }
 
-      let groups = adapter.groupRecordsForFindMany(this, snapshots);
+      let groups;
+      if (adapter.groupRecordsForFindMany) {
+        groups = adapter.groupRecordsForFindMany(this, snapshots);
+      } else {
+        groups = [snapshots];
```
igorT: I think this should be `groups = [[snapshots]];`

PR #6941, opened 2019-12-31
[BUGFIX] fix jsonapi error handling when not using jquery 

### LINKED ISSUE #6936: Adapter error while parsing JSONApi errors
# Thank you for opening this issue!

Hi ember-data team, 

Thanks a lot for the great word that you made to bring ember octane alive. It's so cool. 

I've got a strange error while using that new version of ember into a new project: I can't get the JSON API error parsing working correctly. Despite I'm generating valid error messages, ember is keeping generating adapter errors. 

### Reproduction

To reproduce the issue, I've created a mini-project, consisting of a very simple server, only producing JSONAPI error objects, and a simple frontend, with a single action. 

To use it: 

Clone my repository 
https://github.com/aclarembeau/ember-issue-demo/tree/master

Run the express server in back/ , run the ember project in front/. 

Then, open a browser, click on the link, this will open a debugger session. 
Dump the value of the "e" variable, you'll see that it's an AdapterError, resulting from an incorrect parsing of the JSONAPI error returned by the server. 

### Description

I've investigated a bit, and it seems that the issue comes from the following file: 
https://github.com/emberjs/data/blob/cf153484fd4077eea56ea877c4fac2acaa94fcee/packages/adapter/addon/rest.js

In the definition of the function 
```  handleResponse(status, headers, payload, requestData) { ```

You're doing ```return new InvalidError(payload.errors);```

However, after investigating a bit, it seems that "payload" is not a javascript object, but a string. So, if I add a JSON.parse instruction before that call, everything works well. 

I don't know exactly where should that JSON.parse call go exactly in the code, since, all from the beginning of the ajax respose, payload is always a string (and never a javascript object), so, I'm not feeling confident making a PR to solve this. So, if someone can have a look, this would be really great. 

### Versions

```
yarn list v1.15.2
warning Filtering by arguments is deprecated. Please use the pattern option instead.
└─ ember-source@3.15.0
✨  Done in 0.99s.
yarn list v1.15.2
warning Filtering by arguments is deprecated. Please use the pattern option instead.
└─ ember-cli@3.15.1
✨  Done in 0.79s.
yarn list v1.15.2
├─ @ember-data/adapter@3.15.0
├─ @ember-data/canary-features@3.15.0
├─ @ember-data/debug@3.15.0
├─ @ember-data/model@3.15.0
├─ @ember-data/private-build-infra@3.15.0
├─ @ember-data/record-data@3.15.0
├─ @ember-data/serializer@3.15.0
├─ @ember-data/store@3.15.0
└─ ember-data@3.15.0
```

### PR DESCRIPTION
**Pull request by aclarembeau**
Fixes #6936

### DIFF
--- a/packages/-ember-data/tests/integration/adapter/handle-response-test.js
+++ b/packages/-ember-data/tests/integration/adapter/handle-response-test.js
@@ -95,7 +95,7 @@ module('integration/adapter/handle-response', function(hooks) {
     assert.equal(handleResponseCalled, 1, 'handle response is called');
   });
 
-  test('handleResponse is called on empty string repsonse', async function(assert) {
+  test('handleResponse is called on empty string response', async function(assert) {
     let handleResponseCalled = 0;
 
     this.server.get('/people', function() {
@@ -122,7 +122,7 @@ module('integration/adapter/handle-response', function(hooks) {
     assert.equal(handleResponseCalled, 1, 'handle response is called');
   });
 
-  test('handleResponse is not called on invalid repsonse', async function(assert) {
+  test('handleResponse is not called on invalid response', async function(assert) {
     let handleResponseCalled = 0;
 
     this.server.get('/people', function() {
@@ -149,7 +149,7 @@ module('integration/adapter/handle-response', function(hooks) {
     assert.equal(handleResponseCalled, 0, 'handle response is not called');
   });
 
-  test('handleResponse is called on empty string repsonse with 400 status', async function(assert) {
+  test('handleResponse is called on empty string response with 400 status', async function(assert) {
     let handleResponseCalled = 0;
 
     this.server.get('/people', function() {
@@ -176,10 +176,10 @@ module('integration/adapter/handle-response', function(hooks) {
     assert.equal(handleResponseCalled, 1, 'handle response is called');
   });
 
-  test('handleResponse is called with correct parameters on string repsonse with 422 status', async function(assert) {
+  test('handleResponse is called with correct parameters on string response with 422 status', async function(assert) {
     let handleResponseCalled = 0;
 
-    let errorObject = {errors: {}};
+    let errorObject = { errors: {} };
 
     this.server.get('/people', function() {
       return [422, { 'Content-Type': 'application/json' }, JSON.stringify(errorObject)];
@@ -188,7 +188,7 @@ module('integration/adapter/handle-response', function(hooks) {
     class TestAdapter extends JSONAPIAdapter {
       handleResponse(status, headers, payload, requestData) {
         handleResponseCalled++;
-        assert.deepEqual(payload, errorObject);
+        assert.deepEqual(payload, errorObject, 'payload from handleResponse matches expected error');
 
         return super.handleResponse(status, headers, payload, requestData);
       }
@@ -206,4 +206,3 @@ module('integration/adapter/handle-response', function(hooks) {
     assert.equal(handleResponseCalled, 1, 'handle response is called');
   });
 });
-
--- a/packages/-ember-data/tests/integration/adapter/handle-response-test.js
+++ b/packages/-ember-data/tests/integration/adapter/handle-response-test.js
@@ -175,4 +175,35 @@ module('integration/adapter/handle-response', function(hooks) {
 
     assert.equal(handleResponseCalled, 1, 'handle response is called');
   });
+
+  test('handleResponse is called with correct parameters on string repsonse with 422 status', async function(assert) {
+    let handleResponseCalled = 0;
+
+    let errorObject = {errors: {}};
+
+    this.server.get('/people', function() {
+      return [422, { 'Content-Type': 'application/json' }, JSON.stringify(errorObject)];
+    });
+
+    class TestAdapter extends JSONAPIAdapter {
+      handleResponse(status, headers, payload, requestData) {
+        handleResponseCalled++;
+        assert.deepEqual(payload, errorObject);
+
+        return super.handleResponse(status, headers, payload, requestData);
+      }
+    }
+
+    this.owner.register('adapter:application', TestAdapter);
+
+    try {
+      await this.store.findAll('person');
+      assert.ok(false, 'promise should reject');
+    } catch {
+      assert.ok(true, 'promise rejected');
+    }
+
+    assert.equal(handleResponseCalled, 1, 'handle response is called');
+  });
 });
+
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: packages/-ember-data/tests/integration/adapter/handle-response-test.js [hunk 1]
```diff
@@ -175,4 +175,35 @@ module('integration/adapter/handle-response', function(hooks) {
...
+    let errorObject = {errors: {}};
+
+    this.server.get('/people', function() {
+      return [422, { 'Content-Type': 'application/json' }, JSON.stringify(errorObject)];
+    });
+
+    class TestAdapter extends JSONAPIAdapter {
+      handleResponse(status, headers, payload, requestData) {
+        handleResponseCalled++;
+        assert.deepEqual(payload, errorObject);
```
snewcomer: Could add a failure message - 

`assert.deepEqual(payload, errorObject, 'payload from handleResponse matches expected error');`

packages/-ember-data/tests/integration/adapter/handle-response-test.js
```diff
@@ -175,4 +175,35 @@ module('integration/adapter/handle-response', function(hooks) {
 
     assert.equal(handleResponseCalled, 1, 'handle response is called');
   });
+
+  test('handleResponse is called with correct parameters on string repsonse with 422 status', async function(assert) {
```
snewcomer: `assert.expect(3);`

packages/adapter/addon/rest.js
```diff
@@ -1301,9 +1301,11 @@ function fetchSuccessHandler(adapter, payload, response, requestData) {
 
 function fetchErrorHandler(adapter, payload, response, errorThrown, requestData) {
   let responseData = fetchResponseData(response);
+  payload = adapter.parseErrorResponse(payload);
+
   if (responseData.status === 200 && payload instanceof Error) {
     responseData.errorThrown = payload;
-    payload = responseData.errorThrown.payload;
+    payload = responseData.errorThrown;
```
snewcomer: Do you think we need this reassignment to `payload`?

packages/adapter/addon/rest.js
```diff
@@ -1301,9 +1301,11 @@ function fetchSuccessHandler(adapter, payload, response, requestData) {
 
 function fetchErrorHandler(adapter, payload, response, errorThrown, requestData) {
   let responseData = fetchResponseData(response);
+  payload = adapter.parseErrorResponse(payload);
+
   if (responseData.status === 200 && payload instanceof Error) {
```
snewcomer: It is interesting the error you throw in your test is 422, but the check here is 200.  I thought there was a reason for this but I can't remember why.  Do you happen to know?

packages/-ember-data/tests/integration/adapter/handle-response-test.js
snewcomer: r/repsonse/response

packages/adapter/addon/rest.js
aclarembeau: Indeed, that was an error on my side. 
I shouldn't even change anything in that branch. 
My fix was to handle the case where payload is a javascript string representing a JSONAPI error object. In that branch, payload is an Error object, so, I shouldn't change anything here.

packages/-ember-data/tests/integration/adapter/handle-response-test.js
aclarembeau: I don't get the assert.expect point, could you tell me more?

packages/-ember-data/tests/integration/adapter/handle-response-test.js
aclarembeau: Done :)

packages/adapter/addon/rest.js
aclarembeau: After re-reding the code, it seems that there could be some case where the HTTP status is 200, but an error is thrown (for example due to a runtime error). In that case, we fall into the status=200 branch, so we don't need to parse payload there.

aclarembeau:
@snewcomer , I've updated my code with: 
- better respect of linting rules 
- updated logic (not parsing JSON payload if payload is not JSON, but an error object)

packages/adapter/addon/rest.js
```diff
@@ -1301,11 +1301,13 @@ function fetchSuccessHandler(adapter, payload, response, requestData) {
 
 function fetchErrorHandler(adapter, payload, response, errorThrown, requestData) {
   let responseData = fetchResponseData(response);
+
```
efx: @aclarembeau total nitpick, but can you remove this whitespace?

efx:
The asset size checks are failing by 28 bytes; not sure if that amount should hold up this PR.
@runspired thoughts?

packages/-ember-data/tests/integration/adapter/handle-response-test.js
snewcomer: Yeah!  So because you have assertions inside of stubbed methods that don't have to be called for the test to pass, by adding an `assert.expect`, we ensure that is is called.

packages/adapter/addon/rest.js
aclarembeau: Sure, it's done :)

runspired:
Asset size change is totally fine. I agree there is a bug and that the solution is to JSON.parse something.

What I am not convinced of is that this is the right place in the process to do this. To be convinced I need to spend a little time (or Igor needs to spend a little time) walking through the fetch flow and confirming where the point of divergence in behavior occurs so that we are confident we're fixing it at the source and not much later in the flow.

packages/adapter/addon/rest.js
```diff
@@ -1301,11 +1301,13 @@ function fetchSuccessHandler(adapter, payload, response, requestData) {
...
   if (responseData.status === 200 && payload instanceof Error) {
     responseData.errorThrown = payload;
     payload = responseData.errorThrown.payload;
   } else {
     responseData.errorThrown = errorThrown;
+    payload = adapter.parseErrorResponse(payload);
```
runspired: A few comments after reading through the existing implementation:

I believe the necessary change is either this OR

```
  if (!response.ok) {
    return payload;
  }
```

^ this line in determineBodyPromise should possibly attempt JSON.parse. **I lean heavily towards the former (current PR) solution**

The primary difference is in whether we give the adapter the chance to do the parsing. Note the Ajax codepath calls this for all Ajax errors (I'm not sure if a 404 triggers an error though, as odd as that sounds at first glance, it doesn't trigger one in fetch for instance).

But if it does...

```
let payload = adapter.parseErrorResponse(jqXHR.responseText);
```

^ this is what would be doing the similar parsing for the Ajax branch.

cc @igorT

PR #6949, opened 2020-01-06
improve performance app documentation

efx:
I removed `php` server recommendation since we can use ember-cli's
server.

Is there a specific reason we pin tracerbench to 2.3?

/cc @skaterdav85 @runspired


efx:
For my curiosity as well, why are we using port 500? I assume other parts of the perf script rely upon that? If it isn't a hard requirement I would update the ember serve command to find an available port automatically.

runspired:
RE tracerbench version I'm not sure there is a reason to pin. RE port not sure there's a reason for that either. Feel free to update both if it simplifies things.

PR #6958, opened 2020-01-10
[DOC] Update embedded-records-mixin.js

allthesignals:
Adding documentation on the embedded records mixin explaining that it does not work with JSONAPISerializer. This should be noted more prominently rather than in a passive warning message (which might get buried under a sea of deprecation warnings in the test suite).

### DIFF
--- a/packages/serializer/addon/-private/embedded-records-mixin.js
+++ b/packages/serializer/addon/-private/embedded-records-mixin.js
@@ -19,6 +19,8 @@ import { typeOf } from '@ember/utils';
 
   Note that embedded records will serialize with the serializer for their model instead of the serializer in which they are defined.
 
+  Note also that this mixin does not work with JSONAPISerializer because the JSON:API specification does not describe how to format embedded resources.
+
   Below is an example of a per-type serializer (`post` type).
 
   ```app/serializers/post.js


packages/serializer/addon/-private/embedded-records-mixin.js
```diff
@@ -19,6 +19,8 @@ import { warn } from '@ember/debug';
 
   Note that embedded records will serialize with the serializer for their model instead of the serializer in which they are defined.
 
+  Note also that this mixin does not work with JSONAPISerializer because the JSON API spec does not describe how to format embedded resources.
```
efx: General question I do not know the answer to: what versions of the JSON:API spec does ember-data adhere to? I would like to qualify this caveat with that answer.

allthesignals: image

efx: Thanks! Upon further reflection I think ember-data should describe its compatibility with JSON:API somewhere for transparency. It seems better suited to top-level documentation or some kind of introduction. So we can leave it out of this PR.

runspired: > Upon further reflection I think ember-data should describe its compatibility with JSON:API somewhere for transparency. It seems better suited to top-level documentation or some kind of introduction. 

We do in the serializer documentation overview

efx:
@runspired any objections to merging? Also, I think we can add the changelog:doc label.

hjdivad:
rebased which should get CI green

MelSumner:
@hjdivad looks like a few ember data tests are still failing- any insight there?

allthesignals:
Okay, I went ahead and force-pushed a change with a fresh branch.

I'm not sure why but simply rebasing this branch triggered a few merge conflicts with unrelated files. Any theories on why that might've been happening would be helpful for me in the future :)

hjdivad:
<3 Thanks @allthesignals !

hjdivad:
> @hjdivad looks like a few ember data tests are still failing- any insight there?

Those were failing external partner tests.

In particular [travis-web: Acceptance | repo/trigger build: triggering a custom build via the dropdown
](https://dev.azure.com/ember-data/emberjs/_build/results?buildId=3368&view=logs&j=0f730510-e886-53d0-e9ee-3135be6cf1a2&t=f73f74e7-ab42-5087-af50-a5b3a5b2b9a4&l=2610).  @runspired @igorT is this a test with a race condition or something?  I recall seeing this crop up a lot in false negatives.

PR #6966, opened 2020-01-14
turn on FULL_LINKS_ON_RELATIONSHIPS

runspired:


igorT:
Did we discuss this in the ED meeting?

runspired:
@igorT we discussed a long time ago but then it never got done.

PR #6968, opened 2020-01-14
remove IDENTIFIERS off branches

runspired:


packages/canary-features/addon/default-features.ts
```diff
@@ -173,7 +173,6 @@ export default {
   SAMPLE_FEATURE_FLAG: null,
   RECORD_DATA_ERRORS: null,
   RECORD_DATA_STATE: null,
-  IDENTIFIERS: true,
```
igorT: I think the practice is to keep this right? For example, removing it would break downstream code that is checking whether it is enabled

packages/canary-features/addon/index.ts
```diff
@@ -33,6 +33,5 @@ export const SAMPLE_FEATURE_FLAG = featureValue(FEATURES.SAMPLE_FEATURE_FLAG);
 export const RECORD_DATA_ERRORS = featureValue(FEATURES.RECORD_DATA_ERRORS);
 export const RECORD_DATA_STATE = featureValue(FEATURES.RECORD_DATA_STATE);
 export const REQUEST_SERVICE = featureValue(FEATURES.REQUEST_SERVICE);
```
igorT: same here

packages/store/addon/-private/system/model/internal-model.ts
```diff
@@ -208,22 +207,15 @@ export default class InternalModel {
   }
 
   get id(): string | null {
-    if (IDENTIFIERS) {
-      return this.identifier.id; // || this._id;
-    }
-    return this._id;
+    return this.identifier.id; // || this._id;
```
igorT: should we remove the comment here?

packages/unpublished-adapter-encapsulation-test-app/tests/integration/belongs-to-test.js
```diff
@@ -144,8 +144,8 @@ module('integration/belongs-to - Belongs To Tests', function(hooks) {
     let store = owner.lookup('service:store');
 
     // This code is a workaround for issue https://github.com/emberjs/data/issues/6758
-    // expectedResult is unexpectedly mutated during store.findRecord
-    // if IDENTIFIERS is turned on
+    // expectedResult is mutated during store.findRecord
```
igorT: is it not unexpected?

packages/unpublished-adapter-encapsulation-test-app/tests/integration/belongs-to-test.js
```diff
@@ -260,8 +260,8 @@ module('integration/belongs-to - Belongs To Tests', function(hooks) {
...
```
igorT: Great job on catching the comments

packages/unpublished-adapter-encapsulation-test-app/tests/integration/belongs-to-test.js
runspired: We do expect this yes

packages/store/addon/-private/system/model/internal-model.ts
runspired: it’s deadweight so likely

packages/canary-features/addon/default-features.ts
runspired: Is that the practice? I’d think downstream would treat it as undefined

igorT: In which case it would revert to the pre identifiers behavior, no?

igorT: If wave if(Identifiers) in a m3 version, it ought to continue to work, no? Don't think there's a large cost in keeping. Maybe @rwjblue had thoughts

runspired: Seems like a fine thing just unfamiliar with what we usually do here. Happy to keep if that’s best.

sandstrom:
Awesome! 🚀 

rocket-landing

Issue #6961, opened 2020-01-12
Build failing with ember beta & canary


igorT:
https://github.com/emberjs/data/runs/384084391

@pzuraq is poking at it

hjdivad:
Not a release blocker but nice to get in before we cut a release (est. this evening)

efx:
@igorT I believe https://github.com/emberjs/data/pull/7030 closed this.

igorT:
@efx thank you!

PR #6985, opened 2020-01-21
[FIX] RecordData is not a constructor

runspired:
Fixes issue spotted by partner tests for Storefront/Observer/Travis-web
This fix will need to be back ported alongside the fix for `RecordData not implemented` to 3.16 beta and 3.15 release

Also fixes (separate commit) lint issue that crept in and a lint perf issue where we would attempt lint the dists created for the performance checks.

### DIFF
--- a/packages/store/addon/-private/system/core-store.ts
+++ b/packages/store/addon/-private/system/core-store.ts
@@ -29,6 +29,7 @@ import {
   HAS_ADAPTER_PACKAGE,
   HAS_EMBER_DATA_PACKAGE,
   HAS_MODEL_PACKAGE,
+  HAS_RECORD_DATA_PACKAGE,
   HAS_SERIALIZER_PACKAGE,
 } from '@ember-data/private-build-infra';
 import {
@@ -94,7 +95,9 @@ type AttributesSchema = import('../ts-interfaces/record-data-schemas').Attribute
 type SchemaDefinitionService = import('../ts-interfaces/schema-definition-service').SchemaDefinitionService;
 type PrivateSnapshot = import('./snapshot').PrivateSnapshot;
 type Relationship = import('@ember-data/record-data/-private').Relationship;
+type RecordDataClass = typeof import('@ember-data/record-data/-private').RecordData;
 
+let _RecordData: RecordDataClass | undefined;
 const emberRun = emberRunLoop.backburner;
 
 const { ENV } = Ember;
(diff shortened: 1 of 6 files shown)


packages/store/addon/-private/system/ds-model-store.ts
```diff
@@ -32,6 +30,18 @@ type NotificationManager = import('./record-notification-manager').default;
 type DSModel = import('../ts-interfaces/ds-model').DSModel;
 type ShimModelClass = import('./model/shim-model-class').default;
 type DSModelClass = import('@ember-data/model').default;
+type RecordDataClass = typeof import('@ember-data/record-data/-private').RecordData;
+
+// we can't greedily use require as this causes
+// an order of operations issue which gives us undefined
```
rwjblue: This comment seems to be describing a cycle, can you update the comment to make it clear where the actual cycle is (something like `foo/bar -> baz/qux -> foo/bar`)

packages/store/addon/-private/system/ds-model-store.ts
```diff
@@ -32,6 +30,18 @@ type NotificationManager = import('./record-notification-manager').default;
...
+let _RecordData: RecordDataClass;
```
rwjblue: FWIW, I think the types are not quite correct here, I would expect:

```suggestion
let _RecordData: undefined | RecordDataClass;
```

packages/store/addon/-private/system/ds-model-store.ts
```diff
@@ -32,6 +30,18 @@ type NotificationManager = import('./record-notification-manager').default;
...
+function getRecordData(): RecordDataClass {
+  if (_RecordData !== undefined) {
+    return _RecordData;
+  }
+  _RecordData = HAS_RECORD_DATA_PACKAGE ? require('@ember-data/record-data/-private').RecordData : null;
```
rwjblue: Can we move this lazy require directly into `createRecordDataFor`? It would not need the ternary (because that branch is only used when `HAS_RECORD_DATA_PACKAGE` is present)...

runspired: @rwjblue what is the perf of repeated calls into require? similar cache to our manual one here?

rwjblue: ya, I meant more like:

```ts
let _RecordData: undefined | RecordDataClass;

// ...snip...
createRecordDataFor(modelName: string, id: string | null, clientId: string, storeWrapper: RecordDataStoreWrapper) {
  if (HAS_RECORD_DATA_PACKAGE) {
    if (_RecordData === undefined) {
      _RecordData = require('whatever/path');
    }
   
  }
}
```

packages/store/addon/-private/system/core-store.ts
```diff
@@ -3111,7 +3116,26 @@ abstract class CoreStore extends Service {
     clientId: string,
     storeWrapper: RecordDataStoreWrapper
   ): RecordData {
-    throw new Error(`Expected store.createRecordDataFor to be implemented but it wasn't`);
+    if (HAS_RECORD_DATA_PACKAGE) {
+      // we can't greedily use require as this causes
+      // a cycle we can't easily fix (or clearly pin point) at present.
```
hjdivad:
```suggestion
      // a cycle we can't easily fix (or clearly pin point) at present.
      //
      // it can be reproduced in partner tests by running
      // node ./bin/packages-for-commit.js && cd ../__external-test-cache/ember-observer && ember test
```

hjdivad:
Tests red due to #6961 canary issues that don't impede release but otherwise 🍏

PR #6986, opened 2020-01-22
[BUFGIX BETA] dont run canary model tests in beta

runspired:

### DIFF
--- a/packages/unpublished-model-encapsulation-test-app/package.json
+++ b/packages/unpublished-model-encapsulation-test-app/package.json
@@ -18,6 +18,7 @@
     "test": "ember test --test-port=0"
   },
   "dependencies": {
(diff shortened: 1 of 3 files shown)


packages/unpublished-model-encapsulation-test-app/package.json
```diff
@@ -18,6 +18,7 @@
     "test": "ember test --test-port=0"
   },
   "dependencies": {
+    "@ember-data/private-build-infra": "3.16.0-beta.0",
```
igorT: Do we need `@ember-data/canary-features`

hjdivad: followed up with @runspired; this is a transitive dep so we're good to go for here

PR #6987, opened 2020-01-22
[BUFGIX BETA] dont run feature flag model tests in beta

runspired:
Master version of #6986

### DIFF
--- a/packages/unpublished-model-encapsulation-test-app/ember-cli-build.js
+++ b/packages/unpublished-model-encapsulation-test-app/ember-cli-build.js
@@ -5,8 +5,21 @@
 const EmberApp = require('ember-cli/lib/broccoli/ember-app');
 
 module.exports = function(defaults) {
+  const isProd = process.env.EMBER_ENV === 'production';
+  // allows testing with env config for stripping all deprecations
+  const compatWith = process.env.EMBER_DATA_FULL_COMPAT ? '99.0' : null;
+  const plugins = [...require('@ember-data/private-build-infra/src/debug-macros')(null, isProd, compatWith)];
+
   let app = new EmberApp(defaults, {
+    emberData: {
+      compatWith,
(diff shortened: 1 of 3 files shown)


packages/unpublished-model-encapsulation-test-app/ember-cli-build.js
```diff
@@ -5,8 +5,20 @@
 const EmberApp = require('ember-cli/lib/broccoli/ember-app');
 
 module.exports = function(defaults) {
+  const isProd = process.env.EMBER_ENV === 'production';
+  const compatWith = process.env.EMBER_DATA_FULL_COMPAT ? '99.0' : null;
```
igorT: Whats this about? Can you please add a comment

packages/unpublished-model-encapsulation-test-app/package.json
```diff
@@ -18,6 +18,7 @@
     "test": "ember test --test-port=0"
   },
   "dependencies": {
+    "@ember-data/private-build-infra": "3.17.0-alpha.1",
```
igorT: Do you need to add `ember-data/canary-features` in devDependencies at least?

igorT:
A better title might be > dont run feature flagged tests in beta

packages/unpublished-model-encapsulation-test-app/ember-cli-build.js
runspired: this ensures deprecation stripping when desired as well, matching the config and flag used by the tests in -ember-data package

packages/unpublished-model-encapsulation-test-app/package.json
runspired: we don't have to, as it is a transitive dep of all of these other things, but making it explicit can't hurt

PR #6989, opened 2020-01-22
support other node versions

### LINKED ISSUE #6988: Make sure we test node 10 & node 12 and 13

efx:
fix #6988


efx:
This should actually be done in azure / actions.

PR #7002, opened 2020-01-24
3.15.x release backports

hjdivad:
backport #6960 #6985


packages/-ember-data/addon/index.js
```diff
@@ -60,6 +58,15 @@ import JSONSerializer from '@ember-data/serializer/json';
 import RESTSerializer, { EmbeddedRecordsMixin } from '@ember-data/serializer/rest';
 
 import Model, { attr, belongsTo, hasMany } from '@ember-data/model';
+import Transform from '@ember-data/serializer/transform';
+
+if (VERSION.match(/^1\.([0-9]|1[0-2])\./)) {
```
runspired: Odd that this snuck back in.

packages/unpublished-relationship-performance-test-app/app/routes/application.js
```diff
@@ -19,4 +19,22 @@ export default Route.extend({
       return flattened;
     });
   },
+  afterModel() {
```
runspired: surprising we picked this up but seems fine it makes the perf check more backwards compatible

runspired:
The LTS failure here is due to ember-beta (because the current bug would affect all known versions of ember-data)

packages/unpublished-relationship-performance-test-app/app/routes/application.js
hjdivad: https://github.com/emberjs/data/pull/6985/files#diff-3f38ec8323b0ede1b00d2042d47df6f3

packages/-ember-data/addon/index.js
hjdivad: killed it

PR #7005, opened 2020-01-24
Run partner tests on release

hjdivad:

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -264,9 +264,9 @@ jobs:
     needs: [additional-scenarios, basic-tests, floating-dependencies, lint, lts, releases]
     if: |
       github.event_name == 'pull_request' && (
-        github.base_ref == 'master' || github.base_ref == 'beta'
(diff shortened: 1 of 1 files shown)


igorT:
We run the release branch tests usually after doing point releases. At that point, the partner app might be using newer features, so might fail the tests even if the release branch works. Thats why the beta branch tests are much more useful for new releases.

hjdivad:
> We run the release branch tests usually after doing point releases.

Do you mean manually?

The motivation here is to get partner tests running for PRs that backport fixes before making a point release.

igorT:
@hjdivad I understand the motivation, but the problem is that if the point release is after there has been a newer release the partner app might not be passing anymore

hjdivad:
> @hjdivad I understand the motivation, but the problem is that if the point release is after there has been a newer release the partner app might not be passing anymore

ok.  I'll leave the PR open for now and try to find some time to include a change to the partner test script that skips partners on a newer release than the one being tested.

PR #7019, opened 2020-01-30
[DOC] create a howto for testing pull requests

efx:
Briefly document how to link ember data in your application. Came up as a result of someone wanting to helpfully test an outstanding PR.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -14,7 +14,10 @@ a great place to ask for help!
 
 #### Testing ember data source directly
 
-You can use package linking to test checkouts of ember-data.
+##### monolithic ember-data
+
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -12,6 +12,29 @@ We have provided an [issue template](.github/bug.md) what will help guide you th
 If you are unsure if something is a bug, the `#ember-data` channel on [Discord](https://discord.gg/zT3asNS) is
 a great place to ask for help!
 
+#### Testing ember data source directly
```
efx: this is for depending on monolithic ember-data but not individual packages. It does solve the transitive dependency problem.

efx: updates before merging
- generalize the language for individual packages
- yarn + npm do not mix so we need to caveat and describe this better (see ember.js discussion)

MelSumner:
@efx it looks like there are a couple of failing tests. Can you look to figure out why?

efx:
@MelSumner certainly! I'm swamped with an internal release so will need to take a look at this after 2/3. I plan to revise the verbiage per Rob's suggestions and will dig into whether the CI failures are legitimate.

PR #7020, opened 2020-01-30
Ensure adapters and serializers are destroyed upon store destruction.

### LINKED ISSUE #7018: Adapter instances should be destroyed
It appears that `adapter` instances are never being destroyed. 

Related codepaths:

We configure the `adapter` type to be non-singleton here:

https://github.com/emberjs/data/blob/005ff4ddce35d42ddcfa4148a2ced0b5184a5110/packages/-ember-data/addon/setup-container.js#L37

And then we instantiate the `adapter` instances with `owner.lookup` here:

https://github.com/emberjs/data/blob/005ff4ddce35d42ddcfa4148a2ced0b5184a5110/packages/store/addon/-private/system/core-store.ts#L3254

But in the store's `willDestroy` we do nothing with this._adapterCache (we should loop over them and be calling `.destroy()`) 

https://github.com/emberjs/data/blob/005ff4ddce35d42ddcfa4148a2ced0b5184a5110/packages/store/addon/-private/system/core-store.ts#L3490

---

This is _generally_ not an issue for most users, as they are not commonly doing their own async from within the `adapter` instances. However, as soon as you attempt to do things like tracking response info (by queueing and firing a request) or doing exponential backoff you might find yourself in a situation where you need to clean up a timer (`clearTimeout`, `runDestroyables`, etc). Since we never run `adapterInstance.destroy()`, the user can't implement `willDestroy` and have that cleanup code be ran (ember-concurrency and ember-lifeline both leverage `.destroy()` being called in order to properly cleanup).

rwjblue:
Prior to this change, adapters and serializers were never destroyed.  Commonly (and anecdotally) this is "fine" (heck this is the first report of the issue in years!), but it is pretty common for adapters to do async of their own (e.g. tracking adapter responses, exponential backoffs, etc) and without `.destroy()` being called on them there is no way for those adapters/serializer to ensure that async is canceled appropriately.

Fixes #7018

### DIFF
--- a/packages/store/addon/-private/ts-interfaces/minimum-adapter-interface.ts
+++ b/packages/store/addon/-private/ts-interfaces/minimum-adapter-interface.ts
@@ -699,6 +699,18 @@ interface Adapter {
    * @return {boolean} true if the a new request for all records of the type in SnapshotRecordArray should be made in the background, false otherwise
    */
   shouldBackgroundReloadAll?(store: Store, snapshotArray: SnapshotRecordArray): boolean;
+
+  /**
+   * In some situations the adapter may need to perform cleanup when destroyed,
+   * that cleanup can be done in `destroy`.
+   *
+   * If not implemented, the store does not inform the adapter of destruction.
+   *
+   * @method destroy [OPTIONAL]
+   * @public
+   * @optional
+   */
+  destroy?(): void;
 }
 
 export default Adapter;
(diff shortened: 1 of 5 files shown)


packages/store/addon/-private/ts-interfaces/minimum-adapter-interface.ts
```diff
@@ -699,6 +699,18 @@ interface Adapter {
...
+   * In some situations the adapter may need to perform cleanup when destroyed,
+   * that cleanup can be done in `destroy`.
+   *
+   * If not implemented, the store does not inform the adapter of destruction.
+   *
+   * @method destroy [OPTIONAL]
+   * @public
+   * @optional
+   */
+  destroy?(): void;
```
runspired: Thank you! <3

kiwiupover:
@rwjblue  can we back port this fix to at least 3.11?

igorT:
@kiwiupover 3.11 isn't an lts, would 3.12 work for you?

kiwiupover:
@igorT because of this bug https://github.com/emberjs/data/issues/6867

PR #7067, opened 2020-02-27
Add @canary-features as -build-infra dep

hjdivad:
This is already a de-facto dep as it's required by `features.js`.
Adding it as an explicit dep to deal with dependency issues.


hjdivad:
@igorT going to merge this; LMK when you have a canary release out.

The partner test claims it failed `ember-m3`, but running the partner test locally doesn't regress.  Either way I'll be dropping the highlander checking in https://github.com/hjdivad/ember-m3/pull/567 soon so I don't think it makes too much difference.

PR #6207, opened 2019-06-28
[FEAT request-state-service] implements RFC 466

igorT:
[RFC text](https://github.com/emberjs/rfcs/pull/466)

PR #7078, opened 2020-03-06
Backport #6738 & #7067 for LTS

HeroicEric:
Backports the following PRs to the current LTS branch (lts-3-12):

- [BUGFIX] only allow feature flag alterations in canary #6738 
- Add @canary-features as -build-infra dep #7067

### DIFF
--- a/packages/-build-infra/package.json
+++ b/packages/-build-infra/package.json
@@ -15,6 +15,7 @@
   },
   "dependencies": {
     "@babel/plugin-transform-block-scoping": "^7.5.5",
+    "@ember-data/canary-features": "3.12.5",
     "babel-plugin-debug-macros": "^0.3.2",
     "babel-plugin-feature-flags": "^0.3.1",
     "babel-plugin-filter-imports": "^3.0.0",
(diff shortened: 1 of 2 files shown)


packages/store/addon/-private/system/store.ts
```diff
@@ -2920,14 +2920,30 @@ const Store = Service.extend({
...
+
+    return this._super();
+  },
+
   willDestroy() {
     this._super(...arguments);
     this._pushedInternalModels = null;
     this.recordArrayManager.destroy();
 
-    this._adapterCache = null;
```
HeroicEric: @igorT these lines had already been removed by the time the original changes were made in https://github.com/emberjs/data/pull/7020 I had to also remove them to backport this change in order for the `destroy()` logic above to work. I'm not totally sure what, if any, implications this might have elsewhere. I found a comment related to this https://github.com/emberjs/data/pull/6207/files#r306956786 so I was thinking you might have more context

igorT:
Does this need https://github.com/emberjs/data/pull/7067 to pass CI?

PR #7079, opened 2020-03-06
Fix changelog for v3.16.2

HeroicEric:
This fixes the changelog by moving the changelog entries originally
intended for 3.16.1 to the section for 3.16.2.

For some more context, due to an unfortunate mishap by yours truly,
3.16.1 was never actually published to NPM. Instead, 3.16.2 was
published with what should have been 3.16.1. NPM does not support any
way to remedy this after the fact (and with good reason), but it would
still be nice for the changelog to be correct for anyone viewing it via
GitHub.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,9 +1,13 @@
 # Ember Data Changelog
 
-## Release 3.16.1 (February 27 2020)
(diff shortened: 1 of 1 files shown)


igorT:
We should PR this to master as well

PR #7080, opened 2020-03-06
[BUGFIX] Fix #7015 - calling changedAttributes for non instantiated r…

### LINKED ISSUE #7015: Snapshot.changedAttributes breaks for record-less snapshots in 3.14+
We (smartly) avoided allocating anything for changed attributes for the case where no record exists for the snapshot for perf savings; however, this has an unintended side-effect for users whom are serializing related records which potentially have not been materialized.

For these users, calling `snapshot.changedAttributes` will error with the following:

```
Uncaught (in promise) TypeError: Cannot convert undefined or null to object
    at Function.keys (<anonymous>)
    at Snapshot.changedAttributes (vendor.js:120369)
```

An easy fix for this would be for us to early exit with an empty object when we are in this case.

igorT:
…elationships

### DIFF
--- a/packages/-ember-data/tests/integration/snapshot-test.js
+++ b/packages/-ember-data/tests/integration/snapshot-test.js
@@ -373,6 +373,42 @@ module('integration/snapshot - Snapshot', function(hooks) {
     assert.equal(relationship.attr('title'), 'Hello World', 'post title is correct');
   });
 
+  test('snapshot.belongsTo().changedAttributes() returns an empty object if belongsTo record in not instantiated #7015', function(assert) {
+    assert.expect(2);
+
+    store.push({
+      data: [
+        {
+          type: 'comment',
+          id: '2',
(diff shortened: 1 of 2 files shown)


packages/-ember-data/tests/integration/snapshot-test.js
```diff
@@ -373,6 +373,42 @@ module('integration/snapshot - Snapshot', function(hooks) {
     assert.equal(relationship.attr('title'), 'Hello World', 'post title is correct');
   });
 
+  test('snapshot.belongsTo().changedAttributes() returns an empty object if belongsTo record in not instantiated #7015', function(assert) {
```
HeroicEric:
```suggestion
  test('snapshot.belongsTo().changedAttributes() returns an empty object if belongsTo record is not instantiated #7015', function(assert) {
```

PR #7081, opened 2020-03-06
Backport #7080 for LTS

### LINKED ISSUE #7015: Snapshot.changedAttributes breaks for record-less snapshots in 3.14+

HeroicEric:
Backports [BUGFIX] Fix #7015 - calling changedAttributes for non instantiated r… #7080 for LTS 3.12

### DIFF
--- a/packages/-ember-data/tests/integration/snapshot-test.js
+++ b/packages/-ember-data/tests/integration/snapshot-test.js
@@ -387,6 +387,42 @@ module('integration/snapshot - Snapshot', function(hooks) {
     });
   });
 
+  test('snapshot.belongsTo().changedAttributes() returns an empty object if belongsTo record in not instantiated #7015', function(assert) {
+    assert.expect(2);
+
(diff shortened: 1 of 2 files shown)


igorT:
Do we need #7080 on lts? The issue was reported as having started in 3.13. Unless we backported the problem to 3.12

HeroicEric:
not needed here. Issue didn't start until 3.13

PR #7085, opened 2020-03-06
Backport #7020 for LTS

HeroicEric:
Backports Ensure adapters and serializers are destroyed upon store destruction. #7020 for LTS 3.12

This required a few changes:

- The [changes to packages/-ember-data/node-tests/fixtures/expected.js](https://github.com/emberjs/data/pull/7020/files#diff-90207ce76d6fc168c717eecca7fd82a5R148-R154) were not included because that test file does not exist in the branch
- the documentation for `Adapter#destroy()` was added to the base `Adapter` class rather than the `MinimumAdapterInterface` because that interface did not yet exist

### DIFF
--- a/packages/store/addon/-private/system/store.ts
+++ b/packages/store/addon/-private/system/store.ts
@@ -2920,13 +2920,29 @@ const Store = Service.extend({
     return serializer;
   },
 
+  destroy() {
+    // enqueue destruction of any adapters/serializers we have created
+    for (let adapterName in this._adapterCache) {
+      let adapter = this._adapterCache[adapterName];
+      if (typeof adapter.destroy === 'function') {
+        adapter.destroy();
+      }
+    }
+
+    for (let serializerName in this._serializerCache) {
+      let serializer = this._serializerCache[serializerName];
+      if (typeof serializer.destroy === 'function') {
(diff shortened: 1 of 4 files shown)


packages/store/addon/-private/system/store.ts
```diff
@@ -2920,14 +2920,30 @@ const Store = Service.extend({
...
+    return this._super();
+  },
+
   willDestroy() {
     this._super(...arguments);
     this._pushedInternalModels = null;
     this.recordArrayManager.destroy();
 
-    this._adapterCache = null;
-    this._serializerCache = null;
```
HeroicEric: @igorT these lines had already been removed by the time the original changes were made in https://github.com/emberjs/data/pull/7020 I had to also remove them to backport this change in order for the `destroy()` logic above to work. I'm not totally sure what, if any, implications this might have elsewhere. I found a comment related to this https://github.com/emberjs/data/pull/6207/files#r306956786 so I was thinking you might have more context

igorT: 👍on this PR, but I will double check with @runspired and @rwjblue  whether  this is still correct thinking

PR #7096, opened 2020-03-12
Optimize visiting the relating records

pieter-v:
Visiting all related records during `_cleanupOrphanedRecordDatas` allocates and copies a large amount of unnecessary data. Using an `Iterable` this allocation is not done anymore.
In one of our applications we have a large dataset with complex relations. Destroying a large number of records timed out after 10 minutes, with this fix it still takes two minutes (but it is already a big improvement)

### DIFF
--- a/packages/record-data/addon/-private/record-data.ts
+++ b/packages/record-data/addon/-private/record-data.ts
@@ -465,23 +465,44 @@ export default class RecordDataDefault implements RelationshipRecordData {
   }
 
   /**
-    Computes the set of internal models reachable from `this` across exactly one
+    Iterates over the set of internal models reachable from `this` across exactly one
     relationship.
-
-    @return {Array} An array containing the internal models that `this` belongs
-    to or has many.
-
   */
-  _directlyRelatedRecordDatas(): RecordData[] {
-    let array = [];
+  _directlyRelatedRecordDatasIterable = () => {
+    const initializedRelationships = this._relationships.initializedRelationships;
+    const relationships = Object.keys(initializedRelationships).map(key => initializedRelationships[key]);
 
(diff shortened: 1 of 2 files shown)


packages/record-data/addon/-private/record-data.ts
```diff
@@ -465,23 +465,47 @@ export default class RecordDataDefault implements RelationshipRecordData {
...
+        j = 0;
+        i++;
+      }
+      return undefined;
+    };
+
+    nextMember = findNext();
+
+    return {
+      [Symbol.iterator]() {
```
btecu: `Symbol` is not supported in IE

pieter-v: @btecu 
I understand that IE does not support `Symbol.iterator`, but running our application with this modification works perfectly with IE11. So, I assume that babel is taking care of this.
For the iterator in belongs-to-test, I can easily drop the stub function `_directlyRelatedRecordDatasIterable` and directly stub `_allRelatedRecordDatas` to return just `[this]`.
Or am I missing something?

runspired:
@pieter-v I'm a bit less certain about this change without also adding a perf test. Will likely also want to look at it running manually.

igorT:
@pieter-v Thanks so much for this deep dive! @runspired it seems like we landed more perf tests, is there anything blocking this?

runspired:
@igorT all we need to do is rebase so that the new performance test is run so that we can see the output from this change, I am otherwise 👍 on merging this.

pieter-v:
@runspired @igorT rebase done

runspired:
Performance check indicates a ~6500ms improvement for the unloading scenario, with other scenarios unchanged.

PR #7098, opened 2020-03-15
Revert "Fix nightly run"

Cunero-Lang:
Reverts emberjs/data#7094

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -2,7 +2,7 @@ name: 'Nightly'
 
 on:
   schedule:
-    - cron: '0 14 * * *' # 14:00 in UTC is 6am PST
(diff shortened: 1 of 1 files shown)


rwjblue:
Is there some context here? Why is this being reverted?

Cunero-Lang:
Porque si funciona. Todo método requiere comprobaciones. Porque la info de una rama  y otra rama,  no son tan iguales.

runspired:
Closing this as I would prefer to keep the fix unless that fix wasn't enough, in which case lets determine the correct fix.

Cunero-Lang:
Hey @runspired why (?) I need it running two days. Tomorrows night is other history and components is full for us... please, Dont "snif" here.

Issue #7102, opened 2020-03-18
Reference API is currently incompatible with Tracked


runspired:
If a getter uses the references API since no leaves are tracked it will be memoized by the template and never re-run. To fix this our previously discussed creation of per-record tags for properties to consume should be completed and reference methods should consume those tags.

runspired:
@igorT likely going to implement this soon, but trying to decide what to do about the notification-manager only accepting a single callback-per-identifier currently. Time to change that?

PR #7103, opened 2020-03-18
[BUGFIX] dont use Ember.get on internals

### LINKED ISSUE #7100: Using reference api in a getter fails with "You attempted to update [object Object].hasAnyRelationshipData to "true", but it is being tracked by a tracking context"
### Reproduction

https://ember-twiddle.com/fd9a41b63bc96b2ae43118bad001a224?openFiles=controllers.application%5C.js%2C

runspired:
Resolves #7100 

Surfaced #7102


runspired:
I realized I merged this without fixing the LTS 3.12 test. Will follow up before back porting. The LTS 3.8 test is likely an identical fix but matters less as on master we should be dropping support for that now that 3.16 is LTS.

PR #7104, opened 2020-03-19
[FEAT] RFC-521 Accept Identifiers for store methods

pete-the-pete:
ref https://github.com/emberjs/rfcs/pull/521

### DIFF
--- a/packages/-ember-data/tests/integration/adapter/rest-adapter/find-record-test.js
+++ b/packages/-ember-data/tests/integration/adapter/rest-adapter/find-record-test.js
@@ -1,6 +1,8 @@
 import Pretender from 'pretender';
 import { module, test } from 'qunit';
 
+import { assign } from '@ember/polyfills';
+
 import { setupTest } from 'ember-qunit';
 
 import RESTAdapter from '@ember-data/adapter/rest';
@@ -54,20 +56,20 @@ module('integration/adapter/rest_adapter - REST Adapter - findRecord', function 
 
   // Ok Identifier tests
   [
-    { type: 'post', id: '1', desc: 'type and id' },
-    { type: 'post', id: '1', lid: 'post:1', desc: 'type, id and lid' },
+    { withType: true, withId: true, desc: 'type and id' },
+    { withType: true, withId: true, desc: 'type, id and lid' },
     {
-      type: 'post',
+      withType: true,
+      withLid: true,
       desc: 'type and lid',
-      withLid: true,
     },
     {
-      type: 'post',
-      id: null,
-      desc: 'type, null id, and lid',
+      withType: true,
       withLid: true,
+      extra: { id: null },
+      desc: 'type, null id, and lid',
     },
-  ].forEach(({ type, id, lid, desc, withLid }) => {
+  ].forEach(({ withType, withId, withLid, extra, desc }) => {
     test(`findRecord - basic payload (${desc})`, async function (assert) {
       const Post = Model.extend({
         name: attr('string'),
@@ -76,21 +78,25 @@ module('integration/adapter/rest_adapter - REST Adapter - findRecord', function 
       this.owner.register('adapter:application', RESTAdapter.extend());
       this.owner.register('serializer:application', RESTSerializer.extend());
 
+      let id = '1';
       const store = this.owner.lookup('service:store');
       const adapter = store.adapterFor('application');
-      const ajaxCallback = ajaxResponse(adapter, { posts: [{ id: '1', name: 'Rails is omakase' }] });
-      const allArgs = { type, id, lid };
-      const findRecordArgs = {};
-      Object.keys(allArgs).forEach((key) => {
-        if (typeof allArgs[key] !== 'undefined') {
-          findRecordArgs[key] = allArgs[key];
-        }
-      });
+      const ajaxCallback = ajaxResponse(adapter, { posts: [{ id, name: 'Rails is omakase' }] });
 
+      const findRecordArgs = Object.create(null);
+      if (withType) {
+        findRecordArgs.type = 'post';
+      }
+      if (withId) {
+        findRecordArgs.id = id;
+      }
       if (withLid) {
         // create the identifier without creating a record
-        const identifier = store.identifierCache.getOrCreateRecordIdentifier({ ...findRecordArgs, id: '1' });
+        const identifier = store.identifierCache.getOrCreateRecordIdentifier({ ...findRecordArgs, id });
         findRecordArgs.lid = identifier.lid;
+      }
+      if (extra) {
+        assign(findRecordArgs, extra);
       }
 
       const post = await store.findRecord(findRecordArgs);
@@ -108,17 +114,12 @@ module('integration/adapter/rest_adapter - REST Adapter - findRecord', function 
   // Error Identifier Tests
   [
     {
-      type: 'post',
-      id: null,
+      withType: true,
+      extra: { id: null },
       desc: 'type and no id',
       errorMsg: 'Assertion Failed: Expected an identifier object with (type and id) or lid',
     },
-    {
-      lid: 'post:1',
-      desc: 'lid with no type',
-      errorMsg: 'resource.type needs to be a string',
-    },
-  ].forEach(({ type, id, lid, desc, errorMsg }) => {
+  ].forEach(({ withType, withLid, extra, desc, errorMsg }) => {
     testInDebug(`findRecord - will assert with (${desc})`, async function (assert) {
       const Post = Model.extend({
         name: attr('string'),
@@ -128,13 +129,18 @@ module('integration/adapter/rest_adapter - REST Adapter - findRecord', function 
       this.owner.register('serializer:application', RESTSerializer.extend());
 
       const store = this.owner.lookup('service:store');
-      let allArgs = { type, id, lid };
-      let findRecordArgs = {};
-      Object.keys(allArgs).forEach((key) => {
-        if (typeof allArgs[key] !== 'undefined') {
-          findRecordArgs[key] = allArgs[key];
-        }
-      });
+      const findRecordArgs = Object.create(null);
+      if (withType) {
+        findRecordArgs.type = 'post';
+      }
+      if (withLid) {
+        // create the identifier without creating a record
+        const identifier = store.identifierCache.getOrCreateRecordIdentifier({ ...findRecordArgs, id: '1' });
+        findRecordArgs.lid = identifier.lid;
+      }
+      if (extra) {
+        assign(findRecordArgs, extra);
+      }
 
       assert.expectAssertion(async () => {
         await store.findRecord(findRecordArgs);
--- a/packages/-ember-data/tests/integration/adapter/rest-adapter/find-record-test.js
+++ b/packages/-ember-data/tests/integration/adapter/rest-adapter/find-record-test.js
@@ -57,17 +57,17 @@ module('integration/adapter/rest_adapter - REST Adapter - findRecord', function 
   [
     { type: 'post', id: '1', desc: 'type and id' },
     { type: 'post', id: '1', lid: 'post:1', desc: 'type, id and lid' },
-    {
-      type: 'post',
-      desc: 'type and lid',
-      pushRecord: true,
-    },
-    {
-      type: 'post',
-      id: null,
-      desc: 'type, null id, and lid',
-      pushRecord: true,
-    },
+    // {
+    //   type: 'post',
+    //   desc: 'type and lid',
+    //   pushRecord: true,
+    // },
+    // {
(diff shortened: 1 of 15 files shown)


packages/store/addon/-private/identifiers/cache.ts
```diff
@@ -20,6 +20,15 @@ type ResourceIdentifierObject = import('../ts-interfaces/ember-data-json-api').R
...
 type ConfidentDict<T> = import('../ts-interfaces/utils').ConfidentDict<T>;
 
+/**
+ * Peekable - (id && type) || lid
+ * - Identifiers are guaranteed to have a lid
+ * - ResourceIdentifierObject is Existing or New
+ *    - NewResources are guaranteed to have a lid
+ *    - ExistingResources are guaranteed to have an id AND type
+ */
+export type Peekable = RecordIdentifier | ResourceIdentifierObject;
```
igorT: @runspired lets try to come up with a better name for this type

packages/store/addon/-private/identifiers/cache.ts
```diff
@@ -157,18 +163,31 @@ export class IdentifierCache {
       return resource;
     }
 
+    let lid = coerceId(resource.lid);
+    let identifier: StableRecordIdentifier | undefined = lid !== null ? this._cache.lids[lid] : undefined;
```
igorT: Can we simplify this line please, it's pretty hard to read for me at least

packages/store/addon/-private/identifiers/cache.ts
```diff
@@ -157,18 +163,31 @@ export class IdentifierCache {
...
+
+    if (identifier !== undefined) {
+      return identifier;
+    }
+
+    const _resource = resource as ResourceIdentifierObject;
```
igorT: We don't use const if it's not inside module scope. 
Lets try to rewrite this so TS knows that this is a resourceIdentifier

igorT: The same way we have `isStableIdentifier` this might be more readable if we added `isIdentifier`

packages/store/addon/-private/identifiers/cache.ts
igorT: If identifier is stableRecordIdentifier, wouldn't it have been returned on line 163?

packages/store/addon/-private/system/core-store.ts
```diff
@@ -1467,7 +1471,19 @@ abstract class CoreStore extends Service {
     @param {String|Integer} id
     @return {Model|null} record
   */
-  peekRecord(modelName: string, id: string | number): RecordInstance | null {
+  peekRecord(modelName: string, id: string | number): RecordInstance | null
+  peekRecord(identifier: Peekable): RecordInstance | null
+  peekRecord(identifier: Peekable | string, id?: string | number): RecordInstance | null {
+
+    if (arguments.length === 1) {
+      const _identifier = identifierCacheFor(this).peekRecordIdentifier(identifier as Peekable);
```
igorT: We don't use const if it's not in module scope

packages/store/addon/-private/system/core-store.ts
```diff
@@ -1467,7 +1471,19 @@ abstract class CoreStore extends Service {
...
+      if (_identifier) {
+        return internalModelFactoryFor(this).peek(_identifier);
+      }
+      return null;
+    }
+
+    const modelName = identifier as string;
```
igorT: We should try to not have to cast these

packages/store/addon/-private/identifiers/cache.ts
runspired: Agree but for now I think this is fine and communicates the intent.

pete-the-pete:
Cool...

My life has stabilized more lately, so I should be able to come back to
this, probably this weekend.

packages/store/addon/-private/identifiers/cache.ts
snewcomer: For future knowledge, do we prefix scoped variables with `_`?  Based on usage in other languages, I tend to associate it with an unused variable that is given a descriptive name...

packages/store/addon/-private/identifiers/cache.ts
pete-the-pete: Resolving because this is outdated.

snewcomer:
@pete-the-pete Is this still something you might be interested in picking back up?  If not, I'd be happy to take it on if you are busy.

pete-the-pete:
have at it 👍

packages/store/addon/-private/system/store/internal-model-factory.ts
```diff
@@ -208,7 +208,7 @@ export default class InternalModelFactory {
   }
 
   getByResource(resource: ResourceIdentifierObject): InternalModel {
-    const normalizedResource = constructResource(resource.type, resource.id, resource.lid);
+    const normalizedResource = constructResource(resource.type as string, resource.id, resource.lid);
```
snewcomer, line 211: What do we think about this explicit cast?

packages/-ember-data/tests/integration/adapter/rest-adapter/find-record-test.js
```diff
@@ -51,6 +53,158 @@ module('integration/adapter/rest_adapter - REST Adapter - findRecord', function
...
+      assert.equal(post.get('name'), 'Rails is omakase');
+
+      // stress tests
+      let peekPost = await store.peekRecord(findRecordArgs);
+      assert.strictEqual(peekPost, post, 'peekRecord returns same post');
+
+      let recordReference = store.getReference(findRecordArgs);
+      assert.equal(recordReference.remoteType(), 'identity');
+      assert.equal(recordReference.type, 'post');
+      assert.equal(recordReference.id(), 1);
```
runspired, line 129: should be string '1' 
```suggestion
      assert.equal(recordReference.id(), '1');
```

packages/-ember-data/tests/integration/adapter/rest-adapter/find-record-test.js
```diff
@@ -51,6 +53,158 @@ module('integration/adapter/rest_adapter - REST Adapter - findRecord', function
...
+
+      assert.equal(passedUrl, '/posts/1');
+      assert.equal(passedVerb, 'GET');
+      assert.deepEqual(passedHash.data, {});
+
+      assert.equal(post.get('id'), '1');
+      assert.equal(post.get('name'), 'Rails is omakase');
+
+      // stress tests
+      let peekPost = await store.peekRecord(findRecordArgs);
```
runspired, line 123: peekRecord does not require await

```suggestion
      let peekPost = store.peekRecord(findRecordArgs);
```

packages/-ember-data/tests/integration/adapter/rest-adapter/find-record-test.js
```diff
@@ -51,6 +53,158 @@ module('integration/adapter/rest_adapter - REST Adapter - findRecord', function
...
+    const store = this.owner.lookup('service:store');
+    // create record locally so no fetch is needed
+    const record = store.createRecord('post');
+    const identifier = recordIdentifierFor(record);
+
+    const foundPost = await store.findRecord(identifier, { reload: false, backgroundReload: false });
+    assert.strictEqual(record, foundPost, 'We were able to findRecord');
+
+    // stress tests
+    let peekPost = await store.peekRecord(identifier);
```
runspired, line 150:
```suggestion
    let peekPost = store.peekRecord(identifier);
```

packages/-ember-data/tests/integration/identifiers/cache-test.ts
```diff
@@ -130,6 +131,20 @@ module('Integration | Identifiers - cache', function (hooks) {
...
+        id: '1',
+        attributes: {
+          name: 'runspired',
+        },
+      };
+      cache.createIdentifierForNewRecord(runspiredHash);
+      assert.expectAssertion(() => {
+        cache.createIdentifierForNewRecord(runspiredHash);
+      }, 'The lid generated for the new record is not unique as it matches an existing identifier');
+    });
```
runspired, line 146: awesome catch, great test 👍🏻

packages/-ember-data/tests/integration/references/record-test.js
```diff
@@ -32,75 +33,134 @@ module('integration/references/record', function (hooks) {
     assert.equal(recordReference.id(), 1);
   });
 
-  test('push(object)', function (assert) {
-    var done = assert.async();
+  test('a RecordReference can be retrieved via store.getReference(identifier) without local state', function (assert) {
+    let store = this.owner.lookup('service:store');
+    let recordReference = store.getReference({ type: 'person', id: 1 });
```
runspired, line 38:
```suggestion
    let recordReference = store.getReference({ type: 'person', id: '1' });
```

packages/-ember-data/tests/integration/references/record-test.js
```diff
@@ -32,75 +33,134 @@ module('integration/references/record', function (hooks) {
...
+
+    assert.equal(recordReference.remoteType(), 'identity');
+    assert.equal(recordReference.type, 'person');
+    assert.equal(recordReference.id(), 1);
```
runspired, line 42:
```suggestion
    assert.equal(recordReference.id(), '1');
```

note: the reason these tests pass even though the id will be coerced to string is that `equal` will type-convert. We may want to use strictEqual in the future for id tests.

runspired (approved):
Overall LGTM but there's a few small cleanups to make.

I also find the references-test version of our drier-iteration style test murkier than the corresponding find-record test version. Not asking for a refactor of it, but will probably clean it up a little later.

packages/-ember-data/tests/integration/references/record-test.js
```diff
@@ -32,75 +33,134 @@ module('integration/references/record', function (hooks) {
...
+    },
+  ].forEach(({ withType, withId, withLid, extra, isCreate, generateLid, desc }) => {
+    test(`a RecordReference can be retrieved with ${desc}`, function (assert) {
+      let store = this.owner.lookup('service:store');
+      let person;
+      if (isCreate) {
+        // no id
+        person = store.createRecord('person');
+      } else {
+        person = store.push({
```
runspired, line 74: I don't think we need to push in a record for the non-create case. We should be able to get a reference without a record existing yet.

packages/-ember-data/tests/integration/references/record-test.js
```diff
@@ -32,75 +33,134 @@ module('integration/references/record', function (hooks) {
...
+
+      const getReferenceArgs = Object.create(null);
+      if (withType) {
+        getReferenceArgs.type = 'person';
+      }
+      if (withId && !isCreate) {
+        getReferenceArgs.id = '1';
+      }
+      if (withLid) {
+        if (generateLid) {
```
runspired, line 93: I think this can be simplified to `!isCreate`, and I think we can remove one of the test scenarios from the array config above. we'll always generate in the non-create case. I don't think we need to worry about testing the output of the cache generation vs `recordIdentifierFor` as other tests cover and ensure that these things match.

packages/-ember-data/tests/integration/references/record-test.js
```diff
@@ -32,75 +33,134 @@ module('integration/references/record', function (hooks) {
...
+    { withType: true, withId: true, withLid: true, desc: 'type, id and lid from cache' },
+    { withType: true, withLid: true, exta: { id: null }, desc: 'type, null id, and lid' },
+    {
+      withType: true,
+      withLid: true,
+      exta: { id: null },
+      isCreate: true,
+      desc: 'type, null id, and lid via store.createRecord',
+    },
+  ].forEach(({ withType, withId, withLid, extra, isCreate, generateLid, desc }) => {
```
runspired, line 66: extra will always be undefined as above it is set as `exta`. We should just remove it entirely as it is functioning identically to the idea of `withId` and where we've used it it has been the same as `withId: undefined` which is falsey. No id present in the args used to build the resource-identifier vs null-id isn't material here, and the tests currently are using the no-id present path anyway :)

packages/store/addon/-private/system/core-store.ts
```diff
@@ -843,7 +843,8 @@ abstract class CoreStore extends Service {
     has already been assigned an id, you can find the record again using just the lid.
 
     ```app/routes/post.js
-    store.findRecord({ lid });
+    let identifier = recordIdentifierFor(person);
+    store.findRecord({ lid: identifier.lid });
```
runspired, line 847: since you can just pass the identifier directly, I feel demonstrating this in this way is a bit awkward. Until we are able to document accessing identifiers directly from cache (RecordData) apis we should leave how the lid was obtained somewhat ambiguous. Folks using them will have already brushed up on identifier documentation so I think this should be ok.

PR #7109, opened 2020-04-01
clear relationships during delete

pieter-v:
While investigating memory usage of our application, where we load a large amount of data and afterwards destroy/unload it again, I noticed that the `recordData` objects are not destroyed and kept consuming memory. I tracked it down to `Relationships` not being cleared during unload.


pieter-v:
This also fixes the failing test defined in #6600

runspired:
@pieter-v thanks! could you work with @eodb to clean up #6600 a bit so that we can close that off too?

PR #7126, opened 2020-04-15
Bugfix for wrong hasMany relationship state in scenario of sideposting with lid

andreyfel:
The issue appeared in a sideposting scenario:
- model A has has many relationship to model B.

- instance of model A and instance of model B are created on the client,

- model A is saved along with model B sent in `include` section of payload; the linking is done through `lid`

- server persisted both models and sent the response with server-generated `id` along with `lid` sent by the client.

- when hasMany.updateData() was called while the response payload was  pushed to the store it couldn't match the record which was already in the store with the record data contained in the relationship data because `lid` was omitted.

### DIFF
--- a/packages/-ember-data/tests/integration/identifiers/lid-reflection-test.ts
+++ b/packages/-ember-data/tests/integration/identifiers/lid-reflection-test.ts
@@ -1,11 +1,11 @@
 import { module, test } from 'qunit';
-import { defer } from 'rsvp';
+import { defer, resolve } from 'rsvp';
 
 import { setupTest } from 'ember-qunit';
 
 import Adapter from '@ember-data/adapter';
 import { RECORD_DATA_STATE } from '@ember-data/canary-features';
-import Model, { attr } from '@ember-data/model';
+import Model, { attr, belongsTo, hasMany } from '@ember-data/model';
 import Serializer from '@ember-data/serializer';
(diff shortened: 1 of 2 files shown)


packages/-ember-data/tests/integration/identifiers/lid-reflection-test.ts
```diff
@@ -145,4 +145,79 @@ module('Integration | Identifiers - lid reflection', function(hooks) {
 
     assert.strictEqual(record.name, '@runspired', 'After we finish we use the most recent clean name');
   });
+
+  test('lid works for sideposting scenario', async function(assert) {
```
snewcomer, line 149: Could we beef up this test description?  Maybe something to do with "updating the client record after save" or something?

snewcomer (approved):
This was probably the result of some good/deep investigation.  Great find here.

packages/-ember-data/tests/integration/identifiers/lid-reflection-test.ts
andreyfel: Updated the test name

andreyfel:
@snewcomer yeah, I've spent a day debugging the ember-data internals :)

I've squashed the commits so it is easier to backport the fix to LTS and other branches.

PR #7129, opened 2020-04-22
Remove redundant devDependencies

dcyriller:
### Description

`ember-cli-fastboot`, `ember-cli-fastboot-testing` and `ember-simple-tree` are used in fastboot-test-app workspace only and listed in the workspace package.json. This PR proposes to remove them from the main packages.json

This PR also proposes to bump ember-simple-tree.

### Why

`ember-simple-tree@0.8.0` allows to use NodeJs 13.


dcyriller:
I see that there is a test failing in the external partners test suite:
```
Error: Assertion Failed: Cannot update watchers for `avatarUrl` on `<travis@model:user::ember461:1>` after it has been destroyed.
```
If you have any pointer to help me start digging, I could try to fix it.

runspired:
@dcyriller would love someone to get that fixed, the repo is `travis-web`, I can pair with you to show you how to install local EmberData into the project since linking doesn't always work.

runspired:
@dcyriller I moved travis to allowed-failures for now, we should undo that once we fix: https://github.com/emberjs/data/pull/7142

dcyriller:
I have been able to run travis-web with different ember-data versions. Which allowed me to bisect the issue. It appeared between `ember-data@3.12.6` and `ember-data@3.13.0-beta.0`. Running travis-web test suite with `ember-data@3.12.6` works fine. And the failure is present with `ember-data@3.13.0-beta.0`. In between was released `ember-data@3.13.0-canary.0`, with which I had a different error.

But I believe there is a concurrency issue in travis-web 😢. A 403 failure to fetch the current user seems to trigger a logout and a second attempt at logging in. In parrallel are run a [`this.signOut`](https://github.com/travis-ci/travis-web/blob/master/app/services/auth.js#L214) function that triggers `this.store.unloadAll()` and a second attempt at signing in that triggers a [userRecord.reload](https://github.com/travis-ci/travis-web/blob/master/app/services/auth.js#L209). The second attempt at signing in is triggered by the `application` route's `beforeModel` hook while transitionning to `signin`, which might not be expected.

I'm not sure why it appeared with 3.13. But if there is a bug in ember-data, it will be hard to find because of the concurrency issue 😞.

PR #7146, opened 2020-04-27
[BUGFIX]: Fix errors while running yarn start

### LINKED ISSUE #7072: Running tests via `yarn start` fails
### Reproduction

```
yarn start
```

Then visit the tests page

### Description

Several modules fail to load with messages like the following:

```
Error: Could not find module `@ember-data/unpublished-test-infra/version` imported from `@ember-data/unpublished-test-infra/test-support/deprecated-test`
```

### Versions

On master, doesn't affect published code/versions

snewcomer:
close #7072

Screen Shot 2020-04-26 at 9 26 59 PM

This PR allows a developer to execute `yarn start` and run the tests in the browser.

We add the `/version` file with some tag/sha information at build time iff `EMBER_CLI_TEST_COMMAND` is set by Node on the global variable `process.env` at runtime.  

Another option is to be more specific in the if conditional by allowing only development or test.  However, if this module is excluded from production, then this may be sufficient in its current state.

### DIFF
--- a/packages/unpublished-test-infra/index.js
+++ b/packages/unpublished-test-infra/index.js
@@ -1,5 +1,5 @@
 'use strict';
-
+const merge = require('broccoli-merge-trees');
 const version = require('@ember-data/private-build-infra/src/create-version-module');
 const addonBuildConfigForDataPackage = require('@ember-data/private-build-infra/src/addon-build-config-for-data-package');
 
@@ -8,12 +8,11 @@ const name = require('./package').name;
 const addonBaseConfig = addonBuildConfigForDataPackage(name);
 
 module.exports = Object.assign({}, addonBaseConfig, {
-  treeForAddon() {
-    if (process.env.EMBER_CLI_TEST_COMMAND) {
-      const options = this.getEmberDataConfig();
-      let compatVersion = options.compatWith;
-      let tree = version(compatVersion);
-      return this.debugTree(this._super.treeForAddon.call(this, tree), 'addon-output');
-    }
+  treeForAddonTestSupport(existingTree) {
+    const options = this.getEmberDataConfig();
+    let compatVersion = options.compatWith;
+    let tree = merge([existingTree, version(compatVersion)]);
+
+    return this.debugTree(this._super.treeForAddonTestSupport.call(this, tree), 'addon-output');
   },
 });
--- a/packages/unpublished-test-infra/index.js
+++ b/packages/unpublished-test-infra/index.js
@@ -1,4 +1,5 @@
 'use strict';
+// eslint-disable-next-line node/no-unpublished-require
 const merge = require('broccoli-merge-trees');
 const version = require('@ember-data/private-build-infra/src/create-version-module');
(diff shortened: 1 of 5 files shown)


packages/unpublished-test-infra/index.js
```diff
@@ -9,11 +9,9 @@ const addonBaseConfig = addonBuildConfigForDataPackage(name);
...
-    if (process.env.EMBER_CLI_TEST_COMMAND) {
-      const options = this.getEmberDataConfig();
-      let compatVersion = options.compatWith;
-      let tree = version(compatVersion);
-      return this.debugTree(this._super.treeForAddon.call(this, tree), 'addon-output');
-    }
+    const options = this.getEmberDataConfig();
+    let compatVersion = options.compatWith;
+    let tree = version(compatVersion);
+    return this.debugTree(this._super.treeForAddon.call(this, tree), 'addon-output');
```
runspired, line 15: So this leaks the version module into our asset-size check because we now include it in production builds from unpublished-test-infra.

I don't think it legitimately affects end consumers, but it is frustrating to have it included in the check when it should not be.

There's probably a check we could do here, but it wouldn't just be "is production" since we test production output too. Maybe we check if the command is "serve" ? @rwjblue thoughts?

snewcomer: > we now include it in production builds from unpublished-test-infra.

That is a great point!  I assumed from the name of this package that it was not included in a published prod build; while also noticing it was a `devDependency` for various pkgs.  Side question - is there a short explanation or module that you could point me to in order to understand why `unpublished-test-infra` would be included in the process of a production build?

Also, would an alternative fix involve something like `process.env.EMBER_ENV`?

packages/unpublished-test-infra/index.js
```diff
@@ -8,12 +9,11 @@ const name = require('./package').name;
...
-      let compatVersion = options.compatWith;
-      let tree = version(compatVersion);
-      return this.debugTree(this._super.treeForAddon.call(this, tree), 'addon-output');
-    }
+  treeForAddonTestSupport(existingTree) {
+    const options = this.getEmberDataConfig();
+    let compatVersion = options.compatWith;
+    let tree = merge([existingTree, version(compatVersion)]);
+
+    return this.debugTree(this._super.treeForAddonTestSupport.call(this, tree), 'addon-output');
```
rwjblue, line 17: Can you change the debug tree name (this isn't the addon tree any more)?

packages/unpublished-test-infra/package.json
```diff
@@ -26,6 +26,7 @@
     "testem": "^3.0.3"
   },
   "devDependencies": {
+    "broccoli-merge-trees": "^4.2.0",
```
rwjblue, line 29: This needs to be a dependency not a devDep

runspired: @rwjblue it’s not published

rwjblue: trolol, its literally in the file name 🤦

PR #7167, opened 2020-05-06
[DOC EmbeddedRecordsMixin] Assert from JSONAPISerializer

### LINKED ISSUE #6991: clarify support and behavior around embedded records mixin
Initial documentation was landed in #6958, but we should make it clearer for folks by default. After discussing in the ember-data core meeting on 2020-02-19, we (@hjdivad, @rwjblue, and @efx) propose:

Updating the embedded records mixin to assert when used with `JSONAPISerializer` unless a specific flag to opt out of the assertion is used.

Something like:

```js
assert(
  'The embedded records mixin does not work with all JSON:API\'s, please confirm this works for your specific API and add `this.isEmbeddedRecordsMixinCompatible = true` to your serializer', 
  this.isJSONAPISerializer !== true || this.isEmbeddedRecordsMixinCompatible === true
);
```

allthesignals:
Use of EmbeddedRecordsMixin and JSONAPISerializer together asserts instead of warns. Allows for setting "isEmbeddedRecordsMixinCompatible" on the serializer for when parts of the API are compatible.

Closes #6991

### DIFF
--- a/packages/serializer/addon/json-api.js
+++ b/packages/serializer/addon/json-api.js
@@ -559,6 +559,17 @@ const JSONAPISerializer = JSONSerializer.extend({
 
 if (DEBUG) {
   JSONAPISerializer.reopen({
+    init(...args) {
+      this._super(...args);
+
+      assert(
+        `You've used the EmbeddedRecordsMixin in ${this.toString()} which is not fully compatible with the JSON:API specification. Please confirm that this works for your specific API and add \`this.isEmbeddedRecordsMixinCompatible = true\` to your serializer.`,
+        !this.isEmbeddedRecordsMixin || this.isEmbeddedRecordsMixinCompatible === true,
+        {
+          id: 'ds.serializer.embedded-records-mixin-not-supported',
+        }
+      );
+    },
     willMergeMixin(props) {
       let constructor = this.constructor;
       warn(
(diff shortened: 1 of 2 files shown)


packages/serializer/addon/json-api.js
```diff
@@ -568,9 +568,9 @@ if (DEBUG) {
           id: 'ds.serializer.json-api.extractMeta',
         }
       );
-      warn(
+      assert(
         'The JSONAPISerializer does not work with the EmbeddedRecordsMixin because the JSON API spec does not describe how to format embedded resources.',
-        !props.isEmbeddedRecordsMixin,
+        !props.isEmbeddedRecordsMixin || props.isEmbeddedRecordsMixinCompatible === true,
```
allthesignals, line 573: Probably want a test for allowing this

hjdivad: yep, also mention the opt-in flag in the assertion message

hjdivad:
thanks for taking a look at this @allthesignals!

allthesignals:
Looking at this again tonight, but I'm a bit stuck.

I need a way to assert based on values from both the mixed-in class and the mixin itself (check isEmbeddedRecordsMixin on the mixin and isEmbeddedRecordsMixinCompatible on the user class extension).

Right now, the EmbeddedRecordsMixin assertion lives on the JSONAPISerializer itself, and is run with the `willMergeMixin` hook (which I can't find documentation for).

There are some other assertions sprinkled throughout the JSONAPISerializer class, so I think my best bet is to move the assertion check there, unless there is a better approach.

hjdivad:
@runspired do we expect the IE11 asset size check to be this sensitive?  it's complaining about a 50B (not Kb) increase.

allthesignals:
Just guessing but maybe the inclusion of the code inside `init` means it doesn't get stripped out during some production build step? Might be why some of these other `warn`s are added into the `reopen` inside the debug check.

hjdivad:
thanks @allthesignals!

runspired:
@hjdivad we do expect it to be that sensitive. The idea is that any regression is one we are acutely aware of, vs allowing lots of tiny drift to be added over time unnoticed.

Issue #7170, opened 2020-05-09
[INFRA]: Add Safari support to testing pipeline


snewcomer:
Currently, IE11 and Chrome are tested in CI.  By adding Safari support, we will expand our testing coverage beyond IE11 and Chromium based browsers.

https://github.com/testem/testem/blob/master/README.md#launchers

runspired:
I tried to add support for Safari a while back and found its not currently possible due to limitations with safari's security model

- https://github.com/testem/testem/issues/1387
- https://github.com/karma-runner/karma-safari-launcher/issues/29

As far as I can tell there's been no further movement to make it possible.

snewcomer:
ok closing

PR #7171, opened 2020-05-09
[INFRA]: Add Firefox to test pipeline

snewcomer:
- Add Firefox 
- Improve testing scenarios with IE and CI testem launchers

close https://github.com/emberjs/data/issues/7170

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -152,21 +152,17 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        #scenario: [with-native-fetch, with-ember-fetch-no-jquery, with-ember-fetch-and-jquery, with-jquery]
-        launcher: [Chrome, Firefox, IE, Safari]
+        scenario: [with-native-fetch, with-ember-fetch-no-jquery, with-ember-fetch-and-jquery, with-jquery]
+        launcher: [Chrome, Firefox, IE]
         os: [macos-latest, windows-latest, ubuntu-latest]
         exclude:
           - os: macos-latest
             launcher: IE
-          # - os: macos-latest
-          #   scenario: with-ember-fetch-and-jquery
-          #   launcher: Chrome
+          - os: macos-latest
+            scenario: with-ember-fetch-and-jquery
+            launcher: Chrome
           - os: ubuntu-latest
             launcher: IE
-          - os: windows-latest
-            launcher: Safari
-          - os: ubuntu-latest
-            launcher: Safari
     runs-on: ${{ matrix.os }}
     name: ${{matrix.scenario}} ${{matrix.launcher}} on ${{matrix.os}}
     steps:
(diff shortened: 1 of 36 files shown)


packages/unpublished-debug-encapsulation-test-app/testem.js
```diff
@@ -2,17 +2,34 @@
...
+  if (TestSafari) {
+    return ['Safari'];
+  }
+
+  if (TestIE) {
+    return ['IE'];
+  }
+
+  return ['Chrome'];
+}
```
snewcomer, line 26: If I got this right, perhaps we could move `browserPicker` to `unpublished-test-infra`

.github/workflows/main.yml
```diff
@@ -149,6 +149,31 @@ jobs:
         timeout-minutes: 4
         run: yarn test:fastboot ${{ matrix.scenario }}
 
+  basic-tests-safari:
+    needs: [lint]
+    strategy:
+      matrix:
+        os: [macos-latest, windows-latest, ubuntu-latest]
```
runspired, line 156:
```suggestion
        os: [macos-latest]
```

packages/-ember-data/config/ember-try.js
```diff
@@ -14,6 +14,13 @@ module.exports = function() {
           },
           npm: {},
         },
+        {
+          name: 'with-safari',
+          env: {
+            TARGET_SAFARI: true,
```
runspired, line 20: we shouldn't need this since we've got the other test runs specified with this arg in azure and GitHub CI

packages/-ember-data/testem.js
```diff
@@ -1,16 +1,34 @@
...
 if (TestIE) {
   // eslint-disable-next-line no-console
   console.log('\n\nLaunching with IE\n\n');
 }
+if (TestSafari) {
+  // eslint-disable-next-line no-console
+  console.log('\n\nLaunching with Safari\n\n');
+}
+
+function browserPicker() {
```
runspired, line 15: To avoid the function and the duplication across testem.js files we could change this from boolean process flags to a string.

```
const BrowserLauncher = process.env.TESTEM_CI_LAUNCHER || 'Chrome'; 

...


launch_in_ci: [BrowserLauncher],

Comment (runspired) [reply packages/-ember-data/testem.js:15]: The log could be

```
console.log(`\n\nLaunching with ${BrowserLauncher}\n\n`);
```

File: .github/workflows/main.yml [hunk 5]
```diff
@@ -149,6 +149,31 @@ jobs:
         timeout-minutes: 4
         run: yarn test:fastboot ${{ matrix.scenario }}
 
+  basic-tests-safari:
+    needs: [lint]
+    strategy:
+      matrix:
+        os: [macos-latest]
```
runspired, line 156: this should look like the same strategy as the basic-tests-ie11 but with `runs-on` set to `macos-latest` and with the env set to enable Safari.

snewcomer: great!  Another roadblock it seems.  OpenGL and OpenCL were deprecated in Catalina.  GH Actions doesn't support Mojave anymore.  Looking at alternative solutions....

runspired:
```
Safari 13.1 (15609.1.20.111.8)
SafariDriver 13.1 (15609.1.20.111.8)
```

^ these are supposedly installed but testem isn't seeing them it seems

snewcomer:
Screen Shot 2020-05-14 at 9 33 21 PM

Hopefully this PR is not kaput but I'm not sure we have the right sandbox infra to test Safari.  Since OpenGL and OpenCL were deprecated in Catalina and GH Actions doesn't support Mojave, I don't know of a way around this.

PR #7209, opened 2020-06-02
[DOCS] Remove blank package '@ember-data'

Gaurav0:
A blank package `@ember-data` shows up due to these yuidoc comments.

I tried to get this working correctly so that an `ember-data` package would show up with a DS and VERSION export, but was unable to figure out how to do it.


rwjblue:
Thanks!

PR #7215, opened 2020-06-05
[INFRA] Extend github actions timeouts to prevent CI failures

igorT:
`test-infra` was timing out midrun https://github.com/emberjs/data/runs/741935013 Exteded all of the times by 30-50%, and doubled test-infra

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -14,7 +14,7 @@ on:
 
 jobs:
   lint:
-    timeout-minutes: 4
+    timeout-minutes: 5
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


igorT:
@runspired is there a strong downside to extending? Given that we run into this for test-infra and variable performance of github actions, I would rather extend so we don't run into this again.

runspired:
@igorT changing to a matrix and applying it only to the run actually extends the timeout far more than increasing that number.

Instead of (for instance) 4 minutes to run all of the installs, builds and test runs for infra we would have 4 minutes for each test run and not include the time at all for builds and installs.

PR #7226, opened 2020-06-13
refactor: Native Model

runspired:
Now that our minimum Ember LTS is ~3.12 (and for current canary 3.16! although we didn't go that far here)~  🎉 **3.20** 🎉we can refactor Model to be a native class. (support for this began in 3.6 and decorator support began in 3.10)

Note: ~in 3.16 minimum support~ **we can embrace a @tracked world**, we should prepare :) (this also helps us prepare)

This PR does that refactor as minimally as possible. The few test changes were minor cleanups I did while investigating failures (not fixes, just made debugging easier) that seemed worthwhile to just keep.


runspired:
cc @snewcomer if we can land this then we can probably do some nice things to make sure that folks calling `destroyRecord` today don't get surprised by `isDeleted` suddenly returning false when #7217 lands

packages/model/addon/-private/model.js
```diff
@@ -423,13 +394,32 @@ const Model = EmberObject.extend(DeprecatedEvented, {
...
+    return get(meta, 'isError') || false;
+  }
+  set isError(v) {
+    if (REQUEST_SERVICE && DEBUG) {
+      throw new Error(`isError is not directly settable when REQUEST_SERVICE is enabled`);
+    } else {
+      let meta = getRecordMeta(this);
+      set(meta, 'isError', v);
+    }
+  }
```
snewcomer, line 416: My best guess is b/c Ember's internals relies on observers for query param bindings, we are getting the "used previously in the same computation" error after it was set.

https://github.com/emberobserver/client/blob/2c3cc0f1a1547e716477c9e18a40ebe48a6490a4/app/controllers/admin/size-calculation-results/index.js#L8

runspired: could be, we may need to return the value from the setter or some such (computeds have funny semantics for set)

igorT:
@runspired can we merge this?

snewcomer:
https://github.com/ef4/ember-resource-metadata/pull/25
https://github.com/embermap/ember-data-storefront/pull/396

snewcomer:
https://github.com/ef4/ember-data-relationship-tracker/pull/19

Ok final one to get these tests to pass.

packages/model/addon/-private/model.js
```diff
@@ -1862,9 +1722,36 @@ Model.reopenClass({
    @type Map
    @readOnly
    */
-  relationshipsByName: relationshipsByNameDescriptor,
+  @computed
+  static get relationshipsByName() {
```
igorT, line 1726: We need to cache all of the static relationships stuff for performance.

packages/model/addon/-private/model.js
```diff
@@ -19,16 +20,38 @@ import {
...
+}
+
+function getWithDefault(meta, prop, value) {
+  if (meta[prop] === undefined) {
+    meta[prop] = value;
+  }
+  return get(meta, prop);
+}
+
+const retrieveFromCurrentState = computedMacroWithOptionalParams(function retrieveState() {
```
igorT, line 50: We don't pass args to `retrieveFromCurrentState` so we don't need to use this right?

packages/model/addon/-private/model.js
```diff
@@ -372,14 +341,21 @@ const Model = EmberObject.extend(DeprecatedEvented, {
     @type {Boolean}
     @readOnly
   */
-  isValid: isValid,
+  @computed('currentState', 'errors.length')
```
igorT, line 344: Once ff is on, we should remove `currentState` dependency

packages/model/addon/-private/model.js
snewcomer: Correct.  This was just to reduce some duplication.  We can certainly make it like so if you would like...

```
@computed('currentState')
dirtyType() {
    return this._internalModel.currentState.dirtyType;
}
```

packages/model/addon/-private/model.js
snewcomer: Anything different we need to do or just add a comment?

packages/model/addon/-private/model.js
```diff
@@ -19,16 +20,38 @@ import {
...
+
+const RecordMeta = new WeakMap();
+function getRecordMeta(record) {
+  let meta = RecordMeta.get(record);
+  if (meta === undefined) {
+    meta = Object.create(null);
+    RecordMeta.set(record, meta);
+    if (DEBUG) {
+      record.____recordMetaCache = meta;
+    }
```
snewcomer, line 38: I don't remember the past conversations but I am not sure what the intention here was.

packages/model/addon/-private/model.js
igorT: We need this to noop in the case of feature flagged paths, right?

igorT: In order for `isDeletedCP` and similar to work the same

packages/model/addon/-private/model.js
igorT: We need to cache the map thats being created inside the method, do you know if computed will cache by default or we need to do it manually

packages/model/addon/-private/model.js
igorT: It was to help with debugging/console access

packages/model/addon/-private/model.js
```diff
@@ -422,13 +400,32 @@ const Model = EmberObject.extend(DeprecatedEvented, {
     @type {Boolean}
     @readOnly
   */
-  isError: isError,
+  get isError() {
```
igorT, line 403: This should be a CP right?

packages/model/addon/-private/model.js
```diff
@@ -445,7 +442,24 @@ const Model = EmberObject.extend(DeprecatedEvented, {
     @type {Boolean}
     @readOnly
   */
-  isReloading: isReloading,
+  get isReloading() {
```
igorT, line 445: This should be a cp

packages/model/addon/-private/model.js
runspired: I'll make sure to do that in the next pass (I'm going to make currentState a tracked property and all of these getters, it'll be a much simpler world while we wait for that flag to be on)

packages/model/addon/-private/model.js
runspired: computed will cache by default

runspired: but we should move to `@cached` in the next pass

packages/model/addon/-private/model.js
runspired: partner test is no longer failing

packages/model/addon/-private/model.js
runspired: isDeletedCP uses its own computed and keeps the flags

runspired (commented):
Audited and it seems all of @igorT's feedback was addressed.

runspired:
```
Benchmark Results Summary    

duration phase estimated regression +130ms [103ms to 149ms] OR +4.78% [3.78% to 5.47%]
Phase [navigationStart] => [start-find-all] phase no difference [-11ms to 17ms]
Phase [start-find-all] => [start-outer-materialization] phase no difference [-18ms to 13ms]
Phase [start-outer-materialization] => [stop-outer-materialization] phase estimated regression +132ms [118ms to 143ms] OR +18.93% [17.04% to 20.51%]
Phase [stop-outer-materialization] => [end-find-all] phase no difference [0ms to 0ms]
Phase [end-find-all] => [Test End] phase no difference [-1ms to 0ms]

```

Apparently this is a sizeable regression (~19%) over the .extends syntax. We should look to see if we've messed up anything obvious.

packages/-ember-data/tests/unit/model-test.js
```diff
@@ -373,21 +373,29 @@ module('unit/model - Model', function(hooks) {
 
     test('ID mutation (complicated)', async function(assert) {
       let idChange = 0;
+      let compChange = 0;
       const OddPerson = Model.extend({
         name: DSattr('string'),
-        idComputed: computed('id', function() {}),
-        idDidChange: observer('id', () => idChange++),
+        idComputed: computed('id', function() {
+          this.get('id');
```
runspired, line 380: I need to remove this line again

packages/store/addon/-private/system/model/internal-model.ts
```diff
@@ -165,15 +167,15 @@ export default class InternalModel {
...
-  declare currentState: any;
   declare error: any;
 
   constructor(public store: CoreStore | Store, public identifier: StableRecordIdentifier) {
     if (HAS_MODEL_PACKAGE) {
       _getModelPackage();
     }
-    this._tag = 0;
     this._id = identifier.id;
+    this._tag = 0;
```
igorT, line 177: Can you please add a comment for why the weirdness with the id and why we need the tag

packages/store/addon/-private/system/model/internal-model.ts
```diff
@@ -194,6 +196,10 @@ export default class InternalModel {
     this._isDematerializing = false;
     this._scheduledDestroy = null;
 
+    this._record = null;
```
igorT, line 199: Are these for shape?

packages/store/addon/-private/system/model/internal-model.ts
```diff
@@ -1353,27 +1360,33 @@ export default class InternalModel {
     this.store.recordArrayManager.recordDidChange(this.identifier);
   }
 
-  setId(id: string) {
+  setId(id: string, fromCache: boolean = false) {
+    if (this._isUpdatingId === true) {
```
igorT, line 1364: `if (this._isUpdatingId)`, no?

packages/store/addon/-private/system/model/internal-model.ts
runspired: yes. We previously relied on calling `reset` to set them initially but we no longer want to do so because of the change to `currentState`

packages/store/addon/-private/system/model/internal-model.ts
```diff
@@ -1353,27 +1360,33 @@ export default class InternalModel {
     this.store.recordArrayManager.recordDidChange(this.identifier);
   }
 
-  setId(id: string) {
+  setId(id: string, fromCache: boolean = false) {
```
igorT, line 1363: Can you please add a comment saying what fromCache means

packages/store/addon/-private/system/model/internal-model.ts
runspired: I prefer strict checks myself but yes, we could shorthand.

packages/model/addon/-private/model.js
```diff
@@ -69,88 +86,25 @@ function findPossibleInverses(type, inverseType, name, relationshipsSoFar) {
...
-      return rd.isNew();
-    } else {
-      return get(this._internalModel.currentState, 'isNew');
-    }
-  }).readOnly();
-} else {
-  isNewCP = retrieveFromCurrentState;
-}
+function computeOnce(target, key, desc) {
+  const cache = new WeakMap();
```
igorT, line 90: `let`, no?

packages/model/addon/-private/model.js
```diff
@@ -69,88 +86,25 @@ function findPossibleInverses(type, inverseType, name, relationshipsSoFar) {
...
-    if (rd.isNew) {
-      return rd.isNew();
-    } else {
-      return get(this._internalModel.currentState, 'isNew');
-    }
-  }).readOnly();
-} else {
-  isNewCP = retrieveFromCurrentState;
-}
+function computeOnce(target, key, desc) {
```
igorT, line 89: Can you please add a comment describing what the use case here is and why

packages/model/addon/-private/model.js
igorT: ^^

packages/model/addon/-private/model.js
```diff
@@ -472,26 +512,16 @@ const Model = EmberObject.extend(DeprecatedEvented, {
     @private
     @type {Object}
   */
-  currentState: RootState.empty, // defined here to avoid triggering setUnknownProperty
 
   /**
    @property _internalModel
    @private
    @type {Object}
    */
```
igorT, line 520: We still have the comment, but do we no longer have the prop or do we just not initiate it here? Also for store

packages/model/addon/-private/model.js
```diff
@@ -2184,21 +2100,219 @@ Model.reopenClass({
...
 
    @method toString
    @static
    */
-  toString() {
+  static toString() {
     return `model:${get(this, 'modelName')}`;
+  }
+}
+Model.prototype._internalModel = null;
```
igorT, line 2119: Why do we do this here on the prototype?

igorT (commented):
Left comments

packages/model/addon/-private/model.js
runspired: I think this might be to prevent a setter/getter install happening, but don't fully remember. It's been a thing here for a long time, might be able to remove now.

packages/model/addon/-private/model.js
runspired: we don't initialize here but still have it

snewcomer (approved):

PR #7227, opened 2020-06-15
feat: better debugging of WeakMapped data

runspired:
Converts our most commonly used WeakMap utils to use a WeakCache that handles improved debuggability by attaching a non-enumerable property from the key object pointing to the value when in DEBUG mode.

Basically we replace `new WeakMap<K, V>()` with `new WeakCache<K, V>(fieldName)`.

Once this lands then the following becomes available in DEBUG

- store#Symbol('internal-model-factory')
- Record#Symbol('identifier')
- Reference#Symbol('internal-model')

It also removes our former use of a WeakMap for the store's `identifierCache` as that feature is now fully available via a public property.

`WeakCache` *mostly* is the set of WeakMap APIs we use and just proxies while ensuring the field is added / updated when a new key:value pair appears.

It comes with two additional methods though which handle two patterns for set/lookup we see regularly: (1) `lookup` which will create on get if the value does not exist and (2) `getWithError` which will error on get if the value does not exist.

It additionally takes the `typescript` step of causing `has` to narrow, which is another pattern we would write custom utils for in the past. Specifically:

```ts
let weakCache = new WeakCache<Foo, Bar>();

if (weakCache.has(maybeFoo)) {
  // `maybeFoo` is now narrowed to type `Foo`
  // we encounter this when we use key-present checks to confirm the type of something internally
  // which we do with some regularity.
}
```


packages/store/addon/-private/system/weak-cache.ts
```diff
@@ -0,0 +1,61 @@
...
+    if (DEBUG && this._cache.has(obj)) {
+      throw new Error(`${obj} was already assigned a value for ${this._fieldName}`);
+    }
+    if (DEBUG) {
+      addSymbol(obj, this._symbol!, value);
+    }
+    this._cache.set(obj, value);
+  }
+
+  expect(obj: K): V {
```
igorT, line 37: @igorT think about naming of this

runspired: Yeah I couldn't think of a common "get and error if not present" method name. One option I've been considering is `expectGet`

runspired: renamed to `getWithError`, open to better names but this is also entirely internal

packages/store/addon/-private/system/weak-cache.ts
```diff
@@ -0,0 +1,73 @@
...
+  _fieldName?: string;
+  _generator?: (key: K) => V;
+  _expectMsg?: (key: K) => string;
+}
+
+class WeakCache<K extends object, V> {
+  private _cache = new WeakMap<K, V>();
+
+  constructor(_fieldName?: string, _generator?: (key: K) => V, _expectMsg?: (key: K) => string) {
+    this._generator = _generator;
```
snewcomer, line 16: Another option for a name could be `defaultValueGenerator` or `valueGenerator` as the argument name and/or property on this class.  Both don't seem particularly appealing though.

packages/store/addon/-private/identifiers/cache.ts
```diff
@@ -79,24 +80,11 @@ function defaultGenerationMethod(data: ResourceIdentifierObject, bucket: string)
...
-export function identifierCacheFor(store: CoreStore): IdentifierCache {
-  let cache = IdentifierCaches.get(store);
-
-  if (cache === undefined) {
-    cache = new IdentifierCache();
-    IdentifierCaches.set(store, cache);
-  }
-
-  return cache;
-}
```
snewcomer, line 93: I like this refactor!

packages/store/addon/-private/system/references/belongs-to.js
```diff
@@ -6,7 +6,7 @@ import { DEPRECATE_BELONGS_TO_REFERENCE_PUSH } from '@ember-data/private-build-i
 import { assertPolymorphicType } from '@ember-data/store/-debug';
 
 import recordDataFor from '../record-data-for';
-import { peekRecordIdentifier } from '../store/internal-model-factory';
+import { RecordCache } from '../store/internal-model-factory';
```
snewcomer, line 9: When I see this I think "not an instance of something".  But I see this title case used elsewhere so I assume this was blessed/common convention.

SYU15:
Can you give some scenarios where these would be helpful and/or expected input/output?

packages/store/addon/-private/system/references/belongs-to.js
runspired: We use it fairly consistently for both classes and module scope constants (we also use scream case for module scope constants, we should maybe move to that all the time)

runspired:
@SYU15 sure! This PR is about making the `DEBUG` mode experience of exploring the state of an application easier (so note: this is not for things like the inspector).

Today, in both Ember and EmberData there is a common pattern of managing "private" or "associated" state for a public object via a WeakMapped object in order to avoid polluting the public object. This prevents leaking private API and prevents key collisions. In the case where users can extend the public class (like Records) preventing key collisions is especially important.

An unfortunate side-effect of entangling the associated state to the key object via a WeakMap is that a user debugging the object cannot see or access the associated state from dev tools by any means unless they happen to know the WeakMap exists, a mechanism of accessing it, and jump through significant hurdles to do so when needed.

This PR makes it so that in DEBUG mode we attach a non-enumerable (and so non-serialized and non-iterated) key via either Symbol (if not IE11) or randomized string (if IE11) that explicitly points to the associated state from the object.  The use of this Symbol / randomized string prevents us colliding with existing keys on the object and in production we don't incur the cost of generating and defining these non-enumerable properties.

runspired:
Screenshots showing the results for adding this symbol in a few key places
 
- store#Symbol('internal-model-factory')
- Record#Symbol('identifier')
- Reference#Symbol('internal-model')

Screen Shot 2020-06-16 at 9 56 14 AM
Screen Shot 2020-06-16 at 9 58 07 AM
Screen Shot 2020-06-16 at 9 58 23 AM


**And what these places look like without:**

Screen Shot 2020-06-16 at 10 01 08 AM
Screen Shot 2020-06-16 at 10 01 23 AM
Screen Shot 2020-06-16 at 10 01 57 AM

(Note: for `Reference` the `internalModel` property is deprecated, which is why this *seems* redundant in the screenshot but is not. We will do this for `Model` and many more things soon.)

packages/store/addon/-private/system/weak-cache.ts
```diff
@@ -0,0 +1,70 @@
+import { DEBUG } from '@glimmer/env';
+
+import { addSymbol, symbol } from '../ts-interfaces/utils/symbol';
+
+interface WeakCache<K extends object, V> {
```
stefanpenner, line 5: Thoughts on a name such as `NamedWeakCache` ?

stefanpenner, line 5: Given that this class only provides value in debug, it seems unfortunate to ship the code in production.

stefanpenner: It seems odd to name this differently the `WeakMap`

stefanpenner, line 5: Also, why do we need a separate interface for the concrete class authored beneath?

packages/store/addon/-private/system/weak-cache.ts
```diff
@@ -0,0 +1,70 @@
...
+  _symbol?: Symbol | string;
+  _fieldName?: string;
+  _generator?: (key: K) => V;
+  _expectMsg?: (key: K) => string;
+}
+
+class WeakCache<K extends object, V> {
```
stefanpenner, line 12: Also, authoring this as a subclass removes the need for all the proxy style methods, and inherit all the existing `WeakMap` typings.

```js
class NamedWeakMap extends WeakMap {
  private label: string;
  constructor(label) {
    super();
    this.label = label;
  }

  // ....
}
```

packages/store/addon/-private/system/weak-cache.ts
runspired: Discussed OOB that since this provides additional cache methods `WeakCache` is a good name.

runspired: Discussed OOB (and since implemented) WeakCache directly extending WeakMap so as to reduce the footprint of what we need to do, but keeping the class since the extra sugar provides value even in production.

runspired: I'll leave a comment, this is a pattern in our toolbox we have used a few places where we utilize typescripts declaration merging to our advantage. Basically, it prevents typescript from inserting lots of `void 0` initializers for things that may not be present. This is especially valuable for properties that are only present in DEBUG where typing the concrete class with them would include those fields and their initialization even in production.

packages/store/addon/-private/system/weak-cache.ts
runspired: Implemented

PR #7230, opened 2020-06-16
[CHORE] najax deprecation when ember-fetch is also installed

### LINKED ISSUE #7219: Fetch + fastboot + jquery detection problem
If an app has `ember-cli-fastboot`, `ember-fetch` and `@ember/jquery` as dependencies, code in https://github.com/emberjs/data/blob/35b8abfb7d0fb23ae07a723955d5364d8eee398c/packages/adapter/addon/rest.js#L312-L323 doesn't seem to end up with `fetch`, and uses `najax` instead. Which leads to errors like `najax: method jqXHR."getAllResponseHeaders" not implemented`

Ideally, fetch would be used instead, perhaps with additional check like
```javascript
    if (this.fastboot && hasFetch) {
      return true;
    } else if (jQueryIntegrationDisabled) { 
```

Seems this method was added in https://github.com/emberjs/data/pull/6094/

Not sure if there is a reason for current logic though, or how to proceed.

snewcomer:
close #7219

This PR seeks to add a deprecation notice for najax (in fastboot) if installed.  We want to encourage users to install `ember-fetch` if they haven't already.  `useFetch = true` will be the default for ember data's adapters.


packages/adapter/addon/rest.js
```diff
@@ -1414,4 +1419,23 @@ function ajaxOptions(options, adapter) {
...
+        najax(options);
+      } else {
+        throw new Error(
+          'najax does not seem to be defined in your app. Did you override it via `addOrOverrideSandboxGlobals` in the fastboot server?'
+        );
+      }
+    },
+  });
+}
+
```
snewcomer, line 1440: Ideally this would be done in native class syntax, but I can certainly follow up after this with another PR to migrate

packages/adapter/addon/rest.js
```diff
@@ -1067,9 +1062,19 @@ const RESTAdapter = Adapter.extend(BuildURLMixin, {
...
+    } else if (get(this, 'fastboot.isFastBoot') && DEPRECATE_NAJAX) {
+      if (has('fetch')) {
+        deprecate(
+          'You have ember-fetch and jquery installed. To use ember-fetch, set `useFetch: true` in your adapter.  In 4.0, ember-data will fallback to ember-fetch instead of najax when both of these are installed.',
+          false,
+          {
+            id: 'ember-data:najax-fallback',
+            until: '4.0',
+          }
+        );
```
snewcomer, line 1076: I'm assuming we still want the deprecation since, from what I read, the debug-macros are opt-in (no default by reading the current installed e-d version and stripping away deprecations automatically)

packages/private-build-infra/addon/current-deprecations.ts
```diff
@@ -51,4 +51,5 @@ export default {
   DEPRECATE_SERIALIZER_QUERY_RECORD_ARRAY_RESPONSE: '3.4',
   DEPRECATE_BELONGS_TO_REFERENCE_PUSH: '3.16',
   DEPRECATE_REFERENCE_INTERNAL_MODEL: '3.19',
+  DEPRECATE_NAJAX: '4.0',
```
runspired, line 54: We should maybe clean up the description here, the version here is not "in what version will this deprecation be automatically removed" but rather "in what version was this deprecation introduced", in the case of current master that would be 3.21. That way an application can be on say version 3.25 and maybe not have resolved a deprecation introduced in 3.23 but can still strip resolved deprecations from versions prior by specifying 3.22

packages/adapter/addon/rest.js
```diff
@@ -1067,9 +1062,19 @@ const RESTAdapter = Adapter.extend(BuildURLMixin, {
...
+      }
```
runspired, line 1077: we should add and `else` that straigh up deprecates `najax` even when fetch is not present and tells folks to use something like ember-fetch

packages/adapter/addon/rest.js
```diff
@@ -309,18 +312,26 @@ const RESTAdapter = Adapter.extend(BuildURLMixin, {
...
+      if (this._useFetch !== undefined) {
+        return this._useFetch;
+      }
+      let ENV = getOwner(this).resolveRegistration('config:environment');
+      // TODO: https://github.com/emberjs/data/issues/6093
+      let jQueryIntegrationDisabled = ENV && ENV.EmberENV && ENV.EmberENV._JQUERY_INTEGRATION === false;
+
+      if (jQueryIntegrationDisabled) {
+        return true;
+      } else if (hasNajax || hasJQuery) {
```
runspired, line 326: we should move the hasNajax determination within something featureflagged, so something like

```js
else if (DEPRECATE_NAJAX && typeof najax !== 'undefined') {
  return false;
}
```

then move the jquery check into its own branch so we can strip that too on it's own later once possible.

You can see how this works here using: https://terser-playground.surge.sh/

```js
if (a) {
  c()
} else if (false && true) {
  d()
} else if (false && true) {
  e()
} else {
  b();
}
```

packages/adapter/addon/rest.js
```diff
@@ -299,7 +302,7 @@ const RESTAdapter = Adapter.extend(BuildURLMixin, {
     // Avoid computed property override deprecation in fastboot as suggested by:
     // https://deprecations.emberjs.com/v3.x/#toc_computed-property-override
     get() {
-      if (this._fastboot) {
+      if (this._fastboot !== undefined) {
```
runspired, line 305: since _fastboot is not a boolean (whereas _useFetch is) we don't need to add the undefined check for it, presence check is enough

packages/adapter/addon/rest.js
```diff
@@ -309,18 +312,26 @@ const RESTAdapter = Adapter.extend(BuildURLMixin, {
...
+      let ENV = getOwner(this).resolveRegistration('config:environment');
+      // TODO: https://github.com/emberjs/data/issues/6093
+      let jQueryIntegrationDisabled = ENV && ENV.EmberENV && ENV.EmberENV._JQUERY_INTEGRATION === false;
+
+      if (jQueryIntegrationDisabled) {
+        return true;
+      } else if (hasNajax || hasJQuery) {
+        return false;
+      } else {
+        return true;
```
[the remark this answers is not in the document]
runspired: We appear to still not cache off the value for some of the return branches

packages/adapter/addon/rest.js
```diff
@@ -5,12 +5,14 @@
 */
 
 import { getOwner } from '@ember/application';
+import { deprecate } from '@ember/application/deprecations';
```
rwjblue, line 8: Should import from `@ember/debug`

packages/adapter/addon/rest.js
```diff
@@ -1067,9 +1063,28 @@ const RESTAdapter = Adapter.extend(BuildURLMixin, {
...
+        );
+      } else {
+        deprecate(
+          'In 4.0, ember-data will default to ember-fetch instead of najax in FastBoot.  It is recommended that you install ember-fetch or similar as a fetch polyfill in FastBoot.',
+          false,
+          {
+            id: 'ember-data:najax-fallback',
+            until: '4.0',
+          }
+        );
```
rwjblue, line 1086: Hmmm, maybe we should move this deprecation to happen in the `useFetch` computed and funnel everything through that still? The thing I'm concerned with is a massive flood of deprecations to the console that are all exactly the same. AFAICT, we are going to trigger this deprecation **every** AJAX request we need to make...

packages/adapter/addon/rest.js
```diff
@@ -309,18 +310,47 @@ const RESTAdapter = Adapter.extend(BuildURLMixin, {
...
-    } else if (hasNajax || hasJQuery) {
-      return false;
-    } else {
-      return true;
-    }
+  useFetch: computed({
+    get() {
+      if (this._useFetch !== undefined) {
+        return this._useFetch;
+      }
```
rwjblue, line 317: Can we use a symbol here?

packages/adapter/addon/rest.js
```diff
@@ -309,18 +310,47 @@ const RESTAdapter = Adapter.extend(BuildURLMixin, {
...
+      let ENV = getOwner(this).resolveRegistration('config:environment');
+      // TODO: https://github.com/emberjs/data/issues/6093
+      let jQueryIntegrationDisabled = ENV && ENV.EmberENV && ENV.EmberENV._JQUERY_INTEGRATION === false;
+
+      if (jQueryIntegrationDisabled) {
+        return (this._useFetch = true);
+      } else if (DEPRECATE_NAJAX && typeof najax !== 'undefined') {
+        if (has('fetch')) {
+          deprecate(
+            'You have ember-fetch and jquery installed. To use ember-fetch instead of najax, set `useFetch: true` in your adapter.  In 4.0, ember-data will default to ember-fetch instead of najax when both ember-fetch and jquery are installed in FastBoot.',
```
rwjblue, line 327: Can we tweak this to be "native class" first (e.g. instead of `useFetch: true` say `useFetch = true`)?

packages/adapter/addon/rest.js
```diff
@@ -309,18 +310,47 @@ const RESTAdapter = Adapter.extend(BuildURLMixin, {
...
+            false,
+            {
+              id: 'ember-data:najax-fallback',
+              until: '4.0',
+            }
+          );
+        } else {
+          deprecate(
+            'In 4.0, ember-data will default to ember-fetch instead of najax in FastBoot.  It is recommended that you install ember-fetch or similar fetch polyfill in FastBoot.',
```
rwjblue, line 336: Same here...

packages/adapter/addon/rest.js
```diff
@@ -309,18 +310,47 @@ const RESTAdapter = Adapter.extend(BuildURLMixin, {
...
-    let jQueryIntegrationDisabled = ENV && ENV.EmberENV && ENV.EmberENV._JQUERY_INTEGRATION === false;
-
-    if (jQueryIntegrationDisabled) {
-      return true;
-    } else if (hasNajax || hasJQuery) {
-      return false;
-    } else {
-      return true;
-    }
+  useFetch: computed({
```
rwjblue, line 313: Lets move this down into the `if (DEPRECATE_NAJAX) {` conditional, and here (for all users) we'd say `useFetch: true`.

That way users that opt-in to deprecation stripping of `DEPRECATE_NAJAX` deprecation can reduce the overall payload, and we avoid the work all together...

rwjblue (approved):

PR #7234, opened 2020-06-21
[FEATURE] Pass "fields" through to query params in findRecord and findAll

### LINKED ISSUE #7233: [Proposal]: Pass through `fields` option to findRecord and findAll query parameters
A low impact addition (similar to `include`) to the library would be to pass along sparse fieldsets to the query parameters for a request.

JSONAPI spec: https://jsonapi.org/format/#fetching-sparse-fieldsets

ref https://github.com/emberjs/data/issues/3596

snewcomer:
Current user implementation to support `fields` might include a user adding to adapterOptions but having to explicitly manage adding those query params to the request URL.  Moreover, if a user makes another request with more/less fields, they will have to manage whether or not they should use the cached value or not.

This PR allows a user to pass the `fields` [option](https://jsonapi.org/format/#fetching-sparse-fieldsets) to filter the list of attributes included in an JSONAPI response.  This can help with cases of over-fetching (and reduced payload).

Note - your server needs to be setup to handle parsing and filtering the applicable fields before sending back to the client.  Ember Data will not do any filtering/validations ourselves.

To opt into this feature.

```
export default ApplicationAdapter.extend({
  supportsJSONAPIFields: true,
});
```

If you currently pass `fields` though adapter options, you will get a deprecation warning without this flag.

Here is what the API looks like.

```
store.findRecord('post', 123, { adapterOptions: { fields: { post: 'name,body' } } });
store.findRecord('post', 123, { adapterOptions: { fields: { post: 'name,body', comments: 'title' } }, include: 'comments' });
store.findAll('post', { fields: { adapterOptions: { post: 'name,body' } } });
store.findAll('post', { fields: { adapterOptions: { post: 'name,body', comments: 'title' } }, include: 'comments' });
```

We will not reload the record if -

1. Re-query a record with less fields
2. Re-query a record with same fields 

Will reload the record if -
1. Re-query a record with different fields
2. Re-query a record with more fields

- [x] findRecord
- [ ] findAll
- [ ] others?

close #7233


packages/store/addon/-private/system/snapshot-record-array.ts
```diff
@@ -127,6 +128,14 @@ export default class SnapshotRecordArray {
...
     */
     this.include = options.include;
+
+    /**
+      The specific fields per type to return in a request
+
+      @property fields
+      @type {String}
+    */
+    this.fields = options.fields;
```
snewcomer, line 138: Looks like I need to add documentation to findRecord about fields

packages/-ember-data/tests/unit/adapters/rest-adapter/build-query-test.js
```diff
@@ -27,4 +27,26 @@ module('unit/adapters/rest-adapter/build-query - building queries', function() {
...
+    const snapshotStub = { fields: '[post]=name' };
+
+    const query = adapter.buildQuery(snapshotStub);
+
+    assert.deepEqual(query, { fields: '[post]=name' }, 'query includes `fields`');
+  });
+
+  test('buildQuery() returns query with `fields` and `include` from snapshot', function(assert) {
+    const adapter = DS.RESTAdapter.create();
+    const snapshotStub = { fields: '[post]=name&[comments]=title', include: 'comments' };
```
alechirsch, line 42: In my app, when I pass in fields to the `query` funciton, it is an object like this:
```
{
fields: {
  post: 'name',
  comments: 'title'
}
include: 'comments'
}
```

Can I assume that this will work with `findRecord` as well?

snewcomer: Can can certainly make this work as well!  Good idea

snewcomer: Looks like we properly handle serializing the query params to the url!  Thanks for the comment.  Added a test here -

https://github.com/emberjs/data/pull/7234/commits/79e2507dc9541f8529281ee28dca9526721203cd

packages/adapter/addon/json-api.js
```diff
@@ -221,6 +221,20 @@ const JSONAPIAdapter = RESTAdapter.extend({
...
+    if (snapshot) {
+      let { fields } = snapshot;
+
+      if (fields) {
+        query.fields = fields;
+      }
+    }
+
+    return query;
+  },
```
snewcomer, line 236: TODO - move to adapterOptions and document.

snewcomer: TODO - track fields in shouldReloadRecord,etc and return true/false based on if different than past requests

packages/adapter/addon/json-api.js
```diff
@@ -240,4 +316,58 @@ const JSONAPIAdapter = RESTAdapter.extend({
...
+  let cachedFields = FieldsForRecord.get(record);
+  if (cachedFields && cachedFields.length) {
+    // have seen this record with these fields before - don't fetch
+    if (equalFields(cachedFields, snapshotFields)) {
+      return false;
+    }
+
+    // have seen this record but not these fields - fetch new record
+    cachedFields.push(snapshotFields);
+    FieldsForRecord.set(record, cachedFields);
```
snewcomer, line 362: TODO - associate with returned records and use identifier as a key

snewcomer:
This has been pulled to a separate repo for now.  

https://github.com/snewcomer/ember-data-jsonapi-fields

https://dockyard.com/blog/2020/09/23/introducing-ember-data-jsonapi-fields

snewcomer:
As noted in this blog post, we aren't able to provide a full solution for all use cases.  I'm going to close this and put up a PR with documentation and anything else useful from this PR...

https://dockyard.com/blog/2020/09/23/introducing-ember-data-jsonapi-fields

PR #7237, opened 2020-06-25
Fix broken link

amyrlam:
Relates to: https://github.com/ember-learn/ember-api-docs/pull/726

There seem to be other instances of broken links in this repo with this format:
- Go to https://api.emberjs.com/ember-data/3.10/classes/DS.RESTAdapter and click on the link for `volatile`, which is from this code: https://github.com/emberjs/data/blob/d6bd5f1b82f0ae6424f381f55574a54ea0684aff/packages/adapter/addon/rest.js#L269

I'm not sure how to test this locally with ember-api-docs so any guidance on that would be appreciated!

cc @MelSumner @kategengler from Discord convo

### DIFF
--- a/packages/store/addon/-private/system/promise-proxies.js
+++ b/packages/store/addon/-private/system/promise-proxies.js
@@ -17,7 +17,7 @@ import { Promise } from 'rsvp';
   updated when the promise resolves.
 
   For more information see the [Ember.PromiseProxyMixin
-  documentation](/api/classes/PromiseProxyMixin.html).
+  documentation](/ember/release/classes/PromiseProxyMixin).
 
   Example
 
--- a/packages/store/addon/-private/system/promise-proxies.js
+++ b/packages/store/addon/-private/system/promise-proxies.js
@@ -49,7 +49,7 @@ export const PromiseArray = ArrayProxy.extend(PromiseProxyMixin, {
   be updated when the promise resolves.
 
   For more information see the [Ember.PromiseProxyMixin
-  documentation](/api/classes/PromiseProxyMixin.html).
+  documentation](/ember/release/classes/PromiseProxyMixin.html).
 
   Example
 
--- a/packages/store/addon/-private/system/promise-proxies.js
+++ b/packages/store/addon/-private/system/promise-proxies.js
@@ -17,7 +17,7 @@ import { Promise } from 'rsvp';
   updated when the promise resolves.
 
   For more information see the [Ember.PromiseProxyMixin
-  documentation](/api/classes/Ember.PromiseProxyMixin.html).
+  documentation](/api/classes/PromiseProxyMixin.html).
 
   Example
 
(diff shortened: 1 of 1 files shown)


packages/store/addon/-private/system/promise-proxies.js
```diff
@@ -17,7 +17,7 @@ import { Promise } from 'rsvp';
   updated when the promise resolves.
 
   For more information see the [Ember.PromiseProxyMixin
-  documentation](/api/classes/Ember.PromiseProxyMixin.html).
+  documentation](/api/classes/PromiseProxyMixin.html).
```
amyrlam, line 20: I'm not sure if this link is correct with `.html` as we actually want it to go to https://api.emberjs.com/ember/release/classes/PromiseProxyMixin

Here is the Discord convo: https://discordapp.com/channels/480462759797063690/480777444203429888/725808587343003720

//

FWIW, I also can't search for PromiseProxyMixin, not sure if that's a legit separate issue?

image

igorT:
@Gaurav0 should know. Are there docs for it somewhere?

Gaurav0:
@amyrlam @hjdivad @igorT Docs are in the README of [ember-jsonapi-docs](https://github.com/ember-learn/ember-jsonapi-docs#generate-docs-from-ember-source) repo.

hjdivad:
Thanks for the context @Gaurav0.  Running this locally, these links don't seem to work at all.  The end result looks like it ought to be `/ember/${version}/classes/PromiseProxyMixin` but I'm not seeing a mechanism for version to be inserted or a token replaced.  I suspect these links are quite old and just happened to work.

Given that the links generated from the `@uses` directive work, it's probably better to leave these unlinked rather than broken links.  There's one other occurrence of this style of link in [the REST adapter](https://github.com/emberjs/data/blob/5b2dfaabb4573d10fd8e2c8a71858ae6a2d0276c/packages/adapter/addon/rest.js#L269) and it has the same problem.

@igorT @amyrlam thoughts?

igorT:
@hjdivad @rwjblue mentioned we can use `release` as the version

packages/store/addon/-private/system/promise-proxies.js
rwjblue, line 20: I think we should use:

```suggestion
  documentation](/ember/release/classes/PromiseProxyMixin).
```

Specifically, using `/release/` there will always point to the current Ember release (and avoids the issue that @hjdivad was pointing out).

packages/store/addon/-private/system/promise-proxies.js
```diff
@@ -49,7 +49,7 @@ export const PromiseArray = ArrayProxy.extend(PromiseProxyMixin, {
   be updated when the promise resolves.
 
   For more information see the [Ember.PromiseProxyMixin
-  documentation](/api/classes/Ember.PromiseProxyMixin.html).
+  documentation](/api/classes/PromiseProxyMixin.html).
```
amyrlam, line 52:
```suggestion
  documentation](/ember/release/classes/PromiseProxyMixin.html).
```

I haven't had a chance to test this locally yet but it sounds like from the comments that this will work to map to this link: https://api.emberjs.com/ember/release/classes/PromiseProxyMixin

amyrlam:
Sorry I haven't followed up on this or even looked into the local generation of docs yet. Was this deployed? Doesn't seem to be fixed which is why I'm asking... I can help, just not sure what's going on yet.

To repro: Go to https://api.emberjs.com/ember-data/release/classes/RESTAdapter => click on link for "volatile" => white screen Cannot GET /api/classes/Ember.ComputedProperty.html

PR #7244, opened 2020-07-05
[CHORE]: belongsTo is unset if no relationship data

### LINKED ISSUE #7040: `DS.Snapshot.belongsTo()` does not abide by the documented API
### Description

When an optional `belongsTo` relationship is loaded independently from the model it's defined on, e.g. when invoking `await someRecord.belongsTo('someRel').load()`, and the response is `{ "data": null }` then `DS.Snapshot.belongsTo()` for that relationship returns `undefined` instead of `null`.

Say I have a first record loaded through `store.findRecord('some-model', 1, { include: 'some-rel' })`, intially hydrated from the following JSONAPI document:

```
GET /some-model/1?include=some-rel

{
  data: {
    "type": "some-model",
    "id": 1,
    "attributes": { ... },
    "relationships": {
      "some-rel": {
        "data": null,
        "links": {
          "related": "/some-model/1/some-rel"
        }
      }
    }
  }
}
```

And another one loaded through `store.findRecord('some-model', 2)` (note the lack of include), intially hydrated from the following JSONAPI document:

```
GET /some-model/2

{
  data: {
    "type": "some-model",
    "id": 2,
    "attributes": { ... },
    "relationships": {
      "some-rel": {
        "links": {
          "related": "/some-model/2/some-rel"
        }
      }
    }
  }
}
```

The `someRel` relationship is already loaded on ID=1 (by way of `?include`), however it's not initially loaded on ID=2, but let's say I load that one relationship separately with `await record.belongsTo('someRel').load()`, and let's say the request / response that resulted from that call looks like:

```
GET /some-model/2/some-rel

{
  "data": null
}
```

Suppose now I have a custom serializer for `some-model`:

```javascript
  serialize(snapshot, options) {
    console.log(snapshot.belongsTo('some-rel'));
  },
});
```

Then we can observe the following:

```javascript
someModel1.serialize(); // logs `null`
someModel2.serialize(); // logs `undefined`, expected `null`.
```

The API doc for `DS.Snapshot.belongsTo()` states:

> **returns** *(Snapshot|String|null|undefined)*
> A snapshot or ID of a known relationship or null if the relationship is known but unset. undefined will be returned if the contents of the relationship is unknown.

So I would have expected both calls to return `null`, since both fall under the "known but unset" case.

### Versions

```cli
└─ ember-source@3.12.2
└─ ember-cli@3.14.0
├─ @ember-data/adapter@3.16.0
├─ @ember-data/canary-features@3.16.0
├─ @ember-data/debug@3.16.0
├─ @ember-data/model@3.16.0
├─ @ember-data/private-build-infra@3.16.0
├─ @ember-data/record-data@3.16.0
├─ @ember-data/serializer@3.16.0
├─ @ember-data/store@3.16.0
├─ @types/ember-data@3.1.9
├─ ember-data-copyable@1.1.2
└─ ember-data@3.16.0
```

snewcomer:
close #7040 

@Herriau 👋 I'm unable to reproduce with a test.  How close does this PR look to your original issue?

### DIFF
--- a/packages/-ember-data/tests/integration/references/belongs-to-test.js
+++ b/packages/-ember-data/tests/integration/references/belongs-to-test.js
@@ -492,6 +492,34 @@ module('integration/references/belongs-to', function(hooks) {
     });
   });
 
+  test('load() fetches an unset record', async function(assert) {
+    let store = this.owner.lookup('service:store');
+    let adapter = store.adapterFor('application');
+
+    adapter.findRecord = function(store, type, id, snapshot) {
+      return resolve({ data: null });
+    };
+
+    let person = store.push({
+      data: {
+        type: 'person',
+        id: 1,
+        relationships: {
+          family: {
+            links: {},
+          },
+        },
+      },
+    });
+
+    var familyReference = person.belongsTo('family');
+
+    let record = await familyReference.load();
+    assert.equal(record, null, 'relationship is null');
+
+    assert.equal(person._createSnapshot().belongsTo('family'), null, 'relationship is null off the snapshot');
+  });
+
   test('load() fetches link when remoteType is link', function(assert) {
     var done = assert.async();
 
--- a/packages/-ember-data/tests/integration/snapshot-test.js
+++ b/packages/-ember-data/tests/integration/snapshot-test.js
@@ -338,6 +338,30 @@ module('integration/snapshot - Snapshot', function(hooks) {
     assert.equal(relationship, null, 'relationship is unset');
   });
 
+  test('snapshot.belongsTo() returns null if relationship is unset with no data', function(assert) {
+    assert.expect(1);
+
+    store.push({
+      data: [
+        {
+          type: 'comment',
+          id: '2',
+          attributes: {
+            body: 'This is comment',
+          },
+          relationships: {
+            post: {},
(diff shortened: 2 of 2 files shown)


packages/-ember-data/tests/integration/references/belongs-to-test.js
```diff
@@ -492,6 +492,34 @@ module('integration/references/belongs-to', function(hooks) {
     });
   });
 
+  test('load() fetches an unset record', async function(assert) {
+    let store = this.owner.lookup('service:store');
+    let adapter = store.adapterFor('application');
+
+    adapter.findRecord = function(store, type, id, snapshot) {
```
Herriau, line 499: Shouldn't this be `adapter.findBelongsTo`?

Herriau:
@snewcomer I created a Twiddle that reproduces the issue illustrated in #7040:
https://ember-twiddle.com/e54c989135437db7edba65d2fa440b1e

If you have a console open while loading this twiddle you should see `null` and `undefined` getting logged, instead of `null` and `null`, see the code in `routes/application.js`

runspired:
#7040 seemed to show that this issue was already resolved on master, so closing this

PR #7243, opened 2020-07-03
[INFRA]: test against Node.js 10 and 14

snewcomer:
close #6988

PR #7247, opened 2020-07-15
[CI]: remove erroneous `name` without `run` in node-version-test

snewcomer:
Follow up #7243 

Ref - https://github.com/emberjs/data/actions/runs/170550412

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -261,6 +261,7 @@ jobs:
         run: yarn test:try-one ${{ matrix.scenario }}
 
   node-version-test:
+    name: Use Node.js ${{ matrix.node-version }}
     timeout-minutes: 6
     needs: [lint, basic-tests]
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


.github/workflows/main.yml
```diff
@@ -269,13 +269,12 @@ jobs:
...
     steps:
       - uses: actions/checkout@v1
-      - name: Use Node.js ${{ matrix.node-version }}
       - uses: actions/setup-node@v1
         with:
           node-version: ${{ matrix.node-version }}
       - name: Install dependencies w/o lockfile
         run: yarn install --no-lockfile --non-interactive
-      - name: Basic Tests
+      - name: Basic Tests with Node.js ${{ matrix.node-version }}
```
rwjblue, line 277: You want to add the `name` to the job, not the step

rwjblue: like this https://github.com/ember-cli/ember-cli/blob/21536af8414e70a3e66ab3ae0d64cce24b309f0b/.github/workflows/ci.yml#L49

snewcomer: Oh interesting. Looks like it isn't a pattern we use but certainly a pattern we can start using here.  Fixed.

Issue #5006, opened 2017-06-05
DS.Model.destroyRecord does not unload record from store


kweng2:
#### Scenario:
```
const memberA = store.createRecord('member', { name: 'Bob' })
await memberA.save() // server responds with an id of 5678
await memberA.destroyRecord()
const memberB = store.createRecord('member', { name: 'Bob' })
await memberB.save() // server responds with an id of 5678
```

#### Expectation:
This should work no problem, memberA and memberB should be different records, that have the same 'name'

#### Reality:
```
You cannot update the id index of an InternalModel once set. Attempted to update ${id}.
```

I believe what's happening is that ```destroyRecord``` is not unloading the internalModel, and therefore upon creation throwing this error

On closer inspection, it does seem that ```store._removeFromIdMap``` and ```internalModel.destroy``` only get called from the ```unload``` method

It is unclear from the docs whether destroyRecord also removes internal references

acorncom:
Related issue: https://github.com/emberjs/data/issues/4972

pangratz:
I think this is fixed by #5126, so I am closing.

@kweng2 please reopen if you are still running into this issue after upgrading. Thanks!

SergeAstapov:
This is still an issue in ember-data@2.16.3

I just hit exactly the same problem with exactly the same scenario as described in the first post.
I've tried to do
```
return coupon.destroyRecord().then(() => {
    this.get('store').unloadRecord(coupon);
});
```

But still got error `Error: Assertion Failed: 'coupon' was saved to the server, but the response returned the new id 'XXX', which has already been used with another record.'`

maysam-tayyeb:
I checked the unloadRecod() and realised it does not unload destroyed record and therefore I unloaded my record using:

erichonkanen:
related https://github.com/emberjs/data/issues/5237

runspired:
This should be re-examined following the landing of #5378

runspired:
#5455 is in-progress to address this; however, semantics and teardown timing issues around getting `deleteRecord` and `deleteRecord.save` correct on `isNew` records has held it up.

PR #7258, opened 2020-07-31
feat: record.destroyRecord should also unload the record

snewcomer:
close https://github.com/emberjs/data/issues/5006
close https://github.com/emberjs/data/pull/5455

Port of changes from #7217

### DIFF
--- a/packages/model/addon/-private/attr.js
+++ b/packages/model/addon/-private/attr.js
@@ -152,14 +152,13 @@ function attr(type, options) {
         }
       }
       if (RECORD_DATA_ERRORS) {
-        let oldValue = this._internalModel._recordData.getAttr(key);
-        if (oldValue !== value) {
-          let errors = this.get('errors');
-          if (errors.get(key)) {
-            errors.remove(key);
-            this.___recordState.cleanErrorRequests();
-            if (errors.isEmpty) {
-              this._internalModel.send('becameValid');
+        if (!this.isValid) {
+          let oldValue = this._internalModel._recordData.getAttr(key);
+          if (oldValue !== value) {
+            const { errors } = this;
+            if (errors.get(key)) {
+              errors.remove(key);
+              this.___recordState.cleanErrorRequests();
             }
           }
         }
--- a/packages/model/addon/-private/attr.js
+++ b/packages/model/addon/-private/attr.js
@@ -157,8 +157,11 @@ function attr(type, options) {
           let errors = this.get('errors');
           if (errors.get(key)) {
             errors.remove(key);
+            this.___recordState.cleanErrorRequests();
+            if (errors.isEmpty) {
+              this._internalModel.send('becameValid');
+            }
           }
-          this.___recordState.cleanErrorRequests();
         }
       }
       return this._internalModel.setDirtyAttribute(key, value);
--- a/packages/model/addon/-private/attr.js
+++ b/packages/model/addon/-private/attr.js
@@ -158,7 +158,7 @@ function attr(type, options) {
           if (errors.get(key)) {
             errors.remove(key);
           }
-          this._markInvalidRequestAsClean();
+          this.___recordState.cleanErrorRequests();
         }
       }
       return this._internalModel.setDirtyAttribute(key, value);
--- a/packages/model/addon/-private/model.js
+++ b/packages/model/addon/-private/model.js
@@ -4,63 +4,37 @@
 
 import { assert, deprecate, warn } from '@ember/debug';
 import EmberError from '@ember/error';
-import EmberObject, { computed, get } from '@ember/object';
+import EmberObject, { get } from '@ember/object';
 import { dependentKeyCompat } from '@ember/object/compat';
+import { run } from '@ember/runloop';
 import { isNone } from '@ember/utils';
 import { DEBUG } from '@glimmer/env';
+import { tracked } from '@glimmer/tracking';
 import Ember from 'ember';
 
-import {
-  CUSTOM_MODEL_CLASS,
-  RECORD_DATA_ERRORS,
-  RECORD_DATA_STATE,
-  REQUEST_SERVICE,
-} from '@ember-data/canary-features';
+import { CUSTOM_MODEL_CLASS, REQUEST_SERVICE } from '@ember-data/canary-features';
 import { HAS_DEBUG_PACKAGE } from '@ember-data/private-build-infra';
 import {
   DEPRECATE_EVENTED_API_USAGE,
   DEPRECATE_MODEL_TOJSON,
   DEPRECATE_RECORD_LIFECYCLE_EVENT_METHODS,
 } from '@ember-data/private-build-infra/deprecations';
-import {
-  coerceId,
-  DeprecatedEvented,
-  errorsArrayToHash,
-  InternalModel,
-  PromiseObject,
-  recordDataFor,
-  recordIdentifierFor,
-} from '@ember-data/store/-private';
+import { coerceId, DeprecatedEvented, InternalModel, PromiseObject } from '@ember-data/store/-private';
 
 import Errors from './errors';
+import RecordState, { peekTag, tagged } from './record-state';
 import { relationshipFromMeta } from './system/relationships/relationship-meta';
-
-const RecordMeta = new WeakMap();
-function getRecordMeta(record) {
-  let meta = RecordMeta.get(record);
-  if (meta === undefined) {
-    meta = Object.create(null);
-    RecordMeta.set(record, meta);
-    if (DEBUG) {
-      record.____recordMetaCache = meta;
-    }
-  }
-  return meta;
-}
-
-function getWithDefault(meta, prop, value) {
-  let v = meta[prop];
-  if (v === undefined) {
-    return value;
-  }
-  return v;
-}
 
 const { changeProperties } = Ember;
 
-function isInvalidError(error) {
-  return error && error.isAdapterError === true && error.code === 'InvalidError';
-}
+/*
+ In the non-CUSTOM_MODEL_CLASS world things decorated with flagToggledDecorator
+ have logic that requires a cache and caching key (which tagged provides), but
+ in the CUSTOM_MODEL_CLASS branch it only requires dependentKeyCompat which
+ ensures computeds/observers on the field will fire when anything it accesses
+ that is tracked changes
+*/
+const flagToggledDecorator = CUSTOM_MODEL_CLASS ? dependentKeyCompat : tagged;
 
 function findPossibleInverses(type, inverseType, name, relationshipsSoFar) {
   let possibleRelationships = relationshipsSoFar || [];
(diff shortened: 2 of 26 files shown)


packages/-ember-data/tests/acceptance/relationships/belongs-to-test.js
```diff
@@ -278,11 +278,13 @@ module('async belongs-to rendering tests', function(hooks) {
       petOwner = await tweety.get('petOwner');
       assert.equal(petOwner.get('name'), 'Pete');
 
+      let goofyInternalModel = goofy._internalModel;
       await goofy.destroyRecord();
-      assert.ok(goofy.isDeleted);
+      assert.ok(goofyInternalModel.isDeleted, 'record is deleted');
```
snewcomer, line 283: These changes were required since the original model does not have that property anymore.  The blocking PR helps get us closer to ensuring users can still call `goofy.isDeleted` and returning `true`.

https://github.com/emberjs/data/pull/7226#issuecomment-643552364

runspired: I suspect this was actually an issue with the `DSModel` types that I've fixed in #7471 (but which would also have been improved by the native model class refactor since we could start using it for type information directly). We can probably resume work here, though the native model refactor is close!

runspired: This is also improved in #7475

runspired:
- Won't fix #6147 as that requires new relationship layer
- Should bring in tests from #5480 if it doesn't.
- may fix #5395 should audit that test
- should bring in test for the scenario described in #6600

packages/model/addon/-private/model.js
```diff
@@ -2,8 +2,10 @@ import { assert, deprecate, warn } from '@ember/debug';
 import EmberError from '@ember/error';
 import EmberObject, { computed, get } from '@ember/object';
 import { dependentKeyCompat } from '@ember/object/compat';
+import { run } from '@ember/runloop';
```
runspired, line 5: we need to specify this import in the external dependencies list in packages/model/index.js

jrjohnson:
We addressed the first failure  in `Ilios-frontend` and merged that today, we were relying on the buggy destroyed record not unloading behavior.

 The other test failure seems to be a change in the way `isNew` is set or tracked. We have a [getter](https://github.com/ilios/frontend/blob/master/app/components/school-vocabulary-manager.js#L21-L31) that filters out `isNew` models. Once terms were persisted this getter would re-fire and the newly added term wouldn't be `isNew` anymore. Now the getter doesn't fire again after `isNew` gets set to false (or something else is happening I'm not seeing).

I'm not sure why we're looking at `isNew` here, I'm guessing it is cruft from older async patterns on our end and I'll probably just remove it, but I wanted to mention it here in case this is a bug in this change.

runspired:
@jrjohnson I'd love to investigate that with you before you remove it, we may need to fix a notify somewhere.

runspired (approved):
Overall approved but we need to fix M3 (involves just going back to setting currentState on the model in the FF off branch) / potential issue with ilios (potential lack of notification on isNew)

packages/model/addon/-private/record-state.ts
```diff
@@ -0,0 +1,391 @@
...
+  declare rev: number;
+  declare isDirty: boolean;
+  declare value: any;
+
+  constructor() {
+    this.rev = 1;
+    this.isDirty = true;
+    this.value = undefined;
+  }
+  @tracked _tracking = 0;
```
snewcomer, line 27: Do we need a tracked property for this data structure?  I assumed I would find a getter somewhere that is consuming it.  It seems most of the semantics here are preventing running the tagged property from running/not running if `isDirty`, which is accomplished through an explicit `notify`.  If I am missing something, what is `_tracking` forcing to recompute?

runspired: It's consumed by subscribe which we call on all accesses to a property decorated with @tagged. This consumption is what tells glimmer to recompute once we've decided to notify.

snewcomer: Oh right.  This is the first time I have seen a function call to setup tracking.  The tracking system must be able to "see through" that function call and recompute the original accessor when `_tracking` is updated

packages/model/addon/-private/model.js
```diff
@@ -29,38 +26,15 @@ import {
...
-  }
-  return v;
-}
-
 const { changeProperties } = Ember;
 
-function isInvalidError(error) {
-  return error && error.isAdapterError === true && error.code === 'InvalidError';
-}
+const flagToggledDecorator = CUSTOM_MODEL_CLASS ? dependentKeyCompat : tagged;
```
igorT, line 37: Can you please add a comment describing the goal/purpose of this.

packages/model/package.json
```diff
@@ -27,6 +27,7 @@
     "ember-cli-string-utils": "^1.1.0",
     "ember-cli-test-info": "^1.0.0",
     "ember-cli-typescript": "^4.1.0",
+    "ember-cache-primitive-polyfill": "^1.0.1",
```
igorT, line 30: Do we need this if implementing ourselves?

packages/model/addon/-private/model.js
runspired: sure! the TL;DR is that in the non-CUSTOM_MODEL_CLASS world things decorated with `flagToggledDecorator` have logic that requires a cache and caching key (which tagged provides), but in the CUSTOM_MODEL_CLASS branch it only requires `dependentKeyCompat` which ensures computeds/observers on the field will fire when anything it accesses that is tracked changes. I'll add this to the PR.

packages/record-data/addon/-private/graph/index.ts
```diff
@@ -338,7 +338,7 @@ export class Graph {
 
   willDestroy() {
     this.identifiers.clear();
-    this.store = (null as unknown) as RecordDataStoreWrapper;
+    this.store = null as unknown as RecordDataStoreWrapper;
```
igorT, line 341: Do you find this to be more readable?

packages/model/addon/-private/model.js
```diff
@@ -821,6 +717,19 @@ class Model extends EmberObject {
     return this._internalModel.transitionTo(name);
   }
 
+  notifyPropertyChange(key) {
```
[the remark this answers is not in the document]
runspired: sure. We're doing this because we manage tags for many properties, this is how we ensure those tags are notified when appropriate, not only by our own code but by external code as well.

packages/record-data/addon/-private/graph/index.ts
runspired: I don't find it better one way or the other, its just what prettier does now.

packages/model/package.json
runspired: Yes, it's what polyfills the two helpers that we import for it for 3.20 (they weren't in ember until 3.22)

packages/store/addon/-private/system/model/internal-model.ts
```diff
@@ -435,24 +418,41 @@ export default class InternalModel {
...
   }
 
   deleteRecord() {
-    if (RECORD_DATA_STATE) {
-      if (this._recordData.setIsDeleted) {
-        this._recordData.setIsDeleted(true);
-      }
-    }
-    this.send('deleteRecord');
+    run(() => {
```
igorT, line 430: Why the double runloop?

packages/store/addon/-private/system/model/internal-model.ts
```diff
@@ -944,7 +917,7 @@ export default class InternalModel {
       if (CUSTOM_MODEL_CLASS) {
         this.store._notificationManager.notify(this.identifier, 'relationships');
       } else {
-        this._record.notifyHasManyAdded(key);
+        this._record.notifyPropertyChange(key);
```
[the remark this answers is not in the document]
runspired: it's a similar to change to the other PRs, same effect, once rebased the change will only have occurred once. Basically just ensures we shunt the notification through notifyPropertyChange instead of the misdirection.

packages/store/addon/-private/system/model/internal-model.ts
runspired: outer one is Ember's which ensure the destroy queue will be completed, inner one is ours which ensures the relationship sweep will be completed. Both are necessary to ensure that deleteRecord completes this state updates synchronously.

packages/model/addon/-private/attr.js
```diff
@@ -158,7 +158,7 @@ function attr(type, options) {
           if (errors.get(key)) {
             errors.remove(key);
           }
-          this._markInvalidRequestAsClean();
+          this.___recordState.cleanErrorRequests();
```
igorT, line 161: This used to clear out the invalid requests, but now clears out the error requests?

packages/model/addon/-private/model.js
```diff
@@ -579,6 +468,22 @@ class Model extends EmberObject {
     @private
     @type {Object}
   */
+  @tagged
+  get currentState() {
+    if (CUSTOM_MODEL_CLASS) {
+      return this.___recordState;
+    }
+    if (this.isDestroyed || this.isDestroying) {
```
igorT, line 476: Curious, why do we need this?

packages/model/addon/-private/record-state.ts
```diff
@@ -0,0 +1,392 @@
...
+type RequestCache = import('@ember-data/store/-private/system/request-cache').default;
+type Model = InstanceType<typeof import('./model')>;
+type StableRecordIdentifier = import('@ember-data/store/-private/ts-interfaces/identifier').StableRecordIdentifier;
+type CoreStore = import('@ember-data/store/-private/system/core-store').default;
+
+function isInvalidError(error) {
+  return error && error.isAdapterError === true && error.code === 'InvalidError';
+}
+
+class Tag {
```
igorT, line 18: Can you please write up a comment describing how/why this is used. Tag is a bit too generic of a thing

packages/model/addon/-private/record-state.ts
```diff
@@ -0,0 +1,392 @@
...
+            }
+            this.notify('isLoading');
+            this.notify('isError');
+            this.notify('adapterError');
+            this.notify('isValid');
+            break;
+          case 'fulfilled':
+            this.pendingCount--;
+            this.fulfilledCount++;
+            this.notify('isLoading');
```
igorT, line 188: Can we extract these notifications into a helper method so we don't accidentally forget to call one.

packages/model/addon/-private/record-state.ts
```diff
@@ -0,0 +1,392 @@
...
+        case 'errors':
+          this.notify('isValid');
+          break;
+      }
+    });
+  }
+
+  notify(key) {
+    getTag(this, key).notify();
+    // this.record.notifyPropertyChange(key);
```
igorT, line 225: Whats up with this?

packages/store/addon/-private/system/model/internal-model.ts
igorT: But delete record doesn't cause any destruction? Wouldn't this be fully encompassed by the call to `unloadRecord`?

packages/model/addon/-private/model.js
runspired: In the non CUSTOM_MODEL_CLASS case this is necessary for us to be able to hand the previous state to the user once the record is unloaded. This fixes an issue I noticed after the refactor to native model but was made evident by this PR. When `currentState` was a computed property it would return the cached value once the record was destroyed and not recompute.

packages/model/addon/-private/record-state.ts
```diff
@@ -0,0 +1,392 @@
...
+    notifications.subscribe(id, (identifier, type) => {
+      notifyChanges(identifier, type, record, store);
+      switch (type) {
+        case 'state':
+          this.notify('isNew');
+          this.notify('isDeleted');
+          this.notify('isDirty');
+          break;
+        case 'attributes':
+          this.notify('isEmpty');
```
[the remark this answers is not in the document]
runspired: if you've received data you would no longer be empty

packages/store/addon/-private/system/model/internal-model.ts
runspired: @igorT it does when a new record is deleted.

packages/model/addon/-private/record-state.ts
runspired: deletable

igorT:
Sure but this method calls unload which is how you end up with a full
deletion. No need to have the outer wrapper here, as the call to unload
record should take care of it

PR #7262, opened 2020-08-02
[CHORE]: Extract internalModel access to identifiers for Relationships

snewcomer:
ref #6166 
past work https://github.com/emberjs/data/pull/6715


packages/store/addon/-private/system/core-store.ts
```diff
@@ -1580,7 +1580,7 @@ abstract class CoreStore extends Service {
       typeof modelName === 'string'
     );
 
-    const resource = constructResource(modelName, ensureStringId(id));
+    const resource: ResourceIdentifierObject = constructResource(modelName, ensureStringId(id));
```
snewcomer, line 1583: Do we like to type the return value?  If no value, I can remove

packages/store/addon/-private/system/references/belongs-to.js
```diff
@@ -22,13 +23,18 @@ import Reference, { INTERNAL_MODELS } from './reference';
...
     this.type = belongsToRelationship.relationshipMeta.type;
-    this.parent = parentInternalModel.recordReference;
-    this.parentInternalModel = parentInternalModel;
+    if (RECORD_ARRAY_MANAGER_IDENTIFIERS) {
+      this.parent = internalModelFactoryFor(store).peek(parentIMOrIdentifier).recordReference;
+      this.parentInternalModel = internalModelFactoryFor(store).peek(parentIMOrIdentifier);
+    } else {
+      this.parent = parentIMOrIdentifier.recordReference;
+      this.parentInternalModel = parentIMOrIdentifier;
+    }
```
snewcomer, line 37: If the goal is to remove references to the internal model property as well, I can make this `this.parentIdentifier` and lookup by identifier in each corresponding method.

packages/store/addon/-private/system/references/reference.ts
```diff
@@ -31,6 +34,10 @@ function isResourceIdentiferWithRelatedLinks(
 
 export const INTERNAL_MODELS = new WeakMap<Reference, InternalModel>();
```
snewcomer, line 35: Sort of along the same lines as the other question - The API changed for creating References from InternalModel to ids.  However, do we want to also change our caching layer as well with this refactor?  Or does this seem right to keep as is (various places in belongs-to/has-many reference IM specific properties)?

packages/store/addon/-private/system/references/reference.ts
```diff
@@ -42,8 +49,15 @@ interface Reference {
...
-  constructor(public store: CoreStore, internalModel: InternalModel) {
-    INTERNAL_MODELS.set(this, internalModel);
+  constructor(public store: CoreStore, identifierOrInternalModel: InternalModel | StableRecordIdentifier) {
+    if (RECORD_ARRAY_MANAGER_IDENTIFIERS) {
+      let internalModel = internalModelFactoryFor(this.store).peek(identifierOrInternalModel as StableRecordIdentifier);
+      if (internalModel) {
+        INTERNAL_MODELS.set(this, internalModel);
+      }
+    } else {
+      INTERNAL_MODELS.set(this, identifierOrInternalModel as InternalModel);
```
snewcomer, line 59: Are we ok coercing variables to a specific type?

packages/store/addon/-private/system/core-store.ts
igorT: If it's not super clear, it's :+1:

packages/store/addon/-private/system/references/belongs-to.js
igorT: If we have a path of fully getting away from internal model, we probably want to bundle this up together, landing just the lookup part might be bad for perf in hot paths

packages/store/addon/-private/system/references/reference.ts
igorT: Replacing the caching layer is 👍

packages/store/addon/-private/system/references/reference.ts
igorT: Yeah, usually a type guard would be great, but we don't have one in this case.

packages/store/addon/-private/ts-interfaces/ember-data-json-api.ts
```diff
@@ -33,7 +33,6 @@ export interface ExistingResourceIdentifierObject {
   type: string;
 
   /**
```
igorT, line 35: This seems like an accident

packages/-ember-data/tests/integration/references/belongs-to-test.js
```diff
@@ -144,8 +144,8 @@ module('integration/references/belongs-to', function(hooks) {
     var personReference = store.getReference('person', 1);
     var familyReference = person.belongsTo('family');
 
-    assert.ok(personReference);
-    assert.equal(familyReference.parent, personReference);
+    assert.ok(personReference, 'person reference is present');
```
igorT, line 147: ❤️

packages/store/addon/-private/system/core-store.ts
```diff
@@ -1436,9 +1439,20 @@ abstract class CoreStore extends Service {
     }
     const type = normalizeModelName(modelName);
     const normalizedId = ensureStringId(id);
-    const resource = constructResource(type, normalizedId);
+    const resource: ResourceIdentifierObject = constructResource(type, normalizedId);
```
igorT, line 1442: this type/id should be a recordIdentifer not a resource

packages/store/addon/-private/system/core-store.ts
```diff
@@ -1436,9 +1439,20 @@ abstract class CoreStore extends Service {
...
+
+    if (RECORD_ARRAY_MANAGER_IDENTIFIERS) {
+      const identifier: StableRecordIdentifier = identifierCacheFor(this).getOrCreateRecordIdentifier(resource);
```
igorT, line 1445: We shouldn't `getOrCreateRecordIdentifier` here

packages/store/addon/-private/system/references/belongs-to.js
```diff
@@ -82,7 +88,7 @@ export default class BelongsToReference extends Reference {
   }
 
   _resource() {
-    return INTERNAL_MODELS.get(this)?._recordData.getBelongsTo(this.key);
+    return internalModelForReference(this)?._recordData.getBelongsTo(this.key);
```
igorT, line 91: You should be able to get the record data if you have the identifier right?

packages/store/addon/-private/system/references/belongs-to.js
```diff
@@ -209,10 +215,9 @@ export default class BelongsToReference extends Reference {
    @return {Model} the record in this relationship
    */
   value() {
-    let store = this.parentInternalModel.store;
     let resource = this._resource();
     if (resource && resource.data) {
-      let inverseInternalModel = store._internalModelForResource(resource.data);
+      let inverseInternalModel = this.store._internalModelForResource(resource.data);
```
igorT, line 220: 👍

packages/store/addon/-private/system/references/has-many.js
```diff
@@ -19,19 +21,23 @@ import Reference, { INTERNAL_MODELS } from './reference';
...
+    } else {
+      this.parent = parentIMOrIdentifier.recordReference;
+    }
 
     // TODO inverse
   }
 
   _resource() {
-    return INTERNAL_MODELS.get(this)?._recordData.getHasMany(this.key);
+    return internalModelForReference(this)?._recordData.getHasMany(this.key);
```
igorT, line 40: Should be able to get recordData via reference

packages/store/addon/-private/system/references/reference.ts
```diff
@@ -29,7 +32,15 @@ function isResourceIdentiferWithRelatedLinks(
   return value && value.links && value.links.related;
 }
 
-export const INTERNAL_MODELS = new WeakMap<Reference, InternalModel>();
+export const REFERENCE_CACHE = new WeakMap<Reference, InternalModel | StableRecordIdentifier>();
```
igorT, line 35: Once RECORD_ARRAY_MANAGER_IDENTIFIERS are turned on, lets make sure to simplify this, maybe leave a todo comment and rename to be more descriptive

packages/store/addon/-private/system/references/reference.ts
```diff
@@ -29,7 +32,15 @@ function isResourceIdentiferWithRelatedLinks(
...
+
+export function internalModelForReference(reference: Reference): InternalModel | null | undefined {
+  if (RECORD_ARRAY_MANAGER_IDENTIFIERS) {
+    return internalModelFactoryFor(reference.store).peek(REFERENCE_CACHE.get(reference) as StableRecordIdentifier);
```
igorT, line 39: I am fine with the messy TS as we are turning this FF very soon

packages/store/addon/-private/system/references/reference.ts
```diff
@@ -42,8 +53,12 @@ interface Reference {
 }
 abstract class Reference {
   public recordData: InternalModel['_recordData'];
-  constructor(public store: CoreStore, internalModel: InternalModel) {
-    INTERNAL_MODELS.set(this, internalModel);
+  constructor(public store: CoreStore, identifierOrInternalModel: InternalModel | StableRecordIdentifier) {
+    if (RECORD_ARRAY_MANAGER_IDENTIFIERS) {
+      REFERENCE_CACHE.set(this, identifierOrInternalModel as StableRecordIdentifier);
```
igorT, line 58: I don't think this needs a cast

packages/store/addon/-private/system/core-store.ts
snewcomer: Sounds like we need to adjust the return type of `constructResource` to `ResourceIdentifierObject | RecordIdentifier`?

packages/store/addon/-private/system/references/record.ts
```diff
@@ -17,12 +20,26 @@ type RecordInstance = import('../../ts-interfaces/record-instance').RecordInstan
...
+    if (RECORD_ARRAY_MANAGER_IDENTIFIERS) {
+      let identifier = REFERENCE_CACHE.get(this) as StableRecordIdentifier;
+      if (identifier) {
+        return identifier.id;
+      }
+
+      return null;
+    } else {
+      return internalModelForReference(this)!.id;
+    }
```
snewcomer, line 42: From meeting - an `identifier` getter probably makes sense.  Will discuss in next week's meeting.

packages/store/addon/-private/system/references/has-many.js
snewcomer: Are you suggesting we make a public getter for `recordData` on the Record instance?  

If not, I was assuming here that the only way to get `recordData` was through the InternalModel.  Perhaps there is a better way?

packages/store/addon/-private/system/references/reference.ts
```diff
@@ -61,6 +61,10 @@ abstract class Reference {
     }
   }
 
+  get recordData() {
+    return internalModelForReference(this)?._recordData;
```
igorT, line 65: Can you get the identifier here as we for when we do have it and use it to lookup the record data

snewcomer: Done!

packages/store/addon/-private/system/normalize-model-name.ts
```diff
@@ -16,6 +16,6 @@ import { dasherize } from '@ember/string';
   @param {String} modelName
   @return {String} normalizedModelName
 */
-export default function normalizeModelName(modelName) {
+export default function normalizeModelName(modelName: string): string {
```
igorT, line 19: ❤️

igorT (approved):

sandstrom:
@snewcomer Have you thought about registering for Github Sponsorships?

PR #7263, opened 2020-08-06
[CHORE]: Move symbols out of /ts-interfaces and into -private/utils

snewcomer:
Saw this comment about moving the symbol util.  So just did a quick move out from `-private/ts-interfaces/utils` to `-private/utils`.  I think we still have a bit to go (early 2021) until we drop support for IE11 so there may be some value in this minor refactor.


igorT:
Thank you!

Issue #7265, opened 2020-08-11
Self-Referent Relationship in Save Response Blows Up Save


davelindquist-egistix:
### Reproduction

See the Ember Twiddle at:

https://ember-twiddle.com/5fd0892440b64337440c13d4317ed295

The component provides a 'Save' button and instructions (clicking the 'Save' will immediately cause the error).

### Description

We have a case where we:
- Create a new record
- Fill in a few fields
- Save
- The backend returns the record, with a relationship filled in
- This relationship is self-referential (has its own id)
- Ember Data blows up, complaining that the id is already in the store.

For example, our model looks something like this:

```js
export default class extends Model {
  @attr("string") name
  @belongsTo('company', {inverse: null} ) parent
}
```

and the save returns data something like this:

```js
  data: {
    type: "company",
    id: 1,
    attributes: {
      "name": "Test Company"
    },
    relationships: {
      parent: {
        data: {
          type: "company",
          id: 1
        }
      }
    }
  }
```

The Ember Data sequence of events appears to be:
1. Parse the relationships
2. Create a 'placeholder' company model for the relationship with id of "1", and put it into the map
3. Parse the object itself
4. Check to see if that id is already used (and is not myself) - ```var existingInternalModel = this.peekIdOnly(modelName, id);```
5. Blow up, because it found the "id 1" item created for the relationship, which is not 'me'.

### Versions

Since we do not use yarn, we cannot run the indicated commands to get versions - however, here is what package.json/package-lock.json has to say about the packages of interest:

```cli
    "ember-cli": "3.12.1",
    "ember-source": "3.12.3",
    "ember-data": "3.12.6",
```

runspired:
This can be resolved in your app by reflecting the record identifier's `lid` back in the payload at each location this particular record identity (type id pair) appears. I'd suggest reading the RFC for documentation (it was fully implemented) or the test suite which covers this scenario.

runspired:
Going to reopen this since the proposed fix was simple enough it seems ok to add so that fewer folks need to worry about adding lid for a case we can potentially trivially cover.

runspired:
Closing since this works on master without the need for reflecting the lid back.

PR #7266, opened 2020-08-11
[BUGFIX] New test and possible fix for issue #7265 (self-referent rel)

davelindquist-egistix:
This addresses issue #7265 (Self-Referent Relationship in Save Response Blows Up Save).

It includes a new test that replicates the problem - saving a record, and having the 'backend' return a relationship that is self-referent, thus blowing up Ember Data when it checks to see if the just-saved id has already been used.  (Said id was added to the store as part of the relationships-handling.)

It also includes a patch to record-data.ts that solves the problem in the test, and does not cause any other tests to fail.  HOWEVER, it does this by simply re-ordering the handling of relationships and the 'set the record id' call - I don't know if there is a reason that they were done in the opposite order originally, and thus don't know for certain if this will break something else.  (I'm hoping the tests would have shown that, but perhaps wiser eyes can take a look to make certain.)

Thanks!

### DIFF
--- a/packages/-ember-data/tests/integration/relationships/belongs-to-test.js
+++ b/packages/-ember-data/tests/integration/relationships/belongs-to-test.js
@@ -22,6 +22,13 @@ module('integration/relationship/belongs-to BelongsTo Relationships (new-style)'
   let store;
   setupTest(hooks);
 
+  class Company extends Model {
+    @belongsTo('company', { inverse: null, async: true })
+    parentCompany;
+    @attr()
+    name;
+  };
+
   class Person extends Model {
     @belongsTo('pet', { inverse: 'bestHuman', async: true })
     bestDog;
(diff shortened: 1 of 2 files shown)


davelindquist-egistix:
Hmm, it appears that Travis runs additional tests beyond what happens with the tests locally, and these have failed.  I will close this PR, and resubmit one with only the test in it.

runspired:
Did you ever resubmit? The fix here is simple enough I'd be ok with it since it slightly improves the situation (most of the time this sort of case doesn't have such an easy way to address it in the internals). I'd be happy to look at whatever failed and see if this could work.

runspired:
Confirmed this test passes against master so closing, but I like the ordering change and test since they make more sense anyway so am porting these to a new PR.

PR #7267, opened 2020-08-11
[TEST] Added test for issue #7265 (Self-Referent Relationship)

davelindquist-egistix:
Tests the case where a save-response returns a relationship that includes the self-id.

This passes in master, but is broken in 3.12 LTS - not sure where the 'fix' for this happened along the way?

Can we backport the fix to the 3.12 LTS version?

### DIFF
--- a/packages/-ember-data/tests/integration/relationships/belongs-to-test.js
+++ b/packages/-ember-data/tests/integration/relationships/belongs-to-test.js
@@ -26,6 +26,13 @@ module('integration/relationship/belongs-to BelongsTo Relationships (new-style)'
   let store;
   setupTest(hooks);
 
+  class Company extends Model {
+    @belongsTo('company', { inverse: null, async: true })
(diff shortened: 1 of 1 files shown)


igorT:
Hi @davelindquist-egistix, we usually only support the latest LTS, can you please check whether this fails with 3.16?

davelindquist-egistix:
Hmm, I must be misunderstanding the meaning of "Long Term Support".

Anyway, I did a quick check by updating my Ember Twiddle to 3.16, and it does not appear to occur in that version.

igorT:
Hi @davelindquist-egistix, sorry about the confusion :(

Long Term Support, is long in a sense that it last multiple regular releases, but once there is a newer LTS, we don't indefinitely support the older ones, that wouldn't really be sustainable. You can check the exact versions and timelines at https://emberjs.com/releases/lts/

PR #7269, opened 2020-08-14
Update reference.ts

chrism:
This is to reduce potential confusion over how Ember Data defines the link as using "self" does not work.

https://discordapp.com/channels/480462759797063690/486549196837486592/743781525493121054


snewcomer (approved):
@chrisgame Thanks!  Mind merging in master so we get the tests passing?

PR #7272, opened 2020-08-19
Feature: Allow fetch to accept credentials

kookyungmin:
I want to use option credentials : 'include', but I cannot use it because 'same-origin' is fixed.

### DIFF
--- a/packages/adapter/addon/rest.js
+++ b/packages/adapter/addon/rest.js
@@ -1366,7 +1366,7 @@ function headersToObject(headers) {
  * @returns {Object}
  */
 export function fetchOptions(options, adapter) {
-  options.credentials = 'same-origin';
+  options.credentials = options.credentials || 'same-origin';
 
   if (options.data) {
     // GET and HEAD requests can't have a `body`


packages/adapter/addon/rest.js
```diff
@@ -1366,7 +1366,7 @@ function headersToObject(headers) {
  * @returns {Object}
  */
 export function fetchOptions(options, adapter) {
-  options.credentials = 'same-origin';
+  options.credentials = options.credentials || 'same-origin';
```
snewcomer, line 1369: 👋 Possible we can add a test for this?  Otherwise, this looks great!

kookyungmin: I will modify the source so that only the existing tests can be passed and used.

kookyungmin: I didn't pass all of your tests 😥
How is there no way?

snewcomer: @kookyungmin Mind merging in master?  Tests should pass now.

kookyungmin: Thank you. Let me check.

PR #7273, opened 2020-08-19
[BUGFIX] Entangle Errors.errorsFor properly

pzuraq:
We recently made some changes to the way arrays get consumed when
autotracking in Ember - specifically, whenever an array is return from
`Ember.get`, `@computed`, or `@tracked`, we track that array for changes
in the future. Previously, we also tracked it if it was an _EmberArray_.
This check was a bit redundant, and costly, so we removed it and now
only do it for native arrays.

The reason this is generally safe is that non-native EmberArrays usually
have a contract - you must access the state of the array via `objectAt`,
and you must update via `pushObject`, `popObject`, etc. The array is
consumed when using `objectAt`, so theres no need to do it eagerly in
`Ember.get` (and penalize all other object access).

The `Errors` class is a bit of a special case, because it adds additonal
capabilities - the `errorsFor` method, which returns a subset of errors
in the array. This info is accessed via the `errorsFor()` method, which
does not consume the array proxy at all. So, previously, if a user did
something like:

```js
model.errors.errorsFor(field).map(error => error.message);
```

When `model.errors` was accessed, it was a `@computed` property, and it
consumed the entire array proxy. Later on, when the user goes to add an
error, it adds the error to the main array _and_ the subarray returned
by `errorsFor()`, and dirties both, so the user's code is called again
and updates properly.

With the recent changes, that doesn't happen anymore. Since `errorsFor`
returns a native array (with prototype extensions enabled), `map` does
not use `objectAt` internally either. So, there is nothing that is
properly consuming the array.

This PR adds the consumption manually. In the future, we can likely add
a primitive that allows Errors to create and manually manage tags, or it
can use something along the lines of a TrackedArray for these errors
instead.

### DIFF
--- a/packages/model/addon/-private/errors.js
+++ b/packages/model/addon/-private/errors.js
@@ -127,11 +127,20 @@ export default ArrayProxy.extend(DeprecatedEvented, {
   errorsFor(attribute) {
     let map = get(this, 'errorsByAttributeName');
 
-    if (!map.has(attribute)) {
-      map.set(attribute, A());
+    let errors = map.get(attribute);
+
+    if (errors === undefined) {
+      errors = A();
+      map.set(attribute, errors);
     }
 
-    return map.get(attribute);
+    // Errors may be a native array with extensions turned on. Since we access
+    // the array via a method, and not a computed or using `Ember.get`, it does
+    // not entangle properly with autotracking, so we entangle manually by
+    // getting the `[]` property.
+    get(errors, '[]');
+
+    return errors;
   },
 
   /**


igorT:
@pzuraq thanks a bunch for the fix! Can you point me to where this got changed in ember so I can track what kind of backport needs to happen

andreyfel:
@igorT  this issue appeared in Ember 3.20.4. Can this fix be backported to ember-data@3.20?

igorT:
@andreyfel yup, will do that as part of the release process

PR #7275, opened 2020-08-21
Update rest.js

kookyungmin:


kookyungmin:
I want to use option credentials : 'include', but I cannot use it because 'same-origin' is fixed.

Issue #7256, opened 2020-07-19
API docs do not match Octane blueprints 


maxwondercorn:
The API docs do not reflect the native classes generated with Octane. I didn't see any open issues on this and didn't know if this is in the works or something that's looking for a volunteer? 

I see some of the example code has been updated in 3.20.

edit - add package checklist

Packages:
- [x] adapter - pr #7286
- [x] model - pr #7292 
- [x] serializer - pr #7296
- [x] store - #7297

igorT:
@maxwondercorn oops, missed this. Yeah this would be a great volunteer effort. Do you have time to work on this?

maxwondercorn:
Yes, I can find time to work on this. If it's okay I may want to submit multiple PRs rather than trying to do everything in one large PR. Easier to do and easier to review the PRs.

igorT:
Yup, that sounds like a great way to do it

PR #7286, opened 2020-08-28
[DOC adapter] convert examples to native/Octane class syntax

maxwondercorn:
Updated adapter examples - reference #7256. Will continue work on the remaining packages.

I noticed that the AdapterError [page](https://api.emberjs.com/ember-data/3.20/classes/AdapterError) in the api reference is blank. I don't think that's intentional and I can open a separate issue.


igorT:
Thanks a bunch!

Issue #5204, opened 2017-10-04
hasMany relationship not updating when removing item


ultish:
I currently have a use case when a model is removed from a model's hasMany relationship , ember data isn't pushing that update through. The following events happen:

1. Model is loaded with a hasMany relationship. For example Person model contains hasMany "pets".
2. A web socket notification is received that this Person model was updated from the server so we perform a reload via: this.get('store').query('person', { id: 1, reload: true});
3. This fetches the updated version of the Person, who's "pets" attribute is now empty. Looking at the JSON that's sent down, the Person's pet attribute is an empty array. We're using the DS.RESTAdapter here. 
4. On my display whose template shows {{#each person.pets}}, it still shows the previous pet.

Inspecting the ember store I can see the Person's "pets" attribute has a canonicalState.length === 0, but it's currentState.length remains at 1 with the previous pet.

Of note, when I **add** a new pet to the relationship, this updates fine and the template shows the new pet (when inspecting I can see canonicalState matches currentState). Whenever I remove a pet however, I get into this strange state. 

Any ideas?

Ember version: 2.12.2
Ember Data version: 2.12.2

erichonkanen:
probably related to https://github.com/emberjs/data/issues/5204

erichonkanen:
probably related to https://github.com/emberjs/data/issues/4974

sly7-7:
Closing as duplicate of #4974

erichonkanen:
@ultish this is how we got around a similar scenario for now:
```
    // hack to prevent unload record on existing internal model issue.
    // must come after we unloadRecord() above.
    // https://github.com/emberjs/data/issues/5237
    get(this, 'store')._removeFromIdMap(foo._internalModel);
```

PR #7287, opened 2020-08-28
[BUGFIX]: Ensure empty response for hasMany relationship is properly reflected on client

snewcomer:
close https://github.com/emberjs/data/issues/7318
ref https://github.com/emberjs/data/issues/5204

- Notes and links in each commit when removing from unused methods.
- flushCanonicalLater to sync local state with "persisted" state.

https://discord.com/channels/480462759797063690/486549196837486592/748932008360542480
https://github.com/amk221/-ember-data-remove-has-many/blob/master/tests/unit/controllers/application-test.js#L53

### DIFF
--- a/packages/record-data/addon/-private/relationships/state/has-many.ts
+++ b/packages/record-data/addon/-private/relationships/state/has-many.ts
@@ -110,6 +110,9 @@ export default class ManyRelationship extends Relationship {
     this.removeRecordDataFromOwn(recordData);
   }
 
+  /*
+    sync client and persisted state
+  */
   flushCanonical() {
     let toSet = this.canonicalState;
 
@@ -156,6 +159,9 @@ export default class ManyRelationship extends Relationship {
     this.notifyHasManyChange();
   }
 
+  /*
+    Handle add/removal from adapter
+  */
   computeChanges(recordDatas: RelationshipRecordData[] = []) {
     const members = this.canonicalMembers.toArray();
     for (let i = members.length - 1; i >= 0; i--) {
@@ -164,6 +170,9 @@ export default class ManyRelationship extends Relationship {
     for (let i = 0, l = recordDatas.length; i < l; i++) {
       this.addCanonicalRecordData(recordDatas[i], i);
     }
+
+    // sync with local state
+    this.flushCanonicalLater();
   }
 
   /*
--- a/packages/-ember-data/tests/integration/adapter/rest-adapter-test.js
+++ b/packages/-ember-data/tests/integration/adapter/rest-adapter-test.js
@@ -997,6 +997,45 @@ module('integration/adapter/rest_adapter - REST Adapter', function(hooks) {
       });
   });
 
+  test('updateRecord - hasMany relationships faithfully reflect removal from response', async function(assert) {
+    Post.reopen({ comments: DS.hasMany('comment', { async: false }) });
+    Comment.reopen({ post: DS.belongsTo('post', { async: false }) });
+
+    store.push({
+      data: {
+        type: 'post',
+        id: '1',
+        attributes: {
+          name: 'Not everyone uses Rails',
+        },
+        relationships: {
+          comments: {
+            data: [{ type: 'comment', id: '1' }],
+          },
+        },
+      },
+      included: [
+        {
+          type: 'comment',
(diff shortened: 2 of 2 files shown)


packages/record-data/addon/-private/relationships/state/has-many.ts
```diff
@@ -162,25 +170,9 @@ export default class ManyRelationship extends Relationship {
...
-      }
 
-      this.canonicalMembers.add(recordData);
-      this.members.add(recordData);
-      this.setupInverseRelationship(recordData);
-    }
-
-    this.canonicalState = this.canonicalMembers.toArray();
+    // sync with local state
+    this.flushCanonicalLater();
```
snewcomer, line 175: This was the main change - at the end of `computeChanges`, flush to sync with local state

packages/record-data/addon/-private/relationships/state/has-many.ts
```diff
@@ -225,7 +217,7 @@ export default class ManyRelationship extends Relationship {
     return payload;
   }
 
-  updateData(data, initial) {
+  updateData(data) {
```
snewcomer, line 220: `initial` is never an argument passed

packages/record-data/addon/-private/relationships/state/relationship.ts
```diff
@@ -620,7 +610,7 @@ export default class Relationship {
    `push` use `updateMeta`, `updateData` and `updateLink` to update the state
    of the relationship.
    */
-  push(payload: JsonApiRelationship, initial?: boolean) {
+  push(payload: JsonApiRelationship) {
```
snewcomer, line 613: Same here - never passed as an argument.

This seems to be accomplished with `addRecordData` && `addCanonicalRecordData`

snewcomer: https://github.com/emberjs/data/blob/master/packages/record-data/addon/-private/record-data.ts#L223

A specific use case to show `initial` is not a used argument for `push`

packages/record-data/addon/-private/relationships/state/belongs-to.ts
```diff
@@ -50,20 +50,6 @@ export default class BelongsToRelationship extends Relationship {
...
-    }
-
-    // When we initialize a belongsTo relationship, we want to avoid work like
-    // notifying our internalModel that we've "changed" and excessive thrash on
-    // setting up inverse relationships
-    this.canonicalMembers.add(recordData);
-    this.members.add(recordData);
-    this.inverseRecordData = this.canonicalState = recordData;
-    this.setupInverseRelationship(recordData);
-  }
```
snewcomer, line 65: Without `initial` to `push`, this isn't needed

packages/record-data/addon/-private/relationships/state/relationship.ts
```diff
@@ -331,16 +331,6 @@ export default class Relationship {
     });
   }
 
-  addCanonicalRecordDatas(recordDatas: RelationshipRecordData[], idx: number) {
```
snewcomer, line 334: Searching across the codebase, this isn't used anywhere

amk221:
Got some code I'd love to ship but need this to go in first :D

snewcomer:
@amk221 I'll see if we can discuss in the first e-d meeting of the year!

amk221:
Hi all, really keen to get this fix in. Itching to release some new stuff

amk221:
👋 Do you know of any temporary work arounds I can use in the meantime?

I've tried the following:

* Reloading the record 
* Pushing a new hasMany payload to the record 
* Unloading the record (works, but not a viable solution, because this unloads all other relationships too)

So, in a last ditch effort to ship, I've gone with this:

```javascript
export default class UserSerializer extends BaseSerializer {
  attrs = {
    roles: {
      embedded: 'always'
    }
  };

  normalizeUpdateRecordResponse(store, primaryModelClass, payload) {
    // Horrible hack
    if (payload.user.roles.length === 0) {
      window.location.reload();
    }

    return super.normalizeUpdateRecordResponse(...arguments);
  }
}
```

runspired:
@amk221 this is likely fixed on master already, mind PR'ing just the test?

PR #7292, opened 2020-08-29
[DOC model] convert api examples to native classes/Octane

maxwondercorn:
Updated model examples - reference #7256.

### DIFF
--- a/packages/model/addon/-private/errors.js
+++ b/packages/model/addon/-private/errors.js
@@ -425,11 +425,11 @@ export default ArrayProxy.extend(DeprecatedEvented, {
   /**
     Checks if there are error messages for the given attribute.
 
-    ```app/routes/user/edit.js
-    import Route from '@ember/routing/route';
+    ```app/controllers/user/edit.js
+    import Controller from '@ember/controller';
     import { action } from '@ember/object';
 
-    export default class UserEditRoute extends Route {
+    export default class UserEditController extends Controller {
       @action
       save(user) {
         if (user.get('errors').has('email')) {
--- a/packages/model/addon/-private/errors.js
+++ b/packages/model/addon/-private/errors.js
@@ -53,15 +53,15 @@ import { DeprecatedEvented } from '@ember-data/store/-private';
   - `attribute` The name of the property associated with this error message
 
   ```handlebars
-  <label>Username: <Input @value={{this.model.username}} /> </label>
-  {{#each this.model.errors.username as |error|}}
+  <label>Username: <Input @value={{@model.username}} /> </label>
+  {{#each @model.errors.username as |error|}}
     <div class="error">
       {{error.message}}
     </div>
   {{/each}}
 
-  <label>Email: <Input @value={{this.model.email}} /> </label>
-  {{#each this.model.errors.email as |error|}}
+  <label>Email: <Input @value={{@model.email}} /> </label>
+  {{#each @model.errors.email as |error|}}
     <div class="error">
       {{error.message}}
     </div>
@@ -72,7 +72,7 @@ import { DeprecatedEvented } from '@ember-data/store/-private';
   object to get an array of all the error strings.
 
   ```handlebars
-  {{#each this.model.errors.messages as |message|}}
+  {{#each @model.errors.messages as |message|}}
     <div class="error">
       {{message}}
     </div>
@@ -148,7 +148,7 @@ export default ArrayProxy.extend(DeprecatedEvented, {
     record. This is useful for displaying all errors to the user.
 
     ```handlebars
-    {{#each this.model.errors.messages as |message|}}
+    {{#each @model.errors.messages as |message|}}
       <div class="error">
         {{message}}
       </div>
--- a/packages/model/addon/-private/errors.js
+++ b/packages/model/addon/-private/errors.js
@@ -22,10 +22,10 @@ import { DeprecatedEvented } from '@ember-data/store/-private';
   ```app/models/user.js
   import Model, { attr } from '@ember-data/model';
 
-  export default Model.extend({
-    username: attr('string'),
-    email: attr('string')
-  });
+  export default class UserModel extends Model {
+    @attr('string') username;
+    @attr('string') email;
+  }
   ```
   And you attempted to save a record that did not validate on the backend:
 
@@ -53,15 +53,15 @@ import { DeprecatedEvented } from '@ember-data/store/-private';
   - `attribute` The name of the property associated with this error message
 
   ```handlebars
-  <label>Username: {{input value=username}} </label>
-  {{#each model.errors.username as |error|}}
+  <label>Username: <Input @value={{this.model.username}} /> </label>
+  {{#each this.model.errors.username as |error|}}
     <div class="error">
       {{error.message}}
     </div>
   {{/each}}
 
-  <label>Email: {{input value=email}} </label>
-  {{#each model.errors.email as |error|}}
+  <label>Email: <Input @value={{this.model.email}} /> </label>
+  {{#each this.model.errors.email as |error|}}
     <div class="error">
       {{error.message}}
     </div>
@@ -72,7 +72,7 @@ import { DeprecatedEvented } from '@ember-data/store/-private';
   object to get an array of all the error strings.
 
   ```handlebars
-  {{#each model.errors.messages as |message|}}
+  {{#each this.model.errors.messages as |message|}}
     <div class="error">
       {{message}}
     </div>
@@ -148,7 +148,7 @@ export default ArrayProxy.extend(DeprecatedEvented, {
     record. This is useful for displaying all errors to the user.
 
     ```handlebars
-    {{#each model.errors.messages as |message|}}
+    {{#each this.model.errors.messages as |message|}}
       <div class="error">
         {{message}}
       </div>
@@ -427,17 +427,17 @@ export default ArrayProxy.extend(DeprecatedEvented, {
 
     ```app/routes/user/edit.js
     import Route from '@ember/routing/route';
+    import { action } from '@ember/object';
 
-    export default Route.extend({
-      actions: {
-        save: function(user) {
-          if (user.get('errors').has('email')) {
-            return alert('Please update your email before attempting to save.');
-          }
-          user.save();
+    export default class UserEditRoute extends Route {
+      @action
+      save(user) {
+        if (user.get('errors').has('email')) {
+          return alert('Please update your email before attempting to save.');
         }
+        user.save();
       }
-    });
+    }
     ```
 
     @method has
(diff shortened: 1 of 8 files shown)


packages/model/addon/-private/has-many.js
```diff
@@ -26,18 +26,18 @@ import { computedMacroWithOptionalParams } from './util';
...
 
   ```app/models/comment.js
   import Model, { belongsTo } from '@ember-data/model';
-
-  export default Model.extend({
-    post: belongsTo('post')
-  });
+  
+  export default class CommentModel extends Model {
+    @belongsTo('post') post;
```
snewcomer, line 39: I wonder how much we want to suggest this pattern

```
@belongsTo() post;
```

packages/model/addon/-private/model.js
```diff
@@ -708,20 +708,24 @@ const Model = EmberObject.extend(DeprecatedEvented, {
...
+      @action
+      confirm() {
+        this.controller.get('model').save();
+      }
+
+      @action
+      undo() {
+        this.controller.get('model').rollbackAttributes();
+      }
+    }
```
snewcomer, line 728: Is this possible?  Perhaps we change this to a controller if you can't call an action on a route?

packages/model/addon/-private/model.js
```diff
@@ -2157,7 +2164,7 @@ Model.reopenClass({
       callback.call(binding, name, meta);
     });
   },
-
+  // TODO refactor for native classes/octane??
```
snewcomer, line 2167: 👍

packages/model/addon/-private/has-many.js
maxwondercorn: I converted to decorators because that's what's being used in the Ember guides and other learning materials. I think we should be consistent. IMO, for someone just learning Ember its going to be confusing when they start looking at the API docs and they're doing something different then what they've learned elsewhere.

I'm assuming data's API are versioned the similar to the guides? So the changes would not be reflected in previous versions?  I know how the guides are built - no idea on the data API docs are built.

packages/model/addon/-private/model.js
maxwondercorn: I thought the original construct was a bit "odd" before I converted it. There must have been some reasoning for doing it that way instead of using a controller.

snewcomer: I forgot.  You can call `sendAction` - first looking at the controller then up the active route path.  I'm sure new users would feel confused on how to call this action with the current guides.  I also don't think it is a prescribed path.  Up to you if you want to convert it!

packages/model/addon/-private/model.js
maxwondercorn: The *.extend is the correct syntax for native classes so no refactoring required

packages/model/addon/-private/model.js
maxwondercorn: I'm going to convert to a controller.  `sendAction` is deprecated and you would need to use the `{{action}}` helper to use an action on the route

packages/model/addon/-private/errors.js
```diff
@@ -427,17 +427,17 @@ export default ArrayProxy.extend(DeprecatedEvented, {
...
+    export default class UserEditRoute extends Route {
+      @action
+      save(user) {
+        if (user.get('errors').has('email')) {
+          return alert('Please update your email before attempting to save.');
         }
+        user.save();
       }
-    });
+    }
```
snewcomer, line 440: Should we convert this to a controller as well?

packages/model/addon/-private/model.js
```diff
@@ -706,22 +706,26 @@ const Model = EmberObject.extend(DeprecatedEvented, {
...
+    export default class ModelDeleteController extends Controller {
+      @action
+      softDelete() {
+        this.model.deleteRecord();
       }
-    });
+
+      @action
+      confirm() {
+        this.model').save();
```
snewcomer, line 721: r/this.model').save()/this.model.save()

packages/model/addon/-private/model.js
```diff
@@ -1586,17 +1591,18 @@ Model.reopenClass({
    ```app/models/post.js
    import Model, { hasMany } from '@ember-data/model';
 
-   export default Model.extend({
-      comments: hasMany('message')
-    });
+   export default class PostModel extends Model {
+    @hasMany('message') comments;
```
snewcomer, line 1595: looks like these can be indented 2 spaces (or whatever it was before)

packages/model/addon/-private/model.js
```diff
@@ -2157,7 +2163,7 @@ Model.reopenClass({
       callback.call(binding, name, meta);
     });
   },
-
+  
   /**
```
snewcomer, line 2167: ☝️

packages/model/addon/-private/system/model-for-mixin.ts
```diff
@@ -3,8 +3,8 @@ import { getOwner } from '@ember/application';
 import Model from '../model';
 
 type Store = import('@ember-data/store').default;
-
-/*
+// TODO refactor for native classes/octane??
+/*
```
snewcomer, line 7: 👍

snewcomer (commented):
Phenomenal job on this PR and the previous PRs (that were merged)!

packages/model/addon/-private/model.js
maxwondercorn: fixed

packages/model/addon/-private/model.js
maxwondercorn: fixed

packages/model/addon/-private/errors.js
maxwondercorn: yes - converted to controller

packages/model/addon/-private/model.js
maxwondercorn: fixed indent

packages/model/addon/-private/model.js
maxwondercorn: removed extra chars...

packages/model/addon/-private/model.js
maxwondercorn: @snewcomer I'm assuming you are pointing out the example for `eachAttribute`.  I believe this would be the native class equivalent??

```js
   import Model, { attr } from '@ember-data/model';

   const person = new Model({
      @attr('string') firstName;
      @attr('string') lastName;
      @attr('date') birthday;
    })

   person.eachAttribute(function(name, meta) {
      console.log(name, meta);
    });
```

packages/model/addon/-private/system/model-for-mixin.ts
maxwondercorn: @snewcomer not sure how to convert this example.  Actually, I don't understand the use case...

Maybe 🤷 
```js
import Model, { belongsTo} from '@ember-data/model';
import Mixin from '@ember/object/mixin';

let Comment = new Model ({
  @belongsTo('commentable', { polymorphic: true }) owner;
})

let Commentable = Mixin.create({
  comments: hasMany('comment')
});
```

snewcomer:
```
import Model, { belongsTo} from '@ember-data/model';
import Mixin from '@ember/object/mixin';

let Comment = class CommentModel extends Model {
  @belongsTo('commentable', { polymorphic: true }) owner;
}

let Commentable = Mixin.create({
  comments: hasMany('comment')
});
```

maxwondercorn: So close and yet so far, this also answers the other example I questioned. Updated

packages/model/addon/-private/model.js
maxwondercorn: updated

packages/model/addon/-private/model.js
```diff
@@ -2131,13 +2137,13 @@ Model.reopenClass({
...
-      birthday: attr('date')
-    });
+   const person = class PersonModel extends Model {
+      @attr('string') firstName;
+      @attr('string') lastName;
+      @attr('date') birthday;
+    }
 
-   Person.eachAttribute(function(name, meta) {
+   person.eachAttribute(function(name, meta) {
```
snewcomer, line 2146:
```suggestion
   PersonModel.eachAttribute(function(name, meta) {
```

packages/model/addon/-private/model.js
```diff
@@ -2131,13 +2137,13 @@ Model.reopenClass({
    ```javascript
    import Model, { attr } from '@ember-data/model';
 
-   let Person = Model.extend({
-      firstName: attr('string'),
-      lastName: attr('string'),
-      birthday: attr('date')
-    });
+   const person = class PersonModel extends Model {
```
snewcomer, line 2140:
```suggestion
   class PersonModel extends Model {
```

packages/model/addon/-private/system/model-for-mixin.ts
```diff
@@ -3,15 +3,18 @@ import { getOwner } from '@ember/application';
...
     ```
-      let Comment = Model.extend({
-        owner: belongsTo('commentable'. { polymorphic: true })
-      });
-      let Commentable = Ember.Mixin.create({
-        comments: hasMany('comment')
+      import Model, { belongsTo, hasMany } from '@ember-data/model';
+      import Mixin from '@ember/object/mixin';
+
+      let Comment = class CommentModel extends Model {
```
snewcomer, line 12:
```suggestion
      class CommentModel extends Model {
```

snewcomer: I honestly am not sure the intention of this Mixin block and what it is trying to communicate

locks:
Oops, I mistakenly requested a review. @maxwondercorn do you have some time to address them?

packages/model/addon/-private/model.js
maxwondercorn: Fixed!

packages/model/addon/-private/model.js
maxwondercorn: Fixed

packages/model/addon/-private/system/model-for-mixin.ts
maxwondercorn: fixed

maxwondercorn:
@locks thanks for pinging me. I didn't realize these were out there. Fixes pushed

packages/model/addon/-private/has-many.js
```diff
@@ -88,22 +88,24 @@ import { computedMacroWithOptionalParams } from './util';
...
+  import { hasMany } from '@ember-decorators/data';
 
-  export default Model.extend({
-    comments: hasMany('comment', {
+  export default class PostModel extends Model {
+    @hasMany('comment', {
       inverse: 'redPost'
     })
-  });
+    commnets;
```
snewcomer, line 107:
```suggestion
    comments;
```

snewcomer (approved):
Wow.  Really great PR.  Ppl will surely appreciate this!

sandstrom:
awesome @maxwondercorn 🎉

PR #7303, opened 2020-08-31
Update Changelog for v3.22.0-beta.0

snewcomer:


CHANGELOG.md
```diff
@@ -1,5 +1,39 @@
 # Ember Data Changelog
 
+## Release 3.22.0-alpha.2 (August 31, 2020)
```
igorT, line 3: This should be 3.22-beta.0

igorT:
We should look into whats up with azure builds tests, they might be leftovers/they don't match githubs, but don't think that should block merge/release

PR #7315, opened 2020-09-06
[CI]: remove azure pipeline

snewcomer:
Proposal - Github CI seems to be stable enough that we can rely on one CI test run


igorT:
cc @hjdivad

PR #7319, opened 2020-09-10
[DOC model] Convert model blueprint HELP.md to native class syntax

SergeAstapov:
Command `ember g model` emits native class syntax so this brings example in pair with actual output.


snewcomer (approved):
Great idea!

PR #7325, opened 2020-09-16
[BUGFIX identifiers] Address issue with polymorphic findRecord

krasnoukhov:
Hello! I found this issue while updating our app from 3.12 to 3.16. Previously it was possible to use `findRecord` to retrieve a specific instance of polymorphic model through the "base" model endpoint. This is useful when only the id is known but not the specific type. There was a prior fix in #6644 but the difference in this case is that there's no cache (through the `dealership`) yet.

So far I'm only adding a failing test which runs into this assertion:

https://github.com/emberjs/data/blob/78162a065be99fb3fd284eb944aff98baa483ec7/packages/store/addon/-private/identifiers/cache.ts#L519-L524

My next thought was to remove that assertion (since TODO says that it may be problematic), and I tried that, but I have a feeling that'd not be a complete fix since `foundFerrari.constructor.modelName` in the test resolves to `car` and not `ferrari`. I checked this on 3.12 and I believe it should be `ferrari`. UPD: visible in [this CI run](https://github.com/emberjs/data/pull/7325/checks?check_run_id=1124552339)

I tried digging into the source but I'm a little bit lost really so any feedback would be appreciated. I hope we can address this and backport into 3.16.

Thank you!


krasnoukhov:
Ok, I narrowed it down to the fact that `lid` is assigned when `findRecord` is performed. Seems like not doing that when type does not match would be a safe bet and should only apply in polymorphic context, so a new identifier will be created down the road.

Let me know if I should add some more tests. I'd be happy to consider alternative approaches as well.

snewcomer (approved):
Very nice PR!  Looks like your tests cover when REQUEST_SERVICE is on and off.

krasnoukhov:
Thanks for the approve. Yeah, that integration test is very high level. Sounds like this is going to be merged? I can create a PR into `lts-3-16`, I already have these changes backported. Maybe `lts-3-20` as well? Let me know

packages/-ember-data/tests/integration/identifiers/polymorphic-scenarios-test.ts
```diff
@@ -61,7 +61,31 @@ module('Integration | Identifiers - single-table-inheritance polymorphic scenari
...
+                color: 'red',
+              },
+            },
+          });
+        }
+      }
+      owner.register('adapter:application', TestAdapter);
+
+      const foundFerrari = await store.findRecord('car', '1');
+      assert.strictEqual(foundFerrari.constructor.modelName, 'ferrari', 'We found the right type');
```
hjdivad, line 85: This looks great, thanks.

Do you mind adding a few more assertions

1. assert that `{ type: 'car', id }` is in the cache
2. assert that `{ type: 'ferrari', id }` is in the cache
3. assert that they point to the same record

hjdivad (changes requested):

packages/-ember-data/tests/integration/identifiers/polymorphic-scenarios-test.ts
krasnoukhov: Hi @hjdivad, thanks for the feedback. I just added this to the test, I hope this is what you meant:

```js
const cachedFerrari = await store.peekRecord('ferrari', '1');
assert.strictEqual(cachedFerrari.constructor.modelName, 'ferrari', 'We cached the right type');
```

However I tried adding one for the `car` type of record and it's actually not in cache, `store.peekRecord('car', '1')` return null. Isn't this expected? 🤔

hjdivad: @krasnoukhov depends on how the test sets up identifiers; it should be possible to configure it either way, but the utility of having both `car: 1` and `ferrari: 1` point to the same record in the identity map is that you could then later `store.findRecord('car', 1)` and get back the ferrari without "forgetting" that it had already been loaded

hjdivad (approved):

packages/-ember-data/tests/integration/identifiers/polymorphic-scenarios-test.ts
krasnoukhov: Let me try this now!

hjdivad:
@igorT @runspired do either of you see any issues with the implementation?  Broadly this LGTM

packages/-ember-data/tests/integration/identifiers/polymorphic-scenarios-test.ts
krasnoukhov: However I'm not sure still... I have `store.findRecord('car', '1')` statement in the test already, not sure if I see the point of running it again and then asserting the equality with previous run?

hjdivad: @krasnoukhov `store.peekRecord` would be equivalent to doing another `store.findRecord` here, i.e. verify that it's in the cache.

igorT:
@krasnoukhov thanks a bunch for looking into this, great job dealing with the complex codebase without prior experience.

I have run into this issue previously, and added a fix for the case handling it here: https://github.com/emberjs/data/blob/ff4f9111fcfa7dd9e39804ed17f5af27a4a01378/packages/store/addon/-private/identifiers/cache.ts#L558

I am wondering why this doesn't trigger the merge, it should conceptually be handled by the merge function. @snewcomer do you mind taking a look with @krasnoukhov to see why it doesn't get triggered

krasnoukhov:
Thanks @igorT. I stumbled upon that code as well. I tried adding `debugger` to both `detectMerge` and  `updateRecordIdentifier` and these are not even called for within that test scenario that I've added. Not really sure whether that's expected, any pointers will be appreciated...

igorT:
@snewcomer lets figure out why `detectMerge` isn't getting called

snewcomer:
I took a look at the cache and here are my thoughts.

Seems to me this is the right solution.  `detectMerge` only deals with existing cache items, thus no existing identifier to "merge" as shown in this PR's test case.  Moreover, this solution in this PR is able to get the correct `lid` much earlier in the process (at the outset of find) before propagating through the many layers.

`data` in `detectMerge`: 
Screen Shot 2020-10-24 at 3 31 54 PM

An alternative would be in `detectMerge` to `makeStableRecordIdentifier` from the data if we detect a mismatch of `lid` or `type`.  This new identifier would then be passed to `_mergeRecordIdentifiers`.

snewcomer:
To clarify further, without these changes, the identifier created further upstream looks like this.

Screen Shot 2020-10-28 at 8 52 54 PM

As a result, the `IdentifierCache` by types cache has an entry of `car`.  Because this is actually a `ferrari`, there is no cache entry and thus our logic dictates we should not "merge identifiers".

However, with these changes, not setting the `lid` when we go to `_push` the data into the store will generate a new internalModel for us with the right `lid`/`type` of ferrari.  

If we set the `lid`, then we have what we need and thus don't end up creating an entry in the `IdentifierCache` by type for the new `ferrari`.  The cache just looks like this...

Screen Shot 2020-10-28 at 10 04 32 PM

Lastly, simply commenting and not making any conclusions - I noticed one other thing.  The `internalModel` is in a state of loading [here](https://github.com/emberjs/data/blob/34ee5370d6b3ec6046d53e6c0fe4c9316690aa94/packages/store/addon/-private/system/core-store.ts#L2684) without these changes.  With these changes the rootState is `root.empty`.  I'm not sure why exactly the state is such.

snewcomer:
👋  @krasnoukhov Igor presented another solution.  Although your solution is a great idea and perhaps cleaner (is getting the lid correct further upstream better? I would assume so), the 👇 aligns us closer with how the current system works.  Notice the surrounding code deals with mismatches as well.  Let us know your thoughts (still one thing to deal with on a partner test but close to ready)!!

https://github.com/emberjs/data/pull/7363/files

krasnoukhov:
Thanks @igorT and @snewcomer! Huge 🤦 moment for me since I assumed `updateRecordIdentifier` isn't called but that was totally because of the changes I already made to address the issue.

I think this alternative solution is definitely much more cleaner in a sense that identifier update handling stays close together. Nice work and I appreciate your help with digging into this

krasnoukhov:
Closing in favor of #7363

Issue #7311, opened 2020-09-02
ED + Octane: ManyArray inside POJO is not tracked


urbany:
### Reproduction
https://ember-twiddle.com/26f68cde34e59c46c0c4b8c47d808c42

### Description
Hi, when I sort/filter a ManyArray inside a POJO (check the route) it is not autotracked when adding new elements (in tiwddle: `sortedServices`). If instead I access the ManyArray directly from another model it works correctly (in tiwddle: `sortedThroughBusinessServices`). If I use the old computed sort helper it also works (in twiddle: `sortedServicesComputed`)

@pzuraq said on Discord I should create the issue in ED and also this:
> objectAt is not entangling properly..

Thanks in advance!

### Versions

The twiddle does not have the latest version of ember-source but i tried this in my project today with latest version of cli, source and data.

```cli
ember-cli@3.21.0
ember-data@3.21.0
ember-source@3.21.1
```

urbany:
I just noticed that if a add `get(this.model, 'services.length')` to the sortedServices getter it starts receiving updates of new services added to the ManyArray.

igorT:
@pzuraq is this similar to the bug you were already working on

urbany:
Hi, sorry to bother I know you guys help on your free time, but anyway i can make this easier for you to check if at least it is a bug? Thanks again!

pzuraq:
I haven't had time to dig in, but all that needs to happen is that we should call `get(this, '[]')` at least once whenever the values in the array are accessed. I'm not sure what the best place would be to add that code.

Issue #7328, opened 2020-09-18
Ember.js 3.20.4 no longer compensating for EmberArray implementation oversight in ED


samcic:
This was originally opened in https://github.com/emberjs/ember.js/issues/19139, but @pzuraq asserted that it's an ED issue related to the implementation of EmberArray, which ember.js is no longer compensating for as of Ember 3.20.4, so I've reopened it here.

May be related to https://github.com/emberjs/data/issues/7310 ,which already has a PR ready with a failing test case.

Issue copied below for reference:

### 🐞 Describe the Bug
If I'm tracking a model object with a `hasMany` list of items, I have a getter that uses those items to compute something else, and I clear those items, then the getter doesn't fire (but it does if I *add* items to the list). 

Here's a quick video to describe it visually:

https://recordit.co/OeEey7Edrw

I've also confirmed the issue on the latest Ember.js Beta build (3.22-beta-3).

I'm aware of various other similarly-sounding reports of issues between 3.20.3 and 3.20.4 (e.g. this one on [sort](https://github.com/emberjs/ember.js/issues/19101), and [this one](https://github.com/emberjs/ember.js/issues/19105) which seems rather similar but is apparently fixed in 3.22-beta-3). I don't believe there's an open issue tracking this particular behavior.

### 🔬 Minimal Reproduction
Describe steps to reproduce. If possible, please, share a link with a minimal reproduction.

* `git clone https://github.com/samcic/reactivity-bug.git`
* `cd reactivity-bug`
* `yarn install`
* `ember serve`
* Navigate to http://localhost:4200/
* Click the "Add" button to add some items to the collection and watch the counter increase
* Click the "Clear" button to clear the items collection and watch the counter do nothing => BUG!
* Click the "Add" button again to add another item, and watch the counter increase to 1 (i.e. the collection was cleared but the getter didn't react to the "cleared" state).

### 😕 Actual Behavior
When the collection is cleared, the getter `itemsLength` doesn't fire and the template doesn't update.

### 🤔 Expected Behavior
When the collection is cleared, the getter `itemsLength` should fire and the template should show the count `0`.

### 🌍 Environment

- Ember: 3.20.4 to current (issue doesn't exist in 3.20.3)
- Ember-Data: ~3.21.0
- Node.js/npm: 12.14.0
- OS: Windows 10
- Browser: Chrome 85

snewcomer:
@samcic Do you happen to have a reproduction?  Want to see if #7330 covers your use case

[...]

PR #7330, opened 2020-09-20
[BUG]: Consume array access to autotrack hasMany

snewcomer:
close https://github.com/emberjs/data/issues/7311 https://github.com/emberjs/ember.js/issues/19101 https://github.com/emberjs/data/issues/7328 https://github.com/emberjs/ember.js/issues/19139
ref https://github.com/emberjs/ember.js/issues/19105

Another related autotracking fix for `errorsFor` https://github.com/emberjs/data/pull/7273

## In this PR
Has Many Arrays implement the Mutable Array mixin.  We need to instrument `length` and `objectAt` in order to ensure autotracking works.

### To do
Follow up: convert to class based && ts file

### DIFF
--- a/packages/model/addon/-private/system/many-array.js
+++ b/packages/model/addon/-private/system/many-array.js
@@ -67,7 +67,9 @@ export default EmberObject.extend(MutableArray, DeprecatedEvented, {
     @property {Boolean} isLoaded
     */
     this.isLoaded = this.isLoaded || false;
-    this.length = 0;
+
+    // async has-many rendering tests
+    this._length = 0;
 
     /**
     Used for async `hasMany` arrays
@@ -164,6 +166,17 @@ export default EmberObject.extend(MutableArray, DeprecatedEvented, {
       }
     }
     return false;
+  },
+
+  get length() {
+    // By using `get()`, the tracking system knows to pay attention to changes that occur.
+    get(this, '[]');
+
+    return this._length;
+  },
+
+  set length(value) {
+    return (this._length = value);
   },
 
   objectAt(index) {
--- a/packages/-ember-data/tests/acceptance/relationships/has-many-test.js
+++ b/packages/-ember-data/tests/acceptance/relationships/has-many-test.js
@@ -1,4 +1,8 @@
-import { render } from '@ember/test-helpers';
+import { action } from '@ember/object';
+import { sort } from '@ember/object/computed';
+import { inject as service } from '@ember/service';
+import { click, findAll, render } from '@ember/test-helpers';
+import Component from '@glimmer/component';
 import Ember from 'ember';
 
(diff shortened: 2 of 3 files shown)


packages/model/addon/-private/system/many-array.js
```diff
@@ -166,6 +166,13 @@ export default EmberObject.extend(MutableArray, DeprecatedEvented, {
     return false;
   },
 
+  toArray() {
+    // By using `get()`, the tracking system knows to pay attention to changes that occur.
+    get(this, '[]');
```
snewcomer, line 171: @pzuraq 👋 I'm noticing that caching on `this` once when accessed doesn't pass the tests I added here. We have to consume every time it is accessed. 

I'm thinking this might be expected since `sortBy` (through `toArray`) returns a new array each time.  What do you think?

pzuraq: Discussed this in the data meeting but wanted to note it down here, the issue is that `length` also needs to be instrumented here, along with `objectAt`. This because `objectAt` is not used when the length is 0, but length is read to check that.

rwjblue (approved):

packages/model/addon/-private/system/many-array.js
```diff
@@ -166,7 +165,25 @@ export default EmberObject.extend(MutableArray, DeprecatedEvented, {
     return false;
   },
 
+  get length() {
+    // By using `get()`, the tracking system knows to pay attention to changes that occur.
+    get(this, '[]');
+
+    if (typeof this._length === 'number') {
```
igorT, line 172: Why do we need this check? Should we set the length in init rather than have to do this every time you access length?

snewcomer: Either works!  I did that but then thought I would keep all the logic in one place.  I can do it in init if you would like!

urbany:
Hi everyone, seems like this is ready to go, no? Really sorry for the pressure, but this bug is really making it hard to debug tracking bugs on my update to autotracking. Thanks again! :)

snewcomer:
@urbany 👋  Yes it was!  Fixed in ember source.  EmberArrays are now entangled and should react to adding/removing to them.  I think we might just be waiting for backports to lts.

https://github.com/emberjs/data/pull/7330/commits/cb4ba29d1e8a4cfb42c90cd59a5045877d62ddb2

rwjblue (approved):
Can you drop the second commit here, and we can land this (for release in ember-data@3.22)?

jrjohnson:
Is there a guide for testing an ED branch/pr? I'm not sure how I would install the packages with this fix in an app as my usual NPM tricks don't seem to work here given the monorepo.

rwjblue (approved):

Issue #7193, opened 2020-05-14
[improve] Provide meaningful debug information for identifier update error


lifeart:
I see `Failed to update the 'id' for the RecordIdentifier '[object Object]' to '27830', because that id is already in use by '[object Object]'`. 

`[object Object]` - is not meaningful debug message, and it's tricky to track down issue (modelType? payload?)

![image](https://user-images.githubusercontent.com/1360552/81988908-e9a3f380-9644-11ea-9470-491164141640.png)

snewcomer:
Perhaps we can replace `[object Object]` with `identifier.type || identifier` instead.  Or stringify the identifier object but that not print out pretty.

Here is another solution for that error.

```
const identifierOutput = typeof identifier === 'object' ? `id: ${identifier.id}, type: ${identifier.type}` : identifier;

throw new Error(`Failed to update the 'id' for the RecordIdentifier '${identifierOutput} to '${resourceData.id}', because that id is already in use by '${matchedIdentifierOutput'`)
```

runspired:
I'm guessing this is a production build? In dev we implement `toString` for nice ergonomics.

PR #7337, opened 2020-09-25
[Debug]: improved debug msg instead of [object Object]

snewcomer:
close https://github.com/emberjs/data/issues/7193

### DIFF
--- a/packages/-ember-data/tests/integration/identifiers/cache-test.ts
+++ b/packages/-ember-data/tests/integration/identifiers/cache-test.ts
@@ -79,4 +79,71 @@ module('Integration | Identifiers - cache', function(hooks) {
       );
     });
   });
+
+  module('createIdentifierForNewRecord()', function() {
+    test('returns new identifier', async function(assert) {
+      const runspiredHash = {
+        type: 'person',
+        id: '1',
+        attributes: {
+          name: 'runspired',
+        },
+      };
+      const identifier = cache.createIdentifierForNewRecord(runspiredHash);
+
+      assert.equal(identifier.id, '1', 'identifier has id');
+      assert.equal(identifier.type, 'person', 'identifier has type');
+      assert.ok(identifier.lid, 'identifier has lid');
+    });
+  });
+
+  module('updateRecordIdentifier()', function() {
+    test('returns same identifier', async function(assert) {
+      const runspiredHash = {
+        type: 'person',
+        id: '1',
+        attributes: {
+          name: 'runspired',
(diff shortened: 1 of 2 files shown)


packages/store/addon/-private/system/store/internal-model-factory.ts
```diff
@@ -97,8 +97,9 @@ export default class InternalModelFactory {
       // we cannot merge internalModels when both have records
       // (this may not be strictly true, we could probably swap the internalModel the record points at)
       if (im && otherIm && im.hasRecord && otherIm.hasRecord) {
+        let { type, id, lid } = matchedIdentifier;
         throw new Error(
-          `Failed to update the 'id' for the RecordIdentifier '${identifier}' to '${resourceData.id}', because that id is already in use by '${matchedIdentifier}'`
+          `Failed to update the 'id' for the RecordIdentifier '${identifier}' to '${resourceData.id}', because that id is already in use by ${type}:${id} (${lid})`
```
igorT, line 102: Should we do this for the  '${identifier}' as well?

packages/-ember-data/tests/integration/identifiers/cache-test.ts
```diff
@@ -79,4 +79,71 @@ module('Integration | Identifiers - cache', function(hooks) {
       );
     });
   });
+
+  module('createIdentifierForNewRecord()', function() {
```
igorT, line 83: In general it would be great so split modules into new files

packages/store/addon/-private/identifiers/cache.ts
```diff
@@ -340,17 +340,15 @@ export class IdentifierCache {
...
     let identifier = this.getOrCreateRecordIdentifier(identifierObject);
 
-    let id = identifier.id;
     let newId = coerceId(data.id);
-
-    const keyOptions = getTypeIndex(this._cache.types, identifier.type);
     let existingIdentifier = detectMerge(this._cache.types, identifier, data, newId, this._cache.lids);
 
     if (existingIdentifier) {
+      const keyOptions = getTypeIndex(this._cache.types, identifier.type);
```
igorT, line 347: While you are cleaning up, it would be great to not have this const here, as we do a let keyOptions on line 357

igorT:
I cherry-picked the debug message commits into https://github.com/emberjs/data/pull/7342 to not block merging on the other cleanup review

igorT:
@snewcomer wanna rebase against master, and I'll merge

PR #7350, opened 2020-10-09
[Chore]: convert adapters to class syntax

snewcomer:
- [x] rest
- [x] json
- [x] json-api

Outstanding question - should I convert to `ts` before merging?


packages/adapter/addon/index.ts
```diff
@@ -321,7 +316,6 @@ export default EmberObject.extend({
     @param {Snapshot} snapshot
     @return {Promise} promise
   */
-  createRecord: null,
```
snewcomer, line 324: I removed this because this should be typed as a function. Looking at moving these to `ts` in this PR or another.

packages/adapter/addon/index.ts
```diff
@@ -55,7 +55,7 @@ import EmberObject from '@ember/object';
   @class Adapter
   @extends EmberObject
 */
-export default EmberObject.extend({
+export default class Adapter extends EmberObject {
```
andreyfel, line 58: Does it have to extend EmberObject now?

igorT:
@snewcomer discussed this in the weekly meeting, will split into a class refactor and a TS refactor.

packages/adapter/addon/index.ts
snewcomer: I bet imagine for many users, this would be an unnoticeable change.  However, technically this would be a breaking change.

packages/adapter/addon/index.js
```diff
@@ -321,7 +316,6 @@ export default EmberObject.extend({
     @param {Snapshot} snapshot
     @return {Promise} promise
   */
-  createRecord: null,
```
snewcomer, line 324: We have to remove these because this will take precedence over prototypal methods defined on our own adapters that extend this class or user defined ones.

snewcomer: Let me know if we should replace these with a function + assert or if we would be better served in the ts migration.

igorT: Not quite sure what you mean by 
> We have to remove these because this will take precedence over prototypal methods defined on our own adapters that extend this class or user defined ones.
Can you please explain more?

snewcomer: TS playground [example](https://www.typescriptlang.org/play?#code/MYGwhgzhAEAKYCcCmA7ALtA3gKGtYyYaSASksAPYIAm0AvNAOQAWAlowNzYC+22okGAGE2IWkgAexFNRjxk6LLnyFiZSjQAUASiV48yNAFcEKJgFsAntAgB3JEgzBRtCgDNo51iiSdlvXn4KFAgnemgfW2gRVjEdLkoQihAkADoQCgBzTWBUgiQiUnIqam0OIA).

PR #7356, opened 2020-10-22
[Deps]: add `@ember/string` dependency in preparation of deprecation

snewcomer:
close #5967 

Reimplementation of above PR to see if tests will pass.  Added a few other modules that need the explicit dep install.


locks (approved):
Thanks for working on this!

PR #7358, opened 2020-10-23
[Chore]: TypeScript adapters

snewcomer:
follow up #7350


packages/adapter/addon/rest.ts
```diff
@@ -309,14 +320,15 @@ class RESTAdapter extends Adapter.extend(BuildURLMixin) {
   }
 
   set fastboot(value) {
-    return (this._fastboot = value);
+    this._fastboot = value;
```
snewcomer, line 323: We could take this private property off the class if you so desired with a WeakMap

packages/adapter/addon/index.ts
```diff
@@ -113,6 +122,13 @@ export default class Adapter extends EmberObject {
     @param {Snapshot} snapshot
     @return {Promise} promise
   */
+  findRecord(store: Store, type: ShimModelClass, id: string, snapshot: Snapshot) {
+    warn('You subclassed the Adapter class but missing a findRecord override', false, {
```
snewcomer, line 126: OR something like

```
You subclassed the Adapter class and have not defined a 'findRecord' method.  Please implement 'findRecord(store: Store, type: ShimModelClass, id: string, snapshot: Snapshot) -> Promise<unknown>'
```

packages/adapter/addon/rest.ts
```diff
@@ -1318,15 +1333,15 @@ function ajaxErrorHandler(adapter, jqXHR, errorThrown, requestData) {
   return ajaxError(adapter, payload, requestData, responseData);
 }
 
-function fetchResponseData(response) {
+function fetchResponseData(response: Response): Record<string, any> {
   return {
     status: response.status,
-    textStatus: response.textStatus,
+    textStatus: response.statusText,
```
snewcomer, line 1339: https://developer.mozilla.org/en-US/docs/Web/API/Response/statusText

packages/adapter/addon/-private/build-url-mixin.ts
```diff
@@ -55,7 +58,13 @@ export default Mixin.create({
     @param {Object} query object of query parameters to send for query requests.
     @return {String} url
   */
-  buildURL(modelName, id, snapshot, requestType, query) {
+  buildURL(
+    modelName: string,
+    id: string | string[] | Record<string, any> | null,
```
igorT, line 63: The `Record` would be very confusing for the data repo, so we use Dict and ConfidentDict instead

packages/adapter/addon/-private/build-url-mixin.ts
```diff
@@ -89,9 +98,9 @@ export default Mixin.create({
     @param {String} id
     @return {String} url
   */
-  _buildURL(modelName, id) {
+  _buildURL(modelName: string, id: string): string {
```
igorT, line 101: In the code here it looks like id can be null/undefined?

packages/adapter/addon/-private/build-url-mixin.ts
```diff
@@ -195,7 +204,7 @@ export default Mixin.create({
    @param {String} modelName
    @return {String} url
    */
-  urlForQuery(query, modelName) {
+  urlForQuery(query: Record<string, any>, modelName: string): string {
```
igorT, line 207: Dict here and below

packages/adapter/addon/-private/build-url-mixin.ts
```diff
@@ -388,7 +397,7 @@ export default Mixin.create({
     @param {String} parentURL
     @return {String} urlPrefix
   */
-  urlPrefix(path, parentURL) {
+  urlPrefix(path: string, parentURL: string): string {
```
igorT, line 400: path can be null/undefined/'' by reading the code

packages/adapter/addon/-private/fastboot-interface.ts
```diff
@@ -0,0 +1,24 @@
+export interface Request {
+  method: string;
+  body: unknown;
+  cookies: unknown;
```
igorT, line 4: Lets have a bit more specific types here for objects

packages/adapter/addon/index.ts
```diff
@@ -580,7 +638,7 @@ export default class Adapter extends EmberObject {
     @param {SnapshotRecordArray} snapshotRecordArray
     @return {Boolean}
   */
-  shouldReloadAll(store, snapshotRecordArray) {
+  shouldReloadAll(store: Store, snapshotRecordArray) {
```
igorT, line 641: `snapshotRecordArray` needs a type

packages/adapter/addon/json-api.ts
```diff
@@ -156,7 +160,7 @@ class JSONAPIAdapter extends RESTAdapter {
     @param {Object} options
     @return {Object}
   */
-  ajaxOptions(url, type, options = {}) {
+  ajaxOptions(url: string, type: string, options = {}) {
```
igorT, line 163: options needs a type

packages/adapter/addon/rest.ts
```diff
@@ -522,8 +532,8 @@ class RESTAdapter extends Adapter.extend(BuildURLMixin) {
     @param {SnapshotRecordArray} snapshotRecordArray
     @return {Promise} promise
   */
-  findAll(store, type, sinceToken, snapshotRecordArray) {
-    let query = this.buildQuery(snapshotRecordArray);
+  findAll(store: Store, type: ShimModelClass, sinceToken, snapshotRecordArray: SnapshotRecordArray) {
+    let query: Record<string, any> = this.buildQuery(snapshotRecordArray);
```
igorT, line 536: `Dict`

packages/adapter/addon/rest.ts
```diff
@@ -578,7 +588,7 @@ class RESTAdapter extends Adapter.extend(BuildURLMixin) {
     @param {Object} query
     @return {Promise} promise
   */
-  queryRecord(store, type, query) {
+  queryRecord(store: Store, type: ShimModelClass, query) {
```
igorT, line 591: query needs a type

packages/adapter/addon/rest.ts
```diff
@@ -662,7 +672,7 @@ class RESTAdapter extends Adapter.extend(BuildURLMixin) {
     @param {Object} relationship meta object describing the relationship
     @return {Promise} promise
   */
-  findHasMany(store, snapshot, url, relationship) {
+  findHasMany(store: Store, snapshot: Snapshot, url: string, relationship) {
```
igorT, line 675: relationship needs a type

packages/adapter/addon/rest.ts
```diff
@@ -550,7 +560,7 @@ class RESTAdapter extends Adapter.extend(BuildURLMixin) {
     @param {Object} query
     @return {Promise} promise
   */
-  query(store, type, query) {
+  query(store: Store, type: ShimModelClass, query) {
```
igorT, line 563: We should type a return value for all of these

packages/adapter/addon/rest.ts
```diff
@@ -775,26 +785,26 @@ class RESTAdapter extends Adapter.extend(BuildURLMixin) {
...
 
     return this.ajax(this.buildURL(type.modelName, id, snapshot, 'deleteRecord'), 'DELETE');
   }
 
-  _stripIDFromURL(store, snapshot) {
+  _stripIDFromURL(store: Store, snapshot: Snapshot): string {
     let url = this.buildURL(snapshot.modelName, snapshot.id, snapshot);
 
-    let expandedURL = url.split('/');
+    let expandedURL: string[] = url.split('/');
```
igorT, line 797: do you need this type?

packages/adapter/addon/rest.ts
```diff
@@ -902,11 +912,11 @@ class RESTAdapter extends Adapter.extend(BuildURLMixin) {
     @param  {Object} requestData - the original request information
     @return {Object | AdapterError} response
   */
-  handleResponse(status, headers, payload, requestData) {
+  handleResponse(status: number, headers, payload: Payload, requestData: JQueryAjaxSettings) {
     if (this.isSuccess(status, headers, payload)) {
       return payload;
     } else if (this.isInvalid(status, headers, payload)) {
-      return new InvalidError(payload.errors);
+      return new InvalidError((payload as Record<string, any>).errors);
```
igorT, line 919: Why do you need a cast here?

packages/adapter/addon/rest.ts
```diff
@@ -1206,8 +1215,8 @@ class RESTAdapter extends Adapter.extend(BuildURLMixin) {
     @param  {Snapshot} snapshot
     @return {Object}
   */
-  buildQuery(snapshot) {
-    let query = {};
+  buildQuery(snapshot: Snapshot | SnapshotRecordArray): object {
```
igorT, line 1218: Can we have a better return type here

packages/adapter/addon/rest.ts
```diff
@@ -501,7 +511,7 @@ class RESTAdapter extends Adapter.extend(BuildURLMixin) {
     @param {Snapshot} snapshot
     @return {Promise} promise
   */
-  findRecord(store, type, id, snapshot) {
+  findRecord(store: Store, type: ShimModelClass, id: string, snapshot: Snapshot) {
```
igorT, line 514: Needs a return type

packages/adapter/addon/-private/fastboot-interface.ts
snewcomer: What do you have in mind here?  

Re: cookies...

https://github.com/ember-fastboot/fastboot/blob/9b69cf5a6f093d8068c4108684471dcac296e375/src/fastboot-request.js#L42

An interface for headers I guess could include a few like Content-Type and the like with a catch all.  Is this the idea?

snewcomer: Just to confirm, you generally want only the used properties declared as a part of this interface?

packages/adapter/addon/index.ts
```diff
@@ -55,7 +64,7 @@ import EmberObject from '@ember/object';
   @class Adapter
   @extends EmberObject
 */
-export default class Adapter extends EmberObject {
+export default class Adapter extends EmberObject implements MinimumAdapterInterface {
```
igorT, line 67: Lots of methods here are still missing the return type

packages/adapter/addon/json-api.ts
```diff
@@ -8,6 +8,11 @@ import { pluralize } from 'ember-inflector';
 import { serializeIntoHash } from './-private';
 import RESTAdapter from './rest';
 
+type Dict<T> = import('@ember-data/store/-private/ts-interfaces/utils').Dict<T>;
```
igorT, line 11: Do we need <T>

packages/adapter/addon/json-api.ts
```diff
@@ -221,7 +226,7 @@ class JSONAPIAdapter extends RESTAdapter {
   */
   coalesceFindRequests = false;
 
-  findMany(store, type, ids, snapshots) {
+  findMany(store: Store, type: ShimModelClass, ids: string[], snapshots: Snapshot[]) {
```
igorT, line 229: needs a return type

packages/adapter/addon/json-api.ts
```diff
@@ -231,7 +236,7 @@ class JSONAPIAdapter extends RESTAdapter {
     return pluralize(dasherized);
   }
 
-  updateRecord(store, type, snapshot) {
+  updateRecord(store: Store, type: ShimModelClass, snapshot: Snapshot) {
```
igorT, line 239: also here

packages/private-build-infra/addon/current-deprecations.ts
```diff
@@ -15,7 +15,7 @@
  * An app can use a different version than what it specifies as it's compatibility
  * version. For instance, an App could be using `3.16` while specifying compatibility
  * with `3.12`. This would remove any deprecations that were present in or before `3.12`
- * but keep support for anything deprecated in or abvoe `3.13`.
+ * but keep support for anything deprecated in or above `3.13`.
```
igorT, line 18: good catch

packages/adapter/addon/rest.ts
```diff
@@ -902,11 +938,11 @@ class RESTAdapter extends Adapter.extend(BuildURLMixin) {
     @param  {Object} requestData - the original request information
     @return {Object | AdapterError} response
   */
-  handleResponse(status, headers, payload, requestData) {
+  handleResponse(status: number, headers, payload: Payload, requestData: RequestData) {
     if (this.isSuccess(status, headers, payload)) {
       return payload;
     } else if (this.isInvalid(status, headers, payload)) {
-      return new InvalidError(payload.errors);
+      return new InvalidError((payload as Dict<unknown>).errors);
```
igorT, line 945: Why do we need a cast here

packages/adapter/addon/rest.ts
```diff
@@ -998,13 +1034,13 @@ class RESTAdapter extends Adapter.extend(BuildURLMixin) {
...
         .then(response => {
           _response = response;
-          return determineBodyPromise(response, requestData);
+          return determineBodyPromise(response, requestData as JQueryAjaxSettings);
         })
-        .then(payload => {
+        .then((payload: Payload) => {
           if (_response.ok && !(payload instanceof Error)) {
-            return fetchSuccessHandler(adapter, payload, _response, requestData);
+            return fetchSuccessHandler(adapter, payload, _response, requestData as FetchRequestData);
```
igorT, line 1041: Why do we need casts here

packages/adapter/addon/rest.ts
```diff
@@ -1063,7 +1099,7 @@ class RESTAdapter extends Adapter.extend(BuildURLMixin) {
     @param {Object} options
     @return {Object}
   */
-  ajaxOptions(url, method, options) {
+  ajaxOptions(url: string, method: string, options: ConfidentDict<any>) {
```
igorT, line 1102: options is a Dict<unknown> right?

packages/adapter/addon/json-api.ts
snewcomer: Yep.  The T in `type Dict<T>` is defining the generic and T in `.Dict<T>` is passing in that generic as an argument.

packages/adapter/addon/rest.ts
snewcomer: Payload can be `string | Dict<unknown> | undefined`

packages/adapter/addon/rest.ts
snewcomer: Ya that seems right.  The non class `function ajaxOptions` is a ConfidentDict.

packages/adapter/addon/rest.ts
snewcomer: You make a good point.  I [reworked](https://github.com/emberjs/data/pull/7358/commits/44b02c63c3f3a24fd23df566405476c714a731b6#diff-932693a0d32639d2ef4116b0d4060d31d063f287c74b53bd03466b4a2eb9407fR1038) this.  Cast'ed it because requestData was either for fetch or jquery.  Now each declaration of `requestData` is in it's own conditional and typed accordingly.

packages/adapter/addon/rest.ts
```diff
@@ -1146,12 +1145,14 @@ class RESTAdapter extends Adapter.extend(BuildURLMixin) {
       options = ajaxOptions(options, this);
     }
 
-    options.url = this._ajaxURL(options.url);
+    if (options.url) {
```
igorT, line 1155: Do you think this if changes things?

packages/adapter/addon/rest.ts
```diff
@@ -1463,7 +1464,7 @@ export function fetchOptions(options: ConfidentDict<any>, adapter: RESTAdapter):
   return options;
 }
 
-function ajaxOptions(options: ConfidentDict<any>, adapter): ConfidentDict<any> {
+function ajaxOptions(options: JQueryAjaxSettings, adapter: RESTAdapter): JQueryAjaxSettings {
```
igorT, line 1476: nice

packages/adapter/addon/rest.ts
```diff
@@ -1369,7 +1369,9 @@ function fetchErrorHandler(
     payload = responseData.errorThrown.payload;
   } else {
     responseData.errorThrown = errorThrown;
-    payload = adapter.parseErrorResponse(payload as string);
+    if (typeof payload === 'string') {
```
igorT, line 1372: This seems a bit sketch. We probably want to always call `parseErrorResponse`

snewcomer: The trouble I was coming across here is JSON.parse `undefined` (`Payload: string | undefined | null`).  I know we have the `catch` there but then JSON.parse will complain if we pass anything other than `string`.  Have any thoughts on how we can get around this?

packages/adapter/addon/rest.ts
snewcomer: Ok fixed this!  I created a new variable `reqOptions` with our own "options" type extending the JQuery and Fetch Init typings.

https://github.com/emberjs/data/pull/7358/commits/6a9bf658a51c8eb5338a7ae0f798f21068713278

packages/adapter/addon/rest.ts
```diff
@@ -1063,8 +1117,8 @@ class RESTAdapter extends Adapter.extend(BuildURLMixin) {
     @param {Object} options
     @return {Object}
   */
-  ajaxOptions(url, method, options) {
-    options = assign(
+  ajaxOptions(url: string, method: string, options: JQueryAjaxSettings | RequestInit): Dict<any> {
+    let reqOptions: JQueryRequestInit | FetchRequestInit = assign(
```
snewcomer, line 1121: I took another look at how to type cast this to avoid forcing the cast below. On the face of it, since all these options are optional, I'm not sure we have a reliable way to detect.

https://developer.mozilla.org/en-US/docs/Web/API/Request/Request#Parameters
https://api.jquery.com/jquery.ajax/#jQuery-ajax-settings

packages/adapter/addon/json-api.ts
```diff
@@ -8,8 +8,8 @@ import { pluralize } from 'ember-inflector';
 import { serializeIntoHash } from './-private';
 import RESTAdapter from './rest';
 
-type FetchRequestInit = import('./rest').FetchRequestInit;
-type JQueryRequestInit = import('./rest').JQueryRequestInit;
+type CustomFetchOptions = import('./rest').CustomFetchOptions;
+type CustomJQueryAjaxSettings = import('./rest').CustomJQueryAjaxSettings;
```
snewcomer, line 12: or `ExtendedFetchOptions`?

PR #7363, opened 2020-10-30
[BUGFIX identifiers] Address issue with polymorphic findRecord 

igorT:
Fixes issue when encountering polymorphic records with a subtype on initial load.
Builds on #7325

### DIFF
--- a/packages/store/addon/-private/identifiers/cache.ts
+++ b/packages/store/addon/-private/identifiers/cache.ts
@@ -1,4 +1,5 @@
 import { warn } from '@ember/debug';
+import { assign } from '@ember/polyfills';
 import { DEBUG } from '@glimmer/env';
 
 import coerceId from '../system/coerce-id';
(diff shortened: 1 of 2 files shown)


packages/store/addon/-private/identifiers/cache.ts
```diff
@@ -343,6 +343,15 @@ export class IdentifierCache {
     let newId = coerceId(data.id);
     let existingIdentifier = detectMerge(this._cache.types, identifier, data, newId, this._cache.lids);
 
+    if (!existingIdentifier) {
+      if (identifier.type !== data.type && data.type) {
```
krasnoukhov, line 347: There's `if (existingIdentifier)` statement below, maybe it would be reasonable to have this as `else if ` after that?

I'd also prefer to add a comment about what sort of case this is meant to deal with, but totally up to you folks

krasnoukhov (approved):

krasnoukhov:
Hey @snewcomer @igorT, is there anything I can do to help merge this?

packages/store/addon/-private/identifiers/cache.ts
igorT: 1. We still want the other statement to run, in fact the whole point of doing this here is to get line 357 
`      identifier = this._mergeRecordIdentifiers(keyOptions, identifier, existingIdentifier, data, newId as string);` to perform the merge

igorT: 2. Great point re: comment, adding that and merging

krasnoukhov: OMG makes total sense. Thank you

krasnoukhov:
I see that you've merged this into master and into 3.23, but do you think this is something that can also go into 3.20 LTS?

PR #7372, opened 2020-11-04
Do not explode when getting back an empty payload with the accept json header

igorT:
Even though this goes against the spec, we decided that it is better to be more accepting in the incoming requests, as the downside seemed low, and people seem to be running into this issue when migrating away from jquery to fetch.


snewcomer (approved):
Looks dandy!

PR #7378, opened 2020-11-09
[Deps]: ember-cli-typescript 4.0.0

snewcomer:
https://github.com/typed-ember/ember-cli-typescript/blob/master/CHANGELOG.md#400---2020-09-02

Minor asset size bump due to changes with ember-cli-terser and changes made to ember-cli-typescript to allow ember-cli-babel to manage TS transpilation.

https://github.com/typed-ember/ember-cli-typescript/pull/1018


package.json
```diff
@@ -84,7 +84,7 @@
     "ember-cli-string-utils": "^1.1.0",
     "ember-cli-template-lint": "^2.0.2",
     "ember-cli-test-loader": "^3.0.0",
-    "ember-cli-typescript": "^3.1.3",
+    "ember-cli-typescript": "^4.0.0",
```
rwjblue, line 87: Doesn't this need to be updated in each of the packages `package.json`s too?

igorT:
I have pinged @mike-north for thoughts

package.json
```diff
@@ -43,7 +43,7 @@
     "@ember/optional-features": "^1.3.0",
     "@glimmer/component": "^1.0.0",
     "@tracerbench/core": "4.0.10",
-    "@types/ember": "^3.1.1",
+    "@types/ember": "^3.16.2",
```
mike-north, line 46: This is a significant change. If we don't have any negative test cases for type information in this repo, I would spot check a few things to ensure that there aren't any regressions that inappropriately widen types (i.e., something becomes an `any` when it should be something more specific -- obviously a compiler will not object to this, but it's still a problem)

packages/model/addon/-private/system/many-array.js
```diff
@@ -176,7 +176,7 @@ export default EmberObject.extend(MutableArray, DeprecatedEvented, {
   },
 
   set length(value) {
-    return (this._length = value);
+    this._length = value;
```
igorT, line 179: @snewcomer did you mean to include this

PR #7394, opened 2020-12-03
[Cleanup]: Remove RECORD_ARRAY_MANAGER_IDENTIFIERS

snewcomer:
This flag helped us move from using Internal Models to Identifiers.  With the 3.23 release, we can remove this flag.


snewcomer:
A few possible paths here.  1 - we could consider this feature flag done and any further work to remove internalModel could be under new feature flags.  2 - close this PR and find more spots to remove internalModel (core store?).

One reason we might consider this feature flag completed is because we have ✂️  the use of `internalModel` in record-array and associated modules.

packages/canary-features/addon/default-features.ts
```diff
@@ -177,6 +177,5 @@ export default {
   REQUEST_SERVICE: null,
   CUSTOM_MODEL_CLASS: null,
   FULL_LINKS_ON_RELATIONSHIPS: true,
-  RECORD_ARRAY_MANAGER_IDENTIFIERS: true,
   REMOVE_RECORD_ARRAY_MANAGER_LEGACY_COMPAT: null,
```
igorT, line 180: We probably want to keep the flag exposed

packages/canary-features/addon/index.ts
```diff
@@ -38,7 +38,6 @@ export const REQUEST_SERVICE = featureValue(FEATURES.REQUEST_SERVICE);
 export const IDENTIFIERS = featureValue(FEATURES.IDENTIFIERS);
 export const CUSTOM_MODEL_CLASS = featureValue(FEATURES.CUSTOM_MODEL_CLASS);
 export const FULL_LINKS_ON_RELATIONSHIPS = featureValue(FEATURES.FULL_LINKS_ON_RELATIONSHIPS);
```
igorT, line 40: We probably want to keep the flag exposed

packages/store/addon/-private/system/references/reference.ts
```diff
@@ -32,15 +32,10 @@ function isResourceIdentiferWithRelatedLinks(
   return value && value.links && value.links.related;
 }
 
-// TODO: simplify after 3.23 release and only store identifier
-export const REFERENCE_CACHE = new WeakMap<Reference, InternalModel | StableRecordIdentifier>();
+export const REFERENCE_CACHE = new WeakMap<Reference, StableRecordIdentifier>();
```
igorT, line 35: Thanks for the cleanup

igorT:
@snewcomer we should use a new feature flag for new units of work, so they are easier to back out of. A lot of cleanup might not even need it, depending on how risky it is

PR #7395, opened 2020-12-04
[CHORE] Upgrade ember-inflector to version using babel 7

nlfurniss:
Upgrade `ember-inflector` to a version using babel 7.

**EDIT**: Do I need to bump this dep in all the `packages/unpublished-*`? Would love to remove v3 from the yarn.lock file...


igorT:
@nlfurniss yeah bumping it in all package.json's would be great

PR #7396, opened 2020-12-05
Upgrade ember-inflector in all other packages

nlfurniss:
Finishes what #7395 started.

In case anyone else runs into this error upgrading packages

```bash
cd packages/XXXX
yarn add <pkg-name>
error An unexpected error occurred: "expected workspace package to exist for XXXX
```

You can use `lerna add <package>[@version] --scope=<workspace>` :-)


nlfurniss:
@igorT seems to fail with ember canary on

``` 19) [Chrome 87.0] autotracking has-many: We can re-render hasMany with sort computed macro
     (0 , _destroyables.isDestroyed) is not a function

     TypeError: (0 , _destroyables.isDestroyed) is not a function
```

sandstrom:
@nlfurniss I'm guessing now, but could be related to https://emberjs.github.io/rfcs/0580-destroyables.html

I think there may be an expectation that any class instance with a container managed lifecycle must implement this. So basically, things you can grab via appInstance, for example `appInstance.lookup('service:store')` or `appInstance.lookup('serializer:json-api')`.

So it may not be related to the inflector package (again, it's a guess though).

snewcomer (approved):
Thank you!!

runspired:
Replaced with #7468 after this became more pressing. Thanks for putting this forward!

PR #7399, opened 2020-12-14
Fix deprecation URLs

simonihmig:
The colon in the deprecation ID is transformed into a dash for the URL of the deprecations app, so the URLs were not pointing to the correct section.


lifeart (approved):

snewcomer (approved):
👍 👍

PR #7406, opened 2020-12-29
import deprecate from ember-deprecations

BryanCrotaz:

### DIFF
--- a/packages/-ember-data/addon/setup-container.js
+++ b/packages/-ember-data/addon/setup-container.js
@@ -1,5 +1,6 @@
-import { deprecate } from '@ember/debug';
 import { DEBUG } from '@glimmer/env';
+
+import { deprecate } from 'ember-deprecations';
 
 import Store from '@ember-data/store';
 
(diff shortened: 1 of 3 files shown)


igorT:
Do we need to add a dependency on ember-deprecations?

BryanCrotaz:
it used to be imported from @ember/debug, but that doesn't work any more.

igorT:
@BryanCrotaz my question is more, how come we import from `ember-deprecations` but don't add it to `package.json`?

snewcomer:
Also, do we have a link or RFC describing this change in behaviour?  I'm also unable to find this package.

It looks like this function is still around...

https://api.emberjs.com/ember/3.24/functions/@ember%2Fdebug/deprecate

packages/-ember-data/addon/setup-container.js
```diff
@@ -1,6 +1,7 @@
-import { deprecate } from '@ember/debug';
```
rwjblue, line 1: The original import path is correct, we should not change it.

Can you explain what made you want to update?

rwjblue (changes requested):

runspired:
Concurring with @rwjblue import from `@ember/deprecate` is the correct location. https://api.emberjs.com/ember/3.25/functions/@ember%2Fdebug/deprecate

Some more context on the issue this was attempting to address in a bug ticket would be helpful if this has remained an issue for your app.

PR #7417, opened 2021-01-05
[CI]: simplify test run with smaller footprint for browser tests

snewcomer:
A proposal.  Often, I'm seeing these type of tests flakey.

This PR removes all the different scenarios for `browser tests` and just does on production test run.

Screen Shot 2021-01-05 at 8 46 24 AM

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -147,7 +147,6 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        scenario: [with-native-fetch, with-ember-fetch-no-jquery, with-ember-fetch-and-jquery, with-jquery]
         launcher: [Chrome, Firefox, IE]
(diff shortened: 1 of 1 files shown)


igorT:
I think it might be more important to test scenarios than the browser/OS matrix. Does this PR test the jquery/no-jquery scenarios in a different way?

snewcomer:
So I did not touch the `additional-scenarios` block.  Just removed tagging along the additional scenarios for the `browser-tests` block.  (I hope that is what you were asking 😆)

https://github.com/emberjs/data/blob/8545265fb31b4dae2941c42fee6205e152549fd7/.github/workflows/main.yml#L245-L269

runspired:
This is probably fine; however, the reason I'd done this per-browser was just in case any browsers still treated some xmlhttprequest or fetch requests differently. They probably don't at this point.

PR #7425, opened 2021-02-08
Pass lid from relationship data to get record data

### LINKED ISSUE #7424: Sideposting with lid leaves relationship in invalid state (belongsTo case)
#7127 demonstrates the problem for `hasMany` relationships. Unfortunately, the same issue is present for `belongsTo` relationships and requires a fix.

azhiv:
This PR aims to fix #7424.
The issue appeared in the sideposting scenario:

- model A has belongs to relationship to model B.
- instance of model A and instance of model B are created on the client,
- model A is saved along with model B sent in `include` section of payload; the linking is done through `lid`
- server persisted both models and sent the response with server-generated `id` along with `lid` sent by the client.
- when belongsTo.updateData() was called while the response payload was pushed to the store it couldn't match the record which was already in the store with the record data contained in the relationship data  because `lid` was omitted.

Add a test which ensures that the belongsTo() relationship has correct status in the sideposting with lid scenario.

### DIFF
--- a/packages/-ember-data/tests/integration/identifiers/lid-reflection-test.ts
+++ b/packages/-ember-data/tests/integration/identifiers/lid-reflection-test.ts
@@ -154,7 +154,7 @@ module('Integration | Identifiers - lid reflection', function(hooks) {
 
     class Cake extends Model {
       @attr name;
-      @hasMany('ingredient', { async: false }) ingredients;
+      @hasMany('ingredient', { inverse: null, async: false }) ingredients;
     }
 
     this.owner.register('model:ingredient', Ingredient);
@@ -165,8 +165,10 @@ module('Integration | Identifiers - lid reflection', function(hooks) {
         return payload;
       }
     }
+
     class TestAdapter extends Adapter {
       createRecord(store, ModelClass, snapshot) {
+        const lid = recordIdentifierFor(snapshot.record.ingredients.firstObject).lid;
         return resolve({
           data: {
             type: 'cake',
@@ -180,7 +182,7 @@ module('Integration | Identifiers - lid reflection', function(hooks) {
                   {
                     type: 'ingredient',
                     id: '2',
-                    lid: cheeseIdentifier.lid,
+                    lid,
                   },
                 ],
               },
@@ -190,7 +192,7 @@ module('Integration | Identifiers - lid reflection', function(hooks) {
             {
               type: 'ingredient',
               id: '2',
-              lid: cheeseIdentifier.lid,
+              lid,
               attributes: {
                 name: 'Cheese',
               },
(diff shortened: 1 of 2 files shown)


azhiv:
@snewcomer Can you please review this PR?

azhiv:
@snewcomer Can you please review this PR?

packages/-ember-data/tests/integration/identifiers/lid-reflection-test.ts
```diff
@@ -220,4 +220,76 @@ module('Integration | Identifiers - lid reflection', function(hooks) {
...
+        });
+      }
+    }
+    this.owner.register('serializer:application', TestSerializer);
+    this.owner.register('adapter:application', TestAdapter);
+
+    const cheese = store.createRecord('topping', { name: 'Cheese' });
+    const cake = store.createRecord('cake', { name: 'Cheesecake', topping: cheese });
+
+    const cheeseIdentifier = recordIdentifierFor(cheese);
```
runspired, line 288: depending on these costs getting hoisted is an anti pattern we should avoid, move these declarations above their first usage (in TestAdapter)

runspired (changes requested):
Overall LGTM but one cleanup desired for the test.

packages/-ember-data/tests/integration/identifiers/lid-reflection-test.ts
runspired: (alternatively declare `cheeseIdentifier` as a `let` above and then assign it here. Either is fine but hoisting an assignment is bad.

azhiv: Absolutely agree, done.

runspired: @azhiv I think you forgot to push the changes up

azhiv: @runspired In fact I committed and pushed to another branch, accidentally. The changes are now on master.

runspired (approved):
I'm good with this and will merge once tests pass (had to rekick them) but want to give @igorT a heads up before merging as this is the sort of thing that can sometimes introduce subtle issues.

packages/-ember-data/tests/integration/identifiers/lid-reflection-test.ts
azhiv: @runspired Both adapter and serializer need to be registered before calling `createRecord`, otherwise those will be resolved to default json-api implementations, which will then be stored in ED's cash. This modification breaks the tests, I reverted the change. The tests pass now.

azhiv: @runspired this required an additional fix. The order of method calls is crucial.

andreyfel:
@igorT the branch is green! Can we get it merged?

runspired (approved):

PR #7434, opened 2021-02-19
[Task]: remove meaningless assertion

### LINKED ISSUE #7433: Meaningless assertion
The assertion here is meaningless since the promise will always be defined:

https://github.com/emberjs/data/blob/fe844e484b0200f6ea0e8d7a5efdb79e81f7416a/packages/store/addon/-private/system/core-store.ts#L3678-L3681

It became meaningless here:

https://github.com/emberjs/data/commit/a24cd5011e2f3f9ec910a373400ca6cf1d98afb8#diff-60c8ac6f6fbeb0edd4df31a06f726f0a425f20b6241e73141132974d7070a7e9L2883

snewcomer:
close #7433

### DIFF
--- a/packages/store/addon/-private/system/core-store.ts
+++ b/packages/store/addon/-private/system/core-store.ts
@@ -3675,11 +3675,6 @@ function _commit(adapter, store, operation, snapshot) {
   let serializer = store.serializerFor(modelName);
   let label = `DS: Extract and notify about ${operation} completion of ${internalModel}`;
 
-  assert(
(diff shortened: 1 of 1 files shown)


runspired:
I suspect what we want is to move this assert within the promise wrapper to restore its original intent. Since asserts are stripped this isn't costing us anything currently but it is a good catch and I'm curious what debugging scenario @wagenet was in when he made this realization.

wagenet:
@runspired I was just reading the source trying tp figure out how to do something unrelated and was confused for a moment.

runspired (approved):

PR #7436, opened 2021-03-01
[CHORE] auto publish alpha versions

spham92:
Hi folks,

Similar to https://github.com/emberjs/ember.js/commit/0079b11d0aaec9980aa5e24d165653faf942a791, I'm adding a cron job that runs at 12pm every Monday that checks out the current master and run the publish script for alpha versions.

Repo owners need to add `NODE_AUTH_TOKEN` to the secrets for this to fully work.

### DIFF
--- a/.github/workflows/alpha-release.yml
+++ b/.github/workflows/alpha-release.yml
@@ -1,0 +1,45 @@
+name: 'Alpha Release'
+
+on:
+  schedule:
+    - cron: '0 20 * * 1' # weekly (Monday)
+  push:
+    branches:
+      - master
+
+jobs:
+  test:
+    name: Test latest code
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-node@v2-beta
+        with:
+          node-version: 12.x
+          repo-token: ${{ secrets.GITHUB_TOKEN }}
+      - name: Install dependencies for master
+        run: yarn install --frozen-lockfile --non-interactive
+      - name: Basic Tests
+        env:
+          CI: true
+          ASSERT_ALL_DEPRECATIONS: true
+        run: yarn test
+
+  release:
+    name: Run publish script
+    runs-on: ubuntu-latest
+    needs: [ test ]
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-node@v2
+        with:
+          node-version: 12.x
+          repo-token: ${{ secrets.GITHUB_TOKEN }}
+      - name: Install dependencies for master
+        run: yarn install --frozen-lockfile --non-interactive
+      - name: publish with script
+        run: node bin/publish.js canary
+        env:
+          NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
+      - name: push branch + tag
+        run: git push origin HEAD --follow-tags
--- a/.github/workflows/alpha-release.yml
+++ b/.github/workflows/alpha-release.yml
@@ -2,10 +2,7 @@ name: 'Alpha Release'
 
 on:
   schedule:
-    - cron: '0 20 * * 1' # weekly (Monday)
-  push:
-    branches:
-      - master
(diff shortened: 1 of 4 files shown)


.github/workflows/alpha-release.yml
```diff
@@ -0,0 +1,42 @@
...
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-node@v2
+        with:
+          node-version: 12.x
+          repo-token: ${{ secrets.GITHUB_TOKEN }}
+      - name: Install dependencies for master
+        run: yarn install --frozen-lockfile --non-interactive
+      - name: publish with script
+        run: node bin/publish.js canary
```
rwjblue, line 38: Don't we need to `--distTag` here? Also, what ensures that the version number includes `-alpha.0`?

rwjblue: second question is not an issue:

https://github.com/emberjs/data/blob/fe844e484b0200f6ea0e8d7a5efdb79e81f7416a/bin/publish.js#L231-L236

spham92: Hmm good point, let me figure out some code to follow: 

1. If current version is a minor version release such as `3.15.0`, the version picked for the next alpha is `3.16.0-alpha.0`
2. If current version is an alpha release such as `3.15.1-alpha.0`, the version picked for the next alpha is `3.15.1-alpha.1`
3. Beta releases would not affect the versioning of the next alpha version as the alpha will continue from the last alpha release.

igorT:
@spham92 the yaml file seems to error out https://github.com/emberjs/data/actions/runs/706000564/workflow

bin/publish.js
```diff
@@ -308,22 +310,41 @@ let cli = readline.createInterface({
...
+ * @param {string} otp - Token to make publish requests to npm
+ */
+function publishPackage(distTag, tarball, otp) {
+  let cmd = `npm publish ${tarball} --tag=${distTag} --access=public`;
+
+  if (otp) {
+    cmd += ` --otp=${otp}`;
+  }
+
+  console.log(cmd);
```
igorT, line 328: This log seems like it might get out of sync with the actual command. Do we still want this, and if so, lets use the same line in the exec. It seems like if no otp, cmd will not have it but it will still be in the execWithLog?

bin/publish.js
```diff
@@ -362,7 +391,9 @@ async function main() {
   }
   if (!options.skipPublish) {
     const tarballs = collectTarballPaths();
-    await confirmPublish(tarballs, nextVersion);
+    const npmAuthTokenInEnv = !!process.env.NODE_AUTH_TOKEN;
+    // Assume human ran script if token is missing
```
igorT, line 395: Can we log this please

bin/publish.js
spham92: ahh console log should've been removed. Changed to:

```
function publishPackage(distTag, tarball, otp) {
  let cmd = `npm publish ${tarball} --tag=${distTag} --access=public`;

  if (otp) {
    cmd += ` --otp=${otp}`;
  }

  execWithLog(cmd);
}
```

bin/publish.js
spham92: added

runspired:
@spham92 apparently the version increment script here is buggy. Last night instead of just incrementing the alpha counter it decremented the minor version as well. https://github.com/emberjs/data/commit/2feb0c1a9f2e54fbbcdeaba0d62468a4a5dadebb

spham92:
> @spham92 apparently the version increment script here is buggy. Last night instead of just incrementing the alpha counter it decremented the minor version as well. [2feb0c1](https://github.com/emberjs/data/commit/2feb0c1a9f2e54fbbcdeaba0d62468a4a5dadebb)

Gotcha, will take a look into getting it working correctly

runspired:
@spham92 I pinged you on the revert PR, describe three issues I spotted there. Thanks for your work on this!

PR #7447, opened 2021-03-30
Document adapterOptions on REST adapter

sandstrom:
Seems like we aren't documenting adapterOptions for `query` and `queryRecord` in the JSONAPIAdapter.

I've just mirrored the code in the regular adapter:
https://github.com/emberjs/data/blob/master/packages/adapter/addon/index.ts#L248

There are already tests for this, see:
https://github.com/emberjs/data/blob/ff4f9111fcfa7dd9e39804ed17f5af27a4a01378/packages/-ember-data/tests/integration/adapter/store-adapter-test.js#L1526

Support for `adapterOptions` has existed for quite a while, it was initially added in this PR: https://github.com/emberjs/data/pull/4856/

### DIFF
--- a/packages/adapter/addon/rest.ts
+++ b/packages/adapter/addon/rest.ts
@@ -631,7 +631,12 @@ class RESTAdapter extends Adapter.extend(BuildURLMixin) {
     @param {Object} adapterOptions
     @return {Promise} promise
   */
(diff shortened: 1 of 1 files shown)


sandstrom:
Thanks @igorT I hadn't gotten to it yet 😄 

Also, have you thought about joining the Github sponsorship program? We'd like to donate to some of the core Ember (and Ember Data) contributors, and you're definitely one of them. I know it's not much, but hopefully more companies will join in eventually.

PR #7448, opened 2021-04-01
Bugfix: coalesceFindRequests should work with non native Adapter classes

### LINKED ISSUE #7438: `coalesceFindRequests: true` regression when using classic `RESTAdapter`
### Reproduction

- create "classic" RESTAdapter in `app/adapters/application.js`:
  ```
  export default RESTAdapter.extend({
      coalesceFindRequests: true
  });
  ```
- Run app
- Check `lookup('adapter:application').coalesceFindRequests` at runtime: it's `false`. Expected `true`.

### Description

I have upgraded ember-data from 3.10 to 3.24. And now `coalesceFindRequests: true` doesn't work.
It seems that `coalesceFindRequests` is overridden in constructor: 
image

### Versions

ember-data 3.10 - `coalesceFindRequests: true` works.
ember-data 3.24 - `coalesceFindRequests: true` doesn't work.

snewcomer:
This PR ensures that a user with non native class Adapter will be able to define their own `coalesceFindRequests` field.

Will patch to release and beta.

close #7438 

ref #7358

### DIFF
--- a/packages/-ember-data/tests/integration/adapter/rest-adapter-test.js
+++ b/packages/-ember-data/tests/integration/adapter/rest-adapter-test.js
@@ -47,7 +47,6 @@ module('integration/adapter/rest_adapter - REST Adapter', function(hooks) {
 
     server = new Pretender();
     store = this.owner.lookup('service:store');
-    adapter = store.adapterFor('application');
 
     passedUrl = passedVerb = passedHash = null;
   });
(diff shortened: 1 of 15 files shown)


packages/adapter/addon/index.ts
```diff
@@ -470,6 +470,8 @@ export default class Adapter extends EmberObject implements MinimumAdapterInterf
     return RSVPPromise.resolve();
   }
 
+  _coalesceFindRequests = true;
```
snewcomer, line 473: We could use `#` private fields here but that will throw an error with non native classes

packages/adapter/addon/json-api.ts
```diff
@@ -231,7 +231,8 @@ class JSONAPIAdapter extends RESTAdapter {
     @property coalesceFindRequests
     @type {boolean}
   */
-  coalesceFindRequests: boolean = false;
+
+  _coalesceFindRequests: boolean = false;
```
snewcomer, line 235: We don't want `@private` documentation for this do we?

igorT:
@snewcomer thank you for the fix! Can you please add a test that verifies this fixes the reported issue

packages/-ember-data/tests/integration/adapter/rest-adapter-test.js
```diff
@@ -2137,697 +2242,688 @@ module('integration/adapter/rest_adapter - REST Adapter', function(hooks) {
...
-        {
-          id: 3,
-          name: 'Second comment',
-        },
-      ],
-    });
-    let post;
+  module('init adapter coalesceFindRequests', function() {
+    test('findMany - passes buildURL the requestType', function(assert) {
+      this.owner.register('adapter:application', RESTAdapter.extend({ coalesceFindRequests: true }));
```
snewcomer, line 2841: I think I might just move this to a unit test.  This test file was sort of a pain to work with...

1.  We need to call unregister to register a new adapter with the adapter declaration shown above.
2.  Calling `store.adapterFor` in beforeEach makes it hard to unregister and register a new application adapter since the store `_adapterCache` isn't cleared unless I force clear that private field - which doesn't seem like a good idea.  The approach I took here was to declare two module blocks.  

Still...a unit test is probably the way to go.

packages/adapter/addon/index.ts
```diff
@@ -66,6 +66,8 @@ type SnapshotRecordArray = import('@ember-data/store/-private/system/snapshot-re
   @extends EmberObject
 */
 export default class Adapter extends EmberObject implements MinimumAdapterInterface {
+  declare _coalesceFindRequests: boolean;
```
snewcomer, line 69: https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#the-usedefineforclassfields-flag-and-the-declare-property-modifier

runspired: WAT. I missed this discussion and I am sad. https://github.com/tc39/proposal-class-fields/issues/151#issuecomment-431597270

runspired: I think a better alternative here may be to do this within the constructor.

runspired (approved):
After exploring this is the only implementation that works.

While we could use properties within the constructor for the base Adapter implementation, we would not be able to toggle the default from `true` to `false` for the RESTAdapter and JSONAPIAdapter which extend it for both `.extends` and native class syntax.

A simpler implementation is only possible if we either drop the implementation in the base class entirely, change the base class to have the same default as the others, or stop extending from Adapter (we should probably actually do this, but can investigate it at another time).

PR #7449, opened 2021-04-08
[DOC] Add missing `Ember` namespace to annotations

tricknotes:
These classes are defined in ember package, not in ember-data.


snewcomer (approved):
Great catch!  Thank you.

PR #7450, opened 2021-04-08
Typescript 4.0

runspired:
Updates all things related to eslint, typescript-eslint, typescript, and types packages to be on the latest. Resolves issues.

This also cleans up the eslint file to take advantage of how overrides work (we were overly verbose).

Lastly it deactivates some rules:

- a few  new qunit rules we violate (some of which should possibly cleaned up, some of which our current pattern seems probably fine to me, but a refactor to keep to standard could be best anyway)
- and a few rules which seem to have been turned on in the eslint:recommended packages (prefer-const, prefer-rest-params) which we do not prefer to have on. prefer-const in particular is problematic because it does not detect lazy assignment or usage from an inner scope.
- usage of global Promise is now a warning for TS files and I've deactivated the rule in a few files. This is because the type return of `Promise` is now linted against as part of the no-restricted-globals rule. We should audit promise usage to ensure IE11 wouldn't have an issue but it doesn't seem like the biggest deal.


packages/-ember-data/tests/integration/adapter/find-all-test.js
```diff
@@ -114,7 +114,7 @@ module('integration/adapter/find-all - Finding All Records of a Type', function(
     assert.equal(all.objectAt(0).get('name'), 'Braaaahm Dale', 'the first item in the record array is Braaaahm Dale');
   });
 
-  test('When all records for a type are requested, records that are already loaded should be returned immediately.', async assert => {
+  test('When all records for a type are requested, records that are already loaded should be returned immediately.', async function(assert) {
```
igorT, line 117: Is this a lint complaint?

packages/-ember-data/tests/integration/adapter/queries-test.js
```diff
@@ -76,7 +76,7 @@ module('integration/adapter/queries - Queries', function(hooks) {
     let queryResults = await store.query('person', { page: 1 });
 
     assert.equal(queryResults.length, 2, 'the record array has a length of 2 after the results are loaded');
-    assert.equal(queryResults.isLoaded, true, "the record array's `isLoaded` property should be true");
+    assert.true(queryResults.isLoaded, "the record array's `isLoaded` property should be true");
```
igorT, line 79: Btw, whats the advantage of assert true vs equal to true?

packages/adapter/addon/index.ts
```diff
@@ -9,6 +9,8 @@ type ShimModelClass = import('@ember-data/store/-private/system/model/shim-model
 type Store = import('@ember-data/store/-private/system/core-store').default;
 type Snapshot = import('ember-data/-private').Snapshot;
 type SnapshotRecordArray = import('@ember-data/store/-private/system/snapshot-record-array').default;
+// eslint-disable-next-line no-restricted-globals
+type AdapterPromise<T> = Promise<T> | RSVPPromise<T>;
```
igorT, line 13: Why can't it be Promise<T>?  Do we sometimes rely on RSVP internals?

packages/store/addon/-private/system/core-store.ts
```diff
@@ -1393,7 +1392,8 @@ abstract class CoreStore extends Service {
       // will once again convert the records to snapshots for adapter.findMany()
       let snapshots = new Array(totalItems);
       for (let i = 0; i < totalItems; i++) {
-        snapshots[i] = internalModels[i].createSnapshot(optionsMap.get(internalModel));
+        let internalModel = internalModels[i];
```
igorT, line 1395: Why the multiline?

packages/store/addon/-private/system/core-store.ts
```diff
@@ -3179,7 +3183,7 @@ abstract class CoreStore extends Service {
       return new _RecordData(identifier, storeWrapper);
     }
 
-    assertInDebug(`Expected store.createRecordDataFor to be implemented but it wasn't`, false);
+    assert(`Expected store.createRecordDataFor to be implemented but it wasn't`);
```
igorT, line 3186: Will these be stripped in prod?

packages/store/addon/-private/system/store/finders.js
```diff
@@ -25,7 +25,7 @@ function payloadIsNotBlank(adapterPayload) {
 
 export function _find(adapter, store, modelClass, id, internalModel, options) {
   if (REQUEST_SERVICE) {
-    // assert here
+    assert(`Rquests made when REQUEST_SERVICE=true should not use the legacy finders`);
```
igorT, line 28: Typo

packages/store/addon/-private/ts-interfaces/minimum-adapter-interface.ts
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable no-restricted-globals */
```
igorT, line 1: Can you leave a comment why this is on please

package.json
```diff
@@ -38,21 +38,22 @@
     "test-external:ember-data-relationship-tracker": "./bin/test-external-partner-project.js ember-data-relationship-tracker https://github.com/ef4/ember-data-relationship-tracker.git"
```
igorT, line 38: @mike-north @rwjblue @chriskrycho do these version bumps look :+1:

packages/-ember-data/tests/integration/adapter/find-all-test.js
runspired: yes, and it auto fixed it too

packages/adapter/addon/index.ts
runspired: We sometimes return an RSVPPRomise and we sometimes return a Native Promise, this depends entirely on whether we've created the promise or whether we've received it from a native API.

This mixing is why I think we should do an audit at some point to make sure we haven't accidentally used a native promise somewhere, I'm fairly sure we haven't, but we do have to expect that a promise received/returned could be either.

fwiw `yarn problems` sees no issue with the return types having been just Promise.

typescript-eslint was complaining for things that were typed Promise as an argument or as a return, not because of a ts compiler error but due to the `no-restricted-globals` lint rule. Since we mix the returns I figured I'd go ahead and set the AdapterPromise to both and only disable the rule once vs for the entire file or at every usage of Promise as type. We can probably do this single-location lint override and only mirror Promise as AdapterPromise vs both types.

The best approach is a good question for @mike-north or @chriskrycho.

packages/store/addon/-private/system/core-store.ts
runspired: this was a bug that the upgraded lint and typescript tooling caught. internalModel is always undefined currently. Since we need to access it twice I cached it off.

packages/store/addon/-private/system/core-store.ts
runspired: yes!

packages/-ember-data/tests/integration/adapter/queries-test.js
runspired: huh, apparently the comment I left didn't save :/ 

basically its just a newer thing that's more explicit. There's a lint rule that auto fixes to this thing. I don't care a ton but I do find the new style a little quicker to read and since there's an autofix I'm all for it.

runspired: (comment meaning my first reply to you here)

packages/-ember-data/package.json
```diff
@@ -82,7 +82,7 @@
     "loader.js": "^4.7.0",
     "qunit": "^2.10.0",
     "qunit-dom": "^1.2.0",
-    "typescript": "~3.8.3"
+    "typescript": "~4.2.4"
```
chriskrycho, line 85: These should resolve to the same thing so I'm not especially worried by it, but it's slightly surprising to me to have different version requests for the top-level package vs. the nested package. I assume this was probably a case of 4.2.4 coming out while you were working on this PR, @runspired?

.eslintrc.js
```diff
@@ -87,21 +93,15 @@ module.exports = {
       parserOptions: {
         sourceType: 'module',
       },
-      plugins: ['@typescript-eslint', 'ember-data', 'simple-import-sort', 'import'],
-      extends: ['eslint:recommended', 'plugin:@typescript-eslint/eslint-recommended'],
+      plugins: ['@typescript-eslint', 'ember-data'],
+      extends: ['plugin:@typescript-eslint/eslint-recommended'],
       rules: {
+        'no-restricted-globals': ['off'],
```
chriskrycho, line 99: Question: are we good with *completely* disabling this, or do we want it to be for a specific subset, e.g. `Promise`? (I assume "completely" is fine, but figured I'd double-check!)

packages/store/addon/-private/system/core-store.ts
```diff
@@ -93,6 +93,7 @@ type SchemaDefinitionService = import('../ts-interfaces/schema-definition-servic
 type PrivateSnapshot = import('./snapshot').PrivateSnapshot;
 type Relationship = import('@ember-data/record-data/-private').Relationship;
 type RecordDataClass = typeof import('@ember-data/record-data/-private').RecordData;
+type RequestCache = import('./request-cache').default;
```
chriskrycho, line 96: Note for future fun: you *should* now be able to switch all of these kinds of imports over to the `import type` syntax:

```ts
import type RequestCache from './request-cache';
```

(I'd love to hear if that *doesn't* work with your current tooling configuration for some reason!)

packages/store/addon/-private/system/core-store.ts
```diff
@@ -1403,6 +1403,10 @@ abstract class CoreStore extends Service {
         groups = [snapshots];
       }
 
+      // we use var here because babel transpiles let
+      // in a manner that causes a mega-bad perf scenario here
+      // when targets no longer include IE11 we can drop this.
+      /* eslint-disable no-var */
```
chriskrycho, line 1409: Huh. Out of curiosity, is this still true? The Babel playground just compiles it to `for (var i = 0, ...)`:

Screen Shot 2021-04-13 at 09 22 46

yarn.lock
```diff
@@ -2580,6 +2577,32 @@
     "@types/jquery" "*"
     "@types/rsvp" "*"
 
+"@types/ember@^3.16.5":
```
chriskrycho, line 2580: It may be worth running a `yarn deduplicate` here or otherwise forcing a resolution; `@types/ember-testing-helpers` appears (per previous entry in the lock file) to be resolving `@types/ember@3.16.4`. Given CI is happy and your local build was as well, it seems like that's not currently causing an issue, but I have been bitten enough times that I always make sure there's only a single resolved version of any `@types/*` package where possible.

yarn.lock
```diff
@@ -15630,10 +15727,15 @@ typescript-memoize@^1.0.0-alpha.3:
...
-  integrity sha512-MYlEfn5VrLNsgudQTVJeNaQFUAI7DkhnOjdpAp4T+ku1TfQClewlbSuTVHiA+8skNBgaf02TL/kLOvig4y3G8w==
+typescript@~4.2.3:
+  version "4.2.3"
+  resolved "https://registry.npmjs.org/typescript/-/typescript-4.2.3.tgz#39062d8019912d43726298f09493d598048c1ce3"
+  integrity sha512-qOcYwxaByStAWrBf4x0fibwZvMRG+r4cQoTjbPtUlrWjBHbmCAww1i448U0GJ+3cNNEtebDteo/cHOR3xJ4wEw==
+
+typescript@~4.2.4:
+  version "4.2.4"
+  resolved "https://registry.npmjs.org/typescript/-/typescript-4.2.4.tgz#8610b59747de028fda898a8aef0e103f156d0961"
+  integrity sha512-V+evlYHZnQkaz8TRBuxTA92yZBPotr5H+WhQ7bD3hZUndx5tGOa1fuCgeSjxAzM1RiN5IzvadIXTVefuuwZCRg==
```
chriskrycho, line 15738: Curious: we are actually resolving two different versions of TS here! I'm surprised that it ended up resolving this way, but presumably you added them in separate passes? In any case, probably worth resolving to just 4.2.4?

chriskrycho (changes requested):
Caught a couple small things deps-wise (should be trivially fixed with a `yarn deduplicate`). Otherwise, 👍🏼!

package.json
```diff
@@ -38,21 +38,22 @@
...
   },
   "devDependencies": {
-    "@babel/plugin-transform-typescript": "^7.9.6",
+    "@babel/plugin-transform-typescript": "^7.13.0",
```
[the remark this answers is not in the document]
runspired: We may also have this due to our custom rollup step, I'll try to investigate soon, always happy to maintain less things.

.eslintrc.js
runspired: completely is fine (Promise was the only global we were checking for previously), also we cut to IE11 support in 4 weeks, but also this eslint rule does not appear to have a syntax that allows you to disable only one global, in which case we'd have to manually specify every global we wanted to check except for those we didn't.

packages/store/addon/-private/system/core-store.ts
runspired: I've been considering making the migration, we have a custom lint tule for type only imports that I'd want to update but it is a nicer import syntax :)   Does it also allow you to import types for multiple named declarations in one go?

packages/store/addon/-private/system/core-store.ts
runspired: I checked it and it was still true which is why I added the comment. Though the playground not compiling it to the closed over functions means there may be something for us to tweak that will allow us to remove this.

That said, on your left there note how you didn't change the `var` definitions in the for loops themselves. IIRC only two of these actually have to be vars for us to not transpile to crazy land (`ids` and `internalModel` or maybe it was `group`)

runspired: I checked the playground and if you convert all the vars to let and transpile to IE11 (bc modern doesn't transpile scope things anymore) you do hit the pathological perf case with functions redefined on every iteration that have their own loops redefined with functions etc.

yarn.lock
runspired: fixed

yarn.lock
runspired: done

packages/-ember-data/package.json
runspired: indeed, fixed

packages/store/addon/-private/system/core-store.ts
chriskrycho: Yep, it is perfectly symmetrical with regular import statements! So very much nicer that way.￼

PR #7451, opened 2021-04-09
[DOC] Fix `JSONAPISerializer` import in docs

piotrekbator:
An example in docs used to incorrectly import the serializer from `@ember-data/serializer/json` instead of `'@ember-data/serializer/json-api'`. This PR fixes that.


snewcomer (approved):
Thank you!

PR #7453, opened 2021-04-09
[CHORE] burndown of InternalModel methods that can be eliminated safely

runspired:
Each commit here removes a method or property from InternalModel for which either a different InternalModel codepath is just as efficient to use OR the codepath is unused OR the codepath has a non InternalModel equivalent of the same efficiency.

The motivation is to shrink the surface area of InternalModel until what remains is logic we need to move elsewhere, helping us get a clearer picture of what InternalModel is still handling for us. It comes with the added win of reducing our payload size slightly.

### DIFF
--- a/packages/store/addon/-private/system/model/internal-model.ts
+++ b/packages/store/addon/-private/system/model/internal-model.ts
@@ -919,14 +919,6 @@ export default class InternalModel {
     this.send('loadedData');
   }
 
-  /*
-    @method notFound
-    @private
-  */
-  notFound() {
-    this.send('notFound');
-  }
-
   hasChangedAttributes() {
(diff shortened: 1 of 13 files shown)


packages/store/addon/-private/system/model/internal-model.ts
```diff
@@ -328,10 +328,6 @@ export default class InternalModel {
     return this.currentState.hasDirtyAttributes;
   }
 
-  isSaving() {
-    return this.currentState.isSaving;
-  }
-
```
snewcomer, line 339: 👍

snewcomer (approved):
What an easy PR to review!  Thanks for making this as easy to review.

PR #7454, opened 2021-04-10
[PERF] Class Fields Use Optimization & Made OrderedSet Faster (Again)

runspired:
Evaluating some of our current perf-based decisions to see if they still ring true.

This started off to see if we can clean up some of our patterns, what I found was inconclusive (which means maybe?) but in the process I also found that using `declare` for class fields that are initialized in the constructor is a nice boost to asset size trimming and also discovered that we were still importing all of Ember to grab GUID_KEY in InternalModel which was no longer used. The latter reminded me that OrderedSet usage was probably de-optimized when we moved to RecordData and sure enough we get a ~4.6% win making sure the set is keyed efficiently again. I also found sadness, but that will hopefully get cleaned up soon. 

Findings:

- moving type only field declarations to use `declare` has no evident perf impact but significantly improves build size.
- moving some props from [[DECLARE]] to [[SET]] semantics had no evident perf impact, though this may be due to our babel config (need to investigate)
- moving 1 prop in a hot path from lazy getter to eager initializer of an array did not have any evident perf impact
- I discovered we could remove GUID_FOR usage entirely on InternalModel. It was only there for our use of OrderedSet
- speaking of which `RecordData` needed GUID_FOR, but I went a step further and ported OrderedSet in and tailored it to use `clientId` for now and eliminated our need for `guidFor` and `GUID_FOR`. This better prepares us for refactoring to identifiers anyway, plus it's actually a good sized win for bundle size.
- typing OrderedSet better revealed a few issues in Relationship.ts which have now been cleaned up

**Perf Win**

This is purely from having OrderedSet not need to call guidFor on recordData which does not have a GUID_KEY already.

```
  Benchmark Results Summary    

duration phase estimated improvement -94ms [-133ms to -54ms] OR -3.16% [-4.46% to -1.82%]
Phase [navigationStart] => [start-find-all] phase no difference [-14ms to 14ms]
Phase [start-find-all] => [start-outer-materialization] phase estimated improvement -81ms [-113ms to -50ms] OR -4.63% [-6.46% to -2.89%]
Phase [start-outer-materialization] => [stop-outer-materialization] phase no difference [-25ms to 6ms]
Phase [stop-outer-materialization] => [end-find-all] phase no difference [0ms to 0ms]
Phase [end-find-all] => [Test End] phase no difference [-2ms to 1ms]
```


packages/record-data/addon/-private/ordered-set.ts
```diff
@@ -29,12 +125,13 @@ export default class EmberDataOrderedSet<T> extends EmberOrderedSet<T> {
...
 
-  deleteWithIndex(obj: T | null, idx?: number): boolean {
+  deleteWithIndex(obj: RelationshipRecordData | null, idx?: number): boolean {
     let guid = guidFor(obj);
+    assert(`Expected ${obj} to have an clientId`, typeof guid === 'string' && guid !== '');
     let presenceSet = this.presenceSet;
     let list = this.list;
 
-    if (presenceSet[guid] === true) {
+    if (presenceSet[guid] === obj) {
```
snewcomer, line 134: This seems like a major change?  We are relying on referential equality now, which is different than if we have two instances with the same `guid`.

snewcomer, line 134: This seems like a major change?  We are relying on referential equality now, which is different than if we have two instances with the same `guid`.

packages/record-data/addon/-private/ordered-set.ts
```diff
@@ -1,22 +1,118 @@
...
+    }
+
+    set.list = this.toArray();
+    set.size = this.size;
+
+    return set;
   }
 
-  addWithIndex(obj: T, idx?: number) {
+  addWithIndex(obj: RelationshipRecordData | null, idx?: number) {
```
snewcomer, line 105: Do you happen to know why we have these `withIndex` methods?  It seems like a lot of methods are just looking up on the `presenceSet` by key when retrieving a value.

Otherwise, this data structure just be a simple `Map` perhaps.  It seems like we are emulating a lot of internal behaviour that Map provides - size, iteration, presence.  We would just need to implement our own copy.

snewcomer (commented):
Spicy!  I like this PR a lot.

packages/record-data/addon/-private/ordered-set.ts
```diff
@@ -1,22 +1,118 @@
...
+ While we convert relationships to identifiers this will be something we
+ will be forced to address.
+*/
+function guidFor(obj): string {
+  if (obj === null) {
+    return NULL_POINTER;
+  }
+
+  return obj.clientId || obj.lid;
+}
```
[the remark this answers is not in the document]
runspired: @snewcomer guidFor is the intended name for now. The `obj.lid` path is mostly here for when relationships store identifiers instead of RecordData instances. The concept of OrderedSet will go away soon so I don't think we need to rename this.

packages/record-data/addon/-private/ordered-set.ts
```diff
@@ -1,22 +1,118 @@
...
+
+export default class OrderedSet {
+  declare presenceSet: Dict<RelationshipRecordData | null>;
```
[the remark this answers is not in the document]
runspired: `null` was a mistake in our typing, though we probably shouldn't have been allowing it to be put into the set. I think it happens mostly because we use these for `BelongsTo` relationships as well in which the canonical / current state can often be `null`.

The referenced bug is that we end up keeping destroyed recordData instances in the relationships because the check to remove them fails. This is because when we resurrect a record right before it's discarded we actually hand it a new RecordData instance. Whether the bug is RecordData being a new instance or the bug is not removing the old RecordData, it happens because Ember's `guidFor` builds a new key for each thing it see by-reference. We don't actually care about by-reference here, we care about identity. But for now we've had to preserve the by-reference behavior.

packages/record-data/addon/-private/ordered-set.ts
```diff
@@ -1,22 +1,118 @@
...
+
+  clear() {
+    this.presenceSet = Object.create(null);
+    this.list = [];
+    this.size = 0;
+  }
+
+  add(obj: RelationshipRecordData | null): OrderedSet {
+    let guid = guidFor(obj);
+    assert(`Expected ${obj} to have an clientId`, typeof guid === 'string' && guid !== '');
```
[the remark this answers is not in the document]
runspired: NULL_POINTER is valid (see discussion above), empty string is not: it means somehow we've massively screwed something up.

packages/record-data/addon/-private/ordered-set.ts
runspired: we were already relying on referential equality but now it's explicit instead of implicit.

packages/record-data/addon/-private/ordered-set.ts
runspired: it's because (in theory) we want to preserve order and also (in theory) this is the only data structure within the relationship storing information. Unfortunately over time that became not true. It's something I've tried many times in the past to rectify and which the new rewrite I'm starting this week will try to rectify yet again.

As for why: these methods are called when a user does a splice/insertAt operation on the hasMany. The index is carried through and used here. They were historically also sometimes used when receiving data from the API that re-orders the records we have, though more commonly we blow away these instances and populate new instances in that case today.

packages/record-data/addon/-private/ordered-set.ts
```diff
@@ -1,22 +1,118 @@
 import { assert } from '@ember/debug';
```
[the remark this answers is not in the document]
runspired: Possibly, but I don't anticipate this lasting more than 2 weeks in the code. We've got a 4.0 deadline to hit.

packages/record-data/addon/-private/ordered-set.ts
snewcomer: Ah end user manipulation 👍

snewcomer (approved):
Thank you for the explanations!  I wasn't aware of the short term-ndess of the changes here as well but this looks great!

PR #7455, opened 2021-04-13
[BUGFIX] Updated the yuidocs for the broken ember-data links linking to elsewhere on the page.

broxsonl:
For [issue 737 in the ember-api-docs repository,](https://github.com/ember-learn/ember-api-docs/issues/737) it turned out the fix needed to be made in the yuidocs that are used to generate the markup for the ember-data docs page.

11 links throughout the page, which are intended to link internally to other parts of the page, were all broken because of an incorrect filepath in the docs. They prepended `classes/Store/` to the path unnecessarily.

I am uncertain of valid tests to add for this, as there doesn't appear to be any current testing for the yuidocs generating the docs page.

I also am aware that to fully update the docs--which will likely want to be done for all versions of Ember-Data's docs--the files on s3 will need to be replaced, which is not something I am able to generate or have access to, though would be happy to assist with given the blessing/credentials.

### DIFF
--- a/packages/store/addon/-private/system/core-store.ts
+++ b/packages/store/addon/-private/system/core-store.ts
@@ -187,12 +187,12 @@ function deprecateTestRegistration(
   The store provides multiple ways to create new record objects. They have
   some subtle differences in their use which are detailed below:
 
-  [createRecord](createRecord?anchor=createRecord) is used for creating new
+  [createRecord](../methods/createRecord?anchor=createRecord) is used for creating new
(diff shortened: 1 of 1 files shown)


runspired (approved):

runspired:
@broxsonl I'm unsure what to do about past versions or where to add tests for this (beyond the basic "we generated docs" tests that we have) but those aspects (sadly) aren't something we can do in this repo currently. Probably the learning team folks can help there. Thank you for this fix!

PR #7457, opened 2021-04-13
[CHORE] setup test suite for the record-data package

runspired:
Currently tests all live in the `ember-data` package. Over time we should move tests into the appropriate package. New tests should be added to the appropriate package. This begins that for RecordData.

The asset size increase is artificial, at some point the IE11 test stopped building for IE11 since the check for the env flag was removed. I've fixed this.

~~This does not currently configure the tests to run for ember-try scenarios, one approach for that would be to copy the ember-try config into each package and then use lerna to run all tests for packages that have a try command. I haven't checked yet but I'm hoping we can just keep a single ember-try config and point the try command to where it lives.~~ implemented

ember-source has it's own script for running tests across multiple packages. I don't think we need such a thing yet though it may help reduce maintenance of test-infra across packages were we to find ourselves needing to keep configs in multiple places in sync.


.eslintrc.js
```diff
@@ -75,7 +75,6 @@ module.exports = {
     'qunit/no-conditional-assertions': 'off',
   },
   globals: {
-    heimdall: true,
```
runspired, line 78: we haven't used heimdall since ~3.4

packages/adapter/package.json
```diff
@@ -15,8 +15,6 @@
   "scripts": {
     "build": "ember build",
     "start": "ember serve",
-    "test": "ember test",
-    "test:all": "ember try:each",
```
runspired, line 19: I removed the test commands from packages that we haven't setup testing infra in yet since we will need to update test-helpers/testem etc. still

packages/record-data/ember-cli-build.js
```diff
@@ -3,10 +3,23 @@
...
+      // this ensures that the same build-time code stripping that is done
+      // for library packages is also done for our tests and dummy app
+      plugins: [...require('@ember-data/private-build-infra/src/debug-macros')(null, isProd, compatWith)],
+    },
     'ember-cli-babel': {
       throwUnlessParallelizable: true,
     },
+    'ember-cli-terser': {
+      exclude: ['assets/dummy.js', 'assets/tests.js', 'assets/test-support.js'],
+    },
```
runspired, line 22: we may want to share a base ember-cli-build, testem, and test-helper.js file from our test-infra/build-infra package to reduce maintenance. Copied this over from ember-data for now.

package.json
```diff
@@ -17,14 +17,14 @@
     "lint:js:changed": "./bin/lint-changed-js",
     "problems": "tsc -p tsconfig.json --noEmit --pretty false",
     "start": "yarn workspace ember-data start",
-    "test": "yarn workspace ember-data test",
+    "test": "lerna run --stream --no-private --ignore='@ember-data/{private-build-infra,canary-features,debug}' test",
```
runspired, line 20: `lerna run` runs the specified command (in this case `test`) in all packages that have the command in scripts in their package.json.

`--stream` tells it to pipe the output to stdout right away vs waiting for the command to finish, this is useful for it to not seem like it's hanging / potentially cause CI to timeout.

`--no-private` is a filter telling it to not run this command for any package marked private in package.json.

`--ignore` tells it to additionally ignore the command in these packages if found. We are ignoring these three because if they do ever have more tests of their own they are likely to be node tests vs tests of addon behavior.

packages/record-data/package.json
```diff
@@ -16,7 +16,8 @@
     "build": "ember build",
     "start": "ember serve",
     "test": "ember test",
-    "test:all": "ember try:each"
+    "test:production": "ember test -e production",
+    "test:try-one": "ember try:one --config-path='../-ember-data/config/ember-try.js'"
```
runspired, line 20: use of `--config-path` here ensures all of the packages we add this command to share a single ember-try config for the various dep scenarios we would like to test.

packages/record-data/package.json
```diff
@@ -48,8 +50,13 @@
     "ember-source-channel-url": "^2.0.1",
     "ember-try": "^1.4.0",
     "loader.js": "^4.7.0",
+    "qunit": "^2.10.0",
+    "qunit-dom": "^1.2.0",
     "silent-error": "^1.1.1"
   },
+  "ember": {
+    "edition": "octane"
+  },
```
runspired, line 59: it's safe to octanify this as we're only adding new tests to this package at the moment and we are only a few weeks away from 4.0. Saves us a little deprecation spew.

packages/serializer/package.json
```diff
@@ -14,10 +14,7 @@
   },
   "scripts": {
     "build": "ember build",
-    "lint:hbs": "ember-template-lint .",
```
runspired, line 17: we can bring the lint command back if it turns out later we want it, but I suspect we need to set up a root config similar to eslint and prefer that.

packages/store/package.json
```diff
@@ -14,19 +14,15 @@
...
+    "start": "ember serve"
   },
   "dependencies": {
     "@ember-data/canary-features": "3.28.0-alpha.0",
     "@ember-data/private-build-infra": "3.28.0-alpha.0",
     "@ember/string": "^1.0.0",
     "ember-cli-babel": "^7.26.3",
     "ember-cli-path-utils": "^1.0.0",
-    "ember-cli-typescript": "^4.1.0",
-    "heimdalljs": "^0.3.0"
```
runspired, line 29: we've been forcing this dep on folks for almost 2 years unnecessarily 😱

packages/unpublished-fastboot-test-app/package.json
```diff
@@ -16,7 +16,7 @@
     "lint:js": "eslint --config ../../.eslintrc.js --ignore-path ../../.eslintignore .",
     "start": "ember serve",
     "test": "ember test --test-port=0",
-    "test:try-one": "ember try:one"
+    "test:scenario": "ember try:one"
```
runspired, line 19: I renamed this just because we have the other command running `try-one` in all of the main packages but this one has it's own config.

igorT:
@rwjblue it would be great if you could give this a quick once over

packages/record-data/package.json
```diff
@@ -40,6 +41,7 @@
     "ember-cli-terser": "~4.0.1",
     "ember-disable-prototype-extensions": "^1.1.3",
     "ember-export-application-global": "^2.0.1",
+    "ember-inflector": "^4.0.1",
```
igorT, line 44: Were we needing but missing the inflector?

packages/record-data/tests/test-helper.js
```diff
@@ -0,0 +1,48 @@
+import { setApplication } from '@ember/test-helpers';
```
igorT, line 1: Can you please add a comment that says this is a copy paste (if it is?). Can we make it a symlink instead? Would that be too trolly

packages/record-data/package.json
runspired: only for testing when we don't bring everything

packages/record-data/tests/test-helper.js
runspired: Had another comment that called this out but yes, testem, test-helper and ember-cli-build file changes are all just copy paste. Since they are all JS files I'm considering re-exporting them from the test-infra package instead so that we can maintain in one place. Was thinking that would be a follow up if it is what we want.

igorT (approved):

packages/unpublished-fastboot-test-app/package.json
igorT: Do we not use this in CI?

runspired: We do but we use it via `yarn test:fastboot` in the root package.json and I also renamed it there.

PR #7458, opened 2021-04-14
chore: simpler serializer for perf test app to reduce noise

runspired:
We don't need to process the payloads since they are already JSON:API.

### DIFF
--- a/packages/unpublished-relationship-performance-test-app/app/serializers/application.js
+++ b/packages/unpublished-relationship-performance-test-app/app/serializers/application.js
@@ -1,7 +1,9 @@
-import DS from 'ember-data';
+export default class Serializer {
+  static create() {
+    return new this();
+  }
 
-export default DS.JSONAPISerializer.extend({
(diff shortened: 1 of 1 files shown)


snewcomer (approved):
I'm not terribly familiar with this part of the codebase but it doesn't seem like much harm.  I think you are also indicating just making a simple class isn't critical (nor important) to the actual results of the test suite?

runspired:
For the findAll the perf test makes ember data will attempt to normalize the payload, but we have no normalizing to do so a simple no-op class is far faster to instantiate and just returning the payload prevents what would otherwise be an expensive traversal and copy.

snewcomer:
Oh right.  Perhaps if I flipped the question - I think you are indicating the actual serializer was not terribly important to the realistic results the perf test output'ed.  That is, a "traversal and copy" vs. not doesn't really affect the results of the perf test?

runspired:
Exactly

PR #7459, opened 2021-04-14
cleanup: record-data should use declare syntax

runspired:
similar to the internal-model cleanup earlier this should marginally improve payload size as well as ensure we don't pay the initialization cost for class fields.


snewcomer (approved):
❤️

PR #7460, opened 2021-04-14
fix: IM recordData setter and don't setId if no recordData instance yet

runspired:
neither of these has a big impact but I noticed them a few times working on other PRs and it helps reduce over-notification of property changes.

### DIFF
--- a/packages/store/addon/-private/system/model/internal-model.ts
+++ b/packages/store/addon/-private/system/model/internal-model.ts
@@ -247,7 +247,7 @@ export default class InternalModel {
   get _recordData(): RecordData {
     if (this.__recordData === null) {
       let recordData = this.store._createRecordData(this.identifier);
-      this._recordData = recordData;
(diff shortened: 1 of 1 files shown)


packages/store/addon/-private/system/model/internal-model.ts
```diff
@@ -247,7 +247,7 @@ export default class InternalModel {
   get _recordData(): RecordData {
     if (this.__recordData === null) {
       let recordData = this.store._createRecordData(this.identifier);
-      this._recordData = recordData;
+      this.__recordData = recordData;
```
snewcomer, line 250: This seems like...it was a bug??

snewcomer (approved):
:nod

packages/store/addon/-private/system/model/internal-model.ts
runspired: Causes an extra set but otherwise did work

PR #7461, opened 2021-04-14
chore: relationships should use declare for fields

runspired:

### DIFF
--- a/packages/record-data/addon/-private/relationships/state/relationship.ts
+++ b/packages/record-data/addon/-private/relationships/state/relationship.ts
@@ -8,6 +8,7 @@ import _normalizeLink from '../../normalize-link';
 import OrderedSet from '../../ordered-set';
 import { implicitRelationshipStateFor, relationshipStateFor } from '../../record-data-for';
 
+type Store = import('@ember-data/store/-private/system/core-store').default;
 type PaginationLinks = import('@ember-data/store/-private/ts-interfaces/ember-data-json-api').PaginationLinks;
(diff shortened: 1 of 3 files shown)


packages/record-data/addon/-private/relationships/state/relationship.ts
```diff
@@ -24,39 +25,39 @@ interface ImplicitRelationshipMeta {
...
-    store: any,
+    store: Store,
     inverseKey: string | null,
     relationshipMeta: ImplicitRelationshipMeta,
     recordData: RelationshipRecordData,
     inverseIsAsync?: boolean
   ) {
     this.inverseIsAsync = inverseIsAsync;
-    this.kind = relationshipMeta.kind;
+    this.kind = relationshipMeta.kind || 'implicit';
```
snewcomer, line 60: Seems reasonable

packages/record-data/addon/-private/relationships/state/relationship.ts
```diff
@@ -74,6 +75,11 @@ export default class Relationship {
     this.meta = null;
     this.__inverseMeta = undefined;
 
+    this.links = undefined;
```
snewcomer, line 78: Do you think we need this one (or the one above it)?  I would assume we get that anyways for free without the explicit declaration?

snewcomer (approved):
:love

packages/record-data/addon/-private/relationships/state/relationship.ts
runspired: Yes. It shapes the object so that it doesn't transition later when we do set it.

PR #7463, opened 2021-04-15
chore: cleanup Implicit relationships

runspired:
improves our typing throughout which allows us to drop some checks and extra initializations, makes where/when we are operating on an implicit relationship more clear (by using a flag instead of juggling keys). This allows us to have a single inverse key vs an inverseKey and inverseKeyForImplicit.

Only behavior change is that implicit keys are improved: before we were setting them to `undefined${this.key}` at all times, which is the reason for the (single) test change.

Motivation is to more closely align the current state with my next PR so that these somewhat more cosmetic changes are more easily distinguished from the logical changes.


packages/record-data/addon/-private/ordered-set.ts
```diff
@@ -28,9 +27,9 @@ function guidFor(obj): string {
   return obj.clientId || obj.lid;
 }
 
-export default class OrderedSet {
-  declare presenceSet: Dict<RelationshipRecordData | null>;
-  declare list: (RelationshipRecordData | null)[];
+export default class OrderedSet<T> {
```
igorT, line 30: ❤️

packages/record-data/addon/-private/relationships/state/relationship.ts
```diff
@@ -215,13 +227,13 @@ export default class Relationship {
     return recordData._relationships !== undefined && recordData._relationships !== null;
   }
 
-  get _inverseMeta(): RelationshipSchema {
+  get _inverseMeta(): RelationshipMeta {
     if (this.__inverseMeta === undefined) {
       let inverseMeta = null;
 
-      if (this.inverseKey) {
+      if (!this.inverseIsImplicit) {
```
igorT, line 234: Doesn't this also pass if there is no inverse at all?

packages/record-data/addon/-private/relationships/state/relationship.ts
```diff
@@ -204,7 +216,7 @@ export default class Relationship {
   }
 
   _inverseIsSync(): boolean {
-    return !!(this.inverseKey && !this.inverseIsAsync);
+    return !this.inverseIsImplicit && !this.inverseIsAsync;
```
igorT, line 219: Same here

packages/record-data/addon/-private/relationships/state/relationship.ts
```diff
@@ -362,14 +375,21 @@ export default class Relationship {
       if (!this._hasSupportForImplicitRelationships(recordData)) {
         return;
       }
-      let relationships = recordData._implicitRelationships;
-      let relationship = relationships[this.inverseKeyForImplicit];
+      const relationships = implicitRelationshipsFor(recordData);
```
igorT, line 378: `let` if we are following our current style

packages/record-data/addon/-private/relationships/state/relationship.ts
```diff
@@ -489,22 +515,18 @@ export default class Relationship {
     @private
    */
   removeCompletelyFromInverse() {
```
igorT, line 517: Were we ever hitting this before?

packages/record-data/addon/-private/relationships/state/relationship.ts
```diff
@@ -516,7 +538,7 @@ export default class Relationship {
         const id = guidFor(inverseRecordData);
 
         if (this._hasSupportForImplicitRelationships(inverseRecordData) && seen[id] === undefined) {
-          const relationship = implicitRelationshipStateFor(inverseRecordData, this.inverseKeyForImplicit);
+          const relationship = implicitRelationshipStateFor(inverseRecordData, this.inverseKey);
```
igorT, line 541: `let`

packages/store/addon/-private/system/store/record-data-store-wrapper.ts
```diff
@@ -13,13 +12,17 @@ type IdentifierCache = import('../../identifiers/cache').IdentifierCache;
...
 type RelationshipsSchema = import('../../ts-interfaces/record-data-schemas').RelationshipsSchema;
-
+type RelationshipDefinition = import('@ember-data/model/-private/system/relationships/relationship-meta').RelationshipDefinition;
 /**
   @module @ember-data/store
 */
 
 type StableIdentifierOrString = StableRecordIdentifier | string;
 
+function metaIsRelationshipDefinition(meta: unknown): meta is RelationshipDefinition {
```
igorT, line 22: meta needs to be an object and not null for this to be safe to run

packages/store/addon/-private/system/store/record-data-store-wrapper.ts
```diff
@@ -107,36 +110,46 @@ export default class RecordDataStoreWrapper implements IRecordDataStoreWrapper {
...
         return definition.inverse;
       } else {
         //TODO add a test for this branch
-        if (!definition._inverseKey) {
-          return null;
+        if (metaIsRelationshipDefinition(definition)) {
+          return definition._inverseKey(this._store, modelClass);
         }
-        return definition._inverseKey(this._store, modelClass);
+        return null;
```
igorT, line 125: Why null here now?

packages/store/addon/-private/system/store/record-data-store-wrapper.ts
```diff
@@ -107,36 +110,46 @@ export default class RecordDataStoreWrapper implements IRecordDataStoreWrapper {
...
+      return false;
+    }
     if (CUSTOM_MODEL_CLASS) {
       if (definition.inverse === null) {
         return false;
       }
-      if (definition.inverseIsAsync !== undefined) {
-        return !!definition.inverseIsAsync;
-      } else {
+      if (((definition as unknown) as { inverseIsAsync?: boolean }).inverseIsAsync !== undefined) {
```
igorT, line 142: Whats up with multiple casts?

packages/store/addon/-private/ts-interfaces/record-data-schemas.ts
```diff
@@ -8,15 +6,16 @@ export interface RelationshipSchema {
   /**
    * @internal
    */
-  [BRAND_SYMBOL]: 'RelationshipSchema';
   kind: 'belongsTo' | 'hasMany';
-  type: string;
-  key: string;
+  type: string; // related type
```
igorT, line 10: 👍

packages/record-data/addon/-private/relationships/state/relationship.ts
runspired: no, it was here just for typescript it seems.

packages/store/addon/-private/system/store/record-data-store-wrapper.ts
runspired: we already check that it's not false-y prior to running this.

packages/store/addon/-private/system/store/record-data-store-wrapper.ts
runspired: it was null before too, we just inverted which thing was checked because it allows us to narrow what `definition` is. Now if there is a method for finding the inverse key we invoke it vs before if there was not a prop (that may or may not have been a method) we would return null.

packages/store/addon/-private/system/store/record-data-store-wrapper.ts
runspired: even though RelationshipDefinition implements RelationshipMeta typescript does not allow us to recast to it without first casting to `unknown` unless we change the return type of `relationshipsDefinitionFor` to be a union type. Since this is the only location that we leak our current implementation detail for meta on @ember-data/model it seemed better to leave the return type there alone.

PR #7465, opened 2021-04-16
chore: small improvements to the asset-size report output

runspired:


bin/asset-size-tracking/generate-diff.js
```diff
@@ -188,6 +200,9 @@ function humanizeNumber(n) {
       s = n.toFixed(0);
     }
   }
+  if (n > 0) {
+    s = '+' + s;
+  }
```
snewcomer, line 205: 👍

snewcomer (approved):
Math checks out!

PR #7466, opened 2021-04-16
chore: fix factory-guy link, activate model-fragments, remove ember-data-change-tracker

runspired:
cc @patocallaghan we were skipping the factory-guy tests since they were failing for so long. Now that it's maintained I've updated the link and if passing will allow it to fail our test suite again.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -313,13 +313,13 @@ jobs:
           ]
         include:
           - partner: ember-data-change-tracker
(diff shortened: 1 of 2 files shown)


patocallaghan:
👍 @runspired. Just a heads up that they still _may_ fail as factory guy has some tests based around model fragments which I'm not sure if they pass on their own. I should be working on upgrading things to make them more compatible soon.

snewcomer (approved):

PR #7467, opened 2021-04-16
Cleanup typing in record-data-store-wrapper

igorT:


runspired:
Remaining failure is because master also needs this fix. Merging.

PR #7471, opened 2021-04-19
chore: Improved test coverage for internal relationship state

runspired:
While refactoring the relationship layer in #7470 I noticed that some of the changes I was making did not fail very many tests despite altering semantics of the relationship layer we ought to have been aware of changes happening to.

This adds test coverage to the relationship cache to a wide range of scenarios. I've written the tests in a way where the underlying implementation can be easily swapped out with only minor alterations to how the tests in the test suite gain access to the cache data. This will let us more easily spot semantics changes and evaluate if that change is a regression or improvement.

The test matrix for the graph covers 96 unique situations around what happens when a record is either `unloaded/deleted/deleted+persisted/deleted+persisted+unloaded` from a `clean/dirty/new` state for `1:1 / many:many / 1:1(implicit) / many:many(implicit)` for both the `sync` and `async` cases.

The semantics it covers are at times implementation dependent, but the idea is that we should know when we've changed them.

Asset size increase is because I needed to port the coming additions to `recordDataFor` in order to enable peeking a recordData. In the process this surfaced a bug with our current `recordData` destroy calls that may require amending the v2 RecordData RFC as I do not believe it currently has a sufficient destroy lifecycle hook. More investigation needed.

**Fixed** ~~The failure for the `model-fragments` external partner test is because master is currently failing for that test. A backwards compatible fix has been submitted to that library, see note here: https://github.com/adopted-ember-addons/ember-data-model-fragments/issues/396~~

### DIFF
--- a/packages/record-data/addon/-private/ts-interfaces/relationship-record-data.ts
+++ b/packages/record-data/addon/-private/ts-interfaces/relationship-record-data.ts
@@ -1,3 +1,4 @@
+type StableRecordIdentifier = import('@ember-data/store/-private/ts-interfaces/identifier').StableRecordIdentifier;
 type SingleResourceRelationship = import('@ember-data/store/-private/ts-interfaces/ember-data-json-api').SingleResourceRelationship;
 type CollectionResourceRelationship = import('@ember-data/store/-private/ts-interfaces/ember-data-json-api').CollectionResourceRelationship;
 type RecordData = import('@ember-data/store/-private/ts-interfaces/record-data').RecordData;
@@ -21,6 +22,7 @@ export interface RelationshipRecordData extends RecordData {
   isNew(): boolean;
   modelName: string;
   storeWrapper: RecordDataStoreWrapper;
+  identifier: StableRecordIdentifier;
   id: string | null;
   clientId: string | null;
   isEmpty(): boolean;
--- a/packages/record-data/tests/integration/graph/edge-removal/abstract-edge-removal-test.ts
+++ b/packages/record-data/tests/integration/graph/edge-removal/abstract-edge-removal-test.ts
@@ -1,0 +1,251 @@
+import { assign } from '@ember/polyfills';
+import settled from '@ember/test-helpers/settled';
+
+import { module, test } from 'qunit';
+
+import { setInitialState, testFinalState } from './helpers';
+import { setupGraphTest } from './setup';
+
+type TestConfig = import('./helpers').TestConfig;
+type Context = import('./setup').Context;
+
+module('Integration | Graph | Nodes', function(hooks) {
+  setupGraphTest(hooks);
+
+  /**
+   * These are the various configurations we run tests for on the graph
+   * to ensure things are working.
+   *
+   * We don't currently test 1:many or many:1 relationships. It's unclear
+   * if the semantics of these are different enough to require additional
+   * scenarios.
+   *
+   * name: the name of the test
+   * async: whether the relationship should be async or sync (both sides will conform to this)
+   * relType: whether the relationship should be belongsTo (1:1) or hasMany (many:many)
+   * inverseNull: whether the relationships should specify inverse: null instead of an explicit inverse.
+   */
+  const TestScenarios: TestConfig[] = [
+    {
+      name: 'sync belongsTo',
+      async: false,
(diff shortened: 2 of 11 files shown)


packages/record-data/tests/integration/graph/edge-removal/abstract-edge-removal-test.ts
```diff
@@ -0,0 +1,221 @@
...
+         * For unload, we treat it as a deletion for sync relationships and
+         * as no-change for async relationships.
+         *
+         * Ergo for async we expect no entries removed (`removed: false`) and for no caches
+         * to have been deleted (`cleared: false`)
+         *
+         * And for sync we expect entries removed (`removed: true`) but for no caches to have been
+         * deleted. (`cleared: false`)
+         */
+        // await testFinalState(this, testState, config, { removed: !config.async, cleared: false }, assert);
```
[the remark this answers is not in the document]
runspired: fixed

packages/record-data/tests/integration/graph/edge-removal/abstract-edge-removal-test.ts
```diff
@@ -0,0 +1,221 @@
...
+
+type TestConfig = import('./helpers').TestConfig;
+type Context = import('./setup').Context;
+
+module('Integration | Graph | Nodes', function(hooks) {
+  setupGraphTest(hooks);
+
+  /**
+   * These are the various configurations we run tests on the graph
+   * for to ensure things are working.
```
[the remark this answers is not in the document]
runspired: fixed

packages/store/addon/-private/system/record-data-for.ts
```diff
@@ -19,11 +22,26 @@ interface InternalModel {
...
+export function removeRecordDataFor(identifier) {
+  IdentifierCache.delete(identifier);
+}
 
+export default function recordDataFor(instance: StableRecordIdentifier): RecordData | null;
 export default function recordDataFor(instance: Instance): RecordData;
 export default function recordDataFor(instance: object): null;
 export default function recordDataFor(instance: Instance | object): RecordData | null {
+  if (IdentifierCache.has(instance as StableRecordIdentifier)) {
+    return IdentifierCache.get(instance as StableRecordIdentifier) as RecordData;
```
snewcomer, line 43: I'm surprised `as RecordData` cast is necessary.  I would have thought TS would have seen through to the return type.

packages/record-data/addon/-private/ts-interfaces/relationship-record-data.ts
```diff
@@ -21,6 +22,7 @@ export interface RelationshipRecordData extends RecordData {
   isNew(): boolean;
   modelName: string;
   storeWrapper: RecordDataStoreWrapper;
+  identifier: StableRecordIdentifier;
```
snewcomer, line 25: Would you consider this PR more than test coverage?  I see we are adding a new `recordDataFor` function declaration.  However, I'm unclear as to its usage (currently, unless you have more planned).  Is the "future" plans for building up a weak map captured in the TODO above?

snewcomer (commented):
💯

packages/store/addon/-private/system/record-data-for.ts
runspired: `has` doesn't narrow is why, so TS sees it as `| undefined`. It will narrow using the `get` first and then checking for `undefined` instead of using `has`, which is the pattern we use when we may also need to assign.

packages/record-data/addon/-private/ts-interfaces/relationship-record-data.ts
runspired: It's currently used by the tests and will be used extensively by the new relationship layer. The TODO is for our InternalModel burndown: this iteration logic needs to move into a combination of the graph / store as currently the only thing causing RecordData's to not be retained is that they are only entangled with InternalModel and so when we release an InternalModel we by result release the RecordData.

runspired: (The tests need this because they need an abstract way to peek for whether there is a RecordData for an identifier without accidentally rematerializing it or cancelling destruction)

PR #7472, opened 2021-04-19
Revert "[CHORE] auto publish alpha versions"

runspired:
Reverts #7436 cc @spham92 

See this failure: https://github.com/emberjs/data/commit/2feb0c1a9f2e54fbbcdeaba0d62468a4a5dadebb

I manually ran `npm view ember-data versions --json` and see the following

<details>
<summary>Output</summary>
```json
[
  "1.0.0-beta.10",
  "1.0.0-beta.11",
  "1.0.0-beta.12",
  "1.0.0-beta.14",
  "1.0.0-beta.14.1",
  "1.0.0-beta.15",
  "1.0.0-beta.16",
  "1.0.0-beta.16.1",
  "1.0.0-beta.17",
  "1.0.0-beta.18",
  "1.0.0-beta.19",
  "1.0.0-beta.19.1",
  "1.0.0-beta.19.2",
  "1.13.0",
  "1.13.1",
  "1.13.2",
  "1.13.3",
  "1.13.4",
  "1.13.5",
  "1.13.6",
  "1.13.7",
  "1.13.8",
  "1.13.9",
  "1.13.10",
  "1.13.11",
  "1.13.12",
  "1.13.13",
  "1.13.14",
  "1.13.15",
  "1.13.16",
  "2.0.0-beta.1",
  "2.0.0-beta.2",
  "2.0.0",
  "2.0.1",
  "2.1.0-beta.1",
  "2.1.0-beta.2",
  "2.1.0-beta.3",
  "2.1.0-beta.4",
  "2.1.0-beta.5",
  "2.1.0",
  "2.2.0-beta.1",
  "2.2.0-beta.2",
  "2.2.0-beta.3",
  "2.2.0-beta.4",
  "2.2.0",
  "2.2.1",
  "2.3.0-beta.1",
  "2.3.0-beta.2",
  "2.3.0-beta.3",
  "2.3.0-beta.4",
  "2.3.0-beta.5",
  "2.3.0",
  "2.3.1",
  "2.3.2",
  "2.3.3",
  "2.4.0-beta.1",
  "2.4.0-beta.2",
  "2.4.0-beta.3",
  "2.4.0",
  "2.4.1",
  "2.4.2",
  "2.4.3",
  "2.5.0-beta.1",
  "2.5.0-beta.2",
  "2.5.0-beta.3",
  "2.5.0-beta.4",
  "2.5.0",
  "2.5.1",
  "2.5.2",
  "2.5.3",
  "2.5.4",
  "2.5.5",
[...]

PR #7473, opened 2021-04-20
[CHORE] auto publish attempt 2

spham92:
# Summary
Adding code back revert in #7472 with some differences

- Cron is now scheduled on Wednesday to match emberjs's auto alpha release time
- For finding the next alpha version, I changed the code to just increment the pre-release version in `lerna.json`. This means that we will need a human to bump the minor and major when needed.
- Added `--dryRun` as an option to publish script to skip auto committing and publish when using `lerna` and skipping publishing to npm. This will be useful during local testing

Repo owners need to add `NODE_AUTH_TOKEN` to the secrets for this to fully work.

Use the following command to test what would be the bulk of the GitHub action: `node bin/publish.js canary --dryRun --autoAlphaVersion --force --skipSmokeTest`

# Testing
I tried the GitHub action on a local version of the Ember Data: https://github.com/spham92/data/actions/runs/769888390
It failed because of missing NODE_AUTH_TOKEN which is great since it now throws an error for that

For another instance, I ran a dry run and tested the artifacts: https://github.com/spham92/data/actions/runs/770096237


runspired:
@spham92 I think I see the assumption made in the first version that was incorrect and I think we could make this a bit safer still.

The semantics of our release cycle is that there is always a stable|beta|canary with different version numbers. E.g. if latest stable is `3.27.x` then the current beta will always be `3.28.0-beta.N` and the current alpha will always be `3.29.0-alpha.N`.

I think since there is potential for human error or for version fixes to past versions to be pushed to github that could appear in front of latest stable (assuming github is sorting chronologically vs by incrementing number) I would prefer that we find the version a different way.

Also considering that for this approach the next canary to be a major/minor requires a manual bump to the version anyway, what prevents us from using the version in package.json?

.github/workflows/alpha-release.yml
```diff
@@ -0,0 +1,55 @@
...
+      - uses: actions/checkout@v2
+      - uses: actions/setup-node@v2
+        with:
+          node-version: 12.x
+          repo-token: ${{ secrets.GITHUB_TOKEN }}
+      - name: Install dependencies for master
+        run: yarn install --frozen-lockfile --non-interactive
+      - name: Make sure git user is setup
+        run: |
+          git config --local user.email 'cron@example.com'
```
runspired, line 44: reminder to update this. I believe there's a tomster one to use (see ember's config)

bin/publish.js
```diff
@@ -347,28 +370,47 @@ async function main() {
...
+    }
+    // Assume human ran script if token is missing
+    await confirmPublish(tarballs, options, !npmAuthTokenInEnv);
     console.log(`✅ ` + chalk.cyan(`Successfully Published ${nextVersion}`));
   } else {
     console.log('⚠️ ' + chalk.grey(`Skipping Publishing`));
   }
 }
 
+let cli = readline.createInterface({
```
runspired, line 409: is this left over from something? I don't see it used in this PR

runspired (approved):
Overall 👍🏻 . Wouldn't mind @rwjblue giving it a pass. A couple of minor cleanups to do before we merge.

Also @rwjblue or @igorT will need to setup the secret as I cannot.

bin/publish.js
spham92: good catch, reverted

.github/workflows/alpha-release.yml
spham92: updated to:

```
      - name: Make sure git user is setup
        run: |
          git config --local user.email 'tomster@emberjs.com'
          git config --local user.name 'Ember.js Alpha Releaser'
```

runspired (approved):

PR #7475, opened 2021-04-21
[ON_HOLD] feat: add additional state-machine states to handle unload/loading edge cases

runspired:
resolves https://github.com/emberjs/rfcs/issues/358
resolves https://github.com/emberjs/rfcs/issues/359

While working on #7470 and #7471 I noticed that we don't properly inform RecordData today whether the teardown of relationships is occurring for a new record or not. This is primarily because the context is lost. It's also currently impossible to tell internally if an empty record is empty because it was unloaded after a persisted deletion, has never been fetched, or has been unloaded but retained. This resolves this which will allow us to do nicer things internally in many locations.

We may desire to RFC these, though in most ways this is additive backfill to fix existing bugs with the state machine. It's unclear whether new substates which result in new `currentState.stateName` strings require RFC or not.

I scanned ember-observer and no actively maintained addons will be affected (two that were never published and haven't been touched in years would have had a marginal migration to make).

### DIFF
--- a/packages/-ember-data/tests/integration/records/delete-record-test.js
+++ b/packages/-ember-data/tests/integration/records/delete-record-test.js
@@ -2,6 +2,7 @@
 
 import { get } from '@ember/object';
 import { run } from '@ember/runloop';
+import settled from '@ember/test-helpers/settled';
 
 import { module, test } from 'qunit';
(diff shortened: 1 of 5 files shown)


runspired:
Closing in favor of the state machine refactor in #7258

PR #7477, opened 2021-04-22
[ON_HOLD] feat: custom decorators

runspired:
This allows Ember to more easily remove support for some aspects of ComputedProperty, computed.meta etc. while also giving us better access to the schema at runtime. This is the first step towards being able to compile schema at build time, some deprecations are necessary to take that step.

In theory, this should provide a significant performance boost for accessing static schema and/or accessing a value. However, do to the complexity of also supporting sync-observers and async-proxies there are some caveats that may negate these wins. Benchmarking is needed.

**Edit** Going to put this on hold for now because (1) the various workarounds for `reopen` and `sync observers` seem to nuke perf in addition to adding a lot more size and complexity and (2) there's no clear path forward for this issue: https://github.com/emberjs/ember.js/pull/19504


runspired:
this went on hold because it turned out we could not support classic syntax with custom decorators even with extensive private API abuse.

Since we are phasing out model, this change is now low priority. If we decide to keep a classic-class version of Model in the future we should deprecate classic syntax and then change the implementation.

PR #7480, opened 2021-04-27
CHORE: configure alpha-releases to use deploymeent environment

spham92:
Adding `environment` entry with `deployment` value for the release job

### DIFF
--- a/.github/workflows/alpha-release.yml
+++ b/.github/workflows/alpha-release.yml
@@ -31,6 +31,7 @@ jobs:
     name: Run publish script
     runs-on: ubuntu-latest
     needs: [ test ]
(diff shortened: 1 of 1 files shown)


.github/workflows/alpha-release.yml
```diff
@@ -31,6 +31,7 @@ jobs:
     name: Run publish script
     runs-on: ubuntu-latest
     needs: [ test ]
+    environment: deloyment
```
runspired, line 34: what's the specific need for this?

igorT: I added the secret only for the environment called `deployment`

PR #7493, opened 2021-04-30
Relationship Refactor (part 2): The graph should coordinate state updates

runspired:
Basic premise here is that the individual edges can't optimize updating their inverses, state, or notifying the UI efficiently while the graph, acting as a coordinator, can.

The existing explosion of "remove" and "add" spaghetti (roughly 12 different methods for removing a record from one side last I counted) was primarily due to attempting to prevent "cycling" (an update to one side updating it's inverse which then sends the same updates back to the originating side, potentially bouncing multiple times). These efforts were mostly in vain and we often cycle multiple times even in a first load scenario.

When this PR completes, we should be left in a state of updating the various edges while only visiting each once per relationship payload received (note: a json-api payload will typically contain multiple relationship payloads for the same relationship, so this does not yet do any work to avoid that double pass).


packages/store/addon/-private/system/backburner.js
```diff
@@ -8,11 +8,13 @@ import Ember from 'ember';
 
 // TODO: expose Ember._Backburner as `import { _Backburner } from '@ember/runloop'` in ember-rfc176-data + emberjs/ember.js
 /*
-  flushRelationships is used by the Graph to batch updates
   syncRelationships is used by the UI to grab updates from the graph
   and update the ManyArrays.
+
+  We may be able to remove this once the new relationship layer is
+  complete.
```
snewcomer, line 15: Related to the other question, what will allow it's removal?

snewcomer (approved):
This is really nice and easy to follow.  When I get to work in this area, I am sure it will all be easy to navigate and understand.  Not able to comment on edge cases or potential regressions other than relying on the test suite.  So you get an A+ from me!

packages/model/addon/-private/model.js
```diff
@@ -949,7 +949,9 @@ class Model extends EmberObject {
     @method rollbackAttributes
   */
   rollbackAttributes() {
-    this._internalModel.rollbackAttributes();
+    this.store._backburner.join(() => {
+      this._internalModel.rollbackAttributes();
+    });
```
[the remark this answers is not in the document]
runspired: I noticed we were using a join around the `notify` and `push` schedules, which meant these things were being called without an existing join. I worked through all the possible entry points to that flush and found which ones didn't have a join wrapper already and added them.

Moving the join up means that a lot of scenarios that were creating a flush-per-schedule now create only one flush, which helps us avoid a lot more work. This had the most noticeable effect on teardown/unload.

packages/store/addon/-private/system/backburner.js
runspired: Our scheduler right now does three things (which is why there are three queues)

- it batches payloads that have been pushed for a single flush, thereby allowing us to order the pushes by type (more efficient to push hasMany first. Removing this batching regressed the update scenario by 500%). It's unclear how we'd do this without some form of run loop concept, however we could probably make a simpler queue thing and drop our need for backburner if this was all that was left.
- it batches (and minimizes) updates to hasMany local state in a second pass, this allows us to update the local state with those changes only once instead of potentially many times. This optimization would go away if we were to compute localState lazily on access.
- it notifies the UI layer only once all state has settled (and minimizes those notifications to only one per relationship). We can possibly do away with this once we no longer support sync-observers if state is also lazily calculated.

packages/-ember-data/tests/integration/relationships/inverse-relationships-test.js
```diff
@@ -223,15 +223,15 @@ module('integration/relationships/inverse_relationships - Inverse Relationships'
 
     comment.set('post', post);
 
-    assert.equal(comment.get('post'), post);
-    assert.equal(post.get('bestComment'), comment);
-    assert.strictEqual(post2.get('bestComment'), null);
+    assert.ok(comment.get('post') === post, 'comment post is set correctly');
```
igorT, line 226: Some cleanup here

packages/-ember-data/tests/integration/store-test.js
```diff
@@ -1046,11 +1046,7 @@ module('integration/store - deleteRecord', function (hooks) {
 
     assert.ok(store.hasRecordForId('person', '1'), 'expected the record to be in the store');
```
igorT, line 1047: Is this an intentional removal

PR #7494, opened 2021-04-30
chore: improved monitoring

runspired:
- Fixes issue causing the asset-size comment to not get posted recently.
- Brings back and improves performance monitoring comment.
- Restores all the other performance scenarios
- Cuts fixture generation time out of the part of the report we care about for materialization
- Adds artifact uploading for the performance report

Requires #7499


snewcomer (approved):
:perf

PR #7500, opened 2021-04-30
chore: static fixtures for perf testing & new scenario

runspired:
This should reduce some of the noise we see in the tests by eliminating the most expensive bit of the runtime setup cost with variable perf characteristics.

Update the fixtures by altering `fixtures/index.js` and then running it with `node ./fixtures/index.js` from within the perf app.
We check-in the generated fixtures to `public`.

Perf check fails because markers were changed. Verified locally these routes render as expected.

This also adds explicit testing for the cost of unused relationships (the feature that lazy-relationships initially was built to cover) as the **unused-relationships** perf test.


snewcomer (approved):
🚤

PR #7505, opened 2021-05-04
Perf: Refactor PromiseManyArray and prep for RFC#745

runspired:
This allows us to remove the sync-flush of ManyArray by utilizing a custom enumerable instead of ArrayProxy which was installing array observers. We now only sync-flush if array observers are added by an application.

Refactored in a way that adding deprecations from emberjs/rfcs#745 should be trivial, though will require test changes to resolve any deprecations it results in for our own app (I have a branch with those changes for when the time comes)


packages/store/addon/-private/system/store/finders.js
```diff
@@ -262,9 +262,19 @@ function fixRelationshipData(relationshipData, relationshipKind, { id, modelName
...
     if (relationshipData) {
-      if (
-        !relationshipData.filter((v) => v.type === parentRelationshipData.type && v.id === parentRelationshipData.id)[0]
-      ) {
+      // IE11 does not support array.find
+      // these arrays could be massive so this is better than filter
+      // Note: this is potentially problematic if type/id are not in the
+      // same state of normalization.
+      let found = false;
+      for (let i = 0; i < relationshipData.length; i++) {
```
ksten-ab, line 270: Not sure what your definition of massive is, but once you get in to the 10k+ range this optimization can become worthwhile
```suggestion
      for (let i = 0, len = relationshipData.length; i < len; i++) {
```

packages/model/addon/-private/system/promise-array-proxy.ts
```diff
@@ -0,0 +1,375 @@
...
+  ```js
+  const comments = await post.comments;
+  ```
+
+
+
+  @class PromiseManyArray
+  @private
+*/
+class PromiseArrayProxy {
```
ksten-ab, line 41: doc mismatch

runspired: docs are right, class name is wrong. Used a new class name to avoid the old one while I had both for testing. Will update.

packages/store/addon/-private/system/store/finders.js
runspired: I'm under the impression though @krisselden can correct me that most browsers now optimize length access in this way anyway.

packages/-ember-data/tests/integration/adapter/store-adapter-test.js
```diff
@@ -1169,7 +1169,7 @@ module('integration/adapter/store-adapter - DS.Store and DS.Adapter integration
               },
             },
           });
-          assert.ok(tom.get('dogs') instanceof PromiseArray, 'dogs is a promise');
+          assert.ok(typeof tom.dogs.then === 'function', 'dogs is a thenable');
```
igorT, line 1172: Completely fine if not, but any way to also assert that it's an array?

packages/-ember-data/tests/integration/relationships/nested-relationship-test.js
```diff
@@ -38,112 +37,107 @@ module('integration/relationships/nested_relationships_test - Nested relationshi
...
     let adapter = store.adapterFor('application');
 
     adapter.shouldBackgroundReloadRecord = () => {
       return false;
     };
 
-    run(() => {
-      store.push({
-        data: {
+    store.push({
```
igorT, line 48: I assume the payload is the same here, and github's diff just got confused?

packages/model/addon/-private/model.js
```diff
@@ -1277,13 +1277,6 @@ class Model extends EmberObject {
     return this.constructor.inverseFor(key, this._internalModel.store);
   }
 
-  notifyHasManyAdded(key) {
```
igorT, line 1280: This was a leftover from the previous PR?

packages/store/addon/-private/system/core-store.ts
```diff
@@ -3121,9 +3135,6 @@ abstract class CoreStore extends Service {
     assert('relationshipReferenceFor is only available when CUSTOM_MODEL_CLASS ff is on', false);
   }
 
-  /**
```
igorT, line 3124: Why removing the internal?

igorT:
The list of items in the PR title is a bit confusing, as `eliminate` applies only to the first two elements in the list. Something like: "Eliminate _retainedManyArray cache and sync relationship flushing. Also streamline notifications" Would be more readable.

packages/-ember-data/node-tests/fixtures/expected.js
```diff
@@ -384,5 +382,14 @@ module.exports = {
...
     '(public) Ember.Inflector#uncountable',
+    '(private) @ember-data/model PromiseManyArray#forEach',
+    '(public) @ember-data/model PromiseManyArray#isFulfilled',
+    '(public) @ember-data/model PromiseManyArray#isPending',
+    '(public) @ember-data/model PromiseManyArray#isRejected',
+    '(public) @ember-data/model PromiseManyArray#isSettled',
+    '(public) @ember-data/model PromiseManyArray#length',
+    '(public) @ember-data/model PromiseManyArray#links',
+    '(public) @ember-data/store ManyArray#links',
+    '(public) @ember-data/store Store#identifierCache'
```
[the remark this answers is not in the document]
runspired: Yes. The identifierCache as a public prop on store was what we RFCd.

packages/-ember-data/tests/integration/relationships/nested-relationship-test.js
runspired: Yes

packages/-ember-data/tests/integration/adapter/store-adapter-test.js
runspired: not really, since it wont one. We can assert it's an instance of our class still but the main thing is that its a thenable so I left it at that.

packages/model/addon/-private/system/promise-many-array.ts
```diff
@@ -0,0 +1,322 @@
...
+ @module @ember-data/model
+ */
+/**
+  This class is returned as the result of accessing an async hasMany relationship
+  on an instance of a Model extending from `@ember-data/model`.
+
+  A PromiseManyArray is an array-like proxy that also proxies certain method calls
+  to the underlying ManyArray in addition to being "promisified".
+   
+  Right now we proxy:
```
igorT, line 20: We now proxy more methods right?

runspired: This comment was pulled forward from the existing implementation, and describes only the methods on ManyArray that aren't typical to MutableArray implementations. We of course also proxy all MutableArray methods just by the nature of being an ArrayProxy.

PR #7509, opened 2021-05-07
[BUGFIX] filter @internal, add more robust docs tests, add public docs where missing

runspired:
Resolves https://github.com/ember-learn/ember-api-docs/issues/752

Adds a preprocessor that removes anything with excludedTags, which have been configured to exclude `@internal` and `@feature`.

This latter means that we can document feature flagged code without publishing the docs.

Audited and fixed small mistakes throughout, added/stubbed some missing documentation for public things.

Was able to build and run the api-docs site locally and confirmed things are where they should be.


acorncom (approved):
LGTM 👍🏻

packages/debug/addon/index.js
```diff
@@ -72,7 +71,7 @@ export default DataAdapter.extend({
     Maintains the list of model types without needing the Model package for detection.
 
     @method watchModelTypes
-    @public
+    @private
```
snewcomer, line 74: > Although YUIDoc does not generate documentation for @private blocks

These markers are for internal structure of our source code.

packages/store/addon/-private/system/snapshot.ts
```diff
@@ -54,11 +56,20 @@ export default class Snapshot implements Snapshot {
   public include?: unknown;
   public adapterOptions: Dict<unknown>;
 
+  /**
+   * @method constructor
+   * @constructor
+   * @private
```
snewcomer, line 62: Do you know if this is necessary with the `@constructor` tag?

packages/model/addon/-private/errors.js
```diff
@@ -391,7 +402,8 @@ export default ArrayProxy.extend(DeprecatedEvented, {
    errors.get('messages')
    // => []
    ```
-   @method remove
+   @method clear
+    @public
```
snewcomer, line 406: can't tell if this is a UI issue or there is an actual extra space here

packages/adapter/addon/-private/build-url-mixin.ts
```diff
@@ -382,6 +393,7 @@ export default Mixin.create({
    ```
 
    @method urlForDeleteRecord
+    @public
```
snewcomer, line 396: same here.  Not sure if it is a big deal

snewcomer (approved):
This all looks great!  Didn't see any `@protected` tags.  Just pointing it out in case something came to mind for it's use.

packages/model/addon/-private/errors.js
runspired: just a copy pasta issue

PR #7510, opened 2021-05-07
feat: activate all ffs related to CUSTOM_MODEL_CLASS

runspired:
Activates all Features for 3.28


runspired:
Ilios has 3 test failures, seem to all be in the same unit test module. Will investigate Friday afternoon.

runspired:
The perf jitter here is minimal enough outside of the destroy case that I'm not concerned. The destroy case should be *at least* offset if not fully addressed and then improved by #7505 

The file size increase is slightly concerning: it seems likely we did not flag for removal enough of the old code paths. Hopefully once we begin cleanup where we missed things will become evident.

runspired:
Not sure why the prod IE11 scenario failed here, can't spot anything that specifically seems IE11 problematic, don't have easy access to a VM with IE11 to debug either.

packages/store/addon/-private/system/record-notification-manager.ts
```diff
@@ -9,18 +9,27 @@ type UnsubscribeToken = Object;
...
-    notificationType: 'attributes' | 'relationships' | 'identity' | 'errors' | 'meta' | 'unload' | 'property' | 'state'
-  ): void;
+export type NotificationType =
+  | 'attributes'
+  | 'relationships'
+  | 'identity'
+  | 'errors'
+  | 'meta'
+  | 'unload'
+  | 'property'
```
igorT, line 19: Can we add a comment that this is only ED internal. We should also have a path away from it

PR #7511, opened 2021-05-07
chore: deactivate storefront partner test until polymorphic issue is resolved

runspired:


igorT:
Can you please link the polymorphic issue in this PR so we can track it

PR #7512, opened 2021-05-07
Add has many tests

amk221:
Adds tests that were originally created in this PR: https://github.com/emberjs/data/pull/7287

Splitting off from 7287, because the original problem is [already fixed](https://github.com/emberjs/data/pull/7287#issuecomment-834295693).

### DIFF
--- a/packages/-ember-data/tests/integration/adapter/rest-adapter-test.js
+++ b/packages/-ember-data/tests/integration/adapter/rest-adapter-test.js
@@ -998,6 +998,82 @@ module('integration/adapter/rest_adapter - REST Adapter', function (hooks) {
       });
   });
 
+  test('updateRecord - hasMany relationships faithfully reflect removal from response', async function (assert) {
+    Post.reopen({ comments: DS.hasMany('comment', { async: false }) });
+    Comment.reopen({ post: DS.belongsTo('post', { async: false }) });
+
+    store.push({
+      data: {
+        type: 'post',
+        id: '1',
+        attributes: {
+          name: 'Not everyone uses Rails',
+        },
+        relationships: {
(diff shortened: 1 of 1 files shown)


packages/-ember-data/tests/integration/adapter/rest-adapter-test.js
```diff
@@ -998,6 +998,82 @@ module('integration/adapter/rest_adapter - REST Adapter', function (hooks) {
       });
   });
 
+  test('updateRecord - hasMany relationships faithfully reflect removal from response', async function (assert) {
+    Post.reopen({ comments: DS.hasMany('comment', { async: false }) });
+    Comment.reopen({ post: DS.belongsTo('post', { async: false }) });
```
runspired, line 1003: we should register a fresh class here vs using reopen (we've RFC'd deprecating reopen and need to cleanup our own tests). Same for the test below.

runspired (changes requested):
Overall looks good and I'm glad to see the simpler relationships have resolved this issue already! Small change to make.

packages/-ember-data/tests/integration/adapter/rest-adapter-test.js
amk221: ok, like [that](https://github.com/emberjs/data/pull/7512/commits/9080033d14b6166f19d17fdc3b3ab147dce9842e)?

runspired (approved):

packages/-ember-data/tests/integration/adapter/rest-adapter-test.js
runspired: exactly, thanks!

PR #7513, opened 2021-05-07
chore: fix automation script - remove debugging info

spham92:
Removing the debugging commands from the github action and publish script.

The solution is ultimately setting the `registry-url` in the github action and ensuring the npmjs package setting allows publishing from an automation token i.e.

image

I missed that the recent failure for the publish script was a 403 vs before which was a 404. I was able to recreate the issues with a dummy repo. To summarize these are conditions that causes the following issue:

- Valid token + no set registry-url + whatever npmjs security setting = 404
- Valid token + set-registry - 2fa only npmjs security setting = 403

@igorT we need all the ember-data related repos to allow publishing from automation token

### DIFF
--- a/.github/workflows/alpha-release.yml
+++ b/.github/workflows/alpha-release.yml
@@ -43,8 +43,6 @@ jobs:
         run: |
           git config --local user.email 'tomster@emberjs.com'
           git config --local user.name 'Ember.js Alpha Releaser'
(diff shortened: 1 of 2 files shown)


runspired:
I've updated the settings for all of the packages on npmjs.org, thanks @spham92 !

runspired (approved):

PR #7516, opened 2021-05-08
Perf: eliminate many array retention

runspired:
Streamlines our notifications and eliminates most instances of sync-flush of relationship state. The PromiseManyArray (or some consuming apps) will still add array observers to the many array forcing us to do a more expensive flush that is also synchronous (and therefor happening more often).


packages/model/addon/-private/system/many-array.js
```diff
@@ -72,18 +72,8 @@ export default EmberObject.extend(MutableArray, DeprecatedEvented, {
...
     this._length = 0;
 
-    /**
-    Used for async `hasMany` arrays
-    to keep track of when they will resolve.
-
-    @property {Ember.RSVP.Promise} promise
-    @private
-    */
-    this.promise = null;
```
snewcomer, line 85: Was this effectively unused?

packages/model/addon/-private/system/many-array.js
```diff
@@ -192,35 +207,14 @@ export default EmberObject.extend(MutableArray, DeprecatedEvented, {
...
-      this.set('length', toSet.length);
-      this.currentState = toSet.slice();
-      this.arrayContentDidChange(diff.firstChangeIndex, diff.removedCount, diff.addedCount);
-      if (isInitialized && diff.addedCount > 0) {
-        //notify only on additions
-        //TODO only notify if unloaded
-        this.internalModel.manyArrayRecordAdded(this.get('key'));
-      }
-    }
-  },
```
snewcomer, line 215: This was the sync flush, right?

packages/model/addon/-private/system/many-array.js
```diff
@@ -140,39 +136,33 @@ export default EmberObject.extend(MutableArray, DeprecatedEvented, {
...
-        return true;
-      }
+  notify() {
+    this._isDirty = true;
+    if (this._hasArrayObservers) {
+      this.retrieveLatest();
+    } else {
+      this.notifyPropertyChange('[]');
+      this.notifyPropertyChange('firstObject');
+      this.notifyPropertyChange('lastObject');
```
snewcomer, line 158: Would you mind explaining these two blocks? What are we trying to accomplish with the non array observers block?  What will follow in this case due to the `notifyPropertyChange` on `[]`, `firstObject` and `lastObject`?

packages/model/addon/-private/system/many-array.js
runspired: Yeah. It hasn't been used in many years, unclear if it was ever used at all.

packages/model/addon/-private/system/many-array.js
runspired: retrieveLatest is the syncFlush. If we have observers we must flush immediately. If we don't have any we can notify and wait for any subscribers to pull on the ManyArray content to flush.

In the observers case observers are notified by arrayContentWillChange and arrayContentDidChange

packages/model/addon/-private/system/many-array.js
runspired: This was part of it (called by retrieveLatest). The callback here was a weird backflow we were dependent on that I was able to refactor away. Removing this back flow is where most of this perf improvement is coming from in this PR since we still aren't lazy given pretty much everything is wrapped in an ArrayProxy still.

What the backflow was doing was ensuring unloaded internal models were cleared from relationships, fixing the relationship notification flow ensured this would be true without this.

packages/store/addon/-private/system/model/notify-changes.ts
```diff
@@ -24,14 +24,14 @@ export default function notifyChanges(
...
-        }
-        let manyArray = internalModel._manyArrayCache[key] || internalModel._retainedManyArrayCache[key];
-        if (manyArray && !didRemoveUnloadedModel) {
-          manyArray.retrieveLatest();
+        let manyArray = internalModel._manyArrayCache[key];
+
+        if (manyArray) {
+          manyArray.notify();
+
+          if (!meta.options || meta.options.async || meta.options.async === undefined) {
```
igorT, line 32: Can you move the comment from model notify property change to here, so that we remember why we do this

igorT:
I think the PR title might be better understood as `eliminate retainedManyArrayCache` as we still retain many arrays in other cases

packages/-ember-data/tests/unit/many-array-test.js
```diff
@@ -105,6 +105,7 @@ module('unit/many_array - DS.ManyArray', function (hooks) {
     let willChangeRemoveAmt;
     let willChangeAddAmt;
 
+    DS.ManyArray.proto().__hasArrayObservers = true;
```
igorT, line 108: Would be nice to have a comment here so we don't have to search later why we are doing this

packages/model/addon/-private/system/many-array.js
```diff
@@ -140,39 +136,33 @@ export default EmberObject.extend(MutableArray, DeprecatedEvented, {
...
   },
 
-  // TODO: if(DEBUG)
-  anyUnloaded() {
-    // Use `filter[0]` as opposed to `find` because of IE11
-    let unloaded = this.currentState.filter((im) => im._isDematerializing || !im.currentState.isLoaded)[0];
-    return !!unloaded;
+  // TODO refactor away _hasArrayObservers for tests
+  get _hasArrayObservers() {
+    return this.hasArrayObservers || this.__hasArrayObservers;
```
igorT, line 148: Why do we have both of these?

packages/record-data/addon/-private/relationships/state/has-many.ts
```diff
@@ -197,10 +174,6 @@ export default class ManyRelationship {
       payload.meta = this.meta;
     }
 
-    // TODO @runspired: the @igor refactor is too limiting for relationship state
```
igorT, line 200: What is this about?

igorT:
Do we know why this fails ilios?

igorT (approved):
Seems good after comments and CI passing

packages/model/addon/-private/system/many-array.js
runspired: The first is a built in from Ember, the second is a prop for us to override for tests where we want to trigger the observer codepath but didn't necessarily add an observer (or do, but need to control the timing of when the ManyArray believes it has an observer)

packages/model/addon/-private/system/many-array.js
```diff
@@ -229,32 +223,62 @@ export default EmberObject.extend(MutableArray, DeprecatedEvented, {
...
     }
 
-    this.flushCanonical(internalModels, true);
+    if (this._hasArrayObservers) {
+      // diff to find changes
+      let diff = diffArray(this.currentState, internalModels);
+      // it's null if no change found
+      if (diff.firstChangeIndex !== null) {
+        // we found a change
+        this.arrayContentWillChange(diff.firstChangeIndex, diff.removedCount, diff.addedCount);
```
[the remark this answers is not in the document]
runspired: No, the diff would encapsulate the first change to the last change. An array with several smaller diffs is actually just an array with one larger diff.

PR #7520, opened 2021-05-10
[chore] bump dependencies

runspired:
This PR

- bumps us to ember-source 3.27.1 (real import paths are the cause of the file size increase here)
- restores use of the official ember-cached-decorator-polyfill now that 0.1.3 fixes it for us
- adds ember-auto-import (necessary for unit 5.x) and moves to directly using pretender instead of ember-cli-pretender since we can with that.
- adds webpack as a direct dependency because ember-auto-import requires that
- drops support for node 10 (no longer within our support policy) and shifts default to node 14 from node 12. We now test node 12, 14, 16. Node 10 support was already dropped by multiple dependencies involved in this upgrade including ember-auto-import.
- upgrades us to ember-qunit 5.x (5.1.4)
- adds @ember/test-helpers, qunit, and qunit-dom as direct dependencies to support ember-qunit 5.x
- upgrades qunit config to account for ember-qunit 5.x changes
- bumps mocha
- removes fs-extra and github packages which we no longer use
- updates to latest fastboot and ember-cli-fastboot-testing for the fastboot test suite, which requires us bringing najax in on our own when testing the jquery fastboot support scenario. It also requires adding `node: 'current'` to targets for fastboot builds.
- updates our simple-import-sort plugin to the latest release which required a slight adjustment to our lint config
- fixes the ASSERT_ALL_DEPRECATIONS flag for the nightly job to check for deprecations. This has been passing because it was always undefined before, it will fail now.

In the process we pick up a drift of prettier that causes the bulk of the line changes here due to a change in the formatting of some code.


igorT (approved):
Looks good if CI works

packages/-ember-data/tests/index.html
```diff
@@ -1,41 +1,44 @@
...
-    {{content-for "test-body-footer"}}
-  </body>
-</html>
+
+<head>
+  <meta charset="utf-8">
+  <meta http-equiv="X-UA-Compatible" content="IE=edge">
+  <title>Dummy Tests</title>
+  <meta name="description" content="">
+  <meta name="viewport" content="width=device-width, initial-scale=1">
```
runspired, line 9: I'm not sure why prettier removed the indentation of the `head` tag 🤔. May have to double check that this wasn't my editor doing its own undesirable thing.

packages/-ember-data/tests/integration/identifiers/configuration-test.ts
```diff
@@ -22,7 +22,8 @@ import Store, {
 import { identifierCacheFor } from '@ember-data/store/-private';
 
 type StableRecordIdentifier = import('@ember-data/store/-private/ts-interfaces/identifier').StableRecordIdentifier;
-type ExistingResourceObject = import('@ember-data/store/-private/ts-interfaces/ember-data-json-api').ExistingResourceObject;
+type ExistingResourceObject =
+  import('@ember-data/store/-private/ts-interfaces/ember-data-json-api').ExistingResourceObject;
```
runspired, line 27: in-range updates to lint and prettier enforce the max-line-length for these type declarations now (they didn't before) which is why some of them now split lines.

packages/-ember-data/tests/integration/relationships/belongs-to-test.js
```diff
@@ -889,53 +887,6 @@ module('integration/relationship/belongs_to Belongs-To Relationships', function
     assert.ok(comment instanceof Message, 'a comment is an instance of a message');
   });
 
-  test('relationshipsByName does not cache a factory', async function (assert) {
```
runspired, line 892: this test covered a case where in pre 1.x versions of ember-data we stashed the factory onto relationship meta directly as the `type` field. It has not been the case that we do this since. The `type` field today is a string and other tests would fail were it not. This test no longer has value, so I removed it.

That said, it also failed because the `this` context of the test is registered as a destroyable by `setupTest` and thus cannot be torndown and then reused. We don't have any other tests like this but if we ever add one the solution is to not use `setupTest` at all for this test and to instead setupContext/teardownContext with objects we control.

packages/-ember-data/tests/test-helper.js
```diff
@@ -17,6 +18,13 @@ if (window.Promise === undefined) {
   window.Promise = RSVP.Promise;
 }
 
+// Handle testing feature flags
+if (QUnit.urlParams.enableoptionalfeatures) {
+  window.EmberDataENV = { ENABLE_OPTIONAL_FEATURES: true };
+}
```
runspired, line 24: this used to be in the tests/index.html file but QUnit is no longer a global.

packages/model/index.js
```diff
@@ -10,6 +10,8 @@ module.exports = Object.assign({}, addonBaseConfig, {
   shouldRollupPrivate: true,
   externalDependenciesForPrivateModule() {
     return [
+      'ember-cached-decorator-polyfill',
```
runspired, line 13: the polyfill rewrites the import from `@glimmer/tracking` to be from itself so we declare it here now.

packages/store/addon/-private/system/backburner.js
```diff
@@ -2,7 +2,7 @@
   @module @ember-data/store
 */
 
-import { registerWaiter } from '@ember/test';
+// import { registerWaiter } from '@ember/test';
 import { DEBUG } from '@glimmer/env';
 import Ember from 'ember';
```
runspired, line 7: leaving this commented out was a mistake, will restore

PR #7521, opened 2021-05-12
[WIP] Lazy Relationships & Relationship Diffing

runspired:
This PR aims to make **BOTH** the calculation of canonical state AND the calculation of current state happen on-demand (e.g. only when the UI-layer has "pulled" on the relationship to retrieve the latest state).

The deferring calculation of localState is done by changing the storage mechanism for ManyRelationship and ImplicitRelationship to a `diff` based data structure. This allows us to efficiently update the graph with better performance characteristics when an inverse changes while also avoiding the more-expensive flush of that state until the ui requests it specifically.

Deferring calculation of remoteState is done via a similar mechanism to the lazy-relationships feature in 3.4. Except now, because this mechanism is built into the relationship layer directly, we have the ability to retain laziness for updates even after a relationship has been accessed for the first time. In the 3.4 version once the relationship had materialized the lazy codepaths would no longer be utilized for further updates. The benefit of this new approach is that data that was formerly displayed to the UI but now is not will not result in expensive re-calculations until it is used again.


runspired:
somehow this auto-closed and I can't reopen it :(

PR #7530, opened 2021-05-20
bump dependabot to weekly (vs monthly) and remove unneeded ignores

runspired:
Bumping to weekly because monthly has meant we've lagged on getting updates we actually needed.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,46 +1,7 @@
 version: 2
 updates:
-- package-ecosystem: npm
-  directory: "/"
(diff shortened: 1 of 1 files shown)


igorT:
Should we still not have dependabot not bump lerna and TS, otherwise we might get ourselves into trouble?

runspired:
@igorT it won't get us in trouble unless we merge a red PR, and the current behavior is that it still attempts bumps, the ignores are for specific exact versions.

igorT (approved):

PR #7531, opened 2021-05-21
check debug package presence before import Fix #7053

### LINKED ISSUE #7053: Could not find module `@ember-data/debug` imported from `ember-data/index`
### Description

When I update my ember-data to 3.16.0 and deploy to a production env the error **Could not find module `@ember-data/debug` imported from `ember-data/index`** happen once I try to access the page through a browser.

Maybe it is a compatibility issue with fastboot since the error is being trowed by node and not by the browser.

When I rollback to 3.15.0 it works as expected.

### Versions

```cli
├─ @ember-data/adapter@3.16.0
├─ @ember-data/canary-features@3.16.0
├─ @ember-data/debug@3.16.0
├─ @ember-data/model@3.16.0
├─ @ember-data/private-build-infra@3.16.0
├─ @ember-data/record-data@3.16.0
├─ @ember-data/serializer@3.16.0
├─ @ember-data/store@3.16.0
├─ ember-data-factory-guy@3.9.5
├─ ember-data-storefront@0.17.2
└─ ember-data@3.16.0
```

I've tried to do a reproduction repository, but the error happens only on production env with a node app serving the app with fastboot.

sly7-7:
As per discussion with @runspired, I understood it is a temporary fix, and ideally we should use the `HAS_DEBUG_PACKAGE` flag from build-infra.

### DIFF
--- a/packages/-ember-data/addon/index.js
+++ b/packages/-ember-data/addon/index.js
@@ -89,7 +89,7 @@ DS.errorsArrayToHash = errorsArrayToHash;
 DS.Serializer = Serializer;
 
 if (HAS_DEBUG_PACKAGE) {
-  DS.DebugAdapter = require('@ember/debug').default;
+  DS.DebugAdapter = require('@ember-data/debug').default;
 }
 
 DS.RecordArray = RecordArray;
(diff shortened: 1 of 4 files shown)


runspired (approved):

runspired:
not clear why the its 3.20 and 3.24 tests are timing out here, I'll check locally.

PR #7532, opened 2021-05-21
fix: belongsTo should not attempt load if inverse in payload provided its data (#7049)

### LINKED ISSUE #7049: Regression: Ember Data tries to reload an already-loaded `belongsTo`
This is essentially exactly the same issue that was reported in #4099

### Reproduction

Ember Twiddle: [Here](https://ember-twiddle.com/e03e0f9a6e0a702aad47ea1f180deb9c?openFiles=twiddle.json%2C)

If you have a console open while running this twiddle, you will observe the following output:

```
Requesting http://localhost:3000/authors/1
Requesting http://localhost:3000/posts/1/author
Requesting http://localhost:3000/posts/2/author
Requesting http://localhost:3000/posts/3/author
Requesting http://localhost:3000/posts/4/author
```

Expected output:

```
Requesting http://localhost:3000/authors/1
```

### Description

If a `hasMany` relationship is loaded through includes (e.g. `store.findRecord('author', 1, { include: 'posts' })`) then the inverse `belongsTo` (e.g. `post.author`) is still not considered loaded and consuming it will result in unnecessary additional requests.

### Versions

This behaves properly in EmberData 3.4.4, but it does not in EmberData 3.5.2, and all the way through the current release.

sly7-7:
resolves #7049

### DIFF
--- a/packages/-ember-data/tests/integration/relationships/one-to-many-test.js
+++ b/packages/-ember-data/tests/integration/relationships/one-to-many-test.js
@@ -87,6 +87,61 @@ module('integration/relationships/one_to_many_test - OneToMany relationships', f
     });
   });
 
+  test("Adapter's findBelongsTo must not be hit when the record is included with its owner", function (assert) {
+    let store = this.owner.lookup('service:store');
+
+    this.owner.register(
+      'adapter:message',
+      Adapter.extend({
+        findBelongsTo() {
+          assert.fail('We should not call adapter.findBelongsTo since the owner is already loaded');
+        },
+      })
+    );
+
+    var user;
+    run(function () {
+      user = store.push({
+        data: {
+          id: '1',
+          type: 'user',
+          attributes: {
+            name: 'Stanley',
+          },
+          relationships: {
+            messages: {
+              data: [
+                {
+                  id: '2',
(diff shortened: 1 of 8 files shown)


packages/-ember-data/tests/integration/relationships/one-to-many-test.js
```diff
@@ -126,6 +131,18 @@ module('integration/relationships/one_to_many_test - OneToMany relationships', f
             attributes: {
               title: 'EmberFest was great',
             },
+            relationships: {
+              user: {
+                // data: {
```
sly7-7, line 136: uncomment these lines, and the test becomes green

runspired: Thanks! definitely had hoped the current state of the internals would have cleared this one up, but shouldn't be too hard to fix.

runspired: happy to pair with you on it, can teach you a bit about how it works that way :)

sly7-7: 👍

PR #7537, opened 2021-05-22
chore: upgrades to tsconfig

runspired:
This PR works towards enabling stricter type checking where now feasible. It's the first step towards removing our allowance of `implicitAny`. The goal is to assist in converting remaining JS files to TS.

cc @chriskrycho 

~~I'm also working on activation of `no-implicit-this` but it requires significantly restructuring how we create the build-url-mixin as the stricter checks surface that we don't properly narrow in it, and you can't narrow on an object method via the signature unless you pull the method definition out as a function and author in on conjunction with an `Interface` for the `this` context.~~ noImplicitThis is also activated :)


packages/-ember-data/tests/integration/identifiers/configuration-test.ts
```diff
@@ -328,7 +361,10 @@ module('Integration | Identifiers - configuration', function (hooks) {
     this.owner.register('serializer:application', TestSerializer);
 
     let generateLidCalls = 0;
-    setIdentifierGenerationMethod((resource: ExistingResourceObject) => {
+    setIdentifierGenerationMethod((resource: ResourceData | { type: string }) => {
```
igorT, line 364: It doesn't seem right that you accept `{ type: string }` but then error out if you get it

packages/-ember-data/tests/integration/identifiers/configuration-test.ts
```diff
@@ -254,9 +277,19 @@ module('Integration | Identifiers - configuration', function (hooks) {
     let updateMethodCalls = 0 as number;
     let updateCallback: (...args: any[]) => void;
 
-    function updateMethod(identifier: StableRecordIdentifier, data: ExistingResourceObject, bucket: string) {
-      updateMethodCalls++;
-      updateCallback!(identifier, data);
+    function updateMethod(
+      identifier: StableIdentifier | StableRecordIdentifier,
```
igorT, line 281: Isn't a StableIdentifier a subset of StableRecordIdentifier, you shouldn't need both?

packages/-ember-data/tests/integration/identifiers/configuration-test.ts
runspired: This is because the implementation in this test expects to not be asked to generate a new identifier.

packages/-ember-data/tests/integration/identifiers/configuration-test.ts
runspired: It is but once you activate strict type signatures you can't implement something that accepts only the wider interface.

packages/adapter/addon/-private/build-url-mixin.ts
```diff
@@ -12,6 +11,113 @@ type SnapshotRecordArray = import('@ember-data/store/-private/system/snapshot-re
...
+ The structure of this file is such because typing Mixins is hard. Here we've structured it in
+ such a way as to maximize the type information that a consumer can utilize. There are simpler
+ ways to type a mixin but we would not be able to provide the nice overload signature for buildURL
+*/
+// the interface must fully declare the function signatures that the individual functions
+// will also declare. If instead we try to keep them in sync by doing something like
+// `interface BuildURLMixin { buildURL: typeof buildURL }`
+// then an extending class overwriting one of the methods will break because typescript
+// thinks it is a switch from an instance prop (that is a method) to an instance method.
+interface BuildURLMixin {
```
igorT, line 24: Would love to get @rwjblue or @chriskrycho to take a look here

packages/adapter/addon/-private/utils/determine-body-promise.ts
```diff
@@ -3,8 +3,10 @@ import { DEBUG } from '@glimmer/env';
 
 import continueOnReject from './continue-on-reject';
 
+type Dict<T> = import('@ember-data/store/-private/ts-interfaces/utils').Dict<T>;
+
 type RequestData = import('../../rest').RequestData;
-type Payload = object | string | undefined;
+type Payload = Dict<unknown> | string | undefined;
```
igorT, line 9: Is it guaranteed to not be an array?

packages/record-data/addon/-private/graph/-edge-definition.ts
```diff
@@ -77,7 +77,7 @@ function upgradeMeta(meta: RelationshipSchema): UpgradedMeta {
   niceMeta.isCollection = meta.kind === 'hasMany';
   niceMeta.isPolymorphic = options && !!options.polymorphic;
 
-  niceMeta.inverseKey = options && options.inverse;
+  niceMeta.inverseKey = (options && options.inverse) || STR_LATER;
```
igorT, line 80: Whats `STR_LATER`?

packages/record-data/addon/-private/record-data.ts
```diff
@@ -293,7 +293,7 @@ export default class RecordDataDefault implements RelationshipRecordData {
     }
 
     this._isNew = false;
-    let newCanonicalAttributes: AttributesHash | null = null;
+    let newCanonicalAttributes: AttributesHash | undefined;
```
igorT, line 296: Don't we want explicit null instead of setting to undefined?

packages/adapter/addon/-private/build-url-mixin.ts
runspired: I talked this over with Chris but yes, would still like his review.

packages/record-data/addon/-private/graph/-edge-definition.ts
runspired: its defined above, it's a way for us to initialize to an empty string while making it clear we are going to make it non-empty later within the file.

packages/record-data/addon/-private/record-data.ts
runspired: no, `attributes: null` is not a valid json:api payload, either the hash isn't present at all or its a hash.

packages/adapter/addon/-private/utils/determine-body-promise.ts
runspired: unsure, but `object` didn't cover anything but literally `Object` itself, can check if this should also include array.

runspired: since JSON.parse can parse a top level array then yes, so I've added `unknown[]` and made fixes to types accordingly.

chriskrycho (commented):
Am about ⅔ of the way through this, have a couple initial comments, will finish reviewing later today or tomorrow!

packages/store/addon/-private/system/core-store.ts
```diff
@@ -109,7 +109,7 @@ type PendingFetchItem = {
 };
 type PendingSaveItem = {
   snapshot: Snapshot;
-  resolver: RSVP.Deferred<InternalModel | void>;
+  resolver: RSVP.Deferred<InternalModel> | RSVP.Deferred<void>;
```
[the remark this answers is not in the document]
runspired: The return type isn't "either" but rather changes based on feature flag state. It isn't a `deferred` that may resolve to `InternalModel | void` but rather it is either the case that the deferred always resolves to InternalModel or it is the case that deferred always resolves to void.

packages/store/addon/-private/system/model/shim-model-class.ts
```diff
@@ -57,25 +57,31 @@ export default class ShimModelClass implements ModelSchema {
     return mapFromHash(relationships);
   }
 
-  eachAttribute<T>(callback: (this: T, key: string, attribute: AttributeSchema) => void, binding?: T) {
+  eachAttribute<T>(callback: (this: T | undefined, key: string, attribute: AttributeSchema) => void, binding?: T) {
     let attrDefs = this.__store._attributesDefinitionFor(this.modelName);
     Object.keys(attrDefs).forEach((key) => {
-      callback.call(binding, key, attrDefs[key]);
+      callback.call(binding, key, attrDefs[key] as AttributeSchema);
```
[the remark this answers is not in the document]
runspired: This is just the usual issue with Object.keys, typescript has intentionally *not* typed it to narrow for keys just in case there are other props on the object which keys would see, in our case we are sure that this is not the case and so since the key exists the type is AttributeSchema and not undefined.

packages/store/addon/-private/system/snapshot.ts
```diff
@@ -494,7 +494,7 @@ export default class Snapshot implements Snapshot {
     if (CUSTOM_MODEL_CLASS) {
       let attrDefs = this._store._attributesDefinitionFor(this.modelName, this.identifier);
       Object.keys(attrDefs).forEach((key) => {
-        callback.call(binding, key, attrDefs[key]);
+        callback.call(binding, key, attrDefs[key] as AttributeSchema);
```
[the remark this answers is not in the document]
runspired: see answer above

packages/-ember-data/tests/integration/identifiers/configuration-test.ts
```diff
@@ -148,9 +151,19 @@ module('Integration | Identifiers - configuration', function (hooks) {
...
+      data: ResourceData | unknown,
+      bucket: IdentifierBucket
+    ) {
+      switch (bucket) {
+        case 'record':
+          updateMethodCalls++;
+          updateCallback!(identifier, data);
```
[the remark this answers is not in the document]
runspired: the unsafe cast isn't new here (though it seems it due to the indentation switch) but the TL;DR is it is defined in the the individual test setup, but we define it with `let` because this is is shared setup for each test.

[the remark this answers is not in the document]
chriskrycho: Ah, got it. Thanks for the clarification.

packages/store/addon/-private/system/core-store.ts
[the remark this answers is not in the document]
runspired: so after digging through, looks like the return types here diverged from implementation over time and we only needed Promise<void> (and void | Promise<void> for scheduleFetch). Fixed an implementation bug and fixed up the types.

packages/-ember-data/tests/dummy/app/routes/application/route.ts
```diff
@@ -1,16 +1,21 @@
 import Route from '@ember/routing/route';
+import { inject } from '@ember/service';
+
+type Store = import('@ember-data/store').default;
```
chriskrycho, line 4: Any objection to switching to `import type` for new stuff like this? Less work to do later.

```suggestion
import type Store from '@ember-data/store';
```

packages/adapter/addon/-private/build-url-mixin.ts
chriskrycho: Yeah, it's messy but until we're able to move away from mixins entirely (…could be a while 😬) I think this is least sucky option. It gives us the benefit of generally making sure we're keeping things in sync at the type level, providing decent user feedback via the overrides, and limiting where the damage of the casts has to be done. Mixins have *always* been the part of the system least amenable to type safety, precisely because it's very difficult to teach TypeScript what the type will be—which is fair, because the type in general can be arbitrarily dynamic.

packages/record-data/tests/integration/graph/edge-removal/abstract-edge-removal-test.ts
```diff
@@ -6,6 +6,9 @@ import { module, test as runTest } from 'qunit';
 import { setInitialState, testFinalState } from './helpers';
 import { setupGraphTest } from './setup';
 
+type TestConfig = import('./helpers').TestConfig;
+type Context = import('./setup').Context;
```
chriskrycho, line 10:
```suggestion
import type TestConfig  from './helpers';
import type Context from './setup';
```

packages/store/addon/-private/system/core-store.ts
```diff
@@ -328,10 +328,12 @@ abstract class CoreStore extends Service {
           number: 'NumberTransform',
           string: 'StringTransform',
         };
+        type Mappings = keyof typeof Mapping;
+        const keys = Object.keys(Mapping) as Mappings[];
         let shouldWarn = false;
 
         let owner = getOwner(this);
-        Object.keys(Mapping).forEach((attributeType: keyof typeof Mapping) => {
+        keys.forEach((attributeType: Mappings) => {
```
chriskrycho, line 336: The cast `as Mappings[]` should result in this type inferring correctly, so if you like you can drop the explicit annotation here:

```suggestion
        keys.forEach((attributeType) => {
```

packages/store/addon/-private/system/core-store.ts
```diff
@@ -2487,19 +2489,15 @@ abstract class CoreStore extends Service {
...
-    options
-  ): void | RSVP.Promise<void> {
-    let snapshot = internalModel.createSnapshot(options);
-    if (internalModel._isRecordFullyDeleted()) {
-      resolver.resolve();
-      return resolver.promise as RSVP.Promise<void>;
-    }
-
-    internalModel.adapterWillCommit();
+  scheduleSave(internalModel: InternalModel, resolver: RSVP.Deferred<void>, options): void | RSVP.Promise<void> {
```
chriskrycho, line 2492: 🎉

packages/store/addon/-private/ts-interfaces/identifier.ts
```diff
@@ -3,6 +3,12 @@
 */
 import { symbol } from '../utils/symbol';
 
+type ExistingResourceObject = import('./ember-data-json-api').ExistingResourceObject;
+type ResourceIdentifierObject = import('./ember-data-json-api').ResourceIdentifierObject;
+
+export type ResourceData = ResourceIdentifierObject | ExistingResourceObject;
+export type IdentifierBucket = 'record' | never;
```
chriskrycho, line 10: This is equivalent to `type IdentifierBucket = 'record'`; TypeScript throws away the `| never` because `never` is the "bottom" type, which no value can ever "inhabit". In a parallel to the way that `T | any` is always just `any`: `T | never` is always just `T`.

Based on what I see below, the idea is presumably to make `GenerationMethod` more robust? I think semantically, this ends up being equivalent to doing `bucket?: IdentifierBucket` there.

packages/store/addon/-private/ts-interfaces/identifier.ts
```diff
@@ -194,7 +200,15 @@ export type GenerationMethod = (data: Object, bucket: string) => string;
...
-export type UpdateMethod = (identifier: StableIdentifier, newData: Object, bucket: string) => void;
+export type UpdateMethod = {
+  (identifier: StableRecordIdentifier, newData: ResourceData, bucket: 'record'): void;
+  (identifier: StableIdentifier, newData: unknown, bucket: never): void;
+  (
+    identifier: StableIdentifier | StableRecordIdentifier,
+    newData: ResourceData | unknown,
+    bucket: IdentifierBucket
+  ): void;
+};
```
chriskrycho, line 211: For the second overload, I *think* we want to leave off `bucket` rather than setting it as `never`. If you include the param but set it to never, callers invoking with that signature will get an error if they only pass two arguments. Additionally, I *think* you don't need the final overload at all here – only at the implementation site(s).

```suggestion
export type UpdateMethodAlt = {
  (identifier: StableIdentifier, newData: unknown): void;
  (identifier: StableRecordIdentifier, newData: ResourceData, bucket: 'record'): void;
};
```

See [playground](https://www.typescriptlang.org/play?#code/PTAEElQCwQwNwKagHYHtQEsAmCagO6wAuoRUSGyRCATsjADYDOoqAZqeU0jDQgDSgm6SACsArkxIBbGDk4JpoPgxjUAUCFAwADjpqp9GNQgYBPUACMY3LK2QFiEUACosqF6ADWafNsuo4grSAHSgAFyiqGbh6uqU1DRsMADGSADKRDCWDAjgOFQYbBi0oADeoCmoqAwAQrjITOGgAORmqC2gAL4A3PFUtMlpoJnZuQBKCFU0WPkIhcWlCAAe1MhYLKM5eQVERSU05Y4miDTNbbhQnb3qRGY6SJPC4jRpACJqeAC8RwD6vzkYFBmkQaOIkDc7g8ILt9rRauIUl4ECQflIaJQAOZ9dQrHSoGgkKFIACqOiwJgAsiioKg7D8yupQKAABTYeZ7RZnEZZbaTaazWFcwTIBD4D5ZZpPQKvBASmCCSyI5FEc58AUtACUzTgqGwfWZbKFB2aW1ycwWBxFYvlzXEyB8qHwyEVypRzVFp21oF1+qZrP9zPZltopt55uNpQAPjyxgh+QTBRy4TR+IGUDbPlKEM9ZfLQDH7Y7nWnmcylUj3TDk1yEZWiP7vb6sH0buocClVHxQLkSMtmmSKdRqWQ6X1liyKlUavUYI1mgAidoL7qCFqEE60LU4vEEon3Unkqk0ukAQQYqPK-qNNZNse2Fs5Voz4qzoCLvmQTb1Levwafob3hMUyJo+KbWq+kqgNKLzvJ8rr1mqIEzFqOo-q2OIdl2SC9qAMSgIOx6jlg55EH0ZiTpU1R1A0TSgEuqArl0a4btQpzbnEuEAF4Dkew4niRF6gD8N4htyZo7Le0ZAfGyFJmJEG2u+DqfghKoAPxIRq37YMJAB8V5lpgHAsuuxAIOxmAOP+KaaoZRllloITOemXTqG5QA) for a demo that this simplification works *if* I understand the intent correctly.

chriskrycho (changes requested):
99% there; this is a really good push forward. Had a couple places I think we can simplify these a bit, but 👍🏼

packages/-ember-data/tests/dummy/app/routes/application/route.ts
runspired: I've opened an issue for it and someone has agreed to take it on. I think it's as simple as removing our plugin and configuring the new eslint-typescript rules which autofix, then running the fix. I'm going to let them have a chance to have a stab at it before I jump in though.

packages/store/addon/-private/ts-interfaces/identifier.ts
runspired: Is already noticed we didn't need the implementation version within the interface and could leave only the overloads, so good call I'll remove it.

The thing about the second overload though is it is intended to be an error if invoked. It forces the user to implement both a signature and a JS implementation that can handle extension later. I'll demonstrate this to you if you'd like, maybe we can firm it up a little.

runspired: We discussed this out of band and the two overloads are the desired overloads, as this is an interface for a user implemented function and the second overload is serving as a placeholder to ensure the implemented function is compatible with us extending this interface with additional overloads later.

packages/store/addon/-private/system/core-store.ts
runspired: This can be removed but I additionally cleaned up the naming so done as a commit vs accepting suggestion.

PR #7544, opened 2021-05-25
test: ensure coalesced find requests properly associate options and includes with their snapshot

### LINKED ISSUE #5050: Accessing `include` query parameter with `coalesceFIndRequests = true`

sly7-7:
The underlying issue was resolved with the `optionsMap` passed to _findMany. So the test from #5054 is passing without the other changes. I think this test should be kept, because I didn't see other scenarios where both coalescing+adapterOptions where tested at the same time.

closes #5054 
resolves #5050

### DIFF
--- a/packages/-ember-data/tests/unit/store/adapter-interop-test.js
+++ b/packages/-ember-data/tests/unit/store/adapter-interop-test.js
@@ -116,18 +116,22 @@ module('unit/store/adapter-interop - Store working with a Adapter', function (ho
     });
   });
 
-  test('Coalesced Store#findRecord requests retain the `include` adapter option in the snapshots passed to adapter#findMany and adapter#findRecord', function (assert) {
+  test('Coalesced Store#findRecord requests retain the `include` adapter option in the snapshots passed to adapter#findMany and adapter#findRecord', async function (assert) {
     const includedResourcesForIds = {
       1: 'someResource',
       2: 'differentResource',
       3: 'anotherResource',
     };
-
-    assert.expect(Object.keys(includedResourcesForIds).length);
+    const options = {
+      1: { aQuery: 'for 1' },
+      2: { bQuery: 'for 2' },
+      3: { cQuery: 'for 3' },
+    };
 
     const ApplicationAdapter = Adapter.extend({
       groupRecordsForFindMany(store, snapshots) {
         // this ensures that `findRecord` is also called
+        // by grouping for all but 1
         return [snapshots.slice(0, -1), snapshots.slice(-1)];
       },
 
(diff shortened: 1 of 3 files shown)


packages/-ember-data/tests/unit/store/adapter-interop-test.js
```diff
@@ -116,6 +116,64 @@ module('unit/store/adapter-interop - Store working with a Adapter', function (ho
...
+      coalesceFindRequests: true,
+    });
+
+    this.owner.register('adapter:application', ApplicationAdapter);
+    this.owner.register('serializer:application', JSONAPISerializer.extend());
+    this.owner.register('model:test', Model.extend());
+
+    let store = this.owner.lookup('service:store');
+
+    return run(() => {
```
runspired, line 168: we should make the test async and await the all, no need for the `run` wrapper at all anymore.

runspired (approved):
One small nit but otherwise 👍🏻 

Thanks for reviving this!

packages/-ember-data/tests/unit/store/adapter-interop-test.js
sly7-7: Oh, yes, of course

sly7-7: 🤔  I migrate to async/await only for this test ? Seems like the whole file still rely on run() :(

runspired: Just the new test for now, but test cleanup PRs are always welcome!

PR #7546, opened 2021-05-25
chore: don't minify IE11 builds for tests to improve debugging experience

runspired:


packages/-ember-data/ember-cli-build.js
```diff
@@ -4,7 +4,17 @@ const EmberAddon = require('ember-cli/lib/broccoli/ember-addon');
 
 module.exports = function (defaults) {
   const isProd = process.env.EMBER_ENV === 'production';
+  const needsIE11 = !!process.env.TARGET_IE11;
```
snewcomer, line 7: Does this come across as a string?

runspired: if its defined at all it works, its the same check we do elsewhere. That said we should also ensure we are in the test environment because otherwise this will affect the asset size check build negatively.

snewcomer: Oh right we never set it to `false`

snewcomer (approved):

Issue #7479, opened 2021-04-26
wrong JSONAPISerializer imports


rinoldsimon:
### Description

All the `JSONAPISerializer` imports over here https://api.emberjs.com/ember-data/3.26/classes/JSONAPISerializer/methods/serialize?anchor=serialize points to `@ember-data/serializer/json`.

It should be from `@ember-data/serializer/json-api` ?
### Versions
```
"ember-data": "3.26"
```

runspired:
@skaterdav85 this could be an easy one for you

skaterdav85:
@runspired thanks! will work on it this weekend.

PR #7549, opened 2021-05-26
[DOC] Fixes JSONAPISerializer serialize documentation

skaterdav85:
Fixes https://github.com/emberjs/data/issues/7479

I took the description for `JSONSerializer#serialize` and reworked it for `JSONAPISerializer#serialize` to some symmetry.


packages/serializer/addon/json-api.js
```diff
@@ -471,6 +471,173 @@ const JSONAPISerializer = JSONSerializer.extend({
...
+    }
+    ```
+
+    By default, attributes are passed through as-is, unless
+    you specified an attribute type (`attr('date')`). If
+    you specify a transform, the JavaScript value will be
+    serialized when inserted into the JSON hash.
+
+    By default, belongs-to relationships are converted into
+    IDs when inserted into the JSON hash.
```
runspired, line 523: this isn't correct for JSON:API

skaterdav85: ah right, i need to do another pass on the text. i only updated the code samples.

packages/serializer/addon/json-api.js
```diff
@@ -471,6 +471,170 @@ const JSONAPISerializer = JSONSerializer.extend({
...
+    For example, consider this model:
+
+    ```app/models/comment.js
+    import Model, { attr, belongsTo } from '@ember-data/model';
+
+    export default class CommentModel extends Model {
+      @attr title;
+      @attr body;
+
+      @belongsTo('user') author;
```
runspired, line 488:
```suggestion
      @belongsTo('user', { async: false, inverse: null })
      author;
```

packages/serializer/addon/json-api.js
```diff
@@ -471,6 +471,170 @@ const JSONAPISerializer = JSONSerializer.extend({
...
+          }
+        }
+      }
+    }
+    ```
+
+    By default, attributes are passed through as-is, unless
+    you specified an attribute type (`attr('date')`). If
+    you specify a transform, the JavaScript value will be
+    serialized when inserted into the JSON hash.
```
runspired, line 517:
```suggestion
    normalized when inserted into the attributes hash.
```

packages/serializer/addon/json-api.js
```diff
@@ -471,6 +471,170 @@ const JSONAPISerializer = JSONSerializer.extend({
...
+    Belongs-to relationships are converted into JSON-API
+    resource identifier objects.
+
+    ## IDs
+
+    `serialize` takes an options hash with a single option:
+    `includeId`. If this option is `true`, `serialize` will,
+    by default include the ID in the JSON object it builds.
+
+    The adapter passes in `includeId: true` when serializing
```
runspired, line 528:
```suggestion
    The JSONAPIAdapter passes in `includeId: true` when serializing a record
    for `createRecord` or `updateRecord`.
```

This is true for both, the `rest` adapter sets it to true for `createRecord` which we inherit, and the json-api adapter sets it to true (by omission of last arg to the private `serializeIntoHash` helper) for `updateRecord`

packages/serializer/addon/json-api.js
```diff
@@ -471,6 +471,170 @@ const JSONAPISerializer = JSONSerializer.extend({
...
+    a record for `createRecord`, but not for `updateRecord`.
```
runspired, line 529:
```suggestion
```

packages/serializer/addon/json-api.js
```diff
@@ -471,6 +471,170 @@ const JSONAPISerializer = JSONSerializer.extend({
...
+
+    ## Customization
+
+    Your server may expect a different JSON format than the
```
runspired, line 533:
```suggestion
    Your server may expect a different data or JSON format than the
```

packages/serializer/addon/json-api.js
```diff
@@ -471,6 +471,170 @@ const JSONAPISerializer = JSONSerializer.extend({
...
+    built-in serialization format.
+
+    In that case, you can implement `serialize` yourself and
+    return a JSON hash of your choosing.
```
runspired, line 537:
```suggestion
    return a JSON hash of your choosing, or override this method and do the
    serialization in the adapter directly by using the provided snapshot.
```

packages/serializer/addon/json-api.js
```diff
@@ -471,6 +471,170 @@ const JSONAPISerializer = JSONSerializer.extend({
...
+    {
+      "TITLE": "Rails is omakase",
+      "BODY": "Yep. Omakase.",
+      "COMMENT_IDS": [ "1", "2", "3" ]
+    }
+    ```
+
+    ## Tweaking the Default JSON
+
+    If you just want to do some small tweaks on the default JSON,
```
runspired, line 613:
```suggestion
    If you just want to do some small tweaks on the default JSON:API formatted response,
```

packages/serializer/addon/json-api.js
```diff
@@ -471,6 +471,170 @@ const JSONAPISerializer = JSONSerializer.extend({
     return dasherize(key);
   },
 
+  // SERIALIZE
```
runspired, line 474:
```suggestion
```

packages/serializer/addon/json-api.js
runspired: ensuring the example uses the non-potentially-deprecated syntax

packages/serializer/addon/json-api.js
runspired: trying to avoid using JSON too much to keep disambiguated from the JSONSerializer

packages/serializer/addon/json-api.js
```diff
@@ -471,6 +471,170 @@ const JSONAPISerializer = JSONSerializer.extend({
...
+      "BODY": "Yep. Omakase.",
+      "COMMENT_IDS": [ "1", "2", "3" ]
+    }
+    ```
+
+    ## Tweaking the Default JSON
+
+    If you just want to do some small tweaks on the default JSON,
+    you can call `super.serialize` first and make the tweaks
+    on the returned JSON.
```
runspired, line 615:
```suggestion
    on the returned object.
```

packages/serializer/addon/json-api.js
```diff
@@ -471,6 +471,170 @@ const JSONAPISerializer = JSONSerializer.extend({
...
+
+    ```javascript
+    {
+      "TITLE": "Rails is omakase",
+      "BODY": "Yep. Omakase.",
+      "COMMENT_IDS": [ "1", "2", "3" ]
+    }
+    ```
+
+    ## Tweaking the Default JSON
```
runspired, line 611:
```suggestion
    ## Tweaking the Default Formatting
```

runspired (approved):

PR #7550, opened 2021-05-26
fix: add asserts and assert tests for belongsTo/hasMany/findRecord empty responses

### LINKED ISSUE #6920: TypeError Cannot read property 'id' of null if API response { "data": null }
If API response with `{ "data": null }` to a find request, this warning throws a `TypeError`: https://github.com/emberjs/data/blob/637d606522a49b111df33bc09a3f5fb28e8705d8/packages/store/addon/-private/system/store/finders.js#L53-L59

`payload.data` is `null`. Trying to access `id` on `null` throws.

Actually it doesn't make much sense to return `{ "data": null }` by our API if the resource does not exist but it's explicitly be allowed by JSON:API spec in some cases:

> A server MUST respond to a successful request to fetch an individual resource with a resource object or null provided as the response document’s primary data.
> 
> null is only an appropriate response when the requested URL is one that might correspond to a single resource, but doesn’t currently.
>
> https://jsonapi.org/format/#fetching-resources

I'm facing this issue in Ember Data `3.14.0` but from reading changelog and inspecting the code that has triggered it, I'm quite sure it's still the same for `3.16.0-beta.0`.

### LINKED ISSUE #7547: belongsTo/hasMany fetch by link with missing data key in returned payload should error.
When an async belongsTo or hasMany relationship with no known data (not known empty and not known not-empty but has a link) is fetched via link, an error should be thrown if the returned payload does not have a data key (no error should be thrown if the returned payload has a data key but it specifies `null` for belongsTo or specifies `[]` for hasMany).

See #7534 for a recent adjustment to this codepath that would be the likely point at which to assert this.

It's possible we do error today, but I didn't see a test at first glance.

runspired:
resolves #7547 
resolves #6920

In the case of belongsTo/hasMany this would already have thrown an error but it would have been a confusing one. You will still get a somewhat more confusing one if the response contains only links but not meta as it will fail `validateDocumentStructure` as JSON:API insists that documents have at-least one-of data/errors/meta

### DIFF
--- a/packages/-ember-data/tests/integration/records/load-test.js
+++ b/packages/-ember-data/tests/integration/records/load-test.js
@@ -9,6 +9,7 @@ import JSONAPIAdapter from '@ember-data/adapter/json-api';
 import Model, { attr, belongsTo } from '@ember-data/model';
 import JSONAPISerializer from '@ember-data/serializer/json-api';
 import Store from '@ember-data/store';
+import testInDebug from '@ember-data/unpublished-test-infra/test-support/test-in-debug';
 import todo from '@ember-data/unpublished-test-infra/test-support/todo';
 
 class Person extends Model {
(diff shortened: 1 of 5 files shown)


snewcomer (approved):
Nice debug assertions!

Issue #7401, opened 2020-12-27
pushing a record causes it to jump to the end of any hasMany it is in


BryanCrotaz:
### Reproduction

Writing a PR with a test and a resolution

### Description

pushing an unmodified record with `pushPayload` causes it to jump to the end of any hasMany it is in. It would be desired that the relationship should be unmodified as is the case with attributes.

### Versions

Run the following command and paste the output below: `yarn list ember-source && yarn list ember-cli && yarn list --pattern ember-data`.

```cli
warning Filtering by arguments is deprecated. Please use the pattern option instead.
└─ ember-source@3.19.0
✨  Done in 1.71s.
yarn list v1.22.4
warning Filtering by arguments is deprecated. Please use the pattern option instead.
└─ ember-cli@3.19.0
✨  Done in 1.77s.
yarn list v1.22.4
├─ @ember-data/adapter@3.23.0
├─ @ember-data/canary-features@3.23.0
├─ @ember-data/debug@3.23.0
├─ @ember-data/model@3.23.0
├─ @ember-data/private-build-infra@3.23.0
├─ @ember-data/record-data@3.23.0
├─ @ember-data/rfc395-data@0.0.4
├─ @ember-data/serializer@3.23.0
├─ @ember-data/store@3.23.0
├─ @silvercurve/pixelpipe-ember-data-core@2.0.27
├─ babel-plugin-ember-data-packages-polyfill@0.1.2
├─ ember-data-factory-guy@3.9.6
├─ ember-data-model-fragments@4.0.0
├─ ember-data-save-relationships@0.0.10
└─ ember-data@3.23.0
```

runspired:
I'm fairly sure this has been fixed with my changes that are currently in beta/canary. The test in #7402 should be brought in to confirm that.

runspired:
#7551 shows this issue no longer exists on master branch, which indicates that 3.28 beta 2 (published later this week) and all current canaries and possibly 3.28 beta 1 all have this fixed.

PR #7551, opened 2021-05-26
test: Ensure many-many order preserved when pushing individual records

sly7-7:
This is only a rebase of master against #7402. I've only taken the tests as per @runspired comment: https://github.com/emberjs/data/issues/7401#issuecomment-839466783


runspired:
Looks like the main issue is resolved but that the TODO test needs to remain a TODO for now. @sly7-7 if you'll revert the TODO test change and keep the new test I think this will be good to go.

BryanCrotaz:
Sorry that I was too busy to help but glad it's fixed!

sly7-7:
@runspired This should be good now 
@BryanCrotaz No problem, thanks a lot for your initial work on that

runspired (approved):

Issue #7555, opened 2021-05-27
Automate Projects via Labels


runspired:
Since we use projects to track lots of work, we can automate some of them better using https://github.com/philschatz/project-bot

For instance this could be used to surface PRs that still require a backport or still require tags for lerna-changelog configuration.

PR #6894, opened 2019-12-13
[CHORE] Install Backport bot

Gaurav0:
For #6103 

I forked https://github.com/tibdex/backport and created a version that will work with forked repos. I have opened https://github.com/tibdex/backport/pull/39 but until it is merged I've published the backport bot on my own username in the Github Marketplace, so we can use it right away. I'd like to go ahead and have us test it out in a large codebase like Ember Data

Issue #6103, opened 2019-05-13
🛤️ tracking: Release Automation


runspired:
The `EmberData` team recently landed a number of improvements to the `release publishing` and `CI` pipelines to reduce the cost overhead of releases, enable publishing canary builds to `npm`, and prepare for a more automated future.

This ticket outlines remaining work items we hope to accomplish to complete our quest for low-to-zero-overhead releases:

*CI*

- [x] ensure we match Travis for current Azure setup
   🔒Claimed by @HeroicEric https://github.com/emberjs/data/pull/6098
- [x] ~~Drop `travis.yml` and deactivate `Travis` for `ember-data`~~ We dropped all jobs we could. Once we have allowed-failure equivalent in gh-actions we will drop the rest.
- [x] validate that all PRs have a label for the targeted branch
- [x] validate that all PRs have a label for their kind (fix|chore|test|security etc.)

*Branch Management*
- [ ] document how to use the `backporting bot` feature for backporting fixes to `beta` and `release` branches
- [ ] improve `backporting bot` feature to account for fixes that need to target both beta and release branches
- [ ] configure the `backporting bot` to work via labels
- [ ]  [DISCUSS] `backporting bot` should run `lint --fix` automatically on backported commits

*Changelog Generation*

- [x] move to `lerna changelog` from custom solution
- [x] add changelog generation into the `lerna version` automation step in our release script
- [ ] investigate configuring our monorepo to publish release notes (possible if using `lerna changelog`)
- [x] Improve Changelog generated output (ref https://github.com/emberjs/data/issues/7928)
  - don't include beta/canary builds
  - don't include references to private/unpublished packages

*Publishing*

  - [x] Automate nightly publishing of canary to `npm registry`
     - [ ] failed & successful attempts should notify @core-data in #dev-ember-data on discord
     - [x] should only publish if changes exist since last canary
  - [x] configure CI with a token to be used in place of OTP for publishing to `npm`
  - [x] upgrade release script to confirm or bypass the `lerna version` prompt for the intended version bump
 
*Testing*
Automation also means we will want to ensure more robust testing.

- [ ] add tests for our automated release process (likely via https://verdaccio.org/)
- [x] IE11 integration tests #6431 
- [x] fastboot integration tests #6431 
- [ ] make it clear upstream needs to be set
- [x] lts release bumps patch version when it's just a "new channel", is this necessary?
- [ ] update guidance for using lerna-changelog always, not just lts
[...]

PR #7556, opened 2021-05-27
adds label enforcement for PRs

### LINKED ISSUE #6311: CI should fail if the PR does not have a canary/beta or release label

runspired:
resolves #6311

assists #6103
assists #7555 
assists #6894 

This PR aims to enforce that

- PRs are explicitly labeled for the changelog (or to skip the changelog)
- PRs are explicitly labeled for back porting needs (or whether they are the needed backport PR)
- that labels that shouldn't be applied within a context are not applied.

We will need to update the lts and lts-prev branch names when new lts versions occur.


.github/workflows/enforce-pr-labels-release.yml
```diff
@@ -0,0 +1,35 @@
+name: Enforce Release PR labeling
+
+on:
+  pull_request:
+    types: [labeled, unlabeled, opened, edited, synchronize]
```
snewcomer, line 5: Do you happen to know what a "synchronized" pull request is?  See any other ones we would want to add from the docs below?

https://docs.github.com/en/actions/reference/events-that-trigger-workflows#pull_request

snewcomer (approved):
Would be nice if they allowed a "common-" yml file that has overrided "on" steps for each branch that we target.  Otherwise, this seems like a great idea!

.github/workflows/enforce-pr-labels-release.yml
runspired: I believe it means that the PR is from a fork and the fork was updated. I went through those docs and this is the full list I think we need.

runspired: oooh, other than `reopened`, which I added to some but apparently not all. Will make sure all have it.

PR #7567, opened 2021-05-29
chore: modernized rest-adapter-test | changed .get to dot notation |  replaced post.reopen with individual definitions per test 

runnerboy22:

### DIFF
--- a/packages/-ember-data/tests/integration/adapter/rest-adapter-test.js
+++ b/packages/-ember-data/tests/integration/adapter/rest-adapter-test.js
@@ -313,8 +313,14 @@ module('integration/adapter/rest_adapter - REST Adapter', function (hooks) {
   });
 
   test('updateRecord - hasMany relationships faithfully reflect simultaneous adds and removes', async function (assert) {
-    Post.reopen({ comments: DS.hasMany('comment', { async: false }) });
+    Post = DS.Model.extend({
+      name: DS.attr('string'),
+      comments: DS.hasMany('comment', { async: false }),
+    });
     Comment.reopen({ post: DS.belongsTo('post', { async: false }) });
+
+    this.owner.register('model:post', Post);
+
     adapter.shouldBackgroundReloadRecord = () => false;
 
     store.push({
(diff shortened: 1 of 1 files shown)


packages/-ember-data/tests/integration/adapter/rest-adapter-test.js
```diff
@@ -313,8 +313,14 @@ module('integration/adapter/rest_adapter - REST Adapter', function (hooks) {
   });
 
   test('updateRecord - hasMany relationships faithfully reflect simultaneous adds and removes', async function (assert) {
-    Post.reopen({ comments: DS.hasMany('comment', { async: false }) });
+    Post = DS.Model.extend({
```
runspired, line 316: we should scope the variable to the test now that the definition is entirely within it.

```suggestion
    const Post = Model.extend({
```

Also, since this test file already uses the newer non `DS.` way of accessing the `Model` class and the `attr` `belongsTo` and `hasMany` functions we should use those anywhere we are adding a new definition. This can be done in a later pass if you want.

packages/-ember-data/tests/integration/adapter/rest-adapter-test.js
```diff
@@ -313,8 +313,14 @@ module('integration/adapter/rest_adapter - REST Adapter', function (hooks) {
...
+      name: DS.attr('string'),
+      comments: DS.hasMany('comment', { async: false }),
+    });
     Comment.reopen({ post: DS.belongsTo('post', { async: false }) });
+
+    this.owner.register('model:post', Post);
```
runspired, line 322: The `hooks.beforeEach` callback above still registers Post as well. Can we remove the registration in that callback and ensure that any other tests using `post` also register it directly?

packages/-ember-data/tests/integration/adapter/rest-adapter-test.js
```diff
@@ -952,7 +958,13 @@ module('integration/adapter/rest_adapter - REST Adapter', function (hooks) {
   });
 
   test('findMany - findMany uses a correct URL to access the records', async function (assert) {
-    Post.reopen({ comments: DS.hasMany('comment', { async: true }) });
+    Post = DS.Model.extend({
```
runspired, line 961: same as above, declare `Post` here instead of assigning to the variable with the greater scope access above. I'd do a sweep since there are likely more instances of this. If you can delete the `Post` variable declaration on line 23 you'll know you were successful.

runspired (changes requested):
Overall this lgtm but I'd like to remove the module-scope `Post` and the registration of a Post model in the hooks.beforeEach callback before we merge this.

In addition to assigning Post to a variable within each individual test, this may require adding a `Post` model definition in some tests that were relying on the beforeEach registration before.

Due to the way the beforeEach re-defines Post each time, any test that isn't explicitly registring Post at this time is using a Post model that has no relationships (no hasMany/belongsTo) and just the name `attr`.

PR #7568, opened 2021-05-29
Fixes `DeprecatedEvented` link

skaterdav85:
Closes https://github.com/ember-learn/ember-api-docs/issues/759

image

### DIFF
--- a/packages/store/addon/-private/system/deprecated-evented.js
+++ b/packages/store/addon/-private/system/deprecated-evented.js
@@ -27,10 +27,15 @@ if (DEBUG) {
     return deprecations;
   };
 
+  /**
+   * `DeprecatedEvented` is a mixin that proxies to the `Ember.Evented`
+   * mixin without logging deprecation warnings.
+   *
+   * @class DeprecatedEvented
+   * @private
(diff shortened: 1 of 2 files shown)


packages/store/addon/-private/system/deprecated-evented.js
```diff
@@ -27,10 +27,15 @@ if (DEBUG) {
     return deprecations;
   };
 
+  /**
+   * `DeprecatedEvented` is a mixin that proxies to the `Ember.Evented`
+   * mixin without logging deprecation warnings.
```
runspired, line 32:
```suggestion
   * mixin while logging deprecations. It is used by classes that were previously instrumented with
   * Evented whose evented APIs are now deprecated.
```

Overall LGTM but added some clarification. The `_has` being a workaround to not log the deprecation exists because we don't want to flood the user with deprecation notices while we are merely checking for whether they've registered a deprecated event which we will need to fire.

runspired (changes requested):
LGTM with a small language fix.

runspired (approved):

PR #7570, opened 2021-05-30
[Refactor]: private model store

snewcomer:
WIP

- [x] One single store. We can simplify the system by combining our two stores and use `HAS_MODEL_PACKAGE` to remove code if the end user does not have `@ember-data/model` installed.
~~- [ ] Move private APIs to `_privateStore` as delegate for internal code.~~


runspired (changes requested):
Overall this looks good, and I think we should land just-this and separate out the private stuff in a new PR after this lands. Both because this is still an incremental win *and* because this will help move forward some additional typescript cleanup (setting `allowJs: true`)

I'm happy to pair with you to go over the typescript stuff, I think it will take some tinkering to come up with the right answers. We may not get it ideal on this pass but we can always iterate on it later.

igorT (changes requested):
Results of the ember data meeting today:
1. The class merging moved to be discussed next week
2. Type fixes will be done in a separate PR

runspired:
replaced by #8078 and #8055

PR #7577, opened 2021-06-01
chore: store-adapter-test | removed run loop where possible

runnerboy22:

### DIFF
--- a/packages/-ember-data/tests/integration/adapter/store-adapter-test.js
+++ b/packages/-ember-data/tests/integration/adapter/store-adapter-test.js
@@ -16,12 +16,10 @@ import { Snapshot } from '@ember-data/store/-private';
 import testInDebug from '@ember-data/unpublished-test-infra/test-support/test-in-debug';
 
 function moveRecordOutOfInFlight(record) {
-  run(() => {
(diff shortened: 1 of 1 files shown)


packages/-ember-data/tests/integration/adapter/rest-adapter-test.js
```diff
@@ -313,8 +313,19 @@ module('integration/adapter/rest_adapter - REST Adapter', function (hooks) {
...
-    Comment.reopen({ post: DS.belongsTo('post', { async: false }) });
+    Post = DS.Model.extend({
+      name: DS.attr('string'),
+      comments: DS.hasMany('comment', { async: false }),
+    });
+
+    Comment = DS.Model.extend({
+      name: DS.attr('string'),
+      post: DS.belongsTo('post', { async: false }),
+    });
```
snewcomer, line 324: This is nice!  Btw, while we are here, should we move to `class Comment extends Model {`?

runspired: This was left over from commits in a separate PR included by accident, but yes, starting out new code in a nice state is preferable 👍🏻

PR #7581, opened 2021-06-02
Follow up: store identifier #7104

snewcomer:
ref #7104

### DIFF
--- a/packages/store/addon/-private/system/core-store.ts
+++ b/packages/store/addon/-private/system/core-store.ts
@@ -843,8 +843,7 @@ abstract class CoreStore extends Service {
     has already been assigned an id, you can find the record again using just the lid.
 
     ```app/routes/post.js
-    let identifier = recordIdentifierFor(person);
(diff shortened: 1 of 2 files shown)


packages/store/addon/-private/system/core-store.ts
```diff
@@ -843,8 +843,7 @@ abstract class CoreStore extends Service {
     has already been assigned an id, you can find the record again using just the lid.
 
     ```app/routes/post.js
-    let identifier = recordIdentifierFor(person);
-    store.findRecord({ lid: identifier.lid });
+    store.findRecord(identifier);
```
runspired, line 846:
```suggestion
    store.findRecord({ lid });
```

to make it clear it doesn't need type+id necessarily (which are already in example 2).
we could add a note that this also works by passing an identifier previously obtained from another ember-data api ala

```
Any identifier object returned by another ember-data API is able to be passed to this API.
```

or some such.

runspired (approved):
overall looks good, but seems we have a lint cleanup

PR #7582, opened 2021-06-03
chore: type DeprecatedEvented Mixin and Errors class

runspired:
The mechanics of `.extend()` syntax and Mixins are unwieldy and error prone, so the approach taken here is to build up a more correct type and cast vs relying on `.extend(Mixin)` to get it right. Typescript also does not allow classes to overwrite the method signatures of base classes, though JS does and we do this relatively often when building a primitive by extending from an Ember base class (such as MutableArray) or when using Mixins. Typescript rightly wants to enforce this to preserve the polymorphism of the base-class, but in our cases we do not care about this polymorphism and the base-class is just a transient construct used to compose the various mixins involved. So we use `Omit` to remove things we override the signature for.

I've chatted with @chriskrycho @dfreeman and @jamescdavis a bit about this effort, and these challenges are fairly well known and the patterns used here to solve them fairly typical. I've still requested their review.

**Note:**

I had to copy over all the @types/ember__array definitions and modify them (they were lacking in areas), specific changes include:

Not why I copied over:
- EmberArray.detect an intimate API we use in EmberData still which we've had a types override for for some time.
- any usage converted to unknown (to fit our project's enforcement of no any)

Why I copied over:
- Enumerable (and related) expecting args to be Enumerables when they can be just arrays.
- Some methods were using any instead of the generic type provided
- updating ArrayProxy to understand the difference between the content type and the handed out type.
- updating the type signature for MutableArray.replace to allow null/optional the last param and return void vs any

These are all improvements we should ensure are upstreamed.

**Note: the asset size increase is because Errors is now a native class**


packages/model/addon/-private/errors.ts
```diff
@@ -7,10 +7,27 @@ import { DEBUG } from '@glimmer/env';
...
 /**
   @module @ember-data/store
 */
 
+type ArrayProxyWithDeprecatedEventedType = new () => Omit<MutableArray<ValidationError>, 'clear'> &
+  Omit<ArrayProxy<ValidationError>, 'clear'> & { clear(): void; _has(name: string): boolean } & EmberObject &
+  typeof DeprecatedEvented;
+// we force the type here to our own construct because mixin and extend patterns
+// lose generic signatures. They also lose types generally when Omit is used on a constructable
+// type that is not a class, which we need to do for being able to override 'clear'.
```
dfreeman, line 26: The reason `Omit` breaks here isn't related to constructability (`MutableArray` is already the instance type), but because of the `.extend` call on [this line](https://github.com/emberjs/data/pull/7582/files#diff-5ad47250b4fc3ed0281d5adc0016528660a685c4e6edf234926ab67273e9dd50R26), which ends up leaking an index signature into the type of `MutableArray` and blows up `Omit`.

We should fix this in the upstream types as well—there's no reason to have that `.extend` there in the first place, since `MutableArray` is already included in the interface a couple lines up.

Fixing that would allow this block to become something like this, which to me is a bit more readable:

```ts
type Evented = import('@ember/object/evented').default;
type NativeArray<T> = import('@ember/array/-private/native-array').default<T>;
type ValidationError = {
  attribute: string;
  message: string;
};

/**
  @module @ember-data/store
*/

interface ArrayProxyWithDeprecatedEvented<T> extends Evented, Omit<ArrayProxy<T>, 'clear'> {
  clear(): void;
  _has(name: string): boolean;
}

// we force the type here to our own construct because mixin and extend patterns
// lose generic signatures. They also lose types generally when Omit is used on a constructable
// type that is not a class, which we need to do for being able to override 'clear'.
const ArrayProxyWithDeprecatedEvented = ArrayProxy.extend(DeprecatedEvented) as unknown as new <
  T
>() => ArrayProxyWithDeprecatedEvented<T>;
```

dfreeman: Looking at the source, though, I'm still a little confused why it's necessary to change the signature of `clear`. Is there a reason the `Errors` class couldn't `return this;` and actually adhere to the `MutableArray` protocol?

You're right of course that this is legal JavaScript in that you're not going to get an error directly from you overriding a method with a different return type, but it can be thorny for consumers e.g. if they had some generic logic meant to handle any `MutableArray` that expected that chaining to work (blah blah [Liskov](https://en.wikipedia.org/wiki/Liskov_substitution_principle) et cetera).

runspired: `Errors` may extend from ArrayProxy but it hardly behaves like one, and I don't think we want to increase it's API surface or capabilities (we're actively iterating away from extending Ember APIs at all).

This is also a fairly common pattern in general that we need to be able to have a type pattern for which works. This will especially bite us as we start typing serializers and the EmbeddedRecordsMixin.

PR #7589, opened 2021-06-10
chore: fix or silence-expected deprecations

runspired:
Resolves the issues causing canary publish and nightly chrons to fail.

Note: we will still need to "actually" resolve many of these deprecations for 4.0, but this wraps the ones that we know of currently in `expectDeprecation` to prevent the test suite from failing.

The `ember-global` deprecation when accessing Namespace is not expected, and the message it prints is a red-herring (that ember-cli-babel brought by ember-maybe-import-regenerator-runtime is outdated). That addon is not using Ember in any capacity, so while it should be updated it is not what is causing the deprecation spew.

When accessing:

```ts
import Namespace from '@ember/application/namespace';
Namespace.byName('DS');
```

Ember iterates through all the keys of `window`, and if they start with a capital letter it accesses them and checks if they may be a namespace. This leads to both the `Ember` global and the `Em` alias to the global triggering the globals deprecation, which sees the outdated ember-cli-babel and incorrectly directs attention there (this message would usually be correct, just in this instance it is not). cc @rwjblue 

We also now must set the `Resolver` and `moduleName` for the couple of tests where we create our own Application instance to test our initializer. These tests are mostly unnecessary at this point but for now I've just gone ahead and added these things. I'm not sure what shifted that we didn't encounter the `globals-resolver` deprecation for this previously, but I assume either the `Resolver` property is now eagerly checked or something is attempted to eagerly resolve that didn't before.

### DIFF
--- a/packages/-ember-data/tests/integration/application-test.js
+++ b/packages/-ember-data/tests/integration/application-test.js
@@ -1,6 +1,7 @@
 import Application from '@ember/application';
 import Namespace from '@ember/application/namespace';
 import Controller from '@ember/controller';
+import { assign } from '@ember/polyfills';
 import Service, { inject as service } from '@ember/service';
 
 import { module, test } from 'qunit';
@@ -25,7 +26,7 @@ module('integration/application - Injecting a Custom Store', function (hooks) {
       'controller:baz',
       class Controller {
         constructor(args) {
-          Object.assign(this, args);
+          assign(this, args);
         }
         @service('store') store;
         static create(args) {
@@ -87,7 +88,7 @@ module('integration/application - Injecting the Default Store', function (hooks)
 
   test('If a store is instantiated, it should be made available to each controller.', async function (assert) {
     let fooController = this.owner.lookup('controller:foo');
-    assert.ok(fooController.get('store') instanceof Store, 'the store was injected');
+    assert.ok(fooController.store instanceof Store, 'the store was injected');
   });
 
   test('the DS namespace should be accessible', async function (assert) {
--- a/packages/-ember-data/tests/integration/application-test.js
+++ b/packages/-ember-data/tests/integration/application-test.js
@@ -96,7 +96,7 @@ module('integration/application - Injecting the Default Store', function (hooks)
       () => {
         assert.ok(Namespace.byName('DS') instanceof Namespace, 'the DS namespace is accessible');
       },
-      { id: 'ember-global', count: 2 }
+      { id: 'ember-global', count: 2, when: { ember: '>=3.27.0' } }
     );
   });
 });
--- a/packages/-ember-data/tests/integration/application-test.js
+++ b/packages/-ember-data/tests/integration/application-test.js
@@ -7,11 +7,11 @@ import Service, { inject as service } from '@ember/service';
 import { module, test } from 'qunit';
 
 import initializeEmberData from 'ember-data/setup-container';
-import Store from 'ember-data/store';
 import { setupTest } from 'ember-qunit';
 import Resolver from 'ember-resolver';
 
 import JSONAPIAdapter from '@ember-data/adapter/json-api';
+import Store from '@ember-data/store';
 
 module('integration/application - Injecting a Custom Store', function (hooks) {
   setupTest(hooks);
--- a/packages/-ember-data/tests/integration/application-test.js
+++ b/packages/-ember-data/tests/integration/application-test.js
@@ -69,7 +69,7 @@ module('integration/application - Injecting the Default Store', function (hooks)
       'controller:baz',
       class Controller {
         constructor(args) {
-          Object.assign(this, args);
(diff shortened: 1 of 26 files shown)


packages/-ember-data/tests/integration/injection-test.js
```diff
@@ -53,12 +53,19 @@ module('integration/injection eager injections', function (hooks) {
...
+    // on models (e.g. that owner is properly setup for it).
+    assert.expectDeprecation(
+      () => {
+        let foo = store.createRecord('foo');
+        let apple = foo.get('apple');
+        let appleService = this.owner.lookup('service:apple');
+
+        assert.ok(apple, `'model:foo' instance should have an 'apple' property`);
+        assert.ok(apple === appleService, `'model:foo.apple' should be the apple service`);
+        assert.ok(apple instanceof Apple, `'model:foo'.apple should be an instance of 'service:apple'`);
```
snewcomer, line 66: Is there a specific line that triggers the deprecation?  That is, do we want the callback to only contain the offending line and the rest outside of the expectDeprecation block?

packages/-ember-data/tests/integration/injection-test.js
```diff
@@ -53,12 +53,19 @@ module('integration/injection eager injections', function (hooks) {
...
+      },
+      { id: 'implicit-injections', count: 1 }
```
snewcomer, line 68: I think one piece of this RFC is to remove `owner.inject` in 5.0.  Is the intention with listing this deprecation to leave it until then?  Do we have a spot we can explicitly inject the service on the foo model?

snewcomer (approved):
👍 💯 🥇  Just a few comments!

packages/-ember-data/tests/integration/application-test.js
```diff
@@ -75,9 +106,21 @@ module('integration/application - Using the store as a service', function (hooks
...
+          Object.assign(this, args);
+        }
+        @service('store') store;
+        static create(args) {
+          return new this(args);
+        }
+      },
+      { singleton: true, instantiate: true }
+    );
+    owner.register('controller:application', Controller.extend({ store: service() }));
```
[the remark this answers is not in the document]
runspired: not so much intentional as just these tests have always tried to "use with multiple", I think application + random to simulate the fact that application pretty much always exists by default but random wouldn't, but other than that I have no reasoning for why this test was written this way initially.

packages/-ember-data/tests/integration/injection-test.js
runspired: there isn't a specific line per say, the entirety of what this test is testing has been deprecated by ember.

packages/-ember-data/tests/integration/injection-test.js
runspired: we don't want to remove our injection until defaultStore is killed off from ember. There is no alternative for folks wanting injection other than for them to manually inject.

packages/store/types/ember-compatibility-helpers/index.d.ts
```diff
@@ -0,0 +1,17 @@
...
+export const HAS_NATIVE_COMPUTED_GETTERS: boolean;
+export const IS_GLIMMER_2: boolean;
+export const IS_RECORD_DATA: boolean;
+export const SUPPORTS_FACTORY_FOR: boolean;
+export const SUPPORTS_GET_OWNER: boolean;
+export const SUPPORTS_SET_OWNER: boolean;
+export const SUPPORTS_NEW_COMPUTED: boolean;
+export const SUPPORTS_INVERSE_BLOCK: boolean;
+export const SUPPORTS_CLOSURE_ACTIONS: boolean;
+export const SUPPORTS_UNIQ_BY_COMPUTED: boolean;
```
runspired, line 17: this can be removed once a new release of ember-compatibility-helpers is cut

packages/unpublished-test-infra/addon-test-support/qunit-asserts/assert-deprecation.ts
```diff
@@ -156,12 +161,32 @@ export function configureDeprecationHandler() {
...
+    if (config.when) {
+      let libs = Object.keys(config.when);
+      for (let i = 0; i < libs.length; i++) {
+        let library = libs[i];
+        let version = config.when[library]!;
+        let sanitizedVersion = version.replace(/[\^><=~]/g, '');
+
+        if (version.indexOf('<=') === 0) {
+          if (!lte(library, sanitizedVersion)) {
+            return;
```
runspired, line 173: we need to always execute the callback if it exists, so this isn't quite right.

runspired: we also need to push a "success" status in case any `assert.expect()` counts were configured for us.

PR #7591, opened 2021-06-11
Move files from bin to scripts folder

dcyriller:
This PR proposes to move the files used in the package.json's `scripts` field from the `bin` folder to a `scripts` folder.

It also runs `yarn lint:js --fix`. When the files were in the `bin` folder, the same command did not apply the `simple-import-sort` rules. I don't undestand why.

Why these changes? I was looking at the publish script and figured out these minor changes could make sense. Feel free to discard them if you think they don't!


runspired (approved):
Overall I appreciate this change. Historically we had shell scripts and a real reason for many of these things to be executables, but we no longer have shell scripts and leaving them as executables in a thing called `bin` no longer makes sense. Thanks!

Issue #7590, opened 2021-06-10
Should `normalizeErrorResponse` set error `status` as a string?


nlfurniss:
In the rest adapter, `normalizeErrorResponse` sets the `status` on the [error as a string](https://github.com/emberjs/data/blob/master/packages/adapter/addon/rest.ts#L1278). This seems odd, as all other instances of status in the adapter expect a number, and I believe most devs would as well.

Changing this seems like a breaking change, but happy to make the change if there isn't a good reason to keep it a string.

snewcomer:
Thank you for pointing this out!  So this luckily, the stringified numeric HTTP status code aligns with how JSON-API specifies the errors array should be.

> status: the HTTP status code applicable to this problem, expressed as a string value.

Overall, I agree with your initial thought.  However, this may have been a standard they finished on early in the process

https://jsonapi.org/format/#errors

nlfurniss:
Thanks, would not have guessed!

how would you feel about a PR adding a comment relaying the above info?

PR #7592, opened 2021-06-12
Add a comment explaining setting an errors status as a string

nlfurniss:
"Fixes" #7590.

I imagine I'm not the first nor the last person to wonder why `status` is a string when everywhere else in the adapter it's an integer.


runspired (approved):
Overall thanks for adding this!

I was looking at our documentation for errors lately and feel it's overall pretty lacking, you have to pick together the info from too many sources. We should make sure we update it as part of our sweep through the documentation. Opened #7593 to track that.

Issue #6233, opened 2019-07-09
Internal Typescript Conversion Roadmap


mike-north:
We've allowed implicit `any` in `.ts` files for a while. We should have some sort of plan for eventually requiring them to be _explicit_. This will entail typing function args, among other things.

cc: @runspired

runspired:
## A Roadmap! 🚀

**Background**

For some time we've allowed **implicit-any** since our migration is incremental (and slow due to low contributor availability for doing the work). This also works well for converting new files to typescript or making bugfixes as it means these things were not forcing functions for larger type refactors beyond the scope of the work being done. Recently, we hit enough typescript adoption that it became type to firm up our usage and determine the course to get us all the way. ~60% of library code at this point is written in typescript (however,  most of our tests are still written in javascript). One of the largest hurdles to converting our library code, refactoring `Model` to a native class, has also been cleared. With this in mind we've begun refining our tsconfig, and once #7537 lands `noImplicitAny` will be the *only* deactivated strict rule. Obviously we want to change this.

Relatedly, when we configured this project we kept [`allowJs`](https://www.typescriptlang.org/tsconfig#allowJs) disabled even though we were (and still mostly are when including tests) a JS project. This was because enabling it caused type-checking of JS files, including using JSDoc comments to infer type. Unfortunately, this conflicts with the yuidocs we use for the api documentation, and led to incorrect type inferences and type errors throughout that would have been unfixable. These inferred types would leak into typescript's type checking of TS files, and were _worse than an implicit any_. For these reasons we decided against `allowJs`, feeling there wasn't much of an option.

However, when `allowJs` is disabled, typescript code that calls JS code that calls typescript code loses access to the type information it would have otherwise had. This is also less-than-ideal. Fortunately, the typescript team appears to have realized this wasn't ideal for many teams and provided an escape hatch, albeit one we did not learn about until recently. When `allowJs` is enabled, we can disable JS typechecking as JSDoc annotation usage by disabling **[checkJs](https://www.typescriptlang.org/tsconfig#checkJs)**.  Armed with this context, we've determine our migration can be best completed in the following steps: 

**Caveats**

For the purposes of typescript migration, we are considering three goal posts:

- all library code using typescript in strict mode
- all tests of library code using typescript in strict mode
- all JS files for infra/blueprints etc (e.g. not library code) converted to ts.

[...]

PR #7597, opened 2021-06-18
[Chore]: remove prefer-type-only-import eslint rule

### LINKED ISSUE #7536: Convert to new type import syntax
Currently we enforce type-only imports be declared using the following syntax:

```ts
type Foo = import("./path").default;
type Boo = import("./path").Bar;
```

This has a few disadvantages

1) Some type information is lost
2) The syntax can be a bit unweidly
3) Named imports each require their own line

Fortunately there is a new type-only import syntax we can use:

```ts
import type Foo { Bar } from './path';
```

We should be able to do an automatic conversion by [removing our lint rule](https://github.com/emberjs/data/blob/master/packages/unpublished-eslint-rules/src/rules/prefer-type-only-import.js) and configuring the one that [eslint typescript now provides](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/consistent-type-imports.md)

snewcomer:
- [x] Step 1: remove custom eslint rule
- [x] Step 2: Remove unpublished package - should we keep the shell of this package for future use?
- [x] Step 3: Add `consistent-type-imports`
- [ ] Step 4: run eslint fix

close #7536 
ref https://github.com/emberjs/data/issues/6233#issuecomment-849279594


snewcomer:
close https://github.com/emberjs/data/pull/7598

PR #7598, opened 2021-06-19
[Chore]: prefer static type-only-import typescript eslint rule

### LINKED ISSUE #7536: Convert to new type import syntax

snewcomer:
Rewrite TS rule to prefer static type imports

// TODO
- [ ] Report one single violation, ranges to remove is the accumulation ranges
- [ ] New type imports will be a join of the import types.

close #7536
ref #6233 (comment)


snewcomer:
implemented #7608

PR #7599, opened 2021-06-20
@ember-data/model: Simplify @cached transpilation

dcyriller:
and use ember-cached-decorator-polyfill@0.1.4


igorT:
Perf tests fail due to https://github.com/emberjs/data/pull/7639, should pass after that

igorT:
Beta fails on master right now.

PR #7608, opened 2021-06-25
[Chore]: prefer static type-only-import typescript eslint rule

snewcomer:
This PR migrates from `type X` to static imports `import type X from`

Commit [# 3](https://github.com/emberjs/data/pull/7608/commits/cb6c442c60cb1e50ca2359050ef636f13417daae) is the manual changes to our codebase with the current iteration of the script.  Still some fine tuning to be had it seems
 
ref https://github.com/emberjs/data/pull/7598/files
close https://github.com/emberjs/data/issues/7536
ref https://github.com/emberjs/data/issues/6233


runspired (approved):
overall lgtm, need to fix the Snapshot/PrivateSnapshot lint issue manually but otherwise I think we can merge.

packages/unpublished-eslint-rules/src/rules/prefer-type-only-import.js
```diff
@@ -260,7 +260,7 @@ function lintDeclarationForTypeOnlyImports(declaration, usageData, context) {
 }
 
 module.exports = {
-  name: 'prefer-type-only-import',
+  name: 'prefer-static-type-import',
```
igorT, line 263: Did we want to be making this change?

igorT:
Looks good, except for one minor question

packages/unpublished-eslint-rules/src/rules/prefer-static-type-import.js
```diff
@@ -0,0 +1,109 @@
+const collapseRanges = require('./utils/collapse-ranges');
```
igorT, line 1: It's a bit odd that github is not doing syntax highlighting here, any ideas why?

packages/unpublished-eslint-rules/src/rules/prefer-type-only-import.js
snewcomer: Yep!  b/c now it will lint the fact that you only should use `import type` instead of just `type`.

igorT: isn't that what the other new file is for?

igorT: https://github.com/emberjs/data/pull/7608/files#diff-8eb7a08d0916321018b046fbe96f871b36b23290fcf0f521ebea101e8fb58f2e

igorT: Are you sure we don't want to delete this one?

snewcomer: Yes you are right.  Fixed!

packages/unpublished-eslint-rules/src/rules/prefer-static-type-import.js
snewcomer: It is for me at least!  Also checked vscode

PR #7615, opened 2021-06-30
[DOC] Fix transform example not using native class syntax

brkn:

### DIFF
--- a/packages/model/addon/-private/attr.js
+++ b/packages/model/addon/-private/attr.js
@@ -89,19 +89,19 @@ function getDefaultValue(record, options, key) {
   ```app/transforms/text.js
   import Transform from '@ember-data/serializer/transform';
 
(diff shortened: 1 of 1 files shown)


snewcomer (approved):
Thank you!

runspired (approved):

runspired (approved):
We should also document extending EmberObject directly (or using a static create) in a future path. This is also a reminder that we should look into deprecating the Transform base-class as it is essentially a re-export of an extend of EmberObject without any real purpose.

jenweber:
This ticket came up during learning team triage, since it was marked Docs. Can someone with the permissions make the CI re-run? I think the issues with canary that caused failing tests might be resolved now. Thanks!

jenweber:
I asked in dev-ember-data if anyone can re-run these tests.

PR #7634, opened 2021-07-15
[CHORE]: Modernized store-adapter-test

runnerboy22:

### DIFF
--- a/packages/-ember-data/tests/integration/adapter/store-adapter-test.js
+++ b/packages/-ember-data/tests/integration/adapter/store-adapter-test.js
@@ -1009,7 +1009,7 @@ module('integration/adapter/store-adapter - DS.Store and DS.Adapter integration 
     let adapter = store.adapterFor('application');
 
     class Person extends Model {
-      @hasMany('dog', { async: true }) dogs;
+      @hasMany({ async: true }) dogs;
     }
 
     this.owner.register('model:person', Person);
@@ -1073,7 +1073,7 @@ module('integration/adapter/store-adapter - DS.Store and DS.Adapter integration 
     let store = this.owner.lookup('service:store');
     let adapter = store.adapterFor('application');
     class Person extends Model {
-      @hasMany('dog', { async: true }) dogs;
+      @hasMany({ async: true }) dogs;
     }
 
     this.owner.register('model:person', Person);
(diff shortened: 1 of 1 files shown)


packages/-ember-data/tests/integration/adapter/store-adapter-test.js
```diff
@@ -1459,6 +1800,20 @@ module('integration/adapter/store-adapter - DS.Store and DS.Adapter integration
...
+    }
+
+    class Dog extends Model {
+      @attr('string') name;
+    }
+
+    this.owner.register('model:person', Person);
+    this.owner.register('model:dog', Dog);
+    this.owner.register('adapter:application', JSONAPIAdapter.extend());
+    this.owner.register('serializer:application', JSONAPISerializer.extend());
```
snewcomer, line 1815: Did you find that a beforeEach was too overbearing and it was easier to do it in the specific tests that need it?

runnerboy22: Yes, keeping in the spirit of defining individually per test. What do you think of that design choice, has removing beforeEach made tests on the whole too verbose?

snewcomer: When writing tests, it probably is preferred to take common test setup and only repeat yourself once (or a few times if you need to make some modifications for a few tests).  In this example, we might be able to reduce a few hundred lines of code!

runnerboy22: Thanks for the feedback, I am in complete agreement about minimizing redundant code and making the file much smaller! I was basing this change on: https://github.com/emberjs/data/pull/7567#pullrequestreview-673422850

Why was it more preferred to eliminate the beforeEach in that PR?

snewcomer: Great question.  I think it was a bit more nuanced in that example.  In that PR, it was re-declaring `Post` (via `Post = ...`).  Effectively taking over `Post` from something that had greater scope access. In this case, the thing in the beforeEach here is necessary setup for each test. From looking at the tests briefly, we just may want the most broad test setup. So that looks like a Person with 2 attributes and a relationship....As well as a Dog with one attribute.

runnerboy22: I see, great explanation thank you! I'll roll this back a couple of commits and put the beforeEach back in

PR #7639, opened 2021-07-22
Bring yarn.lock up to date

igorT:
Currently running yarn on master leaves a dirty state


igorT:
Performance checks will fail in this PR because of the dirty state of yarn.lock

dcyriller (approved):

PR #7653, opened 2021-08-07
Nightly test: beta and canary fail b/c semver comparison

snewcomer:
ref https://www.npmjs.com/package/semver

Screen Shot 2021-08-07 at 9 09 28 AM

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -287,13 +287,38 @@ jobs:
         run: yarn test
 
   external-partners:
+    needs: [additional-scenarios, basic-tests, floating-dependencies, lint, lts, releases]
+    if: |
+       github.event_name == 'pull_request' && (
+         github.base_ref == 'master' || github.base_ref == 'beta'
+       ) || github.event_name == 'push' && (
+         endsWith(github.ref, '/master') || endsWith(github.ref, '/beta')
+       )
     runs-on: ubuntu-latest
     strategy:
+      fail-fast: false
       matrix:
         partner:
           [
+            ember-data-relationship-tracker,
             ember-m3,
+            ember-observer,
+            ember-resource-metadata,
+            factory-guy,
+            ilios-frontend,
+            model-fragments,
+            storefront,
+            travis-web,
           ]
+        include:
+          - partner: storefront
+            continue-on-error: true
+          - partner: factory-guy
+            continue-on-error: true
+          - partner: travis-web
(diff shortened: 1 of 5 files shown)


.github/workflows/main.yml
```diff
@@ -334,10 +336,3 @@ jobs:
           CI: true
         run: yarn test-external:${{ matrix.partner }}
         continue-on-error: ${{ matrix['continue-on-error'] == true }}
-      - name: upload built assets
-        if: failure() || success()
-        uses: actions/upload-artifact@v2
-        with:
-          name: ${{ matrix.partner }}-dist
-          path: ${{ github.workspace }}/__external-test-cache/${{ matrix.partner }}/dist/assets
-        continue-on-error: ${{ matrix['continue-on-error'] == true }}
```
snewcomer, line 343: We probably don't need this.  We can just run this locally

.github/workflows/main.yml
```diff
@@ -319,9 +319,11 @@ jobs:
             continue-on-error: true
           - partner: model-fragments
             continue-on-error: true
+          - partner: ember-observer
+            continue-on-error: true
```
snewcomer, line 323: This is currently failing.  Ran locally to confirm.  I've seen this error before but not sure exactly what is going on now.

Screen Shot 2021-08-08 at 2 13 43 PM

.github/workflows/main.yml
```diff
@@ -228,11 +228,11 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        scenario: [ember-release, ember-beta, ember-canary]
+        scenario: [ember-release, ember-beta]
```
snewcomer, line 231: canary is timing out.  If canary is failing now, beta will soon as well.  Tests just aren't starting up.

snewcomer: I think we talked about potentially excluding this.  But perhaps I'm misremembering what that was.

igorT: Yeah that makes sense, we probably want to not have `beta` block CI either.

.github/workflows/main.yml
igorT: I think the idea was it makes debugging faster if I understood correctly, no?

snewcomer: Likely.  We do have those scripts `test-external:*` to run the specific failures and debug just as easily as well.

PR #7668, opened 2021-08-20
feat: allow .load()ing sync relationships

machty:
I think it should be possible to call .load() on a sync belongsTo reference,
but if you do, you get an AssertionError:

Assertion Failed: You looked up the 'team' relationship on a 'person' with id 1
but some of the associated records were not loaded. Either make sure they are
all loaded together with the parent record, or specify that the relationship
is async (`belongsTo({ async: true })`)

This commit adds a breaking test to demonstrate this.

Might be worth checking sync hasMany's handling of this too.

Note: for some reason, `.reload()` works (e.g. `model.belongsTo('foo').reload()`) but not `.load()`.

### DIFF
--- a/packages/model/addon/-private/legacy-relationships-support.ts
+++ b/packages/model/addon/-private/legacy-relationships-support.ts
@@ -749,7 +749,7 @@ function anyUnloaded(store: Store, relationship: ManyRelationship) {
   return unloaded || false;
 }
 
-function areAllInverseRecordsLoaded(store: Store, resource: JsonApiRelationship): boolean {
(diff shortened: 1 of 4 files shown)


runspired:
this is a feature request, albeit a relatively minor one which we can probably skip the RFC on. The EmberDataStoreFront pattern mostly used reload here to work around this. The reason this fails is because we take the same codepath for sync relationships on `load` as if you'd simply accessed the relationship off of the record. That codepath is guarded because (as the error message says) you're accessing it prior to having loaded it. In this case though, the intent is clearly to get around that and we should make it nice.

PR #7681, opened 2021-08-28
typofix `fetchRecord` -> `findRecord`

jfdnc:
Description for `coalesceFindRequests` indicates that it will attempt to coalesce `fetchRecord` calls, but the relevant public API concept is `findRecord`. If using the term `fetchRecord` here is referencing the internal [`_fetchRecord`](https://github.com/emberjs/data/blob/9889ebbd4eec162dba17bb00126ee71e49d04bce/packages/store/addon/-private/system/core-store.ts#L1362) it may be more clear to say something like
```
 By default the store will try to coalesce all `findRecord` and async relationship calls within the same runloop
```
because of `get` and `belongsTo` calls being affected, but just changing to `findRecord` here is probably fine. Either way it is a bit confusing to have `fetchRecord` here since it hints at a framework concept that does not have any definition in the public API.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,13 @@
 # Ember Data Changelog
 
+## Release v3.28.0-beta.2 (Jul 30, 2021)
+
+#### :bug: Bug Fix
+- [#7643](https://github.com/emberjs/data/pull/7643) @ember-data/model: Simplify @cached transpilation (#7599) ([@igorT](https://github.com/igorT))
+
+#### Committers: 1
+- Igor Terzic ([@igorT](https://github.com/igorT))
+
 ## Release v3.28.0-beta.1 (May 27, 2021)
 
 #### :rocket: Enhancement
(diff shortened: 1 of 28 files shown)


runspired:
The change seems good but needs to be targeted against the master branch, could you open a new PR with the correct target or re-target this PR?

jfdnc:
> Description for `coalesceFindRequests` indicates that it will attempt to coalesce `fetchRecord` calls, but the relevant public API concept is `findRecord`. If using the term `fetchRecord` here is referencing the internal [`_fetchRecord`](https://github.com/emberjs/data/blob/9889ebbd4eec162dba17bb00126ee71e49d04bce/packages/store/addon/-private/system/core-store.ts#L1362) it may be more clear to say something like
> 
> ```
>  By default the store will try to coalesce all `findRecord` and async relationship calls within the same runloop
> ```
> 
> because of `get` and `belongsTo` calls being affected, but just changing to `findRecord` here is probably fine. Either way it is a bit confusing to have `fetchRecord` here since it hints at a framework concept that does not have any definition in the public API.

Happy to!

PR #7687, opened 2021-08-30
Failing test: findRecord should not cache if includes differ

kpfefferle:
In 3.28, it seems that #7651 made a change to always return an existing inflight `findRecord` for a given model - even if the request itself would have been different (such as by using different `include` options).

This PR adds a test asserting that the `findRecord` caching (asserted by the previous test) should not apply if the `include` arguments differ.


kpfefferle:
cc: @igorT

igorT:
Thanks for the test case, on it

kpfefferle:
Addressed by #7702

PR #7691, opened 2021-08-31
add test + fix for chained async has many

### LINKED ISSUE #7684: 3.28: Awaiting async relationships does not work anymore
### Reproduction

- clone https://github.com/rust-lang/crates.io/pull/3852
- `yarn install`
- `yarn start:live`
- http://localhost:4200/tests/?mirageLogging&filter=visiting%20a%20crate%20page%20from%20the%20front%20page
- observe that `versions = await crate.get('versions');` resolves before the network request has been performed

### Description

Just like in previous Ember Data releases I expected to be able to `await` async relationships to be able to wait until the content has loaded before continuing, but this appears to have broken with v3.28 by resolving before the network request has resolved.

### Versions

Run the following command and paste the output below: `yarn list ember-source && yarn list ember-cli && yarn list --pattern ember-data`.

```cli
❯ yarn list ember-source && yarn list ember-cli && yarn list --pattern ember-data
yarn list v1.22.4
warning Filtering by arguments is deprecated. Please use the pattern option instead.
└─ ember-source@3.28.0
✨  Done in 1.21s.
yarn list v1.22.4
warning Filtering by arguments is deprecated. Please use the pattern option instead.
└─ ember-cli@3.28.0
✨  Done in 0.90s.
yarn list v1.22.4
├─ @ember-data/adapter@3.28.0
├─ @ember-data/canary-features@3.28.0
├─ @ember-data/debug@3.28.0
├─ @ember-data/model@3.28.0
├─ @ember-data/private-build-infra@3.28.0
├─ @ember-data/record-data@3.28.0
├─ @ember-data/serializer@3.28.0
├─ @ember-data/store@3.28.0
└─ ember-data@3.28.0
✨  Done in 0.90s.
```

sly7-7:
cc/ @Turbo87 Could you confirm your scenario is the same as this one ?

resolves #7684 (at least the mentioned test suite passes locally)


packages/-ember-data/tests/integration/relationships/has-many-test.js
```diff
@@ -848,7 +848,7 @@ module('integration/relationships/has_many - Has-Many Relationships', function (
     });
   });
 
-  test('A hasMany relationship can be reloaded if it was fetched via a link', function (assert) {
+  test('A hasMany relationship can be reloaded if it was fetched via a link', async function (assert) {
```
sly7-7, line 851: Sorry for the noise, this one was just a refactor, getting rid of run().

sly7-7:
@Turbo87 At least the "fix" seem to make crates.io's tests happy ;)

Capture d’écran 2021-08-31 à 11 19 50

packages/-ember-data/tests/integration/relationships/has-many-test.js
```diff
@@ -4095,4 +4088,117 @@ module('integration/relationships/has_many - Has-Many Relationships', function (
...
   });
+
+  test('fetch records with chained async has-many, ensure the leafs are retrieved', async function (assert) {
+    assert.expect(4);
+
+    let store = this.owner.lookup('service:store');
+    let adapter = store.adapterFor('application');
+
+    store.modelFor('user').reopen({
+      messages: hasMany('message', { polymorphic: true, async: true }),
```
Turbo87, line 4099: why do we use `polymorphic: true` here? at least in crates.io we don't seem to have any polymorphic relationships 🤔

sly7-7: just because this is how the relation is defined at the top of the file.  (https://github.com/emberjs/data/blob/df969d59ca2eb0912dcccfb7d59b33332e3a9d55/packages/-ember-data/tests/integration/relationships/has-many-test.js#L28)

I wanted to use the 'same' relation, but async.

packages/-ember-data/tests/integration/relationships/has-many-test.js
igorT: No apology necessary, thanks a bunch for the cleanup?

sly7-7:
Closing because #7696 is more precise test

igorT:
@runspired this looks like the correct fix to me, as if you push the same record twice, the deleted branch of code marks the link as not stale so it does not get fetched, even though it had never been fetched. What are your thoughts

sly7-7:
@igorT Thanks again for looking into it. I've added an assertion to complete the tests (after @runspired pointed me that the state of the hasMany's promise looks not good). 
It turns out, that without this "fix" the new assertion pass, but the relationship is not fetched, and with the "fix", the relationship is fetched, but the promise state looks wrong :(

sly7-7:
As I failed to make the test pass, I tried to make more debugging, to see what happen. I think there are subtle bugs on how the hasMany() computed property is cached/invalidated. For example, considering this code at the end of the added test in `rest-adapter-test`:

```
let commentsPromiseArray = post.comments;
let comments = await commentsPromiseArray;
assert.ok(commentsPromiseArray.isFulfilled, 'referenced commentsPromiseArray variable is fulfilled');
assert.equal(post.comments, commentsPromiseArray, 'promise arrays are the same');
assert.ok(commentsPromiseArray.isFulfilled, 'referenced commentsPromiseArray variable is still fulfilled');
assert.ok(post.comments.isFulfilled, 'post.comments promise is fulfilled');
 assert.equal(comments.length, 3, 'The correct records are in the array');
```

and the "fix" (don't set isStale to false), gives this result:

Capture d’écran 2021-09-08 à 17 44 57

See how the PromiseManyArray state is reset to not fulfilled just because we call again the `post.comments` cp. 
I think it's because awaiting the promise invalidate the cp cache for the hasMany. In the end internalModel.getHasMany() is called, which call PromiseManyArray._update() which cause its flag to be reset. (Though, I think the underlying adapter.fetchHasMany won't be called a second time, which is expected)

Turbo87:
@igorT any news on this? I've just hit the same issue in another project too 😅

benedikt:
We've ran into the issue as well and rolled back to 3.27 for now. If there's anything I can do to help with fixing this, please let me know.

sly7-7:
@benedikt Same here about sticking with 3.27.

> If there's anything I can do to help with fixing this, please let me know.

I would say, if you find how to make the tests pass in this PR, it could be a potential fix for the issue. I really don't know what to do :(

packages/-ember-data/tests/integration/adapter/rest-adapter-test.js
```diff
@@ -1283,6 +1283,59 @@ module('integration/adapter/rest_adapter - REST Adapter', function (hooks) {
...
+    ajaxResponse({
+      comments: [
+        { id: 1, name: 'FIRST' },
+        { id: 2, name: 'Rails is unagi' },
+        { id: 3, name: 'What is omakase?' },
+      ],
+    });
+
+    let comments = await post.get('comments');
+    assert.ok(post.comments.isFulfilled, 'comments promise is fulfilled');
```
benedikt, line 1335: It looks like `post.comments` will trigger an [update of the promise proxy](https://github.com/sly7-7/data/blob/d328ef58c414123e96f8d7f9a4dcd49af15e1725/packages/store/addon/-private/system/model/internal-model.ts#L715). That's why this assertion fails. Removing that proxy update fixes this particular test, but obviously breaks a few others. 

I guess the question is what we expect to get from consecutive `post.comments` calls. Are they supposed to return the cached results or do they all trigger a new fetch?

sly7-7: As I said [in a previous comment](https://github.com/emberjs/data/pull/7691#issuecomment-915358747) I don't know how calls of `post.comments` should behave :(. And it looks like that calling `post.comments` won't always trig the update you mentioned (since DS.hasMany() is a computed property).

benedikt: I should have read all the comments before diving into the code 😬 🙈

benedikt:
Looks like the computed property gets invalidated while loading the data for the first time in the [replace related record operation](https://github.com/sly7-7/data/blob/d328ef58c414123e96f8d7f9a4dcd49af15e1725/packages/record-data/addon/-private/graph/operations/replace-related-records.ts#L341) which eventually calls [`notifyPropertyChange`](https://github.com/sly7-7/data/blob/d328ef58c414123e96f8d7f9a4dcd49af15e1725/packages/store/addon/-private/system/model/internal-model.ts#L918
). 

The first call loads the data, which changes the state, which invalidates the property. The second time (probably?) also loads the data, but the state doesn't change (because it's the same data) and therefore doesn't invalidate the property. 

I don't have a solution, but hopefully that explains why the `post.comments` call sometimes returns a new PromiseManyArray and sometimes not.

esbanarango:
👋. We're also blocked by this one. Back to `3.27`. Thank you @sly7-7 and @igorT for putting the time to fix this 🙏🙏.

snewcomer:
@sly7-7 Mind merging latest master, fixing conflicts and squashing down to one commit?  I prob need to take your commit hash and cherry pick it to other versioned branches.

snewcomer:
mvmd I squashed!

snewcomer:
@sly7-7 What is the current status of the test suite for this PR?  I couldn't parse from this thread but it looks like just the basic test run is failing.

sly7-7:
@snewcomer Well, for me it's complicated. I think async relationships are somewhat broken since 3.28 (you can see https://github.com/emberjs/data/issues/7684 and also issues linked to this one). With (this change)[https://github.com/emberjs/data/pull/7691/files#diff-eeb08072f754311eb1c97e83e5a84fa0c986069652670802caf9d3660fd82a49L156], the issue seemed to be fixed, but at the same time it surfaces an other one, where the state of the underlying promise is unstable (eg, calling twice `post.comments.isFulfilled`, gives different results). And for this, I don't know what should be done. (see https://github.com/emberjs/data/pull/7691#discussion_r719270772).

packages/store/addon/-private/system/model/internal-model.ts
```diff
@@ -902,11 +903,18 @@ export default class InternalModel {
 
   notifyHasManyChange(key: string) {
```
sly7-7, line 904: Thank you so much @runspired. I would never have guessed the culprit was here. I don't know how you found that. All look so easy for you 😅

PR #7696, opened 2021-09-03
Isolated repro for #7691

igorT:
A narrow test case for #7691 that shows that pushing a record twice no longer fetches the link


igorT:
@runspired ^^

igorT:
Going to push to @sly7-7's branch

PR #7697, opened 2021-09-03
[BUGFIX] Update @ember/string to the latest version to fix compatibility with ember-source 4.x

kategengler:
The prior version used deprecated import paths that error under ember-source v4.0.

This will need to be backported to beta and released.


igorT:
@kategengler Thanks for the fix. I have the backports out at https://github.com/emberjs/data/pull/7698 and https://github.com/emberjs/data/pull/7699

igorT:
Published in `v3.28.3` and `v4.0.0-beta.2`

PR #7700, opened 2021-09-05
Fix testcase stub of `createModelDataFor`

machty:
`createModelDataFor` was the old name before it was changed
to `createRecordDataFor`

### DIFF
--- a/tests/main/tests/integration/record-data/unloading-record-data-test.js
+++ b/tests/main/tests/integration/record-data/unloading-record-data-test.js
@@ -260,8 +260,8 @@ module('RecordData Compatibility', function (hooks) {
   });
 
   test(`store.unloadRecord on a record with custom RecordData with relationship to a record with default RecordData does not error`, async function (assert) {
-    const originalCreateRecordDataFor = store.createModelDataFor;
-    store.createModelDataFor = function provideCustomRecordData(identifier, storeWrapper) {
+    const originalCreateRecordDataFor = store.createRecordDataFor;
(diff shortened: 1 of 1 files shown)


igorT:
@machty can you please add an assert in the code we are stubbing to ensure this time it is getting called.

machty:
@igorT while addressing your feedback I noticed these tests had an identical setup and what appeared to be a lot of leftover cruft from the original RFC implementation. It's much smaller now, performs the same tests, and the stub assertion you asked for is accomplished just by testing that that the constructed RecordData for a model has the expected type (which is what one of the tests was doing before).

packages/-ember-data/tests/integration/record-data/store-compatibility-test.js
```diff
@@ -0,0 +1,164 @@
...
+  test(`store.unloadRecord on a record with default RecordData with relationship to a record with custom RecordData does not error`, async function (assert) {
+    run(() => chris.unloadRecord());
+  });
+
+  test(`store.unloadRecord on a record with custom RecordData with relationship to a record with default RecordData does not error`, async function (assert) {
+    run(() => shen.unloadRecord());
+  });
+});
+
+module('RecordData - Instantiation', function (hooks) {
```
igorT, line 99: Can we not split the two modules into two files please?

igorT:
@machty great work!

runspired:
@machty if you hapeen to have time to rebase this I think the tests will pass and I'd be happy to merge

runspired:
something is odd here, the more recent commits clearly show test refactoring that the github diff cannot seem to see :/

PR #7393, opened 2020-12-03
[Cleanup]: Remove FULL_LINKS_ON_RELATIONSHIPS

snewcomer:
PR to turn on - https://github.com/emberjs/data/pull/6966

PR #7703, opened 2021-09-11
fix tests

snewcomer:
further cleanup of https://github.com/emberjs/data/pull/7393


packages/-ember-data/tests/integration/store-test.js
```diff
@@ -175,14 +175,12 @@ module('integration/store - destroy', function (hooks) {
...
+    const result = await nextPromise;
 
     // ensure we allow the internal store promises
     // to flush, potentially pushing data into the store
     await settled();
-    assert.ok(true, 'we made it to the end');
-    await requestPromise;
-    assert.ok(false, 'we should never make it here');
+    assert.equal(result.data.type, 'car', 'we made it to the end');
   });
```
snewcomer, line 184: This test is now failing on the main branch.  I rewrote it a little bit but the main issue was `await requestPromise`.  On first glance, I can't tell what is going wrong here.

packages/-ember-data/tests/integration/store-test.js
```diff
@@ -165,7 +164,8 @@ module('integration/store - destroy', function (hooks) {
 
       throw new Error("We shouldn't be pushing data into the store when it is destroyed");
     };
-    let requestPromise = store.findRecord('car', '1');
+
+    store.findRecord('car', '1');
 
     await nextPromise;
```
snewcomer, line 170: I thought I was reading this right but now I'm questioning myself.  I wouldn't think we would advance from this line because we have yet to `resolve` the promise.  I thought it would just hang here.

packages/-ember-data/tests/integration/store-test.js
snewcomer: Since this was failing on 3.20 (last 3.20 release was https://github.com/emberjs/ember.js/tree/v3.20.7) something has "changed" since the last PRs a week ago.  In any case, I distilled the test down a bit.

sandstrom:
@snewcomer Any plans on upping the tiers on your sponsor page? 

runspired has higher-tiers available (https://github.com/sponsors/runspired), and would be happy to up our sponsorship for you as well. May I suggest a $100 tier?

packages/canary-features/addon/default-features.ts
```diff
@@ -173,10 +173,8 @@ export default {
   SAMPLE_FEATURE_FLAG: null,
   RECORD_DATA_ERRORS: true,
   RECORD_DATA_STATE: true,
-  IDENTIFIERS: true,
```
igorT, line 176: I think we might want to keep these, in case someone was reading them and branching code off of it, removing them would break. Do you think there is a downside in not GCing them?

packages/-ember-data/tests/integration/store-test.js
igorT: It gets resolved on line `143` no?

packages/-ember-data/tests/integration/store-test.js
igorT: Ha, interesting, it seems like we are asserting that the promise never finishes, it makes sense it would hang the test. We can rewrite this as `.then` and not hang on it, or just rely on the `store.push`

packages/-ember-data/tests/integration/store-test.js
snewcomer: [Here](https://www.typescriptlang.org/play?#code/BQQwzgngdgxgBAMwK6wC4EsD2VgEo4DeAUHKXADYCmqcUlAHqgFxwhQQDcJZVNdA7gAUATpgC26MJS5lEKGBmyJMmPIW6zSAkeMmU4AXlqV+cHRKnBgwymEzkAbpXwGAfOs2e4MbHaoA6ckwAc2AAchgQcnIAIxAYAGsw3BkvWTpGQzgbPydUzwBfXH9UAAtKHDU3DzSyHyg-SkCQ8LKK5I1am1QkYShCVhYwkDC4AvzNIq5O2W7e-u1RC2lOgqJO3mzKAEcshBU8CZB+EHQ+E3M9CfrG5tCwgBN0B9YHhzYYSmSJjNRDzpu9iaQXuOXsTgeHVkNxoDxAqBAWWOp3OQiWV06yLOW22118QLuwDhCJSRCK-yAA) is a plan TS example.  We haven't resolved yet.  Clearly a gap in my knowledge somewhere 😆

packages/canary-features/addon/default-features.ts
snewcomer: I only thought to commit these for clarity of the codebase.  Certainly if someone was relying on a forked version of e-d, but is it something we expect users to read?  And if they do, then it is expected they could be removed at any point?  As long as we get the tests passing, I'm ok to do whatever you think is best 👍

packages/-ember-data/tests/integration/store-test.js
snewcomer: I missed that store.findRecord would batch multiple if necessary.  As a result, nextPromise is undefined here.

packages/-ember-data/tests/integration/store-test.js
snewcomer: Fixed with `.then`!  Thank you!

PR #7721, opened 2021-10-17
Handle empty query params properly by allowing in the url

### LINKED ISSUE #7720: Store.query does not respect empty arrays
### Description

When providing a parameter with an empty array to the ember-store query function the resulting url completely omits the parameter.

```
Expected:
store.query('user', {ids: []}) => /users?ids=[]
```
```
Unexpected:
store.query('user', {ids: []}) => /users
```

### Versions
```
└─ ember-data@3.4.4
✨  Done in 1.29s.
```

snewcomer:
https://jsonapi.org/format/#fetching-sparse-fieldsets

> An empty value indicates that no fields should be returned.

The JSONAPI has taken a strict stance on what an empty query param should mean.  They stated it should return no fields instead of leaving it up to the implementer.  I think we should follow.

Original PR that strips falsey values: https://github.com/emberjs/data/pull/6077/files.  We need to keep the query param so that an API can know that, for example, no rows are returned if `ids=[]`.

Ref implementation in other areas: https://github.com/jeregrine/jsonapi/pull/257

close #7720

### DIFF
--- a/packages/-ember-data/tests/unit/adapters/rest-adapter/fetch-options-test.js
+++ b/packages/-ember-data/tests/unit/adapters/rest-adapter/fetch-options-test.js
@@ -200,4 +200,27 @@ module('unit/adapters/rest-adapter/fetch-options', function (hooks) {
       "'options.url' is serialized"
     );
   });
+
+  test('fetchOptions serializes empty query params to the url', function (assert) {
+    assert.expect(1);
+
+    const postData = {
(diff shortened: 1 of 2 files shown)


igorT:
Reviewed with @snewcomer, decided to limit this just to json-api adapter

runspired:
Closing as this has gone stale and we're moving away from adapters. The request utils / request-manager do not exhibit this issue.

PR #7722, opened 2021-10-18
Fix deprecated assign usage from @ember/assign

snewcomer:
prep for 4.0!

### DIFF
--- a/packages/store/addon/-private/system/ds-model-store.ts
+++ b/packages/store/addon/-private/system/ds-model-store.ts
@@ -39,7 +39,7 @@ class Store extends CoreStore {
       _internalModel: internalModel,
       container: null,
     };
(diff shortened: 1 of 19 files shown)


packages/store/addon/-private/system/ds-model-store.ts
```diff
@@ -40,7 +39,7 @@ class Store extends CoreStore {
       _internalModel: internalModel,
       container: null,
     };
-    assign(createOptions, createRecordArgs);
+    createOptions = { ...createOptions, ...createRecordArgs };
```
igorT, line 42: Let's use Object.assign here

igorT (approved):
Looks good after the spread to assign change

kategengler:
@igorT Can this be backported to 4.0 beta? Without this, there are deprecations in brand new 4.0 apps using ember-data.

Issue #7724, opened 2021-10-20
Cleanup for 4.0 release


snewcomer:
- [x] Remove feature flag app code that is enabled - https://github.com/emberjs/data/blob/master/packages/canary-features/addon/default-features.ts
- [x] Object.assign deprecation - https://github.com/emberjs/data/pull/7722
- [x] Drop IE11 from CI suite and build
- [x] [record.toJSON](https://deprecations.emberjs.com/ember-data/v3.x/#toc_record-toJSON)

Work in Progress.  If you know of other items to add (there are many more), let me know in discord or comment here!

SergeAstapov:
@snewcomer [`record.toJSON`](https://deprecations.emberjs.com/ember-data/v3.x#toc_record-toJSON) is still in the codebase and deprecation targets `4.0.0` so should be added to the list

PR #7739, opened 2021-11-04
Remove TARGET_IE11 flag

snewcomer:
prep for 4.0!

There are some comments about IE11 in the codebase and features we couldn't use due to lack of APIs.  Will follow up after to remove all comments and apply necessary changes.

This PR ultimately aligns our browser targets with ember.js

- https://github.com/emberjs/ember.js/blob/master/config/targets.js

### DIFF
--- a/packages/adapter/tests/dummy/config/targets.js
+++ b/packages/adapter/tests/dummy/config/targets.js
@@ -2,12 +2,6 @@
 
 const browsers = ['last 1 Chrome versions', 'last 1 Firefox versions', 'last 1 Safari versions'];
 
-const needsIE11 = !!process.env.TARGET_IE11;
-
-if (needsIE11) {
-  browsers.push('ie 11');
-}
-
 module.exports = {
   browsers,
   node: 'current',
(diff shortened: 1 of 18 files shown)


packages/adapter/tests/dummy/config/targets.js
```diff
@@ -2,12 +2,6 @@
 
 const browsers = ['last 1 Chrome versions', 'last 1 Firefox versions', 'last 1 Safari versions'];
```
snewcomer, line 3: Perhaps we should change this to `last 2`?  Thoughts @igorT ?

igorT: We should do whatever Ember does

igorT (approved):

igorT:
@snewcomer looks like lint is failing

packages/adapter/tests/dummy/config/targets.js
snewcomer:
```
  browsers = ['last 2 Chrome versions', 'last 2 Firefox versions', 'Safari 12', 'last 2 Edge versions'];
```

Good idea.  Here it is and updated in this PR.

PR #7741, opened 2021-11-05
Remove IE11 hacks

snewcomer:
prep for 4.0!

### DIFF
--- a/packages/private-build-infra/src/stripped-build-plugins.js
+++ b/packages/private-build-infra/src/stripped-build-plugins.js
@@ -2,7 +2,6 @@
 
 const FilterImports = require.resolve('babel-plugin-filter-imports');
 const StripClassCallCheck = require.resolve('babel6-plugin-strip-class-callcheck');
-const TransformBlockScoping = require.resolve('@babel/plugin-transform-block-scoping');
 
 function isProduction(environment) {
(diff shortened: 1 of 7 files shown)


runspired (approved):

runspired (approved):
These changes are good but the build tooling will have to be updated to not include IE11 as well, that's the reason for most the infra test failures.

Primarily this means updating our actions configs.

Since the IE11 size is no longer relevant we can drop that bit from the asset size report entirely.

packages/private-build-infra/src/stripped-build-plugins.js
```diff
@@ -22,7 +21,7 @@ module.exports = function (environment, app, compatVersion) {
     plugins.push([FilterImports, { imports: filteredImports }]);
   }
 
-  plugins.push([TransformBlockScoping, { throwIfClosureRequired: true }], ...DebugMacros);
+  plugins.push(...DebugMacros);
```
snewcomer, line 24: @runspired This is the right removal, right?

runspired: No, though it's likely also safe to remove this now. There's several GitHub actions/test run scenario infra bits that expect an IE11 compat build.

PR #7743, opened 2021-11-05
Remove najax fallback in FastBoot

snewcomer:
prep for 4.0!


packages/adapter/addon/rest.ts
```diff
@@ -1561,91 +1551,4 @@ function ajaxOptions(options: JQueryRequestInit, adapter: RESTAdapter): JQueryRe
...
-      return shouldUseFetch;
-    },
-
-    set(value) {
-      addSymbol(this, UseFetch, value);
-      return value;
-    },
-  });
-} else {
-  Object.defineProperty(RESTAdapter.prototype, 'useFetch', {
```
runspired, line 1643: I believe this defineProperty call might have been in use to fix a native-vs-classic class syntax interop issue as much as it was here for the najax fallback. May be worth checking with a test that a consumer can configure this via `extends` vs with `=` vs constructor initialization.

runspired (approved):
overall LGTM, left one bit of feedback that is non-blocking

snewcomer:
This was fixed in #7776

PR #7744, opened 2021-11-05
Disable ilios partner test

snewcomer:
Lots of failures from that package.  I don't have the time to go fix them in the short term.  This should get us back to green.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -304,7 +304,7 @@ jobs:
             ember-observer,
             ember-resource-metadata,
             factory-guy,
-            ilios-frontend,
+            # ilios-frontend,
             model-fragments,
             storefront,
             travis-web,


runspired:
cc @jrjohnson

runspired (approved):

jrjohnson:
Thanks, I'm looking into it now. Initially appears to just be a controller that still has an `actions` hash. Maybe a few other similar touchups. I'll get those merged on our end and PR in here to re-activate this test.

jrjohnson:
Dug deeper into this today - the core issue seems to be that if I have a route class that doesn't inject the store service it will fail when the default `model()` hook is run. So this:
```javascript
import Route from '@ember/routing/route';
import { inject as service } from '@ember/service';

export default class SomeRouter extends Route {
  @service session;

  beforeModel(transition) {
    this.session.requireAuthentication(transition, 'login');
  }
}
```

Is now invalid. But injecting the store like;
```javascript
import Route from '@ember/routing/route';
import { inject as service } from '@ember/service';

export default class SomeRouter extends Route {
  @service session;
  @service store;

  beforeModel(transition) {
    this.session.requireAuthentication(transition, 'login');
  }
}
```
works. I'd guess this is the result of removing the implicit injection in @snewcomer's #7667. This makes sense, but I don't think this is precisely the intended behavior, but I'm not sure if this is an issue here or should be addressed in Ember itself or in docs or blueprints.

snewcomer:
@jrjohnson Can you send me the route this is happening on in your codebase

jrjohnson:
This route: https://github.com/ilios/frontend/blob/master/app/routes/curriculum-inventory-reports.js
Declared here: https://github.com/ilios/frontend/blob/master/app/router.js#L101

snewcomer:
Thanks!  

> This makes sense, but I don't think this is precisely the intended behavior

So explicit injection is expected based on the RFC below.

https://github.com/emberjs/rfcs/pull/680

runspired:
> Thanks!  
> 
> 
> 
> > This makes sense, but I don't think this is precisely the intended behavior
> 
> 
> 
> So explicit injection is expected based on the RFC below.
> 
> 
> 
> https://github.com/emberjs/rfcs/pull/680

I think we should raise this as an issue with that RFC, because defaultModel behavior was still not deprecated (we lost the battle that it needed to be deprecated before global injections sadly). It was one thing to dismiss the defaultModel concern on the premise that folks without a model hook might encounter an error if they were relying on it with an easy enough fix, it's another for it to turn out that every route requires the explicit injection even when no store / default model / model hook behavior should be present.

jrjohnson:
> So explicit injection is expected based on the RFC below.

Yup, I'm all for this! What I didn't expect was the need to inject the store when I wasn't calling `this.store` anywhere in my own code. Obviously I can see that `model()` would need the store I just feel weird about injecting it for magic happening elsewhere. I'm going to run some tests, but I'm also curious what happens if I don't have a backing route class at all and so nowhere to inject the store.

> NVM: I see when I look at the full route the store is expected to be there. I've been working on a codemod for explicit injections, should help folks get past this with fewer hiccups.

Is that in my code? As far as I could see we're not relying on the store anywhere in the route chain up to application, but maybe I missed it? if that's the case then this is a clear issue in our app and easy enough to fix. Otherwise I may try and see if the store can be looked up from the container whenever a default hook expects it to exist.

runspired:
@jrjohnson you linked https://github.com/ilios/frontend/blob/master/app/routes/curriculum-inventory-reports.js

Which contains

```js
async model() {
    return this.store.findAll('school');
  }
 ```

jrjohnson:
Ah hell. Sorry. Extra s. I was aiming for https://github.com/ilios/frontend/blob/master/app/routes/curriculum-inventory-report.js

snewcomer:
So to circle back to the actual failures, this was the errors in our test suite.

Screen Shot 2021-11-10 at 11 20 09 AM

https://github.com/emberjs/data/runs/4113308601?check_suite_focus=true

jrjohnson:
That error was masking the real issue because it triggered our error component. It's fixed now so the real issue, that there is no store on that route, is exposed.

runspired:
@jrjohnson that route *should* have store, it's using it you just aren't aware of it. Notice that the route definition has a url-param, ember's (not ember-data's) default model-hook behavior is to call `store.find` with the type and id being derived from the name and value of that param. Since there is a url-param and no model-hook, you're hitting that behavior.

jrjohnson:
@runspired I agree! However it is super weird to inject the store into code where there is no explicit `this.store`. I'll do it, but I think this is probably a bigger issue that our app is just the first to hit. The implicit `model()` on a route is a basic feature of Ember that probably exists a lot of places. I'll have more OSS time this weekend to check it, but I wonder what would happen if I had *no* backing route class at all and just relied on the default implicit route `model` hook. If that fails I'm not sure where the store would even be injected. 

I think the solution is probably going to be looking up the store in the default route instead of relying on `this.store` to be there. Something like `owner.lookup('service:store')`.

igorT:
@jrjohnson were you getting a deprecation before for this behavior?

jrjohnson:
@igorT not that I can see, our latest run at https://github.com/ilios/frontend/runs/4148647180 doesn't mention anything that I can find about store or injection.

snewcomer:
@jrjohnson Could you take a look at this test run when you get a chance?

https://github.com/emberjs/data/runs/4168965070?check_suite_focus=true

jrjohnson:
Yes. `{"type":"error","text":"Error while processing route: curriculumInventoryReport.index Assertion Failed: CurriculumInventoryReport has no method `find`. ` is what I'm seeing locally as well. It can be resolved by adding `@service store` to the route.

jrjohnson:
I've tracked this down to https://github.com/emberjs/ember.js/blob/master/packages/%40ember/-internals/routing/lib/system/route.ts#L1400 where the store is expected to exist on the route and also confirmed that this happens when there is no model hook and also when there is no backing class at all. It doesn't trigger any deprecations in a new ember app with these feature and fails in this simple case as well. 

I created a [sample Ember CLI 3.28.4 app](https://github.com/jrjohnson/reproduction-data-no-injection) and added two routes to it.  Working with with ED 3.28, but fails with 4.0 beta 4.

snewcomer:
@jrjohnson How does this look? 

https://github.com/emberjs/ember.js/pull/19839

jrjohnson:
@snewcomer it throws the same error as before. It looks like a good incidental change, but I'm not clear on how it solves this issue. I don't think the core problem is with ED at all, the issue is that implicit models didn't get deprecated, but now will suddenly stop working without the store injection.

snewcomer:
@jrjohnson There is one more step I need to take care of. Since we did deprecate implicit injection, I was hoping that the deprecation would surface on your test suite.  But if they aren't, then I missed something.  

1. If there is a deprecation, then we can remove the `assert` since the `protected set store(value: any)` isn't called anymore and Model class instances don't have a `find` method.
2. If no deprecation, then I probably need to for the store on the route somehow in ember-data (not using `inject`)

jrjohnson:
Ah, ok. No I patched in your router code into the sample app I built and still don't see a deprecation, just the assertion failure.

PR #7749, opened 2021-11-09
Remove TARGET_IE11 flag

snewcomer:
ref https://github.com/emberjs/data/pull/7739

While not necessary, further PRs to beta don't need to run on IE11

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -147,13 +147,9 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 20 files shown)


.github/workflows/main.yml
```diff
@@ -147,13 +147,9 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        launcher: [Chrome, Firefox, IE]
```
runspired, line 150: we should still test IE, just not IE11. I believe Edge comes installed on these VMs.

Issue #7529, opened 2021-05-20
Re-activate partner tests


runspired:
Potential new external partners
- https://github.com/hashicorp/vault
- https://github.com/Esri/hub.js
- https://github.com/discourse/docker_manager
- active-model-adapter

runspired:
It's unlikely we'll gather much value for a long time from these given the long-tail of upgrades most of them have failed to keep up with, and considering many of them are concepts that will be baked into EmberData during the 5.x series we're unlikely to gain anything by catching them up.

PR #7752, opened 2021-11-10
Bring back ilios partner test

snewcomer:
ref #7744

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -318,7 +318,7 @@ jobs:
             ember-observer,
             # ember-resource-metadata,
             factory-guy,
-            # ilios-frontend,
+            ilios-frontend,
             model-fragments,
             storefront,
             travis-web,


jrjohnson:
Ilios Frontend should, once again, be testable here. We're now on 4.1. Turns out there were a few bonus breaking changes we discovered around `hasMany('foo').ids()` which became eventually consistent but no longer consistent on first access. I'm not sure if this was intended or a side effect of that data being tracked (which we love BTW!!). We also ended up pushing a bunch of models into octane style and that sorted out some problems we had with CPs that were returning promises.

jrjohnson:
Well this is fun... ugh. We went to a node 16.x minimum to allow us to do overrides which only seem to work in NPM v8. This is necessary to get around an embroider incompatibility between v0.3somethign and 0.50something. I don't know how to handle that here other than pushing this into it's own job and setting the minimum node version.

jrjohnson:
Ok we were able to resolve our issues with embroiderer and now support node 14.x again. This is probably ready for a re-try.

runspired:
closing in favor of #7529

PR #7760, opened 2021-11-16
chore: cancel older CI jobs

runspired:
cancels jobs on PRs that aren't for the current head commit.

### DIFF
--- a/.github/workflows/enforce-pr-labels-lts.yml
+++ b/.github/workflows/enforce-pr-labels-lts.yml
@@ -4,7 +4,12 @@ on:
   pull_request:
     types: [labeled, unlabeled, opened, reopened, edited, synchronize]
     branches:
-      - lts-3-24
+      - lts-3-28
+
+concurrency:
+  group: pr-labels-lts-${{ github.head_ref }}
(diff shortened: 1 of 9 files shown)


.github/workflows/main.yml
```diff
@@ -12,6 +12,10 @@ on:
     tags:
       - '*'
 
+concurrency:
+  group: ci-${{ github.head_ref }}
```
runspired, line 16: we may need to adjust `head_ref` to `ref` for pushes to branches (Vs PRs)

igorT: @runspired I'll merge but remove this for now, until you verify which one we need

.github/workflows/enforce-pr-labels-lts.yml
```diff
@@ -4,7 +4,12 @@ on:
   pull_request:
     types: [labeled, unlabeled, opened, reopened, edited, synchronize]
     branches:
-      - lts-3-24
+      - lts-3-28
```
igorT, line 7: We would want to keep the old lts branch no?

PR #7761, opened 2021-11-17
Cherry pick: chore(deps): bump broccoli-rollup from 4.1.1 to 5.0.0

snewcomer:
Cherry pick https://github.com/emberjs/data/pull/7559 to fix https://github.com/babel/ember-cli-babel/issues/419#issuecomment-971787156

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -31,7 +31,7 @@ jobs:
         run: yarn problems
 
   test-infra:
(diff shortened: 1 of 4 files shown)


igorT:
This seems to be failing due to non lockfile tests catching that newest versions of `cli-table` have dropped node v10 support. I don't think we introduced this/should be present in ember-cli by itself, so still gonna merge @rwjblue

PR #7765, opened 2021-11-20
Place store on Ember.Route to maintain implicit record loading

snewcomer:
ref https://github.com/emberjs/rfcs/pull/774

Fix #7744

### DIFF
--- a/packages/-ember-data/addon/setup-container.js
+++ b/packages/-ember-data/addon/setup-container.js
@@ -51,21 +51,6 @@ function initializeStore(application) {
 }
 
 // Implicit injection was removed. This is a replacement for Ember Route implicit store
-// https://github.com/emberjs/rfcs/pull/774
-// Object.defineProperty(Route.prototype, 'store', {
-//   get() {
-//     if (this[EMBER_DATA_STORE]) {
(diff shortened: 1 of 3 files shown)


packages/-ember-data/addon/setup-container.js
```diff
@@ -46,6 +51,22 @@ function initializeStore(application) {
   }
 }
 
+// Implicit injection was removed. This is a replacement for Ember Route implicit store
+// https://github.com/emberjs/rfcs/pull/774
+Object.defineProperty(Route, 'store', {
```
mixonic, line 56: This would define a property on the `Route` class itself, right? Why not use `reopen`?

igorT:
Once we fix the implicit injection deprecation for 3.28 do we still need to do this?

packages/-ember-data/addon/setup-container.js
snewcomer: You were right.  I has thought defineProperty and reopen would be equivalent but turns out I was wrong.

snewcomer: I'm guessing this has something to do with EmberObject.  But overall I was trying the other way to avoid the `reopen` deprecation

PR #7768, opened 2021-11-21
Add lts 3.28 testing scenario

snewcomer:


.github/workflows/main.yml
```diff
@@ -237,6 +237,8 @@ jobs:
       matrix:
         release: [ember-release, ember-beta]
         include:
+          - release: ember-release
+            continue-on-error: true
```
snewcomer, line 241: temporary as we wait catch up, merge in deprecation removals and then deal with array observer changes made in Ember 4.0.  Do we want to do this?

PR #7776, opened 2021-11-24
Release 4.0

igorT:


ef4:
Please see #7785 which targets this branch. I'm hoping we can get that change in.

snewcomer:
@ef4 Can we cherry pick to this branch?

ef4:
I'll update my PR now that you fixed the test failures here, just to make sure there's no need failures I introduced. Then once it's ready, we can just merge it (it targets this PR).

ef4:
@snewcomer we can just hit merge on that PR whenever and it will end up on this branch. I would do it, but I wanted to get an OK from somebody else first and not just ram it through with no review.

snewcomer:
Will merge and release today

PR #7777, opened 2021-11-24
Sn/inject store route release

snewcomer:
ref https://github.com/emberjs/rfcs/pull/774

Fix #7744


igorT:
Once we fix the implicit injection deprecation for 3.28 do we still need to do this?

snewcomer:
Bringing our conversation here:

In 4.0, if you are relying in implicit record loading and have ember-data installed and generate a new route (that triggers a `store.find` b/c it can parse the [params](https://github.com/emberjs/ember.js/blob/b88bd244cafef582c94b1c3faa8742070c83d697/packages/%40ember/-internals/routing/lib/system/route.ts#L1383)), it will break vs. 3.28.   We technically deprecated implicit store injection in 3.x and will be improved in https://github.com/emberjs/ember.js/pull/19854.  

Do we necessarily want to put the `store` on the Route still?  Are we ok breaking "new" routes?  Or is that bad?

snewcomer:
Not reopening b/c only affected items are blueprints.

PR #7780, opened 2021-11-29
extended lid reflection tests

helgablazhkun:
added a test to demonstrate that child attributes are not updated
correctly by .save() when sideposted record has different
attribute than in the response payload with lid provided

### DIFF
--- a/packages/-ember-data/tests/integration/identifiers/lid-reflection-test.ts
+++ b/packages/-ember-data/tests/integration/identifiers/lid-reflection-test.ts
@@ -293,4 +293,90 @@ module('Integration | Identifiers - lid reflection', function (hooks) {
     assert.deepEqual(cake.belongsTo('topping').id(), '2');
     assert.strictEqual(cake.topping.name, 'Cheese');
   });
+
+  test('child attributes are not updated correctly by .save() when sideposted record has different attribute than in the response payload with lid provided', async function (assert) {
+        class Topping extends Model {
+      @attr name;
+    }
+
+    class Cake extends Model {
+      @attr name;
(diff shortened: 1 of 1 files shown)


packages/-ember-data/tests/integration/identifiers/lid-reflection-test.ts
```diff
@@ -293,4 +293,90 @@ module('Integration | Identifiers - lid reflection', function (hooks) {
...
+          },
+          included: [
+            {
+              type: 'topping',
+              id: '2',
+              lid,
+              attributes: {
+                name: 'Cheese',
+              },
+              relationships: {
```
runspired, line 345: if this relationships has is removed (or is made to include the lid of the cake) does the problem persist?

runspired:
After reviewing, this is expected behavior and not a bug. However, expected does not necessarily mean desired and I'd recommend trying to start a discussion around what the best update semantics here should be.

The reason the record is still in the dirty-state after saving via sidepost and why the local state is left unchanged is because when you sidepost the record itself is never transitioned into a saving state. Due to this, nothing has tracked what attributes and values were given to the request. For all we know these attributes and values should be considered stale compared to what we have locally still. In ember-data today, local values are given precedent over the remote values. You will see this same effect when reloading a record that has dirty attributes / receiving an update in the background for a record that has dirty attributes.

In your application, if you know that you are sideposting and that the responses contain altered values you can use `rollbackAttributes` on the record to move it to a clean state after the save completes. Alternatively, in your adapter you can mark the sideposted records as saving, and resolve their save promises once the primary request has completed.

PR #7782, opened 2021-11-29
Failing test for id of sideposted record

andreyfel:
If the newly created record `id` is consumed before save(), the `id`
won't get updated after the record is sideposted through belongsTo
or hasMany relationship.


runspired:
closing in favor of #7800

PR #7785, opened 2021-12-10
move ember-auto-import from devDeps to deps

ef4:
This is targeting the release-4.0 branch.

The idea here is to be consistent with what we did in ember-source -- within the 4.0 breaking change, we added ember-auto-import 2.0, which imposes the requirement that an app must have ember-auto-import >= 2 in order to use ember-source 4.0.

This PR does the same thing for the ember-data packages.

ember-auto-import 2 serves as a polyfill for embroider that allows addons to move to the new streamlined v2 format. By making it a requirement here at the breaking change, we're free to convert all the ember-data packages to v2 format within the 4.0 series without other breaking changes.


ef4:
As far as I can tell, all the test failures here are also on the release-4.0 branch. As that PR gets updated to green I can keep updating this one to confirm it doesn't introduce any new failures.

snewcomer (approved):
Thank you!

PR #7787, opened 2021-12-11
[bugfix]: fix for chained async has many

snewcomer:
ref https://github.com/emberjs/data/pull/7691


runspired:
closing since the primary fix wasn't completed yet

PR #7796, opened 2021-12-13
feat: autotracking for reference id access

### LINKED ISSUE #7102: Reference API is currently incompatible with Tracked

runspired:
resolves #7102

### DIFF
--- a/packages/-ember-data/tests/integration/references/autotracking-test.js
+++ b/packages/-ember-data/tests/integration/references/autotracking-test.js
@@ -1,0 +1,151 @@
+import EmberObject from '@ember/object';
+import { getRootElement, render } from '@ember/test-helpers';
+import settled from '@ember/test-helpers/settled';
+
+import hbs from 'htmlbars-inline-precompile';
+import { module, test } from 'qunit';
+
+import { setupRenderingTest } from 'ember-qunit';
+
+import Model, { attr, belongsTo, hasMany } from '@ember-data/model';
+import { recordIdentifierFor } from '@ember-data/store';
+
+module('integration/references/autotracking', function (hooks) {
+  setupRenderingTest(hooks);
+
+  class User extends Model {
+    @attr name;
+    @belongsTo('user', { inverse: null, async: false })
+    bestFriend;
+    @hasMany('user', { inverse: null, async: false })
+    friends;
+  }
+
+  let store, user;
+  hooks.beforeEach(function () {
+    const { owner } = this;
+    owner.register('model:user', User);
+    store = owner.lookup('service:store');
+
+    owner.register(
+      'adapter:user',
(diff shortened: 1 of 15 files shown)


packages/store/addon/-private/system/references/belongs-to.ts
```diff
@@ -22,17 +30,76 @@ import Reference from './reference';
...
-  constructor(store, parentIMOrIdentifier, belongsToRelationship, key) {
-    super(store, parentIMOrIdentifier);
+  declare key: string;
+  declare belongsToRelationship: BelongsToRelationship;
+  declare type: string;
+  declare parent: RecordReference;
+  declare parentIdentifier: StableRecordIdentifier;
+
+  #token: Object;
+  #relatedToken: Object | null = null;
```
snewcomer, line 40: Can we put an inline comment on what these are used for?

packages/-ember-data/tests/integration/references/autotracking-test.js
```diff
@@ -0,0 +1,151 @@
...
+    await render(hbs`id: {{if this.context.id this.context.id 'null'}}`);
+
+    assert.strictEqual(getRootElement().textContent, 'id: null', 'the id is null');
+    assert.strictEqual(testContext.id, null, 'the id is correct initially');
+    await dan.save();
+    await settled();
+    assert.strictEqual(getRootElement().textContent, 'id: 6', 'the id updates when the record id updates');
+    assert.strictEqual(testContext.id, '6', 'the id is correct when the record is saved');
+  });
+});
```
snewcomer, line 151: 👍

packages/store/addon/-private/system/references/belongs-to.ts
```diff
@@ -22,17 +30,76 @@ import Reference from './reference';
...
+  get _relatedIdentifier(): StableRecordIdentifier | null {
+    this._ref; // consume the tracked prop
+    if (this.#relatedToken) {
+      unsubscribe(this.#relatedToken);
+    }
+
+    let resource = this._resource();
+    if (resource && resource.data) {
+      const identifier = this.store.identifierCache.getOrCreateRecordIdentifier(resource.data);
+      this.#relatedToken = this.store._notificationManager.subscribe(
```
snewcomer, line 88: The reference API is for `CUSTOM_MODEL_CLASS` true and false right?  Do subscription need to be protected if `CUSTOM_MODEL_CLASS` is turned off?

runspired: that would be the test failure 🤦🏻

PR #7805, opened 2021-12-15
Backport Train for LTS

runspired:
* deactivate broken partner tests

* feat: autotracking for reference id access (#7796)

* feat: autotracking for reference id access

* ensure references are torn down

* fix build

* add dep

* add to deps

* fix invalid json:api support and add valid json:api support

* autotracking tests and cleanup

* fix test failure, add comment

* skip tests when feature not available

* update test and fix lid reflection (#7800)

* update test and fix lid reflection

* remove debugger

* fix ff off branch

* add test and fix push of duplicate identifiers to a relationship (#7801)

* add test + fix for chained async has many (#7691)

* [bugfix]: fix for chained async has many

* add fix and update tests

* remove console.logs

* make work with flags off

* fix test for lts

Co-authored-by: Chris Thoburn <runspired@users.noreply.github.com>

* Fix: assign unknown properties in init after initialization is finished to ensure proper setup timing (#7771)

* Add failing test case which illustrates the createRecord bug

createRecord crashes when a setter which sets an attribute is involved
in the createRecord.

* update test location and add fix

Co-authored-by: Chris Thoburn <runspired@users.noreply.github.com>

* fix: A(PromiseManyArray) should have no-effect (#7802)

Co-authored-by: Sylvain Mina <sylvain.mina@gmail.com>
Co-authored-by: Andrey Fel <andrey.fel@retailnext.net>

PR #7809, opened 2021-12-16
add test + fix for chained async has many (#7691)

snewcomer:
* [bugfix]: fix for chained async has many

* add fix and update tests

* remove console.logs

* make work with flags off

* fix test for lts

Co-authored-by: Chris Thoburn <runspired@users.noreply.github.com>


sly7-7:
@snewcomer isn't already backported by https://github.com/emberjs/data/pull/7805 ?

PR #7825, opened 2021-12-24
[BUG] strictNullChecks TS -> JS

AmilKey:
JS does not know the concept of the Non-null assertion operator since this is a TS feature

TS code
```
  then(s, f) {
    return this.promise!.then(s, f);
  }
  ```
  will compile to 
  
  ```
    then(s, f) {
      return this.promise.then(s, f);
    }
  ```
  
  In this case I get error ` Cannot read properties of null (reading 'then')` when I use `await` helper and hasMany in computed property
  
  this bug also affected 3.28 version
  @snewcomer

### DIFF
--- a/packages/-ember-data/tests/integration/identifiers/configuration-test.ts
+++ b/packages/-ember-data/tests/integration/identifiers/configuration-test.ts
@@ -180,7 +180,7 @@ module('Integration | Identifiers - configuration', function (hooks) {
     assert.ok(updateMethodCalls === 0, 'Precond: We have not updated the identifier yet');
     updateCallback = (updatedIdentifier, resource) => {
       assert.strictEqual(identifier, updatedIdentifier, 'We updated the expected identifier');
-      assert.strictEqual(resource.attributes!.firstName, 'James', 'We received the expected resource to update with');
+      assert.strictEqual(resource.attributes?.firstName, 'James', 'We received the expected resource to update with');
     };
 
     await record.save();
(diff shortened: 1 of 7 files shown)


runspired:
It might help if you explain how you triggered an issue with this, as this is 100% expected behavior. We use `!` only when the value MUST be present. If you've hit a case where it is not, let us know. Converting to conditional path execution is not equivalent.

AmilKey:
@runspired 
please have a look at https://github.com/AmilKey/null-check
also try to change ember-data to 3.27 version, you will see different behavior

runspired:
Assuming line 8 in the controller is the issue (https://github.com/AmilKey/null-check/blob/master/app/controllers/application.js) the issue is that this has never been a valid pattern. If it previously worked it's likely because 3.28 fixed several bugs around unloading that probably resulted in this record more fully tearing down where previously it may have been accidentally retained.

The reason the 'then' explodes is the proxy no longer has anything to proxy to once destroyed. We can probably give you a better error message here though to let you know that you've attempted to access the property on a destroyed record.

AmilKey:
> Assuming line 8 in the controller is the issue (https://github.com/AmilKey/null-check/blob/master/app/controllers/application.js) the issue is that this has never been a valid pattern. If it previously worked it's likely because 3.28 fixed several bugs around unloading that probably resulted in this record more fully tearing down where previously it may have been accidentally retained.
> 
> The reason the 'then' explodes is the proxy no longer has anything to proxy to once destroyed. We can probably give you a better error message here though to let you know that you've attempted to access the property on a destroyed record.

will be nice to see better error message, because I could not completely repeat the error from my application, I get an error in the computed property in my app, now I understand that this is happening because the model is destroyed, thanks

AmilKey:
also in my example there is another bug, I can still see the label 'full' into template after model was destroyed in 3.28 version

AmilKey:
@runspired @snewcomer 
I completely reproduced the bug https://github.com/AmilKey/null-check
an error occurs when a loop is used in a template

Issue #7824, opened 2021-12-22
Queries generated by adapter are putting "+" instead of "%20"


alechirsch:
### Description
We are upgrading from ember data 3.20 to 3.28 and came across a breaking change in the way the ember data generates its urls. There seems to have been a change where ember data now uses `fetch` to make the queries, where as before you were using ajax.

We have been able to get around this issue temporarily by setting `useFetch` on the adapter to `false`, however we would like to use fetch in the future and this issue poses a problem.

When ajax generates the url for the request, it encodes spaces as `%20`, which is what we expect. However, with the change to fetch, ember data needed to generate its own query parameters [](https://github.com/emberjs/data/blob/v4.0.2/packages/adapter/addon/-private/utils/serialize-query-params.ts#L44).
I saw that you are explicity converting `%20` to `+`. This is not backwards compatible with the way ajax encodes the urls.
```
  return buildParams('', queryParamsObject).join('&').replace(/%20/g, '+');
```

We are aware that `+` indicates a space, but we prefer to use the encoding and this is a breaking change.

snewcomer:
Looks like according to the spec, `+` is only allowed if we are posting form data.

```
headers: {
  'Content-Type': 'application/x-www-form-urlencoded',
},
```

https://www.w3.org/TR/html401/interact/forms.html#h-17.13.4.1

Also, this is what `encodeURI` does.  I suspect when the code was implemented/partially copied from Ajax, this was just an oversight.

So overall, this seems like an bug.  But just jumping in and changing this could in theory break for some folks....

PR #7834, opened 2022-01-02
Fix bug with encoding space characters when serializing query params

snewcomer:
close https://github.com/emberjs/data/issues/7824

### DIFF
--- a/packages/adapter/addon/-private/utils/serialize-query-params.ts
+++ b/packages/adapter/addon/-private/utils/serialize-query-params.ts
@@ -41,7 +41,7 @@ export function serializeQueryParams(queryParamsObject: object | string): string
     return s;
   }
 
-  return buildParams('', queryParamsObject).join('&').replace(/%20/g, '+');
+  return buildParams('', queryParamsObject).join('&');
 }
 
 /*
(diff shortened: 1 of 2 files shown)


packages/adapter/addon/-private/utils/serialize-query-params.ts
```diff
@@ -41,7 +41,7 @@ export function serializeQueryParams(queryParamsObject: object | string): string
     return s;
   }
 
-  return buildParams('', queryParamsObject).join('&').replace(/%20/g, '+');
```
snewcomer, line 44: re: https://github.com/emberjs/data/issues/7824#issuecomment-1003642900

@alechirsch I noticed this is only invoked with a GET or HEAD request.  So according to the [spec](https://www.w3.org/TR/html401/interact/forms.html#h-17.13.4.1), the `+` sign should only be encoded with a POST `application/x-www-form-urlencoded` request.  Which would not be this case.  

This is in fact a bug and we should be slightly worried that this might cause issues for users though.

https://github.com/emberjs/data/blob/e076e0ae71ae6426ca53ad3c5501a0af7ceca883/packages/adapter/addon/rest.ts#L1499

PR #7848, opened 2022-01-19
Don't use array.push(...spread) with unknown input

DingoEatingFuzz:
Using `array.push(...spread)` is quite nice and convenient, but unfortunately providing too many arguments to a function call can cause a stack overflow, which happened to us.

[Minimum reproduction Twiddle](https://ember-twiddle.com/a8b1b503f15f0c20fde2afe1be5d5bcb?openFiles=routes.application%5C.js%2C)

Also unfortunately, using `Array#concat` changes the API, since it returns a new array rather than modifying the original. Also also unfortunately, simply iterating over an array and pushing one-by-one is significantly slower than using spread or concat.

[Performance tests JSBin](https://jsbin.com/suropipiyi/edit?js,console)

This implementation is a hybrid approach of looping and spreading that remains stack safe.

If we can guarantee that using concat is safe, that'd be an even better implementation since it's both part of the standard lib and also faster.

### DIFF
--- a/packages/-ember-data/tests/integration/serializers/rest-serializer-test.js
+++ b/packages/-ember-data/tests/integration/serializers/rest-serializer-test.js
@@ -422,6 +422,37 @@ module('integration/serializer/rest - RESTSerializer', function (hooks) {
     assert.strictEqual(superVillainNormalizeCount, 1, 'superVillain is normalized once');
   });
 
+  test('normalizeResponse can handle large included arrays', function (assert) {
+    this.owner.register('serializer:super-villain', DS.RESTSerializer);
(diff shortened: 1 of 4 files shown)


runspired:
Using concat here should be safe

snewcomer:
@DingoEatingFuzz Is this something you would like to push across the line with `concat`? Happy to help if you don't have time as well.

DingoEatingFuzz:
Hey @snewcomer, I'm working on updating the branch right now. Should be done today 🙂

snewcomer (approved):
I think this is ok!  tests in the console seem to verify!

```
const source = new Array(130000);
const input = [];
input(...source);
```

PR #7849, opened 2022-01-20
Remove some usage of runtime require

SergeAstapov:
Came across `embroider-optimized` build issues in https://github.com/poteto/ember-changeset/pull/630 and it appears we need a fix similar to https://github.com/miragejs/ember-cli-mirage/pull/2297.

Looks like TS is not happy as `importSync` return type is `unknown` whereas `require()` was successfully resolved by TS.

@ef4 Would you maybe have some advise here how to make TS happy?

### DIFF
--- a/packages/store/addon/-private/system/model/internal-model.ts
+++ b/packages/store/addon/-private/system/model/internal-model.ts
@@ -5,6 +5,7 @@ import { get } from '@ember/object';
 import { _backburner as emberBackburner, cancel, run } from '@ember/runloop';
 import { DEBUG } from '@glimmer/env';
 
+import { importSync } from '@embroider/macros';
 import RSVP, { Promise } from 'rsvp';
 
 import { HAS_MODEL_PACKAGE, HAS_RECORD_DATA_PACKAGE } from '@ember-data/private-build-infra';
@@ -51,7 +52,7 @@ let _getModelPackage: () => boolean;
 if (HAS_MODEL_PACKAGE) {
   _getModelPackage = function () {
     if (!_found) {
-      let modelPackage = require('@ember-data/model/-private');
+      let modelPackage = importSync('@ember-data/model/-private') as typeof import('@ember-data/model/-private');
       ({ ManyArray, PromiseBelongsTo, PromiseManyArray: _PromiseManyArray } = modelPackage);
       if (ManyArray && PromiseBelongsTo && _PromiseManyArray) {
         _found = true;
@@ -490,7 +491,9 @@ export default class InternalModel {
     if (HAS_RECORD_DATA_PACKAGE) {
       let manyArray = this._manyArrayCache[key];
       if (!definition) {
-        const graphFor = require('@ember-data/record-data/-private').graphFor;
+        const graphFor = (
+          importSync('@ember-data/record-data/-private') as typeof import('@ember-data/record-data/-private')
+        ).graphFor;
         definition = graphFor(this.store).get(this.identifier, key).definition as UpgradedMeta;
       }
 
@@ -511,7 +514,7 @@ export default class InternalModel {
 
       return manyArray;
     }
-    assert(`hasMany only works with the @ember-data/record-data package`, HAS_RECORD_DATA_PACKAGE);
+    assert('hasMany only works with the @ember-data/record-data package');
   }
 
   fetchAsyncHasMany(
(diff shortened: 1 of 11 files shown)


ef4:
You can satisfy typescript like:

```ts
importSync('@ember-data/serializer/-private') as typeof import('@ember-data/serializer/-private')
```

You will probably want to change some of the conditions like:

```js
if (DEBUG && HAS_EMBER_DATA_PACKAGE && HAS_ADAPTER_PACKAGE) {
```

to something like:

```js

import { isDevelopingApp, dependencySatisfies, macroCondition, importSync } from '@embroider/macros';

if (macroCondition(isDevelopingApp() && dependencySatisfies('ember-data', '*') && dependencySatisfies('@ember-data/adapter')) {
  importSync('something');
}
```

so that the importSyncs will get compiled away in builds that don't need them. 

Alternatively, if it's possible to refactor any of this code into places where asynchrony can be absorbed, you can use `import()` instead of `importSync()`, and then typescript will natively understand it and there's no need for macros.

SergeAstapov:
@ef4 thank you for prompt and detailed suggestions! I'll update the PR with these

snewcomer:
@SergeAstapov Happy to merge this once conflicts are resolved!

SergeAstapov:
> @SergeAstapov Happy to merge this once conflicts are resolved!

Unfortunately tests are failing with this change.
I hope I can get it pass by Mon or Tue

SergeAstapov:
@snewcomer @runspired this is ready for review.

Unfortunately I could not replace conditions like

```js
if (DEBUG && HAS_EMBER_DATA_PACKAGE && HAS_ADAPTER_PACKAGE) {
```

with 

```js
if (macroCondition(isDevelopingApp() && dependencySatisfies('ember-data', '*') && dependencySatisfies('@ember-data/adapter')) {
```

as it was suggested by @ef4 in the comment above:
As this is monorepo, `dependencySatisfies` would always satisfy because package would be present in `node_modules` even if not pulled into build by any of the dependencies or transitive dependencies.

snewcomer (approved):
Thank you so much for this PR!  I'm not sure if there are other options than type casting `as typeof import` but otherwise I'm happy with this work you put in!!

snewcomer:
@SergeAstapov Just to confirm

> Unfortunately I could not replace conditions like

This probably means we still have broken tests and work to do elsewhere, but this PR is an improvement overall?

SergeAstapov:
> @SergeAstapov Just to confirm
> This probably means we still have broken tests and work to do elsewhere, but this PR is an improvement overall?

So tests should be fine.

This PR removes *most* of the usage of `require`, some are left in `private-build-infra` package and related to `fetch` presence detection which I'm not sure how to fix as this point of time.

And the last point - yes, this should be an improvement overall as it removes *most* of the runtime `require` usage

runspired:
> As this is monorepo, dependencySatisfies would always satisfy because package would be present in node_modules even if not pulled into build by any of the dependencies or transitive dependencies.

@SergeAstapov after checking with @ef4 it seems that this should already be guarded against.

Looking here at the source seems to confirm that: https://github.com/embroider-build/embroider/blob/7d89b57fbfd9362ae3cf6ac04425d055a8530fc3/packages/macros/src/babel/dependency-satisfies.ts#L25-L35

Are you sure this was something we were hitting? We should investigate this more.

runspired:
PS I noted that some of the require usage was guarding code deprecated for 4.0, I've gone ahead and removed that code, so now there's less usage of require :)

SergeAstapov:
> > As this is monorepo, dependencySatisfies would always satisfy because package would be present in node_modules even if not pulled into build by any of the dependencies or transitive dependencies.
> 
> @SergeAstapov after checking with @ef4 it seems that this should already be guarded against.
> 
> Looking here at the source seems to confirm that: https://github.com/embroider-build/embroider/blob/7d89b57fbfd9362ae3cf6ac04425d055a8530fc3/packages/macros/src/babel/dependency-satisfies.ts#L25-L35
> 
> Are you sure this was something we were hitting? We should investigate this more.

Unfortunately looks like I'm missing something *or* it does not work as expected.

In this PR I replaced only `require()` usage and all tests pass.
In the fork I've created PR that replaces `HAS_*_PACKAGE` with `macroCondition(dependencySatisfies())` and tests fail ([here is the CI run](https://github.com/SergeAstapov/data/runs/5299043073?check_suite_focus=true)).

Below is screenshot of the the same test failed in CI where line

```js
if (macroCondition(dependencySatisfies('@ember-data/record-data', '*'))) {
```

evaluates to

```js
if (runtime.macroCondition(true)) {
```

even though no `@ember-data/record-data` package present in the build
dependencySatisfies false positive

happy to pair on this or try some other approach if you have any ideas about what I'm missing.

@runspired do you think we can/should land this PR as is and iterate on replacing `HAS_*_PACKAGE` in the follow up PR?

runspired:
@SergeAstapov if you rebase I think we can merge

packages/store/addon/-private/system/model/internal-model.ts
```diff
@@ -511,7 +514,7 @@ export default class InternalModel {
 
       return manyArray;
     }
-    assert(`hasMany only works with the @ember-data/record-data package`, HAS_RECORD_DATA_PACKAGE);
```
runspired, line 514: don't delete the boolean here, it's required

runspired: nvm its not and I'm crazy

runspired (approved):

SergeAstapov:
> @SergeAstapov if you rebase I think we can merge

done!

PR #7864, opened 2022-02-11
feat: better debuggability

runspired:
replaces #7227


snewcomer (approved):

packages/store/addon/-private/system/store/internal-model-factory.ts
```diff
@@ -80,14 +81,7 @@ export function setRecordIdentifier(record: RecordInstance, identifier: StableRe
...
 export function internalModelFactoryFor(store: CoreStore): InternalModelFactory {
-  let factory = FactoryCache.get(store);
-
-  if (factory === undefined) {
-    factory = new InternalModelFactory(store);
-    FactoryCache.set(store, factory);
-  }
-
-  return factory;
+  return FactoryCache.lookup(store);
```
snewcomer, line 84: I like the new API.

packages/record-data/addon/-private/graph/index.ts
```diff
@@ -26,7 +27,17 @@ import updateRelationshipOperation from './operations/update-relationship';
...
 type RelationshipEdge = ImplicitRelationship | ManyRelationship | BelongsToRelationship;
 
-const Graphs = new WeakMap<RecordDataStoreWrapper, Graph>();
+const Graphs = new WeakCache<RecordDataStoreWrapper, Graph>(DEBUG ? 'graph' : '');
+Graphs._generator = (wrapper: RecordDataStoreWrapper) => {
+  const graph = new Graph(wrapper);
+
+  // in DEBUG we attach the graph to the main store for improved debuggability
+  if (DEBUG) {
+    Graphs.set(wrapper._store as unknown as RecordDataStoreWrapper, graph);
```
snewcomer, line 36: Doesn't this happen when `lookup` is called?  Or is this adding the graph to an additional object?

runspired: @snewcomer we're adding it to an additional object here. The cache is key'd to the store-wrapper not the store in the map, but for debugging we also attach it to the store.

runspired:
- eslint-plugin-mocha no longer allows node 12, we'll need to update our node scenarios
- ember-m3's proxy breaks when we try to attach the debug symbol: we should try-catch setting it on record instances since that's not owned by us.

packages/store/addon/-private/system/weak-cache.ts
```diff
@@ -0,0 +1,117 @@
...
+  set(obj: K, value: V): this {
+    if (DEBUG && super.has(obj)) {
+      throw new Error(`${Object.prototype.toString.call(obj)} was already assigned a value for ${this._fieldName}`);
+    }
+    if (DEBUG) {
+      if (obj[DEBUG_IDENTIFIER_BUCKET] && this._fieldName !== 'identifier-proxy-target') {
+        const target: K = obj[DebugWeakCache._debugTargetProp as unknown as string] as K;
+        target[this._symbol as unknown as string] = value;
+      } else {
+        obj[this._symbol as unknown as string] = value;
```
snewcomer, line 109: interesting m3 error.  Their proxy is interpolating this._symbol to a string, which can't be done.

runspired: yeah, I just did a try-catch for now since m3 can either fix this on their end or add their own property/symbol for it if they so choose.

PR #7865, opened 2022-02-11
Cleanup eslint config

bmish:
Here are the changes I made. There's essentially no behavior change from any of these except now catching unused disable directives.

* Switch to prettier [recommended](https://github.com/prettier/eslint-plugin-prettier#recommended-configuration) config as a better way of enabling prettier
* Alphabetize lists of rules in .eslintrc.js
* Have ESLint report and autofix [unused disable directives](https://eslint.org/docs/user-guide/command-line-interface#--report-unused-disable-directives)
* ESLint now allows `extends` in an override without a hack so we can use this for `plugin:node/recommended`

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,7 +1,6 @@
 // See https://github.com/lydell/eslint-plugin-simple-import-sort#custom-grouping
 const ImportSortGroups = [
   // Side effect imports.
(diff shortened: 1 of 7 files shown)


ddzz (approved):

snewcomer (approved):
Thanks a ton for this PR!

runspired (approved):

PR #7866, opened 2022-02-11
Autofix most qunit linting

bmish:
Enable qunit lint rules that can be autofixed. This is all autofixed.

https://github.com/platinumazure/eslint-plugin-qunit

### DIFF
--- a/packages/-ember-data/tests/integration/identifiers/configuration-test.ts
+++ b/packages/-ember-data/tests/integration/identifiers/configuration-test.ts
@@ -175,7 +175,7 @@ module('Integration | Identifiers - configuration', function (hooks) {
       'Precond: We receive the expected identifier for a new record'
     );
     assert.strictEqual(identifier.id, null, 'Precond: We have no id yet');
-    assert.ok(updateMethodCalls === 0, 'Precond: We have not updated the identifier yet');
+    assert.strictEqual(updateMethodCalls, 0, 'Precond: We have not updated the identifier yet');
     updateCallback = (updatedIdentifier, resource) => {
       assert.strictEqual(identifier, updatedIdentifier, 'We updated the expected identifier');
       assert.strictEqual(resource.attributes!.firstName, 'James', 'We received the expected resource to update with');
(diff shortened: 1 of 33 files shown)


packages/-ember-data/tests/integration/identifiers/configuration-test.ts
```diff
@@ -507,7 +508,7 @@ module('Integration | Identifiers - configuration', function (hooks) {
     setIdentifierForgetMethod((identifier) => {
       forgetMethodCalls++;
       let expectedIdentifier = expectedIdentifiers.shift();
-      assert.ok(expectedIdentifier === identifier, `We forgot the expected identifier ${expectedIdentifier}`);
+      assert.strictEqual(expectedIdentifier, identifier, `We forgot the expected identifier ${expectedIdentifier}`);
```
bmish, line 511: This improvement has uncovered a type mismatch here.

```
$ tsc -p tsconfig.json --noEmit --pretty false
Error: packages/-ember-data/tests/integration/identifiers/configuration-test.ts(511,4[6](https://github.com/emberjs/data/runs/5158545581?check_suite_focus=true#step:7:6)): error TS2345: Argument of type 'StableIdentifier | StableRecordIdentifier' is not assignable to parameter of type 'StableRecordIdentifier | undefined'.
  Type 'StableIdentifier' is not assignable to type 'StableRecordIdentifier | undefined'.
    Type 'StableIdentifier' is missing the following properties from type 'StableNewRecordIdentifier': id, type
```

ddzz (approved):

runspired (approved):
Can investigate the TS issue is needed but looks at a glance like it's a problem with qunit's types?

packages/-ember-data/tests/integration/identifiers/configuration-test.ts
runspired: Given this works in the previous test when expectedIdentifier is not defined the easy fix is to change the type of the array. This seems like a bug in qunit's types to me though, the comparisons should be comparing `unknown` and not attempting to do a type comparison.

What think you @chriskrycho ?

bmish: I think the issue is that `expectedIdentifiers.shift()` could theoretically return `undefined` if the array is empty. So I added an `if` statement to avoid this assertion if the array is somehow empty.

snewcomer (approved):

PR #7868, opened 2022-02-13
RFC: ember data return promise from model.save

snewcomer:
https://github.com/emberjs/rfcs/pull/795

- New Deprecation
`DEPRECATE_SAVE_PROMISE_ACCESS`

To return a promise, you can set your compatWith to 4.4.
https://api.emberjs.com/ember-data/release/modules/@ember-data%2Fdeprecations


packages/private-build-infra/addon/current-deprecations.ts
```diff
@@ -39,4 +39,5 @@
 export default {
   DEPRECATE_CATCH_ALL: '99.0',
   DEPRECATE_3_12: '3.12',
+  DEPRECATE_SAVE_PROMISE_ACCESS: '4.2',
```
runspired, line 42: since 4.2 has shipped and this will be on 4.4 canary this should now be 4.4

packages/model/addon/-private/model.js
```diff
@@ -846,9 +855,20 @@ class Model extends EmberObject {
...
+    } else {
+      const promise = PromiseObject.create({
+        promise: this._internalModel.save(options).then(() => this),
+      });
+      if (DEPRECATE_SAVE_PROMISE_ACCESS) {
+        return deprecatedPromiseObject(promise);
+      } else {
+        return promise;
+      }
+    }
```
runspired, line 871:
```
const promise = this._internalModel.save(options).then(() => this);

if (DEPRECATE_SAVE_PROMISE_ACCESS) {
  return deprecatedPromiseObject(PromiseObject.create({ promise }));
}
return promise;
```

runspired (approved):
we talked finalized changes and once made and tests green we will ship

packages/store/addon/-private/system/promise-proxies.js
```diff
@@ -88,3 +89,34 @@ export function promiseArray(promise, label) {
...
+            since: {
+              available: '4.4',
+              enabled: '4.4',
+            },
+          }
+        );
+      }
+
+      /* global Reflect */
+      return Reflect.get(...arguments).bind(target);
```
Turbo87, line 116: when used with ember-concurrency, trying to access a `__ec_cancel__` property on the promise object triggers the following error:

```
promise-proxies.js:128 Uncaught TypeError: Cannot read properties of undefined (reading 'bind')
    at Object.get (promise-proxies.js:128:39)
    at TaskInstanceExecutor.handleYieldedValue (executor.js:233:34)
    at TaskInstanceExecutor.handleResolvedContinueValue (executor.js:160:10)
    at TaskInstanceExecutor.proceedSync (executor.js:112:12)
    at TaskInstanceExecutor.start (executor.js:52:10)
    at TaskInstance.start (base.js:77:19)
    at eval (refresh.js:24:65)
    at Array.forEach (<anonymous>)
    at Refresh.process (refresh.js:24:28)
    at EmberScheduler.refresh (scheduler.js:51:34)
```

this appears to be caused by `Reflect.get(...arguments)` returning `undefined` here.

Turbo87: unfortunately it seems like setting `compatWith: '4.4'` does not resolve this issue. this can be reproduced by the frontend test suite in https://github.com/rust-lang/crates.io/pull/4801

rwwagner90: @Turbo87 did you ever find a way around this issue?

Turbo87: I've opened https://github.com/emberjs/data/pull/8000 to fix it, but it's still not merged :-/

PR #7878, opened 2022-02-15
Remove lint rule to prevent global Promise

snewcomer:


.eslintrc.js
```diff
@@ -52,7 +52,7 @@ module.exports = {
     'no-console': 'error', // no longer recommended in eslint v6, this restores it
     'no-eq-null': 'error',
     'no-eval': 'error',
-    'no-restricted-globals': ['error', { name: 'Promise', message: 'Global Promise does not work in IE11' }],
+    'no-restricted-globals': ['error', { name: 'Promise', message: 'Must use RSVP.Promise' }],
```
snewcomer, line 55: We will want to maintain this correct?  Or with IE11 support dropped, can we remove this lint rule @runspired

runspired: with IE11 support dropped we can drop this

snewcomer: Any reason we can't look to drop rsvp?  Need an RFC?

runspired: We may still rely on hooking into it to allow promise cancellation in one situation.

PR #7881, opened 2022-02-17
Simplify import linting

bmish:
Some of these config lines were not needed or unnecessarily disabled.


.eslintrc.js
```diff
@@ -366,11 +365,7 @@ module.exports = {
       plugins: ['node', 'import'],
       extends: 'plugin:node/recommended',
       rules: {
-        'import/first': 'error',
-        'import/newline-after-import': 'error',
```
snewcomer, line 370: Is this a dup?

snewcomer (approved):
Thank yoU!!

PR #7885, opened 2022-02-18
chore: add eslint-plugin-eslint-comments

bmish:
Alternative to using ESLint's built-in [--no-unused-disable-directives](https://eslint.org/docs/user-guide/command-line-interface#--report-unused-disable-directives) CLI option, since that CLI option cannot be disable in a single problematic location, as we encountered issues with in https://github.com/emberjs/data/pull/7882/files#r809271031.

If necessary, we can use:

```js
// eslint-disable-next-line eslint-comments/no-unused-disable
```

to avoid the same problem.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -39,9 +39,16 @@ module.exports = {
     requireConfigFile: false,
   },
   plugins: ['prettier', 'qunit', 'mocha', 'simple-import-sort', 'import'],
-  extends: ['eslint:recommended', 'plugin:prettier/recommended', 'plugin:qunit/recommended'],
+  extends: [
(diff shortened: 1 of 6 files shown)


runspired:
unfortunately this was not enough. when the docs are generated we still end up stripping even with the extra disable. I'm going to change this to warn instead of auto fix.

runspired:
`warn` is not enough either, unfortunately going to have to just accept no reporting of unused disables.

bmish:
How about put this directive at the top of this single file to disable the rule? That should be allowed.

``
/* eslint eslint-comments/no-unused-disable:"off" */
```

PR #7900, opened 2022-02-24
fix: rollbackAttributes|unloadRecord during hasMany forEach should not error

### LINKED ISSUE #7886: rollbackAttributes or unloadRecord while iterating a sync hasMany relationship errors
Affects pretty much all versions of ember-data. Basically an operation that results in removal from the array being iterated affects later iterations, causing an out-of-bounds error. For each record unloaded, iteration will both skip the next item in the array and encounter an undefined index at the end of the array.

runspired:
fixes #7886 

This issue has existed for so long we shouldn't backport unless requested.


runspired:
cc @josh-fnbtech

Issue #7904, opened 2022-02-25
Computed properties dependent on async relationships do not recompute when the relationship fulfills 


kategengler:
### Reproduction

Reproduced here https://github.com/kategengler/datatest
With branches for 3.27 (working) and 4.2.0-beta.0 (not working)

### Description

A computed property dependent upon an async relationship does not recompute when the relationship fulfills.

In the reproduction, post `hasMany` (async) comments which `hasMany` (async) authors

https://github.com/kategengler/datatest/blob/main/app/models/post.js#L11-L18 is the computed property, it is referenced in `application.hbs` where the model is a single Post. 

In the console, you will see that the comments are fetched and the `authors` computed property is only run once, during which `comments.isFulfilled` is `false`.

If you uncomment `comments.isFulfilled` as a dependent key on the CP, you will see it run twice, once when it is fulfilled, and authors are fetched via the API, but still the authors do not list in the template. 

In the template, you can change to using `authorsTracked` and see both authors fetched and displayed in the template.

Similarly, you can switch to branch `ember-data-327` and see expected behavior while using the CP. 
Branch `ember-4` uses ember-data `4.2.0-beta.0` and has the same unexpected behavior as `3.28`

### Versions

Run the following command and paste the output below: `yarn list ember-source && yarn list ember-cli && yarn list --pattern ember-data`.

```cli
└─ ember-source@3.28.8
✨  Done in 0.54s.
yarn list v1.22.17
warning Filtering by arguments is deprecated. Please use the pattern option instead.
└─ ember-cli@3.28.5
✨  Done in 0.54s.
yarn list v1.22.17
├─ @ember-data/adapter@3.28.8
├─ @ember-data/canary-features@3.28.8
├─ @ember-data/debug@3.28.8
├─ @ember-data/model@3.28.8
├─ @ember-data/private-build-infra@3.28.8
├─ @ember-data/record-data@3.28.8
├─ @ember-data/rfc395-data@0.0.4
├─ @ember-data/serializer@3.28.8
├─ @ember-data/store@3.28.8
├─ babel-plugin-ember-data-packages-polyfill@0.1.2
└─ ember-data@3.28.8
```

bmfay:
@kategengler we encountered widespread issues after upgrading to 3.28.  Can confirm it is the same issue with computed properties on async relationships and that it worked through 3.27 and no longer works in 3.28.   It is definitely Ember Data since our CP's work fine with Ember CLI 3.28 and Ember Data 3.27.

[...]

PR #7945, opened 2022-04-14
fix for #7904 computed chains not updated for async hasMany loading

runspired:


bmfay:
Awesome @runspired, thanks for figuring this out.  Will 3.28 LTS get a release with this fix?  Is there a way to be notified if so (perhaps just subscribing to this PR)?  Thanks!

runspired:
@bmfay itll get a release, probably tomorrow. Not sure there's a way to subscribe to anything unless npm allows.

ginomiglio:
Also interested in a release of this fix 🙏

PR #7957, opened 2022-04-18
feat: register the future

runspired:
This PR implements the registry approach we've outlined in past team discussions. This approach so far has turned out to be *achievable* though with caution. It is *extremely* easy to accidentally find yourself in what I can only describe as subtype hell.

I don't recommend trying to change the pattern from what I've established here very much, I've had to find patterns that work effectively without hitting various edge cases that typescript is not yet prepared to handle for generics.

I'm not looking to type 100% of data to ship this, but it is a big enough thread-pull that roughly speaking that's what is happening. We'll ship as soon as we get this to "green" and iterate from there.

**Unanswered Questions**

- ~~How should async relationships be typed?~~
- ~~(depends on prior Q) should we (can we even?) look up inverse types for relationships from the registry.~~
- What patterns should we publicly surface for typing async belongsTo relationship sets?
- Is it better for module-level constants to use the default registry or be typed with unknown/object/string etc. instead? Either way any interaction with a module-level constant requires casting, but default registry *might* help the end-consumer story for importable utils like `recordIdentifierFor`,


runspired:
after further time, exploration and discussion we are not going to pursue the registry approach.

Issue #7997, opened 2022-05-26
Issues with new save promise deprecation


runspired:
The implementation in https://github.com/emberjs/data/pull/7868 has a few issues we should address.

- it does not allow for native object methods such as `hasOwnProperty` etc. that we would still expect to be present.
- it does not proxy properties correctly as it assumes `return Reflect.get(...arguments)` returns a method and calls `.bind(target);`. While we don't want to introduce a new behavior (dot notation lookups) we still do need to allow for the situation in which the property might be expected to be on the promise-object itself.

cc @machty @Turbo87 @snewcomer

Turbo87:
for cross referencing: https://github.com/machty/ember-concurrency/issues/463

runspired:
See #8025

sly7-7:
@runspired I'm sorry, I can't see if this issue is supposed to be closed by #8025. I'm on 4.6.1 and still encounter something like https://github.com/machty/ember-concurrency/issues/463, so I guess some cases are not handled.
I just don't know at all what should be done :(

bartocc:
I was in the process of upgrading our codebase to ember-data@4.5/4.6 this morning and, like @sly7-7 , could not figure if #8025 was supposed to fix everything or not…

runspired:
the original issues have been fixed but it now appears it can lead to infinite recursion of EC tasks in some scenarios: requires investigation.

PR #8000, opened 2022-05-31
promise-proxies: Fix "Cannot read properties of undefined (reading 'bind')" error

Turbo87:
This PR partly resolves https://github.com/emberjs/data/issues/7997 by checking if the property has a `bind` method before using it.


snewcomer (approved):
Thank you!

jrjohnson:
Going to close #7988 as this addresses the problem instead of just adding the test. Thanks @Turbo87!

PR #8027, opened 2022-07-15
fix test runs, update lts support tests

runspired:


packages/-ember-data/package.json
```diff
@@ -31,63 +31,64 @@
...
     "qunit-dom": "^2.0.0",
-    "typescript": "~4.6.2",
-    "webpack": "^5.37.1"
+    "typescript": "~4.7.4",
+    "webpack": "^5.73.0"
   },
   "engines": {
-    "node": "12.* || >= 14.*"
+    "node": "12.* || >= 14.*",
+    "yarn": "1.22.19"
```
Turbo87, line 91: this change is causing issues for people installing ember-data with older yarn versions:

> ember-data@4.6.0: The engine "yarn" is incompatible with this module. Expected version "1.22.19". Got "1.22.18"

runspired: Ah. I forgot that is going to get applied externally. We should remove it then. Annoying cause it's nice for folks that don't use Volta.

PR #8078, opened 2022-07-26
chore: InternalModel burndown

runspired:
Picks up where #8055 left off

- continues to simplify remaining store APIs and align them to the coming network layer rationalization
- continues to replace InternalModel usage with Identifier usage (there's some friction here as we figure out where to store state that isn't necessarily related to DSModel and required for proper book-keeping)
- identifies some locations where code could be conditionally removed if the model package is not present to give those consumers significant savings (core of ED is ~10KB if not using the legacy stuff once this lands)

Big goal for this PR is to get relationship state classes out of core and into the model package. A still unanswered question is what to do with references. This has been a circular debate on the core team in the past, the war of attrition suggests my path forward is the one we take (references should die with @ember-data/model) though I'm open to other paths forward. My reasoning is that directly asking the cache for relationship information is more useful, and we can expose the Graph as public API in the future if we so choose (again more useful). References were always little more than a workaround for the messiness of async relationships, and if those are dying anyway there's not much reason to keep them around in my mind.

----

Important release note: users of only the core of ember-data (for instance users of ember-m3) can expect the total size to shrink between 35-50% once upgrading to a release with these changes and those in #8055


sandstrom:
> References were always little more than a workaround for the messiness of async relationships, and if those are dying anyway there's not much reason to keep them around in my mind.

Just curious, does this mean that there are plans to remove the concept of async relationships?

runspired:
@sandstrom yes! Hopefully for 5.0 this fall (see #8086) but definitely within the year. Basically any replacement to `@ember-data/model` will not include `belongsTo` and `hasMany` in the form they exist today.

PR #8087, opened 2022-07-28
chore: conversion to pnpm

runspired:
In theory this will speed up our CI times. It may also reduce the number of annoying issues we fall into from time to time with yarn bugs.

Mostly a find/replace job, there's probably still some yarn commands to fix and definitely going to be some yarn commands that should have been left that got converted by accident.


runspired:
going to be too difficult with all the tooling to do this easily, skipping for now

PR #8089, opened 2022-07-28
fix: unpin yarn for consumers

runspired:
engines is respected by consuming applications to, so this accidentally forced consumers using yarn to be on latest yarn. We only need to pin in root which is unpublished, so this removes the pin from everything but root that we publish.


runspired:
Seems we have introduced a leaky test, I think we're going to just release this and figure that out after, I suspect I know what changed.

PR #8109, opened 2022-08-09
add optional polyfill

### LINKED ISSUE #8097: uuid generation broken for insecure contexts
We should either revert the change or make the function pluggable so folks can supply one when needed.

see https://github.com/WICG/uuid/issues/23

### LINKED ISSUE #8106: _crypto.randomUUID is not a function
packages/store/addon/-private/identifier-cache.ts:37

Tested in Chrome Android Version 69 -> the functionality is available at Version 92
(https://developer.mozilla.org/en-US/docs/Web/API/Crypto/randomUUID)

The issue occurs on using "createRecord". The fast boot mode is false, so it uses window.crypto that results in an empty object. The consequence, the function "randomUUID" is not defined.

Is there any polyfill, downward compatibility ?

Thank you a lot for your work.

runspired:
resolves #8097 
resolves #8106 

To use the polyfill:

```
let app = new EmberApp(defaults, {
  '@embroider/macros': {
    setConfig: {
      '@ember-data/store': {
        polyfillUUID: true
      },
    },
  },
});
```

### DIFF
--- a/packages/private-build-infra/src/addon-build-config-for-data-package.js
+++ b/packages/private-build-infra/src/addon-build-config-for-data-package.js
@@ -22,6 +22,11 @@ function addonBuildConfigForDataPackage(PackageName) {
       this._prodLikeWarning();
       this.debugTree = BroccoliDebug.buildDebugCallback(`ember-data:${PackageName}`);
       this.options = this.options || {};
+      Object.assign(this.options, {
+        '@embroider/macros': {
+          setOwnConfig: {},
+        },
+      });
     },
 
     _prodLikeWarning() {
--- a/packages/store/addon/-private/caches/identifier-cache.ts
+++ b/packages/store/addon/-private/caches/identifier-cache.ts
@@ -4,6 +4,8 @@
 import { assert, warn } from '@ember/debug';
 import { DEBUG } from '@glimmer/env';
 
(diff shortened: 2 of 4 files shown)


law-rence:
Thank you very much for the quick solution!

ef4 (approved):
This looks good to me.

It's also possible for `@ember-data/store` to set its own macro config based off it's own decisions, like checking for the presence of a package or some other ember-data-specific configuration. But exposing the macros config directly to users is also supported and avoids the need for a lot of extra code.

luxferresum:
Shouldnt this be the default? IMHO removing the support for insecure contexts without a changed configuration is kinda a breaking change :thinking:

At *least* we should have a better error message, helping how to avoid this issue.

runspired:
@luxferresum opened #8301 for this ^

gzurbach:
I spent a couple hours this morning trying to enable the polyfill as described above, until I realized that I needed to install `@embroider/macros` as well in my app:

```
npm i @embroider/macros --save-dev
```

It may seem obvious to some, but not to me. So I am adding this here for anyone else encountering the same issue :)

runspired:
@gzurbach yeah, you only need to do that if we don't forward the config (which we should do). Haven't had time to work on that but if you do it would be hugely appreciated, shouldn't be too complicated we already forward the configs for near everything else.

PR #8112, opened 2022-08-09
update package presence checks to use macros

richgt:
## Type of PR

What kind of change is this?

- [x] refactor
- [ ] internal bugfix
- [ ] user-facing bugfix
- [ ] new feature
- [ ] deprecation
- [ ] documentation
- [ ] tests

## Notes for the release

Updated package presence checks to use use macros instead of simple has checks.

### DIFF
--- a/packages/private-build-infra/src/addon-build-config-for-data-package.js
+++ b/packages/private-build-infra/src/addon-build-config-for-data-package.js
@@ -17,6 +17,14 @@ function addonBuildConfigForDataPackage(PackageName) {
   return {
     name: PackageName,
 
+    options: {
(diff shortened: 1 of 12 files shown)


packages/private-build-infra/src/addon-build-config-for-data-package.js
```diff
@@ -17,6 +17,14 @@ function addonBuildConfigForDataPackage(PackageName) {
   return {
     name: PackageName,
 
+    options: {
+      '@embroier/macros': {
```
runspired, line 21:
```suggestion
      '@embroider/macros': {
```

PR #8116, opened 2022-08-12
Clear subscriptions once unsubscribed, don't unnecessarily churn on subscriptions

### LINKED ISSUE #7857: BelongsTo-Relation changes throw error in Store -> unsubscribe
Hi there,

i have a existing model with attributes and relationships. For example this model has a existent belongsTo relationship named "gender". 

When i set this belongsTo to null, all good at the moment, i get no error.

And if I now but change this nullified belongsTo to another relation, i get a error from stores record-notification-manager like "Passed unknown unsubscribe token to unsubscribe". 

If i change this belongsTo directly to a new relation, i get no error.

In the case that I have loaded an existing model where the belongsTo "gender" is null and I set to a relation, I don't get an error.

This is a bug that I don't think appeared before ember 4. I think the problem comes with this change https://github.com/emberjs/data/commit/856f7b33ab60b9327839fe80b7ae3588e863f657.

I use:
ember-cli 4.1.1
ember-source 4.1.0
ember-data 4.1.0

runspired:
fix #7857

### DIFF
--- a/packages/-ember-data/tests/integration/references/autotracking-test.js
+++ b/packages/-ember-data/tests/integration/references/autotracking-test.js
@@ -32,6 +32,10 @@ module('integration/references/autotracking', function (hooks) {
         createRecord() {
           return { data: { id: '6', type: 'user' } };
         }
+        deleteRecord() {
+          console.log('delete record called');
+          return { data: null };
+        }
       }
     );
     owner.register(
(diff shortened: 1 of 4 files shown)


jherdman:
Any chance we could get a patch release for 3.28 including this? ❤️

runspired:
cc @jrjohnson lets make sure this one gets backported too

jherdman:
@runspired @jrjohnson apologies for nagging, but is there a chance we could get that backport? I'm happy to help.

jherdman:
I'll see what I can do. This is a bit of a show stopper for us at Precision Nutrition on our journey to Ember v4+.

runspired:
I think the file moved but otherwise I suspect the change applies pretty cleanly

runspired:
Also we no longer need beta and release since we cut a release that has this fix in it. We need lts releases though

runspired:
Working on porting this to 3.28

PR #8118, opened 2022-08-12
fix: reopen deprecation should properly pass original arguments

law-rence:
Call original reopen with "...arguments"

## Description

Recall the original reopen function with "arguments" instead of "...arguments".

## Type of PR

What kind of change is this?

- [ ] refactor
- [x] internal bugfix
- [ ] user-facing bugfix
- [ ] new feature
- [ ] deprecation
- [ ] documentation
- [ ] tests

## Notes for the release

### DIFF
--- a/packages/model/addon/-private/model.js
+++ b/packages/model/addon/-private/model.js
@@ -2413,7 +2413,7 @@ if (DEBUG) {
         until: '5.0',
         since: { available: '4.8', enabled: '4.8' },
       });
-      return originalReopen.call(this, arguments);
+      return originalReopen.call(this, ...arguments);
     };
 
     Model.reopenClass = function deprecatedReopenClass() {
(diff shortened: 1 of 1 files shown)


packages/model/addon/-private/model.js
```diff
@@ -2413,7 +2413,7 @@ if (DEBUG) {
         until: '5.0',
         since: { available: '4.8', enabled: '4.8' },
       });
-      return originalReopen.call(this, arguments);
```
runspired, line 2416: 🤦🏻 

Can we use apply instead? IIRC it passes through arguments when proxied like this more cleanly.

runspired: going to merge this, we can always change it later. It's deprecation code anyway.

PR #5931, opened 2019-03-24
[FEAT RecordDataV2] Implements RFC #461 Singleton RecordData

runspired:

PR #8122, opened 2022-08-13
v2.1 cache

runspired:
Performance (and size) will get worse before they get better for this feature. Since we have to wrap every single RecordData in a unique manager so long as V1 exists, I would expect this to initially negate a significant portion of the (sizeable) improvements from the simplification work and removal of InternalModel.

However!

Since the fully v2 cache world is capable of being a singleton (and our implementation will be so), once v1 is not in the system we drop from 2N classes to support N records in the cache to two class instances (total). It is at that point I would expect to see us regain double what we lost from adding in the manager.

My hope is that we can produce a mechanism such that an app may signal that V1 is not in use, which our standard build infra will then make use of to remove the expensive codepath and utilize the cheap one. Ideally this will all be in place prior to merging this PR.

## Release Notes

- [x] Implements NonSingleton Manager
- [x] Implements Cache v2
- [x] Implements Singleton Manager
- [x] Documents the Manager
- [x] Implements Production Manager Mode
- [x] Activates Singleton Cache (but not Singleton manager)

The singleton cache and singleton manager are currently both behind feature flags, only singleton cache is activated (since this is our own implementation this is easier for us to achieve with some safety).

A follow up PR to this one will activate singleton manager as well as document the v2 cache format and deprecate the v1 cache format.

While studying the cost centers of the various benchmarks I came away with the following observations (once singleton-manager is activated)

- our biggest cost centers are down to three things:

1) instantiating things which extend Model
2) populating the graph
3) updating the live RecordArray (specifically updates, not full clears / add from empty)

(1) and (3) are respectively cost centers that will be significantly improved as we see through the plan for 5.0, as eliminating EmberObject / ArrayProxy from the equation will simplify this significantly.

(2) requires rebooting and finishing #7521 with the additional observation that we should eliminate the classes used to hold state for each edge as they are largely dumb containers at this point.


runspired:
this PR replaces #5931

PR #8131, opened 2022-08-17
feat: Lazy Relationships & Relationship Diffing

runspired:
This is a refactor of the internals of the graph into a data structure previously proved out in several older versions of ember-data. Due to various complications these optimizations were never landed; however, the internals (and especially relationships) are now simple enough they may.

The underlying concepts here are around optimizing allocation, instantiation, lookups, iteration and timing.

- [x] **make relationship storage simple objects**

> By moving away from classes to object for storage we both reduce the total number of allocations and the instantiation cost. For 10k records with a single relationship each this is 20k fewer calls to `new Klass()` as well as several fewer objects allocated in the process.

- [ ] **make update propagation more efficient**

> Currently GC accounts for about 15% of the time spent populating the graph. This is due partly to the high temporary allocation overhead operations have and mostly to the temporary arrays and Sets created during the population process. Instead of decomposing operations into more operations we should find a simpler way of executing the instruction set. The allocation of temporary object should be reduced by the diff mechanism itself.

- [ ] **lazy local state updates**

> We do not need to sync remote updates to local updates until the app has requested the current state. To achieve this we notify the change but do not actually do the work until pull.

- [x] **lazy remote state updates**

> We do not need to flush remote state updates unless the app has requested the current state. To achieve this we notify the change but do not actually do the work until pull.

- [ ] **persistent diff**

> The most common operations on the graph are membership checks (which we already optimize by use of a Set) and updates (which require performing a diff). Updates carry through the system: the changes applied to remote state will also need to be applied to local state, the changes applied to local state will need to be applied to the ManyArray, the changes applied to the ManyArray will need to be diffed by glimmer when updating the contents of `#each`. Our goal is to reduce how often we diff and store information such that the diff never needs to be recalculated.

### End Goals

Our long term intent with the graph is to make it entirely public API. This would mean that relationship rollback, relationship diff, and local/remote state information would all be available to consuming applications as needed.

After this PR lands, but before we can make the graph public, we have two primary problems to resolve: (1). how to handle relationship pointers in nested attribute keys. (2) how to embed the proposed pagination structure for collections within the graph.


runspired:
this branch is no longer needed as the diffing work was completed in #8807 and the lazy work is being done in #8806 

all other ideas here have now been ported to #6573 for tracking

PR #8142, opened 2022-08-22
perf: cache access

runspired:
Ports cache access optimizations from https://github.com/emberjs/data/pull/8134. Same as https://github.com/emberjs/data/pull/8139 this is to ensure that the benchmark for that PR is as close as possible to "just the cost of moving to proxies".


runspired:
It's not clear to me yet why two sections of the basic-record-materialization scenario are regressing. Granted overall this change still provides a substantial "win" for that scenario overall, but these regressions are too large to ignore.

On paper, we shouldn't be doing any more (or less) work in either of these sections, and profiling current master against this PR does not show any substantial regression or change in shape of what is called.

My best hypothesis is that somehow we've pushed a lot of GC pressure from earlier in the bench to later, and so gains we saw earlier (in that case) might be in part due to this shift; however, while GC does account for 10-12% locally, it seems to account for an equal share on both the master branch and the PR branch. Will dig a bit more.

runspired:
spent a lot more time poking at this, and what's happening in the basic-materialization case simply does not replicate locally. My best bets are all on some random compile/GC timing shift or change. Since even that benchmark sees substantial improvement with these changes despite the big regression in a key area, and since other benchmarks that also test materialization do not show such a regression, I'm going to merge this.

PR #8153, opened 2022-08-26
fix: account for multi-options-per-identifer possibility

jrjohnson:
backport #8113 to release


jrjohnson:
@runspired I'm stuck here, there is some issue with my merge, but I'm unable to find it. As far as I can tell I've backported the entire change, but tests are failing. Not sure how much value this has since 4.7 is now released anyway and a 4.6.1 release is nice to have, but probably not critical. If you have some time can you take a look and see if you can spot what I missed?

runspired:
likely we don't need this anymore as stable releases with the fix have gone out

PR #8171, opened 2022-09-08
fix: filterBy should compare values

jrjohnson:
## Description
FilterBy can be called with and without a second argument, when the second argument is used the filter should compare that value to the value in the record.

I slapped the test for this in a sort of random spot because I wasn't sure where it goes.

## Type of PR

What kind of change is this?

- [ ] refactor
- [ ] internal bugfix
- [x] user-facing bugfix
- [ ] new feature
- [ ] deprecation
- [ ] documentation
- [ ] tests

### DIFF
--- a/packages/-ember-data/tests/unit/record-arrays/record-array-test.js
+++ b/packages/-ember-data/tests/unit/record-arrays/record-array-test.js
@@ -6,6 +6,7 @@ import { setupTest } from 'ember-qunit';
 import Model, { attr } from '@ember-data/model';
 import { recordIdentifierFor } from '@ember-data/store';
 import { RecordArray, SnapshotRecordArray, SOURCE } from '@ember-data/store/-private';
+import { deprecatedTest } from '@ember-data/unpublished-test-infra/test-support/deprecated-test';
 import testInDebug from '@ember-data/unpublished-test-infra/test-support/test-in-debug';
 
 class Tag extends Model {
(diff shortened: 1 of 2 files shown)


packages/store/addon/-private/record-arrays/identifier-array.ts
```diff
@@ -746,12 +746,12 @@ if (DEPRECATE_ARRAY_LIKE) {
...
-      return this.filter((value) => {
-        return Boolean(get(value, key));
+      return this.filter((record) => {
+        return get(record, key) === value;
       });
     }
-    return this.filter((value) => {
-      return Boolean(get(value, key));
+    return this.filter((record) => {
+      return Boolean(get(record, key));
```
runspired, line 754: :facepalm: you may want to check the other implementations to make sure I didn't make a similar mistake with them

runspired (approved):

packages/-ember-data/tests/unit/record-arrays/record-array-test.js
```diff
@@ -97,6 +97,46 @@ module('unit/record-arrays/record-array - DS.RecordArray', function (hooks) {
     assert.strictEqual(recordArray[3], undefined);
   });
 
+  test('#filterBy', async function (assert) {
```
runspired, line 100: you'll want to use `deprecatedTest` vs `assert.expectDeprecation` both so we know to clean it up later and because it will then get skipped when deprecations are removed and we run the tests.

runspired: this is a perfectly fine place for this test btw

packages/store/addon/-private/record-arrays/identifier-array.ts
jrjohnson: 😆 copy paste so useful, so easy to mess up. I eyeballed the rest and they looked good, decided not to wrote a bunch of soon to be deprecated tests though seemed like overkill.

packages/-ember-data/tests/unit/record-arrays/record-array-test.js
jrjohnson: 🆒 thanks. Fixed.

PR #8179, opened 2022-09-09
fix: filterBy should compare values

jrjohnson:
Backport #8171 FilterBy can be called with and without a second argument, when the second argument is used the filter should compare that value to the value in the record.


runspired:
I'm going to kill the use of the beta branch with this release, so this is no longer necessary, but thank you!

PR #8181, opened 2022-09-12
backport fix for clearing subscriptions from #8116 (fix #7857)

### LINKED ISSUE #7857: BelongsTo-Relation changes throw error in Store -> unsubscribe

runspired:


packages/store/addon/-private/system/references/has-many.ts
```diff
@@ -76,6 +76,7 @@ export default class HasManyReference extends Reference {
   destroy() {
     if (CUSTOM_MODEL_CLASS) {
       unsubscribe(this.#token);
+      this.#token = null as unknown as object;
```
jherdman, line 79: This is a bit out of band, but tell me more about this double `as`.

runspired: if something already has a type, you can't recast it without first casting to unknown

runspired:
The LTS scenario failures appears to be a CI only issue and unrelated to this change

PR #8183, opened 2022-09-13
Convert to pnpm

richgt:
## Description
Convert to pnpm

## Type of PR

What kind of change is this?

- [x ] refactor
## Notes for the release

Converting to use pnpm instead of yarn.


.github/ISSUE_TEMPLATE.md
```diff
@@ -18,20 +18,20 @@ Please provide one of the following:
...
 ### Description
 
-Describe the issue in a few sentences, include both the *expected* result and the observed *unexpected* result.
+Describe the issue in a few sentences, include both the _expected_ result and the observed _unexpected_ result.
 If a previous version of `ember-data` worked as `expected`, which was the most recent version that worked?
 
 ### Versions
 
-Run the following command and paste the output below: `yarn list ember-source && yarn list ember-cli && yarn list --pattern ember-data`.
+Run the following command and paste the output below: `pnpm list ember-source && pnpm list ember-cli && pnpm list --pattern ember-data`.
```
runspired, line 26: we should do a "if using npm/pnpm/yarn" approach to this

.github/workflows/alpha-release.yml
```diff
@@ -40,7 +40,7 @@ jobs:
           node-version: 16.x
           cache: 'yarn'
```
runspired, line 41: update cache

.github/workflows/asset-size-check.yml
```diff
@@ -34,21 +34,21 @@ jobs:
       - name: Checkout master
         run: git checkout master
       - name: Install dependencies for master
-        run: yarn install
+        run: pnpm install
```
runspired, line 37: update cache above

scripts/add-dev-dependency.js
```diff
@@ -6,7 +6,7 @@
  * Usage
  *
  * ```
- * yarn run add-dev <package@version>
+ * pnpm run add-dev <package@version>
```
runspired, line 9: we can just delete this file and anything that references it I think

scripts/sync-dev-dependency.js
```diff
@@ -6,7 +6,7 @@
  * Usage
  *
  * ```
- * yarn run sync-dev
+ * pnpm run sync-dev
```
runspired, line 9: I think we can just delete this file at this point

scripts/test-external-partner-project.js
```diff
@@ -93,7 +93,7 @@ try {
   );
 }
 
-const useYarn = fs.existsSync(path.join(projectTempDir, 'yarn.lock'));
+const usePnpm = fs.existsSync(path.join(projectTempDir, 'yarn.lock'));
```
runspired, line 96: ~~I think we need a use yarn|pnpm|npm check now vs just yarn|npm.~~

Lets see how they run first

yarn.lock
```diff
@@ -1,5 +1,5 @@
 # THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
-# yarn lockfile v1
+# pnpm lockfile v1
```
runspired, line 2: this file should just be deleted

package.json
```diff
@@ -14,22 +11,23 @@
...
     "lint:features": "node ./scripts/lint-features.js",
     "lint:js": "eslint --cache --ext=js,ts .",
     "lint:js:changed": "node ./scripts/lint-changed-js.js",
+    "preinstall": "npx only-allow pnpm",
     "problems": "tsc -p tsconfig.json --noEmit --pretty false",
-    "start": "yarn workspace ember-data start",
+    "start": "pnpm --filter ember-data start",
     "test": "lerna run --stream --no-private --ignore='@ember-data/{private-build-infra,canary-features,debug}' test",
-    "test:all": "yarn workspace ember-data test:all",
+    "test:all": "pnpm --filter ember-data test:all",
```
runspired, line 23: my suspicion is that --filter won't work in a lot of these script scenarios the way we need it to, because it will match anything with an `@ember-data/<foo>` name too :/

package.json
```diff
@@ -1,9 +1,6 @@
 {
   "name": "root",
```
runspired, line 2: we should pin pnpm with Volta in each package.json

in the root package.json we should additionally error for yarn and npm usage, but we don't want to do this in the others as it would break consumers.

```
"engines": {
    "yarn": "use pnpm",
    "npm": "use pnpm",
    "pnpm": "x.y.z"
  },
  "volta": {
    "node": "16.16.0",
    "pnpm": "x.y.z"
  }
    ```

File: .github/workflows/asset-size-check.yml (hunk 3 above)
Comment (runspired) [reply .github/workflows/asset-size-check.yml:37]: all of the workflows likely have caches to update btw, just pointed out the first couple

File: .github/workflows/alpha-release.yml [hunk 10]
```diff
@@ -20,17 +20,17 @@ jobs:
...
       - name: Install dependencies for master
```
NullVoxPopuli, line 25: you may be interested in this way of abstracting the boilerplate from pnpm + cache, etc:
https://github.com/NullVoxPopuli/baobab-ram/pull/1/files#diff-423ef59ce8dec1c2647bc93466fa66fd1eb65a9a65a631d4730fdb753b0ef792

.github/workflows/main.yml
```diff
@@ -170,37 +181,42 @@ jobs:
...
+        with:
+          version: '7.12.2'
       - uses: actions/setup-node@v3
         with:
           node-version: 16.x
-          cache: 'yarn'
+          cache: 'pnpm'
       - name: Install dependencies w/o lockfile
-        run: yarn install --no-lockfile --non-interactive
+        run: pnpm install --non-interactive
```
runspired, line 217: lost the --no-lockfile here

PR #8185, opened 2022-09-16
Use array notation instead of at for accessing values (#8184)

jrjohnson:
Backport #8184 to release:

Array.at isn't supported in Safari until 15.4, using brackets isn't as pretty, but enjoys universal support.


runspired:
@jrjohnson we probably should do this for `objectAt` and `objectsAt` as well

PR #8189, opened 2022-09-21
Use array notation instead of at for accessing object at values

jrjohnson:
## Description

Array.at isn't supported in Safari until 15.4, using brackets isn't as pretty, but enjoys universal support.

## Type of PR

- [x] user-facing bugfix

### DIFF
--- a/packages/store/addon/-private/record-arrays/identifier-array.ts
+++ b/packages/store/addon/-private/record-arrays/identifier-array.ts
@@ -631,12 +631,14 @@ if (DEPRECATE_ARRAY_LIKE) {
 
   IdentifierArray.prototype.objectAt = function (index: number) {
     deprecateArrayLike(this.DEPRECATED_CLASS_NAME, 'objectAt', 'at');
-    return this.at(index);
+    //For negative index values go back from the end of the array
+    let arrIndex = Math.sign(index) === -1 ? this.length + index : index;
+    return this[arrIndex];
   };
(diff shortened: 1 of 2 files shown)


runspired (approved):

packages/store/addon/-private/record-arrays/identifier-array.ts
```diff
@@ -631,12 +631,14 @@ if (DEPRECATE_ARRAY_LIKE) {
 
   IdentifierArray.prototype.objectAt = function (index: number) {
     deprecateArrayLike(this.DEPRECATED_CLASS_NAME, 'objectAt', 'at');
-    return this.at(index);
+    //For negative index values go back from the end of the array
+    let arrIndex = Math.sign(index) === -1 ? this.length + index : index;
```
jrjohnson, line 635: This handles `-1`, which may be the only practical usage here, but it won't handle `-5` in a three element array. I could probably keep looping around. I also considered adding an assertion here, but that seems more likely to be a breaking change than this is. I dunno, thoughts?

runspired: I think not wrapping is fine, if folks open an issue we can point them at `at`, which not only still works but is polyfillable if needed

jrjohnson: 👍

PR #8197, opened 2022-09-30
Fixing broken links in docs

MichalBryxi:
When I try to click links that I'm touching in this PR on the page: https://api.emberjs.com/ember-data/4.6/classes/Store

I will get to: https://api.emberjs.com/ember-data/4.6/methods/createRecord?anchor=createRecord which fails with: `Cannot GET /ember-data/4.6/methods/createRecord`.

etc.

I'm not 100% sure this is correct. I just hand-crafted this PR straight from github following the existing paths. If someone can double-check that would be awesome.

## Type of PR

What kind of change is this?

- [ ] refactor
- [ ] internal bugfix
- [x] user-facing bugfix
- [ ] new feature
- [ ] deprecation
- [ ] documentation
- [ ] tests

## Notes for the release

### DIFF
--- a/packages/store/addon/-private/store-service.ts
+++ b/packages/store/addon/-private/store-service.ts
@@ -134,12 +134,12 @@ export interface CreateRecordProperties {
   The store provides multiple ways to create new record objects. They have
   some subtle differences in their use which are detailed below:
 
-  [createRecord](../methods/createRecord?anchor=createRecord) is used for creating new
+  [createRecord](./classes/Store/methods?anchor=createRecord) is used for creating new
   records on the client side. This will return a new record in the
   `created.uncommitted` state. In order to persist this record to the
   backend, you will need to call `record.save()`.
 
-  [push](../methods/push?anchor=push) is used to notify Ember Data's store of new or
+  [push](./methods/push?anchor=push) is used to notify Ember Data's store of new or
   updated records that exist in the backend. This will return a record
   in the `loaded.saved` state. The primary use-case for `store#push` is
   to notify Ember Data about record updates (full or partial) that happen
@@ -147,7 +147,7 @@ export interface CreateRecordProperties {
   [SSE](http://dev.w3.org/html5/eventsource/) or [Web
   Sockets](http://www.w3.org/TR/2009/WD-websockets-20091222/)).
 
-  [pushPayload](../methods/pushPayload?anchor=pushPayload) is a convenience wrapper for
+  [pushPayload](./methods/pushPayload?anchor=pushPayload) is a convenience wrapper for
   `store#push` that will deserialize payloads if the
   Serializer implements a `pushPayload` method.
 


MichalBryxi:
The CI is not happy with:

> Select at least one label for what release channels to target. Use the label `target:canary` and none of the others in the list if this PR should not be backported. ['target:canary', 'target:beta', 'target:release', 'target:lts', 'target:lts-prev']

I can't edit GitHub labels on the PR though. How do I ... label it?

packages/store/addon/-private/store-service.ts
```diff
@@ -134,20 +134,20 @@ export interface CreateRecordProperties {
   The store provides multiple ways to create new record objects. They have
   some subtle differences in their use which are detailed below:
 
-  [createRecord](../methods/createRecord?anchor=createRecord) is used for creating new
+  [createRecord](../../classes/Adapter/methods?anchor=createRecord) is used for creating new
```
runspired, line 137: Adapter is the wrong class, probably want Store here

runspired: more likely just need to do `./` instead of `../`

MichalBryxi: Adapter -> Store ✔️ do agree.

Page where the link is found:

```
https://api.emberjs.com/ember-data/4.7/modules/@ember-data%2Fstore
```

Page where the user should land:
```
https://api.emberjs.com/ember-data/4.7/classes/Store/methods?anchor=createRecord
```

That's why I did `..` twice as it seems like the link should get out of `@ember-data/store` and then out of `modules` first.

So my conclusion here is that `./` is relative to `https://api.emberjs.com/ember-data/4.7/`?

MichalBryxi: Pushed fix.

packages/store/addon/-private/store-service.ts
```diff
@@ -134,20 +134,20 @@ export interface CreateRecordProperties {
   The store provides multiple ways to create new record objects. They have
   some subtle differences in their use which are detailed below:
 
-  [createRecord](../methods/createRecord?anchor=createRecord) is used for creating new
+  [createRecord](./classes/Store/methods?anchor=createRecord) is used for creating new
```
runspired, line 137: if `./classes/Store/` is necessary for createRecord then it would be necessary for push and pushPayload as well. Probably worth booting up the docs app locally to check.

- generate the api doc data https://github.com/ember-learn/ember-jsonapi-docs
- the app: https://github.com/ember-learn/ember-api-docs/blob/master/README.md

I forget the precise magic formula for getting these things working together locally.

rishiraj88: Thanks.

packages/store/addon/-private/store-service.ts
rishiraj88: 👍

MichalBryxi:
Hmm observation from today's experiments: The links will appear on different URLs, because of `index / methods / properties` tabs on the page. Then one will get two different sets of links:

| Tab | Link |
|-----|-----|
| `index` | `https://api.emberjs.com/ember-data/4.6/methods/createRecord?anchor=createRecord` |
| `methods` | `https://api.emberjs.com/ember-data/4.6/classes/methods/createRecord?anchor=createRecord` |
| `properties` | `https://api.emberjs.com/ember-data/4.6/classes/methods/createRecord?anchor=createRecord` |
| correct URL | `https://api.emberjs.com/ember-data/4.6/classes/Store/methods?anchor=createRecord` |

I think that this might be a challenge to create correct link irrespective to at which tab the user is looking?

runspired:
pretty sure this is no longer necessary as the docs build would error if the link were invalid - closing for now

PR #8215, opened 2022-10-14
fix: inspector type detection 2. attempt #8205

### LINKED ISSUE #8205: no models in ember inspector -> typesMapFor has no entries
currently ember inspector often does not show any models because typesMapFor does not have any entries.

https://github.com/emberjs/data/blob/master/packages/debug/addon/index.js#L85

ember data: 4.7.0

Was it always an initializer? I load most of my data in an initializer .

patricklx:
resolves #8205

### DIFF
--- a/packages/debug/addon/setup.js
+++ b/packages/debug/addon/setup.js
@@ -10,7 +10,6 @@ export function typesMapFor(store) {
   if (typesMap === undefined) {
     typesMap = new Map();
     StoreTypesMap.set(store, typesMap);
-    installHook(store, typesMap);
   }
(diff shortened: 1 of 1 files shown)


patricklx:
@runspired , maybe like this?

runspired (approved):

patricklx:
@runspired tests mostly passed. the error in CI / additional-scenarios (with-ember-fetch-no-jquery) seems unrelated

patricklx:
@runspired _createRecordData does not exist since 4.5.0

PR #8222, opened 2022-10-17
Invoke ember-auto-import analyzer

ef4:
## Type of PR

What kind of change is this?

- [ ] refactor
- [x] internal bugfix
- [ ] user-facing bugfix
- [ ] new feature
- [ ] deprecation
- [ ] documentation
- [ ] tests

## Notes for the release

No release notes required.

## Description

Because the ember-data packages opt out of the standard ember-cli preprocessor registry, they won't get analyzed correctly by ember-auto-import. Instead, they're forced to manually configure the analyzer. That's what this PR does.

The `ember-auto-import/babel-plugin` that was being added before was doing nothing. As documented, it exists to support dynamic imports, which ember-data doesn't use. It's unrelated to the import analysis.

I confirmed manually that these changes make ember-auto-import work in both the public and private parts of the packages. The proper way to add test coverage for this is to start using it, as was begun in the branch ember-auto-import-v2-addon-bug.

### DIFF
--- a/packages/private-build-infra/src/addon-build-config-for-data-package.js
+++ b/packages/private-build-infra/src/addon-build-config-for-data-package.js
@@ -125,7 +125,6 @@ function addonBuildConfigForDataPackage(PackageName) {
         return Array.isArray(plugin) ? plugin : [plugin];
       });
       plugins = plugins.concat(customPlugins.plugins);
-      plugins.push([require.resolve('ember-auto-import/babel-plugin')]);
 
       return {
         loose: true,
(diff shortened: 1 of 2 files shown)


runspired:
> Because the ember-data packages opt out of the standard ember-cli preprocessor registry, they won't get analyzed correctly by ember-auto-import. Instead, they're forced to manually configure the analyzer. That's what this PR does.

Interesting, so I'd tried adding just the analyzer babel plugin but I supposed that wasn't enough?

runspired (approved):
Overall looks good and makes sense, I'll dig into why this causes an asset-size change. It shouldn't unless auto-import is now actually importing something -- which should not currently be the case

PR #8301, opened 2022-11-12
assert missing crypto.randomUUID

luxferresum:
## Description

Add a little assert message when crypto.randomUUID does not exist

Sidenote: I was not able to test this, since I wasnt able to boot an app with the linked ember-data. :(

### DIFF
--- a/packages/store/addon/-private/caches/identifier-cache.ts
+++ b/packages/store/addon/-private/caches/identifier-cache.ts
@@ -41,6 +41,10 @@ if (macroCondition(getOwnConfig<{ polyfillUUID: boolean }>().polyfillUUID)) {
 }
 
 function uuidv4(): string {
+  assert(
+    'crypto.randomUUID needs to be avaliable. Some browsers incorrectly disallow it in insecure contexts. You maybe want to enable the polyfill: https://github.com/emberjs/data#randomuuid-polyfill',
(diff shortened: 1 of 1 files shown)


runspired:
> Sidenote: I was not able to test this, since I wasnt able to boot an app with the linked ember-data. :(

linking usually works better with pnpm than with yarn or npm but as long as you link with whatever you are using to install in the app it should be doable. I tend to just test in the test apps though to avoid these sorts of troubles.

packages/store/addon/-private/caches/identifier-cache.ts
```diff
@@ -41,6 +41,10 @@ if (macroCondition(getOwnConfig<{ polyfillUUID: boolean }>().polyfillUUID)) {
 }
 
 function uuidv4(): string {
+  assert(
```
runspired, line 44: note asserts only trigger in dev build, so if deploying to an insecure environment as mentioned in https://discord.com/channels/480462759797063690/486549196837486592/1040645066504556564 the error will still be simply that `randomUUID` is not a function.

runspired: (also note, even for internal tools, if you aren't deploying production builds with https you very much should be for a host of reasons)

runspired (approved):

PR #8305, opened 2022-11-14
[BUGFIX] Issue 7916 - handle invalid response if payload is null

### LINKED ISSUE #7916: Error if payload===null
Sorry not following the issue reporting instruction but this one is a simple error happening [here](https://github.com/emberjs/data/blob/76f1dcc74221c7d542021ce25df4a7be91d94128/packages/adapter/addon/rest.ts#L1011)
Our backend response payload=null (I know it shouldn't) and because Javascript consider the "null" as an object it throws error when trying resolve "payload.errors"
Indeed it passes through the guard that I think it's provided to prevent this kind of errors.

```
      return payload;
      return new InvalidError(typeof payload === 'object' && 'errors' in payload ? payload.errors : undefined);
    }
```

mani-rsg:
## Description

This fixes issue #7916. Handle an Invalid response though if the payload is `null`.  

Done the change and linked this to a sample Ember app and tested by using a mock API returning status `422` and response as `null`


mani-rsg:
HI @runspired , I tried to add test for this change in `tests/main/tests/integration/adapter/handle-response-test.js` as below:
But this didn't get failed when I removed the condition that I added in `packages/adapter/addon/rest.ts`. I was expecting a `typeError` when payload was passed `null`. I used this command to test `pnpm --filter main-test-app run test`.
```
test('handleResponse is called with correct parameters on null response with 422 status', async function (assert) {
    let handleResponseCalled = 0;

    this.server.get('/people', function () {
      return ['422', { 'Content-Type': 'application/json' }, null];
    });

    class TestAdapter extends JSONAPIAdapter {
      handleResponse(status, headers, payload, requestData) {
        console.log(payload, 'payload');
        handleResponseCalled++;

        return super.handleResponse(status, headers, payload, requestData);
      }
    }

    this.owner.register('adapter:application', TestAdapter);

    try {
      await this.store.findAll('person');
      assert.ok(false, 'promise should reject');
    } catch {
      assert.ok(true, 'promise rejected');
    }

    assert.strictEqual(handleResponseCalled, 1, 'handle response is called');
  });
```

runspired:
@mani-rsg let's still add that test but we will need to tweak it to ensure the error we are catching is the appropriate error. Since in both cases we throw, we just throw something unexpected in the payload=null case.

runspired:
@mani-rsg were you still interested in this?

anandfresh (approved):

runspired:
this was fixed in #10646

Issue #8318, opened 2022-11-24
A 'deleteObject' from a hasMany relation with an object not included in the collection, deletes another object within the collection


eodb:
### Reproduction

Following straight forward qunit test can be copied to quickly reproduce the issue:

```
import { run } from '@ember/runloop';
import Model, { attr, belongsTo, hasMany } from '@ember-data/model';
import { setupTest } from 'ember-qunit';
import { module, test } from 'qunit';

let store;

module('integration/relationship/deleteObject from hasMany relation', function (hooks) {
  setupTest(hooks);

  hooks.beforeEach(function () {
    class Job extends Model {
      @attr name;
      @hasMany('message', { async: true, inverse: 'job' }) messages;
    }

    class Message extends Model {
      @attr msg;
      @belongsTo('job', { async: true, inverse: 'messages' }) job;
    }

    store = this.owner.lookup('service:store');
    this.owner.register('model:job', Job);
    this.owner.register('model:message', Message);
  });

  test('deleteObject of hasMany relation', async function (assert) {
    let job;
    let msg1;
    let msg2;
    let msg3;

    run(() => {
      job = store.createRecord('job', { name: 'my job' });
    });
    assert.ok(job, 'job created');

    run(() => {
      msg1 = store.createRecord('message', { job: job, msg: '1st msg' });
      msg2 = store.createRecord('message', { job: job, msg: '2nd msg' });
      msg3 = store.createRecord('message', { msg: '3rd msg' }); // 3rd msg, not part of the job
    });

    await run(() => {
      return job.messages.then((msgs) => {
        assert.equal(msgs.length, 2, 'job has 2 messages');
        assert.ok(msgs.includes(msg1), 'msg1 included');
        assert.ok(msgs.includes(msg2), 'msg2 included');
        // try to delete an object that is not included in the collection
        assert.notOk(msgs.includes(msg3), 'msg3 is not included');
        msgs.removeObject(msg3);
      });
    });

    run(() => {
      assert.equal(
        job.messages.length,
        2,
        'job still has 2 messages after an attempt to delete an object that is not part of the collection'
      );
    });
  });
});
```

### Description

One expects that an attempt to delete an object from a hasMany relationship that is not included in the collection, should have no effect. Instead some other object from the collection seems to be deleted (see last failing assert: the resulting collection only contains one element anymore instead of the original two elements).

I tested this with an AsyncHasMany relation (async=true), but it also fails when the relation is defined as synchronous (async=false).

### Versions

[...]

PR #8323, opened 2022-11-28
fix: Only remove object(s) from record array if in collection #8318

rossketron:
## Description

Fixes issue #8318.
Changes:
- Only delete from record array if object exists in collection.
- Fixes issue with `splice` accepting the `-1` return value of `indexOf` as a negative index when object to remove was not found in collection.


runspired (approved):

tests/main/tests/integration/record-array-test.js
```diff
@@ -400,4 +400,67 @@ module('unit/record-array - RecordArray', function (hooks) {
...
+
+    const personNotInData = {
+      type: 'person',
+      id: '5',
+      attributes: {
+        name: 'Scumbag Ross',
+      },
+    };
+
+    let recordArray = store.peekAll('person');
```
runspired, line 439: EmberData has 3 kinds of managed record arrays (all of which now utilize `IdentifierArray`, the reason this test fails is because you chose one of the "immutable" ones :D I'll explain

`peekAll` (and by extension `findAll` which returns peekAll once it resolves) return what we refer to as a `LiveArray`, this is a managed record-array that reflects all records loaded for a given type in the store. Since this reflects cache state but cannot mutate cache state, we error if mutation methods are called (this is why this test fails).

`query` returns a managed record array containing a subset of records specific to that query. Historically we referred to this as an `AdapterPopulatedRecordArray` though that is no longer a great name. The distinction now is merely over whether the array reflects a specific collection or "all". In the codebase this is now referred to as a `Collection`. Currently you only gain access to one of these via a `query` but this is about to change (emberjs/rfcs#854 and emberjs/rfcs#860 collectively specify the mechanics of more generalized collections).

`hasMany` relationships are the final managed array (historically called a `ManyArray` currently called a `CollectionRelationship` and soon will just be a `MutableCollection`, and these *are* mutable and their mutations are sent as mutations to the cache. So that is what this test will need to do: test a hasMany relationship. It doesn't hugely matter whether we test `async` or `sync` hasMany relationships, though if testing `async` I would test both and then you'll have to write a `deprecatedTest` for the `async` version since async promise proxies are deprecated now.

rossketron: Thank you! I'll update to test both `async` and `sync` `hasMany` relationships and will include a `deprecatedTest` for `async`. Your explanation is greatly appreciated!

tests/main/tests/integration/record-array-test.js
```diff
@@ -400,4 +400,100 @@ module('unit/record-array - RecordArray', function (hooks) {
...
+
+    let scumbagsToRemove = [scumbagInRecordArray, scumbagNotInRecordArray];
+    recordArray.removeObjects(scumbagsToRemove);
+
+    didRemoveObject = recordArray.length === 1 && !recordArray.includes(scumbagInRecordArray);
+    assert.true(didRemoveObject, 'sync record array only removes objects in list that are found in collection');
+
+    recordArray.push(scumbagInRecordArray);
+    this.owner.unregister('model:tag');
+    this.owner.register('model:tag', AsyncTag);
```
runspired, line 476: we should do this as a separate test, unregister/register with a booted application isn't safe

PR #8324, opened 2022-11-28
add @ember-data/tracking to Advanced Installation

luxzeitlos:
## Description

It seems `@ember-data/tracking` is required now, so we add it to the readme under "Advanced Installation"


runspired:
we might want to make this an optional peer-dep and use macros to skip the path if not present

PR #8335, opened 2022-12-06
fix: <Cache>.changedAttrs should be empty if no local mutations exist

### LINKED ISSUE #8329: Model.changedAttributes not cleaned out after save
### Reproduction

I have a repro case in a simple ember application: https://github.com/timmorey/ember-data-changed-attrs.  You can see the failure by cloning and then running `npm install && npm run test`.

The application defines a single model and configures the application REST adapter/serializer.  ember-cli-mirage provides a mock backend to support the model.  The test case is coded as an acceptance test that will fail when it hits the issue, and succeed if `changedAttribute` behaves as expected.

See test case in https://github.com/timmorey/ember-data-changed-attrs/blob/main/tests/acceptance/model-changed-attributes-test.js.  A basic summary:

```
const thing = await store.findRecord('thing', '1');
thing.name = 'new name';
// thing.changedAttributes() now returns { name: ... }
await thing.save();
// thing.changedAttributes() still returns { name: ... }
```

I'm also going to try to take a stab at coding a failing test in ember-data itself, but I've never done that before...

### Description

The `Model.changedAttributes()` method continues to indicate that attributes have been changed after the changes to the model have been saved.  In previous versions of ember-data (< 4.7.0), this method would indicate that there are no changed attributes after a `save()`.

This is confounding some logic we have in another application, where we look at `changedAttributes()` in the adapter to determine how to save the changes.

### Versions

I cannot reproduce in 4.6.3, but can in 4.7.0.  Also validated that the issue is reproducible in 4.8.3 and 4.9.0-beta.3.

timmorey:
resolves #8329

### DIFF
--- a/packages/record-data/addon/-private/record-data.ts
+++ b/packages/record-data/addon/-private/record-data.ts
@@ -599,6 +599,7 @@ function setupRelationships(
 function patchLocalAttributes(cached: CachedResource): boolean {
   const { localAttrs, remoteAttrs, inflightAttrs, changes } = cached;
   if (!localAttrs) {
+    cached.changes = null;
     return false;
   }
   let hasAppliedPatch = false;
--- a/tests/main/tests/unit/model-test.js
+++ b/tests/main/tests/unit/model-test.js
@@ -1217,43 +1217,45 @@ module('unit/model - Model', function (hooks) {
       await cat.save();
     });
 
-    test('changedAttributes() reset after save', async function(assert) {
-      adapter.updateRecord = function(store, type, snapshot) {
+    test('changedAttributes() reset after save', async function (assert) {
+      adapter.updateRecord = function (store, type, snapshot) {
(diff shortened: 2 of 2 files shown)


runspired:
I think it was but yes, if not please open a PR

ptgamr:
@runspired Nah, it wasn't back ported to the LTS 4.8.7, still happening, this is a blocker for our app -  what's the processs of creating a back port? I can help with that

ptgamr:
Tried to use `patch-package` to have a quick fix, but it's not possible with the way `ember-data` organize its package.

runspired:
> Tried to use patch-package to have a quick fix, but it's not possible with the way ember-data organize its packages.

It's certainly possible, we don't do anything non-standard for our packages. Likely you're just noticing that the `ember-data` package simply re-exports all the actual packages?

PR #8395, opened 2023-01-13
Camila/fix without for deprecation

camilabrowne:


runspired:
we're working on getting 3.28 (an old lts) into shape for Square, so this patch is no longer needed.

PR #8399, opened 2023-01-18
fix: deprecated record-array rejectBy use optional value if supplied

mrloop:
## Description

If an optional value supplied as second arg to `rejectBy` it should be compared to the value retrieved using the `key`.

## Notes for the release

Fix regression in deprecated `rejectBy` behaviour when optional value to test against is used.

### DIFF
--- a/tests/main/tests/unit/record-arrays/record-array-test.js
+++ b/tests/main/tests/unit/record-arrays/record-array-test.js
@@ -142,6 +142,49 @@ module('unit/record-arrays/record-array - DS.RecordArray', function (hooks) {
   );
 
   deprecatedTest(
+    '#rejectBy',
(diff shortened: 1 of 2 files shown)


tests/main/tests/unit/record-arrays/record-array-test.js
```diff
@@ -141,6 +141,49 @@ module('unit/record-arrays/record-array - DS.RecordArray', function (hooks) {
     }
   );
 
+  deprecatedTest(
+    '#rejectBy',
+    { id: 'ember-data:deprecate-array-like', until: '5.0', count: 3 },
```
mrloop, line 146: not sure what value `count` should be.

runspired: count is number of expected times this deprecation will be seen

PR #8402, opened 2023-01-20
Fix ember data v4.9 for embroider & package.json version

### LINKED ISSUE #8396: [Bug] Ember data doens't work with embroider starting with 4.8.x
### Description

There is not anymore possible to use ember-data with embroider, starting with 4.8.x.

In v4.9.1 we get this error:
Error `Error while processing route: index Assertion Failed: Expected store.createRecordDataFor to be implemented but it wasn't`

https://github.com/emberjs/data/blob/f8acfc162fb126efe8e2603359438b551347de41/packages/store/addon/-private/store-service.ts#L2392-L2406

This is because the private-build-infra `HAS_RECORD_DATA_PACKAGE` says, that the package `@ember-data/record-data` is not present.

https://github.com/emberjs/data/blob/f8acfc162fb126efe8e2603359438b551347de41/packages/private-build-infra/addon/available-packages.ts#L5

### Fix
We can fix this problem by replacing the line with `HAS_RECORD_DATA_PACKAGE: '@ember-data/record-data/-private',`

### LINKED ISSUE #8401: [Bug] `package.json` has wrong version in 4.9.1
I have seen, that the `4.9.1` has in `package.json` as version `4.9.0-alpha-14`.

https://github.com/emberjs/data/blob/f8acfc162fb126efe8e2603359438b551347de41/package.json#L1-L8

I think it was forget to replace...

mkszepp:
fix #8396 & fix #8401


runspired:
the `master` branch is `canary` and always has an `alpha` tag.

PR #8403, opened 2023-01-23
Fix store.createRecordDataFor for embroider

mkszepp:
This should be fix the error `store.createRecordDataFor` for build with embroider


runspired:
I don't think we want to make a private module public ....

runspired:
closing in favor of using dependencySatisfies instead of moduleExists

PR #8410, opened 2023-02-17
fix: add better embroider test and fix embroider compat

mkszepp:
Add embroider try scenarios for tests


mkszepp:
@runspired maybe we get already with this tests the same error like explained in https://github.com/emberjs/data/issues/8396, otherwise we need to add some other tests, which should fail

runspired:
@mkszepp ended up adding a test which failed because we can't use @embroider/test-setup currently as it results in downstream build errors: https://github.com/glimmerjs/glimmer.js/issues/406

I think the solution here is to not use moduleExists (since no module does exist) and to use dependencySatisfies. This was failing because (as you noticed but I'm not sure quite communicated well enough in #8403 for me to understand why that was what you were doing) there is no module for moduleExists to find.

PR #8427, opened 2023-02-25
fix: embroider compat

### LINKED ISSUE #8396: [Bug] Ember data doens't work with embroider starting with 4.8.x

runspired:
resolves #8396

### DIFF
--- a/packages/private-build-infra/src/transforms/babel-plugin-convert-existence-checks-to-macros/index.js
+++ b/packages/private-build-infra/src/transforms/babel-plugin-convert-existence-checks-to-macros/index.js
@@ -1,5 +1,12 @@
 const { ImportUtil } = require('babel-import-util');
 
+function parentIsUnary(node) {
+  if (node.parent.type === 'UnaryExpression' && node.parent.operator === '!') {
+    return true;
+  }
+  return false;
+}
+
 module.exports = function (babel) {
   const { types: t } = babel;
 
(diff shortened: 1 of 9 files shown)


Techn1x:
This feels like it should be backported to 4.8 LTS, but I don't know how complicated that task would be

(as a result I'm skipping all the way to 4.12.. hitting other issues instead with that)

runspired:
It's too difficult to support 4.8 at this point, and I suspect this was one of dozens of PRs that would need backported for this fix to actually apply there. Sorry for the inconvenience :(

Techn1x:
That's okay :) totally understand.

PR #8487, opened 2023-03-19
feat: relationship fetching as legacy op

runspired:
### ⚠️ This change is potentially breaking. We don't really have an option. A short story.

- Ember overrides the (already not a native microtask) flush of RSVP.Promise to schedule an autorun / join an existing run's actions queue.
- Ember's override has a number of horrid consequences:
  - RSVP Promises that are resolved can be re-resolved synchronously
  - RSVP Promise chains with no other async mechanism can be run start-to-finish synchronously by wrapping the whole chain in its own `run`
  - resolving or rejecting an RSVP Promise always schedules a synchronous Render
  
This has historically limited EmberData in a number of ways:

- EmberData could not use `async/await` or native promises as it would interrupt the otherwise entirely RSVP Promise-chain (breaking 1 & 2)
- EmberData could get away with mutating cache-state within any link of the chain as the entire chain was guaranteed to complete before the next render.

We were always going to have to break something if we started moving the internals away from RSVP, it's been a question of "when" more than "whether". Unfortunately, the work in this PR exposed that "when" is "now".

First, a note: in order for RequestManager to support the RFC'd range of capabilities, it cannot build entirely over `async/await` anyway. We carefully manage Promise chains and theoretically could have built over RSVP.Promise (initially we did, this PR surfaced the larger problem to us and shifts it back to native promises).

What goes wrong is this, if two RSVP promises are in the chain and do not resolve within the same runloop, you get two renders. This is key to understand why the next point matters.

In the old paradigms, data-fetching was a bunch of spaghetti managed across a dozen different locations. There was no coordinated request pipeline. This meant that the final promise in the chain was both the promise the resolved to the UI *and* the promise that updated the cache and triggered property change notifications.

Refactoring this logic to support the RFC means that now the cache update happens several promises before the final promise in the chain. When using native microtasks that do not flush within a runloop that also sync-flushes render, this is safe. But, when using RSVP Promises, the state update occurs within a runloop and render immediately occurs.

In real world effect, this means that templates would infinitely re-render any asynchronous values, especially noticeable on reject (where the value EmberData hands back is less stable than records that maintain stable proxies) as the final promise which updates the object handed to render is not the promise which triggered the render. 

Basically: `render -> fetch -> cache update -> complete -> render` accidentally becomes `render -> fetch -> cache update+render -> fetch -> cache udpate + render -> fetch -> cache update + render` and so on.

What does this mean for your apps?

1. Well, some apps may find that continuing to use RSVP Promises shows no ill-effects. If you aren't triggering fetch via templates with async-relationship, aren't rendering promise proxies in templates, or never have API requests that fail then setting `window.Promise = RSVP.Promise` would seemingly maintain the old behavior and may even work for some set of safer async-from-template cases. So if this upgrade is difficult, you might consider temporarily doing that. If your apps do do these things, you're the reason this change has to happen now, else your app would blow up during render.

2. Likely most apps won't notice a thing, they probably will even see nicer performance because of the reduction in runloop usage and extra promise generation that this PR now brings. What *might* get affected is tests that depended on the sync-resolve capabilities of `run(rsvpPromiseChain)` which ... we were never going to be able to detect and deprecate and has been a liability ever since `async/await` came into common usage. Given the prevalence of async/await, ember-concurrency, and native promise usage in Ember apps today, there's even a chance this change goes by unnoticed. Which would be a relief. If it doesn't, reach out and I'll see how I can help mitigate.

3. Leaky tests will be a lot more noticeable. If a promise escapes the test bounds but resolves within the same microtask queue it generally will complete before any of the `afterEach` callbacks begin, allowing a small amount of un-awaited async to just-work unnoticed, most commonly triggered by actions, observers, side-effects within tests etc. This refactor to native promises and native promise chains almost certainly means that any unawaited unsettled ember-data state would trigger an error. Good news, easy to fix, add `await settled()` before the test completes if you don't have any other need to await this state more directly.

### Original Change Description

(4/4ths of) the literal final step to complete #6166 💚

This PR:
- [x] `Model.reload` as request op
- [x] `findBelongsTo` as request op
- [x] `findHasMany` as request op
- [x] remove _fetchManager temporary additions
- [x] cleanup util duplicates

To Go (for full RequestManager support): (not this PR)
- [-] enforce immutability in dev
- [-] handle error documents in cache
- [-] handle meta documents in cache
- [-] additional test coverage for Store.request, errors thrown during request
- [-] additional test coverage for cache.peek / cache.put / cache.peekRequest
- [-] update docs for RequestManager.useCache
- [-] implement new `Collection` mode to properly reflect `ResourceDocument`
- [-] ensure single-resource-document requests are notified when `data` changes
- [-] determine Cache "mutability" of `CollectionResourceDocument`
- [-] add tests for enhancing `context.request` in RequestManager before calling next

Other Notes:

- clarify intended Error handling mechanisms/propagation
- update RFC for RequestManager.useCache
- update RFC for `lifetimes.isSoftExpired/isHardExpired`
- update RFC for rename `data` => `content`


robbytx:
Additional question -- this PR is titled "relationship fetching as legacy op", which suggests to me that Ember Data (at some future point) will no longer auto-fetch relationships upon access.  Am I understanding this correctly?  Is there an RFC or other design doc that explains the reasoning and future state?  Sorry for my ignorance, I'm just now getting up to speed on the variety of changes occurring.

robbytx:
Thank you for taking the time to respond so thoroughly.  I was not aware of the RequestManager RFC until now, and that is certainly an ambitious scope to achieve, so I have a much better understanding of the motivation for these changes.  (I was previously under the impression that the motivation was primarily for modularity under the linked Project Trim.)  You've also made it clear to me why the transition to native promises is necessary/desirable now, so thank you for indulging my questioning on that.

I don't think I have any further questions on the PR itself.  If I can find some extra time, I will try to review the changes more thoroughly, although I'm sure you don't need to wait on me for that. :)

For my own understanding on this point:

> Native promises would result in the following: '1' 'A' '2' 'B'
> RSVP with the backburner flush results in the following: '1', '2', 'A' ('B' is dropped because 'A' won't have flushed yet)

When you say that 'B' is dropped, do you just mean that it is not included in the sync resolution?  It would still fire later, right?  Or is there something else going on that I'm not appreciating?

runspired:
> When you say that 'B' is dropped, do you just mean that it is not included in the sync resolution? It would still fire later, right? Or is there something else going on that I'm not appreciating?

No. The render is de-duped and so only one render occurs, in this case 'A' because it was the first scheduled. All state updates that have occurred when A occurs will be a part of that render. Effectively 'A' + 'B' but not double work.

PR #8499, opened 2023-03-26
chore: refactor model hook support to live in the model package

runspired:


packages/-ember-data/addon/-private/index.ts
```diff
@@ -3,17 +3,48 @@ import ArrayProxy from '@ember/array/proxy';
...
+import type { RecordInstance } from '@ember-data/types/q/record-instance';
 
 export class Store extends BaseStore {
+  declare _modelFactoryCache: Record<string, ModelFactory>;
   constructor(args: Record<string, unknown>) {
     super(args);
     this.requestManager = new RequestManager();
     this.requestManager.use([LegacyNetworkHandler, Fetch]);
     this.requestManager.useCache(CacheHandler);
+    this._modelFactoryCache = Object.create(null) as Record<string, ModelFactory>;
```
runspired, line 26: we should have the model package do this on-demand

packages/-ember-data/addon/-private/index.ts
```diff
@@ -3,17 +3,48 @@ import ArrayProxy from '@ember/array/proxy';
...
+  createCache(storeWrapper: CacheStoreWrapper) {
+    return new Cache(storeWrapper);
+  }
+
+  instantiateRecord(
+    this: ModelStore,
+    identifier: StableRecordIdentifier,
+    createRecordArgs: Record<string, unknown>
+  ): DSModel {
+    return instantiateRecord.call(this, identifier, createRecordArgs);
```
runspired, line 39: before shipping check that these can be installed as `instantiateRecord = instantiateRecord;` If not, may as well just pass store as first arg instead of binding this.

packages/store/package.json
```diff
@@ -36,17 +36,13 @@
   ],
   "peerDependencies": {
     "@ember-data/graph": "workspace:5.3.0-alpha.4",
-    "@ember-data/json-api": "workspace:5.3.0-alpha.4",
     "@ember-data/legacy-compat": "workspace:5.3.0-alpha.4",
     "@ember-data/model": "workspace:5.3.0-alpha.4",
```
runspired, line 40: this PR is complete when we can also remove this.

legacy-compat and graph will be handled in a separate PR

PR #8517, opened 2023-03-30
fix: ensure cache-handler works as expected

### LINKED ISSUE #8510: [Store] Documentation should tell something about store lifetimes
Sorry for flooding with issues :cry: But hopefully it will help hammering out the nasty details.

When requesting data through the store you will get the error
```
Error while processing route: index store.lifetimes is undefined calcShouldFetch
```
I found the function here: https://github.com/emberjs/data/blob/main/packages/store/src/-private/cache-handler.ts#L68 and in particular https://github.com/emberjs/data/blob/main/packages/store/src/-private/cache-handler.ts#L15

So I added in the `constructor` of my route
```javascript
constructor(...args) {
  super(...args);
  this.store.lifetimes = {
    isHardExpired() { return true; }
  }
}
```
Which solves the error but to me this basically disables the cache right? So maybe in the documentation the lifetimes of the `store` can be explained in more detail?

### LINKED ISSUE #8515: [Store] recordArrayManager._keyedArrays is undefined
This is on `alpha.14`.
After fetching the data it goes through the Cache but errors out on
```
Uncaught (in promise) TypeError: recordArrayManager._keyedArrays is undefined
```
Looking through the code it seems that `_keyedArray is not defined in the `constructor`. See https://github.com/emberjs/data/blob/15049beed153f1375b3eb2bfd9ba31da744b7e14/packages/store/src/-private/managers/record-array-manager.ts#L81

Manually adding `this._keyedArrays = new Map();` to the `constructor` fixes this problem but I don't know if this is the right way.

For me though I got around :smiley: I did a `this.store.findAll('model')` and got finally my data from the `store` :tada:.

### LINKED ISSUE #8516: [Store / Cache] The key from `cacheOptions` is not used
When I fetch data from my JSON-RPC backend, the url is not changing but stays the same. This will basically kill caching because the key used for caching is the url. So the latest response overrides the rest.
Thankfully there is an option to add my own key to the `cacheOptions` however this key is never used. Even the code seems to have some checks for it, the key is never used.

My first impression was https://github.com/emberjs/data/blob/15049beed153f1375b3eb2bfd9ba31da744b7e14/packages/store/src/-private/cache-handler.ts#L128 because there the `cacheOptions.key` isn't picked up. When I force it, further down the stream the key is changed back to the url.

But as said in https://github.com/emberjs/data/issues/8507#issuecomment-1487916795 there are a few things in the way which have to be fixed. This is a reminder :reminder_ribbon: .

BTW, the key I generate is based on the `method` and `params`. Both together create a unique string from which I create a MD5 hash. That's the key I also send to the backend to uniquely identify the request.

runspired:
- Adds additional testing for the CacheHandler. resolves #8510 
- Ensures `cacheOptions.key` is used before `url` (needs test). resolves #8516 
- resolves #8515 by ensuring keyedArrays is setup properly (needs test)

### DIFF
--- a/tests/main/app/cache-handler/store-package-setup-test.ts
+++ b/tests/main/app/cache-handler/store-package-setup-test.ts
@@ -1,0 +1,70 @@
+import { service } from '@ember/service';
+
+import { module, test } from 'qunit';
+
+import { setupTest } from 'ember-qunit';
+
+import Cache from '@ember-data/json-api';
+import { LegacyNetworkHandler } from '@ember-data/legacy-compat';
+import RequestManager from '@ember-data/request';
+import Fetch from '@ember-data/request/fetch';
+import Store, { CacheHandler } from '@ember-data/store';
+import type { SingleResourceDataDocument } from '@ember-data/types/cache/document';
+import type { CacheStoreWrapper } from '@ember-data/types/q/cache-store-wrapper';
+
+module('Store | CacheHandler - setup with @ember-data/store', function (hooks) {
+  setupTest(hooks);
+
+  test('When using @ember-data/store, we can use @ember-data/request as a service', async function (assert) {
+    const { owner } = this;
+
+    class RequestManagerService extends RequestManager {
+      constructor() {
+        super(...arguments);
+        this.use([LegacyNetworkHandler, Fetch]);
+        this.useCache(CacheHandler);
+      }
+    }
+
+    class TestStore extends Store {
+      @service('request') declare requestManager: RequestManager;
+
+      createCache(wrapper: CacheStoreWrapper) {
+        return new Cache(wrapper);
+      }
+    }
+
+    owner.register('service:store', TestStore);
+    owner.register('service:request', RequestManagerService);
+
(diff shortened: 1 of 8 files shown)


tests/main/app/cache-handler/store-package-setup-test.ts
```diff
@@ -0,0 +1,70 @@
...
+    assert.deepEqual(
+      userDocument.content.meta,
+      {
+        expiration: 120000,
+      },
+      'we get access to the document meta'
+    );
+
+    const record = store.peekRecord(userDocument.content.data!);
+    assert.strictEqual(record?.name, 'Chris Thoburn');
```
runspired, line 68: This isn't passing yet, which indicates that the store is not properly inserting the resource from the document yet 🤔

Jopie01:
Thanks :tada: ! I upgraded to `alpha.15` and I'm able to drop the `constructor` from both the `store` and `application` route. No need to set `store.lifetimes` and the store is now using the general `requestManager` through a service.

Take a look at my demo repo to see the differences.

Issue #8519, opened 2023-03-30
[Cache] attributes of records got missing


Jopie01:
With some changes and upgrade to `alpha.15` I accidentally bypassed the `Cache` so I added the `CacheHandler` back into my `requestManager` service and now the `cache` is hit again. Because the response is JSON, I extended the `Cache` and converted the JSON into JSONAPI data so that the default JSONAPI cache consumes the data.

```javascript
import Cache from '@ember-data/json-api';

class RPCCache extends Cache {
  constructor(...args) {
    super(...args);
  }

  // convert the JSON response into a JSONAPI response
  put(document) {
    let res = [];
    const method = document.request.data.method.split('.');
    document.content.forEach((item) => {
      res.push({
        id: item.id,
        type: method.slice(1, method.length - 1).join('.'),
        attributes: item,
      });
    });
    document.content.data = res;
    return super.put(...arguments);
  }
}
```
```hbs
{{#each @model as |menu|}}
  {{menu.attributes.icon}} {{menu.attributes.name}}<br/>
{{/each}}
```
This worked prior to `alpha.15`. In my template I had to add the `attributes` key but that's ok. I also tested it with converting the response directly after the `fetch` but it doesn't make a difference.

BTW, also the `instantiateRecord` isn't called when doing a `store.findAll(model)` or `store.query(model, options)` so some wire seems to be broken.

runspired:
can you push this to the repo? I'll want to debug it

Jopie01:
> can you push this to the repo? I'll want to debug it

Sorry for not mention it. I've already done that.

runspired:
Currently, the CacheHandler only instantiates records in the response documents for `op` codes that it knows about, I think this check needs to be loosened to being only for whether a hidden symbol is present, so that hydration is optional. What happened here is that while fixing some of the other issues you've hit I scoped down which requests the cache-handler attempts to "hydrate", and so bumping the version you now get back just the identifiers. Fix inbound.

I suspect there's also some more documentation to add here around where/how to do what kinds of work.

runspired:
> Sorry for not mention it. I've already done that.

No worries, figured that out

PR #8528, opened 2023-04-01
feat: enable request cache without hydration

runspired:
work for #8519

- ensures any opcode can be cached and hydrated
- ensures non-store issued requests can be cached if desired (by supplying the store)


Jopie01:
Thanks for the hard work! It works like a charm now.

PR #8576, opened 2023-04-18
LegacyHandler providing frozen options to Adapters

kiosion:
## Description
This PR addresses an issue with the LegacyHandler in 4.12 - The StoreRequestContext obj passed to legacy Adapters is frozen, causing issues when modifying request params, i.e., modifying the `query` param within an Adapter's `query` or `urlForQuery` methods. I've added a failing test case as a reproduction of how I ran into this issue.


amk221:
Thanks. I'm also experiencing this from:

```javascript
urlForQuery(query) {
  let url = super.urlForQuery(...arguments);

  if (query.something) {
    url = `${url}/${query.something}`;
    delete query.something;
  }

  return url;
}
```

```javascript
this.store.query('foo', {
  something: 'abc'
});
```

The actual URL used will be `/foo/abc?something=abc`

But it should be `/foo/abc`

PR #8664, opened 2023-07-03
Clearer instructions for polyfillUUID

Techn1x:
## Description

Clearer instructions for polyfillUUID.

The need to install embroider/macros has tripped me up at least 3 times now (When I go to update a repo's ember-data version, I keep forgetting what I did to make it work on a previous repo months prior). Not to mention, the first time I encountered it it took me a long while to figure out why the config entry wasn't doing anything.

### DIFF
--- a/README.md
+++ b/README.md
@@ -122,6 +122,7 @@ let app = new EmberApp(defaults, {
 the necessary feature if your browser support or deployment environment demands it. To
 activate this polyfill;
 * install `@embroider/macros` if not already installed
+* install `@ember-data/store` if not already installed
 * add this to your ember-cli-build.js file:
 
 ```ts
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -120,7 +120,10 @@ let app = new EmberApp(defaults, {
 
 *Ember***Data** uses `UUID V4` by default to generate identifiers for new data created on the client. Identifier generation is configurable, but we also for convenience will polyfill
 the necessary feature if your browser support or deployment environment demands it. To
-activate this polyfill:
+activate this polyfill;
+* install `@embroider/macros` if not already installed
+* install `@ember-data/store` if not already installed
+* add this to your ember-cli-build.js file:
```
runspired, line 126: we should align the config to the others so that we don't have to make folks do this. For all the other configs we just copy-forward into the appropriate packages from the `emberData` config, including setting up macros configs when needed.

Techn1x: is that something likely to happen soon? Is it worth not having clear instructions until then?

runspired: Open an issue its something I'll get done for 4.12.x + 5.2+

Techn1x: Thanks a bunch - done https://github.com/emberjs/data/issues/8675

PR #8713, opened 2023-07-14
Notify on length when notifying that many-array has changed

richgt:
Port of https://github.com/emberjs/data/pull/8413 to 4.4


runspired:
not sure whats up with the try scenarios etc. but I'm pretty sure this fix was in 3.28 too so going to accept.

PR #8717, opened 2023-07-17
Switch from local and @types/ember types to ember-source types

BradBarnich:
## Description

It seemed like a first step to ember-data shipping types was to use the ember-source types directly. This was a helpful to me when I was updating the @types/ember-data for ember-source types. Even if types are still shipped separately, this would make it easier to harmonize the shipped types to the internal types.

## Notes for the release

not sure, seems like it is only visible internally

^ note from @runspired: this is correct. We intentionally do not ship types though this PR likely means we could ship types for @ember-data/tracking @ember-data/request @ember-data/request-utils @ember-data/rest and @ember-data/active-record starting as soon as 5.2. Would require a small cleanup pass.


packages/store/src/-private/store-service.ts
```diff
@@ -238,7 +236,7 @@ class Store extends EmberObject {
     @private
   */
   constructor(createArgs?: Record<string, unknown>) {
-    super(createArgs);
+    super();
```
runspired, line 239: I believe as long as we extend EmberObject we need to super the args to ensure services are available within `init` / after `super` should someone be extending the store and relying on that.

packages/model/src/-private/errors.ts
```diff
@@ -1,5 +1,4 @@
-import { A } from '@ember/array';
-import type NativeArray from '@ember/array/-private/native-array';
+import { A, NativeArray } from '@ember/array';
```
runspired, line 1:
```suggestion
import { A, type NativeArray } from '@ember/array';
```

packages/model/src/-private/errors.ts
```diff
@@ -13,7 +12,7 @@ type ValidationError = {
 /**
   @module @ember-data/model
 */
-interface ArrayProxyWithCustomOverrides<T, M = T> extends Omit<ArrayProxy<T, M>, 'clear' | 'content'> {
+interface ArrayProxyWithCustomOverrides<T> extends Omit<ArrayProxy<T>, 'clear' | 'content'> {
```
runspired, line 15: I believe we had a types override for this. The reason we do this is that the existing types for array proxy are incorrect: they assume the content type is the same as the materialized type, when in fact ArrayProxy typically transforms content by running an operation for each member of content on access.

runspired: The above said, we *might* not need this anymore given we don't actually use ArrayProxy anywhere but Errors at this point. Since it's not throwing a bunch of tsc errors that we've silenced (from what I can tell) lets go with the "delete shit" approach for now, just wanted to add context.

runspired (commented):
I'd say this is 99% mergeable 💜. left a few comments to chat through first.

packages/store/src/-private/store-service.ts
BradBarnich: this is the error, I wasn't sure what to do.

```
packages/store/src/-private/store-service.ts(239,11): error TS2345: Argument of type 'Record<string, unknown> | undefined' is not assignable to parameter of type 'Owner | undefined'.
  Type 'Record<string, unknown>' is missing the following properties from type 'Owner': register, lookup, factoryFor
```

`super(...arguments)` passes tsc but I don't know how correct that is

packages/model/src/-private/legacy-relationships-support.ts
```diff
@@ -122,6 +122,7 @@ export class LegacySupport {
     relationship.state.shouldForceReload = true;
     let promise = this._findBelongsTo(key, resource, relationship, options);
     if (this._relationshipProxyCache[key]) {
+      // @ts-expect-error
       return this._updatePromiseProxyFor('belongsTo', key, { promise });
```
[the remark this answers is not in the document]
BradBarnich:
```
packages/model/src/-private/legacy-relationships-support.ts(125,7): error TS2741: Property '[Symbol.toStringTag]' is missing in type 'PromiseBelongsTo' but required in type 'Promise<RecordInstance | null>'.
```

tests/main/tests/integration/model-errors-test.ts
```diff
@@ -45,6 +44,7 @@ module('integration/model.errors', function (hooks) {
     let { owner } = this;
 
     owner.register('model:tag', Tag);
+    // @ts-expect-error
```
BradBarnich, line 47:
```
tests/main/tests/integration/model-errors-test.ts(48,65): error TS2345: Argument of type 'import("/Users/brad/Code/data/node_modules/.pnpm/ember-cli-htmlbars@6.2.0/node_modules/ember-cli-htmlbars/lib/index").TemplateFactory' is not assignable to parameter of type 'import("/Users/brad/Code/data/node_modules/.pnpm/@glimmer+interfaces@0.84.2/node_modules/@glimmer/interfaces/dist/types/lib/template").TemplateFactory'.
  Type 'TemplateFactory' provides no match for the signature '(owner?: object | undefined): Template'.
```

packages/store/src/-private/store-service.ts
runspired: Ember's actual API here is that owner is key'd onto the actual create args, owner itself is not the first argument. I think I remember @chriskrycho kinda fudged that for glimmer components for various reasons to keep it simpler (that or glimmer components take a different code path than everything else), but if the types are acting like this should be owner the types are wrong.

See implementation: https://github.com/emberjs/ember.js/blob/03e2e64166e39b2e3e3cd9661418830105344570/packages/%40ember/-internals/container/lib/container.ts#L527-L563

runspired: If Ember's service types are wrong I suspect we should just @ts-expect-error that line, supering with the createArgs, and move on for now.

runspired, line 239:
```suggestion
    // @ts-expect-error
    super(createArgs);
```

esbanarango:
What would be the next step once this PR is merged? 

Is the end goal here to port the types of `@ember-data/*` into the actual releases of `@ember-data`, isn't it? I would like to help if more hands are needed.

runspired:
@esbanarango I *think* the next step is to port the DT types into the project, putting them into their respective packages within a folder named something like `unstable-legacy-types` or in `<root>/unstable-legacy-types` packages. We'll want to update the tsconfig to somehow ignore them (and give them their own tsconfig) and figure out a publishing strategy.

Issue #8765, opened 2023-08-03
Friendlier inverse error


MichalBryxi:
### Reproduction

- Will provide PR.

#### Unexpected
When I declare `belongsTo` relationship without specifying `inverse` option ember-data gives me following error:

```
Error: Assertion Failed: Expected belongsTo options.inverse to be either null or the string type of the related resource.
```

The error comes [from here](https://github.com/emberjs/data/blob/main/packages/model/src/-private/belongs-to.js#L109).

This is problematic because it lacks the basic information that I will need to action upon this:
1. What model class was the issue found in?
2. What relationship is it referring to?

#### Expected
- Ideally `ember-data` points the user nicely to respective lines in code that they need to adjust.
- But for practical reasons giving at least model class & relationship definition should be enough.

### Versions

```cli
pnpm list ember-source && pnpm list ember-cli && pnpm list --pattern ember-data
Legend: production dependency, optional only, dev only

frontend@0.0.0 /Users/michal/quick-e-vote.cz/app/ui/frontend

devDependencies:
ember-source 5.1.1
Legend: production dependency, optional only, dev only

frontend@0.0.0 /Users/michal/quick-e-vote.cz/app/ui/frontend

devDependencies:
ember-cli 5.1.0
 ERROR  Unknown option: 'pattern'
Did you mean 'test-pattern'? Use "--config.unknown=value" to force an unknown option.
For help, run: pnpm help list
```

PR #8766, opened 2023-08-03
fix: More context for error message in belongsTo 

MichalBryxi:
## Description

More context for error message in belongsTo when missing inverse or async attribute
Fixes: #8765 

## Notes for the release

More context for error message in belongsTo when missing inverse or async attribute

### DIFF
--- a/packages/model/src/-private/belongs-to.js
+++ b/packages/model/src/-private/belongs-to.js
@@ -104,9 +104,12 @@ function belongsTo(modelName, options) {
   let opts = options;
   let userEnteredModelName = modelName;
 
-  assert(`Expected belongsTo options.async to be a boolean`, opts && typeof opts.async === 'boolean');
   assert(
-    `Expected belongsTo options.inverse to be either null or the string type of the related resource.`,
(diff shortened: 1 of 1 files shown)


MichalBryxi:
Wait, the PR is wrong. The case for me was complaining about:

```js
export default class PunchcardModel extends Model {
  @belongsTo('delegate', { async: false }) delegate;
}
```

And it was ok after I changed it to:

```js
export default class PunchcardModel extends Model {
  @belongsTo('delegate', { async: false, inverse: 'punchcards' }) delegate;
}
```

So it's not really about the model, but about the relationship defined. I can't see how to access the information about said model though.

MichalBryxi:
Ok, force pushed an update. I think it might be a good idea to show the most likely syntax the user used so that it's somewhat easier for them to find respective source code. Not sure if there are some traps with such approach.

runspired:
> I can't see how to access the information about said model though.

because these are computed properties under the hood still (vs decorators) we have no ability to access schema information about the host. Unfortunately as long as we continue to support classic (`Model.extend({ foo: attr() })`) we have to be based on computed :(

Issue #8262, opened 2022-10-27
[Ember Data 4.7] Error: Assertion Failed: Expected to receive a stable Identifier to subscribe to


mkszepp:
### Reproduction

1. Use a stored ember data model, which was returned from server
2. add in a hasMany (which has already some stored models) a new ember model with `createRecord` without saving
3. remove added model with `destroyRecord`
4. `Error: Assertion Failed: Expected to receive a stable Identifier to subscribe to` by accessing the function `some`, `filter`....

The problem doens't appear when i use in step 3 `rollbackAttributes` instead of `destroyRecord`/`deleteRecord`

### Description

For better understanding i have created a new clean ember project with a example (branch identifier-subscribe).
Example repo: https://github.com/mkszepp/ember-stable-identifier/tree/identifier-subscribe

So i have check the difference between `rollbackAttributes` and `destroyRecord` i found a possible reason.

The difference is, that the `unloadRecord` calls immidiatly the `unloadRecord` while `deleteRecord` in the next run loop.

https://github.com/emberjs/data/blob/c165e7efd062e186d53bb4ed689116c1ddb9d836/packages/model/addon/-private/model.js#L848-L860

https://github.com/emberjs/data/blob/c165e7efd062e186d53bb4ed689116c1ddb9d836/packages/store/addon/-private/store-service.ts#L632-L649

Is it possible, that the reported bug is a timing problem?

An other problem is, when i create a new record and delete it (with deleteRecord) the model stays in the store and the states (isNew, isDeleted...) are false and dirtyType is empty string. Maybe those two problems are interconnected

### Versions

Ember Js: 4.7.0+
Ember Data: 4.7.3

runspired:
I suspect you haven't diagnosed what has gone wrong here quite right yet, the reason it appears there are separate run-loops is because we immediately fully dismantle everything for new records when any of `unloadRecord/deleteRecord/rollbackAttributes/destroyRecord` are called. Any one of these when called should fully complete the process.

runspired:
I'm unable to reproduce this using the repo you provided

mkszepp:
@runspired i'm using npm v14

The error occures, when i access to a deleted model, like in the provided repo.

https://github.com/mkszepp/ember-stable-identifier/blob/241d3f25559f98068a465c1c25a565af90f05af6/app/controllers/application.js#L61-L66

runspired:
[...]

PR #8791, opened 2023-08-23
fix: clear relationships properly when unloading new records

Windvis:
## Description

Unloading a new record that is related to another record does not remove it from the relationships. This then causes issues if the other record also is unloaded or if code expects the relationships to no longer include the unloaded record. This adds a failing test case.

This behavior seems to have changed in EmberData v4.7+, older versions seemingly work as expected.

Related Discord thread: https://discord.com/channels/480462759797063690/1143193491250368583
Similar issue: https://github.com/emberjs/data/issues/8262

### DIFF
--- a/tests/main/tests/integration/records/new-record-unload-test.js
+++ b/tests/main/tests/integration/records/new-record-unload-test.js
@@ -93,6 +93,12 @@ module('Integration | Records | New Record Unload', function (hooks) {
     assert.false(Matt.isDestroyed, 'record is not yet destroyed');
     assert.true(Matt.isDestroying, 'record is destroying');
     assert.strictEqual(people.length, 0, 'precond - no person left in the store');
+
+    await settled();
+
+    assert.true(Matt.isDestroyed, 'record is destroyed');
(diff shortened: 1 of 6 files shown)


tests/main/tests/integration/records/new-record-unload-test.js
```diff
@@ -60,6 +60,41 @@ module('Integration | Records | New Record Unload', function (hooks) {
...
+
+    assert.true(Pat.isDestroyed, 'record is destroyed');
+    assert.true(Pat.isDestroying, 'record is destroying');
+    assert.strictEqual(friends.length, 0, 'Matt has no friends');
+    assert.strictEqual(people.length, 1, 'precond - one person left in the store');
+
+    Matt.rollbackAttributes();
+    assert.false(Matt.isDestroyed, 'record is not yet destroyed');
+    assert.true(Matt.isDestroying, 'record is destroying');
+    assert.strictEqual(people.length, 0, 'precond - no person left in the store');
```
runspired, line 95: let's `await settled()` and do the final assertions here too, I'll get started on a fix.

PR #8825, opened 2023-09-04
feat: more docs for builders

runspired:
Documents `findRecord` and `query` for ActiveModel/Rest/JSON:API as well as their respective `/request` sub packages

### DIFF
--- a/packages/json-api/src/-private/builders/query.ts
+++ b/packages/json-api/src/-private/builders/query.ts
@@ -1,3 +1,6 @@
+/**
+ * @module @ember-data/json-api/request
+ */
 import { pluralize } from 'ember-inflector';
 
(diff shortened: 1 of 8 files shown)


packages/json-api/src/-private/builders/query.ts
```diff
@@ -1,10 +1,63 @@
...
+ * **With Query Params**
+ *
+ * ```ts
+ * import { query } from '@ember-data/json-api/request';
+ *
+ * const options = query('person', { include: ['pets', 'friends'] });
+ * const data = await store.request(options);
+ * ```
+ *
+ * **Supplying Options to Modify the Request Behavior**
```
Baltazore, line 32: @runspired do you think we should add similar section of documentation to `findRecord`  builder?

runspired: yep!

PR #8829, opened 2023-09-04
chore: eliminate dead build code

runspired:
EmberData moved to a new strategy that uses rollup pre-publish some time ago. This removes the vestigial code (that no longer was in use at all) so folks like @mansona stop thinking we still use it 😅 

TODO

- simplify buildBabelOptions
- determine if cacheKeyForTree is still required for engines


packages/private-build-infra/src/addon-build-config-for-data-package.js
```diff
@@ -54,69 +42,15 @@ function addonBuildConfigForDataPackage(pkg) {
...
-          );
-          throw message.message;
-        }
-      }
-      next(message);
-    },
-
     shouldIncludeChildAddon(addon) {
-      if (addon.name.startsWith('@ember-data')) {
+      if (addon.name.startsWith('@ember-data/')) {
```
runspired, line 53: note this is not the broccoli node issue (see the `@`) but since it could manifest in a similar incorrect check should someone choose a github/npm org starting with @ember-data* we add the `/`

PR #8845, opened 2023-09-05
feat: SchemaRecord, never write a Model again

runspired:
image


runspired:
everything in this PR as extracted over time

PR #8849, opened 2023-09-06
feat: docs, tests and fixes for create/update/deleteRecord builders

Baltazore:
## Documentation for rest of the builder functions

This PR aims to cover create-record / update-record / delete-record documentation

Should be added to json-api / rest / active-record packages


packages/json-api/src/-private/builders/save-record.ts
```diff
@@ -52,6 +52,48 @@ export function deleteRecord(record: unknown, options: ConstrainedRequestOptions
...
+/**
+ * Builds request options to create new record for resources,
+ * configured for the url, method and header expectations of most JSON:API APIs.
+ *
+ * **Basic Usage**
+ *
+ * ```ts
+ * import { createRecord } from '@ember-data/json-api/request';
+ *
+ * const person = this.store.createRecord('person', { name: 'Ted' });
```
Baltazore, line 64: @runspired am I correct here that this is how new record will be created? 

It seems like I am missing something here... kinda feels not right with `this.store.createRecord` and `create-record` helper

Baltazore: confirmed on ember-data meeting

packages/json-api/src/-private/builders/save-record.ts
```diff
@@ -52,10 +94,53 @@ export function deleteRecord(record: unknown, options: ConstrainedRequestOptions
...
+ * @for @ember-data/json-api/request
+ * @param record
+ * @param options
+ */
 export function createRecord(record: unknown, options: ConstrainedRequestOptions = {}): CreateRequestOptions {
   const identifier = recordIdentifierFor(record);
   assert(`Expected to be given a record instance`, identifier);
-  assert(`Cannot delete a record that does not have an associated type and id.`, isExisting(identifier));
+  // TODO: seems to be wrong assert here, at least message is confusing
+  // assert(`Cannot delete a record that does not have an associated type and id.`, isExisting(identifier));
```
leepfrog, line 143: +1 to wrong assertion, you should be able to create a record without an identifier.

runspired: @leepfrog not exactly. You do need an identifier, the identifier does not need an `id`.

leepfrog: Whoops!  Yes, agree!

tests/builders/tests/test-helper.js
```diff
@@ -6,13 +8,18 @@ import assertAllDeprecations from '@ember-data/unpublished-test-infra/test-suppo
...
+import config from '../config/environment';
+
 // Handle testing feature flags
 if (QUnit.urlParams.enableoptionalfeatures) {
   window.EmberDataENV = { ENABLE_OPTIONAL_FEATURES: true };
 }
 
 configureAsserts();
 
+setApplication(Application.create(config.APP));
```
runspired, line 21: in follow up we should eliminate this and the new app files in favor of in-test-file registration

runspired (approved):

PR #8868, opened 2023-09-11
Add tests for filter-empty request util

Baltazore:
@runspired small quick one PR here, please let me know if there is some cases I missed. Please also check if I put this file in correct place. 

there is 2 commented out fields, it fails `SerializablePrimitive` type, so I cannot even put it there... I am not sure if those should be even tested.

### DIFF
--- a/packages/request-utils/src/index.ts
+++ b/packages/request-utils/src/index.ts
@@ -350,8 +350,8 @@ export function filterEmpty(source: Record<string, Serializable>): Record<string
   const result: Record<string, Serializable> = {};
   for (const key in source) {
     const value = source[key];
-    // Allow falsy values except for undefined and null
-    if (value !== undefined && value !== null) {
+    // Allow `0` and `false` but filter falsy values that indicate "empty"
+    if (value !== undefined && value !== null && value !== '') {
       if (!Array.isArray(value) || value.length > 0) {
         result[key] = source[key];
       }
--- a/packages/request-utils/src/index.ts
+++ b/packages/request-utils/src/index.ts
@@ -350,7 +350,8 @@ export function filterEmpty(source: Record<string, Serializable>): Record<string
   const result: Record<string, Serializable> = {};
   for (const key in source) {
     const value = source[key];
-    if (value) {
+    // Allow falsy values except for undefined and null
+    if (value !== undefined && value !== null) {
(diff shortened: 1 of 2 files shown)


tests/builders/tests/unit/filter-empty-test.ts
```diff
@@ -0,0 +1,28 @@
+import { module, test } from 'qunit';
+
+import { filterEmpty } from '@ember-data/request-utils';
+
+module('filterEmpty', function () {
```
runspired, line 5: this is probably more useful filtering only `undefined` `null` and `[]`. `false` and `0` tend to have real meaning.
`[[]]` and similar is interesting. I think we leave those, not sure how those even serialize today would have to check

packages/request-utils/src/index.ts
```diff
@@ -350,7 +350,8 @@ export function filterEmpty(source: Record<string, Serializable>): Record<string
   const result: Record<string, Serializable> = {};
   for (const key in source) {
     const value = source[key];
-    if (value) {
+    // Allow falsy values except for undefined and null
+    if (value !== undefined && value !== null) {
```
runspired, line 354:
```suggestion
    if (value !== undefined && value !== null && value !== '') {
```

packages/request-utils/src/index.ts
```diff
@@ -350,7 +350,8 @@ export function filterEmpty(source: Record<string, Serializable>): Record<string
   const result: Record<string, Serializable> = {};
   for (const key in source) {
     const value = source[key];
-    if (value) {
+    // Allow falsy values except for undefined and null
```
runspired, line 353:
```suggestion
    // Allow `0` and `false` but filter falsy values that indicate "empty"
```

tests/builders/tests/unit/filter-empty-test.ts
```diff
@@ -0,0 +1,28 @@
...
+  });
+
+  test('it returns an object with truthy values and meaningful falsy values like `false` and `0`', function (assert) {
+    assert.deepEqual(
+      filterEmpty({
+        foo: 'bar',
+        baz: null,
+        zero: 0,
+        booleanFalse: false,
+        emptyArray: [],
```
runspired, line 17:
```suggestion
        emptyString: '',
        emptyArray: [],
```

PR #8878, opened 2023-09-18
test: add basic test for Fetch handler

runspired:
Adds `@ember-data/mock-server`, a private (for now) mock-server based on real HTTP requests.

The mock server can integrate with any app's test suite via an additional handler and a simple beforeEach/afterEach setup.

The mock server has the ability to both record and replay. Once a recording exists, replaying is extremely cheap due to http/2, careful mitigation of CORS pre-flighting, and all requests having been pre-compressed with brotli at 11.

Currently all mocking is done within the app and posted to the server.


packages/active-record/src/-private/builders/find-record.ts
```diff
@@ -99,7 +99,7 @@ export function findRecord(
 
   const url = buildBaseURL(urlOptions);
   const headers = new Headers();
-  headers.append('Content-Type', 'application/json; charset=utf-8');
+  headers.append('Accept', 'application/json; charset=utf-8');
```
runspired, line 102: I removed `Content-Type` from the builders because we want the out-of-the-box CORS experience to be great, and CORS only allows `Content-Type` to be `application/x-www-form-urlencoded` `multipart/form-data` or `text/plain` before it will opt for a preflight request.

Issue #8924, opened 2023-09-26
[4.12.3] Bug: Loading / unloading the same resource multiple times results in lingering reference in cache


bk-pb:
### Reproduction

[Diff from ember-cli blueprint](https://github.com/bk-pb/data-multiple-unload/commit/7077897b584560ca6a7784ae5060ab378d2e8dde)

```
git clone git@github.com:bk-pb/data-multiple-unload.git
cd data-multiple-unload
npm i
ember s
```

**NOTE:** Do not open **Ember Inspector** before receiving the error. Having it open on the **Data** tab during reproduction seems to hide the issue.

1. Open http://localhost:4200/home
2. Navigate to **Company List**
3. Notice that `@model.length` is `2`
4. Navigate to **Home**
5. Navigate to **Company List** again
6. Notice that `@model.length` is still `2`
7.  Navigate to **Home**
8. Navigate to **Company List** again
9. Notice that `@model.length` is now `4` and an error is thrown:
```
index.js:153 Uncaught (in promise) Error: Assertion Failed: Expected to receive a stable Identifier to subscribe to
    at assert (index.js:153:1)
    at NotificationManager.subscribe (index-36c7ac56.js:3571:1)
    at new RecordState (has-many-c1bfdebf.js:3242:1)
    at CompanyModel.init [as _super] (has-many-c1bfdebf.js:3626:1)
    at CompanyModel.init (has-many-c1bfdebf.js:5503:1)
    at CompanyModel.superWrapper (index.js:366:1)
    at initialize (core.js:74:1)
    at CompanyModel.create (core.js:398:1)
    at Store.instantiateRecord (index-36c7ac56.js:5529:1)
    at InstanceCache.getRecord (index-36c7ac56.js:2953:1)
```

### Description

Loading / unloading the same resource (i.e. record with the same `type` and `id`) multiple times results in lingering record reference in cache, which triggers a missing stable identifier error on any subsequent peeks.

Upgrading from `3.24.2` to `4.12.3`. Quite a jump, but it feels like this use case is simple enough that it shouldn't be blowing up like this.

### Versions

```cli
Ember: 4.12.3
Ember Data: 4.12.3
```

runspired:
Why are you unloading multiple times? That's not something that is explicitly supported.

PR #8926, opened 2023-09-27
fix: unloadAll should cleanup live-array membership

runspired:
test for #8924 
fix is likely just to wipe record-array membership at the point of the unloadAll

This feels similar to an issue reported before that was already fixed, we might have a fix in later versions already. Will port the test forward to find out.


bk-pb:
I should mention that the issue in [#8924](https://github.com/emberjs/data/issues/8924) also occurs when iterating over the record list and unloading each record manually. i.e. `this.store.peekAll(...).forEach(record => record.unloadRecord())`

PR #9000, opened 2023-10-15
feat(private): native test runner

runspired:
Begins construction of a test runner to replace Testem, ember-exam, and ember-cli when using diagnostic.


Baltazore:
It's over 9000 ! 😃 🎉

PR #9003, opened 2023-10-16
decycle property lookups on promise proxies to fix usage in ember-concurrency

fivetanley:
backporting https://github.com/emberjs/data/pull/8435 to 4.4


runspired:
merged because basic tests pass. All other scenarios on branches that old are often unreliable due to dep drift etc.

PR #9054, opened 2023-10-25
Initial lint config for tests/blueprints

gitKrystan:
## Description

- Adds ESLint config for `tests/blueprints` project.
- Adds `eslint-plugin-mocha` and ESLint config for Mocha tests.
- Autofixes the resulting lint.

### DIFF
--- a/tests/blueprints/.eslintrc.cjs
+++ b/tests/blueprints/.eslintrc.cjs
@@ -1,0 +1,32 @@
+const base = require('@warp-drive/internal-config/eslint/base.cjs');
+const ignore = require('@warp-drive/internal-config/eslint/ignore.cjs');
+const imports = require('@warp-drive/internal-config/eslint/imports.cjs');
+const isolation = require('@warp-drive/internal-config/eslint/isolation.cjs');
+const mocha = require('@warp-drive/internal-config/eslint/mocha.cjs');
+const node = require('@warp-drive/internal-config/eslint/node.cjs');
+const parser = require('@warp-drive/internal-config/eslint/parser.cjs');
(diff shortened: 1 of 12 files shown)


config/eslint/node.cjs
```diff
@@ -1,16 +1,25 @@
...
       browser: false,
       node: true,
       es6: true,
+      ...config.env,
+    },
+    globals: config.globals || {},
+    plugins: ['n', 'import'],
+    extends: 'plugin:n/recommended',
+    rules: {
+      'import/order': ['error', { 'newlines-between': 'always' }],
```
gitKrystan, line 20: Should we always set `newlines-between` to `always`?

tests/blueprints/.eslintrc.cjs
```diff
@@ -0,0 +1,43 @@
...
+      rules: {
+        'n/no-unpublished-require': 'off',
+      },
+    }),
+    node.defaults({
+      useModules: true,
+      files: ['fixtures/**/*.{js,ts}'],
+      rules: {
+        'import/order': 'off',
+        'simple-import-sort/imports': 'off',
```
gitKrystan, line 39: Should we just autofix these instead of turning them off?

runspired: iirc it doesnt work right for node files is why I turned them off, but that was a long time ago.

config/eslint/node.cjs
runspired: if we can match the main library code pattern we should, there we do newlines between groups

gitKrystan: AFAICT this rule is redundant. Removing it. (Was just trying it out based on `._eslintrc`)

PR #9055, opened 2023-10-25
Fix ESLint for VSCode

gitKrystan:
## Description

Fix ESLint for VSCode

### DIFF
--- a/config/eslint/base.cjs
+++ b/config/eslint/base.cjs
@@ -1,5 +1,17 @@
 const path = require('path');
-const prettierPath = path.join(process.cwd(), '../../.prettierrc.js');
+
+let isRoot = false;
+try {
+  const dir = process.cwd();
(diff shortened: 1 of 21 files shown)


config/eslint/base.cjs
```diff
@@ -20,13 +32,13 @@ function rules() {
     'prefer-rest-params': 'off',
     'prefer-const': 'error',
 
-    "prettier/prettier": [
-      "error",
+    'prettier/prettier': [
```
gitKrystan, line 35: NOTE: We currently don't lint for prettier (or anything AFAICT) in this part of the project. Should we?

runspired: probably. I want to separate prettier out like we did in AB frontend, but haven't had a chance to. Would handle this.

runspired (approved):

PR #9063, opened 2023-10-26
docs: add requests guide

runspired:
fills out additional terminology in the new guide


Baltazore:
@runspired I think its more or less ready to merge. Should be good addition to guides

PR #9069, opened 2023-10-29
feat: Improve extensibility

runspired:
This PR introduces two small features to improve extension and configuration

1) `ember-data/store` will only create a `RequestManager` if one does not exist.
2) `@ember-data/json-api/request` now extends and re-exports `setBuildURLConfig` to allow users to specify which json-api extensions and profiles are in use.

### DIFF
--- a/packages/request-utils/src/index.ts
+++ b/packages/request-utils/src/index.ts
@@ -52,7 +52,7 @@ type Store = {
 // host and namespace which are provided by the final consuming
 // class to the prototype which can result in overwrite errors
 
-interface BuildURLConfig {
+export interface BuildURLConfig {
   host: string | null;
   namespace: string | null;
 }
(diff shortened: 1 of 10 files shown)


packages/request-utils/src/index.ts
```diff
@@ -52,7 +52,7 @@ type Store = {
 // host and namespace which are provided by the final consuming
 // class to the prototype which can result in overwrite errors
 
-interface BuildURLConfig {
+export interface BuildURLConfig {
```
Baltazore, line 55: @runspired I have a question, do you think it would be useful to allow users to overwrite function that does `resourcePath` conversion right here as part of BuildURLConfig? 

Something like

```ts 
BuildURLConfig({
  host: "https://myhost.com",
  namespace: "api/v1",
  resourcePath: (resourceName) => {
    return camelize(pluralize(resourceName));
  }
});
``` 

or is it too crazy and I should go to sleep ?

runspired: go to sleep :P

This is mostly already configurable in this way though maybe we could make it even more configurable

Baltazore: Yeah I know you can configure it using builder options, but would be nice to have it in single place like it used to be in adapter. It would cover 90% cases, and you can still change it on builder level

PR #9073, opened 2023-10-30
spec: JSON:API Graphs

runspired:
As EmberData builds out its JSON:API support and Schema driven behaviors, we've noticed that some useful capabilities are less usable by JSON:API users due to that spec not yet considering the overall ergonomics of querying and consuming graphs.

This adds a specification with an extension and profile that APIs may implement to make use of these capabilities.

[Rendered]()

### DIFF
--- a/packages/json-api-graph-spec/README.md
+++ b/packages/json-api-graph-spec/README.md
@@ -1,5 +1,8 @@
 # :electron: JSON:API Graphs
 
+> **Note**
+> Users of this specification must also utilize the [Atomic Operations Extension](https://jsonapi.org/ext/atomic/) and [Cursor Pagination Profile](https://jsonapi.org/profiles/ethanresnick/cursor-pagination/)
+
 The `JSON:API Graphs` Specification provides three things
 
 - A [profile](https://jsonapi.org/format/1.2/#profile-rules) which adds support for linkage via `*` members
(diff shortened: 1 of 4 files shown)


packages/json-api-graph-spec/spec/aql.md
```diff
@@ -0,0 +1,72 @@
...
+# This is a Comment
+QUERY company { # we wrap our query definition in QUERY {} and specify the primary resource type
+  data { # the graph of data we want back goes in data
+    name
+    ceo { # relationships with {} after them will be included
+      name
+      # if ceo is an employee and something else
+      # also declares fields for an employee, the
+      # field sets will be merged
+      profileImage
```
dsmyda-ab, line 44: for exclusion could do something like:

```
{
  -name
}
```

and if the only fields are exclusion fields in the brackets ({}) then you receive all fields minus excludes. You may not mix included and excluded fields.

vlascik:
Great idea, JSON:API needs this.

IMO though, for a query language to be really useful it should also contain comparison operators, such as:
```
eq (equals)
neq (not equals)
gt (greater than)
lt (less than)
gte (greater than or equal to)
lte (less than or equal to), 
in (list)
not_in (list), 
contains (string) or like
```

and perhaps some basic grouping and logic `(and, or, not)` as in https://dgraph.io/docs/graphql/queries/and-or-not/
and maybe even aggregate functions `(sum, avg, min, max)` as in https://dgraph.io/docs/graphql/queries/aggregate/
and multi-field ordering `(order_by: ['date': 'desc', 'name': 'asc']`, or something similar to https://dgraph.io/docs/graphql/queries/order-page/

PR #9097, opened 2023-11-12
fix: allow decorator syntax in code comments during yui doc processing

jaredgalanis:
## Description

This is a fix for decorator handling in code comment blocks. This introduces part of https://github.com/cibernox/ember-cli-yuidoc, a dep of ember.js and is what processes the ember.js docs and allows ember.js yui doc generation to handle decorator syntax. Since we have our own bespoke doc compilation process in EmberData that the addon doesn't address, we are introducing this 1 change here rather than adding `ember-cli-yuidoc` itself as a dependency.

See: https://github.com/cibernox/ember-cli-yuidoc/blob/master/lib/broccoli-yuidoc.js

This also fixes a few code blocks that were unclosed and only became exposed in failing tests once decorators were being handled properly and the docs were fully building (previously we were missing some of the code comments in docs because of the decorator issue).

BEFORE:
Screenshot 2023-11-12 at 8 49 34 AM

AFTER:

Screenshot 2023-11-12 at 8 41 17 AM

### DIFF
--- a/docs-generator/compile-docs.js
+++ b/docs-generator/compile-docs.js
@@ -4,6 +4,45 @@ const path = require('path');
 const Y = require('yuidocjs');
 const getVersion = require('git-repo-version');
 
+/**
+ * This is a fix for decorator handling in code comment blocks that is
+ * part of https://github.com/cibernox/ember-cli-yuidoc and is what allows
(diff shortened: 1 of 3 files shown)


mansona (approved):
Looks good to me 👍

runspired (approved):

PR #9126, opened 2023-11-18
fix: mutating ManyArray should handle duplicates gracefully

### LINKED ISSUE #9119: bug: Setting hasMany relationship IDs to an array with duplicate IDs should error
Issue confirmed in 4.12

As of version 4.6, duplicate records added to ManyArrays were de-duped. This unspecified behavior was accidentally removed in 4.12+.

runspired:
resolves #9119 

New Tests

**Graph Tests** (we will likely want these in 5.3 version of this, but not necessarily)

- [-] addRelatedRecords mutation dedupes
- [-] addRelatedRecords mutation asserts duplicates if we want it to
- [x] replaceRelatedRecords mutation dedupes
- [-] replaceRelatedRecords mutation can assert duplicates if we want it to
- [-] test local graph mutation notification behaviors for addRelatedRecords, removeRelatedRecords, replaceRelatedRecords

**Model/Main Tests**

- [x] ManyArray mutation dedupes for replace
- [x] ManyArray mutation dedupes for splice in with adds
- [x] ManyArray mutation dedupes for splice in with adds & removes
- [x] ManyArray mutation dedupes for push
- [x] ManyArray mutation dedupes for unshift
- [x] ManyArray is not overly notified/churning on mutation
- [x] ManyArray can optionally assert for duplicates


packages/store/src/-private/record-arrays/identifier-array.ts
```diff
@@ -355,11 +362,9 @@ class IdentifierArray {
               }
               const args: unknown[] = Array.prototype.slice.call(arguments);
               assert(`Cannot start a new array transaction while a previous transaction is underway`, !transaction);
+
               transaction = true;
-              let result: unknown = Reflect.apply(target[prop] as ProxiedMethod, receiver, args);
-              self[MUTATE]!(prop as string, args, result);
-              addToTransaction(_TAG);
-              // TODO handle cache updates
+              const result = self[MUTATE]!(target, receiver, prop as string, args, _TAG);
```
runspired, line 367: one thing I've been constantly debating around this line is whether we should try/finally it to ensure transaction is set back to false if we error.

PR #9132, opened 2023-11-20
Add auth handler guides

Baltazore:
## Description

Working on guides here 

## Notes for the release

Auth Handler guides section

### DIFF
--- a/guides/requests/examples/1-auth.md
+++ b/guides/requests/examples/1-auth.md
@@ -4,14 +4,12 @@
 - Next → [Not sure what next?](./something.md)
 - ⮐ [Requests Guide](../index.md)
 
-## Basic Usage Example
+## JWT Token 
 
 > **Note**
-> This example uses [Ember](https://emberjs.com/)
-> for convenience.
+> This example uses [Ember](https://emberjs.com/) for convenience.
 >
-> `@ember-data/request` works with raw javascript
-> or any framework of your choosing.
+> `@ember-data/request` works with raw javascript or any framework of your choosing.
 
 Many public APIs require authentication. Most common pattern nowadays is the use of an `Authorization` header with a bearer token.
 
@@ -20,6 +18,8 @@ GET /api/companies?fields[company]=name&fields[employee]=name,profileImage&inclu
 Accept: application/vnd.api+json; profile="https://jsonapi.org/profiles/ethanresnick/cursor-pagination"
 Authorization: Bearer <token>
 ```
+
+### Basic example
 
 In Ember Data we can create our own custom handler to add authentication header to all requests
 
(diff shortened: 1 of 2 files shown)


guides/requests/examples/1-auth.md
```diff
@@ -0,0 +1,167 @@
...
+```
+
+Usually this token is stored in a cookie, so we need to extract it from there. Also this token is usually sent only with `POST`, `PUT`, `PATCH` and `DELETE` requests. Let's create a handler that will do just that.
+
+```js
+const MUTATION_OPS = new Set(['createRecord', 'updateRecord', 'deleteRecord']);
+
+const AuthHandler = {
+  request({ request }, next) {
+    if (MUTATION_OPS.has(context.request.op)) {
```
SergeAstapov, line 133:
```suggestion
    if (MUTATION_OPS.has(request.op)) {
```

guides/requests/examples/1-auth.md
```diff
@@ -0,0 +1,167 @@
...
+
+    this.use([authHandler, Fetch]);
+    this.useCache(CacheHandler);
+  }
+}
+```
+
+Apart of this little twist with dependency injection, everything else is the same as in previous example.
+
+## CSRF Token
```
sandstrom, line 111: A little known thing regarding CSRF, is that as long as:

1. You have the client set a custom request header, 
2. check for it on the backend and
3. CORS isn't whitelisting all domains (`*`)

you don't need to fiddle with CSRF tokens.

I'm not saying we should change this section of the docs though, because some backends may still require this and maybe the backend is managed by another team.

But we could do something like this:

```suggestion
## CSRF Token

### Simple

The easy way of protecting against Cross Site Request Forgery (CSRF) is to set a static custom header in every HTTP Request (for example `X-CSRF-Protection: static`), check for it on the backend and make sure that CORS is only accepting data from trusted domains.

If you do this, you don't have to generate dynamic CSRF tokens for every request.

More information:

https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS#simple_requests

### Advanced

```

(then put the current text under advanced)

guides/requests/examples/1-auth.md
```diff
@@ -0,0 +1,167 @@
+# Auth Handler
+
+- Previous ← [Requests](./0-basic-usage.md)
+- Next → [Not sure what next?](./something.md)
+- ⮐ [Requests Guide](../index.md)
```
sandstrom, line 5: Minor; maybe skip these, it's some work to maintain and keep up-to-date.

sandstrom (commented):
Small comments

guides/requests/examples/1-auth.md
Baltazore: Thanks for review and this proposal. I think it makes sense to add it. @runspired what do you think ? Should I also put code related to adding header in simple section ?

PR #9198, opened 2024-01-03
fix: mutating ManyArray should handle duplicates gracefully (with deprecation)

gitKrystan:
## Description

Port of https://github.com/emberjs/data/pull/9126 and https://github.com/emberjs/data/pull/9189 (see https://github.com/emberjs/data/issues/9119)

Re-enables deduplication of duplicate records added to ManyArrays, but with a deprecation (`ember-data:deprecate-many-array-duplicates`) as described below:

```js
/**
 * **id: ember-data:deprecate-many-array-duplicates**
 *
 * When the flag is `true` (default), adding duplicate records to a `ManyArray`
 * is deprecated in non-production environments. In production environments,
 * duplicate records added to a `ManyArray` will be deduped and no error will
 * be thrown.
 *
 * When the flag is `false`, an error will be thrown when duplicates are added.
 *
 * @property DEPRECATE_MANY_ARRAY_DUPLICATES
 * @since 5.3
 * @until 6.0
 * @public
 */
```

### DIFF
--- a/tests/graph/tests/integration/graph/edge-removal/abstract-edge-removal-test.ts
+++ b/tests/graph/tests/integration/graph/edge-removal/abstract-edge-removal-test.ts
@@ -116,7 +116,7 @@ module('Integration | Graph | Edge Removal', function (hooks) {
          * However: for a newly created record any form of rollback, unload or persisted delete
          * will result in it being destroyed and cleared
          */
-        await testFinalState(
+        testFinalState(
           this,
           testState,
           config,
@@ -178,7 +178,7 @@ module('Integration | Graph | Edge Removal', function (hooks) {
         // we clear new records, or sync non-implicit relationships (client side delete semantics)
         let cleared = config.useCreate || (!config.async && !config.inverseNull);
 
-        await testFinalState(this, testState, config, { removed, cleared, implicitCleared: true }, assert);
+        testFinalState(this, testState, config, { removed, cleared, implicitCleared: true }, assert);
       });
     }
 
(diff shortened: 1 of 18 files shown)


tests/graph/tests/integration/graph/edge-removal/abstract-edge-removal-test.ts
```diff
@@ -116,7 +116,7 @@ module('Integration | Graph | Edge Removal', function (hooks) {
          * However: for a newly created record any form of rollback, unload or persisted delete
          * will result in it being destroyed and cleared
          */
-        await testFinalState(
+        testFinalState(
```
gitKrystan, line 119: IDK why the linter was complaining about these in this PR, but this change appears correct based on the code.

tests/graph/tests/integration/graph/edge-removal/helpers.ts
```diff
@@ -125,24 +132,29 @@ export async function setInitialState(context: Context, config: TestConfig, asse
...
+      // @ts-expect-error
       john.bestFriends = chris;
+      // @ts-expect-error
       chris.bestFriends = john;
     }
   }
 
-  await settled();
+  // give ourselves a tick in case there was async work
+  await Promise.resolve();
```
gitKrystan, line 157: This change came from copy-pasta. Should I keep it?

runspired: I think this should be fine

runspired (approved):

PR #9215, opened 2024-01-29
Docs: Add guide for incremental adoption

Baltazore:
## Description

Working on guides for incremental adoption of new APIs of EmberData

[Rendered](https://github.com/Baltazore/data/blob/incremental-migration-guide/guides/cookbook/incremental-adoption-guide.md)

## Notes for the release

guide for incremental adoption


guides/drafts/incremental-adoption-guide.md
```diff
@@ -0,0 +1,115 @@
...
+
+Nav links
+
+---
+
+This guide is for existing projects that want to adopt the new APIs of the EmberData incrementally. If you are starting a new project, you should use the [new project guide](./new-project-guide.md).
+
+## Step 1: Upgrade to EmberData 4.12.x
+
+This version of EmberData is the first version that supports the new APIs. It also a LTS version, so you can stay on it for a while. You can refer the [EmberData Compatibility table](https://github.com/emberjs/data/blob/main/README.md#compatibility) to see which version of EmberData is compatible with your Ember version.
```
runspired, line 11:
```suggestion
This version of EmberData is the first version that supports the new APIs. It is also a LTS version, so you can stay on it for a while. You can refer the [EmberData Compatibility table](https://github.com/emberjs/data/blob/main/README.md#compatibility) to see which version of EmberData is compatible with your Ember version.
```

guides/drafts/incremental-adoption-guide.md
```diff
@@ -0,0 +1,115 @@
...
+
+## Step 2: Add `Store` service to your application
+
+You would need to create you own store service. Before it was automatically injected by Ember Data.
```
runspired, line 15:
```suggestion
You will need to create your own store service. Before, a store service was automatically injected by EmberData. 
```

guides/drafts/incremental-adoption-guide.md
```diff
@@ -0,0 +1,115 @@
...
+import { service } from '@ember/service';
+import RequestManager from '@ember-data/request';
+
+export default class MyStore extends Store {
+  @service declare requestManager: RequestManager;
+}
+
+```
+
+Notice we still want to import the `Store` class from `ember-data/store` package. You might have a lint rule that says don't do it. You can disable it for this import. The reason we want to import it from `ember-data/store` is because we want to use Ember Data models, serializers, adapters, etc. and alongside we want to start utilizing new APIs.
```
runspired, line 30:
```suggestion
Notice we still want to import the `Store` class from `ember-data/store` package. You might have a lint rule that says don't do it. You can disable it for this import. The reason we want to import it from `ember-data/store` is because we want to use EmberData models, serializers, adapters, etc. while alongside we want to start utilizing new APIs.
```

guides/drafts/incremental-adoption-guide.md
```diff
@@ -0,0 +1,115 @@
...
+}
+```
+
+Let's go over the code above:
+
+1. `LegacyNetworkHandler` is the handler that is responsible for sending requests to the server using the old APIs. It will interrupt handlers chain if it detects request using old APIs. It will process it as it used to be doing with Adapters/Fetch/Serializers workflow.
+
+2. Next is `TestHandler`. It is a handler that is responsible for logging requests. It is a quick example of how you can add your own handlers to the request manager. We will take a look at more useful examples later.
+
+3. Lastly `Fetch`. It is a handler that is responsible for sending requests to the server using the new Ember Data APIs. It must be last handler you put in the chain. After finishing request it will enrich handlers context with `response` property. And pass it back to the handlers chain in reverse order. So `TestHandler` will receive `response` property first, and so on if we would have any.
```
runspired, line 70:
```suggestion
3. Lastly `Fetch`. It is a handler that sends requests to the server using the `fetch` API. It expects responses to be JSON and when in use it should be the last handler you put in the chain. After finishing each request it will convert the response into json and pass it back to the handlers chain in reverse order as the request context's response. So `TestHandler` will receive `response` property first, and so on if we would have any.
```

guides/drafts/incremental-adoption-guide.md
```diff
@@ -0,0 +1,115 @@
...
+
+You can read more about request manager in the [request manager guide](./request-manager-guide.md).
+
+## Step 4: Install `@ember-data/json-api`, `@ember-data/request-utils` packages
+
+If you was using JSON:API adapter/serializer for your backend communication, you can use `@ember-data/json-api` package. It is a package that contains predefined builders for JSON:API requests. You can read more about it in the [`@ember-data/json-api` guide](TODO: add link).
```
runspired, line 76:
```suggestion
If you were using JSON:API adapter/serializer for your backend communication, you can use `@ember-data/json-api` package. It is a package that contains predefined builders for JSON:API requests. You can read more about it in the [`@ember-data/json-api` guide](TODO: add link).
```

guides/drafts/incremental-adoption-guide.md
```diff
@@ -0,0 +1,115 @@
...
+
+If you have different backend format - Ember Data provides you with builders for `REST`(@ember-data/rest) and `ActiveRecord`(@ember-data/active-record).
```
runspired, line 78:
```suggestion
If you have different backend format - EmberData provides you with builders for `REST`(@ember-data/rest) and `ActiveRecord`(@ember-data/active-record).
```

runspired (commented):
Overall looking good, left some small cleanups

Baltazore:
hey @fry69 thanks for review and testing, good inputs - I will update accordingly 
Also please keep in mind there code here not works just yet, some fixes were made, but not yet released, especially around service injection area.

Baltazore:
@runspired where do we put it? I need to update the location of file, and then fix nav links. But overall I think this is ready

guides/cookbook/incremental-adoption-guide.md
```diff
@@ -0,0 +1,146 @@
...
++    const { content } = await this.store.request(query('project', {}, { host: config.api.host }));
++    this.projects = content.data;
+  });
+```
+
+You most likely would need to add Auth Handler to your request manager to add `accessToken` to your requests.
+Let's say you have your `accessToken` in the `session` service. Here is how you can add it to the request manager:
+
+```js
+import { inject as service } from '@ember/service';
```
sly7-7, line 106: I saw you use `import { service } from '@ember/service'`, maybe remove `inject`here for consistency ? (I think both are valid though)

guides/cookbook/incremental-adoption-guide.md
```diff
@@ -0,0 +1,146 @@
...
+import config from 'base-ember-typescript-app/config/environment';
++import { setBuildURLConfig } from '@ember-data/request-utils';
+
+export default class App extends Application {
+  modulePrefix = config.modulePrefix;
+  podModulePrefix = config.podModulePrefix;
+  Resolver = Resolver;
+}
+
+loadInitializers(App, config.modulePrefix);
```
sly7-7, line 140: I'm not sure if this could happen (so maybe just ignore this), but if an initializer somehow uses ED, then `setBuildURLConfig` should be called before `loadInitializers`.

tests/incremental-json-api/app/routes/application.ts
```diff
@@ -0,0 +1,36 @@
...
+import { query } from '@ember-data/json-api/request';
+import { setBuildURLConfig } from '@ember-data/request-utils';
+import type Store from '@ember-data/store';
+import type { Document } from '@ember-data/store/-private/document';
+
+import type Author from '../models/author';
+import type Book from '../models/book';
+import type Genre from '../models/genre';
+
+setBuildURLConfig({
```
sly7-7, line 13: Above this is recommended to call this function in `app/app.js`, for consistency, maybe it would be better to use the recommandation ?

Baltazore: same as above, up to change still

Baltazore:
@sly7-7 thanks for review, would need to finalize some questions

tests/incremental-json-api/app/routes/application.ts
runspired: app.js makes the most sense

sly7-7:
@Baltazore One more time I want to thank you for this work. I've just migrated to ED 5.3.3, and got my app working with the Legacy Store. Now I'm trying to switch to the new paradigm (ie using `@ember-data/json-api/request` instead of adapter/serializer. I really think this work will help me to move forward. Eventually if I find something missing, or don't understand something I'll let you know here.

guides/cookbook/incremental-adoption-guide.md
```diff
@@ -0,0 +1,148 @@
...
++    const { content } = await this.store.request(query('project', {}, { host: config.api.host }));
++    this.projects = content.data;
+  });
+```
+
+You most likely would need to add Auth Handler to your request manager to add `accessToken` to your requests.
+Let's say you have your `accessToken` in the `session` service. Here is how you can add it to the request manager:
+
+```js
+import { inject as service } from '@ember/service';
```
runspired, line 104:
```suggestion
import { service } from '@ember/service';
```

runspired (approved):

sly7-7:
@Baltazore Did you just do a rebase ?

Baltazore:
> @Baltazore Did you just do a rebase ?

yes, did I lost something ?

sly7-7:
I don't think so, actually I was speaking with @runspired and he asked me if I could make the rebase, because he wanted to merge :)

sly7-7:
@runspired, @Baltazore has been quickier 😄 so I guess you can merge

PR #9216, opened 2024-02-01
Fix trailing `?` in URL  of query builder

Baltazore:
## Description

Fix trailing ? in built url for query builder across json-api/rest/active-record

Mirage have troubles recognizing URL with trailing `?` 
I think other routers might have problems as well

### DIFF
--- a/packages/active-record/src/-private/builders/query.ts
+++ b/packages/active-record/src/-private/builders/query.ts
@@ -76,12 +76,14 @@ export function query(
 
   copyForwardUrlOptions(urlOptions, options);
 
-  const url = buildBaseURL(urlOptions);
+  const url = new URL(buildBaseURL(urlOptions));
+  url.search = buildQueryParams(query, options.urlParamsSettings);
+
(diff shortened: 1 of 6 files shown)


packages/active-record/src/-private/builders/query.ts
```diff
@@ -76,12 +76,14 @@ export function query(
...
 
-  const url = buildBaseURL(urlOptions);
+  const url = new URL(buildBaseURL(urlOptions));
+  url.search = buildQueryParams(query, options.urlParamsSettings);
+
   const headers = new Headers();
   headers.append('Accept', 'application/json;charset=utf-8');
 
   return {
-    url: `${url}?${buildQueryParams(query, options.urlParamsSettings)}`,
```
runspired, line 84: I think instead we should check if there is a query / check if buildQueryParams has output and use that to conditionally join with a `?`

runspired:
Ended up porting the tests into #9220 with a modified approach there

PR #9218, opened 2024-02-08
Backport #8850 (Overwrite addMixin)

kategengler:
Backports #8850 to LTS 4.12

* use addMixin

* here too

(cherry picked from commit 5a771b39124ab7ee60953c95624c556e9491ad06)

### DIFF
--- a/packages/model/src/-private/promise-many-array.ts
+++ b/packages/model/src/-private/promise-many-array.ts
@@ -58,7 +58,7 @@ export default class PromiseManyArray {
 
     if (DEPRECATE_A_USAGE) {
       const meta = Ember.meta(this);
-      meta.hasMixin = (mixin: Object) => {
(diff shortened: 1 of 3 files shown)


packages/model/src/-private/promise-many-array.ts
```diff
@@ -71,9 +71,17 @@ export default class PromiseManyArray {
         }
         return false;
       };
+      meta.addMixin = (mixin: object) => {
+        deprecate(`Do not use A() on an EmberData PromiseManyArray`, false, {
```
Baltazore, line 75: @runspired  I had to put deprecation call here for both `hasMixin` and `addMixin` to satisfy the tests... Not sure if this is the correct approach...

runspired:
this won't be needed anymore with the #9598 approach

PR #9240, opened 2024-02-29
feat: implement managed array for schemaRecord

richgt:
Adds support for simple autotracked arrays to schema-record.

basic schema:

```ts
{
  name: 'favoriteNumbers',
  type: null,
  kind: 'array',
},
```

per-item transformation is supported:

```ts
 {
  name: 'favoriteNumbers',
  type: 'string-from-int',
  kind: 'array',
}
```

The fields will appear and act as native arrays.

### DIFF
--- a/config/package.json
+++ b/config/package.json
@@ -18,7 +18,7 @@
     "typescript": "~5.3.3"
   },
   "engines": {
-    "node": ">= 18.19.1"
+    "node": ">= 18.17.1"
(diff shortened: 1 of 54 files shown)


runspired (approved):
In follow up we should add tests around recoverability/behavior if the transform errors on item insertion / an out-of-bounds index set or length set occurs.

PR #9246, opened 2024-03-05
normalization in json-api serializer preserves lid #7956

### LINKED ISSUE #7956: JSONAPISerializer should preserve `lid` in payloads
While we don't care to maintain that serializer that much, we should make sure it takes advantage of the feature if someone is using it and configures their API :)

sly7-7:
## Description

Fix #7956


sly7-7:
@runspired I've labelled this as feature to keep consistency with the issue

PR #9253, opened 2024-03-11
feat: typed serializers

runspired:
nowhere near ready to ship.

I suspect the js files like these that remain are skewing the d.ts types in canary and need to be killed off.


runspired:
this prove to be too hard to do. We may at some point want to ship some d.ts types or some such

PR #9261, opened 2024-03-16
wip

NullVoxPopuli:

### DIFF
--- a/packages/ember/src/-private/request.gts
+++ b/packages/ember/src/-private/request.gts
@@ -23,6 +23,8 @@ export class Throw extends Component {
   <template></template>
 }
 
+const and = (x: unknown, y: unknown) => x && y;
+
 export class Request extends Component {
   @provide('store') declare _store: Store;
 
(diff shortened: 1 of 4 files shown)


packages/store/package.json
```diff
@@ -31,6 +31,11 @@
       "default": "./addon/*.js"
     }
   },
+  "typesVersions": {
+    "*": {
+      "*": ["./unstable-preview-types/index.d.ts"]
```
NullVoxPopuli, line 36: I noticed that these weren't built, so this won't work

packages/ember/src/-private/request.gts
```diff
@@ -55,13 +57,13 @@ export class Request extends Component {
 
   <template>
     {{#if this.reqState.isLoading}}
```
NullVoxPopuli, line 59: named args must always come after positional args

packages/ember/src/-private/request.gts
```diff
@@ -23,6 +23,8 @@ export class Throw extends Component {
   <template></template>
 }
 
+const and = (x: unknown, y: unknown) => x && y;
```
NullVoxPopuli, line 26: and wasn't defined or imported

See also: https://github.com/emberjs/rfcs/issues/816

runspired:
thanks for poking around! got the branch working and applied the suggestions on gts usage there

PR #9275, opened 2024-03-22
doc: don't mention unexisting ESA auth handler

sly7-7:
## Description

I think this could be confusing, and unless I miss something such handler does not exist, and I didn't see any open issue/PR in the ESA repo mentionning it


runspired (approved):
yeah I put this there as an example, as this is what ESA should refactor to do vs the approach it uses today. Thanks!

PR #9299, opened 2024-03-27
doc: use store for save-record docs 

Yelinz:
## Description
Sometimes was `this.store` other times `store`, changed to `this.store` consistently

### DIFF
--- a/packages/rest/src/-private/builders/save-record.ts
+++ b/packages/rest/src/-private/builders/save-record.ts
@@ -33,7 +33,7 @@ function isExisting(identifier: StableRecordIdentifier): identifier is StableExi
  * ```ts
  * import { deleteRecord } from '@ember-data/rest/request';
  *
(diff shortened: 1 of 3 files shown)


Yelinz:
saw a few more inconsistencies in json-api and in the other files in rest.
Now I'm not sure what the intent is. If it should be only `store` or `this.store`

packages/rest/src/-private/builders/save-record.ts
```diff
@@ -36,10 +36,10 @@ function isExisting(identifier: StableRecordIdentifier): identifier is StableExi
  * const person = this.store.peekRecord('person', '1');
  *
  * // mark record as deleted
- * store.deleteRecord(person);
+ * this.store.deleteRecord(person);
```
runspired, line 39: we should make all examples use just `store` and only use `this.store` when in the example we are inside a class method

runspired (approved):

PR #9316, opened 2024-04-02
Notify on length when notifying that many-array has changed

grayt0r:
## Description

Port https://github.com/emberjs/data/pull/8413 to 4.6.

### DIFF
--- a/packages/model/addon/-private/many-array.ts
+++ b/packages/model/addon/-private/many-array.ts
@@ -217,6 +217,7 @@ export default class ManyArray extends MutableArrayWithObject<StableRecordIdenti
     } else {
       this._hasNotified = true;
       this.notifyPropertyChange('[]');
+      this.notifyPropertyChange('length');
       this.notifyPropertyChange('firstObject');
       this.notifyPropertyChange('lastObject');
     }


grayt0r:
Hopefully targeting the `release-4-6` branch is the correct approach here, let me know if not.

The original fix didn't come with a test, but I'm more than happy to add one if necessary.

grayt0r:
@runspired would it be possible to get this released? 🙏

grayt0r:
@runspired apologies for chasing, but this is the final blocker before I can get our app onto a more modern version of ember - please could you release `4.6.5`?

runspired:
@grayt0r its hard to release these older versions (a lot of careful manual steps) so sometimes it takes feeling like I've got a free hour or so 🙈

I should be able to get to this later today. In the meantime, `pnpm patch` or yarn patching would let you do your upgrade!

grayt0r:
@runspired thanks for explaining, I didn't mean to assume it was a simple process. I'm currently a bit limited with regards to patching options but realise that's a me problem :)

runspired:
@grayt0r haven't forgotten about you, going to be doing some releases of 4.6 and 4.12 tomorrow

grayt0r:
@runspired awesome, thank you!

runspired:
published!

dwickern:
Model-fragments uses the same code. Do we need to port this fix over there?
https://github.com/adopted-ember-addons/ember-data-model-fragments/blob/316e9be882baf3a9b20b8d6243de3e9fb518beb0/addon/array/stateful.js#L61-L71

Issue #8990, opened 2023-10-10
Assertion Failed exceptions: Cannot remove a resource that is not present


eodb:
### Reproduction

It took some time to find a reproducible qunit test but here it is:

```javascript showLineNumbers
import { run } from '@ember/runloop';
import Model, { attr, belongsTo, hasMany } from '@ember-data/model';
import { setupTest } from 'ember-qunit';
import { module, test } from 'qunit';

let store;

module('integration/relationship/clear hasMany relation twice', function (hooks) {
  setupTest(hooks);

  hooks.beforeEach(function () {
    class DocPage extends Model {
      @hasMany('entry', { inverse: 'docPageId', async: false }) entries;
    }

    class Entry extends Model {
      @attr state;
      @belongsTo('doc-page', { async: true, inverse: 'entries' }) docPageId;
      @belongsTo('page-assembly', { async: false, inverse: 'entries' }) pageAssembly;
    }

    class PageAssembly extends Model {
      @hasMany('entry', { async: false, inverse: 'pageAssembly' }) entries;
    }

    store = this.owner.lookup('service:store');

    this.owner.register('model:doc-page', DocPage);
    this.owner.register('model:entry', Entry);
    this.owner.register('model:page-assembly', PageAssembly);
  });

  const hashDocPage = {
    data: [
      {
        id: '0',
        type: 'doc-page',
      },
    ],
  };

  const hashEntries = {
    data: [
      {
        id: '0',
        type: 'page-assembly',
        relationships: { entries: { data: [{ id: '0', type: 'entry' }] } },
      },
    ],
    included: [
      {
        id: '0',
        type: 'entry',
        attributes: { state: 'CREATED' },
        relationships: {
          docPageId: { data: { id: '0', type: 'doc-page' } },
        },
      },
    ],
  };

  const hashClearDocPageEntries = {
    data: [
      {
        id: '0',
        type: 'page-assembly',
        relationships: { entries: { data: [{ id: '0', type: 'entry' }] } },
      },
    ],
    included: [
      {
        id: '0',
        type: 'entry',
        attributes: { state: 'CLEARED' },
        relationships: {
          docPageId: { data: null },
        },
      },
    ],
  };

  async function pushDocPage() {
    let docPages;
    await run(() => {
      docPages = store.push(hashDocPage);
    });
    return docPages;
  }

  async function pushEntries() {
    let pageAssemblies;
    await run(() => {
[...]

PR #9326, opened 2024-04-05
Added failing test: "Fetching a belongsTo that is set to null removes the record from a relationship - sync - and do it twice" 

eodb:
This is a PR to facilitate the fixing of https://github.com/emberjs/data/issues/8990.
An extra test "Fetching a belongsTo that is set to null removes the record from a relationship - sync - and do it twice", has been added.

## Description

It is actually the same test as the existing one that precedes it: "Fetching a belongsTo that is set to null removes the record from a relationship - sync" but only run twice immediately after each other.
If the user.accounts are not accessed in between (see the assert.strictEqual - put in comment), the test fails, otherwise the test succeeds.


runspired:
found and fixed

PR #9343, opened 2024-04-10
@ember-data/codemods package

gitKrystan:
A new package for EmberData codemods. The first will be the `legacy-compat-builders` codemod to convert `const foo = await store.findRecord('foo', '1')` to `const { content: foo } = await store.request(findRecord('foo', '1')`, etc.

## Description

- **Basic project setup**
- **legacy-compat-builders codemod**


packages/codemods/rollup.config.mjs
```diff
@@ -0,0 +1,28 @@
+import babel from '@rollup/plugin-babel';
```
gitKrystan, line 1: FYI: Assuming we run codemods through the `jscodeshift` CLI, we might only need rollup for testing purposes. The `jscodeshift` CLI will handle TS files JIT.

tests/codemods/tests/index.spec.ts
```diff
@@ -0,0 +1,168 @@
...
+  expectedOutput: string,
+  testOptions?: TestOptions
+) {
+  const output = applyTransform(module, options, input, testOptions);
+  expect(output).toEqual(expectedOutput.trim());
+  return output;
+}
+
+// prettier-ignore
+runTests(
```
gitKrystan, line 165: jscodeshift's default testing setup is missing some features, so I wrote my own.

For each codemod, we can add a folder in `__testfixtures__` with the dasherized codemod name, and this test setup will run that codemod against all of the files contained therein. `testName.input.ts` will be modded and the result will be compared with `testName.output.ts`. You can also add `testName.info.json` with other useful stuff (e.g. config--not set up yet--or expected log output--see e.g. below).

example:
```
// testName.info.json
{
  "expectedLogs": [
    [
      "warn",
      "oh noes"
    ],
  ]
}

```

packages/codemods/rollup.config.mjs
gitKrystan: Indeed, was able to remove this.

packages/codemods/package.json
```diff
@@ -0,0 +1,62 @@
+{
+  "name": "@ember-data/codemods",
+  "private": true,
```
gitKrystan, line 3:
```suggestion
```

packages/codemods/src/legacy-compat-builders/legacy-store-method.ts
```diff
@@ -0,0 +1,201 @@
...
+}
+
+function legacyStoreMethodFinder(importInfo: ParsedImportInfo) {
+  return function filter(value: CallExpression): value is LegacyStoreMethodCallExpression {
+    return (
+      value.callee.type === 'MemberExpression' &&
+      value.callee.property.type === 'Identifier' &&
+      importInfo.importedName === value.callee.property.name &&
+      (value.callee.object.type === 'Identifier' ||
+        (value.callee.object.type === 'MemberExpression' && value.callee.object.property.type === 'Identifier'))
```
gitKrystan, line 151: Look for "store" by default but add config to allow users to pass other names, e.g. `"storeService"` or `"db"`

PR #9355, opened 2024-04-12
Fix: @attr defaultValue() results should persist after initialization

### LINKED ISSUE #9315: Attributes with defaultValue functions not persisting
### High Level Overview

When an attribute is defined with `defaultValue` as a function, those _specific_ default values are not serialized.  Instead, the `defaultValue` function is re-executed upon serialization, resulting in serialized data that may not match the actual model data.

### Reproduction

https://github.com/christophersansone/ember-data-attr-default-value

Once the application is running...

The `Attribute Value` is the value of the attribute that was generated by the `defaultValue` method.
The `Serialized Value` is the value of the attribute as it is output by `model.serialize()`.

Expected Result: The two values equal.
Actual Result: The two values do not equal.

Bonus:

* Clicking `Re-serialize` will call `model.serialize()` again.  The `Serialized Value` is different each time.
* You can manually set the attribute by entering a value and clicking `Set`.  Once that occurs, the `Attribute Value` then equals the specified value, and subsequent serializations output the expected value.

### What's Happening?

Given an attribute definition with a dynamic default value, such as:

```javascript
import Model, { attr } from '@ember-data/model';

export default class Post extends Model {
  @attr('number', { defaultValue: () => Math.round(Math.random() * 10000) })
  likes;
}
```

When the model is instantiated, the attribute is correctly instantiated:

```javascript
let post = this.store.createRecord('post');
post.likes; // 1234
```

But when serializing it, it outputs a new default value every time:

```javascript
post.likes; // 1234
post.serialize().data.attributes.likes; // 4958
post.serialize().data.attributes.likes; // 2423
post.serialize().data.attributes.likes; // 8463
post.likes; // still 1234
```

Once the value is manually set, the serialized value stabilizes:

```javascript
post.set('likes', 9999);
post.serialize().data.attributes.likes; // 9999
post.serialize().data.attributes.likes; // 9999
post.serialize().data.attributes.likes; // 9999
```

### Why It Matters

This example is a minimal reproduction of the issue and is fairly contrived, but the issue has real world implications.

If the default value is a primitive string, number, boolean, etc., it is typically defined as a static value, e.g.:

```javascript
@attr('boolean', { defaultValue: false }) isPublished;
```

When the default value is a complex object such as an object or array, it must be defined as a function in order to get uniquely instantiated values for each model.  In fact, there is a [nice little assertion](https://github.com/emberjs/data/blob/900ce877f041d2c2736ce33716582f95f56bd936/packages/json-api/src/-private/cache.ts#L1217) to ensure this is the case.

In my usage, I have a custom transform that serializes an array of objects.  I instantiate the attribute as an empty array up front to avoid needing to check and instantiate it downstream.

```javascript
@attr('item-array', { defaultValue: () => [] }) items;
```

Because of this issue, if I add items to this default array, the items will be in the attribute, but it will always serialize as an empty array (because it re-executes `defaultValue()`.

### Research

After doing some digging, this is what I was able to put together.

I added a `debugger` statement in one of my `defaultValue` methods to confirm that it is called again upon serialization, and to view the call stack at that point.

The culprit appears to be the JSON:API cache.  It is hitting [this line](https://github.com/emberjs/data/blob/900ce877f041d2c2736ce33716582f95f56bd936/packages/json-api/src/-private/cache.ts#L904) during serialization, when the expectation is that it should use the value that was already initialized on the attribute.

From what I can tell, it looks like `localAttrs` should contain all the values that have been modified from the original.  In this case, it seems like it should contain these instantiated `defaultValue` values, as if they were set manually, but it does not.

### Versions

```cli
├─┬ @ember/test-helpers@3.3.0
│ └── ember-source@5.6.0 deduped
├─┬ ember-cli-app-version@6.0.1
│ └── ember-source@5.6.0 deduped
├─┬ ember-data@5.3.3
│ └─┬ @ember-data/tracking@5.3.3
│   └─┬ ember-cached-decorator-polyfill@1.0.2
│     └── ember-source@5.6.0 deduped
├─┬ ember-modifier@4.1.0
│ └── ember-source@5.6.0 deduped
├─┬ ember-page-title@8.2.3
│ └── ember-source@5.6.0 deduped
├─┬ ember-qunit@8.0.2
│ └── ember-source@5.6.0 deduped
├─┬ ember-resolver@11.0.1
│ └── ember-source@5.6.0 deduped
└── ember-source@5.6.0

├─┬ ember-cli-dependency-checker@3.3.2
│ └── ember-cli@5.6.0 deduped
└── ember-cli@5.6.0

└── ember-data@5.3.3
```

First noticed on Ember Data 4.12.3

christophersansone:
## Description

Fixes #9315. 

(cc @runspired)

## Notes for the release

Before this change, when `@attr` defines a `defaultValue` as a function, the default value instantiated on a model instance will not necessarily be the value that is serialized.  If the attribute was not `set`, and therefore the default value is used, then the serialization would call the `defaultValue` function again instead of returning the value on the model.  This is mostly a problem when the attribute value is an array or object, because changes can be made to the object, resulting in those changes being lost upon serialization.  With this change, the original `defaultValue` is persisted at the cache level to avoid this issue.

### DIFF
--- a/packages/json-api/src/-private/cache.ts
+++ b/packages/json-api/src/-private/cache.ts
@@ -70,6 +70,7 @@ interface CachedResource {
   id: string | null;
   remoteAttrs: Record<string, Value | undefined> | null;
   localAttrs: Record<string, Value | undefined> | null;
+  defaultAttrs: Record<string, Value | undefined> | null;
   inflightAttrs: Record<string, Value | undefined> | null;
   changes: Record<string, [Value | undefined, Value]> | null;
   errors: JsonApiError[] | null;
@@ -1036,6 +1037,7 @@ export default class JSONAPICache implements Cache {
     // we report as `isEmpty` during teardown.
     cached.localAttrs = null;
     cached.remoteAttrs = null;
+    cached.defaultAttrs = null;
     cached.inflightAttrs = null;
 
     const relatedIdentifiers = _allRelatedIdentifiers(storeWrapper, identifier);
@@ -1096,11 +1098,18 @@ export default class JSONAPICache implements Cache {
       return cached.inflightAttrs[attr];
     } else if (cached.remoteAttrs && attr in cached.remoteAttrs) {
       return cached.remoteAttrs[attr];
+    } else if (cached.defaultAttrs && attr in cached.defaultAttrs) {
+      return cached.defaultAttrs[attr];
     } else {
       const attrSchema = this._capabilities.schema.fields(identifier).get(attr);
 
       upgradeCapabilities(this._capabilities);
-      return getDefaultValue(attrSchema, identifier, this._capabilities._store);
+      const defaultValue = getDefaultValue(attrSchema, identifier, this._capabilities._store);
+      if (typeof attrSchema?.options?.defaultValue === 'function') {
+        cached.defaultAttrs = cached.defaultAttrs || (Object.create(null) as Record<string, Value>);
+        cached.defaultAttrs[attr] = defaultValue;
+      }
+      return defaultValue;
     }
   }
 
@@ -1131,6 +1140,10 @@ export default class JSONAPICache implements Cache {
     } else if (cached.localAttrs) {
       delete cached.localAttrs[attr];
       delete cached.changes![attr];
+    }
+
+    if (cached.defaultAttrs && attr in cached.defaultAttrs) {
+      delete cached.defaultAttrs[attr];
     }
 
     this._capabilities.notifyChange(identifier, 'attributes', attr);
@@ -1195,6 +1208,7 @@ export default class JSONAPICache implements Cache {
     }
 
     cached.inflightAttrs = null;
+    cached.defaultAttrs = null;
 
     if (cached.errors) {
       cached.errors = null;


packages/json-api/src/-private/cache.ts
runspired, line 1: I believe we also need to clear defaultAttrs during patchLocalAttributes

runspired: > This seems like quite an edge case, but I think my opinion would be that the original default value would take effect again, as opposed to a new one being created.

If the server has previously informed us of a state, and that state later disappears, imo the correct thing to do is start over.

>  I was hoping to find a nice set of unit tests for this cache class, but no luck

The cache requires integration tests. Newest ones are here https://github.com/emberjs/data/tree/main/tests/ember-data__json-api/tests/integration/cache and here https://github.com/emberjs/data/blob/main/tests/main/tests/integration/cache/json-api-cache-test.js

but most of them are scattered through the main test app, with most having been written from the perspective of user observable behaviors (due largely to how cache grew out of Store/Model/InternalModel).

I'd recommend adding new tests at that first link above since this is a behavior specific to JSON:API cache.

runspired: @christophersansone got any more time for this one? I can poke at it a bit if needed

christophersansone:
Sorry!  The day job got in the way this week.  Planning to spend some time on it this weekend.

christophersansone:
@runspired Sorry for the delay!

Okay, I added the change to `patchLocalAttributes`, which I am pretty sure is correct.  I also wrote a few tests that I believe cover the basics here, really just focused on when the default values are cached and / or regenerated.  Thanks for pointing me in the right direction on these!

I did not figure out how to adequately test the change for `patchLocalAttributes`.  If the server informed us of an attribute, the server can't really clear the entire state of that attribute.  The attribute can either be stated in `attributes` or not.  If it's stated, that value will take precedence.  If it's not stated, the current local value will be retained.

Let me know if there's anything else that makes sense to test.

runspired (approved):

daniellubovich:
@runspired Is this something you think could be easily backported into 4.12?

runspired:
@daniellubovich the backport commit shouldn't be too hard

PR #9360, opened 2024-04-16
fix: Make IS_MAYBE_MIRAGE work in Firefox

MichalBryxi:
## Description

- FF vs Chrome produces different `toString()` for native `window.fetch`.
- This patch fixes that.

## Notes for the release

Fix of a bug where Mirage identification failed for Firefox.


MichalBryxi:
One smarter person than me suggested that maybe a better option would be to make the `toString()` whitespace agnostic and thus avoid any formatting issues whatsoever:

```
window.fetch.toString().replace(/\s+/g, '') !== 'function fetch() { [native code] }'.replace(/\s+/g, '');
```

runspired:
@MichalBryxi yeah I thought of that too, am down. Runs in debug only though so not a huge thing either way

PR #9380, opened 2024-04-23
[WIP] issue 9232

Mikek2252:


runspired (commented):
needs a bit of cleanup and we should make sure to also PR this to main

packages/model/addon/-private/record-state.ts
```diff
@@ -233,6 +233,9 @@ export default class RecordState {
     notifications.subscribe(identity, (identifier: StableRecordIdentifier, type: NotificationType, key?: string) => {
       switch (type) {
         case 'state':
+          this.notify('isSaved');
```
Mikek2252, line 236: @runspired what do you think is the best way to test this? Can try to write ones based of an existing one if you have an example

PR #9383, opened 2024-04-24
fix: ensure cache-handler clones full errors

runspired:
Fixes an external report that surfaced that the CacheHandler would discard much of the useful error when cloning to prevent mutation.


packages/holodeck/bin/ensure-cert.js
```diff
@@ -0,0 +1,40 @@
...
+    process.env.HOLODECK_SSL_KEY_PATH = KEY_PATH;
+    execSync(`echo '\nexport HOLODECK_SSL_KEY_PATH="${KEY_PATH}"' >> ${getShellConfigFilePath()}`);
+    console.log(`Added HOLODECK_SSL_KEY_PATH to ${getShellConfigFilePath()}`);
+  }
+
+  if (!fs.existsSync(CERT_PATH) || !fs.existsSync(KEY_PATH)) {
+    console.log('SSL certificate or key not found, generating new ones...');
+
+    execSync(`mkcert -install`);
+    execSync(`mkcert -key-file ${KEY_PATH} -cert-file ${CERT_PATH} localhost`);
```
Baltazore: should we mention that `mkcert` would be required to have this auto-cert feature?

runspired: yes

PR #9384, opened 2024-04-25
Implement schema-array for schema-record

richgt:


packages/schema-record/src/managed-array.ts
```diff
@@ -159,6 +159,41 @@ export class ManagedArray {
...
+                // for stability. Default should be `@identity` which means that
+                // same object reference from cache should result in same SchemaRecord
+                // embedded object.
+                recordPath.push(index as unknown as string);
+                record = new SchemaRecord(
+                  store,
+                  self.owner[Identifier],
+                  { [Editable]: self.owner[Editable], [Legacy]: self.owner[Legacy] },
+                  true,
+                  field.type,
```
runspired, line 184: we'll need to potentially pull additional info from the object for this

PR #9385, opened 2024-04-28
fix: Make IS_MAYBE_MIRAGE simplified

MichalBryxi:
## Description

- In previous version we fixed the code for FireFox that can now work in FireFox which has different `.toString()` implementation
- This change should make it more error prone as it completely removes whitespace from the strings and then compares them
- Follow-up for #9360

### DIFF
--- a/packages/request/src/fetch.ts
+++ b/packages/request/src/fetch.ts
@@ -40,8 +40,7 @@ if (DEBUG) {
     Boolean(
       typeof window !== 'undefined' &&
         ((window as { server?: { pretender: unknown } }).server?.pretender ||
-          (window.fetch.toString() !== 'function fetch() { [native code] }' &&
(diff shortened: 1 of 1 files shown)


runspired (approved):

runspired (commented):
seems to fail both lint and ts parsing, probably a missed paren.

Thanks for the follow up :D

MichalBryxi:
Sigh me :)

Maybe would help if I would not just ignore the hooks setup and figure out what is this error about:

```sh
git commit --amend
.git/hooks/pre-commit: line 2: ./node_modules/pre-commit/hook: No such file or directory
```

Ok, let's try one more time.

runspired:
@MichalBryxi we have no hooks! that means you cloned us at a point long ago when we did. You can safely remove that hook file/directory.

PR #9319, opened 2024-04-04
Add @ember-data/legacy-compat/builders

gitKrystan:
## Description

Adds `@ember-data/legacy-compat/builders` module with builders for migrating from `store` methods to `store.request`.

These builders enable you to migrate your codebase to using the correct syntax for `store.request` while temporarily preserving legacy behaviors. This is useful for quickly upgrading an entire app to a unified syntax while a longer incremental migration is made to shift off of adapters and serializers.

- **findRecord**
- **query**
- **queryRecord**
- **findAll**
- **saveRecord**

PR #9386, opened 2024-04-29
Backport #9319 to lts-4-12: Add @ember-data/legacy-compat/builders

gitKrystan:


packages/legacy-compat/src/builders/find-all.ts
```diff
@@ -3,24 +3,26 @@
...
-import { SkipCache } from '@warp-drive/core-types/request';
-import type { RequestSignature } from '@warp-drive/core-types/symbols';
-
+import type Model from '@ember-data/model';
+import { SkipCache } from '@ember-data/request';
+import type { ImmutableRequestInfo } from '@ember-data/request/-private/types';
 import { normalizeModelName } from './utils';
 
-type FindAllRequestInput<T extends string = string, RT = unknown[]> = StoreRequestInput & {
+// Keeping unused generics for consistency with 5x types
```
gitKrystan, line 11: I know we aren't publishing these types, but removing them was more work than just keeping them so....

runspired (approved):

PR #9394, opened 2024-05-07
Add cookbook page about model names convention

Baltazore:


guides/cookbook/index.md
```diff
@@ -1,3 +1,4 @@
 # Cookbook
 
 - [Incremental Adoption Guide](./incremental-adoption-guide.md)
+- [Model Name: singular or plural? What to choose? Why is that?](./model-name.md)
```
runspired, line 4:
```suggestion
- [Naming Conventions: Should resource types be singular or plural? What to choose? Why is that?](./model-name.md)
```

I'd suggest we also call the file `naming-conventions.md`. That way we can also add a bit on `snake_case` vs `camelCase` key conventions etc. if desired.

guides/cookbook/model-name.md
```diff
@@ -0,0 +1,25 @@
+# Model Name: singular or plural? What to choose? Why is that?
+
+- ⮐ [Cookbook](./index.md)
+
+## Model name convention (or why it was singular)
```
runspired, line 5:
```suggestion
## Resource Type (model name) conventions – or, why it was singular
```

guides/cookbook/model-name.md
```diff
@@ -0,0 +1,25 @@
...
+
+If you have been working with EmberData for a while, you might remember convention about model names. It was a convention that model names should be singular. But why is that? Why not plural?
```
runspired, line 7:
```suggestion
If you have been working with EmberData for a while, you might remember a convention about singular-dasherized resource types (or modelNames). It was a convention that model names should be singular. But why is that? Why not plural? And why dasherized?
```

guides/cookbook/model-name.md
```diff
@@ -0,0 +1,25 @@
...
+
+There is no strict rule about model names. You was just using singular names because you had default Serializer configured in your app. It was doing all job of data normalization for you. Mainly it was singularizing `type` of response you had received from server.
```
runspired, line 9:
```suggestion
There is no longer any strict rule in EmberData governing what naming convention to use for resource types. Before, you may have been using singular names because you had default Serializer configured in your app. The default serializers assume types should be singular and dasherized, and since they do the job of data normalization for you they would singularize and dasherize the `types` received from your server.
```

runspired: I think we should add a bit more in here around the history of why it used to be that you needed singular, dasherized: mainly its the path-on-disc convention in Ember that drove this.

guides/cookbook/model-name.md
```diff
@@ -0,0 +1,25 @@
...
+
+## So what to choose?
+
+When using EmberData without Legacy setup, you are responsible for data normalization. You can choose whatever you want. You can use singular or plural names. It is up to you. Or up to your backend to be precise, as it would be beneficial for you to not do all that normalization on frontend. Just have it as a part of API contract of your app. But remember, you need to be consistent. If you choose singular names, stick with it. If you choose plural names, stick with it.
```
runspired, line 13:
```suggestion
When using EmberData without Legacy setup, you are responsible for data normalization. You can choose whatever you want. You can use singular or plural names. It is up to you. Or up to your backend to be precise, as it would be beneficial for you to not do all that normalization on frontend. Just have it as a part of API contract of your app. But remember, you need to be consistent. If you choose singular names, stick with it. If you choose plural names, stick with it. **Be Consistent!**

What does consistency look like?

#### Let's say your convention is singular dasherized, e.g. `user-setting`:

- the API should respond with `user-setting` (or your handler/serializer should normalize the type to)
- calls to store methods should use the same format: `store.findRecord('user-setting', '1')`
- relationship definitions should also use this format:
  \`\`\`ts
  class User extends Model {
    @hasMany('user-setting', { async: false, inverse: null }) userSettings;
  }
  \`\`\`
- The model files should also use this format, e.g. the model would be located in `app/models/user-setting.{js,ts}`

#### But what about plural and snake case?


- the API should respond with `user_settings`
- calls to store methods: `store.findRecord('user_settings', '1')`
- relationship definitions:
  \`\`\`ts
  class User extends Model {
    @hasMany('user_settings', { async: false, inverse: null }) userSettings;
  }
  \`\`\`
- The model file would be located in `app/models/user_settings.{js,ts}`

```

PR #9396, opened 2024-05-10
fix: Resolve promise types for props passed to `store.createRecord()`

seanCodes:
## Description

Fix incorrect types being expected by the store service’s `createRecord()` method. This could happen when the record being created had, for example, a `belongsTo` relationship, in which case the type for the property with the relationship was `PromiseBelongsTo<SomeModel>` rather than `SomeModel`.

## Cause

When determining the TS type for properties that can be passed to `createRecord()` a `CreateRecordProperties` utility type is used. It in turn uses a `FilteredKeys` type to narrow the properties to only the model attributes. Promise types were not being handled here and only filtered, leading to the issue.

## Fix

To resolve any promise types like `PromiseBelongsTo`, a new `AwaitedKeys<Type>` utiltity type was added that wraps `FilteredKeys`. It’s a mapped type that applies [`Awaited`](https://www.typescriptlang.org/docs/handbook/utility-types.html#awaitedtype) to all keys, resolving promise types.

## Notes for the release

N/A

### DIFF
--- a/packages/store/src/-private/store-service.ts
+++ b/packages/store/src/-private/store-service.ts
@@ -66,7 +66,10 @@ type CompatStore = Store & {
 };
 function upgradeStore(store: Store): asserts store is CompatStore {}
 
-type FilteredKeys<T> = Omit<T, typeof ResourceType | keyof EmberObject | 'constructor'>;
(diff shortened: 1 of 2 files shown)


runspired (approved):

runspired (commented):
needs a lint fix but otherwise 👍🏻

PR #9402, opened 2024-05-11
[WIP] fix: resolve and unwrap `AsyncHasMany` for `store.createRecord()`

seanCodes:
_This is a WIP. The code works as expected but a cyclical dependency was introduced that needs to be resolved._

## Description

In #9396 the prop types expected by `store.createRecord()` were updated to resolve `AsyncBelongsTo` to the underlying model.

`AsyncHasMany` wasn't handled though, so this PR addresses that. To do so the `CreateRecordProperties` type util was updated to look for `RelatedCollection`s (which are the result of `Awaited`-ing the `AsyncHasMany` type) and then the model used in the collection is turned into a plain array of that model type. This allows the following:

```ts
const parentModel = store.createRecord<ParentModel>('parent', { children: [existingChildModel] })
```

Previously this would error because `[existingChildModel]` in the model props wasn't of type `RelatedCollection<ChildModel>`.

## Notes for the release

N/A

### DIFF
--- a/packages/store/src/-private/store-service.ts
+++ b/packages/store/src/-private/store-service.ts
@@ -5,6 +5,8 @@
 import { assert } from '@ember/debug';
 import EmberObject from '@ember/object';
 
+import type Model from '@ember-data/model';
+import type RelatedCollection from '@ember-data/model/-private/many-array';
 import type RequestManager from '@ember-data/request';
 import type { Future } from '@ember-data/request/-private/types';
 import { LOG_PAYLOADS, LOG_REQUESTS } from '@warp-drive/build-config/debugging';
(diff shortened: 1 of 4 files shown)


packages/store/src/-private/store-service.ts
```diff
@@ -66,10 +68,15 @@ type CompatStore = Store & {
 };
 function upgradeStore(store: Store): asserts store is CompatStore {}
 
-type AwaitedKeys<T> = { [K in keyof T]: Awaited<T[K]> };
+type UnwrapRelationshipKeys<T> = {
+  // For `AsyncHasMany`: apply `Awaited` and check if the type is a `RelatedCollection`. If so,
+  // unwrap it and convert it to a plain array of the model type. For everything else (including
+  // `AsyncBelongsTo`): apply `Awaited`.
+  [K in keyof T]: NonNullable<Awaited<T[K]>> extends RelatedCollection<infer M> ? Array<M> : Awaited<T[K]>;
```
runspired, line 75: it should work here to check if it extends `Array<infer M>`

PR #9440, opened 2024-05-15
Update dom.ts DOM text reinterpreted as HTML

Shivam7-1:
## Description

By using innerText, it will avoid the risk of HTML injection, as these properties automatically escape any HTML special characters in the provided text. This helps prevent cross-site scripting (XSS) vulnerabilities by treating the input as plain text rather than interpreted HTML.


runspired:
In this case we're clearing the DOM, never appending. Alternatives are setting via `textContent` or using `element.replaceChildren()`, though for the "clear all" case the difference is negligible. The only reason to avoid `innerHTML` would be if some security scanner that matters is particularly dogmatic about it, which .... they shouldn't be.

Setting via `innerText` isn't an alternative, it would behave differently.

PR #9458, opened 2024-05-25
Remove ember-source from package.json

Baltazore:


runspired:
need to run `pnpm install` to get the lockfile changes as well

PR #9476, opened 2024-06-14
chore: cleanup symbol usage

runspired:


Baltazore:
love to see just `Type` there, I am always worried to use it, as sometimes this so common word might backfire :D

PR #9494, opened 2024-06-18
Use vite's built-in package.json loading to inline names and versions

ef4:
This eliminates the need for embroider macros to provide packages their own name and version. That is not a good use case for macros, since it can never change post-publication. The prepublication Vite build here has built-in support for reading from package.json.

(Also, this is being motivated  by the fact that addon-main.js in v2 addons in not guaranteed to be run by embroider. It's a classic file that does nothing under our Vite build. Having an alternative that allows apps to manipulate the macros config is planned as a blocker for releasing the vite blueprint.)

This PR is only a WIP, because I suspect the package publication process here needs to change to ensure that `vite build` gets run after each of the mirror package names and/or versions gets set. There's a lot going on in the publication step, so I figured it would be best to just open this for discussion first.

### DIFF
--- a/packages/-ember-data/src/version.ts
+++ b/packages/-ember-data/src/version.ts
@@ -1,4 +1,1 @@
-import { getOwnConfig } from '@embroider/macros';
-
-const VERSION = getOwnConfig<{ VERSION: string }>().VERSION;
-export default VERSION;
+export { version as default } from '../package.json';
--- a/packages/-ember-data/addon-main.cjs
+++ b/packages/-ember-data/addon-main.cjs
@@ -3,12 +3,7 @@
 const { addonShim } = require('@warp-drive/build-config/addon-shim.cjs');
 
 const addon = addonShim(__dirname);
-addon.options = addon.options || {};
-addon.options['@embroider/macros'] = addon.options['@embroider/macros'] || {};
 const pkg = require('./package.json');
-addon.options['@embroider/macros'].setOwnConfig = {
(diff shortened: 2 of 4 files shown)


packages/-ember-data/src/version.ts
```diff
@@ -1,4 +1 @@
-import { getOwnConfig } from '@embroider/macros';
-
-const VERSION = getOwnConfig<{ VERSION: string }>().VERSION;
-export default VERSION;
+export { version as default } from '../package.json';
```
runspired, line 1: is this guaranteed to tree-shake? Or does it pull the full json object into the built asset?

NullVoxPopuli: it replaces with the string value -- doesn't even keep the import in the output

ef4:
```sh
$ cat dist/version.js                                                                                                        [version-inlining][ruby-2.7.4]
const version = "5.4.0-alpha.88";
export { version as default }             
```

runspired: Nice, then we should be able to merge this as-is. The mirror publish step uses a regex to replace package names where-ever they appear and the version would match since they are always in sync.

mansona:
Hey folks, what's the plan for backporting this to the current stable release? We're being hit by this in the vite preview app blueprint still: https://github.com/embroider-build/app-blueprint/actions/runs/9589542127/job/26443484423

PR #9503, opened 2024-06-29
feat: request deduping & <Request /> invalidate subscriptions

runspired:
Follow up from https://github.com/emberjs/data/pull/9363

- [x] dedupe requests: before if multiple requests required bypassing cache/fulfilling from handler chain each would run in its entirely. This primarily negatively affected first-load scenarios / thundering-herd reload scenarios. Now requests with a unique identifier will only maintain one active-request per-id per-store at a time.
- [x] invalidate subscription: we've added a new `invalidate` event to the notification manager, and upgraded the CachePolicy provided by request-utils to send this event / upgraded `<Request />` to subscribe to this event
- [x] time based refresh on `<Request />`
- [x] invalidate event based refresh on `<Request />`
- [x] background reload subscriptions on `<Request />`
- [x] reload subscriptions on `<Request />`


runspired:
this needs some tests but the implementation is in place

PR #9505, opened 2024-07-01
Fix/code snippet json api/request.ts

MehulKChaudhari:
## Description
In the JSON API request documentation, there's an issue with a code snippet. The snippet is intended to be displayed as a comment, but it instead shows up with unwanted characters.

resolves: 
Screenshot 2024-06-10 at 11 14 11 PM

### DIFF
--- a/packages/active-record/src/request.ts
+++ b/packages/active-record/src/request.ts
@@ -31,7 +31,7 @@ import { findRecord } from '@ember-data/active-record/request';
 
 const options = findRecord('ember-developer', '1', { include: ['pets', 'friends'] });
 
-/\*
+/*
   => {
     url: 'https://api.example.com/v1/ember_developers/1?include=friends,pets',
     method: 'GET',
@@ -39,7 +39,7 @@ const options = findRecord('ember-developer', '1', { include: ['pets', 'friends'
     op: 'findRecord';
     records: [{ type: 'ember-developer', id: '1' }]
   }
-*\/
+*\
 ```
 
 Request builder output may be used with either `requestManager.request` or `store.request`.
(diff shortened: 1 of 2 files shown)


packages/json-api/src/request.ts
```diff
@@ -41,7 +41,7 @@ const options = findRecord('ember-developer', '1', { include: ['pets', 'friends'
     op: 'findRecord';
     records: [{ type: 'ember-developer', id: '1' }]
   }
-*\/
+*\
```
runspired, line 44: I wonder if we need `\/*` and `\*/` instead

MehulKChaudhari: I think this should work.

That's how we do it in other files for example [here](https://github.com/emberjs/data/blob/3ec0359017d321d43d914f4fffda51c0672f58b5/packages/rest/src/request.ts#L34)

MehulKChaudhari: If we want to change it to `\*/`.  we can do it.

packages/active-record/src/request.ts
```diff
@@ -31,15 +31,15 @@ import { findRecord } from '@ember-data/active-record/request';
...
+/*
   => {
     url: 'https://api.example.com/v1/ember_developers/1?include=friends,pets',
     method: 'GET',
     headers: <Headers>, // 'Content-Type': 'application/json;charset=utf-8'
     op: 'findRecord';
     records: [{ type: 'ember-developer', id: '1' }]
   }
-*\/
+*\
```
runspired, line 42:
```suggestion
* /
```

packages/json-api/src/request.ts
runspired, line 44:
```suggestion
* /
```

packages/active-record/src/request.ts
MehulKChaudhari: Ohh I see, got it. 

Thank you!

anandfresh (approved):

Issue #9502, opened 2024-06-25
[5.4.0-alpha.91] Incorrect/Missing typings


mkszepp:
I have recently updated in our app the native type packages from `5.4.0-alpha.49` to `5.4.0-alpha.91`

After updating we have seen that there already fixed other incorrect types (thanks for that), but other types are now incorrect or maybe other improvements are bringing now incorrect TS errors

1. The option `include` was defined as `array`. it should be a `string`, otherwise the JSONAPI call is incorrect. Passing as array generates url `&include[]=pets`, which is not correct. [demo](https://github.com/mkszepp/ember-data-ts-errors/blob/8aad8a964fe787ac1a16dac68402889d8b81cc73/app/routes/index.ts#L11-L14) - [JSON:API Specs](https://jsonapi.org/format/#fetching-includes). This error is in `findAll`, `findRecord`, `query`... (i think in all legacy functions)
  ```ts
  const users = await this.store.findAll<UserModel>('user', {
      // Error "Type 'string' is not assignable to type 'undefined[]'.ts(2322)"
      include: 'pets',
    });
``` 

2. When we a `findAll` returns a list of models and we take for example the first to make additional calls like `findRecord`, we see error `Error Argument of type 'string | null' is not assignable to parameter of type 'string | number'. Type 'null' is not assignable to type 'string | number'.ts(2345)` [demo](https://github.com/mkszepp/ember-data-ts-errors/blob/8aad8a964fe787ac1a16dac68402889d8b81cc73/app/routes/index.ts#L19-L23)
  Expected: `id` should be `string | number`, instead of `string | null`
```ts
user = await this.store.findRecord<UserModel>('user', firstUser.id, {
        include: 'pets',
      });
```

3. We have used in some models `this.constructor.relationshipsByName;`. In this case we get error "Property 'relationshipsByName' does not exist on type 'Function'.". This works in JS... but typing looks like incorrect or missing [see](https://github.com/mkszepp/ember-data-ts-errors/blob/main/app/models/user.ts#L10-L13)
```ts
 relationshipNames() {
    // Error "Property 'relationshipsByName' does not exist on type 'Function'."
    return this.constructor.relationshipsByName;
  }
```

All TS errors you can find in this repo: https://github.com/mkszepp/ember-data-ts-errors

runspired:
these are all expected

[...]

PR #9507, opened 2024-07-09
Include param should always reference array in docs

Baltazore:
Due to confusion described in https://github.com/emberjs/data/issues/9502 we should update docs and maybe internal usage

TODO:
- [ ] Add lint rule with autofix 
- [ ] Verify all docs updated 
- [ ] Verify guides updated https://github.com/ember-learn/guides-source/pull/2048

### DIFF
--- a/guides/requests/examples/0-basic-usage.md
+++ b/guides/requests/examples/0-basic-usage.md
@@ -138,7 +138,7 @@ import fetch from './fetch';
 
 // ... execute a request
 const { content: collection } = await fetch.request(query('company', {
-  include: 'ceo',
+  include: ['ceo'],
   fields: {
     company: 'name',
     employee: ['name', 'profileImage']
@@ -182,7 +182,7 @@ import { query } from '@ember-data/json-api/request';
 
 // ... execute a request
 const { content: collection } = await store.request(query('company', {
-  include: 'ceo',
+  include: ['ceo'],
   fields: {
     company: 'name',
     employee: ['name', 'profileImage']
@@ -249,7 +249,7 @@ import fetch from './fetch';
 // ... execute a request
 try {
   const result = await fetch.request(query('company', {
-    include: 'ceo',
+    include: ['ceo'],
     fields: {
       company: 'name',
       employee: ['name', 'profileImage']
--- a/packages/store/src/-private/store-service.ts
+++ b/packages/store/src/-private/store-service.ts
@@ -1278,7 +1278,7 @@ export class Store extends BaseClass {
     ```app/routes/post.js
     export default class PostRoute extends Route {
       model(params) {
-        return this.store.findRecord('post', params.post_id, { include: 'comments' });
+        return this.store.findRecord('post', params.post_id, { include: ['comments'] });
       }
     }
     ```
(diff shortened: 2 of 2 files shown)


mkszepp:
In you checklist we need to add one point, that we need todo, before we ship lint, updating docs... otherwise api isn't working correctly with includes, when users are using legacy code-parts.

Atm all legacy codes like `this.store.findAll` / `this.store.findRecord` are passing `include` not correctly to api.

When we pass a string[] to include (`include: 'comments,author'`) the urls is:
`api/post?include[]=comments&include[]=author`

Correctly is:
`api/post?include=comments,author`

runspired:
@mkszepp as I said in the ticket, we'd need to look at your particular setup to determine why that is, as that's up to the adapter to do

mkszepp:
> @mkszepp as I said in the ticket, we'd need to look at your particular setup to determine why that is, as that's up to the adapter to do

its not my stetup directly... the bug is already with default adapter... here the repo with the bug (like already shared in issue https://github.com/mkszepp/ember-data-ts-errors/commit/152a1055c53978644e3c8c99c42a24ab701c5223)

grafik

anandfresh (approved):

Baltazore:
Closing this as superseded.

#9583 (`feat|doc|fix: cleanup includes story`, landed 2024-10-21 as 55714da220) made the same two documentation changes this PR proposes, and went further. Checked against `main` today:

- `guides/requests/examples/0-basic-usage.md` has moved to `guides/the-manual/cookbook/basic-usage.md`, and all three `include: 'ceo'` samples now read `include: ['ceo']`.
- `packages/store/src/-private/store-service.ts` has moved to `warp-drive-packages/core/src/store/-private/store-service.ts`, where the four `findRecord` / `findAll` docblock samples are arrays and both prose paragraphs say "a list of relationship names" rather than "a comma-separated list (without white-space)".

On @mkszepp's point that legacy code paths emit `include[]=comments&include[]=author`, that was handled in the same PR and still holds in `main`:

- The JSON:API adapter joins arrays before serializing (`warp-drive-packages/legacy/src/adapter/json-api.ts:292`), so `include: ['comments', 'author']` produces `?include=comments,author`.
- The REST adapter passes `include` through untouched on purpose and now documents that (`warp-drive-packages/legacy/src/adapter/rest.ts:1247-1249`), naming `buildQuery` as the override point for apps that want the comma form.

`include` remains typed `string | string[]` (`warp-drive-packages/core/src/types/request.ts:436`), so this was always a docs-consistency change rather than an API change. #9502 is closed as completed.

Two checklist items are still open, and neither needs this branch:

- No lint rule with autofix exists. `packages/eslint-plugin-warp-drive/src/rules/` has nothing include-related. Worth a fresh issue if it is still wanted.
- ember-learn/guides-source#2048 lives in another repo and tracks separately.

One small thing survived the supersede. `warp-drive-packages/core/src/store/-private/store-service.ts:1119` reads "a list **or** relationship names" where the matching `findRecord` paragraph correctly reads "of". That is a one-word fix on `main`, not a reason to rebase a branch that is two years behind and conflicting against a repository that has since restructured its package layout.

PR #9510, opened 2024-07-11
Backport #9511 to lts-4-12: Add @ember/string v4 to peerDependencies

SergeAstapov:
## Description

@ember/string v4 was recently published https://github.com/emberjs/ember-string/releases/tag/v4.0.0-%40ember%2Fstring

it's not possible to upgrade as `@ember-data/store` still requires v3.

### DIFF
--- a/packages/-ember-data/package.json
+++ b/packages/-ember-data/package.json
@@ -17,7 +17,7 @@
   "author": "",
   "license": "MIT",
   "peerDependencies": {
-    "@ember/string": "^3.0.1"
+    "@ember/string": "^3.0.1 || ^4.0.0"
   },
   "dependencies": {
     "@ember-data/adapter": "workspace:4.12.8",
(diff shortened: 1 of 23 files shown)


anandfresh (approved):

PavinStrickland:
The error message indicates that the self-installer failed with code 1, which could be the cause of the setup job failure. To resolve this, you may need to investigate the pnpm/action-setup step and check for any issues or misconfigurations in the self-installer script.

PavinStrickland:
Also, the checks for lint, special-build-tests, and Test Chrome have failed shortly after beginning, which might suggest a common issue affecting them. Given these checks failed quickly, it might be due to misconfiguration, missing dependencies, or other immediate errors in the setup. Investigating the details of these failed checks could offer insights into the cause of the failures and guide necessary fixes. The 'lts' and 'releases' checks being skipped should be reviewed to understand why they were not executed. The fact that there are no conflicts with the base branch means that once the issues causing the failures are resolved, the pull request should be mergeable.

PavinStrickland:
jobs:
       lint:
         runs-on: ubuntu-latest

         steps:
           - name: Check out repository
             uses: actions/checkout@v3

           - name: Set up Node.js
             uses: actions/setup-node@v2
             with:
               node-version: '16.x'
         
           - name: Run custom setup
             run: ./.github/actions/setup
             continue-on-error: false # ensure the job stops if this step fails

           - name: Lint JavaScript
             if: success()
             run: npm run lint:js

wagenet (approved):

PR #9511, opened 2024-07-11
Add @ember/string v4 to peerDependencies

SergeAstapov:
## Description

@ember/string v4 was recently published https://github.com/emberjs/ember-string/releases/tag/v4.0.0-%40ember%2Fstring

it's not possible to upgrade as `@ember-data/store` still requires v3.

### DIFF
--- a/packages/request-utils/package.json
+++ b/packages/request-utils/package.json
@@ -52,7 +52,7 @@
     "version": 2
   },
   "peerDependencies": {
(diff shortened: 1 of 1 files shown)


SergeAstapov:
@runspired my assumption is that if it's optional and not present, then we are ok. but if it's present but v4 - then it's a trouble

PR #9517, opened 2024-07-30
Allow `@ember/string` v4 as dependency & allow `ember-inflector` v5

mkszepp:
- Some week ago there was released `@ember/string` v4. We should also allow v4 as peerDependency
- Allow `ember-inflector` v5 as peerDependency

### DIFF
--- a/packages/request-utils/package.json
+++ b/packages/request-utils/package.json
@@ -52,9 +52,9 @@
     "version": 2
   },
   "peerDependencies": {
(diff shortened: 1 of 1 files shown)


runspired:
cam you PR this to main instead?

mkszepp:
suppressed with #9519

as note: ember/string v4 was already added on main, there was not anymore necessary

PR #9525, opened 2024-08-27
Allow `@ember/test-helpers` v4 as peerDependency

mkszepp:
This PR allows consumer apps also `@ember/test-helpers` v4 version (test-helper is now a v2 addon)


anandfresh (approved):

mkszepp:
@runspired can we get a review also for this one?

Baltazore:
@runspired I don't really understand why tests are failing on master. Preston mentioned it might pick up wrong version of test helpers.

mkszepp:
The reason why it was still using v3.x of test-helper is, that there is a override in `package.json` and a patch for v3... Its not so nice to make a patch for a active addon which is provided from ember team.

I have now created a new patch for test-helper v4 and updated on all points to v4.

After this change, we were running into `ember-cli-fastboot-testing` error with test-helper, because there is missing the peerDependency (opend a issue https://github.com/embermap/ember-cli-fastboot-testing/issues/872)

Preston has told me, that its possible fixed with updating all `@embroider` & `auto-import` patches to latest, but it wasn't... so i have added a patch for now

mkszepp:
[https://github.com/embermap/ember-cli-fastboot-testing/releases/tag/v0.6.2](ember-cli-fastboot-testing) has added now peerDependency of `@ember/test-helper`. Patching ember-cli-fastboot-testing is now not anymore necessary (already changes in this PR)

Baltazore:
@mkszepp thank you for working on that !

mkszepp:
[ember-provide-consume-context](https://github.com/customerio/ember-provide-consume-context/releases/tag/ember-provide-consume-context%400.4.0) supports now also v4.0.0 of `test-helpers`... now all packages are supporting v3 and v4 of `test-helpers`

For `test-helpers` there is now no override anymore necessary. Only the patches are still needed like there were before.

@Baltazore @runspired can we get a review and release with this changes?

packages/-ember-data/package.json
```diff
@@ -125,7 +125,7 @@
     "@warp-drive/build-config": "workspace:0.0.0-alpha.46"
   },
   "peerDependencies": {
-    "@ember/test-helpers": "^3.3.0",
+    "@ember/test-helpers": "^3.3.0 || ^4.0.2",
```
runspired, line 128: probably this should be `^4.0.4` ?

package.json
```diff
@@ -157,7 +157,8 @@
     "patchedDependencies": {
       "qunit@2.19.4": "patches/qunit@2.19.4.patch",
       "testem@3.11.0": "patches/testem@3.11.0.patch",
-      "@ember/test-helpers@3.3.0": "patches/@ember__test-helpers@3.3.0.patch"
+      "@ember/test-helpers@3.3.0": "patches/@ember__test-helpers@3.3.0.patch",
```
runspired, line 160: do we still need both?

packages/ember/package.json
```diff
@@ -45,7 +45,7 @@
     "@ember-data/request": "^4.12.7 || > 5.3.3",
     "@ember-data/request-utils": "^4.12.7 || > 5.3.3",
     "@warp-drive/core-types": "workspace:0.0.0-alpha.95",
-    "ember-provide-consume-context": "^0.3.1",
+    "ember-provide-consume-context": "^0.3.1 || ^0.4.0",
```
runspired, line 48: we can just shift this to `^0.4.0` since this package is still in alpha mode

packages/unpublished-test-infra/package.json
```diff
@@ -75,7 +75,7 @@
     "@ember-data/tracking": "workspace:5.4.0-alpha.109",
     "@warp-drive/diagnostic": "workspace:0.0.0-alpha.95",
     "@warp-drive/core-types": "workspace:0.0.0-alpha.95",
-    "@ember/test-helpers": "^3.3.0"
+    "@ember/test-helpers": "^3.3.0 || ^4.0.2"
```
runspired, line 78:
```suggestion
    "@ember/test-helpers": "^3.3.0 || ^4.0.4"
```

runspired (approved):
overall seems good, little bit of cleanup to do

package.json
mkszepp: @runspired yes we need both, since alle patches will be added into v3 & v4 of test-helpers...

runspired (approved):

package.json
runspired: the answer here is "we only need the 3.x patch because of the ember-try scenario for 3.28", I don't want to remove that scenario just yet so left it for now

PR #9529, opened 2024-08-30
Widen inflector range

NullVoxPopuli:


NullVoxPopuli:
ope, nevermind: https://github.com/emberjs/data/pull/9519/files

PR #9531, opened 2024-09-01
feat: adds AliasField to allowed schema fields

runspired:
also fixes several issues around writes with SchemaObject and ManagedObject, and improves some factoring

### DIFF
--- a/tests/warp-drive__schema-record/tests/reads/alias-test.ts
+++ b/tests/warp-drive__schema-record/tests/reads/alias-test.ts
@@ -1,0 +1,114 @@
+import { module, test } from 'qunit';
+
+import { setupTest } from 'ember-qunit';
+
+import { recordIdentifierFor } from '@ember-data/store';
+import type { StableRecordIdentifier } from '@warp-drive/core-types';
+import { Type } from '@warp-drive/core-types/symbols';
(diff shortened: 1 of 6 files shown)


tests/warp-drive__schema-record/tests/reads/alias-test.ts
```diff
@@ -0,0 +1,116 @@
...
+    );
+
+    const record = store.createRecord<User>('user', {
+      rawNetWorth: '1000000.009',
+    });
+    const identifier = recordIdentifierFor(record);
+    const resource = store.cache.peek(identifier)!;
+
+    assert.strictEqual(record.rawNetWorth, '1000000.009', 'netWorth is accessible in raw form');
+    assert.strictEqual(record.netWorth, 1_000_000.009, 'netWorth is accessible in numeric form');
```
Baltazore, line 81: Hey I might be missing something, but I would expect `netWorth` to be 1_000_000.01 as it has transformer in alias setup

runspired: hydration doesn't change the value in the transform we registered, only serialization does. Since we set the value here via the raw form, we don't expect serialization to run. This test is basically checking that set via the original property is unaffected.

PR #9532, opened 2024-09-16
fix: Be more helpful when failing

MichalBryxi:
## Description

- Current message of "Error: Multiple versions of WarpDrive detected, the application will malfunction." is hard to action on as it does not give the user enough context on what it sees.
- Make it so that the user has _some_ idea what configuration made the app fail.

## Notes for the release

- When multiple versions of warp-drive are found, tell the user which ones while failing.


runspired (approved):

MichalBryxi:
Fixed the prettier error, but not sure what does the "Main / releases (ember-canary) (pull_request)" complain about? @runspired?

anandfresh (approved):

PR #9539, opened 2024-09-27
docs(cookbook): add two-store migration guide

Baltazore:
## Why

This PR sat for two years while the docs tree moved underneath it. `guides/cookbook/` became `guides/the-manual/cookbook/`, and `upgrading/v5/two-store-migration.md` landed and took ownership of mirror packages, store configuration, and the constraints of running two copies of the library.

The strategy is still worth documenting. It is now the recommended v5 migration path, and mirror packages still ship. What was missing then and is still missing now is the application half, meaning what your components do once the second store exists. So the page keeps only that, links to the upgrading section for everything it already owns, and adopts the `v2-store` naming that section established.

## Scope

- New `guides/the-manual/cookbook/migrating-with-two-stores.md`. Route-level service swap, the vertical-slice rule, `@provide`/`@consume` from `ember-provide-consume-context` to carry the store through a subtree, and the shared-leaf-component boundary case.
- `upgrading/v5/two-store-migration.md` gains a link to the new page from its Caveat Emptor section.
- `guides/the-manual/cookbook/index.md` lists the new page, plus `basic-usage.md` and `auth-handlers.md`, which the index was already missing.
- Second commit closes an unterminated `ts` fence in `guides/the-manual/relational-data/configuration/many-to-many.md`, found by the fence check run over this change.

Out of scope: install and build configuration for mirror packages, which Steps 1 through 4 of the v4 to v5 guide already cover.

## Tradeoffs

The original draft opened with `ember install ember-data-mirror` and a `mirror-store` service. Both are replaced by links rather than updated names, because duplicating setup instructions across two pages is how they drift apart in the first place.

The `Model` to `ReactiveResource` case is no longer listed as a reason to run two stores. `useLegacyStore`'s `schemas` option lets both live in one store, and LegacyMode is the documented path, so the page now says so and points there.

## Verification

Checked every API in the samples against this tree. `record.save()` asserts when `ENABLE_LEGACY_REQUEST_METHODS` resolves and `store.findRecord` is deprecated until 6.0, so the samples use `store.request` with builders from `@warp-drive/utilities/json-api`. `store.createRecord` is synchronous. `@consume` returns `undefined` with no provider above it, so the shared component falls back to the service and the page says why.

Ran a fence, smart-quote, and link check over `guides`, `upgrading`, and `blog`. Zero problems. The same check reproduces two defects in the previous revision of this branch, a smart quote inside a JS fence and an unclosed `hbs` fence, and found the `many-to-many.md` fence fixed here.

Rendering is confirmed by the deploy preview rather than locally.


runspired (approved):
Broad strokes seems good.

I wrote a tool that generates a package for this for AuditBoard to make the migration easier. I suspect I should make an OSS version of the tool as it makes managing this much simpler, especially for monorepos, if I do we can update this.

Baltazore:
@runspired I've validate agains new ember app and guide holds up, do you think we merge it or we do not want to encourage people to use that path ?

guides/the-manual/cookbook/migrating-with-two-stores.md
```diff
@@ -0,0 +1,205 @@
+# Migrating with two stores
+
+- ⮐ [Cookbook](./index.md)
```
runspired: in follow up lets add a "migrating from 1.x / 2.x / 3.x / 4.x ember-data to latest WarpDrive" that suggests using 2 stores. since the `ember-data` package (and @ember-data packages) do not overlap with the modern warp-drive/@warp-drive packages these migrations don't even need to use the mirror packages - with one exception: if the lower version is 4.13-canary then the modern version needs to use mirror packages.

runspired (approved):

guides/the-manual/cookbook/migrating-with-two-stores.md
Baltazore: Folded this into this PR rather than a follow-up (4340487f7d). The page has a new "Coming from ember-data 4.12 or earlier" section. It says 1.x through 4.12 can go straight to the latest WarpDrive in a second store with plain `@warp-drive/*`, because those releases only publish `ember-data` / `@ember-data/*`. It sends readers to the Migration steps of the 4.x → 5.x guide, swapping each `@warp-drive-mirror/*` name for `@warp-drive/*`. 4.13 is called out as the exception that still needs the mirrors: its canaries already depend on `@warp-drive/core-types` and `@warp-drive/build-config`. I checked that against the tags: `v4.12.8` has no `@warp-drive/*` dependency and `v4.13.0-alpha.9` has both.

---

guides/the-manual/cookbook/migrating-with-two-stores.md
```diff
@@ -0,0 +1,231 @@
...
```
runspired: this manual nav link probably doesn't make sense to keep anymore considering how major navigation concerns are handled by the documentation site itself instead of these being standalone markdown in the repo

guides/the-manual/cookbook/migrating-with-two-stores.md
```diff
@@ -0,0 +1,231 @@
...
+`v2-store` to `store` so injections and providers collapse back to the default.
+[Post Migration](/upgrading/v5/index.md#post-migration) covers the version-upgrade case.
+
+Because each store is configured independently, the second store is also where you drop what you
+no longer want. Adapters, serializers, and `Model` can stay in the old store and never be installed
+in the new one.
+
+---
+
+- ⮐ [Cookbook](./index.md)
```
runspired: same as above, remove this

runspired (changes requested):

guides/the-manual/cookbook/migrating-with-two-stores.md
Baltazore: Removed in 8e397b4a21, both this one and the footer link with its `---` rule.

---

guides/the-manual/cookbook/migrating-with-two-stores.md
Baltazore: Removed in 8e397b4a21, along with the `---` rule above it.

---

guides/the-manual/cookbook/migrating-with-two-stores.md
Baltazore: Update, ae30379afa: the 4.12-or-earlier guidance moved out of the cookbook. It now lives on `upgrading/v5/two-store-migration.md` as "Coming from ember-data 4.12 or earlier", under Mirror Versions, with the 4.13 exception. Step 1 of the 4.x → 5.x guide has a tip pointing there, since the 4.6 and 4.12 apps that guide targets are exactly the ones that can skip mirrors.

The cookbook page is now "Migrating route by route" (`migrating-route-by-route.md`), so the title makes sense to readers who don't already know the two-store approach. It keeps the app-side patterns and has a one-paragraph pointer for which packages to install.

---

Baltazore: Final layout, 6172acb357: this is now its own Upgrading section, `upgrading/v1-v4/` ("1.x – 4.x → 5.x"), next to 4.x → 5.x and shaped the same way. The page is the route-by-route guide aimed at 1.x–4.12 apps. It installs `@warp-drive/*` directly, follows the 4.x → 5.x Migration steps with the mirror names swapped, and has a one-line note sending 4.13 to the mirror path. The v5 two-store page and Step 1 of the v5 guide each carry a short tip pointing there, and the page is no longer in the cookbook. The PR title and description are updated to match.

---

PR #9543, opened 2024-10-01
Improvement: Setting Up The Project

leoeuclids:
## Description

This PR aims to improve the `setting-up-the-project.md` file, based on our findings during the setup of the project on my local environment.

1. Moves to recommending the usage of `pnpx` instead of `npx`
  . Since the project relies on pnpm, recommend running scripts with pnpm commands
  . The only script affected is the holodeck ensure-cert
2. It adds steps related to holodeck certificates
  . Install `mkcert` and `nss` (as needed)
  . Run the ensure-cert script to create the certificates
3. Remove the env manipulation from the holodeck `index.js` file
  . Rely on running the ensure-cert script instead
4. Add more granular log and error messages

### DIFF
--- a/packages/holodeck/server/index.js
+++ b/packages/holodeck/server/index.js
@@ -35,24 +35,13 @@ function getShellConfigFilePath() {
 function getCertInfo() {
   let CERT_PATH = process.env.HOLODECK_SSL_CERT_PATH;
   let KEY_PATH = process.env.HOLODECK_SSL_KEY_PATH;
-  const configFilePath = getShellConfigFilePath();
 
-  if (!CERT_PATH) {
-    CERT_PATH = path.join(homedir(), 'holodeck-localhost.pem');
-    process.env.HOLODECK_SSL_CERT_PATH = CERT_PATH;
-    execSync(`echo '\nexport HOLODECK_SSL_CERT_PATH="${CERT_PATH}"' >> ${configFilePath}`);
-    console.log(`Added HOLODECK_SSL_CERT_PATH to ${configFilePath}`);
-  }
-
-  if (!KEY_PATH) {
-    KEY_PATH = path.join(homedir(), 'holodeck-localhost-key.pem');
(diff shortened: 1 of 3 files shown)


runspired (approved):
Thank so much! ❤️

packages/holodeck/server/index.js
```diff
@@ -5,54 +5,27 @@ import { Hono } from 'hono';
...
-
 function getCertInfo() {
   let CERT_PATH = process.env.HOLODECK_SSL_CERT_PATH;
   let KEY_PATH = process.env.HOLODECK_SSL_KEY_PATH;
-  const configFilePath = getShellConfigFilePath();
 
-  if (!CERT_PATH) {
-    CERT_PATH = path.join(homedir(), 'holodeck-localhost.pem');
-    process.env.HOLODECK_SSL_CERT_PATH = CERT_PATH;
-    execSync(`echo '\nexport HOLODECK_SSL_CERT_PATH="${CERT_PATH}"' >> ${configFilePath}`);
```
runspired, line 43: IIRC the setup here exists for CI environments.

We should probably gate it with `process.env.CI` instead of removing it (this is the reason for the failing test scenarios in this PR), though there may be some other nuance that allows us to still remove this but make CI work.

I *think* the issue is that we don't re-source the ENV after running ensure-cert (see https://github.com/emberjs/data/blob/536192b15822c49226b462a905fa39e47a790fea/.github/actions/setup/action.yml#L75) so the paths aren't seen even though they are already present

anandfresh (approved):

packages/holodeck/server/index.js
leoeuclids: Yeah, I tried adding the source command to the ensure-cert script, but was unable to make it work, probably because the script is running on a child process.

Since the CI setup is running the ensure-cert script, I'll attempt to add the re-sourcing step to it.

leoeuclids: Added back the default values, only removing the bash file editing, so it should work for CI and locally as well.

PR #9581, opened 2024-10-21
Update docs to ensure we're consistent on async/await usage

acorncom:
## Description

Updated async/await usage for handlers


runspired (commented):
The lack of showing `async/await` is intentional as with handlers we want to discourage using it for anything but the leaf-most handler when possible.

PR #9614, opened 2024-12-06
allow @ember-data/request to work with test-waiters@4

mansona:
## Description

This allows @ember-data/request (and therefore ember-data) to be used with test-waiters@4 which was recently converted to a v2 addon. This is a very naive implementation so please feel free to correct me (or push any fixes) to get this working right 👍  

## Notes for the release

Allow @ember-data/request to be used with @ember/test-waiters v4

PR #9619, opened 2024-12-15
Support @ember/test-waiters v4

NullVoxPopuli:
## Description

v4 is a v2 addon

## Notes for the release

Widen support of `@ember/test-waiters` to include v3 and v4


anandfresh (approved):

runspired:
this should replace #9614 but iirc there should be a lockfile change here too?

NullVoxPopuli:
I ran pnpm locally, and it did not generate one -- I suspect because y || x still satisfies y? :shrug:

PR #9621, opened 2024-12-18
Fix vite app via fixin' deps

NullVoxPopuli:

### DIFF
--- a/package.json
+++ b/package.json
@@ -144,7 +144,6 @@
     },
     "overrides": {
       "ember-auto-import": "^2.10.0",
-      "@embroider/macros": "^1.16.6",
       "broccoli-funnel": "^3.0.8",
       "broccoli-merge-trees": "^4.2.0",
       "@glimmer/validator": "^0.92.3",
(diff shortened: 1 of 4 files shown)


tests/vite-basic-compat/tsconfig.json
NullVoxPopuli, line 1: without the changes there are types-squigglies in the app

package.json
```diff
@@ -144,7 +144,6 @@
     },
     "overrides": {
       "ember-auto-import": "^2.10.0",
-      "@embroider/macros": "^1.16.6",
```
NullVoxPopuli, line 147: without this change the wrong version of macros ends up being used (along with the wrong shared-internals, etc)

tests/vite-basic-compat/package.json
```diff
@@ -119,7 +119,7 @@
     "ember-load-initializers": "^3.0.1",
     "ember-modifier": "^4.2.0",
     "ember-page-title": "^8.2.3",
-    "ember-qunit": "8.0.2",
```
NullVoxPopuli, line 122: v8 used AMD to load tests via the TestLoader

PR #9622, opened 2024-12-19
[NEVER MERGE] v4-main delta comparison

runspired:
This PR exists purely to compare the delta between the main branch and the v4-main branch, so as to eliminate any unexpected differences before we continue with releasing 4.3 and 5.4


packages/store/src/-private/record-arrays/identifier-array.ts
```diff
@@ -396,6 +446,7 @@ export class IdentifierArray<T = unknown> {
 
         const original: StableRecordIdentifier | undefined = target[index];
         const newIdentifier = extractIdentifierFromRecord(value);
+        // FIXME this line was added on main and I'm not sure why
```
runspired, line 449: investigate v4-main for this

PR #9624, opened 2024-12-20
feat(cache): add peekRemoteState to cache to view remote state

richgt:
## Description

PeekRemoteState gives the remote view of an object. 

## Notes for the release

- [x] update comments
- [x] confirm vite config should be updated.


packages/graph/src/-private/edges/collection.ts
```diff
@@ -68,12 +68,15 @@ export function createCollectionEdge(definition: UpgradedMeta, identifier: Stabl
...
+export function legacyGetCollectionRelationshipData(
+  source: CollectionEdge,
+  getRemoteState: boolean
+): CollectionRelationship {
   source.accessed = true;
   const payload: CollectionRelationship = {};
 
   if (source.state.hasReceivedData) {
-    payload.data = computeLocalState(source);
+    payload.data = getRemoteState ? source.remoteState.slice() : computeLocalState(source);
```
runspired, line 79: I love that this means we dont flush the computation unless needed <3

runspired (approved):
Generally this looks good. I suspect we can simplify the JSON:API cache implementation but I'd need to pull it down and work through it a bit to be sure. "works as expected" (at least according to the tests which look good) is more than good enough for the first implementation.

Because in SchemaRecord these code paths will be the default (while in Model the mutated code paths were the default) I suspect we will want to make sure to do an optimization pass soon.

packages/graph/src/-private/edges/resource.ts
```diff
@@ -40,16 +40,19 @@ export function createResourceEdge(definition: UpgradedMeta, identifier: StableR
...
   let data: StableRecordIdentifier | null | undefined;
   const payload: ResourceRelationship = {};
-  if (source.localState) {
+  if (getRemoteState && source.remoteState) {
+    data = source.remoteState;
+  } else if (!getRemoteState && source.localState) {
     data = source.localState;
   }
-  if (source.localState === null && source.state.hasReceivedData) {
+  if (((getRemoteState && source.remoteState === null) || source.localState === null) && source.state.hasReceivedData) {
```
robbytx, line 52: @richgt in my testing (on 4.13, backported from 5.3), this line seems incorrect in the situation where a relationship was non-null, but is set to null locally and not yet saved.  I think the line should be:
```
if ((getRemoteState ? source.remoteState : source.localState) === null) && source.state.hasReceivedData) {
```

runspired: assuming:
- `remoteState !== null`
- `localState === null`

for `getRemoteState === true`

- line 48 gives us the remote value
- check on line 52 becomes: (((true && false) || true) && true) which evals to `true`
- we assign `null` to the data incorrectly

for `getRemoteState === false`

- conditions on lines 47 and 49 are false
- line 52 evals to (((false && false) || true) && true) which evals to `true`
- we assign `null` to the data correctly

I believe the mistake here is actually that the check on line 52 is an `if` instead of an `else if`

PR #9654, opened 2025-02-07
perf: dont notify for attrs on insertions or committed changes that are referentially equal

runspired:
While debugging 4.13 with @gitKrystan I noticed that the attribute notifications were noisier than I would have expected, this will reduce that noise especially for newly created records.

This PR also fixes a few invariant method signatures and drops use of the global IDENTIFIERS set as it is no longer required.

This PR also improves the logging output, and updates our perf-test harness. However, due to an issue somewhere with chrome+tracerbench I've not been able to get it to run yet.

### DIFF
--- a/packages/core-types/src/identifier.ts
+++ b/packages/core-types/src/identifier.ts
@@ -2,13 +2,19 @@
   @module @ember-data/store
 */
 
+import { DEBUG } from '@warp-drive/build-config/env';
+
 // provided for additional debuggability
 export const DEBUG_CLIENT_ORIGINATED: unique symbol = Symbol('record-originated-on-client');
 export const DEBUG_IDENTIFIER_BUCKET: unique symbol = Symbol('identifier-bucket');
 export const DEBUG_STALE_CACHE_OWNER: unique symbol = Symbol('warpDriveStaleCache');
 
+function ProdSymbol<T extends string>(str: T): T {
+  return DEBUG ? (Symbol(str) as unknown as T) : str;
+}
+
 // also present in production
-export const CACHE_OWNER: unique symbol = Symbol('warpDriveCache');
+export const CACHE_OWNER = ProdSymbol('__$co');
 
 export type IdentifierBucket = 'record' | 'document';
 
--- a/packages/json-api/src/-private/cache.ts
+++ b/packages/json-api/src/-private/cache.ts
@@ -516,7 +516,7 @@ export default class JSONAPICache implements Cache {
     data: ExistingResourceObject,
     calculateChanges?: boolean
   ): void | string[] {
-    let changedKeys: string[] | undefined;
+    let changedKeys: Set<string> | undefined;
     const peeked = this.__safePeek(identifier, false);
     const existed = !!peeked;
     const cached = peeked || this._createCache(identifier);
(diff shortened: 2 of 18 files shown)


packages/core-types/src/identifier.ts
runspired, line 1: the basic idea here is that symbol keys are slower so in Prod we use a less safe string key

packages/json-api/src/-private/cache.ts
```diff
@@ -525,38 +526,52 @@ export default class JSONAPICache implements Cache {
...
-      this._capabilities.notifyChange(identifier, 'state');
+      this._capabilities.notifyChange(identifier, 'identity', null);
+      this._capabilities.notifyChange(identifier, 'state', null);
     }
 
-    if (calculateChanges) {
-      changedKeys = existed ? calculateChangedKeys(cached, data.attributes) : Object.keys(data.attributes || {});
+    // if no cache entry existed, no record exists / property has been accessed
+    // and thus we do not need to notify changes to any properties.
+    if (calculateChanges && existed && data.attributes) {
```
runspired, line 558: this turns out to not be true for "emergent behaviors" in prod when belongsTo relationships are illegally accessed before load.

we fixed this by making the consuming record pay the cost for this by subscribing to its own "added" event and invalidating its attrs when so.

packages/json-api/src/-private/cache.ts
```diff
@@ -567,11 +582,11 @@ export default class JSONAPICache implements Cache {
       setupRelationships(this.__graph, this._capabilities, identifier, data);
     }
 
-    if (changedKeys && changedKeys.length) {
+    if (changedKeys?.size) {
       notifyAttributes(this._capabilities, identifier, changedKeys);
     }
 
-    return changedKeys;
+    return changedKeys?.size ? Array.from(changedKeys) : undefined;
```
runspired, line 589: we should consider updating the signature of this from array to set

packages/json-api/src/-private/cache.ts
```diff
@@ -1661,14 +1676,18 @@ function getDefaultValue(
...
   if (!keys) {
-    storeWrapper.notifyChange(identifier, 'attributes');
+    storeWrapper.notifyChange(identifier, 'attributes', null);
     return;
   }
 
-  for (let i = 0; i < keys.length; i++) {
-    storeWrapper.notifyChange(identifier, 'attributes', keys[i]);
+  for (const key of keys) {
+    storeWrapper.notifyChange(identifier, 'attributes', key);
```
runspired, line 1690: it may be worth adding a batch signature so we just forward these as a single batch. We already have the set and it would avoid N (record)*M (field) notifications in favor of just N.

packages/model/src/-private/notify-changes.ts
```diff
@@ -16,26 +16,33 @@ export default function notifyChanges(
...
-      notifyRelationship(identifier, key, record, meta);
-    } else {
-      record.eachRelationship((name, meta) => {
-        notifyRelationship(identifier, name, record, meta);
-      });
-    }
-  } else if (value === 'identity') {
-    record.notifyPropertyChange('id');
+  switch (value) {
+    case 'added':
```
runspired, line 20: "added" is a scenario a record will never encounter unless it manages to get created in a prod build when no data exists for it through an illegal sync-relationship access.

so by subscribing to it here, we invalidate any attribute caches once we have data available, and only pay the cost in records which have done this vs needing to notify of attribute changes on first-load

packages/store/src/-private/caches/identifier-cache.ts
```diff
@@ -39,11 +39,10 @@ import { hasId, hasLid, hasType } from './resource-utils';
 
 type ResourceData = unknown;
 
-const IDENTIFIERS = getOrSetGlobal('IDENTIFIERS', new Set());
```
runspired, line 42: we always were checking the CACHE_OWNER symbol path first so this was just dead-perf

PR #9680, opened 2025-02-12
chore: improve handling of rollback scenarios

runspired:


packages/graph/src/-private/-diff.ts
```diff
@@ -70,11 +75,43 @@ function _deprecatedCompare<T>(
...
+          }
+        } else {
+          if (i < priorLocalLength) {
+            const priorLocalMember = priorLocalState![j];
+            if (priorLocalMember !== member) {
+              changed = true;
+            }
+          } else {
+            changed = true;
+          }
```
anandfresh, line 101: @runspired Why do we need this if else when you are performing same operations inside ? 
Can we just do it with just one if condition like below 
```
if (DEPRECATE_RELATIONSHIP_REMOTE_UPDATE_CLEARING_LOCAL_STATE) {
if ((!remoteClearsLocal && i < priorLocalLength) || i < priorLocalLength || remoteClearsLocal ) {
           if (!changed && j < priorLocalLength) {
             const priorLocalMember = priorLocalState![j];
              changed  = priorLocalMember !== member ? true : false;
           }
         }
       }
}

```

packages/graph/src/-private/-diff.ts
```diff
@@ -143,11 +183,43 @@ function _compare<T>(
...
+        } else {
+          if (i < priorLocalLength) {
+            const priorLocalMember = priorLocalState![i];
+            if (priorLocalMember !== member) {
+              changed = true;
+            }
+          } else {
+            changed = true;
```
anandfresh, line 208: Even this if conditions should be optimised as like suggested above

anandfresh (changes requested):

tests/main/tests/integration/application-test.js
```diff
@@ -135,62 +132,3 @@ module('integration/application - Using the store as a service', function (hooks
     assert.notStrictEqual(store, secondService, 'the store can be used as a service');
   });
 });
-
-module('integration/application - Attaching initializer', function (hooks) {
```
gitKrystan, line 139: Are these test changes intentional?

runspired: these tests test a removed thing, by testing the removed thing they cause a deprecation because its not really there. They were unrelated but I was tired of the noise when running the test suite

packages/graph/src/-private/-diff.ts
runspired: `!remoteClearsLocal && ` is the difference. The way we approach deprecation flagging (so that deprecated code can be fully eliminated) is to wrap the full version of the deprecated code in its flag and the else be the full new version.

RE `changed  = priorLocalMember !== member ? true : false;`, we don't do this because once `changed` is true we never revert it to false.

anandfresh: @runspired ok , in that case make it as `changed  = priorLocalMember !== member && true ;`

PR #9698, opened 2025-02-24
fix: Don't notify changes for attributes not registered with the schema

gitKrystan:
## Description

Fixes a bug where we would notify changes to data that wasn't actually registered with the schema. This seems to only cause issues if the data is coincidentally tracked for some other reason (e.g. `id` or a tracked property with the same name).

NOTE:
Currently (including with this change), we do not discard this data in the cache. So when you get the resource data out of the cache, we include it. Since some APIs are dependent on payloads being complete when you send them back, some users may rely on the current behavior. So we don't delete the value itself so that it's still merged in. We should probably deprecate this eventually.

### DIFF
--- a/packages/json-api/src/-private/cache.ts
+++ b/packages/json-api/src/-private/cache.ts
@@ -560,7 +560,7 @@ export default class JSONAPICache implements Cache {
     // if no cache entry existed, no record exists / property has been accessed
     // and thus we do not need to notify changes to any properties.
     if (calculateChanges && existed && data.attributes) {
-      changedKeys = calculateChangedKeys(cached, data.attributes);
+      changedKeys = calculateChangedKeys(cached, data.attributes, this._capabilities.schema.fields(identifier));
     }
 
     cached.remoteAttrs = Object.assign(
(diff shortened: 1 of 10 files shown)


runspired (approved):

runspired (approved):

packages/json-api/src/-private/cache.ts
```diff
@@ -557,10 +557,12 @@ export default class JSONAPICache implements Cache {
       this._capabilities.notifyChange(identifier, 'state', null);
     }
 
+    const fields = this._capabilities.schema.fields(identifier);
```
gitKrystan, line 560: Technically I could only set this if one of the conditions below are met. Don't want to over-optimize

packages/model/src/-private/model.ts
```diff
@@ -1740,7 +1740,7 @@ class Model extends EmberObject implements MinimalLegacyRecord {
     this.eachComputedProperty((name, meta) => {
       if (isAttributeSchema(meta)) {
         assert(
-          "You may not set `id` as an attribute on your model. Please remove any lines that look like: `id: attr('<type>')` from " +
+          "You may not set 'id' as an attribute on your model. Please remove any lines that look like: `id: attr('<type>')` from " +
```
gitKrystan, line 1743: Making this error message match the other one so the tests don't care about where specifically the error is thrown.

tests/main/tests/unit/store/push-test.js
```diff
@@ -527,24 +527,6 @@ module('unit/store/push - Store#push', function (hooks) {
     assert.notOk(store._instanceCache.peek(pushResult, { bucket: 'record' }), 'record is not materialized');
   });
 
-  test('_push does not require a modelName to resolve to a modelClass', function (assert) {
-    const store = this.owner.lookup('service:store');
-    const originalCall = store.modelFor;
-    store.modelFor = function () {
```
gitKrystan, line 533: Chris says we don't need this optimization anymore.

PR #9702, opened 2025-02-27
Don't overnotify for updates to added state that match local updates

gitKrystan:

### DIFF
--- a/packages/graph/src/-private/-diff.ts
+++ b/packages/graph/src/-private/-diff.ts
@@ -1,6 +1,5 @@
 import { deprecate } from '@ember/debug';
 
-import { recordIdentifierFor } from '@ember-data/store';
 import {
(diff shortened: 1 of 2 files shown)


packages/graph/src/-private/-diff.ts
```diff
@@ -52,9 +54,17 @@ function _deprecatedCompare<T>(
         adv = true;
 
         if (!prevSet.has(member)) {
-          changed = true;
-          added.add(member);
-          onAdd(member);
+          if (!prevSet.has(member)) {
```
bertdeblock, line 57: `if (!prevSet.has(member)) {` is defined twice, is this intended?
Sorry for the out-of-the-blue review, stumbled upon this PR and thought I should ask / mention it.

gitKrystan: Def not intentional. Thanks for the catch.

PR #9754, opened 2025-03-10
Add: tests verify behavior mutating a record 

MehulKChaudhari:
## Description

Added tests to determine if mutating a related record from a checked-out record  
triggers an automatic checkout or results in an error

### DIFF
--- a/tests/warp-drive__schema-record/tests/polaris/basic-fields-field-read-test.ts
+++ b/tests/warp-drive__schema-record/tests/polaris/basic-fields-field-read-test.ts
@@ -225,9 +225,9 @@ module('Reads | basic fields', function (hooks) {
         id: '1',
         type: 'user',
         attributes: {
-          name: 'Rey Skybarker'
-        }
-      }
+          name: 'Rey Skybarker',
+        },
+      },
     });
 
     assert.strictEqual(immutableRecord.id, '1', 'id is accessible');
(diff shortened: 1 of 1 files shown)


tests/warp-drive__schema-record/tests/reads/basic-fields-test.ts
```diff
@@ -192,4 +193,24 @@ module('Reads | basic fields', function (hooks) {
...
+    registerDerivations(schema);
+
+    schema.registerResource(
+      withDefaults({
+        type: 'user',
+        fields: [{ name: 'name', kind: 'field' }],
+      })
+    );
+
+    const record = store.createRecord('user', { name: 'Rey Skybarker' }) as User;
```
runspired, line 209: when we call `createRecord` we actually SHOULD be editable as this is explicitly asking for a new record in an editable state.

To test that the field is immutable, use `push` to add the record to the cache and get a record back.

```ts
store.push<User>({
  data: {
    id: '1',
    type: 'use',
    attributes: {
      name: 'Rey Skybarker'
    }
  }
});
```

insider baseball: I use `Rey Skybarker` when writing tests as thats my dog. I always recommend contributors immortalize their own pets/friends/family/things as desired 😍

MehulKChaudhari: Thanks for explaining! I'll fix the tests. 
Oh, I had no idea that was your dog's name—love that!

runspired (approved):

PR #9761, opened 2025-03-12
chore: reduce simple Map/Set ops

runspired:


packages/json-api/src/-private/cache.ts
runspired, line 1: for small arrays and sets, checking value directly is 30% faster (or more)

packages/store/src/-private/managers/notification-manager.ts
runspired, line 1: the change here is that we use the callback function itself as the token, stashing a bit of info directly on it. This eliminates the double map lookup and simplifies iteration.

We also convert CacheOperations (a simple set) into a direct value comparison.

PR #9776, opened 2025-03-21
chore: remove some deprecations in `lts-3-28`

yangpiao:
## Description

- `@ember/error': `EmberError` -> `Error`
- `@ember/polyfills': `assign` -> `Object.assign`
- guard `application.inject` with ember-compatibility-helpers

We'd like to use ember-data 3.28 with ember-source 5.12 and in my testing these are the minimum change to achieve it.


runspired (approved):

packages/-ember-data/tests/acceptance/relationships/has-many-test.js
runspired, line 1: check how we managed this on main as we have to support 3.28 -> as far as we can go (and for this branch ~2.x -> as far as we can go)

https://github.com/emberjs/data/pull/9773

yangpiao: Thanks for the tip. I've updated the code

packages/-ember-data/tests/integration/application-test.js
```diff
@@ -1,7 +1,7 @@
 import Application from '@ember/application';
 import Namespace from '@ember/application/namespace';
 import Controller from '@ember/controller';
-import Service, { inject as service } from '@ember/service';
```
runspired, line 4: I suspect this needs to drop the `Service, ` portion and do `const Service = s.default` below

packages/store/index.js
```diff
@@ -8,6 +8,15 @@ const addonBaseConfig = addonBuildConfigForDataPackage(name);
 
 module.exports = Object.assign({}, addonBaseConfig, {
   shouldRollupPrivate: true,
+
+  included() {
```
runspired, line 12: note for myself: before merging check this works correctly given what we do in addonBaseConfig

PR #9880, opened 2025-03-28
feat: @warp-drive/core/{build-config,types}

runspired:
This is the beginning of figuring out the mechanics of implementing the new core package.

One of the harder threads to pull is the build-config. The docs are generated from code that is not exported, many of the modules are intended to be virtual, and it has a combination of raw cjs, compiled to cjs, and esm modules.

I've been playing with three potential directions:

A) build-config going away in favor of just having @warp-drive/core
- upside all in one place
- downside | impossible as the individual packages would have a cycle

B) build-config re-exporting from @warp-drive/core
- upside, single source of truth
- downside | impossible as the individual packages would have a cycle

C) core re-exporting from @warp-drive/build-config
- upside, no cycles.
- downside, multiple sources of truth, docs must live in different files in core than where they live in @warp-drive/build-config

(C) Is the current implementation in this PR


warp-drive-packages/core/src/build-config/env.ts
```diff
@@ -0,0 +1,6 @@
+export const DEBUG: boolean = true;
+export const PRODUCTION: boolean = true;
+export const TESTING: boolean = true;
```
lifeart, line 3: this could be a `define` plugin config - https://vite.dev/config/shared-options.html#define (no need to use babel)

runspired: (but also the safety of having the import seems better)

lifeart: > (but also the safety of having the import seems better)

It's may leads to confusion, because we magically re-defining imported constant values in runtime.
If `@warp-drive` could run in browser without compilation, it totally make sense, but if we always expect to build it, it's better to explicitly define configurable parts, and `vite`, `webpack` and other bundlers support `define` for years (easy to debug, common pattern for external contributors, no confusion if code running without compilation).
In addition, modern tooling may have limited babel support because of custom rust/go js compilers, and reducing amount of babel plugins seems a pragmatic way to keep maintenance surface smaller.

lifeart: Sidenote: I had `define` plugin problem with predefined constants (in @glimmer/env) - https://github.com/lifeart/demo-ember-vite/blob/63519d5ddf32d91fc76ae9b6385bb25b5494a605/compat/glimmer-env/index.ts#L3 because (at least in vite) we could not override `Identifier`s to workaround it I have to re-export names to be able to re-define it by value:

```js
const ENV_DEBUG = true;
const ENV_CI = false;

export const DEBUG = ENV_DEBUG;
export const CI = ENV_CI;
``` 

As outcome of it, I think that having defined flags as part of codebase and re-defining it in babel plugin adding more complexity, and less benefits comparing to `define` approach.

runspired: I think you misunderstand how this works

runspired:
no longer needed

PR #9881, opened 2025-03-28
fix: graph isReordered should return false when localState is null

BoussonKarel:


packages/graph/src/-private/graph.ts
```diff
@@ -807,7 +807,11 @@ function isReordered(relationship: CollectionEdge): boolean {
   }
 
   const { remoteState, localState, additions, removals } = relationship;
-  assert(`Expected localSate`, localState);
+
+  if (localState === null) {
```
BoussonKarel, line 811: is this enough or do we also need to check if `relationship.state.hasReceivedData` is false? Because I saw this on the CollectionEdge interface:
image

tests/main/tests/integration/relationships/rollback-test.ts
runspired, line 1: we also need a test for rollback called when a relationship wasn't accessed, yeah?

packages/graph/src/-private/graph.ts
runspired: if we have never received data we also cant be reordered, so while `hasReceivedData` matters for isDirty, it doesn't matter for this.

in terms of `isDirty`, it doesn't matter if localState is out-of-sync if its currently null, as that still indicates it is in the same state as the remote once it eventually does sync.

Where it may matter is when localState is not null. It's possible we need to "flush" the changes here (or before we get here in the calling function) in order to do the re-order detection. I'd say punt on that question for the moment.

tests/main/tests/integration/relationships/rollback-test.ts
BoussonKarel: Well `pages` isn't accessed in the other 4 test cases. So it is tested (implicitly). Do you want me to explicitly test this? Maybe assert that is has not been accessed (how to check this?)

runspired: I think explicitly testing it via a descriptive test name and a comment that "hi this relationship exists and was never accessed + has remote state / does not have remote state" would be good

BoussonKarel: I've updated the tests

runspired (approved):

PR #9884, opened 2025-03-28
feat: cache patch operations for relationships and documents

runspired:
This builds upon #9883 to allow additional cache patch operations for relationships and documents

The implementation here is mostly complete, I've left an assert preventing removing documents from the cache because adding support for removing documents/requests requires a separate refactor to ensure we can cleanup reactive documents as well.

### Testing Status

| Passing | Failing | Broken | Under Construction | Skipped | Todo | Total |
| ------- | ------- | ------ | ------------------ | ------- | ---- | ----- |
| ✅ 30 | ❌ 0 | 💥 0 | 🚧 0 | ⚠️ 0 | 🛠️ 1 | 31 |

- 🛠️ We can remove a document from the cache
- ✅ We can add a resource to the cache
- ✅ We can perform bulk operations on the cache
- ✅ We can remove a resource from the cache
- ✅ We can update a resource in the cache
- ✅ We can update a single resource field (an attribute) in the cache
- ✅ We can update a single resource field (a relationship) in the cache
- ✅ We can add to a resource relationship in the cache
- ✅ We can remove from a resource relationship in the cache
- ✅ We can add to a collection relationship in the cache
- ✅ We can add multiple to a collection relationship in the cache
- ✅ We can add to a collection relationship in the cache with an index
- ✅ We can add multiple to a collection relationship in the cache with an index
- ✅ We can remove from a collection relationship in the cache
- ✅ We can remove multiple from a collection relationship in the cache
- ✅ We can remove from a collection relationship in the cache with an index
- ✅ We can remove multiple from a collection relationship in the cache with an index
- ✅ We can add to data on a single resource document in the cache
- ✅ We can remove from data on a single resource document in the cache
- ✅ We can add to data on a collection resource document in the cache
- ✅ We can add multiple to data on a collection resource document in the cache
- ✅ We can add to data on a collection resource document in the cache with an index
- ✅ We can add multiple to data on a collection resource document in the cache with an index
- ✅ We can remove from data on a collection resource document in the cache
- ✅ We can remove multiple from data on a collection resource document in the cache
- ✅ We can remove from data on a collection resource document in the cache with an index
- ✅ We can remove multiple from data on a collection resource document in the cache with an index
- ✅ We can add to included on a resource document in the cache
- ✅ We can add multiple to included on a resource document in the cache
- ✅ We can remove from included on a resource document in the cache
- ✅ We can remove multiple from included on a resource document in the cache


packages/graph/src/-private/graph.ts
runspired, line 1: the asserts here may be able to just be deleted, iirc we were guarding the feature's use but its actually implemented

packages/json-api/src/-private/cache.ts
```diff
@@ -405,13 +416,43 @@ export default class JSONAPICache implements Cache {
         console.log(`EmberData | Operation - patch ${op.op}`, op);
```
runspired, line 416: we should take the opportunity to update to the new logging syntax for operations

PR #9921, opened 2025-04-08
Support build-config being used outside of ember

NullVoxPopuli:
- If this PR updates API docs, preview them by running `bun preview-api-docs`
- Read the full [contributing documentation](https://github.com/emberjs/data/blob/main/contributing/become-a-contributor.md)


packages/build-config/src/index.ts
```diff
@@ -102,8 +102,9 @@ function recastMacrosConfig(macros: object): MacrosWithGlobalConfig {
   return macros as MacrosWithGlobalConfig;
 }
 
+export function setConfig(macros: object, config: WarpDriveConfig): void;
 export function setConfig(context: object, appRoot: string, config: WarpDriveConfig): void {
```
runspired, line 106: In order to use function overloading you need both signatures AND the implementation signature.  

```ts
setConfig(macros: object, config: WarpDriveConfig): void;
setConfig(context: object, appRoot: string, config: WarpDriveConfig): void;
export function setConfig(context: object, appRootOrConfig: string | WarpDriveConfig, config?: WarpDriveConfig): void {
```

NullVoxPopuli: yes, but I was lazy and only used the web ui haha

runspired: coulda still be done :P

PR #9452, opened 2024-05-22
feat: pagination data utils

runspired:
This PR introduces a suite of data-utilities for pagination to go along with `PromiseState / <Await />` and `RequestState / <Request />`

### Docs (in PR)

- [x] [<:idle>](https://github.com/emberjs/data/tree/pagination-utils/packages/ember#idle-is-an-additional-state)
- [ ] PaginationState
- [ ] getPaginationState
- [x] [\<Paginated />](https://github.com/emberjs/data/tree/pagination-utils/packages/ember#paginate-)
- [ ] \<EachLink />

---

### Implementation Tracking

- [ ] <:idle>
- [ ] PaginationState
- [ ] getPaginationState
- [ ] \<Paginated />
- [ ] \<EachLink />

---

### Below the Line Features

(features we intend to bake into these primitives but may not be added in this PR)
- PaginationState
  - [ ] Reload Strategy
  - [ ] SelectionState
  - [ ] Local state
  - [ ] Commit Changes
  
> [!Note]
> We do not intend for `<Paginate />` to be used to manage Tree and LinkedList 
> structures, which inherently have different requirements around async loading
> and pagination. We are planning a separate suite of data-utils for these structures
> in the near future.

PR #10014, opened 2025-05-27
Pagination utils

leoeuclids:
> [!Note]
> This PR is a migration from the work done on #9452

This PR introduces a suite of data-utilities for pagination to go along with `PromiseState / <Await />` and `RequestState / <Request />`

### Docs (in PR)

- [x] [<:idle>](https://github.com/emberjs/data/tree/pagination-utils/packages/ember#idle-is-an-additional-state)
- [x] PaginationState
- [x] getPaginationState
- [x] [\<Paginated />](https://github.com/emberjs/data/tree/pagination-utils/packages/ember#paginate-)
- [x] \<EachLink />

---

### Implementation Tracking

- [x] <:idle>
- [x] PaginationState
- [x] getPaginationState
- [x] \<Paginated />
- [x] \<EachLink />

---

### Below the Line Features

(features we intend to bake into these primitives but may not be added in this PR)
- PaginationState
  - [ ] Reload Strategy
  - [ ] SelectionState
  - [ ] Local state
  - [ ] Commit Changes
  
> [!Note]
> We do not intend for `<Paginate />` to be used to manage Tree and LinkedList 
> structures, which inherently have different requirements around async loading
> and pagination. We are planning a separate suite of data-utils for these structures
> in the near future.


warp-drive-packages/core/src/store/-private/new-core-tmp/paginated-state.ts
```diff
@@ -0,0 +1,119 @@
...
+    this.#pageList = {
+      prev: null,
+      self: getRequestState(request),
+      next: null,
+      isVirtual: false,
+    };
+  }
+
+  // TODO: pagination-utils Add loading state tracking to the ReactiveDocument interface
+  @cached
```
leoeuclids, line 55: Should we still use `@cached` and `@tracked` in this class?

warp-drive-packages/core/src/store/-private/new-core-tmp/paginated-state.ts
```diff
@@ -0,0 +1,119 @@
...
+ * const state = getPaginationState(future);
+ * ```
+ *
+ * @public
+ * @static
+ * @for @warp-drive/ember
+ * @param future
+ * @return {PaginationState}
+ */
+export function getPaginationState<T, RT extends ReactiveDocument<T[]>, E = ResourceErrorDocument>(
```
leoeuclids, line 105: The `RequestCacheRequestState` now required a "Error" type. Not sure if it should be deferred of defined here.

warp-drive-packages/core/src/store/-private/new-core-tmp/paginated-state.ts
runspired: no, peek at how the reactivity is handled for `PromiseState` and `RequestState` you'll want to mirror those.

In general, even within the `ember` package I want to avoid using tracked and cached in favor of our own signals approach to ensure broaded compat

warp-drive-packages/core/src/store/-private/new-core-tmp/paginated-state.ts
```diff
@@ -0,0 +1,119 @@
...
+
+  @cached
+  get isSuccess() {
+    return !this.isError;
+  }
+
+  // TODO: pagination-utils Add error state tracking to the ReactiveDocument interface
+  @cached
+  get isError() {
+    return this.pages.some((page) => page.isError);
```
leoeuclids, line 68: `isError` doesn't exist on `ReactiveDocument`. Should it be added there or we just check the `errors` property here?

warp-drive-packages/core/src/store/-private/new-core-tmp/paginated-state.ts
```diff
@@ -0,0 +1,119 @@
...
+      self: getRequestState(request),
+      next: null,
+      isVirtual: false,
+    };
+  }
+
+  // TODO: pagination-utils Add loading state tracking to the ReactiveDocument interface
+  @cached
+  get isLoading() {
+    return this.pages.some((page) => page.isLoading);
```
leoeuclids, line 57: `isLoading` doesn't exist in `ReactiveDocument` and the object also doesn't have reference to it's running promise.

leoeuclids:
@runspired It's green 🚀

runspired (approved):

PR #10028, opened 2025-06-18
docs: Unify naming for the tabs cases

MichalBryxi:
- Make it so that it's obvious what case we're talking about
- Follow the convention from [here](https://emberjs.github.io/data/guides/1-configuration/2-setup/1-universal)
- Fixes: #100027


runspired:
that conversation was mistaken

PR #10041, opened 2025-06-28
Add svelte signal configuration package

jacojoubert:

---

- Read the full [contributing documentation](https://github.com/emberjs/data/blob/main/contributing/become-a-contributor.md)


runspired (approved):
generally this seems like the right approach. We'll still want to implement the Request component, though with svelte having unfortunately deprecated?(demoted?) its slots API it may be more ergonomic to just use a RequestSubscription directly in svelte.

I'll get this rebased and bring it up to speed with the isolated-declarations work over the coming week.

anandfresh (approved):

runspired:
> What blocks us from using snippets? They dropped slots because they dropped web component support. As far as I know there isn't any downside to snippets? I can build out the request component.

The trouble with snippets is just the verbosity and the inverted control flow. By the time someone wires them up they may as well have just written their own if/else using the reactive SubscriptionState class directly. Reminds me of all the worst parts of contextual components in ember.

PR #10083, opened 2025-07-20
Fix deprecation handling for blueprint/tutorial

### LINKED ISSUE #9310: Deprecations will fire spuriously under strict ES module builds
Using ember-data 5.3.0 and given either:
 - the proposed [strict es modules RFC](https://github.com/emberjs/rfcs/pull/938) 
 - or an app built with the current Embroider main branch

ember-data emits deprecation warning for features that the app isn't actually using. 

```
DEPRECATION: You are relying on ember-data auto-magically installing the BooleanTransform. Use `export { BooleanTransform as default } from 'ember-data/serializer/transform';` in app/transforms/boolean.js instead [deprecation id: ember-data:deprecate-legacy-imports] This will be removed in ember-data 6.0.
DEPRECATION: You are relying on ember-data auto-magically installing the DateTransform. Use `export { DateTransform as default } from 'ember-data/serializer/transform';` in app/transforms/date.js instead [deprecation id: ember-data:deprecate-legacy-imports] This will be removed in ember-data 6.0.
DEPRECATION: You are relying on ember-data auto-magically installing the NumberTransform. Use `export { NumberTransform as default } from 'ember-data/serializer/transform';` in app/transforms/number.js instead [deprecation id: ember-data:deprecate-legacy-imports] This will be removed in ember-data 6.0.
DEPRECATION: You are relying on ember-data auto-magically installing the StringTransform. Use `export { StringTransform as default } from 'ember-data/serializer/transform';` in app/transforms/string.js instead [deprecation id: ember-data:deprecate-legacy-imports] This will be removed in ember-data 6.0.
```

Historically Ember has allowed "modules" to remain un-evaluated until they are synchronously required. That is not possible under true ES modules. So deprecations in module scope like here:

https://github.com/emberjs/data/blob/72b1e210091cdb59c61e911b50294b2a7fc941aa/packages/-ember-data/app/transforms/string.js#L5-L6

will fire before the app boots regardless of whether anybody is actually using that feature.

(One might also ask whether we can stop putting these modules in the build at all when they're not in use, and that's a good goal for the next-gen ember-data API, but these particular modules exist for backward compatibility and probably need to continue getting the backward-compatible behavior that pushes them into the build to make them synchronously available at runtime just in case they're used.)

runspired:
resolves #9310

### DIFF
--- a/packages/-ember-data/app/services/store.js
+++ b/packages/-ember-data/app/services/store.js
@@ -1,17 +1,23 @@
 import { deprecate } from '@ember/debug';
 
-export { default } from 'ember-data/store';
+import DataStore from 'ember-data/store';
 
-deprecate(
-  "You are relying on ember-data auto-magically installing the store service. Use `export { default } from 'ember-data/store';` in app/services/store.js instead",
-  false,
-  {
-    id: 'ember-data:deprecate-legacy-imports',
-    for: 'ember-data',
-    until: '6.0',
-    since: {
-      enabled: '5.2',
-      available: '4.13',
-    },
+export default class Store extends DataStore {
+  constructor(...args) {
+    super(...args);
+
+    deprecate(
+      "You are relying on ember-data auto-magically installing the store service. Use `export { default } from 'ember-data/store';` in app/services/store.js instead",
+      false,
+      {
+        id: 'ember-data:deprecate-legacy-imports',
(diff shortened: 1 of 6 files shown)


ef4:
Thanks for getting to this before I could. Looks good.

PR #10084, opened 2025-07-20
chore: scaffold react using vite

runspired:
doesn't seem at all like what we'd want, but its what vite outputs. Need a test app as well.

React doesn't offer any of the ergonomics features that component frameworks do like slots, so its not clear if there's a version of `<Request />` that actually makes sense ... though a class component using suspense under the hood + an error boundary could be pretty nifty. I'm thinking we could overload the children/Children API for this, though with jsx using props isn't too bad either

I'm roughly thinking the base API will be

```ts
import { useRequest } from '@warp-drive/react';

function MyComponent($props) {
  const { isLoading, isError, isSuccess, reason, value, future } = useRequest(
    findRecord('user', $props.id)
  );
}
```

If we do build a component API it would likely be something like this (with the states prop moved to being children if the children API approach isn't too brittle)

```tsx
import { Request } from '@warp-drive/react';

function MyComponent($props) {
  return (
    <Request
      query={findRecord('user', $props.id)}>
      states={{
        error: (error, state) => <MyError />,
        cancelled: (error, state) => <MyCancelled />,
        loading: (state) => <MyLoading />,
        content: (result, state) => <MyContent />,
        idle: () => <MyIdle />
      }}
    />
  );
}
```


runspired:
none of this PR is needed (except the notes in the description) since we landed react infra better suited to library development

PR #10094, opened 2025-07-31
fix: use HashFn for value comparisons in the cache when present for schema-objects

runspired:


warp-drive-packages/json-api/src/-private/cache.ts
```diff
@@ -1891,6 +1901,48 @@ function candidateKeys(
...
+/**
+ * Whether two values of `field` count as the same value for change detection.
+ *
+ * A `schema-object`, and each element of a `schema-array`, compares by the identity hash its
+ * `ObjectSchema` declares (`identity: { kind: '@hash', ... }`), so the schema decides what "same"
+ * means for it. With no hash declared, and for every other field kind, only the same reference
+ * counts. That knowingly over-notifies for equal-content objects: content equality is the schema's
+ * to define, not the cache's to guess.
+ */
+function attrValuesEqual(schema: SchemaService, field: FieldSchema, a: unknown, b: unknown): boolean {
```
gitKrystan: The one definition of "same value" for change detection, used by both `calculateChangedKeys` and `patchLocalAttributes` so a confirmed edit is recognized the same way a moved projection is. Dispatches on field kind: schema-object and schema-array go through the identity hash, everything else stays `===`. That fallback knowingly over-notifies for equal-content objects; per the original TODO here, content equality is the schema's to define, not the cache's to guess. An earlier revision on #11159 used a generic structural walk instead; it was dropped for exactly that reason.

warp-drive-packages/json-api/src/-private/cache.ts
```diff
@@ -1891,6 +1901,48 @@ function candidateKeys(
...
+  if (field.kind === 'schema-object') return schemaObjectsEqual(schema, field, a, b);
+  if (field.kind === 'schema-array') return schemaArraysEqual(schema, field, a, b);
+  return false;
+}
+
+/**
+ * A schema-array has no hash of its own: two arrays are the same when they have the same length
+ * and every element compares equal as a schema-object of the element type.
+ */
+function schemaArraysEqual(schema: SchemaService, field: SchemaArrayField, a: unknown, b: unknown): boolean {
```
gitKrystan: A schema-array has no hash of its own (`key: '@hash'` and `type: '@hash'` on the field both refer to the *element* type's hash), so array equality is length plus per-element `schemaObjectsEqual`. Split out from `attrValuesEqual` so the array and object rules sit side by side.

warp-drive-packages/json-api/src/-private/cache.ts
```diff
@@ -1891,6 +1901,48 @@ function candidateKeys(
...
+  if (!Array.isArray(a) || !Array.isArray(b) || a.length !== b.length) return false;
+  for (let i = 0; i < a.length; i++) {
+    if (!schemaObjectsEqual(schema, field, a[i], b[i])) return false;
+  }
+  return true;
+}
+
+function schemaObjectsEqual(
```
gitKrystan: Type first, then hash. Two elements of a polymorphic array with different concrete types are unequal before any hash runs, and the hash is looked up on the concrete type's `ObjectSchema`. `schemaObjectHash` returns `null` for `identity: null`, which lands in the `===` fallback via `hash !== null`.

warp-drive-packages/json-api/src/-private/cache.ts
```diff
@@ -1900,6 +1952,7 @@ function candidateKeys(
  * result depend on whether the caller had already mutated it.
  */
 function calculateChangedKeys(
+  schema: SchemaService,
```
gitKrystan: New first parameter. The cache already holds the `SchemaService` as `this._capabilities.schema`; passing it in keeps `calculateChangedKeys` a pure function of its inputs, which the existing doc block above it promises.

warp-drive-packages/json-api/src/-private/cache.ts
```diff
@@ -1942,8 +1995,8 @@ function calculateChangedKeys(
     const wasLocal = isEdited ? localEdit : isPromoted ? promoted[key] : wasRemote;
     const nowLocal = isEdited ? localEdit : nowRemote;
 
-    const remoteMoved = wasRemote !== nowRemote;
-    const localMoved = wasLocal !== nowLocal;
+    const remoteMoved = !attrValuesEqual(schema, field, wasRemote, nowRemote);
```
gitKrystan: This is the line the original TODO sat above. `wasRemote !== nowRemote` became `!attrValuesEqual(...)`; the four before/after values and the bucket logic are unchanged.

warp-drive-packages/json-api/src/-private/cache.ts
```diff
@@ -2059,8 +2117,10 @@ function patchLocalAttributes(cached: CachedResource, changedRemoteKeys?: Set<st
         : remoteAttrs && attr in remoteAttrs
           ? remoteAttrs[attr]
           : undefined;
+    const field = fields.get(attr);
 
-    if (existing === localAttrs[attr]) {
+    // the same equality `calculateChangedKeys` uses, so a confirmed edit is recognized the same way
```
gitKrystan: Without this, a nested local edit to a hashed schema-object could never be recognized as confirmed: the cloned local array is never `===` the pushed one, so the record stayed dirty forever after the server agreed. `fields.get(attr)` can be undefined for a key the schema does not know, hence the `===` branch.

warp-drive-packages/core/src/store/-private/utils/schema-object-identity.ts
```diff
@@ -0,0 +1,43 @@
...
+import type { ObjectSchema, SchemaArrayField, SchemaObjectField } from '../../../types/schema/fields.ts';
+import type { SchemaService } from '../../../types/schema/schema-service.ts';
+
+/**
+ * The concrete schema-object type behind a `schema-object` field, or one element of a
+ * `schema-array` field. Non-polymorphic fields name it directly; polymorphic fields read it off
+ * the raw value at `options.type` (default `'type'`), or compute it with the hash function named
+ * by `field.type` when `options.type` is `'@hash'`.
+ */
+export function resolveSchemaObjectType(
```
gitKrystan: Extracted, not new. `ManagedArray` and `getSchemaObjectField` each carried a copy of this (the `options.type` path, or a `HashFn` named by `field.type` when `options.type` is `'@hash'`). The cache needs the same rule to compare `schema-array` elements, so all three now share it. Under `store/-private` because both the reactive layer and json-api can import from there.

warp-drive-packages/core/src/store/-private/utils/schema-object-identity.ts
```diff
@@ -0,0 +1,43 @@
...
+  }
+  assert(`A non-polymorphic ${field.kind} field must provide a SchemaObject type in its definition`, field.type);
+  return field.type;
+}
+
+/**
+ * The identity hash an `ObjectSchema` declares for `rawValue`, or `null` when the schema declares
+ * none (`identity: null`), in which case callers fall back to referential identity.
+ */
+export function schemaObjectHash(schema: SchemaService, objectType: string, rawValue: object): string | null {
```
gitKrystan: Returns `null` rather than throwing for `identity: null`, so callers choose their own fallback: the cache falls back to `===`, `getSchemaObjectField` to `field.name`, and `ManagedArray` asserts (a `key: '@hash'` array whose element type has no hash is a schema mistake).

warp-drive-packages/core/src/reactive/-private/fields/managed-array.ts
```diff
@@ -189,34 +190,9 @@ export class ManagedArray {
...
-                typePath && objectType && typeof objectType === 'string'
-              );
-            }
-          } else {
-            assert(`A non-polymorphic SchemaArrayField must provide a SchemaObject type in its definition`, field.type);
-            objectType = field.type;
-          }
+          // schema-object identity is only required to be unique by type, so a polymorphic array
+          // resolves the concrete type first
+          const objectType = resolveSchemaObjectType(schema, field, rawValue as object);
```
gitKrystan: Behavior unchanged; the 22-line inline block this replaces moved to `resolveSchemaObjectType` verbatim, including both asserts.

warp-drive-packages/core/src/reactive/-private/fields/managed-array.ts
```diff
@@ -227,9 +203,9 @@ export class ManagedArray {
            */
           let schemaObjectKeyValue: string | number | object;
           if (KeyMode === '@hash') {
-            const hashField = schema.resource({ type: objectType! }).identity as HashField;
-            const hashFn = schema.hashFn(hashField);
-            schemaObjectKeyValue = hashFn(rawValue as object, hashField.options ?? null, hashField.name);
+            const hash = schemaObjectHash(schema, objectType, rawValue as object);
```
gitKrystan: Previously this cast `identity as HashField` and would have thrown inside `schema.hashFn` for an `identity: null` element type. Now it asserts with a message that names the type.

warp-drive-packages/core/src/reactive/-private/kind/schema-object-field.ts
```diff
@@ -39,35 +40,8 @@ export function getSchemaObjectField(context: KindContext<SchemaObjectField>): u
...
-    assert(`A non-polymorphic SchemaObjectField must provide a SchemaObject type in its definition`, field.type);
-    objectType = field.type;
-  }
-
-  const hashField = (schema.resource({ type: objectType }) as ObjectSchema).identity;
-  const identity = hashField
-    ? schema.hashFn(hashField)(rawValue, hashField.options ?? null, hashField.name)
-    : field.name;
+  const objectType = resolveSchemaObjectType(schema, field, rawValue);
+  const identity = schemaObjectHash(schema, objectType, rawValue) ?? field.name;
```
gitKrystan: Same fallback as before (`field.name` when no hash), via the shared helper. Net −27 lines in this file.

tests/json-api/tests/integration/cache/hash-equality-test.ts
```diff
@@ -0,0 +1,268 @@
...
+  unhashedProfile: { bio: string } | null;
+  settings: Record<string, unknown> | null;
+}
+
+interface CustomContext extends TestContext {
+  store: Store;
+}
+
+// only `id` and `state` participate, so a change to `note` alone is not a change
+function hashMessage(data: object): string {
```
gitKrystan: The hash deliberately omits `note`, so the "the hash decides" test below can prove that the hash, not a content walk, is what decides.

tests/json-api/tests/integration/cache/hash-equality-test.ts
```diff
@@ -0,0 +1,268 @@
...
+    const local = watch(store, user, 'local');
+
+    pushUser(store, { messages: [{ id: 'm1', state: 'executed' }] });
+    flush(store);
+
+    assert.true(remote.includes('messages'), `remote heard the changed element (saw ${remote.join()})`);
+    assert.true(local.includes('messages'), `local heard the changed element (saw ${local.join()})`);
+  });
+
+  test<CustomContext>('the hash decides: a change to a field the hash ignores does not notify', function (assert) {
```
gitKrystan: The load-bearing test for the design. A change to an un-hashed field is silent because the schema said so. A structural compare would notify here.

tests/json-api/tests/integration/cache/hash-equality-test.ts
```diff
@@ -0,0 +1,268 @@
...
+    pushUser(store, { attachments: [attachments[0], { kind: 'reaction', emoji: '👋', count: 3 }] });
+    flush(store);
+    assert.deepEqual(remote, ['attachments'], 'a changed hashed field on one element notifies');
+
+    pushUser(store, { attachments: [attachments[0], { kind: 'sticker', code: 'wave' }] });
+    flush(store);
+    assert.deepEqual(remote, ['attachments', 'attachments'], 'an element changing type notifies');
+  });
+
+  test<CustomContext>('a schema-object with identity: null compares by reference and notifies on an equal-content re-push', function (assert) {
```
gitKrystan: Control: passes on main too. Documents the `===` fallback for `identity: null`, and the `object` field test below it does the same for an unstructured kind.

tests/json-api/tests/integration/cache/hash-equality-test.ts
```diff
@@ -0,0 +1,268 @@
...
+    const remote = watch(store, user, 'remote');
+
+    pushUser(store, { settings: { a: 1, b: 2 } });
+    flush(store);
+
+    assert.deepEqual(remote, ['settings'], 'object fields have no schema to hash against');
+  });
+
+  // the scalar counterpart, which holds by reference equality alone, is in did-commit-notification-test.ts
+  test<CustomContext>('a nested local edit confirmed by an equal-hash push is dropped', async function (assert) {
```
gitKrystan: Asserts only the dirty state here. Which channels hear a confirming push is #11159's change (on main the confirmed key is pruned from the notification set), so this file stays out of that question.

tests/core/tests/polaris-mode/commit-reactivity-test.gts
```diff
@@ -61,6 +61,20 @@ class NestedRow extends Component<{ Args: { saved: User; edits: EditableUser } }
   </template>
 }
 
+let rowRenderCount = 0;
+
+function hashMessage(data: object): string {
```
gitKrystan: Without this hash the `message` schema is `identity: null`, elements compare by reference, and an equal-content push rebuilds every row exactly as on main. Declaring the hash is the app-side opt-in the design expects.

tests/core/tests/polaris-mode/commit-reactivity-test.gts
```diff
@@ -238,4 +255,47 @@ module('Reactivity | committing updates a rendered immutable record', function (
     assert.dom('[data-test-saved]').hasText('Christopher', 'saved renders the saved value');
     assert.dom('[data-test-edits]').hasText('Christopher', 'edits renders the saved value');
   });
+
+  test('a remote push with equal-hash messages does not rebuild the rendered rows', async function (this: RenderingTestContext, assert) {
```
gitKrystan: Counts component constructions rather than reading a property back. On main an equal-content push of `messages` rebuilds every row: `ManagedArray` keys child records by raw element object under the default `@identity` mode, and every element in a fresh payload is a new object.

warp-drive-packages/json-api/src/-private/cache.ts
```diff
@@ -1891,6 +1894,49 @@ function candidateKeys(
...
+/**
+ * Whether two values of `field` count as the same value for change detection.
+ *
+ * A `schema-object`, and each element of a `schema-array`, compares by the identity hash its
+ * `ObjectSchema` declares (`identity: { kind: '@hash', ... }`), so the schema decides what "same"
+ * means for it. With no hash declared, and for every other field kind, only the same reference
+ * counts. That knowingly over-notifies for equal-content objects: content equality is the schema's
+ * to define, not the cache's to guess.
+ */
+function attrValuesEqual(schema: SchemaService, field: FieldSchema, a: unknown, b: unknown): boolean {
```
gitKrystan, line 1906: The one definition of "same value" for change detection, used by both `calculateChangedKeys` and `patchLocalAttributes` so a confirmed edit is recognized the same way a moved projection is. Dispatches on field kind: schema-object and schema-array go through the identity hash, everything else stays `===`. That fallback knowingly over-notifies for equal-content objects; per the original TODO here, content equality is the schema's to define, not the cache's to guess. An earlier revision on #11159 used a generic structural walk instead; it was dropped for exactly that reason.

warp-drive-packages/json-api/src/-private/cache.ts
```diff
@@ -1891,6 +1894,49 @@ function candidateKeys(
...
+  if (field.kind === 'schema-object') return schemaObjectsEqual(schema, field, a, b);
+  if (field.kind === 'schema-array') return schemaArraysEqual(schema, field, a, b);
+  return false;
+}
+
+/**
+ * A schema-array has no hash of its own: two arrays are the same when they have the same length
+ * and every element compares equal as a schema-object of the element type.
+ */
+function schemaArraysEqual(schema: SchemaService, field: SchemaArrayField, a: unknown, b: unknown): boolean {
```
gitKrystan, line 1917: A schema-array has no hash of its own (`key: '@hash'` and `type: '@hash'` on the field both refer to the *element* type's hash), so array equality is length plus per-element `schemaObjectsEqual`. Split out from `attrValuesEqual` so the array and object rules sit side by side.

warp-drive-packages/json-api/src/-private/cache.ts
```diff
@@ -1891,6 +1894,49 @@ function candidateKeys(
...
+  if (!Array.isArray(a) || !Array.isArray(b) || a.length !== b.length) return false;
+  for (let i = 0; i < a.length; i++) {
+    if (!schemaObjectsEqual(schema, field, a[i], b[i])) return false;
+  }
+  return true;
+}
+
+function schemaObjectsEqual(
```
gitKrystan, line 1925: Type first, then hash. Two elements of a polymorphic array with different concrete types are unequal before any hash runs, and the hash is looked up on the concrete type's `ObjectSchema`. Both come from `schema.fieldValueIdentity(field, value)`, which returns `{ type, hash }` with `hash: null` for an `identity: null` type, so that case lands in the `===` fallback via `hash !== null`. A `SchemaService` that does not implement the optional method gets the same reference-equality fallback.

warp-drive-packages/json-api/src/-private/cache.ts
```diff
@@ -1900,6 +1946,7 @@ function candidateKeys(
  * result depend on whether the caller had already mutated it.
  */
 function calculateChangedKeys(
+  schema: SchemaService,
```
gitKrystan, line 1949: New first parameter. The cache already holds the `SchemaService` as `this._capabilities.schema`; passing it in keeps `calculateChangedKeys` a pure function of its inputs, which the existing doc block above it promises.

warp-drive-packages/json-api/src/-private/cache.ts
```diff
@@ -1942,8 +1989,8 @@ function calculateChangedKeys(
     const wasLocal = isEdited ? localEdit : isPromoted ? promoted[key] : wasRemote;
     const nowLocal = isEdited ? localEdit : nowRemote;
 
-    const remoteMoved = wasRemote !== nowRemote;
-    const localMoved = wasLocal !== nowLocal;
+    const remoteMoved = !attrValuesEqual(schema, field, wasRemote, nowRemote);
```
gitKrystan, line 1992: This is the line the original TODO sat above. `wasRemote !== nowRemote` became `!attrValuesEqual(...)`; the four before/after values and the bucket logic are unchanged.

warp-drive-packages/json-api/src/-private/cache.ts
```diff
@@ -2059,8 +2111,10 @@ function patchLocalAttributes(cached: CachedResource, changedRemoteKeys?: Set<st
         : remoteAttrs && attr in remoteAttrs
           ? remoteAttrs[attr]
           : undefined;
+    const field = fields.get(attr);
 
-    if (existing === localAttrs[attr]) {
+    // the same equality `calculateChangedKeys` uses, so a confirmed edit is recognized the same way
```
gitKrystan, line 2116: Without this, a nested local edit to a hashed schema-object could never be recognized as confirmed: the cloned local array is never `===` the pushed one, so the record stayed dirty forever after the server agreed. `fields.get(attr)` can be undefined for a key the schema does not know, hence the `===` branch.

warp-drive-packages/core/src/reactive/-private/fields/managed-array.ts
```diff
@@ -189,34 +189,11 @@ export class ManagedArray {
...
-                typePath && objectType && typeof objectType === 'string'
-              );
-            }
-          } else {
-            assert(`A non-polymorphic SchemaArrayField must provide a SchemaObject type in its definition`, field.type);
-            objectType = field.type;
-          }
+          // schema-object identity is only required to be unique by type, so a polymorphic array
+          // resolves the concrete type first
+          const identity = schema.fieldValueIdentity(field, rawValue);
```
gitKrystan, line 194: Behavior unchanged; the 22-line inline block this replaces became `SchemaService.fieldValueIdentity`, asserts included.

warp-drive-packages/core/src/reactive/-private/fields/managed-array.ts
```diff
@@ -227,9 +204,8 @@ export class ManagedArray {
            */
           let schemaObjectKeyValue: string | number | object;
           if (KeyMode === '@hash') {
-            const hashField = schema.resource({ type: objectType! }).identity as HashField;
-            const hashFn = schema.hashFn(hashField);
-            schemaObjectKeyValue = hashFn(rawValue as object, hashField.options ?? null, hashField.name);
+            assert(`Expected the '${objectType}' schema to declare an identity hash for key: '@hash'`, hash !== null);
```
gitKrystan, line 207: Previously this cast `identity as HashField` and would have thrown inside `schema.hashFn` for an `identity: null` element type. Now it asserts with a message naming the type.

warp-drive-packages/core/src/reactive/-private/kind/schema-object-field.ts
```diff
@@ -38,36 +38,14 @@ export function getSchemaObjectField(context: KindContext<SchemaObjectField>): u
...
-
-  const hashField = (schema.resource({ type: objectType }) as ObjectSchema).identity;
-  const identity = hashField
-    ? schema.hashFn(hashField)(rawValue, hashField.options ?? null, hashField.name)
-    : field.name;
+  assert(
+    `The SchemaService must implement fieldValueIdentity to support schema-object fields`,
+    typeof store.schema.fieldValueIdentity === 'function'
+  );
+  const resolved = store.schema.fieldValueIdentity(field, rawValue);
```
gitKrystan, line 45: Same fallback as before (`field.name` when no hash), via the service. Net −27 lines in this file.

tests/json-api/tests/integration/cache/hash-equality-test.ts
gitKrystan, line 44: The hash deliberately omits `note`, so the "the hash decides" test below can prove that the hash, not a content walk, is what decides.

tests/json-api/tests/integration/cache/hash-equality-test.ts
gitKrystan, line 174: The load-bearing test for the design. A change to an un-hashed field is silent because the schema said so. A structural compare would notify here.

tests/json-api/tests/integration/cache/hash-equality-test.ts
gitKrystan, line 207: Control: passes on main too. Documents the `===` fallback for `identity: null`, and the `object` field test below it does the same for an unstructured kind.

tests/json-api/tests/integration/cache/hash-equality-test.ts
gitKrystan, line 230: Asserts only the dirty state here. Which channels hear a confirming push is #11159's change (on main the confirmed key is pruned from the notification set), so this file stays out of that question.

tests/core/tests/polaris-mode/commit-reactivity-test.gts
gitKrystan, line 66: Without this hash the `message` schema is `identity: null`, elements compare by reference, and an equal-content push rebuilds every row exactly as on main. Declaring the hash is the app-side opt-in the design expects.

tests/core/tests/polaris-mode/commit-reactivity-test.gts
gitKrystan, line 259: Counts component constructions rather than reading a property back. On main an equal-content push of `messages` rebuilds every row: `ManagedArray` keys child records by raw element object under the default `@identity` mode, and every element in a fresh payload is a new object.

warp-drive-packages/core/src/types/schema/schema-service.ts
```diff
@@ -157,6 +169,27 @@ export interface SchemaService {
...
+   *
+   * Every other field kind, and any non-object value, has no identity and
+   * returns `null`.
+   *
+   * Optional: a SchemaService with no `@hash` support may omit it, and
+   * consumers then treat every value as having no identity.
+   *
+   * @public
+   */
+  fieldValueIdentity?(field: FieldSchema, value: unknown): SchemaObjectIdentity | null;
```
gitKrystan: New optional method on the public `SchemaService`. It takes any field and answers "what is the identity of this value as a value of that field": for a `schema-object`, or one element of a `schema-array`, the concrete type plus the hash the type's `ObjectSchema` declares (`hash: null` for `identity: null`); for every other kind, `null`. This is the polymorphic type resolution and hash lookup that `ManagedArray` and `getSchemaObjectField` each carried a copy of, and that the cache now needs too, so it moved onto the service that already owns `hashFn()` and `resource()` rather than into a shared util. Optional so third-party schema services keep compiling; the reactive layer asserts it exists (schema-object fields cannot work without it) and the cache treats absence as "no identity, compare by reference". Named for the field's *value* on purpose, so it cannot be misread as record identity.

warp-drive-packages/core/src/reactive/-private/schema.ts
```diff
@@ -811,6 +812,36 @@ export class SchemaService implements SchemaServiceInterface {
     );
     return this._hashFns.get(field.type)!;
   }
+  fieldValueIdentity(field: FieldSchema, value: unknown): SchemaObjectIdentity | null {
```
gitKrystan: The one implementation. The body is the former inline block from `ManagedArray` verbatim (both asserts included), followed by the hash lookup from `getSchemaObjectField`.

warp-drive-packages/legacy/src/model/migration-support.ts
```diff
@@ -633,6 +634,14 @@ export class DelegatingSchemaService implements SchemaService {
   hashFn(field: HashField | { type: string }): HashFn {
     return this._preferred.hashFn(field);
   }
+  fieldValueIdentity(field: FieldSchema, value: unknown): SchemaObjectIdentity | null {
```
gitKrystan: Fragments are schema-objects served through this delegating service, so it has to forward. It prefers the app's service and falls back to the model-derived one, and asserts if neither implements the method, mirroring how `hashFn` delegates above.

warp-drive-packages/core/src/reactive/-private/schema.ts
```diff
@@ -811,6 +812,38 @@ export class SchemaService implements SchemaServiceInterface {
...
+      assert(`A non-polymorphic ${field.kind} field must provide a SchemaObject type in its definition`, field.type);
+      type = field.type;
+    }
+
+    const hashField = (this.resource({ type }) as ObjectSchema).identity;
+    const hash =
+      hashField && hashField.kind === '@hash'
+        ? this.hashFn(hashField)(rawValue, hashField.options ?? null, hashField.name)
+        : null;
+    return { type, hash };
```
gitKrystan: Are there other types of identity we can/should handle here?

And should we always return `{ type }` even if there is no `hash`?

gitKrystan:
Superseded by #11181, re-opened as a new PR so the squash commit carries the right author. Same diff (66f85c9879) and the inline explanations are copied over.

PR #10099, opened 2025-08-01
docs: Missing export in the example

MichalBryxi:
Since few lines below say:

```
import AuthHandler from './auth-handler.js';
```

we should make sure it's exported.

### DIFF
--- a/guides/the-manual/cookbook/auth-handlers.md
+++ b/guides/the-manual/cookbook/auth-handlers.md
@@ -50,10 +50,10 @@ This handler would need to be added to the request manager configuration:
 ```ts
 import RequestManager from '@ember-data/request';
 import Fetch from '@ember-data/request/fetch';
-import { AuthHandler } from './auth-handler.ts';
+import { AuthHandler } from './auth-handler.ts'; // [!code focus]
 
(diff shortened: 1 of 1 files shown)


runspired:
lets update to a named export/import as well.

We can also take advantage of the `focus` feature to help narrow in what you should pay attention to while still making the full thing easy to copy/paste or see https://vitepress.dev/guide/markdown#focus-in-code-blocks

MichalBryxi:
1) Moved imports of auth-handler to `import { AuthHandler } from './auth-handler.ts';` and adjusted exports accordingly
2) Added also named export for `RequestManager`

Now:
- [ ] Should the handlers be defined in `app/handlers/auth-handler.ts` instead in `app/services/auth-handler.ts`? I believe that is/used to be the standard in the docs at some point? It's not a service anyways, right?
- [ ] Should also `app/services/request-manager.js` be moved to `app/services/request-manager.ts`? It's a detail, but again - consistency

MichalBryxi:
Ok, I've added `// [!code focus]` to the sections that:
1) I touched
2) Seems to be the tofu of the problem discussed in given section

PR #10108, opened 2025-08-06
Add `unknownProperty` and `size` to `IgnoredGlobalFields` list

### LINKED ISSUE #10096: Passing a SchemaRecord to ember `isEmpty` brings `Error: No field named unknownProperty on user`
While making tests with schema record / warp-drive (polaris mode) i was running into error `Error: No field named unknownProperty on user`.

This occors, if you pass a schema record to `isEmpty()`, imported from `@ember/utils`.

While i confirm, that using `isEmpty` make not really sense with warp-drive, its a possible issue for users which are moving an old app to warp-drive

Is there any was to fix this?

Here my used schemas:

```ts
const LoginUserSchema = withDefaults({
  type: 'login/user',
  fields: [
    {
      kind: 'field',
      name: 'guid',
      type: 'guid',
    },
    {
      kind: 'field',
      name: 'userName',
      type: 'string',
    },
    {
      kind: 'field',
      name: 'emailAddress',
      type: 'string',
    },
    {
      kind: 'field',
      name: 'language',
      type: 'number',
    },
    {
      kind: 'field',
      name: 'creationDate',
      type: 'date',
      options: { nullable: false },
    },
    {
      kind: 'field',
      name: 'modificationDate',
      type: 'date',
      options: { nullable: false },
    },
    {
      kind: 'belongsTo',
      name: 'privilegeRole',
      type: 'login/privilege-role',
      options: { async: false, inverse: 'users', linksMode: true },
    },
  ],
});

const LoginPrivilegeRoleSchema = withDefaults({
  type: 'login/privilege-role',
  fields: [
    {
      kind: 'field',
      name: 'name',
      type: 'string',
    },
    {
      kind: 'field',
      name: 'name',
      type: 'string',
    },
    {
      kind: 'field',
      name: 'privileges',
      type: 'array',
    },
    {
      kind: 'hasMany',
      name: 'users',
      type: 'login/user',
      options: { async: false, inverse: 'privilegeRole', linksMode: true },
    },
  ],
});
```

Ember: 6.6
EmberData / WarpDrive: 5.6

mkszepp:
fix #10096

While using `isEmpty` and `@ember/utils` with schema-record i was running into error `Error: No field named unknownProperty on user.`

To resolve this issue we need to add `unknownProperty` and `size` to `IgnoredGlobalFields`

### DIFF
--- a/warp-drive-packages/core/src/reactive/-private/record.ts
+++ b/warp-drive-packages/core/src/reactive/-private/record.ts
@@ -28,7 +28,7 @@ import { peekManagedObject } from './fields/managed-object.ts';
 import type { SchemaService } from './schema.ts';
 import { Checkout, Commit, Context, Destroy } from './symbols.ts';
 
-const IgnoredGlobalFields = new Set<string>(['length', 'nodeType', 'then', 'setInterval', 'document', STRUCTURED]);
+const IgnoredGlobalFields = new Set<string>(['length', 'nodeType', 'then', 'setInterval', 'document', 'unknownProperty', 'size', STRUCTURED]);
 const symbolList = [Context, Destroy, RecordStore, Checkout, Commit];
 const RecordSymbols = new Set(symbolList);
 


warp-drive-packages/core/src/reactive/-private/record.ts
```diff
@@ -28,7 +28,7 @@ import { peekManagedObject } from './fields/managed-object.ts';
 import type { SchemaService } from './schema.ts';
 import { Checkout, Commit, Context, Destroy } from './symbols.ts';
 
-const IgnoredGlobalFields = new Set<string>(['length', 'nodeType', 'then', 'setInterval', 'document', STRUCTURED]);
+const IgnoredGlobalFields = new Set<string>(['length', 'nodeType', 'then', 'setInterval', 'document', 'unknownProperty', 'size', STRUCTURED]);
```
runspired, line 31: size is a pretty odd one to need, what's the story there? 'length' is here because its what ember uses for isEmpty and JS uses to determine if something is ArrayLike, I'm not familiar with anything using size

mkszepp: after adding `unknownProperty` to ignore list, i was running into `size` error... for this reason i have also added... after both, app works fine

PR #10112, opened 2025-08-08
feat: @warp-drive/react <Request />

runspired:
Implements `<Request />` overtop of `<ReactiveContext />`


warp-drive-packages/react/src/-private/request.tsx
```diff
@@ -0,0 +1,234 @@
...
+ * ### Example Usage
+ *
+ * ```tsx
+ * import { Request } from "@warp-drive/react";
+ *
+ * export function UserPreview($props: { id: string | null }) {
+ *   return (
+ *    <Request
+ *       query={findRecord('user', $props.id)}
+ *       states={{}
```
Shaglock, line 137:
```suggestion
 *       states={{
```

PR #10364, opened 2025-09-03
Add React tests for await, promise and request state specs

MehulKChaudhari:
This PR adds React implementations for the `await-component`, `get-promise-state`and `get-request-state` rendering tests using the new spec-based architecture.

Files Added:
* `tests/framework-react/tests/await-component.spec-test.tsx`
* `tests/framework-react/tests/get-promise-state.spec-test.tsx`
* `tests/framework-react/tests/get-request-state-rendering.spec-test.tsx`


runspired:
Kinda impressive that without needing to change the specs we were able to get all but two tests passing so elegantly. Shows its pretty well done. Spot checking those two failures, we may be able to address within the react spec impl too, meaning no code changes to the spec or the library implementation.

MehulKChaudhari:
> Kinda impressive that without needing to change the specs we were able to get all but two tests passing so elegantly. Shows its pretty well done. Spot checking those two failures, we may be able to address within the react spec impl too, meaning no code changes to the spec or the library implementation.

Yup, looks like we can pull this off pretty smoothly across different frameworks. 

I tried around with two failures but couldn’t get them to pass feels like either the tests are off or we might need a small specs tweak.

What do you think?

PR #10372, opened 2025-09-06
Missing Context Param in Code Example for setTestId

lucasxzh:
This context parameter is necessary in QUnit as well, not just other test launchers.

---


runspired (approved):
ah thanks, we aren't using this with qunit in the main test suite which is the only place I think we copied this mistake (also the teardown needs to be `setTestId(this, null)`.

Out of curiosity - how'd you bump into this?

PR #10393, opened 2025-09-21
fix: improve docs and make peek mutation safe

runspired:


warp-drive-packages/utilities/src/-private/json-api/save-record.ts
```diff
@@ -25,6 +25,11 @@ function isExisting(identifier: ResourceKey): identifier is PersistedResourceKey
 }
 
 /**
+ * :::warning ⚠️ **These Mutation Builders DO NOT Set the Request Body**
```
MichalBryxi, line 28: `Set` -> `set`

PR #10400, opened 2025-09-25
add(doc): ResourceSchema

MehulKChaudhari:


guides/the-manual/schemas/resource-schemas.md
```diff
@@ -3,3 +3,85 @@ order: 1
 ---
 
 # Resource Schemas
+
+Resource Schemas are the foundation of ***Warp*Drive**’s data layer.  
+They describe the shape of your resources in a **plain object**, replacing the old `DS.Model` pattern with something universal, explicit, and future-ready.
```
runspired, line 8: I'd like to keep the main guide authored from the perspective that `Legacy` does not exist, or where "additional context for Legacy Ember" is kept in separate follow on pages, this way we prevent confusing new users - especially those coming from non-Ember backgrounds.

For instance, how under `Configuration` we have

- setup
- setup - Advanced
- setup - Legacy (Ember)

To that end, I feel like the references to Models are better left to a migration guide or docs page specifically explaining how Model concepts fit into Schema concepts.

Issue #10399, opened 2025-09-25
lint: default import from @ember-data/store should convert to named Store import from @warp-drive/core


runspired:
currently the eslint rule converts

```diff
-import type Store from '@ember-data/store';
+import type Store from '@warp-drive/core';
```

Issue #10396, opened 2025-09-22
lint: default RequestManager import from @ember-data/request is converted to @warp-drive/ember default import


runspired:
see title

runspired:
this was fixed by #10546

PR #10405, opened 2025-09-30
WIP: Fixes to legacy import lint

### LINKED ISSUE #10394: no-legacy-imports should account for more recent exports
We added (and thus exposed via *) some exports, mostly types. We should either (A) also account for new exports added over time in our book-keeping or (B) say the older packages have their exports frozen entirely and to gain access to new things you must shift. I suspect (A) is better, and what we should do is create a mapping of file-to-file for any unknown tokens so we aren't constantly updating the list.

We also don't error for legacy import paths if they import tokens not in the current list - we should.

Baltazore:
Should fix #10394 #10396 #10399


runspired:
closing due to staleness - I had claude figure out if any of this work was still relevant, will open a PR for the piece that is ... accounting for `export *`

PR #10412, opened 2025-10-04
Add(doc) - Derivations and Traits section

MehulKChaudhari:
This PR adds two new docs pages: Derivations and Traits section

### DIFF
--- a/guides/the-manual/schemas/derivations.md
+++ b/guides/the-manual/schemas/derivations.md
@@ -1,5 +1,5 @@
 ---
-order: 7
+order: 6
 ---
 
 # Derivations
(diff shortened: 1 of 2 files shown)


guides/the-manual/schemas/derivations.md
```diff
@@ -3,3 +3,89 @@ order: 7
...
+```ts [schemas/user.ts]
+import { withDefaults } from '@warp-drive/core/reactive';
+
+export const UserSchema = withDefaults({
+  type: 'user',
+  fields: [
+    { name: 'id', kind: '@id' },
+    { name: 'firstName', kind: 'field' },
+    { name: 'permissions', kind: 'field' },
+    { name: 'canEdit', kind: 'derived', type: 'boolean' }
```
runspired: `type` here would be the name used to register the derivation so likely `canEdit`, perhaps choosing a different derivation name from field name will make this more clear

PR #10413, opened 2025-10-04
Add(docs) - Transformations

MehulKChaudhari:


runspired:
I re-reviewed this after looking at the current transformation docs - we may want to add Map handling as a cookbook recipe in the future, but for now it's probably easier to leave this be.

PR #10414, opened 2025-10-05
docs: First mention of @warp-drive/build-config

MichalBryxi:
This is the first place in guides when `@warp-drive/build-config` is mentioned & used. The user had no chance to install the package yet. Not sure if this is the best place.


runspired:
Installation is not required (or desired) when using the new packages.

Issue #10425, opened 2025-11-06
splice(0) does not work properly for hasMany's


BoussonKarel:
Failing test created here in `core` tests: https://github.com/BoussonKarel/warp-drive/tree/splice-without-delete-count-fails

BoussonKarel:
I also had a different issue in the tests of my addon.
I was using splice(0) on a legacy hasMany with 2 items which cleared both from the array, but it did not clear the inverses correctly.
Maybe it's related to this bug, maybe it's something different?

PR #10426, opened 2025-11-06
Splice without delete count fails

BoussonKarel:
attempt at fixing #10425

### DIFF
--- a/warp-drive-packages/core/src/store/-private/record-arrays/legacy-many-array.ts
+++ b/warp-drive-packages/core/src/store/-private/record-arrays/legacy-many-array.ts
@@ -342,8 +342,11 @@ function _MUTATE<T>(
         }
       }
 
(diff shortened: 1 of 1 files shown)


warp-drive-packages/core/src/store/-private/record-arrays/legacy-many-array.ts
```diff
@@ -325,10 +325,25 @@ function _MUTATE<T>(
...
+          // do not remove any elements
+          deleteCount = 0;
+        } else {
+          deleteCount = _deleteCount;
+        }
+      }
 
       // detect a full replace
-      if (start === 0 && deleteCount === collection[Context].source.length) {
+      if (start === 0 && deleteCount >= collection[Context].source.length) {
```
runspired: if > length we should sanitize by adjusting it down to match length so that we can trust the number

runspired (approved):

PR #10448, opened 2026-01-14
Adds Graphql support

arthurrsouza:
This PR adds support to GraphQL in WarpDrive. It adds optional `graphql` and `graphql-tag` dependencies to `@warp-drive/utilities` and some files to help calling GraphQL APIs.

### GraphQL Query Builder:
Responsible for building the request data for a giving GraqphQL query + schema type.

### GraphQL Request Handler
Responsible for handling an GraphQL request and transforming into a valid JSON:API response, that will be used to create all records in WarpDrive cache.

TODO:

- [x] GraphQL query builder
- [x] Tests for Query Builder
- [x] Graphql Request Handler
- [x] Tests for Request Handler
- [x] Error Policy - used to return partial results when the API has some errors
- [ ] Pagination Support


warp-drive-packages/utilities/src/-private/graphql/request-handler.ts
```diff
@@ -0,0 +1,346 @@
...
+
+      return resources;
+    }
+
+    if (typeof data === 'object') {
+      if ('edges' in data && data.edges) {
+        const resources = (data.edges as ArrayValue)
+          .map((item: Value) => {
+            if (item && typeof item === 'object' && 'node' in item && item.node) {
+              return this.transformObjectToResource(item.node as Value, typeName, included, processedIds, meta);
```
johanrd, line 234: this discards all other info on the edge than the node, no?

johanrd (commented):
Thanks for exploring this. It would be very cool to see a GraphQL story in WarpDrive. Some comments from the perspective of an ember app on Apollo (plus some RequestManager) today:

> transforming into a valid JSON:API response

Wouldn't this be somewhat working against the grains of GraphQL? (E.g. `parseAsRelationship()` only calls `isHasMany()`, so a nested single entity lands in `attributes` unnormalized). 

Maybe model at the Cache / schema layer instead to match the schema strengths of GraphQL with the schema strengths of WarpDrive? (WarpDrive now has `@id` / `@hash` `schema-object` / polymorphic, as I understand it).

warp-drive-packages/utilities/src/-private/graphql/request-handler.ts
arthurrsouza: not all the other info. we retrieve the `pageInfo` [here](https://github.com/warp-drive-data/warp-drive/pull/10448/changes#diff-710d803aef3c0676a86592d969bee0dba69768798da61d1973d310d03edc3612R240) and add it to the meta

arthurrsouza:
@johanrd I agree we could start digging into this to make it a full GQL cache, not a transformation. our initial conversations was around `How can we start making GQL requests?`, and one of the suggestions from @runspired was to start by converting a GQL into JSON:API and using all the existing APIs from WarpDrive.
Since that conversation (a year ago almost), WarpDrive evolved a lot. Maybe it's time to go for the next step.

runspired:
@arthurrsouza I think @johanrd is just suggesting mapping the schema concepts more directly. `schema-object` and `schema-array` allow for embedded complex structures, meaning that not all embedded objects and arrays need to be hoisted up as relationships. I don't think we need a cache implementation here, normalizing to json:api will be fundamentally more correct and faster than normalizing in a graph format like Apollo does. It'll also keep us leaner - there's a reason Apollo is so bloated :P 


Where I'd push back - something WarpDrive can offer that Apollo can't is proper normalization of things that are relationships but not properly expressed as relationships in GQL queries. If the entity has a distinct cache-key in your DB it should in WarpDrive too.

warp-drive-packages/utilities/src/-private/graphql/request-handler.ts
johanrd: ah, I meant on the edge itself, like:

```graphql
type MemberProjectEdge {
  cursor: String!
  node: Member!
  periodList: [IDTimePeriod!]!
}
```

Here the edge is effectively the assignment resource (project ↔ member, with its own periods).

johanrd:
Sounds like a good approach to keep going down that path until we see real blockers. 

**Noting down where I expect it to be not-straightforward (based on our app):**
• **Fields with arguments under an entity**. Consider `project(id) { memberList(from, to) }`, with `from` and `to` from a date picker. Every range writes to the same relationship on the project. If the user goes back to a range that is still cached, the template shows the list for the last fetched range. For that query, the handler can put the list in the document data. But our project list also asks for `memberList(from, to)` under each project in a connection. No field kind takes arguments, so how does a template read the list for its own range?
• **Writes to lists after mutations.** In Apollo, one `cache.modify` can add a new item to all cached variants of a connection (and filter based on `storeFieldName`). With `cache.patch`, we need a way to know the cache key of each variant (maybe easy? dunno).
• **The handler needs the `DocumentNode` and the variables, not only the response**. Without them, it cannot see aliases or arguments.
• **POST requests need a cache key** (`cacheOptions.key`) made from the operation and the variables.
• **Mutations that do not return the record.** Some of ours (for instance deletes) return only a Boolean or an `{ id }`, so they do not map directly to create/update/deleteRecord.

PR #10455, opened 2026-01-20
Add extensions docs and start model fragments examples

RobbieTheWagner:

---


warp-drive-packages/core/src/types/schema/fields.ts
```diff
@@ -2360,7 +2386,7 @@ export interface ObjectSchema {
    * to objects to support migrating away from deprecated patterns
    * like custom getters, computeds, and methods
    */
-  objectExtensions?: string[];
+  objectExtensions?: (BuiltInObjectExtension | string)[];
```
runspired: we don't automatically make these available ... but hinting at them could be useful. I wonder if its enough to link them in the comment?

RobbieTheWagner: I was just trying to expose the possible ones people could use. We could make it generic I suppose and try to dynamically populate them somehow.

PR #10467, opened 2026-02-12
turn off fastboot tests

mansona:
Right now all the PRs that are opened against main are failing on the fastboot tests, and really "fastboot support" is not a concern of a V2 addon any more

I am proposing that we remove this always failing job for now, and if it's something we want to add back later we can 👍


runspired:
we've removed the fastboot check at this point in order to continue pace with ember a a project overall - but we should add an SSR/ReactNative env check at some point to ensure our use of globals is appropriately guarded.

PR #10474, opened 2026-02-17
codemods: model-to-schema typescript support

BobrImperator:
Pulls changes from https://github.com/warp-drive-data/warp-drive/pull/10487

- Remove remnants of previous Signature type generation
- Fix and normalize type imports
- Normalizes schema names for Traits and Extensions especially.
- Normalizes interface names to always include `Trait` or `Extension`
  - Names without suffixes are reserved for `Resource` `Brand`s 
- Adds a solution for constructing resource `Brand`s using Typescript interface merging.
- Fixes Mixin.create mixins being left over in generated traits code
- Fixes incorrect path files being used depending on `disableTypescriptSchemas` and `combineSchemasAndTypes` settings
This does rely on the fact that some imports are effectively circular, but Typescript appears to be fine with that.

```ts
// --
/*
 Generated Type Artifacts:

 For a Model:
 - just its fields and its traits fields (if any) (UserResource)
 - its fields, traits, extensions and WithLegacy (LegacyUserResource)


 For a Mixin: Updateable
 - just its fields and its traits fields (if any) (UpdateableTrait)
 - its fields, traits and extensions and WithLegacy (LegacyUpdateableTrait)
 
*/
type WithLegacy<T extends object> = T & { belongsTo(key: keyof T): void; hasMany(): void }

/**
 * Just the exact fields on this resource
 */
interface UserResource extends TimestamppedTrait {
  firstName: string;
  lastName: string;
  name: 'user fields'
  createdAt: number
}

// give people the nice import for just the "schema" bit
type User = WithLegacy<UserResource>;

// @ts-expect-error
interface UserExtension extends User, TimestamppedExtension {}
class UserExtension {
  get fullName(): string {
    return this.firstName + this.lastName;
  }
}

// give folks a nice import for the full thing
type UserWithExtensions = UserExtension;

// --

const a = {} as User;
a.createdAt;
a.belongsTo('firstName'); // expect works since in the resource schema
a.belongsTo('_firstName'); // expect error since not exists
a.belongsTo('deletedAt'); // expect works because inherited from trait of trait
a.belongsTo('fullName'); // expect error since from extension

interface DeleteableTrait {
    deletedAt: string;
}
interface DeleteableExtension extends DeleteableTrait {}
class DeleteableExtension {
    get timeSinceDeleted(): number {
     return 1;
    }
}

interface TimestamppedTrait extends DeleteableTrait {
  createdAt: string | number;
}

interface TimestamppedExtension extends TimestamppedTrait, DeleteableExtension {}
class TimestamppedExtension {
  get timeSince(): string | number {
    return this.createdAt;
  }

  name: string = 'user ext'
  
}
---

### DIFF
--- a/packages/codemods/src/schema-migration/tasks/migrate.ts
+++ b/packages/codemods/src/schema-migration/tasks/migrate.ts
@@ -186,7 +186,13 @@ function buildOutputFileName(
   }
 
   if (config.preserveExtension) {
-    const extension = sourceFilePath.endsWith('.ts') ? '.ts' : '.js';
+    const extension = suggestedFileName?.endsWith('.ts')
+      ? '.ts'
+      : suggestedFileName?.endsWith('.js')
+        ? '.js'
+        : sourceFilePath.endsWith('.ts')
+          ? '.ts'
+          : '.js';
     return relativePath.replace(/\.(js|ts)$/, `${config.suffix}${extension}`);
   }
 
@@ -227,7 +233,7 @@ function getArtifactOutputPath(
   // Build the output filename
   const outputName = config.useSuggestedFileName
     ? buildOutputFileName('', filePath, config, artifact.suggestedFileName) || 'unknown'
-    : buildOutputFileName(relativePath, filePath, config);
+    : buildOutputFileName(relativePath, filePath, config, artifact.suggestedFileName);
 
   const outputPath = join(resolve(outputDir), outputName);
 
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: packages/codemods/src/schema-migration/processors/mixin.ts [hunk 1]
```diff
@@ -229,7 +229,7 @@ function generateMixinArtifacts(
 
   collectTraitImports(extendedTraits, imports, options);
 
-  const traitSchemaName = traitInterfaceName;
+  const traitSchemaName = `${toPascalCase(baseName)}Schema`;
```
BobrImperator: A recurring problem so far is shotgun surgeries like this which are also very error prone that Claude was very happy with. Additionally the codemod often lacks the ability to quickly refer from a file to it's dependencies for the exact same reason.

There's on-going effort to wrangle this up https://github.com/BobrImperator/data/compare/codemod/typescript-support...BobrImperator:data:codemod/poc-typescript?expand=1

PR #10507, opened 2026-03-30
FIx SignalHooks import path in @warp-drive/ember

SergeAstapov:
Noticed broken import path when tried to use latest version in the app.

---

### DIFF
--- a/warp-drive-packages/ember/src/install.ts
+++ b/warp-drive-packages/ember/src/install.ts
@@ -6,8 +6,7 @@ import { importSync } from '@embroider/macros';
 
 import { DEPRECATE_COMPUTED_CHAINS } from '@warp-drive/core/build-config/deprecations';
 import { TESTING } from '@warp-drive/core/build-config/env';
-import { setupSignals } from '@warp-drive/core/configure';
-import type { SignalHooks } from '@warp-drive/core/store/-private';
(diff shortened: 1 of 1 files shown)


warp-drive-packages/ember/src/install.ts
```diff
@@ -6,8 +6,7 @@ import { importSync } from '@embroider/macros';
 
 import { DEPRECATE_COMPUTED_CHAINS } from '@warp-drive/core/build-config/deprecations';
 import { TESTING } from '@warp-drive/core/build-config/env';
-import { setupSignals } from '@warp-drive/core/configure';
-import type { SignalHooks } from '@warp-drive/core/store/-private';
+import { setupSignals, type SignalHooks } from '@warp-drive/core/configure';
```
SergeAstapov, line 9: [`@warp-drive/core/store/-private`](https://github.com/warp-drive-data/warp-drive/blob/main/warp-drive-packages/core/src/store/-private.ts) does not export `SignalHooks`

SergeAstapov:
can this be released as v3.8.2 patch? which branch should I point this change then?

runspired:
@SergeAstapov I assume you mean 5.8 not 4.8, it could be but that's close enough to latest that probably just use the canary? not much delta

SergeAstapov:
My bad indeed meant 5.8.2 @runspired!

Using canary would be totally working option, thank you for looking into this!

PR #10509, opened 2026-04-09
remove import Ember from 'ember'

mansona:
I'm not even sure when this file is being loaded but if we keep the `import Ember from "ember"` call it will break **eagerly** in Ember 7.0 because Vite freaks out.

Unfortunately, this isn't a complete 1-1 migration because:

- Ember.cacheFor was removed and will not be replaced https://deprecations.emberjs.com/id/deprecate-import-cache-for-from-ember
- I could not find any evidence of `decrementProperty`, `incrementProperty`, or `toggleProperty` exports from the global Ember namespace 🤔 or at least they weren't listed in the deprecation and the API docs seem to suggest they are member functions on EmberObjects not individiual functions 🤔 

I would appreciate some feedback on how to proceed here since this is blocking the Ember 7.0 release

### DIFF
--- a/warp-drive-packages/legacy/src/compat/extensions.ts
+++ b/warp-drive-packages/legacy/src/compat/extensions.ts
@@ -1,6 +1,13 @@
-import { type default as EmberObject, get, set } from '@ember/object';
+import {
+  type default as EmberObject,
+  get,
+  getProperties,
+  notifyPropertyChange,
+  set,
+  setProperties,
+} from '@ember/object';
+import { addObserver, removeObserver } from '@ember/object/observers';
 import { compare } from '@ember/utils';
-import Ember from 'ember';
 
 import { assert } from '@warp-drive/core/build-config/macros';
 import type { CAUTION_MEGA_DANGER_ZONE_Extension } from '@warp-drive/core/reactive';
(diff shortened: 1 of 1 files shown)


NullVoxPopuli:
do you have a stack trace for what caused the import from 'ember'?

mansona:
Here is the full stack from a `pnpm start` when ember-soruce is set to 7

```
  this error and leave the unresolved path in the bundle.

/Users/mansona/temp/ember-7-alpha/node_modules/.pnpm/esbuild@0.27.7/node_modules/esbuild/lib/main.js:1748
  let error = new Error(text);
              ^

Error: Error during dependency optimization:

✘ [ERROR] Could not resolve "ember.json" [plugin embroider-esbuild-resolver]

    node_modules/.pnpm/@warp-drive+legacy@5.8.1_@babel+core@7.29.0_@warp-drive+core@5.8.1_@babel+core@7.29.0___8d06b82c56d19c8e5d04916c261d8db9/node_modules/@warp-drive/legacy/dist/compat/extensions.js:3:18:
      3 │ import Ember from 'ember';
        ╵                   ~~~~~~~

  You can mark the path "ember.json" as external to exclude it from the bundle, which will remove this error and leave the unresolved path in the bundle.


    at failureErrorWithLog (/Users/mansona/temp/ember-7-alpha/node_modules/.pnpm/esbuild@0.27.7/node_modules/esbuild/lib/main.js:1748:15)
    at /Users/mansona/temp/ember-7-alpha/node_modules/.pnpm/esbuild@0.27.7/node_modules/esbuild/lib/main.js:1207:25
    at /Users/mansona/temp/ember-7-alpha/node_modules/.pnpm/esbuild@0.27.7/node_modules/esbuild/lib/main.js:1626:9
    at process.processTicksAndRejections (node:internal/process/task_queues:103:5) {
  errors: [Getter/Setter],
  warnings: [Getter/Setter]
}

Node.js v24.13.1
```

warp-drive-packages/legacy/src/compat/extensions.ts
```diff
@@ -1,6 +1,13 @@
...
+  type default as EmberObject,
+  get,
+  getProperties,
+  notifyPropertyChange,
+  set,
+  setProperties,
+} from '@ember/object';
+import { addObserver, removeObserver } from '@ember/object/observers';
 import { compare } from '@ember/utils';
-import Ember from 'ember';
```
NullVoxPopuli, line 3: how dare this exist!!!!

NullVoxPopuli:
Finished here: https://github.com/warp-drive-data/warp-drive/pull/10511

runspired:
closing since handled by another PR

PR #10511, opened 2026-04-09
remove all `import Ember from 'ember'`

NullVoxPopuli-ai-agent:
## Summary

Completes the work started in #10509 to remove all `import Ember from 'ember'` calls, which break eagerly under Ember 7.0 + Vite.

### Changes

- **`warp-drive-packages/legacy/src/compat/extensions.ts`**: Replace the `Ember[method](this, ...args)` dynamic dispatch with a `switch`/`case` using specific imports from `@ember/object` and `@ember/object/observers`. `decrementProperty`, `incrementProperty`, and `toggleProperty` are implemented inline using `get`/`set` since they are EmberObject instance methods without standalone function equivalents. `cacheFor` throws as it has been [removed with no replacement](https://deprecations.emberjs.com/id/deprecate-import-cache-for-from-ember).
- **`packages/-ember-data/src/-private/core.ts`**: Remove `Ember.libraries.registerCoreLibrary()` entirely — `Ember.libraries` is deprecated with no replacement per [RFC 1003](https://rfcs.emberjs.com/id/1003-deprecation-import-ember-from-ember/).
- **`vite.config.mjs` (both packages)**: Remove `'ember'` from externals lists, add `'@ember/object/observers'`.

### CI verification (local)

- Build: 33/33 tasks pass
- Lint: 78/78 tasks pass
- Type check: 47/47 tasks pass

## Test plan

- [ ] CI passes on this PR
- [ ] Verify extensions still work correctly in Ember 6.x (backward compatible — all replaced APIs exist in Ember 6.x)
- [ ] Verify no eager failures in Ember 7.0 environment

Supersedes #10509

### DIFF
--- a/warp-drive-packages/legacy/src/compat/extensions.ts
+++ b/warp-drive-packages/legacy/src/compat/extensions.ts
@@ -36,7 +36,7 @@ EmberObjectMethods.forEach((method) => {
       case 'decrementProperty': {
         const key = args[0] as string;
         const decrement = (args[1] as number | undefined) ?? 1;
-        return set(this, key, (get(this, key) as number) - decrement);
+        return set(this, key, ((get(this, key) as number) || 0) - decrement);
       }
       case 'get':
         return (get as (...args: unknown[]) => unknown)(this, ...args);
@@ -45,7 +45,7 @@ EmberObjectMethods.forEach((method) => {
       case 'incrementProperty': {
         const key = args[0] as string;
         const increment = (args[1] as number | undefined) ?? 1;
-        return set(this, key, (get(this, key) as number) + increment);
+        return set(this, key, (parseFloat(get(this, key) as string) || 0) + increment);
       }
       case 'notifyPropertyChange':
         return (notifyPropertyChange as (...args: unknown[]) => unknown)(this, ...args);
--- a/warp-drive-packages/legacy/src/compat/extensions.ts
+++ b/warp-drive-packages/legacy/src/compat/extensions.ts
@@ -34,18 +34,26 @@ EmberObjectMethods.forEach((method) => {
       case 'cacheFor':
         throw new Error('cacheFor has been removed and will not be replaced');
       case 'decrementProperty': {
-        const key = args[0] as string;
-        const decrement = (args[1] as number | undefined) ?? 1;
-        return set(this, key, (parseFloat(get(this, key) as string) || 0) - decrement);
+        const keyName = args[0] as string;
+        const decrement = (args[1] as number) ?? 1;
+        assert(
+          'Must pass a numeric value to decrementProperty',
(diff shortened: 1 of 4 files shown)


NullVoxPopuli:
Here is the CI result on the fork:
- https://github.com/NullVoxPopuli-ai-agent/warp-drive/pull/1

warp-drive-packages/legacy/src/compat/extensions.ts
```diff
@@ -21,7 +28,38 @@ const EmberObjectMethods = [
...
+        const decrement = (args[1] as number | undefined) ?? 1;
+        return set(this, key, (get(this, key) as number) - decrement);
+      }
+      case 'get':
+        return (get as (...args: unknown[]) => unknown)(this, ...args);
+      case 'getProperties':
+        return (getProperties as (...args: unknown[]) => unknown)(this, ...args);
+      case 'incrementProperty': {
+        const key = args[0] as string;
+        const increment = (args[1] as number | undefined) ?? 1;
```
runspired: does this match the original upstream? feels like it would be buggy for properties that don't start as numbers

NullVoxPopuli: Image

this is what is upstream https://github.com/emberjs/ember.js/blob/1886f8de49cabe805b76e963db9c95a7a66d2aeb/packages/%40ember/object/observable.ts#L516

NullVoxPopuli-ai-agent: Good catch — updated to match the upstream Ember implementations. Both now use `|| 0` to handle undefined/non-numeric properties, and `incrementProperty` uses `parseFloat()` matching [the original](https://github.com/emberjs/ember.js/blob/1886f8de49cabe805b76e963db9c95a7a66d2aeb/packages/%40ember/object/observable.ts#L516).

NullVoxPopuli: @NullVoxPopuli-ai-agent originally it also has parseFloat, which implies that we support string increment-properties, and coercing them back to numbers 🙈 this data carelessness is gross, but it's what the original supported.

NullVoxPopuli-ai-agent: Added `parseFloat` to `decrementProperty` as well for consistency. Both now coerce string values to numbers, matching the upstream's implicit support for string increment-properties.

NullVoxPopuli-ai-agent: Updated — both `incrementProperty` and `decrementProperty` now copy the upstream implementations verbatim, including the `assert()` validation and `parseFloat` coercion.

NullVoxPopuli: it matches now

runspired (approved):

NullVoxPopuli:
is this meant to be tagged canary? the branch target is/was `release`

PR #10512, opened 2026-04-09
add ember-cli-deprecation-workflow to all Ember test apps

NullVoxPopuli-ai-agent:
## Summary

- Adds `ember-cli-deprecation-workflow` v4 to all 11 Ember test applications
- Creates `app/deprecation-workflow.{ts,js}` config files with empty workflow lists
- Imports the deprecation workflow in each app's entry point (`app/app.{ts,js,gts}`)

This enables deprecation management across the test suite, which is useful for the Ember 7.0 upgrade — deprecations can be silenced/tracked per-app rather than flooding test output.

### Apps updated

- `dont-write-new-tests-here`
- `ember-data__adapter`
- `ember-data__graph`
- `ember-data__model`
- `ember-data__request`
- `example-app-ember`
- `fastboot`
- `json-api`
- `performance`
- `utilities`
- `vite-basic-compat`

### CI verification (local)

- Prettier: pass
- Lint: 78/78 tasks pass
- Type check: 47/47 tasks pass

## Test plan

- [ ] CI passes
- [ ] Verify `deprecationWorkflow.flushDeprecations()` works in browser console during `pnpm test --serve`

### DIFF
--- a/tests/dont-write-new-tests-here/app/deprecation-workflow.ts
+++ b/tests/dont-write-new-tests-here/app/deprecation-workflow.ts
@@ -1,0 +1,6 @@
+import setupDeprecationWorkflow from 'ember-cli-deprecation-workflow';
+
+setupDeprecationWorkflow({
+  throwOnUnhandled: false,
+  workflow: [],
+});
--- a/tests/dont-write-new-tests-here/app/deprecation-workflow.ts
+++ b/tests/dont-write-new-tests-here/app/deprecation-workflow.ts
@@ -1,6 +1,6 @@
 import setupDeprecationWorkflow from 'ember-cli-deprecation-workflow';
 
 setupDeprecationWorkflow({
-  throwOnUnhandled: false,
+  throwOnUnhandled: true,
   workflow: [],
 });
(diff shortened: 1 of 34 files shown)


tests/dont-write-new-tests-here/app/deprecation-workflow.ts
```diff
@@ -0,0 +1,6 @@
+import setupDeprecationWorkflow from 'ember-cli-deprecation-workflow';
+
+setupDeprecationWorkflow({
+  throwOnUnhandled: false,
```
NullVoxPopuli: let's set throw on unhandled to true (everywhere).

as library maintainers, we never want to ship code with deprecations

NullVoxPopuli-ai-agent: Done — set `throwOnUnhandled: true` in all 11 apps.

runspired:
we never want to ignore deprecations and we already error for any deprecation encountered, so this feels like an unnecessary and redundant thing to setup.

PR #10534, opened 2026-06-29
Update pluralization rules for 'bus' and 'tus'

gossi:
I did this entirely on github UI.

My cases, I have the word `apparatus` (in gymnastics) which pluralizes to `apparatuses`, but wasn't captured in the rules.

PS. I tried to use `irregular('apparatus', 'apparatuses');` - which alone does the job. But when `query()` is called, it reads from a different `IRREGULAR` module. I had two versions installed in `node_modules`, after cleanup still the same problem. That's another issue.


runspired (approved):

anandfresh (approved):

runspired:
Note I generally don't support expanding the default ruleset - but this seems like generally a good thing to have in the base rules. `irregular` is the correct thing for apps to be using here - if multiple copies are present you likely have a build bug to figure out :D

PR #10543, opened 2026-08-01
feat/fix: new approach to mirage detection

### LINKED ISSUE #10535: Incorrectly hitting IS_MAYBE_MIRAGE
I'm attempting to move to Request Handlers but when I attempt to modify headers I'm hitting:

```
Failed to execute 'set' on 'Headers': Headers are immutable
```

it is coming from: 

https://github.com/warp-drive-data/warp-drive/blob/main/warp-drive-packages/core/src/request/-private/fetch.ts#L143

`window.fetch` in my project is indeed patched, by Sentry, not mirage. There are so many things that patch fetch -- this really cannot be a determiner of mirage.

runspired:
resolves #10535

### DIFF
--- a/warp-drive-packages/core/src/index.ts
+++ b/warp-drive-packages/core/src/index.ts
@@ -16,7 +16,7 @@ import { CacheHandler, type CachePolicy, Store } from './store/-private.ts';
 import { recordIdentifierFor } from './store/-private.ts';
 import type { CacheCapabilitiesManager, ResourceKey } from './types.ts';
 import type { Cache } from './types/cache.ts';
-import { getRuntimeConfig, setLogging } from './types/runtime.ts';
+import { getRuntimeConfig, setIsMaybeMirage, setLogging } from './types/runtime.ts';
 import type { Derivation, HashFn, Transformation } from './types/schema/concepts.ts';
 import type { ObjectSchema, PolarisResourceSchema, Trait } from './types/schema/fields.ts';
 
@@ -30,6 +30,9 @@ if (TESTING) {
 
   // @ts-expect-error adding to globalThis
   globalThis.getWarpDriveRuntimeConfig = getRuntimeConfig;
+
+  // @ts-expect-error adding to globalThis
+  globalThis.setWarpDriveIsMaybeMirage = setIsMaybeMirage;
 }
 export { Store, CacheHandler, type CachePolicy };
 
--- a/warp-drive-packages/core/src/index.ts
+++ b/warp-drive-packages/core/src/index.ts
@@ -30,10 +30,14 @@ if (TESTING) {
 
   // @ts-expect-error adding to globalThis
   globalThis.getWarpDriveRuntimeConfig = getRuntimeConfig;
-
-  // @ts-expect-error adding to globalThis
-  globalThis.setWarpDriveIsMaybeMirage = setIsMaybeMirage;
 }
+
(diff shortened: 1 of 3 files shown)


warp-drive-packages/core/src/index.ts
runspired, line 1: since mirage is used outside of testing - likely need a non-env dependent config for this

warp-drive-packages/core/src/request/-private/fetch.ts
```diff
@@ -139,9 +148,12 @@ const Fetch = {
     const isMutationOp = Boolean(op && MUTATION_OPS.has(op));
 
     if (!isError && !isMutationOp && response.status !== 204 && !response.headers.has('date')) {
-      if (IS_MAYBE_MIRAGE()) {
```
runspired, line 142: I'm not convinced this is a good idea - too many things override fetch with a non-compliant implementation. a compliant implementation will always be immutable.

warp-drive-packages/core/src/request/-private/fetch.ts
```diff
@@ -33,13 +34,21 @@ function cloneResponse(response: Response, overrides: Partial<Response>) {
...
+// patched" as "this is Mirage" produces false positives for those apps.
+// See https://github.com/warp-drive-data/warp-drive/issues/10535
 let IS_MAYBE_MIRAGE = () => false;
 if (DEBUG) {
   IS_MAYBE_MIRAGE = () =>
     Boolean(
-      typeof window !== 'undefined' &&
-      ((window as { server?: { pretender: unknown } }).server?.pretender ||
-        window.fetch.toString().replace(/\s+/g, '') !== 'function fetch() { [native code] }'.replace(/\s+/g, ''))
+      getRuntimeConfig().mirage ||
```
runspired: this would be a breaking change for anyone actually using mirage today - we'll need to figure out a way to default to still making this check or some such

warp-drive-packages/core/src/index.ts
runspired: Fixed - moved this out of the `TESTING` gate. It's now attached unconditionally, since Mirage can be running in a plain dev build (`ember serve`) that isn't a `TESTING` build at all. See d09245e.

warp-drive-packages/core/src/request/-private/fetch.ts
runspired: Agreed, and reverted - the try/catch is no longer the primary signal. `IS_MAYBE_MIRAGE()` is back as the gate (compliant `fetch()` responses are always immutable per spec, so a real request would throw+catch on every single request under the old approach - wasteful for the common case). The try/catch now only wraps the already-detected-as-mirage branch, as a safety net in case that detection is itself a false positive. See d09245e.

warp-drive-packages/core/src/request/-private/fetch.ts
runspired: Restored the legacy `window.fetch.toString()` check as part of the default detection, so behavior for existing Mirage users is unchanged (no breaking change). The override now takes precedence over the heuristic in both directions (`true` or `false`, checked via `!== undefined`) rather than only OR-ing in a forced-true, so apps that get false-positived by the legacy check (like the reporter of #10535) can explicitly opt out with `setWarpDriveIsMaybeMirage(false)` if the try/catch safety net alone isn't enough (e.g. for the streaming-vs-text branch, which has no such safety net). See d09245e.

Issue #10498, opened 2026-03-17
ModelSchemaProvider._loadModelSchema causes infinite recursion via ShimModelClass


paddor:
## Bug

`ModelSchemaProvider._loadModelSchema(type)` in `@warp-drive/legacy@5.8.1` causes `RangeError: Maximum call stack size exceeded`.

## Stack trace

```
handlers.js:11 Uncaught (in promise) RangeError: Maximum call stack size exceeded
    at invoke (handlers.js:11:27)
    at deprecate (deprecate.js:187:5)
    at Store.modelFor (index-CQP2NSqg.js:5823:26)
    at ModelSchemaProvider._loadModelSchema (schema-provider-DJCV_6AF.js:2127:35)
    at ModelSchemaProvider.fields (schema-provider-DJCV_6AF.js:2160:12)
    at DelegatingSchemaService.fields (migration-support.js:470:28)
    at get attributes (index-CQP2NSqg.js:5546:40)
    at ModelSchemaProvider._loadModelSchema (schema-provider-DJCV_6AF.js:2128:37)
    at ModelSchemaProvider.fields (schema-provider-DJCV_6AF.js:2160:12)
```

## Root cause

In `@warp-drive/legacy` `schema-provider`, `_loadModelSchema(type)` calls `this.store.modelFor(type)`. In `@warp-drive/core` 5.8.1, `Store.prototype.modelFor` returns a `ShimModelClass` (not the actual Model class). The `ShimModelClass.get attributes()` getter calls `this.__store.schema.fields({type})`, which goes through `DelegatingSchemaService.fields()` → `ModelSchemaProvider.fields()`. Since the schema isn't cached yet (we're still inside `_loadModelSchema`), `fields()` calls `_loadModelSchema(type)` again — infinite recursion.

The cycle:

1. `_loadModelSchema(type)` — line 2150: `const modelClass = this.store.modelFor(type)` → returns `ShimModelClass`
2. Line 2151: `modelClass.attributes` → triggers `ShimModelClass.get attributes()`
3. `get attributes()` calls `store.schema.fields({type})` (core index line 5567)
4. → `DelegatingSchemaService.fields()` → `ModelSchemaProvider.fields()` → checks cache → **miss** → calls `_loadModelSchema(type)` → goto 1

## Fix

[...]

Issue #10493, opened 2026-03-10
`@warp-drive/legacy`: Model schema provider fields method with non-existing type causes infinite loop


BoussonKarel:
I've stumbled upon an infinite loop in the fields() method on the legacy Model schema provider in the legacy package. This only happens in **production**.

When calling fields with a nonexistent type, it calls _loadModelSchema.
https://github.com/warp-drive-data/warp-drive/blob/b3372e5274ace464093ae693c3aa22c0b52a50a8/warp-drive-packages/legacy/src/model/-private/schema-provider.ts#L141-L148

This then calls `this.store.modelFor`. Because the type does not exist, it calls `super.modelFor`, which returns a shim:
https://github.com/warp-drive-data/warp-drive/blob/b3372e5274ace464093ae693c3aa22c0b52a50a8/warp-drive-packages/legacy/src/store.ts#L206-L213

And this shim calls `this.__store.schema.fields`, causing a loop:
https://github.com/warp-drive-data/warp-drive/blob/b3372e5274ace464093ae693c3aa22c0b52a50a8/warp-drive-packages/core/src/store/deprecated/-private.ts#L410-L421

BoussonKarel:
In development there is an error, thrown because of assertType in the modelFor on the store.

<img width="1583" height="213" alt="Image" src="https://github.com/user-attachments/assets/f01a5383-f29b-42b9-a4c3-3ead8cf5995c" />

runspired:
looking into a fix for this and the slight variant in #10498

PR #10544, opened 2026-08-01
fix: prevent infinite recursion during model lookup

runspired:

### DIFF
--- a/tests/legacy/tests/unit/schema-provider-test.ts
+++ b/tests/legacy/tests/unit/schema-provider-test.ts
@@ -1,0 +1,43 @@
+import { Type } from '@warp-drive/core/types/symbols';
+import { module, setupTest, test } from '@warp-drive/diagnostic/ember';
+import Model, { attr } from '@warp-drive/legacy/model';
+
+import Store from '../serializer/store';
+
+class Person extends Model {
+  @attr
+  firstName;
+
+  @attr
+  lastName;
+
+  declare [Type]: 'person';
+}
+
+module('Unit | Model | ModelSchemaProvider', function (hooks) {
+  setupTest(hooks);
+
+  hooks.beforeEach(function () {
+    this.owner.register('service:store', Store);
+    this.owner.register('model:person', Person);
+  });
+
+  test('fields() does not recurse infinitely for a type with a registered model', function (assert) {
+    const store = this.owner.lookup('service:store') as Store;
+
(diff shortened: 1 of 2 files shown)


runspired:
## Root cause

Both #10493 and #10498 (a duplicate) trace to the same bug: `ModelSchemaProvider._loadModelSchema()` in [schema-provider.ts:106](https://github.com/warp-drive-data/warp-drive/blob/main/warp-drive-packages/legacy/src/model/-private/schema-provider.ts#L106) called `this.store.modelFor(type)`. In `@warp-drive/legacy`, `store.modelFor` (defined in [store.ts:206](https://github.com/warp-drive-data/warp-drive/blob/main/warp-drive-packages/legacy/src/store.ts#L206)) doesn't return the real Model class — it returns a `ShimModelClass` ([store/-private.ts:417](https://github.com/warp-drive-data/warp-drive/blob/main/warp-drive-packages/legacy/src/store/-private.ts#L417)) whose `.attributes` getter calls back into `store.schema.fields()`.

That creates a cycle: `_loadModelSchema` → `modelFor` → shim → `.attributes` → `schema.fields()` → cache miss (the schema isn't cached yet, since we're still inside the call that would cache it) → `_loadModelSchema` again → ∞.

- For a type with a real registered model (#10498), this cost one extra wasted recursive round-trip before eventually working, but could also blow the stack depending on the schema-service wrapper in play (e.g. `DelegatingSchemaService`).
- For a nonexistent type (#10493), `store.modelFor` guards against this with an `assert(...)`, but `assert` is a macro that's stripped entirely in production builds — so in prod it just returns the shim, and the recursion never terminates, producing a real infinite loop / stack overflow. This matches the "only happens in production" report.

## Fix

Changed `_loadModelSchema` to use `getModelFactory(this.store, type)` — the same helper already used by `hasResource()` in this file — which resolves the actual Ember-registered Model class directly from the container instead of going through `store.modelFor`. If no factory is found, it now throws a clear, production-safe `Error` (not gated by the `assert` macro) instead of falling into the shim/recursion path.

Also removed the now-unused `Model` type import.

Added a regression test in `tests/legacy/tests/unit/schema-provider-test.ts` covering both cases (existing model resolves fields correctly; nonexistent type throws a clear error instead of recursing). Full legacy test suite (107 tests) passes, and `ember-tsc --build` shows no new type errors introduced by this change.

Issue #10494, opened 2026-03-10
Repeated remote updates can leave accessed linksMode hasMany relationships stale


alexraputa:
## Summary

When a `linksMode` `hasMany` relationship has already been accessed/materialized, a later remote update can change the canonical relationship state without resynchronizing the materialized local state when there is no local mutation to reconcile.

## Reproduction

1. Seed two `user` records and two `pet` records with `user.pets <-> pet.owner`.
2. Start from:
   - `user:1.pets = ["1"]`
   - `user:2.pets = ["2"]`
3. Access `user1.pets` so the `linksMode` `hasMany` is materialized.
4. Send `PATCH /api/user/1` updating `pets` to `[{ type: "pet", id: "1" }, { type: "pet", id: "2" }]`.
5. Send another `PATCH /api/user/1` updating `pets` to `[{ type: "pet", id: "1" }]`.

## Expected

After each remote update, if there are no local additions, removals, or local reordering to preserve, the materialized `linksMode` `hasMany` should match the canonical remote state.

## Actual

The canonical/inverse state updates, but the already-accessed materialized `hasMany` can remain stale after repeated remote updates.

For example, after the second `PATCH`, `pet2.owner` is cleared but `user1.pets` can still contain `["1", "2"]` instead of `["1"]`.

## Suspected Area

This looks like a policy bug in `replaceRelatedRecordsRemote`: https://github.com/warp-drive-data/warp-drive/blob/b3372e5274ace464093ae693c3aa22c0b52a50a8/warp-drive-packages/core/src/graph/-private/operations/replace-related-records.ts#L178

The relevant case is:
- `hasMany`
- `linksMode`
- relationship already accessed/materialized
- remote membership changed
- no local mutation is present

___

**Environment**

- WarpDrive: `5.8.1`
- Node: `25.8.0`
- pnpm: `10.31.0`

alexraputa:
Hey @runspired - thanks for digging into this. I took a closer look at the TODO https://github.com/warp-drive-data/warp-drive/blob/5222bb03eecce247713de4c3c4da7407aa99b147/tests/json-api/tests/integration/cache/mutation-request-test.ts#L137-L154 and agree with the diagnosis: after the first remote update, the materialized `linksMode` array reads via `getRemoteRelationship()`, so `localState` is never refreshed. The next update can then calculate `diff.changed` against that stale snapshot.

[...]

PR #10560, opened 2026-08-06
feat/fix: dual channel notifications for cache updates

### LINKED ISSUE #9688: feat: NotificationManager should accept a flag for whether a change to local vs remote state results in the notification
now that SchemaRecord only shows remote state by-default, updates generated by changes to local-state are unnecessary.

additionally: updates resulting from application of local state become necessary where before they were unnecessary (this is essentially a bug in the SchemaRecord world).

having separate channels will enable better performance tuning

runspired:
resolves #9688
partially addresses #10494 (does not close it — see note below)

## Summary

Adds a `'local' | 'remote'` notification channel to `NotificationManager`/`CacheCapabilitiesManager`, so a change to a resource's `attributes`/`relationships` can be routed separately to its mutable/editable (local) view versus its remote-state-only (Polaris default) view — and fixes the notification-starvation bugs (#9688, and the `isDirty`-latch half of #10494) that came from using local state as a proxy for "should we notify."

> **Note on #10494:** this PR fixes the notification-layer half — the first remote update to a materialized `linksMode` hasMany now notifies correctly, since the `flushCanonical` gate no longer relies on the `isDirty` latch (which stuck `true` forever for readers that never call `computeLocalState`). The *repeated*-update staleness in #10494's repro remains: a `linksMode` hasMany caches its rendered array on first read (`getHasManyField`), so `graph.getData`/`computeLocalState` — the only refresh point for a `CollectionEdge`'s `isDirty`/`localState` — never runs again for that field, and a later remote update diffs against the stale `localState`. That is a read-path bug (pre-existing on `main`), not a notification bug, and is tracked by the `todo()` test `update hasMany with repeated patch` in `tests/json-api/tests/integration/cache/mutation-request-test.ts`, which asserts the currently-failing behavior and will flip when the read path is fixed.

## New public API surface

- `NotificationChannel = 'local' | 'remote'`, exported from `@warp-drive/core` alongside `NotificationType`/`NotifyKeys`.
- An optional trailing `channel?: NotificationChannel` on three existing methods: `NotificationManager.subscribe` (ResourceKey overload only), `NotificationManager.notify`, and `CacheCapabilitiesManager.notifyChange` (the latter two only for the `'attributes'`/`'relationships'` namespaces — `notify` asserts if a channel is passed for any other namespace).

Delivery matrix, documented on all three APIs — the only skipped combination is an explicit `'local'` notify to an explicit `'remote'` subscriber:

| notify ↓ subscribe → | `'local'` | `'remote'` | omitted (= `'local'`) |
| -------------------- | --------- | ---------- | --------------------- |
| `'local'`            | ✅        | ❌         | ✅                    |
| `'remote'`           | ✅        | ✅         | ✅                    |
| omitted              | ✅        | ✅         | ✅                    |

## Backwards-compatibility design

The channel is additive and the filtering is deliberately one-directional, which is what makes it safe:

- **Omitted = old behavior on both sides.** An unscoped `notify` reaches every subscriber (custom caches and every pre-channel producer keep working); an unscoped `subscribe` is defined as `'local'`, which receives everything (every pre-channel subscriber keeps working).
- **Fail-open tagging.** Because local views derive from remote state, a producer that *fails* to tag can only over-notify a remote reader, never starve one. Only an affirmative `'local'` tag — a producer asserting "only local state changed" — suppresses anything, and only for subscribers that affirmatively opted into `'remote'`.
- **Narrow filtering scope**: `errors`/`identity`/`state`/cache-operations are never channel-filtered.
- **No churn at legacy callsites**: legacy `Model`, `RecordState`, both references, and the deprecated `RecordReference` are untouched; the entire consumer-visible change is `notify('local')` tags in the cache/graph plus one `subscribe('remote')` for PolarisMode's default immutable record.
- The hasMany `flushCanonical` gate was iterated carefully (isDirty transition → `remoteOrderChanged` → final: `diff.changed`) specifically to preserve the deprecated `resetOnRemoteUpdate` reconciliation path, validated against the 5,344-test legacy app suite.

## Internal architectural changes

- **`NotificationManager` buffer is channel-aware.** The per-identifier flush buffer's `'attributes'`/`'relationships'` buckets went from key-sets to `Map<key, ChannelSet>` (or a wildcard `ChannelSet` for keyless touches), recording *which channels each key was touched on* during the flush window. Delivery is decided per subscriber at flush: a `'remote'` subscriber is skipped only when a key's set is exclusively `{'local'}`. A keyless (wildcard) touch folds in all previously-recorded per-key channels so the wildcard stays a superset both key-wise and channel-wise. The subscriber's channel is stashed concretely on the callback token at subscribe time.
- **`_pendingNotifies` in `CacheCapabilitiesManager`** changed from `Map<ResourceKey, Set<string>>` to `Map<ResourceKey, Map<key, Set<channel | undefined>>>`: relationship notifies are batched per-identifier *before* they reach `notify()`, and a channel-blind `Set<string>` would erase the tag at the batching layer — either losing a `'local'` tag (remote readers woken for purely-local changes, defeating the feature) or wrongly collapsing distinct channels onto one. At flush it regroups keys by channel and emits one `notify(identifier, 'relationships', Set, channel)` per distinct channel — batching preserved, and the NotificationManager's own buffer still coalesces down to one delivery per subscriber.
- **Graph operations**: purely-local mutation paths (`replace-related-records` local, `add/remove-from-related-records` with `isRemote=false`, the local branches of `addToInverse`/`removeFromInverse`) tag `'local'`; remote-state changes notify unconditionally and immediately (new notifies in the remote paths and in `merge-identifier`, which now notifies when either identity substitution or a local reference to the merged identifier changed) rather than relying on the `isDirty` latch — which could stick `true` forever for readers that never call `computeLocalState`. The `flushCanonical` gate is `diff.changed`.
- **json-api cache**: `setAttr` (flat and deep-path) tags `'local'`; upsert/didCommit/rollback notifies stay unscoped so they reach both channels; `notifyAttributes` was inlined away.
- **`ReactiveResource`**: one subscription per instance — `'remote'` for the default immutable Polaris instance, `'local'` for legacy/editable instances.

## Test coverage

- **`notification-channel-test.ts` (new)**: each row of the delivery matrix (default sub hears everything; `'local'` sub hears everything; `'remote'` sub skips only explicit-`'local'`); non-field notification types ignore channel; and three coalescing cases within a single flush window — same key touched `'local'`+unscoped (exactly one delivery to each subscriber), `'local'`+`'remote'` (one each), and *exclusively* `'local'` (zero for the remote sub, proving coalescing doesn't blur the skip).
- **`mutation-request-test.ts`**: end-to-end regression for the starvation bug — a local edit on a checked-out editable copy that matches the subsequent PATCH response still updates the default remote-only reader (incorporates and extends the regression test from #10495); plus the repeated-patch hasMany case (the #10494 repro) converted to `todo()` documenting the remaining pre-existing read-path bug described in the note above.
- Suites verified green: `tests/json-api` (91), `tests/ember-data__graph` (143), `tests/legacy` (132), `tests/dont-write-new-tests-here` (5,344).

https://claude.ai/code/session_01UtMujm1y6hRCYEjHTR1zSr


warp-drive-packages/json-api/src/-private/cache.ts
```diff
@@ -1824,9 +1826,14 @@ function getDefaultValue(
   }
 }
 
-function notifyAttributes(storeWrapper: CacheCapabilitiesManager, identifier: ResourceKey, keys?: NotifyKeys) {
+function notifyAttributes(
+  storeWrapper: CacheCapabilitiesManager,
+  identifier: ResourceKey,
+  keys?: NotifyKeys,
```
runspired: let's drop `notifyAttributes` as we can now inline this signature trivially

warp-drive-packages/core/src/graph/-private/operations/add-to-related-records.ts
```diff
@@ -57,7 +57,9 @@ export default function addToRelatedRecords(
     addRelatedRecord(graph, relationship, record, value, index ?? null, isRemote);
   }
 
-  notifyChange(graph, relationship);
+  // a purely local mutation (isRemote=false) has no remote implication, so
+  // remote-only readers don't need to be woken for it.
+  notifyChange(graph, relationship, isRemote ? undefined : 'local');
```
runspired: since `remote` is the same as an omitted notify, lets keep the signature monomorphic by passing 'remote'` here

warp-drive-packages/core/src/graph/-private/operations/remove-from-related-records.ts
```diff
@@ -50,7 +50,9 @@ export default function removeFromRelatedRecords(
     removeRelatedRecord(graph, record, relationship, value, op.index ?? null, isRemote);
   }
 
-  notifyChange(graph, relationship);
+  // a purely local mutation (isRemote=false) has no remote implication, so
+  // remote-only readers don't need to be woken for it.
+  notifyChange(graph, relationship, isRemote ? undefined : 'local');
```
runspired: same change here as above, keep the signature monomorphic

warp-drive-packages/core/src/reactive/-private/record.ts
```diff
@@ -252,7 +252,12 @@ export class ReactiveResource {
 
             break;
         }
-      }
+      },
+      // PolarisMode's default immutable instance shows only remote state, so it
+      // subscribes 'remote' to skip purely-local edits it can't see. Legacy-mode
+      // schemas and any editable instance (including a checked-out PolarisMode
+      // copy) show a reconciled local view and so hear everything (the default).
+      context.legacy || context.editable ? undefined : 'remote'
```
runspired: lets be explicit with `'local'` here

PR #10663, opened 2026-08-10
fix: restore docs build broken by tsdown/rolldown migration

runspired:
## Summary
The canary.warp-drive.io deployment workflow ([failing run](https://github.com/warp-drive-data/warp-drive/actions/runs/31432432562/job/93598780186)) has been red since the core (#10643) and build-config (#10651) tsdown/rolldown migrations, which switched `.d.ts` generation from a glob that typed every `src/**` file to a curated per-entry-point list. Two latent issues surfaced as a result:

- `core/src/types/runtime.ts` imported a type-only value from build-config's `-private/utils/logging` subpath, which was never a curated tsdown entry point and so no longer has a corresponding declaration file to resolve against. Fixed by deriving the same `LOG_CONFIG` shape locally via a namespace import of the already-public `@warp-drive/build-config/debugging` entry, instead of reaching into a private internal path.
- `legacy/src/store.ts` assigns its own `RecordReference` class (a long-standing full duplicate of core's `store/deprecated/-private.ts` version) onto `getReference`, which core's now-split-out `store/-private/store-service.ts` types as returning its own copy. The two classes are structurally identical but TypeScript treats their private fields as distinct nominal brands — the core migration's file split just made this previously-inert mismatch visible to typedoc's cross-package declaration checking. Fixed by casting only the return value (not the whole function), so `this` stays contextually typed inside the function body.

Both are minimal, non-behavioral fixes scoped to satisfying the type-checker; no runtime logic changed.

## Test plan
- [x] `pnpm install` (which runs `build:pkg` via turbo across all 35 workspace packages) succeeds
- [x] `tests/core` and `tests/legacy` `check:types` (`ember-tsc --noEmit`) pass cleanly
- [x] `eslint` on both touched files passes cleanly
- [ ] CI's `canary.warp-drive.io deployment` workflow goes green on this branch (the docs-build pipeline itself is flaky/non-deterministic locally across repeated runs, so real CI is the authoritative check)


warp-drive-packages/build-config/src/debugging.ts
```diff
@@ -4,6 +4,8 @@
  *
  * Visit {@link @warp-drive/core!build-config/debugging | @warp-drive/core debugging} for documentation of the debugging features.
  *
+ * See {@link @warp-drive/build-config!LOG_CONFIG} for the shape of the runtime logging configuration these flags feed into.
+ *
```
runspired: drop this

warp-drive-packages/build-config/src/index.ts
```diff
@@ -50,6 +50,8 @@ import { createLoggingConfig } from './-private/utils/logging.ts';
 import type { PluginItem } from '@babel/core';
 import { buildMacros } from '@embroider/macros/src/babel.js';
 
+export type { LOG_CONFIG } from './-private/utils/logging.ts';
```
runspired: drop this

warp-drive-packages/core/src/types/runtime.ts
runspired: link to the debugging page

PR #10928, opened 2026-08-28
chore: bump pnpm to 12.0.0, replace Volta with mise

runspired:
## Summary
- Bump pnpm from 11.22.0 to 12.0.0 (root, tests/core, tests/legacy, warp-drive-packages/schema-dsl). 12.0.0 was published 2026-08-26 and is still inside the 7-day `minimumReleaseAge` window as of this PR -- added a one-off, dated `minimumReleaseAgeExclude` entry to unblock it (safe to drop after ~2026-09-02).
- pnpm 12 renamed `managePackageManagerVersions` to `initPackageManager` and now hard-fails (`ERR_PNPM_UNRECOGNIZED_WORKSPACE_SETTINGS`) on any unrecognized `pnpm-workspace.yaml` key instead of silently ignoring it -- updated the setting name accordingly.
- **Replaced Volta with mise for local node/pnpm version management.** pnpm 12.0.0 ships ESM-only bins (`bin/pnpm.mjs`, whose own header comment says it's Corepack's entry point -- there's no `bin/pnpm.cjs` in the published package anymore). Volta (tested through the latest 2.0.2) synthesizes a global shim assuming the old `.cjs` layout and crashes with `MODULE_NOT_FOUND`; this is a Volta-side limitation, not something fixable from the repo. mise already had a foothold here (`mise.toml` pinning `node`) and installs pnpm from standalone per-platform GitHub release binaries rather than the npm package's bin script, sidestepping the issue entirely.
- Removed the `"volta"` field (and its `extends` pointer back to root) from every workspace `package.json` (62 files), and the now-dead volta-specific code paths: the mirror-tarball publish step no longer needs to rewrite `volta.extends` paths, and docs-viewer's unused `maybeMakePNPMInstallable`/`getCurrentVersion` helpers (which pinned `volta.node`/`volta.pnpm` in cloned example-app repos) are gone.
- Updated contributor docs (`setting-up-the-project.md`, `RELEASE.md`, `canary.md`, PR template) to reference mise instead of Volta.
- Regenerated `pnpm-lock.yaml` -- purely additive: pnpm 12 now self-tracks its own per-platform binaries as a `packageManagerDependencies` entry in a new leading YAML document ahead of the existing lockfile content. No dependency resolutions changed.

## Test plan
- [x] `pnpm install` from a clean `node_modules`, via `mise exec -- pnpm install` -- no warnings or errors
- [x] `pnpm check:types` -- 83/83 tasks pass
- [x] `pnpm lint` -- 86/86 tasks pass
- [x] `pnpm lint:prettier` -- clean
- [x] Verified lockfile diff is additive only (new pnpm self-tracking metadata), no dependency version changes
- [ ] CI green
- [ ] Contributors with Volta installed locally will need to switch to mise (`brew install mise`, activate in shell profile, `mise install`) -- flagged in the updated setup docs


packages/diagnostic/package.json
```diff
@@ -2,7 +2,7 @@
   "name": "@warp-drive/diagnostic",
   "version": "0.0.3-alpha.151",
   "private": true,
-  "description": "⚡️ A Lightweight Modern Test Runner",
+  "description": "\u26a1\ufe0f A Lightweight Modern Test Runner",
```
runspired: explain this unrelated change

docs-viewer/src/-utils.ts
runspired, line 1: explain these unrelated deletions

packages/holodeck/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@warp-drive/holodeck",
-  "description": "⚡️ Simple, Fast HTTP Mocking for Tests",
+  "description": "\u26a1\ufe0f Simple, Fast HTTP Mocking for Tests",
```
runspired: explain this unrelated change

packages/diagnostic/package.json
runspired: Bug on my end: the script I used to strip the `volta` field from all 62 workspace `package.json` files used Python's `json.dump` without `ensure_ascii=False`, which silently re-encoded this file's ⚡️ emoji in `description` into a `⚡️` escape sequence. Fixed in 8baba59cc, along with the same issue in `packages/holodeck/package.json` (the only two files affected).

packages/holodeck/package.json
runspired: Same root cause as the diagnostic package.json comment -- fixed in 8baba59cc.

docs-viewer/src/-utils.ts
runspired: These are the `maybeMakePNPMInstallable`/`getCurrentVersion` helpers, which existed solely to pin `volta.node`/`volta.pnpm` in cloned example-app repos (e.g. ember-api-docs) so their own `pnpm install` would work under Volta. Neither function is imported/called anywhere else in the tracked repo -- I checked before removing them. Since this PR drops Volta from the toolchain entirely, they were dead code the moment volta stopped being how we manage pnpm, so I deleted them along with the volta field removal rather than leave Volta-specific dead code behind. Happy to revert if you'd rather keep them (e.g. if something outside the repo calls them, or you want a slower deprecation).

PR #10992, opened 2026-09-02
Defer relationship notifications to store transaction notify phase

runspired:
Related RFC https://github.com/emberjs/rfcs/pull/1232 

Note this is not currently intended to be the implementation of that RFC, but contains groundwork exploration of moving towards transactions that will be used to inform the RFCs design.

## Summary

Refactors notification handling to defer `'relationships'` notifications that arrive during store transactions (within `store._run` or `store._join`) to the transaction's notify phase, rather than flushing them immediately mid-mutation. This ensures the graph can finish syncing relationship state before subscribers are notified.

### NotificationManager
- Added `_hasTransactionFlush` flag to track pending transaction-deferred flushes
- Modified `notify()` to detect `'relationships'` notifications arriving while `store._cbs` is active and schedule them into the transaction's notify phase via `store._schedule('notify', ...)`
- Added `_flushTransaction()` method to handle the deferred flush at the end of the transaction
- Refactored `_flush()` to handle partial flushes when a transaction-deferred relationships flush is still pending, extracting and delivering non-relationship notifications while preserving relationship buckets for the transaction's notify phase
- Extracted notification delivery logic into a new `_deliver()` helper method

### CacheCapabilitiesManager
- Removed `_willNotify`, `_pendingNotifies`, `_scheduleNotification()`, and `_flushNotifications()` methods
- Removed special-case batching logic for relationship notifications — this is now handled directly by NotificationManager's buffer and transaction scheduling
- Simplified `notifyChange()` to pass all notifications directly to `store.notifications.notify()`
- Removed cleanup of `_pendingNotifies` in `disconnectRecord()`

### Tests
- Updated test expectations in `notification-batch-test.ts` to reflect that relationship notifications are now delivered per-key rather than batched into a single `notify()` call
- Removed assertions about `notify()` call counts and Set types, as the batching now happens at the buffer level rather than at the `notify()` call level

## Rationale

The previous approach batched relationship notifications at the `CacheCapabilitiesManager` level before passing them to `NotificationManager`. This change moves that responsibility entirely to `NotificationManager`, which can now defer relationship notifications to the transaction's notify phase while allowing other notification types to flush immediately. This ensures:

1. Relationship state syncing completes before subscribers are notified
2. Other notification types (e.g., `'state'`) are still delivered immediately, as consumers depend on this timing
3. Subscribers still receive exactly one notification per changed relationship, delivered synchronously before the transaction returns

## Test Plan

Existing tests pass. The `notification-batch-test.ts` test has been updated to verify that subscribers receive exactly one notification per changed relationship, which is the key invariant this change maintains.

https://claude.ai/code/session_017CMTUUPwpZXGjqCM8nUDrP

### DIFF
--- a/tests/json-api/tests/integration/cache/notification-batch-test.ts
+++ b/tests/json-api/tests/integration/cache/notification-batch-test.ts
@@ -202,7 +202,7 @@ module('Integration | NotificationManager batch notifications', function () {
     store.notifications.unsubscribe(token);
   });
 
-  test('relationship changes on the same identifier are flushed to `notify` as a single batch', function (assert) {
+  test('relationship changes on the same identifier are delivered exactly once per changed relationship', function (assert) {
     const store = new RelationshipTestStore();
 
     const doc = store.cache.put(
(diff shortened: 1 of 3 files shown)


tests/json-api/tests/integration/cache/notification-batch-test.ts
```diff
@@ -202,7 +202,7 @@ module('Integration | NotificationManager batch notifications', function () {
     store.notifications.unsubscribe(token);
   });
 
-  test('relationship changes on the same identifier are flushed to `notify` as a single batch', function (assert) {
+  test('relationship changes on the same identifier are delivered exactly once per changed relationship', function (assert) {
```
runspired, line 205: how is this test actually testing this claim?

PR #10993, opened 2026-09-02
Notify remote-state changes per notification channel (fixes stale linksMode hasMany)

### LINKED ISSUE #10494: Repeated remote updates can leave accessed linksMode hasMany relationships stale

runspired:
## Description

Fixes #10494: a materialized `linksMode` hasMany went stale after repeated remote updates because the second update's diff compared against a `localState` snapshot that remote-only readers never refresh, so `diff.changed` came out `false` and no notification fired.

### The Problem

A `CollectionEdge`'s `localState` snapshot is only ever refreshed by `computeLocalState`, which only runs on the *local*-state read path. A readonly (non-editable) `linksMode` ManyArray renders and re-syncs from *remote* state, so after the first remote update the snapshot stays permanently stale — and the notification gate (`diff.changed`) trusted that snapshot as "what the app currently shows". A later remote change that happened to match the stale snapshot was judged to be no change, and the rendered array never updated.

### The Solution

Rather than making reads refresh the snapshot (the first iteration of this PR), notification correctness no longer depends on `localState` freshness for remote readers at all. With dual-channel notifications, each copy of a record subscribes to the channel of the projection it renders and re-pulls that projection when notified:

- The delivery matrix is now symmetric: a notify tagged `'remote'` is delivered only to `'remote'` subscribers, `'local'` only to `'local'`/default subscribers, and an unscoped notify still reaches everyone (pre-channel compatibility).
- `replaceRelatedRecordsRemote` notifies per projection: `'remote'` when `diff.remoteOrderChanged` (remote membership/order actually changed), `'local'` when `diff.changed` (the reconciled local view changed). When both change, the graph's flush queue — which now tracks the channel per edge — coalesces them into a single unscoped notification.
- A remote push that merely confirms a committed local mutation now notifies exactly the remote channel: the readonly projection genuinely changed (the commit landed) while the local view did not.
- `getHasManyField`'s readonly path now reads remote state, mirroring `getBelongsToField` and `ManyArrayManager._syncArray`, so the readonly `linksMode` hasMany renders, refreshes, and is notified about the same projection.
- Remote add/remove operations, which change both projections, notify unscoped instead of relying on the old "`'remote'` reaches every subscriber" behavior.

### Changes

- **core/graph**: per-channel notification gates in `replaceRelatedRecordsRemote`; channel-aware flush queue in `Graph`; unscoped notifies for remote add/remove ops
- **core/store**: symmetric channel delivery in `NotificationManager` (docs/matrices updated here and in `CacheCapabilitiesManager`)
- **core/reactive**: readonly hasMany initial read pulls remote state; subscription-channel comment updated
- **tests/json-api**: `todo` → `test` for "update hasMany with repeated patch" (now passes); channel-matrix tests updated to the symmetric contract
- **tests/ember-data__graph**: diff-preservation tests now expect exactly one remote-channel notification when a push confirms a committed mutation (still zero for the local view)
- **unpublished-test-infra**: `watchNotifications` no longer drops the `channel` argument when forwarding to the real `notify`

### Test Plan

All green locally on Node 26.8.1: `tests/json-api` (92 pass / 0 fail, including the previously-failing repeated-patch test), `tests/ember-data__graph` (143/143), `tests/legacy` (132 / 0), main-test-app (5344 / 0), `tests/core` (202 / 0), plus lint, format, and type checks.

https://claude.ai/code/session_01N4uRTAvfmeoeZVNjcqJ9os

### DIFF
--- a/tests/ember-data__graph/tests/integration/graph/diff-preservation-test.ts
+++ b/tests/ember-data__graph/tests/integration/graph/diff-preservation-test.ts
@@ -892,7 +892,7 @@ module('Integration | Graph | Diff Preservation', function (hooks) {
     );
   }
 
-  test('updateRelationship operation matching localState does not produce a notification for a committed removal (resetOnRemoteUpdate=false)', function (assert) {
+  test('updateRelationship operation matching localState notifies only the remote channel for a committed removal (resetOnRemoteUpdate=false)', function (assert) {
     class User extends Model {
       @attr declare name: string;
       @hasMany('comment', { async: false, inverse: 'user', resetOnRemoteUpdate: false }) declare comments: Comment[];
@@ -1008,8 +1008,8 @@ module('Integration | Graph | Diff Preservation', function (hooks) {
       userIdentifier,
       'relationships',
       'comments',
-      0,
-      'We should have no notifciations after remote push matches local state'
+      1,
+      'one remote-channel notification after a remote push that matches local state: the remote view changed (the commit landed) while the local view did not'
     );
 
     // push an update that does not match the local state
@@ -1041,7 +1041,7 @@ module('Integration | Graph | Diff Preservation', function (hooks) {
     assert.arrayStrictEquals(data4.data, [comment1Identifier, comment4Identifier], 'state is updated');
   });
 
-  test('updateRelationship operation matching localState does not produce a notification for a committed addition (resetOnRemoteUpdate=false)', function (assert) {
+  test('updateRelationship operation matching localState notifies only the remote channel for a committed addition (resetOnRemoteUpdate=false)', function (assert) {
     class User extends Model {
       @attr declare name: string;
       @hasMany('comment', { async: false, inverse: 'user', resetOnRemoteUpdate: false }) declare comments: Comment[];
@@ -1134,8 +1134,8 @@ module('Integration | Graph | Diff Preservation', function (hooks) {
       userIdentifier,
       'relationships',
       'comments',
-      0,
-      'no notifications after remote push matches local state'
+      1,
+      'one remote-channel notification after a remote push that matches local state: the remote view changed (the commit landed) while the local view did not'
     );
 
     // check state is still the same
@@ -1181,7 +1181,7 @@ module('Integration | Graph | Diff Preservation', function (hooks) {
     );
   });
 
-  test('updateRelationship operation matching localState does not produce a notification for a committed removal', function (assert) {
+  test('updateRelationship operation matching localState notifies only the remote channel for a committed removal', function (assert) {
     class User extends Model {
       @attr declare name: string;
       @hasMany('comment', { async: false, inverse: 'user', resetOnRemoteUpdate: true }) declare comments: Comment[];
@@ -1297,8 +1297,8 @@ module('Integration | Graph | Diff Preservation', function (hooks) {
       userIdentifier,
       'relationships',
       'comments',
-      0,
-      'We should have no notifciations after remote push matches local state'
+      1,
+      'one remote-channel notification after a remote push that matches local state: the remote view changed (the commit landed) while the local view did not'
     );
 
     // push an update that does not match the local state
@@ -1330,7 +1330,7 @@ module('Integration | Graph | Diff Preservation', function (hooks) {
     assert.arrayStrictEquals(data4.data, [comment1Identifier, comment4Identifier], 'state is updated');
   });
 
-  test('updateRelationship operation matching localState does not produce a notification for a committed addition', function (assert) {
+  test('updateRelationship operation matching localState notifies only the remote channel for a committed addition', function (assert) {
     class User extends Model {
       @attr declare name: string;
       @hasMany('comment', { async: false, inverse: 'user', resetOnRemoteUpdate: true }) declare comments: Comment[];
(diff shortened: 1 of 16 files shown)


tests/ember-data__graph/tests/integration/graph/diff-preservation-test.ts
```diff
@@ -1008,8 +1008,8 @@ module('Integration | Graph | Diff Preservation', function (hooks) {
       userIdentifier,
       'relationships',
       'comments',
-      0,
-      'We should have no notifciations after remote push matches local state'
+      1,
+      'one remote-channel notification after a remote push that matches local state: the remote view changed (the commit landed) while the local view did not'
```
runspired: how can we assert this was a remote and not a local notification? and how can we assert the local was not notified?

warp-drive-packages/core/src/store/-private/managers/notification-manager.ts
```diff
@@ -214,13 +206,13 @@ function mergeIntoBuffer(
...
- * and a touch-set that was exclusively tagged `'local'`. A `'local'` (or
- * defaulted) subscriber receives everything.
+ * should be delivered to (see the delivery matrix on
+ * {@link NotificationChannel}): a subscriber hears a touch-set if any of its
+ * touches was unscoped or tagged with the subscriber's own channel; a
+ * touch-set exclusively tagged with the *other* channel is skipped.
  */
 function shouldDeliverToChannel(channels: ChannelSet, subscriberChannel: NotificationChannel): boolean {
-  return subscriberChannel !== 'remote' || channels.has(undefined) || channels.has('remote');
+  return channels.has(undefined) || channels.has(subscriberChannel);
```
runspired: we should use a well-known key here that isn't `undefined`

tests/ember-data__graph/tests/integration/graph/diff-preservation-test.ts
runspired: Added per-channel tracking to `assert.watchNotifications` and a new `assert.notifiedOn(channel, identifier, bucket, key, count, message?)` probe (`channel` is `'local' | 'remote' | 'unscoped'`, where `'unscoped'` counts notifies that carried no channel). This site now asserts all three facts explicitly:

```ts
assert.notifiedOn('remote', userIdentifier, 'relationships', 'comments', 1, ...);
assert.notifiedOn('local', userIdentifier, 'relationships', 'comments', 0, ...);
assert.notified(userIdentifier, 'relationships', 'comments', 1, ...);
```

`notifiedOn` is a non-clearing probe; `notified` still closes out the step by clearing both the total and per-channel counts. Also applied to the other steps in these four tests: the mutation now asserts it notified `'local'`, and the non-matching push asserts the two channel schedules coalesced into a single `'unscoped'` notification. (While in there I fixed `watchNotifications`' notify interceptor, which was silently dropping the `channel` argument when forwarding to the real `notify`.)

Pushed in 7382cd6.

---

warp-drive-packages/core/src/store/-private/managers/notification-manager.ts
runspired: Replaced with a module-level well-known sentinel:

```ts
const UnscopedChannel = Symbol('unscoped');
type ChannelSet = Set<NotificationChannel | typeof UnscopedChannel>;
```

`mergeIntoBuffer` now records `channel ?? UnscopedChannel` and `shouldDeliverToChannel` checks `channels.has(UnscopedChannel)`; the docs that described "`undefined` in the set" are updated to match. A `Symbol` keeps the sentinel out of the `NotificationChannel` value space so a future channel name can never collide with it.

Pushed in 7382cd6.

---

PR #10994, opened 2026-09-02
Feat(schema-dsl) add schemaObject and schemaArray

MehulKChaudhari:
This PR adds schemaObject and schemaArray support to the DSL


runspired (approved):
there's some follow up around the flattening of options we should do - but schema-dsl repo is private currently anyway so merging this is safe.

PR #11043, opened 2026-09-08
Parameterize reactive document types over `meta` and `errors`

gitKrystan:
### What

The reactive document types gain three optional type params, so a request can declare what its endpoint actually returns instead of every caller coercing.

**`M extends Meta = Meta` for `meta`.** Today `meta` is hardcoded to `Meta` — `ObjectValue`, an arbitrary JSON object — so reading a documented key gives `unknown` and every callsite coerces:

```ts
Number(content.meta?.total ?? 0); // meta.total is unknown
```

With the param:

```ts
type PageMeta = { page: { limit: number; offset: number }; total?: number };

const { content } = await store.request(query<User, PageMeta>('user'));
content.meta?.total; // number | undefined
```

**`E extends object = object` for `errors`.** `ReactiveErrorDocument['errors']` was `object[]`. The obvious "fix" — retyping it `ApiError[]` — turns out to be a claim the runtime does not make: `fromStructuredError` ([`json-api/src/-private/cache.ts:2124-2130`](https://github.com/warp-drive-data/warp-drive/blob/main/warp-drive-packages/json-api/src/-private/cache.ts#L2124-L2130)) assigns `doc.content.errors` straight through without validating it, so a non-{json:api} endpoint can land anything in that array. Parameterizing it keeps `object` as the honest default and lets the request say more when it knows more.

The builders in `@warp-drive/utilities/json-api` default `E` to [`ApiError`](https://github.com/warp-drive-data/warp-drive/blob/main/warp-drive-packages/core/src/types/spec/error.ts) — the [{json:api} error object](https://jsonapi.org/format/#error-objects) — because that is the shape the {json:api} spec calls for. The `rest` and `active-record` builders leave it `object`, because neither REST nor ActiveRecord specifies an error shape.

**`EM extends Meta = M` for the error document's `meta`.** A failed request need not carry the same `meta` a successful one does, and the union was claiming the success meta on both arms. `EM` defaults to `M`, so an API that returns one envelope either way is unaffected.

Each arm names its own meta as its **second** param; the **fourth** is the other variant's, needed only so `next`/`prev`/`fetch` can type the union they resolve with:

```ts
type UsersDocument = ReactiveDataDocument<User[], PageMeta, ApiError, ErrorMeta>;
type UsersErrorDocument = ReactiveErrorDocument<User[], ErrorMeta, ApiError>;
```

**`ApiError['source']['pointer']` is now optional.** The spec says `source` "SHOULD include one of the following members or be omitted: `pointer`, `parameter`, `header`" — an error sourced from a query parameter legitimately carries `parameter` and no `pointer`, which the required type could not express.

### Where the params are threaded

- `ReactiveDocumentBase`, `ReactiveDataDocument`, `ReactiveErrorDocument`, `ReactiveDocument`, `createReactiveDocument` — all three params
- `withReactiveResponse<T, M, E, EM>` and the deprecated `Document<T, M, E, EM>` alias — all three
- `query`, `postQuery`, `findRecord` and `updateRecord` across `json-api`, `rest` and `active-record` — `M` and `E` only

`EM` is deliberately not on the builders: it would put a fourth positional on nine signatures for a case you reach at the consumption site anyway (see the note below). `fetch`, `next`, `prev`, `first` and `last` carry every param through to the document they resolve with. Every default preserves today's behavior.

`meta` also moved from `ReactiveDocumentBase` onto each arm, since the two now hold different types. The base is not exported.

### Tests

**`tests/utilities/tests/unit/document-type-params-test.ts`** — type-level assertions that each param reaches its member, that the defaults are unchanged, that link-following preserves them, and that the per-builder `E` defaults are what they claim.

**`tests/json-api/tests/integration/cache/error-document-meta-test.ts`** — mocks a 422 whose body carries `meta`, and reads it back through `getRequestState<Doc, ErrDoc>`. This is the typed path: a failed request *rejects* rather than resolving with the error arm, and `getRequestState<RT, E>` takes the error content type as its own param. It asserts the value survives `copyLinksAndMeta` at runtime, and that `EM`, `E` and the `EM = M` default reach the members they should.

Two traps the assertions had to work around, both noted in the files:

- `ApiError` is a **weak type** — every member is optional — so `object` is assignable to it and an exactness assertion on `errors` passes even when `E` stops being applied. Each error assertion also probes a member (`['status']`).
- `E` never appears on the data arm, only in method return positions, which TypeScript compares bivariantly. The assertions reach the error arm through `next()`'s union.

Verified load-bearing — each of these fails `check:types`:

- reverting `meta?: M` to `meta?: Meta`
- reverting `errors: E[]` to `errors: object[]`
- changing the `json-api` builders' `E` default from `ApiError` back to `object`
- reverting the error arm's `meta?: EM` to `meta?: M`

### Docs

`guides/the-manual/requests/typing-requests.md` gains "Typing the Document's `meta`", "Typing Errors" and a nested "Typing the Error Document's `meta`", covering all three params, the per-builder `E` defaults, why `M` must be a `type` alias rather than an `interface` (TypeScript gives implicit index signatures to aliases only), the `never` idiom for a meta-only endpoint, and the `getRequestState` path to a typed error document. Typedoc added on both arms' `meta`, on `errors`, on `withReactiveResponse` and on `ApiError['source']`; the `json-api` module doc states its own `E` default.

### Notes for review

@runspired — judgment calls worth your eye:

**`Meta` is now imported into the `rest` and `active-record` builders.** `Meta` has exactly one home, `@warp-drive/core/types/spec/json-api-raw`, so those two families now import from a module named for {json:api}. The coupling is pre-existing rather than introduced here — they already return `ReactiveDataDocument` on `main`, and that type's `meta` has always been `Meta` — but if you would rather they not name that module, the alternatives are dropping `M` from them or re-exporting `Meta` from a neutral path.

**`E` defaults to `ApiError` only in the `json-api` builders.** That is the one place the contract is actually specified. Happy to make all three `object` and have {json:api} callers pass `ApiError` explicitly if you would rather no builder make a claim about a payload it does not validate.

**The two paths that hand back an error document are mistyped in opposite directions, and this PR does not fix either.** `next`/`prev`/`fetch` declare `Promise<ReactiveDocument<…> | null>` but the implementation returns the data arm and throws on failure; the throw path sets `newError.content` to the hydrated error document while `StructuredErrorDocument<T>` declares `content?: T` as the *success* type. `getRequestState<RT, E>` is the one place you can name the error content today, which is why the new test goes through it. Worth a follow-up.

**The cache-side type still makes the claim this PR backs away from.** `ResourceErrorDocument.errors` is `ApiError[]` on `main` (`types/spec/document.ts:120`), filled from unvalidated payload — the same overclaim, one layer down. Left alone here since changing it reaches into the `Cache` interface (`put`, `getErrors`, `commitWasRejected`).

**A meta-only response has no dedicated arm on `ReactiveDocument`.** `ResourceMetaDocument` (`meta` required, no `data`, no `errors`) is a first-class cache document that flows through the same hydration path, so it lands on `ReactiveDataDocument`. With these params that is expressible — `ReactiveDataDocument<never, CountMeta>`, which the guide documents — where `data` reads as `never`. Flagging it because the alternative, a third `ReactiveMetaDocument` arm, would change narrowing for every consumer of `ReactiveDocument`.

### Verification

- `pnpm check:types` — 83/83
- `pnpm lint` — 44/44, `pnpm lint:oxlint` and `pnpm lint:oxfmt` clean
- `utilities-tests`, `core-tests`, `@warp-drive/legacy-tests`, `json-api`, `main-test-app` — 38/38 tasks, 5876 pass, 0 fail


guides/the-manual/requests/typing-requests.md
```diff
@@ -177,6 +177,75 @@ function getUsers() {
 
 :::
 
+## Typing the Document's `meta`
+
+A reactive document also carries the response's [meta](https://jsonapi.org/format/#document-meta). By
+default it is typed `Meta` — an arbitrary JSON object — so reading a key off it gives you `unknown`
+and every callsite ends up coercing:
```
gitKrystan:
```suggestion
default it is typed `Meta` — an arbitrary JSON object — so reading a key off it gives you `unknown`:
```

guides/the-manual/requests/typing-requests.md
```diff
@@ -177,6 +177,75 @@ function getUsers() {
...
+
+```ts
+const { content } = await store.request(getUsers());
+
+Number(content.meta?.total ?? 0); // meta.total is unknown
+```
+
+Both `ReactiveDataDocument` and `withReactiveResponse` take a second type param for the meta, so the
+request declares once what its endpoint returns:
```
gitKrystan:
```suggestion
Both `ReactiveDataDocument` and `withReactiveResponse` take an optional second type param for the meta:
```

guides/the-manual/requests/typing-requests.md
```diff
@@ -177,6 +177,75 @@ function getUsers() {
...
+  return withReactiveResponse<User[], PageMeta>({
+    url: '/users'
+  });
+}
+
+const { content } = await store.request(getUsers());
+content.meta?.total; // number | undefined
+```
+
+The param must be a `type` alias rather than an `interface`: `Meta` is an index-signature type, and
```
gitKrystan:
```suggestion
NOTE: The param must be a `type` alias rather than an `interface`. This is because `Meta` is an index-signature type, and
```

warp-drive-packages/utilities/src/-private/active-record/find-record.ts
```diff
@@ -5,6 +5,7 @@ import type {
   FindRecordRequestOptions,
   RemotelyAccessibleIdentifier,
 } from '@warp-drive/core/types/request';
+import type { Meta } from '@warp-drive/core/types/spec/json-api-raw';
```
gitKrystan, line 8: It doesn't make sense for active-record builders to be importing from json-api stuff. Meta might not be applicable here? Please confirm.

warp-drive-packages/utilities/src/-private/rest/find-record.ts
```diff
@@ -5,6 +5,7 @@ import type {
   FindRecordRequestOptions,
   RemotelyAccessibleIdentifier,
 } from '@warp-drive/core/types/request';
+import type { Meta } from '@warp-drive/core/types/spec/json-api-raw';
```
gitKrystan, line 8: Same comment here as I had for active-record

warp-drive-packages/core/src/reactive/-private/document.ts
```diff
@@ -130,9 +143,12 @@ export interface ReactiveErrorDocument<T> extends ReactiveDocumentBase<T> {
   /**
    * The errors returned by the API for this request, if any
    *
+   * Each entry is an {@link ApiError}, the same shape the cache stores for
+   * an error document.
+   *
    * @public
    */
-  readonly errors: object[];
+  readonly errors: ApiError[];
```
gitKrystan: Is this type re-used by active-record and/or rest utils? If so, that might be why it was `object` before and we might need to do this `errors: ApiError[]` change somewhere else

warp-drive-packages/core/src/types/spec/error.ts
```diff
@@ -52,12 +52,16 @@ export interface ApiError {
   code?: string;
   /**
    * an object containing references to the primary source of the error
+   *
+   * The spec says this SHOULD include one of `pointer`, `parameter` or
+   * `header` — an error sourced from a query param carries `parameter` and no
+   * `pointer` — so every member is optional here.
```
gitKrystan:
```suggestion
  * an object containing references to the primary source of the error
   *
   * The spec says this SHOULD include one of `pointer`, `parameter` or
   * `header`. For example, an error sourced from a query param includes `parameter` and no
   * `pointer`.
```

guides/the-manual/requests/typing-requests.md
runspired, line 1: I think one question I'm left with is do we use the same type for error and success? or do we instead come up with something that could discriminate? or for the error case does meta drop back to being `Meta`?

guides/the-manual/requests/typing-requests.md
```diff
@@ -177,6 +177,100 @@ function getUsers() {
...
+
+The error variant of a document, [ReactiveErrorDocument](/api/@warp-drive/core/reactive/interfaces/ReactiveErrorDocument),
+exposes `errors`. The cache stores whatever the API sent without validating it, so by default the
+type promises no shape — `errors` is `object[]`.
+
+A third type param declares what the endpoint actually returns. The builders in
+`@warp-drive/utilities/json-api` default it to
+[ApiError](/api/@warp-drive/core/types/spec/error/interfaces/ApiError), the
+[{json:api} error object](https://jsonapi.org/format/#error-objects), because that is what the
+dialect specifies. `status` there is a string, not a number, and every member is optional:
```
gitKrystan:
```suggestion
`@warp-drive/utilities/json-api` default it to
[ApiError](/api/@warp-drive/core/types/spec/error/interfaces/ApiError), the
[{json:api} error object](https://jsonapi.org/format/#error-objects):
```

guides/the-manual/requests/typing-requests.md
```diff
@@ -177,6 +177,100 @@ function getUsers() {
...
+const { content } = await store.request(query<User>('user'));
+const nextPage = await content.next(); // resolves with the document union
+
+if (nextPage?.errors) {
+  nextPage.errors[0].status; // string | undefined
+  nextPage.errors[0].source?.pointer; // string | undefined
+}
+```
+
+The `rest` and `active-record` builders leave it as `object`, since neither dialect specifies an
```
gitKrystan: Is "dialect" in use elsewhere in these docs or does this PR add it?

warp-drive-packages/utilities/src/json-api.ts
```diff
@@ -45,6 +45,10 @@
  * - URLs are stable. The same query will produce the same URL every time, even if the order of keys in
  *   the query or values in an array changes.
  * - URLs follow the most common {json:api} format (dasherized pluralized resource types)
+ * - The response document's `errors` are typed as `ApiError`, the
+ *   [{json:api} error object](https://jsonapi.org/format/#error-objects), since that is the shape
+ *   the spec calls for. Pass the builder's third type param to say otherwise. The `rest` and
+ *   `active-record` builders leave it as `object`, because neither dialect specifies an error shape.
```
gitKrystan:
```suggestion
 * - The response document's `errors` are typed as `ApiError`, the
 *   [{json:api} error object](https://jsonapi.org/format/#error-objects).
```

warp-drive-packages/core/src/reactive/-private/document.ts
```diff
@@ -17,7 +17,7 @@ function urlFromLink(link: Link): string {
   return link.href;
 }
 
-export interface ReactiveDocumentBase<T> {
+export interface ReactiveDocumentBase<T, M extends Meta = Meta, E extends object = object> {
```
gitKrystan: NOTE: The two params currently have different constraints for backwards compat:

- `M extends Meta`, where `Meta = ObjectValue` — a string index signature plus recursive JSON-serializability.
- `E extends object` — rules out primitives and nothing else.

Do we want to keep this asymmetry or fix it now?

Claude dump:

The reason is narrow: `ApiError` is declared as an `interface` (`types/spec/error.ts:18`), and
interfaces get no implicit index signature, so it can't satisfy `E extends Meta`. Flipping the
constraint to check produces `Type 'ApiError' does not satisfy the constraint 'ObjectValue'` at the
`json-api` builder callsites.

On the merits the two members want the same constraint. Both are unvalidated payload passthrough,
both round-trip the cache, and both get copied into `toJSON()` — so the serializability guarantee
`M` carries is one `errors` arguably wants too. It also leaks to users in an inconsistent way: `M`
must be supplied as a `type` alias and an `interface` fails, while `E` takes either.

Three possible fixes:

1. **Make `ApiError` a `type` alias**, then use `E extends Meta`. Cheapest, and errors gain the same
   serializability guarantee. But `ApiError` is public and load-bearing — `Cache.getErrors`,
   `commitWasRejected`, `AdapterRequestError.errors` — and it would break anyone relying on
   declaration merging.
2. **Leave it as-is.** Errors stay less constrained than meta; the type-alias rule the guide
   documents for `M` is silently not true for `E`.
3. **Loosen `M` to `object` too.** Symmetric, but drops a guarantee `meta` actually depends on.

guides/the-manual/requests/typing-requests.md
gitKrystan: I ended up adding an `EM extends Meta = M`

guides/the-manual/requests/typing-requests.md
```diff
@@ -177,6 +177,131 @@ function getUsers() {
...
+
+```ts
+const future = store.request<UsersDocument>({ url: '/users' });
+const state = getRequestState<UsersDocument, UsersErrorDocument>(future);
+
+if (state.isError) {
+  state.reason.content?.meta?.requestId; // string | undefined
+  state.reason.content?.errors[0].status; // string | undefined
+}
+```
```
gitKrystan:
```suggestion
### Typing the Error Document's `meta`

Because error responses may return different `meta` than successful responses, you can pass an optional fourth type param to type the
error document's `meta` separately (otherwise it will fall back to the same `meta` type as the success response):

```ts
type PageMeta = { page: { limit: number; offset: number } };
type ErrorMeta = { requestId: string };

type UsersDocument = ReactiveDataDocument<User[], PageMeta, ApiError, ErrorMeta>;
type UsersErrorDocument = ReactiveErrorDocument<User[], ErrorMeta, ApiError>;
```

You can access the error document through [getRequestState](/api/@warp-drive/core/reactive/functions/getRequestState),
which takes the error content type as an optional second param:

```ts
const future = store.request<UsersDocument>({ url: '/users' });
const state = getRequestState<UsersDocument, UsersErrorDocument>(future);

if (state.isError) {
  state.reason.content?.meta?.requestId; // string | undefined
  state.reason.content?.errors[0].status; // string | undefined
}
```
```

File: tests/json-api/tests/integration/cache/error-document-meta-test.ts [hunk 13]
```diff
@@ -0,0 +1,121 @@
...
+}
+
+/** What this endpoint returns alongside a successful collection. */
+type UserPageMeta = { page: { limit: number; offset: number }; total?: number };
+
+/** What it returns alongside a failure — deliberately a different shape. */
+type UserErrorMeta = { requestId: string };
+
+type UsersDocument = ReactiveDataDocument<User[], UserPageMeta, ApiError, UserErrorMeta>;
+type UsersErrorDocument = ReactiveErrorDocument<User[], UserErrorMeta, ApiError, UserPageMeta>;
```
gitKrystan: It's this now, right?

```suggestion
type UsersErrorDocument = ReactiveErrorDocument<User[], UserErrorMeta, ApiError>;
```

runspired (approved):

PR #11088, opened 2026-09-11
fix(json-api): notify the remote channel for attrs "promoted" to remote state by `commit`

gitKrystan:
## The bug

**After a save succeeds, a rendered component can keep showing the pre-save value.**

The cache itself is correct (the new value is in `remoteAttrs`), but we don't notify as expected. Thus, anything already consuming the old value is never told to re-read it and stays stale until something else coincidentally invalidates it.

The affected reader is the **default immutable record**. The editable copy from `checkout()` shows the new value correctly. For example:

```ts
import { recordIdentifierFor } from '@warp-drive/core';
import { checkout } from '@warp-drive/core/reactive';
import { findRecord, serializePatch, updateRecord } from '@warp-drive/utilities/json-api';

// cache holds firstName: 'Chris'
const { content } = await store.request(findRecord<User>('user', '1'));

// immutable — rendered with `Saved: {{user.firstName}}`
const user = content.data;

// editable - rendered with `Edits: {{editable.firstName}}`
const editable = await checkout<EditableUser>(user);

// Works as expected
// Saved: Chris
// Edits: Christopher
editable.firstName = 'Christopher';

const saveInit = updateRecord(editable);
saveInit.body = JSON.stringify(serializePatch(store.cache, recordIdentifierFor(editable)));

await store.request(saveInit);                          // 200, server echoes 'Christopher'

// BUG!
// Saved: Chris
// Edits: Christopher
```

## Affected Cases

These cases are demonstrated in `tests/core/tests/polaris-mode/commit-reactivity-test.gts` — they fail on `main` and pass with this change, alongside a remote-push control test that passes on both.

1. **Your API returns the resource it just updated.** You edit a name in a form, save, and your API responds `200` with the updated user. Anything bound to the immutable record does not update.

2. **Your API returns `204 No Content`.** You edit a name in a form, save, and your API responds `204 No Content`. Anything bound to the immutable record does not update.

3. **Your API echoes back only some of what you sent.** You save three fields; the response includes only the one the server normalized. That one field updates on screen and the other two go stale.

4. **You use `commit()` directly instead of saving.** ([Documented as a non-recommended but still public API.](https://warp-drive.io/api/@warp-drive/core/reactive/functions/commit#commit)) `checkout()` → mutate → `commit()` applies local edits to the record without a round-trip — e.g. for state your API doesn't persist. Nothing rendering the immutable record updates.

Editing a nested field is affected the same way: `editable.messages[0].state = 'executed'` through a checkout, then `commit(editable)`, leaves the immutable record stale.

## Why it happens

Glossary, for anyone following along:

- **`remoteAttrs`** — the last-known-persisted values. What the immutable record reads.
- **`localAttrs`** — uncommitted edits made through a checkout. What the editable copy reads.
- **`inflightAttrs`** — edits moved out of `localAttrs` when a commit starts, usually  this is what's being saved (though if you call `commit` directly, that's not necessarily the case).

Under the hood:

`didCommit` merges everything saved into the persisted values, then notifies only `changedKeys`:

```js
const changedKeys = newCanonicalAttributes && calculateChangedKeys(cached, newCanonicalAttributes, fields);

cached.remoteAttrs = Object.assign(cached.remoteAttrs || {}, cached.inflightAttrs, newCanonicalAttributes);
//                                                           ^^^^^^^^^^^^^^^^^^^ all of it lands here
…
if (changedKeys?.size) notifyChange(identifier, 'attributes', changedKeys);   // but only these are announced
```

`changedKeys` comes from `calculateChangedKeys`, whose baseline already includes `inflightAttrs`:

```js
const original = Object.assign(Object.create(null), cached.remoteAttrs, cached.inflightAttrs);
```

So `changedKeys` answers **"did the server disagree with what I sent?"** — the right question for the editable copy, the wrong one for the immutable record, which needs **"did the persisted value change?"**

For the example above those answers diverge:

| question | answer |
| --- | --- |
| did the server disagree with what we sent? | no — it echoed `'Christopher'` back |
| did the persisted value change? | **yes — `'Chris'` → `'Christopher'`** |

`changedKeys` is empty, nothing is announced, and the immutable record stays on `'Chris'`.

**It does not happen when the server contradicts you.** Then `changedKeys` is non-empty and the existing notification fires — which is why a save that normalizes or overwrites your value has always worked.

## The fix

`collectPromotedRemoteKeys` collects the keys a commit moves into `remoteAttrs` that nothing else announced, and `didCommit` notifies them on the `'remote'` channel.

`'remote'` rather than unscoped because the editable copy's value doesn't move across a commit — it read `localAttrs`/`inflightAttrs` before and `remoteAttrs` after, the same value either way. Notifying it would be wasted work, not a correctness problem.

Two exclusions keep it from announcing more than it should:

- keys already in `changedKeys`, which are notified unscoped — a server-driven difference genuinely does change both views
- keys whose persisted value didn't actually change, so a no-op commit stays silent

This lives in `didCommit` because that's the one place that knows the save's values were just promoted. Fixing it there covers `commit()` and every response shape together, rather than patching `_COMMIT` (the internal `commit()` implementation) or individual call sites.

<details>
<summary><b>Why this surfaced now</b></summary>

Before notification channels, `setAttr` notified unscoped. The immutable record was invalidated by the *edit*, so its next render re-read the cache — and that render landed after the commit had resolved, so it happened to pick up the right value. The commit was never what notified it.

Once `setAttr` became `'local'`-tagged, that notification stopped reaching the immutable record, and the absence of one in `didCommit` became observable.

</details>

## Tests

Two layers.

**`tests/json-api/.../did-commit-notification-test.ts`** — 8 tests, 5 failing without the source change, asserting on the notifications themselves.

**`tests/core/tests/polaris-mode/commit-reactivity-test.gts`** — 5 tests, 4 failing without it, asserting on rendered output. This is the layer that speaks to a consumer rather than to cache internals, and it's the one the existing coverage was missing.

Each renders both projections side by side and asserts all three stages, so the tests state the intended behavior rather than only the failure:

| stage | `[data-test-saved]` (remote) | `[data-test-edits]` (local) |
| --- | --- | --- |
| after checkout | Chris | Chris |
| after the edit | Chris | Christopher |
| after commit | Christopher | Christopher |

The middle row is the deliberate buffering of local edits, not the bug — pinning it down stops a future change from "fixing" the third row by leaking local edits early. Pre-fix, the only failing assertion in each test is the third row; every stage-1 and stage-2 assertion already passes, which isolates the defect to exactly the claimed behavior.

One of the five covers a nested `schema-array` write (`messages[0].state`), which had no rendered coverage and fails pre-fix the same way.

Neither layer asserts a property read, and that distinction is why this went unnoticed. Outside a tracking frame a getter re-reads the cache on every access, and the cache is already correct — so `assert.strictEqual(user.firstName, 'Christopher')` passes with and without the bug. Only a rendered component or a notification assertion sees it.

Confirming the mechanism rather than assuming it: `ReactiveResource`'s notification handler has no `case 'state':` at all — it handles `identity`, `attributes`, and `relationships` only. So the `notifyChange(identifier, 'state', null)` that `didCommit` always fires cannot invalidate a field signal, and nothing else was rescuing the render.

<details>
<summary><b>Test design, and a related note on an existing test</b></summary>

Each test watches three subscribers so the whole delivery picture is visible rather than half of it: `remote`; `local` from the start, which also catches the `setAttr` from the edit; and `local` subscribed after the edit, which sees only what the commit or save itself emits.

Three of the eight pass either way and exist to pin down behavior a later change shouldn't break — a no-op commit stays silent, `commit()` announces nothing to the editable copy, and a server-contradicting response reaches each channel once rather than twice.

On that last one: the exclusion of `changedKeys` keys is not what prevents double-delivery. I removed the exclusion and all 108 tests still pass, because the notification buffer coalesces by key within a flush window — two notifications for one key merge into a single delivery carrying both channels. The exclusion avoids putting a redundant channel tag in the buffer; it isn't load-bearing for correctness.

Two existing tests are worth a look, because the same blind spot hides the bug in both:

**`tests/json-api/.../mutation-request-test.ts:305`** — "a local edit matching the persisted server response still notifies the remote channel" — already describes this bug in its comments, including that `changedKeys` never sees a difference. But it asserts `user?.firstName` rather than the notification, so it passes before and after.

**`tests/core/tests/edit-workflow-test.ts:50`** — its handler echoes the request body straight back, which is exactly the verbatim-echo case, and it asserts `user.name` updated "in the cache and shows in the immutable record". Same property read, same false pass.

I left both alone rather than widen this PR. Happy to convert them to notification assertions here or separately, whichever you'd rather.

</details>

<details>
<summary><b>Verification</b></summary>

- `tests/json-api` — 102 pass / 0 fail / 6 todo, against 97 pass / 5 fail without the source change
- `tests/core`, `tests/legacy`, `ember-data__graph`, `utilities`, `ember-data__model` — all green
- `check:types` clean on `warp-drive-packages/json-api` and `tests/json-api`; `lint:oxfmt` and `lint:oxlint` clean

</details>

## Open questions

- **`relationships` may have the same gap** via `inflightRelationships`, which `didCommit` clears without diffing. I didn't investigate and this change doesn't touch it. Worth a follow-up issue if you agree it's plausible.
- Anything shipped on `5.10.0-alpha.x` has this behavior, so it may warrant a patch release rather than waiting.

<sub>**AI attribution:** :robot: _Claude using Opus 5. Used skills including `gitkrystan-claude-plugin:doc-coauthoring`, `satl-pr-description`, and `gitkrystan-claude-plugin:comment-triage`. The code change has been reviewed by the author._</sub>


warp-drive-packages/json-api/src/-private/cache.ts
```diff
@@ -1826,6 +1826,36 @@ function getDefaultValue(
...
 
+/**
+ * The keys a commit promotes from `inflightAttrs` into `remoteAttrs` that the remote projection
+ * has not otherwise been told about: saved keys the server echoed verbatim or omitted from its
+ * response. Keys already in `changedKeys` are excluded because `didCommit` notifies those
+ * unscoped, and keys whose remote value is unchanged are excluded so a no-op commit stays silent.
+ *
+ * Must be called before `remoteAttrs` is merged, while it still holds the pre-commit values.
+ */
+function collectPromotedRemoteKeys(
```
gitKrystan: Two decisions worth stating here rather than leaving to the diff.

**Why this is a separate pass instead of folding the keys into `changedKeys`.** The two sets want different channels. A server-driven difference changes both projections, so `changedKeys` is correctly notified unscoped; a promoted key changes only the remote projection. Merging them would force one channel on both and wake local subscribers whose value never moved.

**Why it runs before the `remoteAttrs` merge.** The comparison needs the pre-commit remote values, and the merge on the next line destroys them. That ordering is load-bearing and easy to break by moving this call for readability — hence the note in the JSDoc.

**On the `changedKeys` exclusion specifically:** it is not what prevents double-delivery. I removed it and all 108 tests still pass, because the notification buffer coalesces by key within a flush window — two notifications for one key merge into a single delivery carrying both channels. The exclusion avoids putting a redundant channel tag into that buffer. I kept it for intent and to skip needless work, but I don't want it read as a correctness guarantee it isn't providing. Happy to drop it if you'd rather lean on the coalescing.

<sub>**AI attribution:** :robot: _Claude using Opus 5. Used skills including `gitkrystan-claude-plugin:comment-triage`. The code change has been reviewed by the author._</sub>

tests/core/tests/polaris-mode/commit-reactivity-test.ts
```diff
@@ -0,0 +1,131 @@
...
+  async function renderUser(store: ReturnType<typeof setup>) {
+    const resource = store.schema.resource({ type: 'user' });
+    const user = store.push({
+      data: { type: 'user', id: '1', attributes: { firstName: 'Chris' } },
+    }) as User;
+    await reactiveContext(user, resource);
+    return user;
+  }
+
+  test('control: a remote push updates the DOM', async function (assert) {
```
gitKrystan: This one passes on both main and this branch. The others fail on main and pass on this branch.

tests/json-api/tests/integration/cache/did-commit-notification-test.ts
```diff
@@ -0,0 +1,321 @@
...
+            { name: 'id', kind: 'field' },
+            { name: 'state', kind: 'field' },
+          ],
+        },
+      ],
+    });
+    this.store = new TestStore();
+  });
+
+  test<CustomContext>('control: a remote push reaches a remote subscriber', function (assert) {
```
gitKrystan, line 96: `control` = passes on both main and this branch. The others fail on main and pass on this branch.

warp-drive-packages/json-api/src/-private/cache.ts
```diff
@@ -2493,6 +2523,11 @@ function didCommit(
     newCanonicalAttributes = data.attributes;
   }
   const changedKeys = newCanonicalAttributes && calculateChangedKeys(cached, newCanonicalAttributes, fields);
+  // `calculateChangedKeys` baselines against `inflightAttrs`, so keys the server echoed
+  // verbatim or omitted are absent from `changedKeys` -- yet they do move into `remoteAttrs`
+  // below. Tagged 'remote', not unscoped: the local projection already read these values, so
+  // its value does not move across the commit.
+  const promotedKeys = collectPromotedRemoteKeys(cached, changedKeys, fields);
```
runspired: likely the correct fix is to calculateChangedKeys against both local and remote, probably as a set intersection: local-only, remote-only, both

gitKrystan: makes sense

gitKrystan: Claude took a stab at this. I'm reviewing it now.

warp-drive-packages/json-api/src/-private/cache.ts
```diff
@@ -1826,49 +1826,141 @@ function getDefaultValue(
...
-  const localAttrs = cached.localAttrs;
-
-  const original: Record<string, unknown> = Object.assign(
-    Object.create(null) as Record<string, unknown>,
-    cached.remoteAttrs,
-    cached.inflightAttrs
-  );
+  updates: ExistingResourceObject['attributes'],
+  fields: ReturnType<Store['schema']['fields']>,
+  promoted: Record<string, unknown> | null
```
gitKrystan: should the order be: cached, updates, promoted, fields ?

warp-drive-packages/json-api/src/-private/cache.ts
```diff
@@ -1826,49 +1826,152 @@ function getDefaultValue(
...
+ * projection and says nothing about the remote one, which is how a save whose response echoed our
+ * own values back could leave an immutable record rendering the pre-save value: the value moved
+ * into `remoteAttrs`, and nothing reported it as changed.
+ *
+ * So partition instead. Each bucket maps onto the channel that should carry it — `localOnly` and
+ * `remoteOnly` to their own channels, `both` unscoped — and `remoteChanged` is the union the
+ * `remoteAttrs` reconciliation in {@link patchLocalAttributes} needs.
+ *
+ * Buckets are `undefined` rather than empty Sets: this runs on every `upsert`, where the usual
+ * outcome is that nothing moved, and an unchanged resource should cost no allocation.
```
gitKrystan, line 1861: This is TMI for people like @runspired that have this codebase memorized

warp-drive-packages/json-api/src/-private/cache.ts
```diff
@@ -1826,49 +1826,152 @@ function getDefaultValue(
...
+
+/**
+ * Partition the keys an incoming payload touches by which projection each one moves.
+ *
+ * `promoted` is the `inflightAttrs` a commit is about to merge into `remoteAttrs`, or `null` for
+ * an upsert that merges only `updates`. It has to be passed rather than read off `cached` because
+ * the two callers merge different things, and the post-merge remote value is what decides the
+ * partition.
+ *
+ * Must be called before `remoteAttrs` is merged, while it still holds the pre-merge values.
```
gitKrystan, line 1889: This seems brittle. Is there a way to avoid this?

warp-drive-packages/json-api/src/-private/cache.ts
```diff
@@ -2187,7 +2276,16 @@ function cacheUpsert(
   // if no cache entry existed, no record exists / property has been accessed
   // and thus we do not need to notify changes to any properties.
   if (calculateChanges && existed && data.attributes) {
-    changedKeys = calculateChangedKeys(cached, data.attributes, fields);
+    // an upsert merges only `updates`, so nothing is being promoted. Channel-scoping this path
+    // off the partition is a separate change -- for now it keeps announcing the union unscoped,
+    // exactly as before.
+    changedKeys = calculateChangedKeys(
```
gitKrystan, line 2282: Does calculateChangedKeys end up doing anything unnecessary in this invocation since we only need `remoteChanged` from it?

warp-drive-packages/json-api/src/-private/cache.ts
gitKrystan: Definitely got harder for me to understand which probably means you'll like it?

warp-drive-packages/json-api/src/-private/cache.ts
```diff
@@ -2492,22 +2590,41 @@ function didCommit(
...
   if (cached.errors) {
     cached.errors = null;
     cache._capabilities.notifyChange(identifier, 'errors', null);
   }
 
-  if (changedKeys?.size) cache._capabilities.notifyChange(identifier, 'attributes', changedKeys);
+  // each bucket goes out on the channel it actually moved. Pruning only happens for a resource
+  // that had local edits the server confirmed, and `patchLocalAttributes` reports whether it
+  // removed anything -- so the re-filter is skipped entirely on the common path.
+  const { remoteChanged, localOnly } = changes;
```
runspired, line 2620: I'm fairly sure there's inefficiency here where the changeset is already known and doesn't need recalculated. That said - make it work is more important than make it fast.

runspired (approved):

PR #11142, opened 2026-09-21
feat: implement resource and collection relationship field kinds

runspired:
Ships the `resource` and `collection` schema field kinds for `ReactiveResource` in both LegacyMode and PolarisMode. Until now both kinds were declared in the types but stubbed in the reactive layer (`resource` returned an ad-hoc object with no reactivity, `collection` asserted on access). PolarisMode relationships were only possible through the legacy `belongsTo`/`hasMany` kinds in LinksMode.

### Relationship documents

The value of a `resource` or `collection` field is a `ReactiveRelationshipDocument<T>` (new public type exported from `@warp-drive/core/reactive`) whose `data`, `links` and `meta` mirror the relationship payload in the cache:

```ts
user.bestFriend.data;  // User | null | undefined
user.bestFriend.links; // { related: '/users/1/best-friend' }
user.friends.data;     // User[] | undefined (a reactive array)
```

- `data` is `undefined` when the payload carried no `data` member (links-only payload), `null` / `[]` when the relationship is known to be empty.
- The document and (for collections) its `data` array are stable instances. Their properties are gates that recompute from the cache when the relationship is notified; the array re-syncs through its manager like the LinksMode `ManyArray` does.
- Immutable PolarisMode records render remote state (`cache.getRemoteRelationship`); LegacyMode records and checked-out PolarisMode records render local state (`cache.getRelationship`), matching the existing `belongsTo`/`hasMany` LinksMode split.
- `doc.fetch()` requests the `related` (falling back to `self`) link as a normal top-level request.

### Payload constraints that `belongsTo`/`hasMany` do not have

These are collected in a new normative **[Relationship Specification](https://canary.warp-drive.io/pr-preview/pr-11142/guides/the-manual/relational-data/spec)** page in the manual.

1. **Full inclusion.** Every resource referenced in a `resource`/`collection` relationship's `data` MUST be included in the payload that references it. The JSON:API validator reports a missing resource as an error (with a kind-specific message; the generic 7.4 full-linkage check skips these fields to avoid double-reporting), and materializing `data`/`remoteData` for a member with no data in the cache **throws** instead of producing an empty record.
2. **`async` is a contract about the payload.**
   - `async: true`: every payload MUST carry `links.related`. Enforced by the validator (error) and by a graph assertion on every payload, including `store.push`.
   - `async: false` (default): whenever the relationship is present its `data` MUST be present, so links-only or meta-only payloads are a validator error; and it SHOULD NOT carry `links`, which the validator reports as a warning promotable to an error via the new `strict.syncRelationshipLinks` reporter setting.

### Mutation

Relationships are mutated through the document's `data`, never by assigning the field:

```ts
editable.bestFriend.data = otherUser;   // or null → replaceRelatedRecord
editable.friends.data.push(a);          // push/pop/unshift/shift/splice/sort/index assignment
editable.friends.data = [a, b];         // → replaceRelatedRecords
editable.bestFriend = otherUser;        // ❌ asserts in every mode
```

Array mutations are translated into granular graph ops (`add`, `remove`, `replaceRelatedRecords`). Duplicates throw. On an immutable PolarisMode record every setter asserts. Inverses are updated per the usual local/remote split.

### Server-owned `links`/`meta`, plus `remoteData` and `isDirty`

Only `data` is mutable. `links` and `meta` are server-owned and always reflect the last payload received from the API, on editable and immutable records alike. That is aligned with JSON:API, but it means `meta` derived from membership (e.g. `count`) drifts from `data` while a relationship has unsaved changes. To make the drift visible the document also exposes:

```ts
editable.friends.data.push(a);
editable.friends.isDirty;           // true until the change is confirmed by the API
editable.friends.remoteData.length; // membership the API last confirmed (read-only reactive array)
editable.friends.meta.count;        // still describes remoteData, not data
```

- `remoteData` is the membership the API last confirmed. On an immutable record it is the same value as `data`.
- `isDirty` reports whether the relationship has unconfirmed local changes (via `cache.changedRelationships`). It is a property of the relationship, so it reads the same on the immutable record and its checked-out copy.
- Notification callbacks now receive the resolved channel (`'local'`, `'remote'`, or `undefined` for unscoped/mixed) as a fourth argument. Relationship documents use it to leave `links`, `meta` and `remoteData` untouched on purely local notifications, so their identity is preserved and consumers of them do not recompute for edits that cannot have changed them.

### `createRecord` and the JSON:API cache

- `normalizeProperties` converts `resource`/`collection` create args to `ResourceKey`s like it does for `belongsTo`/`hasMany`.
- `clientDidCreate` seeds the graph for the new kinds instead of passing the values through as plain props (which would have hit the asserting field setter).
- The DEBUG in-flight relationship bookkeeping also covers `resource`.

### Schema and types

- `ResourceField | CollectionField` added to `LegacyModeFieldSchema` and `PolarisModeFieldSchema`; the "NOT YET IMPLEMENTED" notes are replaced with real docs, including the `async` contract above.
- `SchemaService.registerResource` treats the new kinds as relationships for abstract-type (`options.as`) registration. The deprecated legacy `relationships` map continues to hold only `belongsTo`/`hasMany`.
- `CreateRecordProperties` unwraps `ReactiveRelationshipDocument<T>` to `T` so `store.createRecord('user', { bestFriend, friends })` type-checks against document-typed interfaces.
- `UpgradedMeta` records the original `fieldKind` alongside the graph-internal `kind`.

### `maxCollectionRelationshipSize`

New opt-in guard configured via `useRecommendedStore({ maxCollectionRelationshipSize: 100 })` or as a `Store` property (default `null`, off). When a remote `collection` payload's `data` array is longer than the limit, the payload is still applied and an error is thrown asynchronously (from a microtask, in all builds) so it reaches `window.onerror`/error reporters and fails tests without leaving the cache half-updated. It is a per-payload member count, applies only to `collection` fields (not legacy `hasMany`), and does not fire for local mutations. The error message points at the new large-collections guide.

### Documentation

- New **Relationship Specification** page (`guides/the-manual/relational-data/spec.md`): normative MUST/SHOULD rules for payload shape, the strict-kind constraints and their enforcement (validator / graph / runtime), the legacy kinds, values and mutation, and the size guard.
- Fills in the relational-data outline in the manual: resource relationships, collection relationships, links vs identifiers (now a per-`async` validity table), sync vs async (rewritten around the contract), adding & removing, saving, sorting & filtering, the graph, pagination (via top-level requests, `ReactiveDocument.next()`, `<Paginate>`), large collections (the size guard plus a per-`async` payload table), and directionality. Also fills in `schemas/relational-fields.md`, fixes the schemas overview example, and updates the PolarisMode limitation note and the caching aside about relationship pagination.
- The guides state that `links`/`meta` are server-owned, explain the drift while editing and how `isDirty`/`remoteData` expose it, and establish the serialization rule: request builders send only the identifiers of `data`, never `links` or `meta`.

### Tests

- `tests/core` PolarisMode: read, reactivity, write (checkout, inverses, payload confirmation, createRecord, `remoteData`/`isDirty`, link and meta identity across local edits), size-limit, materialization-error and async-link-assertion tests for both kinds; the previously skipped resource reactivity test is enabled.
- `tests/legacy` LegacyMode: read and edit tests, including a `collection` kept in sync with a legacy `belongsTo` inverse.
- `tests/json-api` validator: the new per-field rules (valid sync, missing inclusion, sync with links → warning, sync without data → error, async without related link → error, async with link with/without data, async data still requires inclusion).

## Worth knowing

- **`commit()` does not promote relationship local state.** `didCommit` without a payload only touches attributes; relationships are confirmed by the response payload (`setupRelationships`). This is pre-existing behavior for every relationship kind, so the tests confirm changes via a pushed payload, mirroring a save response.
- **Local `sort()` is a replace.** The graph has no local `sortRelatedRecords` operation, so sorting a collection's `data` issues `replaceRelatedRecords` with the new order, which the graph diff preserves.
- **`fetch()` does not write back.** The response is a top-level document; it is not merged into the relationship's `data`. There is no autofetch for `async: true`.
- **Pagination is descoped.** `next`/`prev`/`first`/`last` links on a relationship are surfaced on `doc.links` but never merged into membership. The docs steer paginated, sorted or filtered lists to top-level requests.
- **Where each constraint is enforced.** The validator only runs in development builds when `JSON_API_CACHE_VALIDATION_ERRORS` (or `LOG_CACHE`) is active and only for request documents, not `store.push`. The async-link rule is additionally a dev-build graph assertion on every payload. The full-inclusion rule is additionally a runtime throw in every build when the relationship is materialized. `strict.*` reporter settings, including the new `syncRelationshipLinks`, are still internal defaults (pre-existing TODO to expose them to apps).
- **Stale `meta` after a save that omits it.** Relationship payloads replace `links`/`meta` per key, so if the save response does not echo `meta`, the pre-save value persists. `isDirty` goes back to `false` once the response confirms `data`.
- **`NotificationCallback` gained an optional fourth `channel` argument.** Additive and backwards compatible; existing subscribers ignore it.
- **LinksMode validation is not applied** to the new kinds: they have their own, stricter rules above and no `linksMode` option.
- **Legacy adapter/serializer/snapshot paths** still only understand `belongsTo`/`hasMany`; `resource`/`collection` in LegacyMode are reactive-only and should be saved through request builders.
- `resetOnRemoteUpdate` is forced to `false` for the new kinds by the existing `upgradeMeta` conversion, so remote updates never discard local changes.

## Verification

| Suite | Result |
| --- | --- |
| `tests/core` | 239 passed, 1 pre-existing skip |
| `tests/legacy` | 141 passed, 1 skip |
| `tests/json-api` | 111 passed |
| `tests/ember-data__graph` | 143 passed |

`check:types` passes for `@warp-drive/core`, `@warp-drive/json-api`, `@warp-drive/legacy` and the core, legacy and json-api test apps; oxlint, oxfmt and eslint are clean.

https://claude.ai/code/session_019oXtgKwnvLbBmgibREnpWF


warp-drive-packages/core/src/reactive/-private/schema.ts
```diff
@@ -855,7 +873,9 @@ export class SchemaService implements SchemaServiceInterface {
         const ownField = fields.get(name);
         if (!ownField) {
           fields.set(name, contribution.field);
-          relationships[name] = contribution.field;
+          if (isLegacyRelationshipField(contribution.field)) {
```
runspired: why is this only something we need to do for legacy relationships

runspired: Because the `relationships` map is only the Model-era view of the schema. Its type is `Record<string, LegacyRelationshipField>` (`InternalSchema.relationships`, line 639) and its only reader is the deprecated `relationshipsDefinitionFor`, which returns it verbatim (line 1189); everything non-deprecated goes through `schema.fields()`, which does include the new kinds. Widening the map would push `resource`/`collection` fields into the deprecated adapter/serializer/snapshot paths that consume `relationshipsDefinitionFor` and only know how to handle `belongsTo`/`hasMany`, so I kept it to the kinds those callers can handle. Same reasoning for the abstract-type contribution branch here and in `_registerAbstractTypeImplementation`.

If you'd rather the deprecated API expose the new kinds as well, it is a two-line change (widen the type to `RelationshipField` and drop the three guards); say so and I'll push it.

---

warp-drive-packages/core/src/store/-private/managers/notification-manager.ts
```diff
@@ -94,7 +94,19 @@ const UnscopedChannel: '___(unique) Symbol(UnscopedChannel)' = getOrSetGlobal('U
...
+   * during this flush: `'local'` or `'remote'` when all touches agreed, and
+   * `undefined` when any touch was unscoped or the touches disagreed (in which
+   * case the subscriber should assume both projections may have changed).
+   */
+  (
+    cacheKey: ResourceKey,
+    notificationType: 'attributes' | 'relationships',
+    key?: string,
+    channel?: NotificationChannel
+  ): void;
```
runspired: why do we need this change? aren't the existing legacy relationship implementations already using dual channel?

runspired: They are, but dual channel today only decides *which subscribers* a notification reaches, not *what kind* of change it was. A `'local'` subscriber (every editable record) receives both `'local'`-tagged notifications and unscoped ones (`shouldDeliverToChannel`, line 234: unscoped reaches everyone, and remote pushes notify unscoped). That is enough for `belongsTo`/`hasMany`, which have a single signal per field and re-pull the whole value on either kind of change.

The relationship document is one object with two projections. `data` and `isDirty` follow local state, while `links`, `meta` and `remoteData` are server-owned and only change on a remote update. Without knowing the channel, an editable document would have to stale all five on every notification, so a `push()` into a collection would invalidate `links` and `meta` and hand consumers new frozen objects for values that cannot have changed. That is the drift the earlier `remoteData`/`isDirty` review wanted to avoid. With the channel the document skips the server-owned three on a purely `'local'` notification (`notifyRelationshipDocument` in #11217). Immutable documents subscribe `'remote'` and never hit this.

It is additive and existing subscribers ignore it. The alternative without an API change is a second, `'remote'`-channel subscription per editable record that stales only the server-owned properties (unscoped pushes would reach both subscriptions, local mutations only the `'local'` one). That works too, at the cost of a second callback and dispatch per record. Happy to switch to it if you prefer keeping the callback signature as is.

---

warp-drive-packages/core/src/store/-private/store-service.ts
```diff
@@ -2536,12 +2536,12 @@ function normalizeProperties(
 
         if (!field) continue;
 
-        if (field.kind === 'hasMany') {
+        if (field.kind === 'hasMany' || field.kind === 'collection') {
```
runspired: should we consider moving to a util like `isManyKind` `isSingleKind`?

runspired: Done in e4e8361: `isManyKind`, `isSingleKind` and `isRelationshipKind` now live in `@warp-drive/core/types/schema/fields` next to the existing `isResourceSchema`/`isLegacyResourceSchema` helpers, so core and the JSON:API cache share one definition. Used here in `normalizeProperties`, in the schema service's abstract-implementer check, in the graph's `isRelationshipField` (which replaces the `RELATIONSHIP_KINDS` array), and in the cache's `isRelationship`, `willCommit` and `didCommit`. The `switch` cases in `clientDidCreate` stay as cases. Types, lint and the core, legacy and json-api suites pass locally; the rest of the stack is rebased onto it.

---

